PR #443, opened 2019-11-19
Remove orphan lock file

tamilmani1989:
This PR make sure that  CNI process is not blocked bacause of orphaned lock file. The cni will remove lock file if the process that is holding lockfile exited

**Release note**:

```release-note
```

### DIFF
--- a/cni/ipam/plugin/main.go
+++ b/cni/ipam/plugin/main.go
@@ -28,11 +28,19 @@ func main() {
 
 	if err := ipamPlugin.Plugin.InitializeKeyValueStore(&config); err != nil {
 		fmt.Printf("Failed to initialize key-value store of ipam plugin, err:%v.\n", err)
+
+		if ipamPlugin.Plugin.IsSafeToRemoveLock(ipamPlugin.Plugin.Name) {
+			log.Printf("[CNI] Removing lock file as process holding lock exited.")
+			if errUninit := ipamPlugin.Plugin.UninitializeKeyValueStore(true); errUninit != nil {
+				log.Errorf("Failed to uninitialize key-value store of network plugin, err:%v.\n", errUninit)
+			}
+		}
+
 		os.Exit(1)
 	}
 
 	defer func() {
-		if errUninit := ipamPlugin.Plugin.UninitializeKeyValueStore(); errUninit != nil {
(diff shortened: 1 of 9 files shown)


cni/ipam/plugin/main.go
```diff
@@ -28,11 +28,19 @@ func main() {
 
 	if err := ipamPlugin.Plugin.InitializeKeyValueStore(&config); err != nil {
 		fmt.Printf("Failed to initialize key-value store of ipam plugin, err:%v.\n", err)
+
+		if ipamPlugin.Plugin.IsSafeToRemoveLock(ipamPlugin.Plugin.Name) {
```
jaer-tsun: duplicate code, maybe make this a func

cni/plugin.go
```diff
@@ -186,9 +188,9 @@ func (plugin *Plugin) InitializeKeyValueStore(config *common.PluginConfig) error
 }
 
 // Uninitialize key-value store
-func (plugin *Plugin) UninitializeKeyValueStore() error {
+func (plugin *Plugin) UninitializeKeyValueStore(forceUnlock bool) error {
```
jaer-tsun: force is good enough

cni/plugin.go
```diff
@@ -198,3 +200,37 @@ func (plugin *Plugin) UninitializeKeyValueStore() error {
...
+			return false
+		}
+
+		if len(content) <= 0 {
+			log.Errorf("Num bytes read from lock file is 0")
+			return false
+		}
+
+		pid := string(content)
+		pid = strings.Trim(pid, "\n")
```
jaer-tsun: pid := strings.Trim(string(content), "\n") seems cleaner

cni/plugin.go
```diff
@@ -198,3 +200,37 @@ func (plugin *Plugin) UninitializeKeyValueStore() error {
 
 	return nil
 }
+
+// check if safe to remove lockfile
+func (plugin *Plugin) IsSafeToRemoveLock(processName string) bool {
+	if plugin.Store != nil {
```
jaer-tsun: you may need to check if plugin is nil as well

cni/ipam/plugin/main.go
tamilmani1989: cannot do this..will explain offline

PR #449, opened 2019-11-26
Fix NPM Regression & Remove TelemetryBuffer Sidecar

jaer-tsun:
- give precedence to drop rules (over allow)
- remove telemetry buffer from NPM (removed sidecar as well)
- write logs to stdout (and file) so that we may see logs using 'kubectl logs...'
- remove kube-system chain
- add drop entries in corresponding chain for specific policies (i.e. non ALLOW-ALL* entries)


tamilmani1989:
@jaer-tsun can you create separate PRs for telemetry change and drop entry precedence

PR #451, opened 2019-12-02
Join vnet as a part of NC publish

ashvindeodhar:
Join vnet before publishing NC. This is required as nmagent may be
restarted which causes the NC publishing to fail.

**Release note**:

```release-note
```


cns/restserver/restserver.go
```diff
@@ -113,6 +113,7 @@ func NewHTTPRestService(config *common.ServiceConfig) (HTTPService, error) {
 
 	serviceState := &httpRestServiceState{}
 	serviceState.Networks = make(map[string]*networkInfo)
+	serviceState.joinedNetworks = make(map[string]struct{})
```
jaer-tsun: caps J for joinedNetworks

PR #455, opened 2019-12-06
AITelemetry support for CNS

tamilmani1989:
This PR adds AI Telemetry support for CNS component

**Release note**:

```release-note
```


cns/service/main.go
```diff
@@ -271,29 +270,26 @@ func main() {
...
-
 		err = httpRestService.Start(&config)
 		if err != nil {
-			log.Errorf("Failed to start CNS, err:%v.\n", err)
+			logger.Errorf("Failed to start CNS, err:%v.\n", err)
 			return
 		}
 	}
 
+	if !cnsconfig.TelemetrySettings.DisableAll {
```
ashvindeodhar: can this be moved to else on line 237?

tamilmani1989: this can't be in else

cns/logger/heartbeat.go
```diff
@@ -1,28 +1,91 @@
-// Copyright Microsoft. All rights reserved.
+// Copyright 2018 Microsoft. All rights reserved.
```
ashvindeodhar: I think we are not adding the year. That would need us to update those every year ( when we touch those files )

PR #461, opened 2019-12-17
Prevent Namespace Race

jaer-tsun:
Adding namespace to npMgr if it doesn't exist. This fixes reboot bug where entries don't get added after failing to append non-existing namespace to all-namespaces.

When NPM reboots, namespace details do not always sync before policies and pods do. In those cases the namespace will need to be added before we proceed with add/update pod/networkpolicy workflow.

### DIFF
--- a/npm/namespace.go
+++ b/npm/namespace.go
@@ -86,14 +86,8 @@ func (npMgr *NetworkPolicyManager) UninitAllNsList() error {
 	return nil
 }
 
-// AddNamespace handles adding namespace to ipset; if reEntrant is set to true, this func will
-// not acquire npMgr lock.
-func (npMgr *NetworkPolicyManager) AddNamespace(nsObj *corev1.Namespace, reEntrant bool) error {
-	if !reEntrant {
-		npMgr.Lock()
-		defer npMgr.Unlock()
-	}
-
+// AddNamespace handles adding namespace to ipset
+func (npMgr *NetworkPolicyManager) AddNamespace(nsObj *corev1.Namespace) error {
 	var err error
 
 	nsName, nsLabel := "ns-" + nsObj.ObjectMeta.Name, nsObj.ObjectMeta.Labels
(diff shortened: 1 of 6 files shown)


npm/namespace.go
```diff
@@ -135,6 +132,14 @@ func (npMgr *NetworkPolicyManager) AddNamespace(nsObj *corev1.Namespace) error {
 	return nil
 }
 
+// AddNamespaceWithLock acquires NetworkPolicyManager lock before adding namespace to ipset
```
saiyan86: I think a cleaner way to do this is to unlock & lock in the caller instead of having this function.

npm/pod.go
```diff
@@ -35,6 +36,20 @@ func (npMgr *NetworkPolicyManager) AddPod(podObj *corev1.Pod) error {
...
+	// Add pod namespace if it doesn't exist
+	if _, exists := npMgr.nsMap[podNs]; !exists {
+		nsObj := &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   podObj.ObjectMeta.Namespace,
+				Labels: make(map[string]string),
+			},
+		}
+		
+		if err = npMgr.AddNamespace(nsObj); err != nil {
```
saiyan86: Do you know why it worked without this change previously? The functions are locked so there shouldn't be any race conditions...

npm/namespace.go
matmerr: https://github.com/Azure/azure-container-networking/pull/461/files/63dd9d5784f75c11d09d573d5274d90a773bd646#diff-88c55b4ac2573a453141838261b6b413R91

AddNamespace used to take a bool to determine that, but having every signature with a bool seemed a bit like an antipattern, but it can be changed back

npm/pod.go
jaer-tsun: it's only when they have a namespace that hasn't been registered yet

so if sharedinformercache triggers the addnetworkpolicy or addnode call before addnamespace, there will be a failure (on reboot)

npm/namespace.go
jaer-tsun: That was the patter I had it in before. And I don't feel strongly about either to change it back and forth

PR #468, opened 2019-12-21
Merge Multiple Network Policies

### LINKED ISSUE #465: azure-npm networking policies are not compatible with Kubernetes Network Policy
**Is this a request for help?**: NO

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**Which release version?**: v1.0.30

---

**Which component (CNI/IPAM/CNM/CNS)**: NPM

---

**Which Operating System (Linux/Windows)**: Linux

---

Linux aks-nodepool1-XXXXXXX-vmss000000 4.15.0-1061-azure #66-Ubuntu SMP Thu Oct 3 02:00:50 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

---

Kubernetes 1.14.8

---

**What happened**:

To make things simple I am providing a minimal subset of our networking policies. First, we apply a restrictive set of networking policies and then allow some internal communication along with egress to specified out of cluster IPs, etc.

The following network policy to block all inbound and outbound traffic.

```yaml
kind: NetworkPolicy
metadata:
  name: deny-all-ingress-egress
  namespace: our-namespace
spec:
  podSelector: {}
  policyTypes:
  - Ingress
  - Egress
```

then we allow DNS

```yaml
kind: NetworkPolicy
metadata:
  name: dns
  namespace: our-namespace
spec:
  podSelector: {}
  egress:
  - to:
    - namespaceSelector: {}
    ports:
    - protocol: UDP
      port: 53
```

this is fine until `azure-npm` starts to diverge from the other networking policy managers (which is not mentioned in [the documentation](https://docs.microsoft.com/en-us/azure/aks/use-network-policies))

if we apply the following policy inbound traffic remains blocked

```yaml
kind: NetworkPolicy
metadata:
  name: allow-nginx-ingress
  namespace: our-namespace
spec:
  podSelector:
    matchLabels:
      app: nginx-ingress
  ingress:
  - {}
```

to make it work, we had to change this policy to this

```yaml
kind: NetworkPolicy
metadata:
  name: allow-nginx-ingress
  namespace: our-namespace
spec:
  podSelector:
    matchLabels:
      app: nginx-ingress
  ingress:
  - from: []
    ports:
    - protocol: TCP
```

We had to update all of our policies to define the protocol in all of them. However, even we do so, things start to fall over after a while and we couldn't find the reason and decided to switch another networking policy manager.

---

I expect `azure-npm` to be compatible with [Kubernetes Network Policy](https://kubernetes.io/docs/concepts/services-networking/network-policies/)

---

You can apply the above networking policies to reproduce it. If you need a more precise example I can provide it by creating a new repository

---

We already have an open ticket on Azure Portal about `azure-npm` and `azure-ip-masq-agent`.

---

jaer-tsun:
Union between multiple network policies was broken because current implementation only scopes per policy. This PR will prepend all ipt entries except for chain jumps to prioritize the union of allows.

fixes #465 along with https://github.com/Azure/azure-container-networking/pull/466

### DIFF
--- a/npm/iptm/iptm.go
+++ b/npm/iptm/iptm.go
@@ -32,6 +32,7 @@ type IptEntry struct {
 	Chain                 string
 	Flag                  string
 	LockWaitTimeInSeconds string
+	IsJumpEntry           bool
 	Specs                 []string
 }
 
@@ -300,7 +301,11 @@ func (iptMgr *IptablesManager) Add(entry *IptEntry) error {
 		return nil
 	}
 
-	iptMgr.OperationFlag = util.IptablesAppendFlag
+	if entry.IsJumpEntry {
+		iptMgr.OperationFlag = util.IptablesAppendFlag
+	} else {
+		iptMgr.OperationFlag = util.IptablesInsertionFlag
+	}
 	if _, err := iptMgr.Run(entry); err != nil {
 		log.Errorf("Error: failed to create iptables rules.")
 		return err
--- a/npm/translatePolicy.go
+++ b/npm/translatePolicy.go
@@ -239,31 +239,6 @@ func translateIngress(ns string, targetSelector metav1.LabelSelector, rules []ne
 		for _, fromRule := range rule.From {
 			// Handle IPBlock field of NetworkPolicyPeer
 			if fromRule.IPBlock != nil {
-				if len(fromRule.IPBlock.Except) > 0 {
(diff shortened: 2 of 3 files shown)


jaer-tsun:
may/will swap out the contains(..., <partial jump chain comment>) and add a flag in IptEntry struct

PR #470, opened 2020-01-03
Added config option for disabling iptable lock

tamilmani1989:
Added config option for disabling iptable lock. By default, iptable lock is enabled. If users want to disable iptable lock, they have to set disableIPTableLock field to true in conflist.

**Release note**:

```release-note
```


tamilmani1989:
> In my opinion, we should update the conflist also. But it is upto you. I understand that it is workaround for short time, but change looks incomplete.

By default iptable lock is enabled. Even if we specify this in conflist, they have to manually flip the flag  before rollout. I don't want to expose this to user to disable it unless there is a need.

PR #471, opened 2020-01-03
Add build and publish image capabilities for cns

jaer-tsun:
This is just for testing purposes; (runners)

**Release note**:

```release-note
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,7 @@ CNS_ARCHIVE_NAME = azure-cns-$(GOOS)-$(GOARCH)-$(VERSION).$(ARCHIVE_EXT)
 NPM_ARCHIVE_NAME = azure-npm-$(GOOS)-$(GOARCH)-$(VERSION).$(ARCHIVE_EXT)
 NPM_IMAGE_ARCHIVE_NAME = azure-npm-$(GOOS)-$(GOARCH)-$(VERSION).$(ARCHIVE_EXT)
 TELEMETRY_IMAGE_ARCHIVE_NAME = azure-vnet-telemetry-$(GOOS)-$(GOARCH)-$(VERSION).$(ARCHIVE_EXT)
+CNS_IMAGE_ARCHIVE_NAME = azure-cns-$(GOOS)-$(GOARCH)-$(VERSION).$(ARCHIVE_EXT)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -267,6 +271,23 @@ azure-vnet-telemetry-image: azure-vnet-telemetry
 publish-azure-vnet-telemetry-image:
 	docker push $(AZURE_VNET_TELEMETRY_IMAGE):$(VERSION)
 
+# Build the Azure CNS image.
+.PHONY: azure-npm-image
```
tamilmani1989: is this azure-cns-image or npm?

cns/Dockerfile
```diff
@@ -0,0 +1,16 @@
+# Use a minimal image as a parent image
+FROM ubuntu:19.10
+ARG CNS_BUILD_DIR
+
+# Update
+RUN apt-get update
+
+# Install plugin
+COPY $CNS_BUILD_DIR/azure-cns /usr/bin
```
tamilmani1989: can we do it in another location instead of /usr/bin because im adding config file and it should be in same directory as that of azure-cns. Can we copy this to /usr/local dir? 
also we have to copy gz file and extract that. If tomorrow we have more than 1 file then we have to edit this docker file everytime

Makefile
jaer-tsun: cns

cns/Dockerfile
jaer-tsun: moved

Makefile
```diff
@@ -267,6 +271,23 @@ azure-vnet-telemetry-image: azure-vnet-telemetry
...
+ifeq ($(GOOS),linux)
+	docker build \
+	-f cns/Dockerfile \
+	-t $(AZURE_CNS_IMAGE):$(VERSION) \
+	--build-arg CNS_BUILD_ARCHIVE=$(CNS_BUILD_DIR)/$(CNS_IMAGE_ARCHIVE_NAME) \
+	.
+	docker save $(AZURE_CNS_IMAGE):$(VERSION) | gzip -c > $(CNS_BUILD_DIR)/$(CNS_IMAGE_ARCHIVE_NAME)
+endif
+
+# Publish the Azure NPM image to a Docker registry
```
tamilmani1989: Azure CNS image

PR #472, opened 2020-01-06
Config option to not enable hairpin on host interface

vipulha:
Adds an option to not enable Hairpin on the host interface. 
Setting hairpin on the host interface creates a problem when the upstream network devices have STP enabled. This prevents the CNI to be used on bare-metal hosts.

Note:

1. The default behavior of CNI is left as-is. Only a optional parameter is added for scenarios where enabling hairpin on host doesn't work.
2. Hairpin will still be enabled on all the virtual interfaces, so a pod will still be able to connect itself via its ServiceIP.

**Release note**:

```release-note
```

### DIFF
--- a/cni/netconfig.go
+++ b/cni/netconfig.go
@@ -42,22 +42,23 @@ type RuntimeDNSConfig struct {
 
 // NetworkConfig represents Azure CNI plugin network configuration.
 type NetworkConfig struct {
-	CNIVersion                 string   `json:"cniVersion"`
-	Name                       string   `json:"name"`
-	Type                       string   `json:"type"`
-	Mode                       string   `json:"mode"`
-	Master                     string   `json:"master"`
-	Bridge                     string   `json:"bridge,omitempty"`
-	LogLevel                   string   `json:"logLevel,omitempty"`
-	LogTarget                  string   `json:"logTarget,omitempty"`
-	InfraVnetAddressSpace      string   `json:"infraVnetAddressSpace,omitempty"`
-	PodNamespaceForDualNetwork []string `json:"podNamespaceForDualNetwork,omitempty"`
-	MultiTenancy               bool     `json:"multiTenancy,omitempty"`
-	EnableSnatOnHost           bool     `json:"enableSnatOnHost,omitempty"`
-	EnableExactMatchForPodName bool     `json:"enableExactMatchForPodName,omitempty"`
-	DisableIPTableLock         bool     `json:"disableIPTableLock,omitempty"`
-	CNSUrl                     string   `json:"cnsurl,omitempty"`
-	Ipam                       struct {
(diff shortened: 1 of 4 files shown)


cni/netconfig.go
```diff
@@ -56,6 +56,7 @@ type NetworkConfig struct {
 	EnableSnatOnHost           bool     `json:"enableSnatOnHost,omitempty"`
 	EnableExactMatchForPodName bool     `json:"enableExactMatchForPodName,omitempty"`
 	CNSUrl                     string   `json:"cnsurl,omitempty"`
+	HairpinOffOnHostInterface  bool     `json:"hairpinOffOnHostInterface,omitempty"`
```
tamilmani1989: can we define this above CNSUrl so that all bools will be in one place.

Instead of HairpinOffOnHostInterface  can we name it as DisableHairpinOnHostInterface  ? Its' upto you.

cni/network/network.go
```diff
@@ -425,11 +425,12 @@ func (plugin *netPlugin) Add(args *cniSkel.CmdArgs) error {
...
-			BridgeName:       nwCfg.Bridge,
-			EnableSnatOnHost: nwCfg.EnableSnatOnHost,
-			DNS:              nwDNSInfo,
-			Policies:         policies,
-			NetNs:            args.Netns,
+			BridgeName:       	        nwCfg.Bridge,
+			EnableSnatOnHost: 	        nwCfg.EnableSnatOnHost,
+			DNS:              	        nwDNSInfo,
+			Policies:         	        policies,
+			NetNs:                      args.Netns,
```
tamilmani1989: space damage

cni/netconfig.go
vipulha: Updated the order.

I went back & forth between DisableHairpin*** vs HairpinOff***. Finally settled on the current name because this flag isn't actually "disabling" the hairpin on the interface, it is never enabling it.

cni/network/network.go
vipulha: Updated

cni/netconfig.go
vipulha: Updated the ordering of the parameters.

I'll prefer keeping the name as HairpinOff***. I had considered Disable*** earlier, but this feature isn't really disabling hairpin instead it is not enabling it :).

khenidak: flag name is super confusing. It should be `DisableHairpin..` with default false.

tamilmani1989: @vipulha the argument goes same for hairpinoff. If this is true, its setting that to off instead its not turning it on.

vipulha: Renamed the parameter.

vipulha:
1. Renamed the parameter.
2. Rebased all commits.

PR #484, opened 2020-01-25
Fix the code used to get the working directory

ashvindeodhar:

**Release note**:

```release-note
```

### DIFF
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -57,7 +57,7 @@ func ReadConfig() (CNSConfig, error) {
 		//dir + string(os.PathSeparator) + defaultConfigName
 	}
 
-	logger.Printf("Config path:%s", configpath)
+	logger.Printf("[Configuration] Config path:%s", configpath)
 
 	content, err := ioutil.ReadFile(configpath)
 	if err != nil {
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 
 	"github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/Azure/azure-container-networking/common"
 )
 
 const (
@@ -46,22 +47,7 @@ func ReadConfig() (CNSConfig, error) {
 	// Check if env set for config path otherwise use default path
 	configpath, found := os.LookupEnv("CNS_CONFIGURATION_PATH")
 	if !found {
-		var (
-			dir string
-			ex  string
-			err error
-		)
-		ex, err = os.Executable()
-		if err == nil {
-			dir = filepath.Dir(ex)
-		} else {
-			var exReal string
-			exReal, err = filepath.EvalSymlinks(ex)
-			if err == nil {
-				dir = filepath.Dir(exReal)
-			}
-		}
-
+		dir, err := common.GetExecutableDirectory()
 		if err != nil {
 			logger.Errorf("[Configuration] Failed to find exe dir:%v", err)
 			return cnsConfig, err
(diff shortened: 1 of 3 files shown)


cns/configuration/configuration.go
```diff
@@ -46,7 +46,7 @@ func ReadConfig() (CNSConfig, error) {
 	// Check if env set for config path otherwise use default path
 	configpath, found := os.LookupEnv("CNS_CONFIGURATION_PATH")
 	if !found {
-		dir, err := filepath.Abs(filepath.Dir(os.Args[0]))
+		dir, err := os.Getwd()
```
tamilmani1989: It was stated that "Getwd returns a rooted path name corresponding to the current directory". So if we run the exe from different directory, will this work?

cns/configuration/configuration.go
```diff
@@ -46,7 +46,22 @@ func ReadConfig() (CNSConfig, error) {
 	// Check if env set for config path otherwise use default path
 	configpath, found := os.LookupEnv("CNS_CONFIGURATION_PATH")
 	if !found {
-		dir, err := filepath.Abs(filepath.Dir(os.Args[0]))
+		var (
+			dir string
+			ex  string
+			err error
+		)
+		ex, err = os.Executable()
```
tamilmani1989: can we make this as common function..

cns/configuration/configuration.go
```diff
@@ -46,7 +46,22 @@ func ReadConfig() (CNSConfig, error) {
...
+		if err == nil {
+			dir = filepath.Dir(ex)
```
tamilmani1989: is this test both in linux and windows?

cns/configuration/configuration.go
ashvindeodhar: done

cns/configuration/configuration.go
ashvindeodhar: yes

cns/configuration/configuration.go
```diff
@@ -56,7 +57,7 @@ func ReadConfig() (CNSConfig, error) {
 		//dir + string(os.PathSeparator) + defaultConfigName
```
tamilmani1989: its not your change...can you please remove this?

common/utils.go
```diff
@@ -323,3 +324,23 @@ func GetAzureCloud(url string) (string, error) {
...
+
+func GetExecutableDirectory() (string, error) {
+	var (
+		dir string
+		ex  string
+		err error
+	)
+	ex, err = os.Executable()
+	if err == nil {
+		dir = filepath.Dir(ex)
```
tamilmani1989: can you add a comment on why are we doing filepath.dir on nil case and EvalSymlinks on err case

cns/networkcontainers/networkcontainers_windows.go
```diff
@@ -250,12 +250,13 @@ func getAzureNetworkContainerBinaryPath() (string, error) {
 		err        error
 	)
 
-	if workingDir, err = filepath.Abs(filepath.Dir(os.Args[0])); err != nil {
+	workingDir, err = common.GetExecutableDirectory()
+	if err != nil {
 		return binaryPath,
 			fmt.Errorf("[Azure CNS] Unable to find working directory. Error: %v. Cannot continue", err)
```
tamilmani1989: make sure we are logging this somewhere

cns/configuration/configuration.go
ashvindeodhar: done

cns/networkcontainers/networkcontainers_windows.go
ashvindeodhar: this error bubbles up and gets printed

common/utils.go
ashvindeodhar: done

PR #489, opened 2020-01-29
Fix CNS regression due to logger change

ashvindeodhar:
log.SetTarget creates the log file under log directory using golang os package. Whenever code sets the log directory, it needed to call SetTarget to create the actual log file under that directory. In the recent logger changes, InitLogger by default set the log directory to the current folder. This created the log file in the current folder. The code then set the log directory to a different location without a subsequent call to log.SetTarget. This resulted into the logger to not find the actual log file in the set log directory.

This fix updates the logger InitLogger function to accept the log directory to create the file in correct log directory. To avoid having such issue, this fix also combines the function calls to set log directory and set target into a single function. This prevents any out of order calls resulting into such issue.

**Release note**:

```release-note
```

### DIFF
--- a/cns/logger/log.go
+++ b/cns/logger/log.go
@@ -28,9 +28,9 @@ type CNSLogger struct {
 }
 
 // Initialize CNS Logger
-func InitLogger(fileName string, logLevel, logTarget int) {
+func InitLogger(fileName string, logLevel, logTarget int, logDir string) {
 	Log = &CNSLogger{
-		logger: log.NewLogger(fileName, logLevel, logTarget),
+		logger: log.NewLogger(fileName, logLevel, logTarget, logDir),
 	}
 }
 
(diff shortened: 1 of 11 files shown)


cns/logger/log.go
```diff
@@ -61,8 +61,8 @@ func Close() {
 	}
 }
```
ninzavivek: could you add in a comment why we need to combine both the function?

cns/logger/log.go
```diff
@@ -61,8 +61,8 @@ func Close() {
...
 
-func SetLogDirectory(dir string) {
-	Log.logger.SetLogDirectory(dir)
+func SetTargetLogDirectory(target int, dir string) error {
```
ninzavivek: I am assuming, the components in aquarius repo, that has vendor depedencies, will update it, when pull in the latest change or it was build break there?

ashvindeodhar: yes once vendor dep are updated

cns/logger/log.go
ashvindeodhar: Since, this has passed all the checks, let me avoid updating this PR. I'll add the comment in my next PR. I have made a note of this.

PR #498, opened 2020-02-06
NPM Version Validation

### LINKED ISSUE #494: NPM doesn't work with Patched Kubernetes
Yes
---

Issue
---

**Which release version?**:
1.0.29 or above
---

NPM
---

Linux
---

Any
---

**For windows: provide output of "$(Get-ItemProperty -Path "C:\windows\system32\hal.dll").VersionInfo.FileVersion"**

---

Kubernetes
---

**What happened**:
Bug in parsing the k8s version:
https://github.com/Azure/azure-container-networking/blob/aeacc60dfee84f414bd0bbaf8ac07848d8b35c10/npm/util/util.go#L101-L105

In a patched k8s, the version.Info apiserver returns looks like

> version.Info{
>   Major: "1",
   **Minor: "14+",**
>   GitVersion: "1.14.8-hotfix.20200127",
> }

While "14+" is not a valid semver, since it only allows number identifier in the minor. https://semver.org/#spec-item-11
The current code to semver.NewVersion(Major+Minor) will get a semantic error.
https://play.golang.org/p/7j2Dcx5_fpH

Further, Major+Minor comes up with "111" for 1.11, which has potential issue if k8s start to bump up the Major version (111 > 21).

---

NPM should work with patched kubernetes apiserver.
---

Build k8s cluster (AKS-Engine)
Replace the apiserver image by mcr.microsoft.com/oss/kubernetes/hyperkube:v1.14.8_f0.0.3
Install the NPM 1.0.30
You will find NPM pod crash with the following error message:
https://github.com/Azure/azure-container-networking/blob/aeacc60dfee84f414bd0bbaf8ac07848d8b35c10/npm/npm.go#L223

If you run `kubectl version`, you will get the following version information.
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.8-hotfix.20200127", GitCommit:"c52f59bbba5fbf21fbb18e9a06f96e563fe4c20a", GitTreeState:"clean", BuildDate:"2020-01-28T02:47:05Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}

---

I'd propose to use the GitVersion to get the k8s version.
---

jaer-tsun:
The value of minor was incorrectly assumed to be e.g. 14.8-hotfix.20191113 instead of 14+

Fixes #494

### DIFF
--- a/npm/util/util.go
+++ b/npm/util/util.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"hash/fnv"
 	"os"
+	"regexp"
 	"strings"
 	"sort"
 
@@ -15,6 +16,9 @@ import (
 
 // IsNewNwPolicyVerFlag indicates if the current kubernetes version is newer than 1.11 or not
 var IsNewNwPolicyVerFlag = false
+
+// regex to get minor version
+var re = regexp.MustCompile("[0-9]+")
 
 // Exists reports whether the named file or directory exists.
 func Exists(filePath string) bool {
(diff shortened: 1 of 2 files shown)


jchauncey:
You should add the test I have in this pr - https://github.com/Azure/azure-container-networking/pull/497

jaer-tsun:
thanks @jchauncey , I've added it but in a different order, since second version for us will always be 1.11

npm/util/util.go
```diff
@@ -98,11 +102,15 @@ func GetHashedName(name string) string {
...
+	minorVers := re.FindAllString(firstVer.Minor, -1)
+	if len(minorVers) < 1 {
+		return -2
+	}
+	v1, err := semver.NewVersion(firstVer.Major+"."+minorVers[0])
 	if err != nil {
 		return -2
 	}
-	v2, err := semver.NewVersion(secondVer.Major+secondVer.Minor)
+	v2, err := semver.NewVersion(secondVer.Major+"."+secondVer.Minor)
```
robbiezhang: as a utility function, I think you should not assume the caller always give a right format in the secondVer.

PR #508, opened 2020-02-21
Add IPAM unit tests && Using ginkgo replace the origin go test

DancingLinks:

**Release note**:

```release-note
```


DancingLinks:
@tamilmani1989, @matmerr, @jaer-tsun. Hi all, I'm an SWE Intern to ASK China team. My leader and mentor want me to be familiar with Azure CNI, and add some unit tests for it. So I mentioned this PR after learning the code of Azure CNI.

This PR mainly adds some unit tests of IPAM to ensure that each function can get the expected results. To make the test code look clearer, I used ginkgo as the unit test framework.

I hope you can spare time to review my PR. If there is any problem, I will correct it in time. You can also connect me on Teams, my alias is t-yuqian :)

At present, I am adding unit tests for the network package.

ipam/null.go
```diff
@@ -59,7 +59,9 @@ func (s *nullSource) refresh() error {
 	}
 
 	// Set the local address space as active.
-	s.sink.setAddressSpace(local)
+	if err := s.sink.setAddressSpace(local); err != nil {
```
jaer-tsun: return s.sink.setAddressSpace(local)

cni/ipam/ipam_test.go
```diff
@@ -4,17 +4,20 @@
 package ipam
 
 import (
+	"encoding/json"
 	"fmt"
+	"github.com/Azure/azure-container-networking/common"
+	"github.com/Azure/azure-container-networking/platform"
+	cniSkel "github.com/containernetworking/cni/pkg/skel"
+	cniTypesCurr "github.com/containernetworking/cni/pkg/types/current"
+	. "github.com/onsi/ginkgo"
```
jaer-tsun: we don't use submodules in this repo so please vendor via go-dep

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,191 @@ var ipamQueryResponse = "" +
...
+func handleIpamQuery(w http.ResponseWriter, r *http.Request) {
+	w.Write([]byte(ipamQueryResponse))
+}
 
-	// Create a fake local agent to handle requests from IPAM plugin.
-	u, _ := url.Parse("tcp://" + ipamQueryUrl)
-	testAgent, err := common.NewListener(u)
+func parseResult(stdinData []byte) (*cniTypesCurr.Result, error) {
+	result := &cniTypesCurr.Result{}
+	err := json.Unmarshal(stdinData, result)
```
jaer-tsun: for single returns, current style is to keep it on the same line

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,191 @@ var ipamQueryResponse = "" +
...
-func TestAddSuccess(t *testing.T) {
-}
 
-func TestDelSuccess(t *testing.T) {
-}
+var (
+	_ = Describe("Test IPAM", func() {
+
+		var (
+			config common.PluginConfig
```
jaer-tsun: looks like linter didn't work here

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,191 @@ var ipamQueryResponse = "" +
...
+
+			Context("When ADD with nothing (request pool)", func() {
+				It("Request pool and ADD successfully", func() {
+					arg.StdinData = getStdinData("0.4.0", ifName, "", "")
+					err = plugin.Add(arg)
+					Expect(err).ShouldNot(HaveOccurred())
+					result, err = parseResult(arg.StdinData)
+					Expect(err).ShouldNot(HaveOccurred())
+					address1, _ := platform.ConvertStringToIPNet("10.0.0.5/16")
+					address2, _ := platform.ConvertStringToIPNet("10.0.0.6/16")
```
jaer-tsun: do we need address2 here?

ipam/azure.go
```diff
@@ -177,7 +177,9 @@ func (s *azureSource) refresh() error {
 	}
 
 	// Set the local address space as active.
-	s.sink.setAddressSpace(local)
+	if err := s.sink.setAddressSpace(local); err != nil {
```
jaer-tsun: just return s.sink.setAddressSpace(local)

ipam/azure_test.go
```diff
@@ -0,0 +1,193 @@
+// Copyright 2017 Microsoft. All rights reserved.
+// MIT License
+
+package ipam
+
+import (
+	"encoding/json"
+	"github.com/Azure/azure-container-networking/common"
+	cniTypesCurr "github.com/containernetworking/cni/pkg/types/current"
+	. "github.com/onsi/ginkgo"
```
jaer-tsun: we don't use submodules in this repo, please vendor libs via go-dep

ipam/azure_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"</Interfaces>"
+
+// Handles queries from IPAM source.
+func handleIpamQuery(w http.ResponseWriter, r *http.Request) {
+	w.Write([]byte(ipamQueryResponse))
+}
+
+func parseResult(stdinData []byte) (*cniTypesCurr.Result, error) {
+	result := &cniTypesCurr.Result{}
+	err := json.Unmarshal(stdinData, result)
```
jaer-tsun: current style is to merge with if statement for single returns
e.g.

if err := json.Marshal...; err != nil {
    ...
}

ipam/fileIpam_test.go
```diff
@@ -1,290 +1,371 @@
...
-			t.Fatalf("default file path set incorrectly")
-		}
+func (sink *addressManagerMock) newAddressSpace(id string, scope int) (*addressSpace, error) {
+	if sink.newAddressSpaceSucc {
+		return &addressSpace{
+			Id:    id,
+			Scope: scope,
+			Pools: make(map[string]*addressPool),
+		}, nil
 	} else {
```
jaer-tsun: no need for else block

ipam/fileIpam_test.go
```diff
@@ -1,290 +1,371 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "MAS Suite")
 }
 
-func TestNewFileIpamSource(t *testing.T) {
-	options := make(map[string]interface{})
-	options[common.OptEnvironment] = common.OptEnvironmentFileIpam
-	fileIpam, _ := newFileIpamSource(options)
+type addressManagerMock struct {
+	newAddressSpaceSucc bool
```
jaer-tsun: type out Succ

ipam/fileIpam_test.go
```diff
@@ -1,290 +1,371 @@
...
-		t.Fatal("didn't throw error on invalid file")
-	}
-
-	interfaces, err = getSDNInterfaces(nonexistentFileName)
-	if interfaces != nil || err == nil {
-		t.Fatal("didn't throw error on nonexistent file")
+func (sink *addressManagerMock) setAddressSpace(*addressSpace) error {
+	if sink.setAddressSpaceSucc {
+		return nil
+	} else {
```
jaer-tsun: no need for else block

ipam/manager_test.go
```diff
@@ -141,233 +146,279 @@ func cleanupTestAddressSpace(am AddressManager) error {
...
+					am := &addressManager{}
+					options := map[string]interface{}{}
+					options[common.OptEnvironment] = common.OptEnvironmentAzure
+					err := am.StartSource(options)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(am.source).NotTo(BeNil())
+				})
+			})
+
+			Context("When environment is mas", func() {
```
jaer-tsun: It doesn't look like we're testing address manager functions anymore?
e.g. RequestAddress, ReleaseAddress, ...

cni/ipam/ipam_test.go
DancingLinks: I think may yes, while testing, plugin.Add(arg) would return 10.0.0.5 or 10.0.0.6 randomly.

ipam/azure.go
DancingLinks: Get!

ipam/azure_test.go
DancingLinks: OK, thanks!

cni/ipam/ipam_test.go
DancingLinks: Get, thanks!

cni/ipam/ipam_test.go
DancingLinks: I will move config to the Context where it works

ipam/fileIpam_test.go
DancingLinks: Ok!

ipam/fileIpam_test.go
DancingLinks: I'll remove it

ipam/fileIpam_test.go
DancingLinks: I'll remove it

ipam/manager_test.go
DancingLinks: I found that functions such as RequestAddress just call the functions in pool.go and raise errors, there is no complicated work in itself. So I didn't test them. And the unit tests of functions in pool.go were all covered in pool_test.go in next PR :)

ipam/null.go
DancingLinks: Get

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,191 @@ var ipamQueryResponse = "" +
...
+}
 
-	err = testAgent.Start(make(chan error, 1))
-	if err != nil {
-		fmt.Printf("Failed to start agent, err:%v.\n", err)
-		return
-	}
+func getStdinData(cniversion, ifname, subnet, ipAddress string) []byte {
+	stdinData := fmt.Sprintf(
+`{
```
tamilmani1989: space damage

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,191 @@ var ipamQueryResponse = "" +
...
+	return []byte(stdinData)
+}
 
-	// Create the plugin.
-	plugin, err = NewPlugin("ipamtest", &config)
-	if err != nil {
-		fmt.Printf("Failed to create IPAM plugin, err:%v.\n", err)
-		return
-	}
+func getIfName() (string, error) {
```
tamilmani1989: why do we need this ? What this function expected to return? I think this is for pod interface name right?

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,191 @@ var ipamQueryResponse = "" +
...
+				Expect(err).NotTo(HaveOccurred())
+			})
+		})
+
+		Describe("Test IPAM ADD and DELETE pool", func() {
+
+			var result *cniTypesCurr.Result
+
+			Context("When ADD with nothing (request pool)", func() {
+				It("Request pool and ADD successfully", func() {
```
tamilmani1989: is this for request pool or request address or for both?

cni/ipam/ipam_test.go
DancingLinks: Yes I copy this code from network.go, for getting the interface name of the test environment.

cni/ipam/ipam_test.go
DancingLinks: Both. I found that this case may happen when request pool.

cni/ipam/ipam_test.go
DancingLinks: I'm now using go-dep instead, adding github.com/onsi/ginkgo, github.com/onsi/gomega and gopkg.in/fsnotify.v1

Gopkg.toml
```diff
@@ -24,6 +24,9 @@
 #   go-tests = true
 #   unused-packages = true
 
+[[override]]
```
jaer-tsun: what is this package for?

cni/ipam/ipam_test.go
jaer-tsun: @tamilmani1989 , should this be random behaviour? I'd assume it'll always be 10.0.0.5 from a clean cluster

Gopkg.toml
DancingLinks: When I run dep ensure, an error occurred.
https://github.com/golang/dep/issues/1799
https://github.com/cloudfoundry-incubator/disaster-recovery-acceptance-tests/commit/983620d3add93622b67c9d5abcd1abaa315c96c7

cni/ipam/ipam_test.go
DancingLinks: In Golang, traversal map does not seem to be sorted by key, so when traversing Addresses map [string]*addressRecord, it may not necessarily take out 10.0.0.5?
````go
	// If no address was found, return any available address.
	if ar == nil {
		for _, ar = range ap.Addresses {
			if !ar.InUse && ar.ID == "" {
				break
			}
			ar = nil
		}

		if ar == nil {
			return "", errNoAvailableAddresses
		}
	}
```

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,192 @@ var ipamQueryResponse = "" +
...
+					address, _ := platform.ConvertStringToIPNet("10.0.0.5/16")
+					Expect(result.IPs[0].Address.IP).Should(Equal(address.IP))
+					Expect(result.IPs[0].Address.Mask).Should(Equal(address.Mask))
+				})
+			})
+		})
+
+		Describe("Test IPAM DELETE", func() {
+
+			Context("When address and subnet is given", func() {
```
mainred: We may need tests to cover specifying an address and empty string.

ipam/manager_test.go
```diff
@@ -4,11 +4,17 @@
...
+	"github.com/Azure/azure-container-networking/platform"
+	"github.com/Azure/azure-container-networking/store"
+	"github.com/Azure/azure-container-networking/testutils"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 	"net"
 	"testing"
-
-	"github.com/Azure/azure-container-networking/common"
+	"time"
```
mainred: The following package groups sequence convention is very popular:
1. standard library
2. third-party packages
3. local packages
Could you please adjust your code to comply with this?

CC @tamilmani1989

ipam/null.go
```diff
@@ -59,7 +59,5 @@ func (s *nullSource) refresh() error {
 	}
 
 	// Set the local address space as active.
-	s.sink.setAddressSpace(local)
-
-	return nil
+	return s.sink.setAddressSpace(local)
```
mainred: ❤️

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,192 @@ var ipamQueryResponse = "" +
...
+			arg = &cniSkel.CmdArgs{IfName:ifName}
+		})
+
+		AfterSuite(func() {
+			// Cleanup.
+			plugin.Stop()
+			testAgent.Stop()
+		})
+
+		Context("IPAM start", func() {
```
mainred: If this context assertion break, all the following specs will be executed, right?

ipam/azure_test.go
```diff
@@ -0,0 +1,192 @@
...
+					pool, ok = as.Pools["10.1.0.0/16"]
+					Expect(ok).To(BeTrue())
+
+					_, ok = pool.Addresses["10.1.0.4"]
+					Expect(ok).To(BeTrue())
+				})
+			})
+		})
+	})
+)
```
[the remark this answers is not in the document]
DancingLinks: Ok, I'll add a newline here :) forgot

cni/ipam/ipam_test.go
DancingLinks: In cni/network/network.go 145, it will detect the in used interface name. For network.go get the same interface name as the test code, I added this function...

ipam/manager_test.go
DancingLinks: ok!

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,192 @@ var ipamQueryResponse = "" +
...
+				plugin.SetOption(common.OptEnvironment, common.OptEnvironmentAzure)
+				plugin.SetOption(common.OptAPIServerURL, "null")
+				plugin.SetOption(common.OptIpamQueryUrl, "http://"+ipamQueryUrl)
+				// Start the plugin.
+				err = plugin.Start(&config)
+				Expect(err).NotTo(HaveOccurred())
+			})
+		})
+
+		Describe("Test IPAM ADD and DELETE pool", func() {
```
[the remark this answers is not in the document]
DancingLinks: address is correct, not pool

cni/ipam/ipam_test.go
DancingLinks: I'll add a case of release pool, when address is empty

Gopkg.toml
tamilmani1989: is this signed off by lca?

cni/ipam/ipam_test.go
```diff
@@ -4,17 +4,20 @@
...
 
 import (
+	"encoding/json"
 	"fmt"
+	"github.com/Azure/azure-container-networking/common"
+	"github.com/Azure/azure-container-networking/platform"
+	cniSkel "github.com/containernetworking/cni/pkg/skel"
+	cniTypesCurr "github.com/containernetworking/cni/pkg/types/current"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
tamilmani1989: are these 3rd party packages signed by microsoft legal (lca)

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,192 @@ var ipamQueryResponse = "" +
...
+
+		var (
+			plugin *ipamPlugin
+			testAgent *common.Listener
+			arg *cniSkel.CmdArgs
+			ifName string
+			err error
+		)
+
+		BeforeSuite(func() {
```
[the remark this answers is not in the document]
DancingLinks: I'll move them outside of the Describe, thx!

cni/ipam/ipam_test.go
tamilmani1989: the problem is if there are multiple interfaces, if it returns lo adapter or docker0 bridge as 1st interface, then this will break. for simplicity you can keep it as eth0

cni/ipam/ipam_test.go
tamilmani1989: Then update description saying this is test for ipam add call

cni/ipam/ipam_test.go
tamilmani1989: @jaer-tsun  yes it can return in any order

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,192 @@ var ipamQueryResponse = "" +
...
+					Expect(err).ShouldNot(HaveOccurred())
+					address1, _ := platform.ConvertStringToIPNet("10.0.0.5/16")
+					address2, _ := platform.ConvertStringToIPNet("10.0.0.6/16")
+					Expect(result.IPs[0].Address.IP).Should(Or(Equal(address1.IP), Equal(address2.IP)))
+					Expect(result.IPs[0].Address.Mask).Should(Equal(address1.Mask))
+				})
+			})
+
+			Context("When DELETE with subnet and address", func() {
+				It("DELETE address successfully", func() {
```
tamilmani1989: update description saying its delete call for ipam triggering release address

cni/ipam/ipam_test.go
```diff
@@ -27,70 +30,192 @@ var ipamQueryResponse = "" +
...
+					arg.StdinData = getStdinData("0.4.0", ifName, "10.0.0.0/16", result.IPs[0].Address.IP.String())
+					err = plugin.Delete(arg)
+					Expect(err).ShouldNot(HaveOccurred())
+				})
+			})
+
+			Context("When DELETE with subnet (release pool)", func() {
```
tamilmani1989: update description saying its delete call for ipam triggering releasing pool

cni/ipam/ipam_test.go
tamilmani1989: also add a case where release pool fails when address is allocated in pool and not released

ipam/azure_test.go
```diff
@@ -0,0 +1,192 @@
...
+					options[common.OptIpamQueryUrl] = "http://"+ipamQueryUrl
+					source, err = newAzureSource(options)
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(source.name).Should(Equal("Azure"))
+					Expect(source.queryUrl).Should(Equal("http://"+ipamQueryUrl))
+					Expect(source.queryInterval).Should(Equal(azureQueryInterval))
+				})
+			})
+
+			Context("When time interval is less", func() {
```
tamilmani1989: couldnot understand what this test for..what we are testing here

ipam/azure_test.go
```diff
@@ -0,0 +1,192 @@
...
+					source.queryInterval = time.Hour
+					err = source.refresh()
+					Expect(err).To(BeNil())
+					source.queryInterval = time.Nanosecond
+				})
+			})
+
+			Context("When newAddressSpace err", func() {
+				It("Exit with error when refresh", func() {
+					sink := &addressManagerMock{false, true}
```
tamilmani1989: what this mock does..can you add a comment

ipam/manager_test.go
tamilmani1989: my only concern with 3rd party is we  have to get approval from lca before using that..did we get required approval?

ipam/manager_test.go
```diff
@@ -141,233 +141,279 @@ func cleanupTestAddressSpace(am AddressManager) error {
...
+					timeReboot, _ := platform.GetLastRebootTime()
+					am.store = &testutils.KeyValueStoreMock{
+						ModificationTime: timeReboot.Add(time.Hour),
+					}
+					ap := &addressPool{
+						Id:			"ap-test",
+						RefCount: 	1,
+						Addresses: 	make(map[string]*addressRecord),
+					}
+					ap.Addresses["ar-test"] = &addressRecord{
```
tamilmani1989: can we also add some addr and check if addrbyid returns that address...thats the main purpose of addrsbyid

ipam/manager_test.go
```diff
@@ -141,233 +141,279 @@ func cleanupTestAddressSpace(am AddressManager) error {
...
-	if err != nil {
-		t.Errorf("ReleaseAddress failed, err:%v", err)
-	}
-
-	err = am.ReleasePool(LocalDefaultAddressSpaceId, poolId)
-	if err != nil {
-		t.Errorf("ReleasePool failed, err:%v", err)
-	}
-}
+var (
```
tamilmani1989: was unit test for requestpool, requestaddr, releaseddr, releasepool moved somewhere else?

cni/ipam/ipam_test.go
mainred: Sorry, my fault, please keep your code as it is.

ipam/manager_test.go
mainred: Thanks so much for pointing this out. I am not sure about the newly imported packages in this PR.
I saw from our Golang files that we are using different package group conventions, so I here just want to share my idea to follow one convention.

cni/ipam/ipam_test.go
mainred: Thank @DancingLinks for your clarification. We do use a real 
> the problem is if there are multiple interfaces, if it returns lo adapter or docker0 bridge as 1st interface, then this will break. for simplicity you can keep it as eth0

After re-reading the code, @DancingLinks and I found that we do not have an interface check, so both lo and docker0 will still work. But in practise, `lo` and `docker0` will not be picked for mac address mismatch.

Also, I found that interface name is not specified for the Kubernetes case[1]. How about we first cover the path without an interface name when invoking the IPAM plugin.
If we have to select a real interface name, how about removing `lo` and `docker0` from the interface list and pick one randomly, in case we may have interface names other than `eth0`, in different OS releases? This seems to be more real.

[1]https://github.com/kubernetes/kubernetes/blob/581d3e26c97ade852b54bc7f102f1a4d16b74437/pkg/kubelet/dockershim/network/cni/cni.go#L353

ipam/azure_test.go
mainred: Neither can I. Let's remove this test unless you can make it more descriptive and please also remember test will also be maintained as part of our code.

cni/ipam/ipam_test.go
mainred: I am not sure about this, but we have these two packages used in repos like aks-engine already.
https://github.com/search?q=org%3AAzure+github.com%2Fonsi%2Fginkgo&type=Code

mainred:
@DancingLinks, besides rebase your code, could you please also squash your commits and reword your commit messages to explain your changes neatly before we merge this PR?

cni/ipam/ipam_test.go
DancingLinks: finished

DancingLinks:
> @DancingLinks Can you rebase and push it again?

Rebased finished :)

DancingLinks:
@tamilmani1989 rebased finished :)

DancingLinks:
@mainred Thanks a lot for your suggestion! I've now merged them in a single commit.

ipam/azure_test.go
DancingLinks: I will change the description as "refresh interval is too short". It may be more clearly now.

PR #509, opened 2020-02-22
Fix unnecessary consecutiveAdd for windows

### LINKED ISSUE #367: Optimize attach endpoint for repetitive ADD calls

---

Optimization
---

**Which release version?**:
v1.0.22
---

CNI
---

Windows
---

---

**For windows: provide output of "$(Get-ItemProperty -Path "C:\windows\system32\hal.dll").VersionInfo.FileVersion"**
10.0.17763.529 (WinBuild.160101.0800)
---

k8s
---

**What happened**:
For the repetitive CNI ADD calls for the same pause container, azure CNI will try to invoke HNS and attach the endpoint to the container again. consecutiveAdd is not needed for the pause container ADD.
---

Azure CNI can optimize this by not invoking HNS and returning early.
---

Create container on windows.
---

---

csfmomo:
Added endpoint checking logic before calling consecutiveAdd. If endpoint is attached, return directly without calling consecutiveAdd. 
When endpoint is attached, it means ADD has been done successfully previously. No consecutiveAdd is needed. 

**Which issue this PR fixes**
fixes #367

### DIFF
--- a/cni/network/network_windows.go
+++ b/cni/network/network_windows.go
@@ -39,25 +39,20 @@ func handleConsecutiveAdd(args *cniSkel.CmdArgs, endpointId string, nwInfo *netw
 		return nil, err
 	}
 
-	/*
-	 * Return in case of endpoint is already attached and consecutive add call doesn't need to be handled
-	 */
-	endpoint, err := GetHNSEndpointByID(endpointId)
-	isAttached, err := endpoint.IsAttached(args.ContainerID)
-	if isAttached {
-		log.Printf("No consecutive Add needed since endpoint is already attatched")
-		return nil, err
-	}
-
 	hnsEndpoint, err := hcsshim.GetHNSEndpointByName(endpointId)
 	if hnsEndpoint != nil {
 		log.Printf("[net] Found existing endpoint through hcsshim: %+v", hnsEndpoint)
 		log.Printf("[net] Attaching ep %v to container %v", hnsEndpoint.Id, args.ContainerID)
 
-		err := hcsshim.HotAttachEndpoint(args.ContainerID, hnsEndpoint.Id)
-		if err != nil {
-			log.Printf("[cni-net] Failed to hot attach shared endpoint[%v] to container [%v], err:%v.", hnsEndpoint.Id, args.ContainerID, err)
-			return nil, err
+		endpoint, _ := hcsshim.GetHNSEndpointByID(hnsEndpoint.Id)
+		isAttached, _ := endpoint.IsAttached(args.ContainerID)
+		// Attach endpoint if it's not attached yet. If attached, populate result directly should be good enough.
+		if !isAttached {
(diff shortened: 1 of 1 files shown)


cni/network/network_windows.go
```diff
@@ -39,6 +39,16 @@ func handleConsecutiveAdd(args *cniSkel.CmdArgs, endpointId string, nwInfo *netw
 		return nil, err
 	}
 
+	/*
+	 * Return in case of endpoint is already attached and consecutive add call doesn't need to be handled
+	 */
+	endpoint, err := GetHNSEndpointByID(endpointId)
```
jaer-tsun: if you don't need to check err

endpoint, _ := ...

cni/network/network_windows.go
```diff
@@ -39,6 +39,16 @@ func handleConsecutiveAdd(args *cniSkel.CmdArgs, endpointId string, nwInfo *netw
...
+	isAttached, err := endpoint.IsAttached(args.ContainerID)
```
jaer-tsun: if isAttached, err := ...; isAttached {...}

jaer-tsun: actually, nvm, just follow the style in this file

cni/network/network_windows.go
```diff
@@ -44,10 +44,15 @@ func handleConsecutiveAdd(args *cniSkel.CmdArgs, endpointId string, nwInfo *netw
 		log.Printf("[net] Found existing endpoint through hcsshim: %+v", hnsEndpoint)
 		log.Printf("[net] Attaching ep %v to container %v", hnsEndpoint.Id, args.ContainerID)
 
-		err := hcsshim.HotAttachEndpoint(args.ContainerID, hnsEndpoint.Id)
-		if err != nil {
-			log.Printf("[cni-net] Failed to hot attach shared endpoint[%v] to container [%v], err:%v.", hnsEndpoint.Id, args.ContainerID, err)
-			return nil, err
+		endpoint, _ := hcsshim.GetHNSEndpointByID(hnsEndpoint.Id)
```
jaer-tsun: check if endpoint is nil

cni/network/network_windows.go
```diff
@@ -44,10 +44,15 @@ func handleConsecutiveAdd(args *cniSkel.CmdArgs, endpointId string, nwInfo *netw
...
+		isAttached, _ := endpoint.IsAttached(args.ContainerID)
+		// Attach endpoint if it's not attached yet. If attached, populate result directly should be good enough.
```
jaer-tsun: change to // Attach endpoint if it's not attached yet

and move above line 47, since that's what the couple lines after does

cni/network/network_windows.go
csfmomo: Need to sync up with hns team to see whether its necessary to check hnsEndpoint.Id nil when hnsEndpoint is not nill. If it's necessary, let me add in another separate CR.

csfmomo:
Test is finished and it generated logs which proved this CR change is expected.

PR #546, opened 2020-04-14
Check Processed NP Map & Add Update Pod Conditions

### LINKED ISSUE #428: NPM does not remove succeeded pod ips from firewall ipsets
---
Yes

---
ISSUE

**Which release version?**:
---
1.0.28

---
NPM

---
Linux

---
Linux aks-agentpool-34239724-1 4.15.0-1059-azure #64-Ubuntu SMP Fri Sep 13 17:02:44 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

---
Kubernetes

**What happened**:
---
Pods that go to a completed state before being deleted are not removed from the relevant ipset. 

This eventually causes incorrect network policy/firewall rules to be applied when an unrelated pod happens to re-use the ipaddress of the completed pod. 

---

A pod that goes to a completed state should be removed from all of the ipsets that it belongs to, similar to when it is deleted.

---

Create a Job and delete some time after completion. e.g.
```
apiVersion: batch/v1
kind: Job
metadata:
  name: hello-world
spec:
  template:
    spec:
      containers:
      - name: pi
        image: perl
        command: ["perl",  "-e", "print 'hello world...'; sleep 30; print 'goodbye world';"]
      restartPolicy: Never
```

Observe the logs of one of the azure-npm pods and note that the pod creation is logged but no deletion event is ever recorded. As you might expect, this is also reflected in the underlying ipsets used by the firewall rules.

---

I think the problem is in npm/pod.go where the UpdatePod function should handle the pod going into a Succeeded or Failed state. Currently any pod in those states is ignored.

jaer-tsun:
- check both processed and raw network policy maps when updating
- when pods get updated, only add the entries back if pod is not in completed or failed state since we're assuming that they will release their IPs

**Which issue this PR fixes**:
fixes #428

### DIFF
--- a/npm/pod.go
+++ b/npm/pod.go
@@ -7,6 +7,7 @@ import (
 	"github.com/Azure/azure-container-networking/npm/util"
 
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 )
 
 func isValidPod(podObj *corev1.Pod) bool {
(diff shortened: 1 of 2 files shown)


npm/pod.go
```diff
@@ -99,7 +101,9 @@ func (npMgr *NetworkPolicyManager) UpdatePod(oldPodObj, newPodObj *corev1.Pod) e
 		return err
 	}
 
-	if newPodObj.ObjectMeta.DeletionTimestamp == nil && newPodObj.ObjectMeta.DeletionGracePeriodSeconds == nil {
+	// Assume that the pod IP will be released when pod moves to succeeded or failed state.
+	// If the pod transitions back to an active state, then add operation will re establish the updated pod info.
+	if newPodObj.ObjectMeta.DeletionTimestamp == nil && newPodObj.ObjectMeta.DeletionGracePeriodSeconds == nil && newPodObjPhase != v1.PodSucceeded && newPodObjPhase != v1.PodFailed {
```
csfmomo, line 106: Do we need to consider about PodUnknown situation?

PR #547, opened 2020-04-15
Update action to block and increase priority for allow

aegal:

**Release note**:

```release-note
```

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -245,12 +245,9 @@ type ValidAclPolicySetting struct {
 }
 
 const (
-	ActionTypeAllow string = "Allow"
-
-	ActionTypeBlock string = "Block"
-
-	DirectionTypeIn string = "In"
-
+	ActionTypeAllow  string = "Allow"
+	ActionTypeBlock  string = "Block"
+	DirectionTypeIn  string = "In"
 	DirectionTypeOut string = "Out"
 )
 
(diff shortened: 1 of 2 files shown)


cns/NetworkContainerContract.go
```diff
@@ -244,6 +244,16 @@ type ValidAclPolicySetting struct {
 	Priority        uint16 `json:","`
 }
 
+const (
+	ActionTypeAllow string = "Allow"
+
```
tamilmani1989, line 249: remove space

cns/NetworkContainerContract.go
```diff
@@ -244,6 +244,16 @@ type ValidAclPolicySetting struct {
...
+	ActionTypeBlock string = "Block"
+
```
tamilmani1989, line 251: no space between each constants

cns/hnsclient/hnsclient_windows.go
```diff
@@ -67,6 +67,9 @@ const (
 	// aclPriority200 indicates the ACL priority of 200
 	aclPriority200 = 200
 
+	// aclPriority1000 indicates the ACL priority of 1000
+	aclPriority1000 = 1000
```
tamilmani1989, line 71: why choose 1000? does this signify higher or lower priority

aegal: Higher priority for allows, right now we block at 2000 and allow at 200. It allows for the orchestrator to specify a rule that is block (example: 999 block all below) and allow rules to exclude each port/protocol/src/dest address.

cns/NetworkContainerContract.go
aegal: changed.

PR #548, opened 2020-04-23
Fix cni logs

pjohnst5:
Empty CNI telemetry logs were being sent and cluttering jarvis. This PR disallows empty messages being sent.

**Which issue this PR fixes**
https://msazure.visualstudio.com/One/_workitems/edit/7041879

@tamilmani1989 , I tried adding a function in that defer call we were talking about yesterday, but it broke the connectivity of new pods. Opted for this simpler fix of just checking the error and event message for empty instead.

### DIFF
--- a/cni/network/plugin/main.go
+++ b/cni/network/plugin/main.go
@@ -259,10 +259,11 @@ func main() {
 	reflect.ValueOf(reportManager.Report).Elem().FieldByName("CniSucceeded").SetBool(true)
 	reflect.ValueOf(reportManager.Report).Elem().FieldByName("OperationDuration").SetInt(executionTimeMs)
 
-	if err = reportManager.SendReport(tb); err != nil {
-		log.Errorf("SendReport failed due to %v", err)
-	} else {
-		log.Printf("Sending report succeeded")
+	if cniReport.ErrorMessage != "" || cniReport.EventMessage != "" {
+		if err = reportManager.SendReport(tb); err != nil {
+			log.Errorf("SendReport failed due to %v", err)
+		} else {
+			log.Printf("Sending report succeeded")
+		}
 	}
-
 }
--- a/cni/network/plugin/main.go
+++ b/cni/network/plugin/main.go
@@ -242,13 +242,14 @@ func main() {
 	}
 
 	executionTimeMs := time.Since(startTime).Milliseconds()
-	cnimetric.Metric = aitelemetry.Metric{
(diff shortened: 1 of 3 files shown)


cni/network/plugin/main.go
```diff
@@ -259,10 +259,11 @@ func main() {
...
 
-	if err = reportManager.SendReport(tb); err != nil {
-		log.Errorf("SendReport failed due to %v", err)
-	} else {
-		log.Printf("Sending report succeeded")
+	if cniReport.ErrorMessage != "" || cniReport.EventMessage != "" {
+		if err = reportManager.SendReport(tb); err != nil {
+			log.Errorf("SendReport failed due to %v", err)
+		} else {
+			log.Printf("Sending report succeeded")
```
neaggarwMS, line 266: Shouldn't this be logged even if ErrorMessage/EventMessage is empty?

pjohnst5: When ErrorMessage and EventMessage are both empty, we don't want to send the report because it causes empty log statements in telemetry like [this](https://jarvis-west.dc.ad.msft.net/6753C8EF) (look at "Raw" column"

cniReport is a pointer to the report object inside of reportManager. reportManager.SendReport() is sends the cniReport object to telemetry. That's why were checking for cniReport.ErrorMessage and cniReport.EventMessage first before reportManager.SendReport().

telemetry/aiwrapper.go
```diff
@@ -47,8 +47,10 @@ func SendAITelemetry(cnireport CNIReport) {
...
 	if cnireport.ErrorMessage != "" {
 		msg = cnireport.ErrorMessage
-	} else {
+	} else if cnireport.EventMessage != "" {
 		msg = cnireport.EventMessage
+	} else {
+		return
 	}
 
 	report := aitelemetry.Report{
```
tamilmani1989, line 56: can you also remove cni prefix from msg, no need to set prefix

line 57: Message:          "CNI:" + msg,

cni/network/plugin/main.go
```diff
@@ -248,7 +248,9 @@ func main() {
 		CustomDimensions: make(map[string]string),
 	}
 	network.SetCustomDimensions(&cnimetric, nil, err)
-	telemetry.SendCNIMetric(&cnimetric, tb)
+	if cniReport.ErrorMessage != "" || cniReport.EventMessage != "" {
```
tamilmani1989, line 251: can move this condition before line 245 where we initialize cnimetric?

PR #549, opened 2020-04-25
Cns yaml

pjohnst5:
The beginnings of the CNS yaml which will deploy a CNS instance on every AKS worker node for AKS swift.

### DIFF
--- a/cns/azure-cns.yaml
+++ b/cns/azure-cns.yaml
@@ -1,0 +1,63 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: azure-cns
+  namespace: kube-system
+  labels:
+    app: azure-cns
+    addonmanager.kubernetes.io/mode: Reconcile
+spec:
+  selector:
+    matchLabels:
+      k8s-app: azure-cns
+  template:
+    metadata:
+      labels:
+        k8s-app: azure-cns
+      annotations:
+        cluster-autoscaler.kubernetes.io/daemonset-pod: "true"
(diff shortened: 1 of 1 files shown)


cns/azure-cns.yaml
```diff
@@ -0,0 +1,62 @@
...
+        k8s-app: azure-cns
+      annotations:
+        cluster-autoscaler.kubernetes.io/daemonset-pod: "true"
+    spec:
+      priorityClassName: system-node-critical
+      tolerations:
+        - key: CriticalAddonsOnly
+          operator: Exists
+      nodeSelector:
+        beta.kubernetes.io/os: linux
```
thatmattlong, line 24: We'll want to support Windows hosts too, but I imagine a few things will change (like the hostpath file mounts). I think we can cover this case in Helm but let's discuss if we want to try supporting both linux and windows hosts in this manifest file, or if we should go with an alternative like Helm or separate manifests

cns/azure-cns.yaml
```diff
@@ -0,0 +1,62 @@
...
+            - name: CNSIpAddress
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+      hostNetwork: true
+      volumes:
+        - name: log
+          hostPath:
+            path: /var/log
+            type: Directory
```
thatmattlong, line 50: If we log to stdout instead of writing to a file mounted inside the Pod, we could capture the logs with kube-apiserver, which should help with ops

PR #556, opened 2020-05-05
Ignore Non-Actionable Updates

jaer-tsun:
Ignore non-actionable namespace and pod updates.


jaer-tsun:
initial tests should fail; will need to pull latest changes from #555

PR #561, opened 2020-05-16
Explicitly add content type header.

ninzavivek:
If CNS Client happen to use a library which checks for the content type before decoding the response ( ex: Python -aiohttp, then reading the response will always fail. 

We are defaulting to "Content-Type: text/plain; charset=utf-8"

As the response is always JSON ended. The fix basically always add the Content-Type header as  "application/json; charset=UTF-8"


ninzavivek:
Considering the response is always JSON encoded, I think it is fine.

PR #568, opened 2020-05-26
Enable hairpin on VM Nic by default

tamilmani1989:
Enable hairpin on VM Nic by default. This will address for LB scenarios where VM acts as proxy or router.

**Release note**:

```release-note
```

### DIFF
--- a/cni/azure-linux.conflist
+++ b/cni/azure-linux.conflist
@@ -6,7 +6,6 @@
          "type":"azure-vnet",
          "mode":"bridge",
          "bridge":"azure0",
-         "disableHairpinOnHostInterface":true,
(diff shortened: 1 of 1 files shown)


cni/azure-linux.conflist
```diff
@@ -6,7 +6,6 @@
          "type":"azure-vnet",
          "mode":"bridge",
          "bridge":"azure0",
-         "disableHairpinOnHostInterface":true,
```
pjohnst5, line 9: Will we get duplicate UDP/Tcp packets from the load balancers like we did before? https://github.com/Azure/azure-container-networking/pull/530

tamilmani1989: yes we cannot support UDP.

PR #570, opened 2020-05-28
NetworkConfig CRD defintions

thatmattlong:
This PR adds definitions for the NetworkConfig CRD that will be used to pass goal state to CNS

We need to confirm that only `bufferSize` is passed to CNS via the status, and no other options regarding the pool of allocated IP addresses. In other words, is CNS responsible for setting its own thresholds based on the buffer size, or should we also pass it a `threshold`?


networkconfig/api/v1/networkconfig_types.go
```diff
@@ -0,0 +1,78 @@
...
+
+// Important: Run "make" to regenerate code after modifying this file
+
+// AvailableIPMapping groups an IP address and Name. This is a struct and not a Map for future extensibility
+type AvailableIPMapping struct {
+	Name string `json:"name,omitempty"`
+	IP   string `json:"ip,omitempty"`
+}
+
+// NetworkContainerSpec defines the structure of a Network Container as found in NetworkConfigStatus
```
paulgmiller, line 31: redundant definition. People in k8s world have already have a notion on what a contaienr is so might be worth definiting abit more.

networkconfig/api/v1/networkconfig_types.go
```diff
@@ -0,0 +1,78 @@
...
+type NetworkContainerSpec struct {
+	PrimaryIP      string               `json:"primaryIP,omitempty"`
```
paulgmiller, line 33: Will CNS consume anythign besides the available IPs? if not should we be yagni and remove remove some of these  until their needed to keep the interface thin?

networkconfig/api/v1/networkconfig_types.go
```diff
@@ -0,0 +1,78 @@
...
+
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Important: Run "make" to regenerate code after modifying this file
+
+// AvailableIPMapping groups an IP address and Name. This is a struct and not a Map for future extensibility
```
paulgmiller, line 25: define what a name is. It's a guid but who generates it does it persist for the lifetime of the subnet/vnet?

networkconfig/api/v1/networkconfig_types.go
```diff
@@ -0,0 +1,78 @@
...
+type NetworkConfigSpec struct {
+	Count       int64    `json:"count,omitempty"`
+	IPsNotInUse []string `json:"ipsNotInUse,omitempty"`
+}
+
+// NetworkConfigStatus defines the observed state of NetworkConfig
+type NetworkConfigStatus struct {
+	Count      int64                  `json:"count,omitempty"`
+	BufferSize int64                  `json:"bufferSize,omitempty"`
+	NCs        []NetworkContainerSpec `json:"ncs,omitempty"`
```
paulgmiller, line 51: wierd for the status to contain something with spec in it's name
Also why is this is list. seems like NetworkContainerSpec  should just be inside here. NetworkContainerSpec

networkconfig/api/v1/networkconfig_types.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// +kubebuilder:object:root=true
+
+// NetworkConfig is the Schema for the networkconfigs API
+// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:subresource:status
+type NetworkConfig struct {
```
paulgmiller, line 59: NodeNetworConfig?

networkconfig/api/v1/networkconfig_types.go
thatmattlong: Agreed, but that is what the swift component is called. Let me see if we can call it something else for this purpose

networkconfig/api/v1/networkconfig_types.go
thatmattlong: CNS will be consuming all of these fields

networkconfig/api/v1/networkconfig_types.go
thatmattlong: Agreed, renamed to NetworkContainer (pending another rename to avoid using "container"). This is a list because there can be more than one NC per Node

networkconfig/api/v1/networkconfig_types.go
```diff
@@ -0,0 +1,78 @@
...
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Important: Run "make" to regenerate code after modifying this file
+
+// AvailableIPMapping groups an IP address and Name. This is a struct and not a Map for future extensibility
+type AvailableIPMapping struct {
```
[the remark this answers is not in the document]
thatmattlong: I don't want to call it an `IP` because this fundamentally represents an IP *allocation*, since an IP address could be allocated to this node, deallocated, then reallocated under a different "IP allocation". Similarly, deletions happen via the IP allocation's UUID, not IP. How about `IPAllocation` instead of `AvailableIPMapping`?

networkconfig/api/v1/networkconfig_types.go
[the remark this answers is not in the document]
thatmattlong: Yes, will do, pending a potential name change that Paul M brought up in https://github.com/Azure/azure-container-networking/pull/570#discussion_r432173572

networkconfig/api/v1alpha/networkconfig_types.go
```diff
@@ -0,0 +1,81 @@
...
+// AvailableIPMapping groups an IP address and Name. This is a struct and not a Map for future extensibility
+type AvailableIPMapping struct {
+	Name string `json:"name,omitempty"`
+	IP   string `json:"ip,omitempty"`
+}
+
+// NetworkContainer defines the structure of a Network Container as found in NetworkConfigStatus
+type NetworkContainer struct {
+	PrimaryIP      string               `json:"primaryIP,omitempty"`
+	ID             string               `json:"id,omitempty"`
```
neaggarwMS, line 34: nitpick: Can we move ID to the top, This ID represents an ID for the NetworkContainer

networkconfig/api/v1alpha/networkconfig_types.go
```diff
@@ -0,0 +1,81 @@
...
+	Subnet         string               `json:"subnet,omitempty"`
+	AvailableIPs   []AvailableIPMapping `json:"availableIPs,omitempty"`
+	IPCount        int64                `json:"ipCount,omitempty"`
+	DefaultGateway string               `json:"defaultGateway,omitempty"`
+	Netmask        string               `json:"netmask,omitempty"`
+}
+
+// NetworkConfigSpec defines the desired state of NetworkConfig
+type NetworkConfigSpec struct {
+	IPCount     int64    `json:"ipCount,omitempty"`
```
neaggarwMS, line 44: My 2 cents, Can we rename it as RequestedIPCount? It will be more explicit. IPCount may still confuse as a reference count to the IPsNotInUse list.

networkconfig/api/v1alpha/networkconfig_types.go
```diff
@@ -0,0 +1,81 @@
...
+	IPsNotInUse []string `json:"ipsNotInUse,omitempty"`
+}
+
+// NetworkConfigStatus defines the observed state of NetworkConfig
+type NetworkConfigStatus struct {
+	IPCount                 int64              `json:"count,omitempty"`
```
neaggarwMS, line 50: Again, can we make it more explicit as AllocatedIPCount?

networkconfig/api/v1/networkconfig_types.go
thatmattlong: Spoke with the team and we are going to stick with this naming

nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -0,0 +1,81 @@
...
+
+// NodeNetworkConfigSpec defines the desired state of NetworkConfig
+type NodeNetworkConfigSpec struct {
+	RequestedIPCount int64    `json:"requestedIPCount,omitempty"`
+	IPsNotInUse      []string `json:"iPsNotInUse,omitempty"`
+}
+
+// NodeNetworkConfigStatus defines the observed state of NetworkConfig
+type NodeNetworkConfigStatus struct {
+	AssignedIPCount         int64              `json:"assignedIPCount,omitempty"`
```
paulgmiller, line 55: the assigned up count here and in network container seem unnecessary can just len() IpAssignments.  It's a slice not a stream :)

nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -0,0 +1,81 @@
...
+	BatchSize               int64              `json:"batchSize,omitempty"`
+	ReleaseThresholdPercent int64              `json:"releaseThresholdPercent,omitempty"`
+	RequestThresholdPercent int64              `json:"requestThresholdPercent,omitempty"`
+	NetworkContainers       []NetworkContainer `json:"networkContainers,omitempty"`
```
paulgmiller, line 59: Can we give an example if why we'd have multiple network containers on a single node? i

paulgmiller: When would a single node have different different primary ips and different subnets.

nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
thatmattlong: @neaggarwMS wanted these for easier debugging/ops. If you don't object @paulgmiller, I'll leave them in. If you do, we can discuss with Neha.

nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -0,0 +1,81 @@
...
+
+// NetworkContainer defines the structure of a Network Container as found in NetworkConfigStatus
+type NetworkContainer struct {
+	ID              string         `json:"id,omitempty"`
+	PrimaryIP       string         `json:"primaryIP,omitempty"`
+	Subnet          string         `json:"subnet,omitempty"`
+	IPAssignments   []IPAssignment `json:"iPAssignments,omitempty"`
+	AssignedIPCount int64          `json:"assignedIPCount,omitempty"`
+	DefaultGateway  string         `json:"defaultGateway,omitempty"`
+	Netmask         string         `json:"netmask,omitempty"`
```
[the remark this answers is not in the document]
thatmattlong: An NC can have only one subnet. I'll make this subnetId.

nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
thatmattlong: Right now we are planning one NC per node and thus one primary IP and one subnet. However, this imposes a limit on the number of IPs, so we want to support multiple NCs in the future. It will be easier to add multi-NC support when we can support it on the other tooling if the CRD already supports it.

PR #573, opened 2020-06-03
Limit and rotate NPM logs

jaer-tsun:
Limit and rotate NPM logs
image


npm/plugin/main.go
```diff
@@ -26,6 +26,7 @@ func initLogging() error {
 		log.Logf("Failed to configure logging, err:%v.", err)
 		return err
 	}
+	log.SetLogFileLimits(log.MaxLogFileSize, log.MaxLogFileCount)
```
neaggarwMS, line 29: :)

jaer-tsun: Supposedly, I'm still testing

jaer-tsun: it works

PR #574, opened 2020-06-06
CNS to DNC communication in Managed DNC

jaer-tsun:
1. CNS talks to DNC to get node goalstate
2. CNS checks with NMA if it has goalstate for NC

re-review after pretty fat rebase


cns/restserver/restserver.go
```diff
@@ -1252,7 +1263,31 @@ func (service *HTTPRestService) getNetworkContainerResponse(req cns.GetNetworkCo
...
+				service.lock.Lock()
+				if getNetworkContainerResponse.Response.ReturnCode == NotFound {
+					return getNetworkContainerResponse
+				}
+
+				containerID = service.state.ContainerIDByOrchestratorContext[context]
+			}
+		} else if isManagedDnc {
+			_, getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.isNCWaitingForUpdate(service.state.ContainerStatus[containerID].CreateNetworkContainerRequest.Version, containerID)
+		}
```
tamilmani1989, line 1289: move to separate function

cns/restserver/restserver.go
```diff
@@ -1252,7 +1263,31 @@ func (service *HTTPRestService) getNetworkContainerResponse(req cns.GetNetworkCo
...
+			dncEP        = service.GetOption(acn.OptPrivateEndpoint).(string)
+			infraVnet    = service.GetOption(acn.OptInfrastructureNetwork).(string)
+			nodeID       = service.GetOption(acn.OptNodeID).(string)
+			isManagedDnc = dncEP != "" && infraVnet != "" && nodeID != ""
+		)
+
+		containerID, exists = service.state.ContainerIDByOrchestratorContext[context]
+		if !exists {
+			if isManagedDnc {
+				service.lock.Unlock()
```
tamilmani1989, line 1278: why do you unlock?..no lock acquired before this code

cns/restserver/restserver.go
```diff
@@ -1252,7 +1263,31 @@ func (service *HTTPRestService) getNetworkContainerResponse(req cns.GetNetworkCo
...
+				getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.SyncNodeStatus(dncEP, infraVnet, nodeID, req.OrchestratorContext)
+				service.lock.Lock()
```
tamilmani1989, line 1280: where do you unlock this.?

cns/NetworkContainerContract.go
```diff
@@ -75,6 +75,7 @@ type CreateNetworkContainerRequest struct {
 	AllowHostToNCCommunication bool
 	AllowNCToHostCommunication bool
 	EndpointPolicies           []NetworkContainerRequestPolicies
+	WaitingForUpdate           bool // True when NC is waiting for NMA to sync versions/rules
```
tamilmani1989, line 78: why this need to be part of cns dnc contract?

jaer-tsun: We need to keep track of whether the NC version is in sync with NMA version

cns/restserver/restserver.go
jaer-tsun: it's actually supposed to be RUnlock and RLock

cns/restserver/restserver.go
jaer-tsun: move what to separate function?

cns/restserver/restserver.go
```diff
@@ -1252,7 +1263,31 @@ func (service *HTTPRestService) getNetworkContainerResponse(req cns.GetNetworkCo
...
 
 		logger.Printf("pod info %+v", podInfo)
-		containerID = service.state.ContainerIDByOrchestratorContext[podInfo.PodName+podInfo.PodNamespace]
+
+		var (
+			context      = podInfo.PodName + podInfo.PodNamespace
+			dncEP        = service.GetOption(acn.OptPrivateEndpoint).(string)
+			infraVnet    = service.GetOption(acn.OptInfrastructureNetwork).(string)
+			nodeID       = service.GetOption(acn.OptNodeID).(string)
+			isManagedDnc = dncEP != "" && infraVnet != "" && nodeID != ""
```
ashvindeodhar, line 1272: Consider a case where partner team starts CNS without specifying infraVnet- We are not throwing error and because of logic on line 1272, we are not going to do things needed for managed dnc. This would go undetected and CNS won't get any data for that NC.
In CNS is used with mdnc, we need to validate these params at cns start and throw and error.

cns/restserver/restserver.go
```diff
@@ -1252,7 +1263,31 @@ func (service *HTTPRestService) getNetworkContainerResponse(req cns.GetNetworkCo
...
+		)
+
+		containerID, exists = service.state.ContainerIDByOrchestratorContext[context]
+		if !exists {
```
ashvindeodhar, line 1276: can we combine the code for isManagedDnc path in one scope?
if isManagedDnc {
    if !exists {
    } else {
    }
}

cns/restserver/restserver.go
```diff
@@ -1252,7 +1263,31 @@ func (service *HTTPRestService) getNetworkContainerResponse(req cns.GetNetworkCo
...
+			if isManagedDnc {
+				service.RUnlock()
+				getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.SyncNodeStatus(dncEP, infraVnet, nodeID, req.OrchestratorContext)
```
ashvindeodhar, line 1279: what happens if we crash in line 1279. in existing defer you would try unlock the lock which you already unlocked on 1278

cns/restserver/restserver.go
```diff
@@ -2055,3 +2115,140 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+
+	// check if the version is valid and save it to service state
+	for _, nc := range ncsToBeAdded {
+		versionURL := fmt.Sprintf(nodeInfoResponse.GetNCVersionURLFmt,
+			nmagentclient.WireserverIP,
+			nc.PrimaryInterfaceIdentifier,
+			nc.NetworkContainerid,
+			nc.AuthorizationToken)
+		versionURLs.Store(nc.NetworkContainerid, versionURL)
+		waitingForUpdate, tmpReturnCode, tmpErrStr := service.isNCWaitingForUpdate(nc.NetworkContainerid, nc.Version)
```
ashvindeodhar, line 2223: change the variable names from tmp to something descriptive of the operation.

cns/restserver/restserver.go
```diff
@@ -2055,3 +2115,140 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+		if tmpErrStr != "" && errStr == "" && bytes.Compare(nc.OrchestratorContext, contextFromCNI) == 0 {
```
ashvindeodhar, line 2224: Add comment explaining the logic. I don't understand what's happening here.

ashvindeodhar: why is errStr == "" check is needed here?

cns/service/main.go
```diff
@@ -187,6 +216,35 @@ func printVersion() {
...
 }
 
+// Try to register node with DNC when CNS is started in managed DNC mode
+func registerNode(httpRestService restserver.HTTPService, dncEP, infraVnet, nodeID string) {
+	var (
+		numCPU   = runtime.NumCPU()
+		response *http.Response
+		err      = fmt.Errorf("")
+		body     bytes.Buffer
+		httpc    = &http.Client{Timeout: time.Second * 30}
```
ashvindeodhar, line 226: reuse the httpclient we already have

cns/service/main.go
```diff
@@ -187,6 +216,35 @@ func printVersion() {
...
+	)
+
+	for err != nil {
+		response, err = httpc.Post(fmt.Sprintf("%s/%s/node/%s/cores/%d%s", dncEP, infraVnet, nodeID, numCPU, dncApiVersion), "application/json", &body)
```
ashvindeodhar, line 230: same comment as before.

cns/service/main.go
```diff
@@ -187,6 +216,35 @@ func printVersion() {
...
+		if err == nil {
+			if response.StatusCode == http.StatusCreated {
+				var req cns.SetOrchestratorTypeRequest
+				json.NewDecoder(response.Body).Decode(&req)
+				httpRestService.SetNodeOrchestrator(&req)
+			} else {
+				logger.Errorf("[Azure CNS] Failed to register node with managed DNC with http status code: " + strconv.Itoa(response.StatusCode))
```
ashvindeodhar, line 237: Can you include node id and infravnetid in the log? It would help when looking at these logs in kusto

cns/service/main.go
```diff
@@ -187,6 +216,35 @@ func printVersion() {
...
+			}
+
+			response.Body.Close()
+		}
+		time.Sleep(time.Second * 5)
```
ashvindeodhar, line 242: why sleep for 5 seconds if err == nil?
You should only sleep if err != nil
Create a const for 5 second sleep.

cns/service/main.go
```diff
@@ -187,6 +216,35 @@ func printVersion() {
...
+	}
+
+	logger.Printf("[Azure CNS] Node Registered")
```
ashvindeodhar, line 245: Can you add node id and infravnetid in this log? looking via kusto will help

cns/service/main.go
```diff
@@ -200,15 +258,18 @@ func main() {
...
+	ipamQueryUrl := acn.GetArg(acn.OptIpamQueryUrl).(string)
+	ipamQueryInterval := acn.GetArg(acn.OptIpamQueryInterval).(int)
 	startCNM := acn.GetArg(acn.OptStartAzureCNM).(bool)
 	vers := acn.GetArg(acn.OptVersion).(bool)
 	createDefaultExtNetworkType := acn.GetArg(acn.OptCreateDefaultExtNetworkType).(string)
 	telemetryEnabled := acn.GetArg(acn.OptTelemetry).(bool)
 	httpConnectionTimeout := acn.GetArg(acn.OptHttpConnectionTimeout).(int)
 	httpResponseHeaderTimeout := acn.GetArg(acn.OptHttpResponseHeaderTimeout).(int)
 	storeFileLocation := acn.GetArg(acn.OptStoreFileLocation).(string)
+	privateEndpoint := acn.GetArg(acn.OptPrivateEndpoint).(string)
```
ashvindeodhar, line 270: can you rename privateEndpoint to also include dnc since this is the dnc endpoint?

cns/service/main.go
```diff
@@ -316,6 +377,22 @@ func main() {
 		go httpRestService.SendNCSnapShotPeriodically(cnsconfig.TelemetrySettings.SnapshotIntervalInMins, stopSnapshots)
 	}
 
+	// If CNS is running on managed DNC mode
+	if privateEndpoint != "" && infravnet != "" && nodeID != "" {
```
ashvindeodhar, line 381: See comment above. How can we know if CNS is running for mdnc mode? We need to validate this and throw error

cns/service/main.go
```diff
@@ -187,6 +216,35 @@ func printVersion() {
...
+				json.NewDecoder(response.Body).Decode(&req)
+				httpRestService.SetNodeOrchestrator(&req)
+			} else {
+				logger.Errorf("[Azure CNS] Failed to register node with managed DNC with http status code: " + strconv.Itoa(response.StatusCode))
+			}
+
+			response.Body.Close()
+		}
+		time.Sleep(time.Second * 5)
+	}
```
ashvindeodhar, line 243: if err != nil let's log an error.

cns/service/main.go
```diff
@@ -316,6 +377,22 @@ func main() {
...
+	if privateEndpoint != "" && infravnet != "" && nodeID != "" {
+		httpRestService.SetOption(acn.OptPrivateEndpoint, privateEndpoint)
+		httpRestService.SetOption(acn.OptInfrastructureNetwork, infravnet)
+		httpRestService.SetOption(acn.OptNodeID, nodeID)
+
+		registerNode(httpRestService, privateEndpoint, infravnet, nodeID)
+		go func(ep, vnet, node string) {
+			// Periodically poll (30s) DNC for node updates
+			for {
+				<-time.NewTicker(time.Second * 30).C
```
ashvindeodhar, line 390: move 30 sec to a const

common/config.go
```diff
@@ -98,4 +98,16 @@ const (
 	// Store file location
 	OptStoreFileLocation      = "store-file-path"
 	OptStoreFileLocationAlias = "storefilepath"
+
+	// Private Endpoint
+	OptPrivateEndpoint      = "private-endpoint"
```
ashvindeodhar, line 103: move to config file

ashvindeodhar:
Can you add the UTs to test register, sync and NMA api? You will need to inject the dependency / mock those components

cns/NetworkContainerContract.go
```diff
@@ -285,3 +285,15 @@ func (networkContainerRequestPolicy *NetworkContainerRequestPolicies) Validate()
...
 }
+
+// NodeInfoResponse - Struct to hold the node info response.
+type NodeInfoResponse struct {
+	NetworkContainers  []CreateNetworkContainerRequest
+	GetNCVersionURLFmt string
+}
+
+// JsonErr - Struct to hold err info
+type JsonErr struct {
```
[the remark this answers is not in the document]
jaer-tsun: removed

cns/NetworkContainerContract.go
```diff
@@ -285,3 +285,15 @@ func (networkContainerRequestPolicy *NetworkContainerRequestPolicies) Validate()
 	}
 	return nil
 }
+
+// NodeInfoResponse - Struct to hold the node info response.
+type NodeInfoResponse struct {
+	NetworkContainers  []CreateNetworkContainerRequest
+	GetNCVersionURLFmt string
```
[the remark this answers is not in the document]
jaer-tsun: DNC holds the general NMA URL fmts

cns/restserver/restserver.go
```diff
@@ -2055,3 +2115,140 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+		return
+	}
+
+	response, err := nmagentclient.GetNetworkContainerVersion(ncid, getNCVersionURL.(string))
+	if err == nil {
+		if response.StatusCode == http.StatusOK {
+			var versionResponse nmagentclient.NetworkContainerResponse
+			rBytes, _ := ioutil.ReadAll(response.Body)
+			json.Unmarshal(rBytes, &versionResponse)
+			if versionResponse.ResponseCode == "200" {
```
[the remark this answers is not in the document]
jaer-tsun: one's a string and the latter's an int

cns/restserver/restserver.go
```diff
@@ -2055,3 +2115,140 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+				returnCode = UnexpectedError
+				message = fmt.Sprintf("[Azure-CNS] Failed to get NC version from response %s for NC %s", rBytes, ncid)
+			}
+		} else {
+			returnCode = UnexpectedError
+			message = fmt.Sprintf("[Azure-CNS] Failed to get NC version with http status %d", response.StatusCode)
+		}
+	}
+
+	return
```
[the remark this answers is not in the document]
jaer-tsun: what do you mean?

cns/restserver/restserver.go
```diff
@@ -2055,3 +2115,140 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+
+// SetNodeOrchestrator :- Set node orchestrator after registering with mDNC
+func (service *HTTPRestService) SetNodeOrchestrator(r *cns.SetOrchestratorTypeRequest) {
+	body, _ := json.Marshal(r)
+	req, _ := http.NewRequest(http.MethodPost, "", bytes.NewBuffer(body))
+	req.Header.Set("Content-Type", "application/json; charset=UTF-8")
+	service.setOrchestratorType(httptest.NewRecorder(), req)
+}
+
+// SyncNodeStatus :- Retrieve the latest node state from DNC & returns the first occurence of returnCode and error with respect to contextFromCNI
```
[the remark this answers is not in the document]
jaer-tsun: what do you mean?

cns/restserver/restserver.go
```diff
@@ -2055,3 +2115,140 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+		response         *http.Response
+		err              error
+		nodeInfoResponse cns.NodeInfoResponse
+		req              *http.Request
+		body             []byte
+		httpc            = &http.Client{Timeout: time.Second * 30}
+	)
+
+	// try to retrieve NodeInfoResponse from mDNC
+	response, err = httpc.Get(fmt.Sprintf("%s/%s/node/%s%s", dncEP, infraVnet, nodeID, dncApiVersion))
```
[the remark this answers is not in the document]
jaer-tsun: I only do that when the lines are too long

cns/restserver/restserver.go
jaer-tsun: but these are actually tmp...

cns/restserver/restserver.go
jaer-tsun: removed

cns/service/main.go
```diff
@@ -179,6 +187,27 @@ var args = acn.ArgumentList{
 		Type:         "string",
 		DefaultValue: platform.CNMRuntimePath,
 	},
+	{
```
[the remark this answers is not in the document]
jaer-tsun: we can do both

cns/service/main.go
jaer-tsun: fixed the logic and 5s sleep is only used here

cns/service/main.go
jaer-tsun: what do you mean, I think it's aptly named in this instance

cns/service/main.go
jaer-tsun: it's only used here

common/config.go
jaer-tsun: not moving, we can add to config

cns/restserver/restserver.go
jaer-tsun: if it crashes then the unlock will be a runtime err

cns/restserver/restserver.go
```diff
@@ -2055,3 +2115,140 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+	if err != nil {
+		returnCode = UnexpectedError
+		errStr = fmt.Sprintf("[Azure-CNS] Failed to sync node with error: %+v", err)
+		logger.Errorf(errStr)
+		return
+	}
+
+	var (
+		ncsToBeAdded   = make(map[string]cns.CreateNetworkContainerRequest)
+		ncsToBeDeleted = make(map[string]bool)
```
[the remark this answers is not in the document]
jaer-tsun: not sure what you mean

we add all the NCs to ncsToBeDeleted and then we remove from the map based on DNC node state

cns/restserver/restserver.go
```diff
@@ -2055,3 +2115,140 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+	// determine new NCs and NCs to be deleted
+	service.lock.Lock()
+	for ncid := range service.state.ContainerStatus {
+		ncsToBeDeleted[cns.SwiftPrefix+ncid] = true
+	}
+
+	for _, nc := range nodeInfoResponse.NetworkContainers {
+		ncid := cns.SwiftPrefix + nc.NetworkContainerid
+		delete(ncsToBeDeleted, ncid)
+		if savedNc, exists := service.state.ContainerStatus[ncid]; !exists || savedNc.CreateNetworkContainerRequest.Version < nc.Version {
```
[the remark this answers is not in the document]
jaer-tsun: the code itself is explanatory enough here no?

if NC doesn't exist or if existing NC version is not up to date with programmed version add it to ncsToBeAdded

cns/restserver/restserver.go
```diff
@@ -2055,3 +2115,140 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+
+	if err != nil {
+		returnCode = UnexpectedError
+		errStr = fmt.Sprintf("[Azure-CNS] Failed to sync node with error: %+v", err)
+		logger.Errorf(errStr)
+		return
+	}
+
+	var (
+		ncsToBeAdded   = make(map[string]cns.CreateNetworkContainerRequest)
```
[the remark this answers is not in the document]
jaer-tsun: they're NCs to be added or updated

cns/restserver/restserver.go
msvik: I agree 100%. Instead of "calculating" that we are in managed dnc mode, pass in an explicit argument that it is managed DNC mode and then validate all the necessary arugments that are required for that mode.

cns/restserver/restserver.go
```diff
@@ -2055,3 +2122,143 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+	}
+
+	response, err := nmagentclient.GetNetworkContainerVersion(ncid, getNCVersionURL.(string))
+	if err == nil {
+		if response.StatusCode == http.StatusOK {
+			var versionResponse nmagentclient.NetworkContainerResponse
+			rBytes, _ := ioutil.ReadAll(response.Body)
+			json.Unmarshal(rBytes, &versionResponse)
+			if versionResponse.ResponseCode == "200" {
+				programmedVersion, _ := strconv.Atoi(ncVersion)
```
ashvindeodhar, line 2142: programmedVersion means 'programmed vfp' -> this is the one nma has. You meant expectedVersion?

cns/restserver/restserver.go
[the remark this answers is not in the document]
ashvindeodhar: The default value for waitingForUpdate is false. If NMA api returns an error ( err != nil ), you just return on 2159.
This would assume that vfp programming is done, no?

cns/restserver/restserver.go
```diff
@@ -2055,3 +2122,143 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+	// check if the version is valid and save it to service state
+	for _, nc := range ncsToBeAdded {
+		versionURL := fmt.Sprintf(nodeInfoResponse.GetNCVersionURLFmt,
+			nmagentclient.WireserverIP,
+			nc.PrimaryInterfaceIdentifier,
+			nc.NetworkContainerid,
+			nc.AuthorizationToken)
+		ncVersionURLs.Store(nc.NetworkContainerid, versionURL)
+		waitingForUpdate, tmpReturnCode, tmpErrStr := service.isNCWaitingForUpdate(nc.NetworkContainerid, nc.Version)
+		if tmpReturnCode != Success && bytes.Compare(nc.OrchestratorContext, contextFromCNI) == 0 {
```
ashvindeodhar, line 2230: can you add comment here in the code?
I don't understand what's happening here
What are we doing with byte compare?
if tmpReturnCode != Success && bytes.Compare(nc.OrchestratorContext, contextFromCNI) == 0

cns/service/main.go
```diff
@@ -187,6 +216,38 @@ func printVersion() {
...
+
+	for err != nil {
+		response, err = httpc.Post(fmt.Sprintf("%s/%s/node/%s/cores/%d%s", dncEP, infraVnet, nodeID, numCPU, dncApiVersion), "application/json", &body)
+		if err == nil {
+			if response.StatusCode == http.StatusCreated {
+				var req cns.SetOrchestratorTypeRequest
+				json.NewDecoder(response.Body).Decode(&req)
+				httpRestService.SetNodeOrchestrator(&req)
+			} else {
+				logger.Errorf("[Azure CNS] Failed to register node %s/%s with managed DNC with http status code %s", infraVnet, nodeID, strconv.Itoa(response.StatusCode))
```
ashvindeodhar, line 237: looks like there is bug here - when err == nil you break out of the for loop. But you may have an error from line 237. 
This means the registration failed. In the CNS main file, you are not checking the error but going ahead with the sync loop. We need to make sure the registration succeeds before moving forward.

cns/service/main.go
```diff
@@ -187,6 +216,38 @@ func printVersion() {
...
+			}
+
+			response.Body.Close()
+			time.Sleep(time.Second * 5)
```
ashvindeodhar, line 241: My original comment was to not sleep in the err == nil. It's still the case. Maybe we can discuss this.

cns/service/main.go
```diff
@@ -187,6 +216,38 @@ func printVersion() {
...
+		} else {
+			logger.Errorf("[Azure CNS] Failed to register node %s/%s with err: %+v", infraVnet, nodeID, err)
+			return
```
ashvindeodhar, line 244: Do you intend to return here? If you return, you are not retrying when err is nil. I think you added return here by mistake.

cns/service/main.go
ashvindeodhar: Can you respond to the comment saying 'this is why I don't want to address it'  if you don't intend to address it?

cns/service/main.go
ashvindeodhar: isn't that PE specific to dnc? I just thought having dnc in the name would clarify that.

cns/service/main.go
ashvindeodhar: my comment in general is to not hardcode values like 5 seconds, 30 seconds. Instead declare those as variable like syncNodeInterval.

cns/service/main.go
```diff
@@ -316,6 +380,33 @@ func main() {
 		go httpRestService.SendNCSnapShotPeriodically(cnsconfig.TelemetrySettings.SnapshotIntervalInMins, stopSnapshots)
 	}
 
+	// If CNS is running on managed DNC mode
+	if privateEndpoint != "" && infravnet != "" && nodeID != "" {
```
ashvindeodhar, line 384: Let's not decide if we need to use mdnc based on privateEndpoint != "" && infravnet != "" && nodeID != ""
Let's add a flag in config and based on that decide. The logic becomes simpler and clear intent is provided to CNS from client

common/config.go
ashvindeodhar: If we keep on providing them commandline params, no one will adopt the config file usage. The decision was to stop adding the params to commandline and have those in config instead.

cns/restserver/restserver.go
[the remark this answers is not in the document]
ashvindeodhar: added comment below where you are comparing byte array. Resolving this.

cns/NetworkContainerContract.go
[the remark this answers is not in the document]
jaer-tsun: Since DNC held the other formats I've left it on that side

cns/restserver/restserver.go
jaer-tsun: programmed = published version

NMA needs to sync with published version

cns/restserver/restserver.go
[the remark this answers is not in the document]
jaer-tsun: fixed, added err msg and returncode

cns/restserver/restserver.go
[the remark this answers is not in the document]
jaer-tsun: the map is used so that the key can be deleted after looping through NCs on node

cns/restserver/restserver.go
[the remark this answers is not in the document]
jaer-tsun: in this instance, I think the code is concise enough to avoid over commenting

cns/restserver/restserver.go
jaer-tsun: It's described in function comments

// SyncNodeStatus :- Retrieve the latest node state from DNC & returns the first occurence of returnCode and error with respect to contextFromCNI

cns/service/main.go
jaer-tsun: it goes from CNS->PE->PL->ILB->DNC, from CNS POV it just talks to PE as if it's DNC

common/config.go
jaer-tsun: I don't remember that decision, we can chat about this

cns/service/main.go
jaer-tsun: removed

cns/service/main.go
jaer-tsun: fixed

cns/service/main.go
jaer-tsun: fixed

cns/service/main.go
ashvindeodhar: I just finished reviewing the DNC PR. We are retrieving all the NCs every 30 seconds. let's expose this as a configurable param via config file. If we see we are hitting cosmosdb RU limit due to NC processing in DNC we can increase this w/o needing to update the code.

cns/restserver/api.go
```diff
@@ -1415,3 +1419,107 @@ func (service *HTTPRestService) deleteHostNCApipaEndpoint(w http.ResponseWriter,
 	err = service.Listener.Encode(w, &response)
 	logger.Response(service.Name, response, response.Response.ReturnCode, ReturnCodeToString(response.Response.ReturnCode), err)
 }
+
+// SetNodeOrchestrator :- Set node orchestrator after registering with mDNC
+func (service *HTTPRestService) SetNodeOrchestrator(r *cns.SetOrchestratorTypeRequest) {
```
neaggarwMS, line 1424: Looks like this is called internally by CNS. Can you move this to internalapi.go? This file only captures HTTP APIs exposed to external clients/components

cns/restserver/api.go
```diff
@@ -1415,3 +1419,107 @@ func (service *HTTPRestService) deleteHostNCApipaEndpoint(w http.ResponseWriter,
...
+	body, _ := json.Marshal(r)
+	req, _ := http.NewRequest(http.MethodPost, "", bytes.NewBuffer(body))
+	req.Header.Set("Content-Type", "application/json; charset=UTF-8")
+	service.setOrchestratorType(httptest.NewRecorder(), req)
+}
+
+// SyncNodeStatus :- Retrieve the latest node state from DNC & returns the first occurence of returnCode and error with respect to contextFromCNI
+func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string, contextFromCNI json.RawMessage) (returnCode int, errStr string) {
```
neaggarwMS, line 1432: Similar for this function as well

cns/common/service.go
```diff
@@ -35,20 +36,22 @@ type ServiceConfig struct {
 	Listener *acn.Listener
 	ErrChan  chan error
 	Store    store.KeyValueStore
+	Managed  bool
```
neaggarwMS, line 39: Instead of defining a bool, can we define it as ChannelType? (with values as Direct, Managed) Later that can be extended to CRD as well

cns/restserver/api.go
jaer-tsun: moved

cns/restserver/api.go
jaer-tsun: moved

cns/NetworkContainerContract.go
jaer-tsun: moved

cns/common/service.go
jaer-tsun: I have no context on this, why channel?

cns/restserver/api.go
```diff
@@ -1415,3 +1419,107 @@ func (service *HTTPRestService) deleteHostNCApipaEndpoint(w http.ResponseWriter,
...
+		waitingForUpdate, tmpReturnCode, tmpErrStr := isNCWaitingForUpdate(nc.NetworkContainerid, nc.Version)
+		if tmpReturnCode != Success && bytes.Compare(nc.OrchestratorContext, contextFromCNI) == 0 {
+			returnCode = tmpReturnCode
+			errStr = tmpErrStr
+			logger.Errorf(tmpErrStr)
+		}
+
+		if tmpReturnCode == UnexpectedError {
+			continue
+		}
```
tamilmani1989: was this case handled?  tmpReturnCode !=success && tmpReturnCode !=unexpected error && bytes.Compare(nc.OrchestratorContext, contextFromCNI) != 0

cns/restserver/api.go
```diff
@@ -1415,3 +1419,107 @@ func (service *HTTPRestService) deleteHostNCApipaEndpoint(w http.ResponseWriter,
...
+
+		nc.WaitingForUpdate = waitingForUpdate
+		body, _ = json.Marshal(nc)
+		req, _ = http.NewRequest(http.MethodPost, "", bytes.NewBuffer(body))
```
tamilmani1989: i think you don;t need to call http within process...neha s implementing internal api calls,..sync with her

cns/restserver/util.go
```diff
@@ -201,7 +206,34 @@ func (service *HTTPRestService) getNetworkContainerResponse(req cns.GetNetworkCo
...
 
 		logger.Printf("pod info %+v", podInfo)
-		containerID = service.state.ContainerIDByOrchestratorContext[podInfo.PodName+podInfo.PodNamespace]
+
+		var (
+			context   = podInfo.PodName + podInfo.PodNamespace
+			dncEP     = service.GetOption(acn.OptPrivateEndpoint).(string)
+			infraVnet = service.GetOption(acn.OptInfrastructureNetwork).(string)
+			nodeID    = service.GetOption(acn.OptNodeID).(string)
+			managed   = service.GetOption(acn.OptManaged).(bool)
```
tamilmani1989, line 215: can we do this one time..when service starts and save it in memory..instead of doing at multiple places

cns/restserver/util.go
```diff
@@ -299,9 +331,35 @@ func (service *HTTPRestService) attachOrDetachHelper(req cns.ConfigureContainerN
...
+		} else {
+			returnCode, msg := service.SyncNodeStatus(dncEP, infraVnet, nodeID, json.RawMessage{})
+			if returnCode != Success {
+				return cns.Response{
+					ReturnCode: returnCode,
+					Message:    msg}
+			}
 
-	if !ok {
+			existing, _ = service.getNetworkContainerDetails(cns.SwiftPrefix + req.NetworkContainerid)
```
tamilmani1989, line 360: dont we need to call isNCWaitingForUpdate after this?

cns/restserver/util.go
```diff
@@ -441,6 +499,45 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
 	}
 }
 
+// isNCWaitingForUpdate :- Determine whether NC version on NMA matches programmed version
+func isNCWaitingForUpdate(ncVersion, ncid string) (waitingForUpdate bool, returnCode int, message string) {
+	getNCVersionURL, ok := ncVersionURLs.Load(ncid)
```
tamilmani1989, line 504: can you add log at start of function with ncid and ncversion

cns/restserver/util.go
```diff
@@ -441,6 +499,45 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
...
+			rBytes, _ := ioutil.ReadAll(response.Body)
+			json.Unmarshal(rBytes, &versionResponse)
+			if versionResponse.ResponseCode == "200" {
+				programmedVersion, _ := strconv.Atoi(ncVersion)
+				nmaVersion, _ := strconv.Atoi(versionResponse.Version)
+				if programmedVersion > nmaVersion {
+					waitingForUpdate = true
+					returnCode = NetworkContainerPendingStatePropagation
+					message = fmt.Sprintf("[Azure-CNS] Network container %s v%d had not propagated to respective NMA w/ v%d", ncid, programmedVersion, nmaVersion)
+				}
```
tamilmani1989, line 524: do we need to log if programmedversion < nmaversion?

cns/service/main.go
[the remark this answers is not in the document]
tamilmani1989: can we just put it in config? whats the reason for having both?

cns/service/main.go
tamilmani1989: i think he added in start of function

cns/service/main.go
tamilmani1989: i think ashvin point here is it would be more readable ..by looking at code tomorrow anyone can get some sense this PE is intended for dnc communication

cns/nmagentclient/nmagentclient.go
```diff
@@ -62,3 +73,16 @@ func UnpublishNetworkContainer(
 
 	return response, err
 }
+
+// GetNetworkContainerVersion :- Retrieves NC version from NMAgent
+func GetNetworkContainerVersion(
+	networkContainerID,
```
tamilmani1989, line 79: where this networkContainerID used?

cns/restserver/restserver.go
tamilmani1989: github doesnt have good gui...

`Comment on lines 1266 to 1289`

cns/common/service.go
tamilmani1989: its not go channel..she s asking you to define enum of strings..so that it can take multiple values instead of bool

cns/nmagentclient/nmagentclient.go
jaer-tsun: in the log

cns/nmagentclient/nmagentclient.go
```diff
@@ -62,3 +73,16 @@ func UnpublishNetworkContainer(
...
+	getNetworkContainerVersionURL string) (*http.Response, error) {
+	logger.Printf("[NMAgentClient] GetNetworkContainerVersion NC: %s", networkContainerID)
+
+	response, err := common.GetHttpClient().Get(getNetworkContainerVersionURL)
```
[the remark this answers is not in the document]
jaer-tsun: it's init at start

cns/restserver/util.go
jaer-tsun: no it's already called in SyncNodeStatus

cns/restserver/util.go
jaer-tsun: either the caller will log or GetNetworkContainerVersion will

cns/restserver/util.go
jaer-tsun: no, CNS should be receiving latest

cns/restserver/util.go
jaer-tsun: the options are set after HTTPRestService has been init

cns/common/service.go
jaer-tsun: in dnc there's already direct and privatelink channel modes (for cns), are they related to this at all?

e.g. managed here could mean privatelink on dnc side but it may not be guaranteed

cns/configuration/configuration.go
```diff
@@ -17,6 +18,8 @@ const (
 
 type CNSConfig struct {
 	TelemetrySettings TelemetrySettings
+	ManagedSettings   ManagedSettings
+	ChannelMode       string
```
ashvindeodhar, line 22: is ChannelMode supposed to be specified in config file? I don't see it in the config above.

cns/nmagentclient/nmagentclient.go
```diff
@@ -9,6 +9,17 @@ import (
 	"github.com/Azure/azure-container-networking/common"
 )
 
+const (
+	WireserverIP = "168.63.129.16"
+)
+
+// NetworkContainerResponse - NMAgent response.
+type NetworkContainerResponse struct {
```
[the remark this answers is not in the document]
ashvindeodhar: +1. NMANetworkContainerVersionResponse

cns/restserver/internalapi.go
```diff
@@ -14,3 +27,123 @@ func (service *HTTPRestService) GetPartitionKey() (dncPartitionKey string) {
...
+
+		ncVersionURLs.Store(nc.NetworkContainerid, versionURL)
+		waitingForUpdate, tmpReturnCode, tmpErrStr := isNCWaitingForUpdate(nc.Version, nc.NetworkContainerid)
+		if tmpReturnCode != Success && bytes.Compare(nc.OrchestratorContext, contextFromCNI) == 0 {
+			returnCode = tmpReturnCode
+			errStr = tmpErrStr
+		}
+
+		if tmpReturnCode == UnexpectedError {
+			continue
```
ashvindeodhar, line 109: log the error

cns/restserver/internalapi.go
```diff
@@ -14,3 +27,123 @@ func (service *HTTPRestService) GetPartitionKey() (dncPartitionKey string) {
...
+				nmagentclient.WireserverIP,
+				nc.PrimaryInterfaceIdentifier,
+				nc.NetworkContainerid,
+				nc.AuthorizationToken)
+			w = httptest.NewRecorder()
+		)
+
+		ncVersionURLs.Store(nc.NetworkContainerid, versionURL)
+		waitingForUpdate, tmpReturnCode, tmpErrStr := isNCWaitingForUpdate(nc.Version, nc.NetworkContainerid)
+		if tmpReturnCode != Success && bytes.Compare(nc.OrchestratorContext, contextFromCNI) == 0 {
```
ashvindeodhar, line 103: what's the logic behind bytes compare here? Pls add comment in the code

cns/service/main.go
```diff
@@ -187,6 +223,44 @@ func printVersion() {
...
+	for sleep := true; err != nil; sleep = true {
+		response, err = httpc.Post(url, "application/json", &body)
+		if err == nil {
+			if response.StatusCode == http.StatusCreated {
+				var req cns.SetOrchestratorTypeRequest
+				json.NewDecoder(response.Body).Decode(&req)
+				httpRestService.SetNodeOrchestrator(&req)
+				sleep = false
+			} else {
+				logger.Errorf("[Azure CNS] Failed to register node with http status code %s", strconv.Itoa(response.StatusCode))
```
ashvindeodhar, line 248: This is a bug: for loop exits even if registration fails on line 248.

cns/service/main.go
```diff
@@ -187,6 +223,44 @@ func printVersion() {
...
+	var (
+		numCPU   = runtime.NumCPU()
+		url      = fmt.Sprintf(acn.RegisterNodeURLFmt, dncEP, infraVnet, nodeID, numCPU, dncApiVersion)
+		response *http.Response
+		err      = fmt.Errorf("")
+		body     bytes.Buffer
+		httpc    = acn.GetHttpClient()
+	)
+
+	for sleep := true; err != nil; sleep = true {
```
ashvindeodhar, line 239: nitpick:
sleep := true
for err != nil {
...
}

cns/service/main.go
```diff
@@ -187,6 +223,44 @@ func printVersion() {
...
+				json.NewDecoder(response.Body).Decode(&req)
+				httpRestService.SetNodeOrchestrator(&req)
+				sleep = false
+			} else {
+				logger.Errorf("[Azure CNS] Failed to register node with http status code %s", strconv.Itoa(response.StatusCode))
+			}
+
+			response.Body.Close()
+		} else {
+			logger.Errorf("[Azure CNS] Failed to register node with err: %+v", err)
```
ashvindeodhar, line 253: I had mentioned this in earlier review comments - 
Can you add infravnet guid and node id in the error messages in register and sync funcs? It will help when we are debugging the issue from kusto logs

jaer-tsun: it's already there

logger.Printf("[Azure CNS] Registering node %s with Infrastructure Network: %s PrivateEndpoint: %s", nodeID, infraVnet, dncEP)

cns/restserver/internalapi.go
jaer-tsun: this err is returned to cni specifically

cns/restserver/internalapi.go
jaer-tsun: not sure what you mean, the func def describes this already

// SyncNodeStatus :- Retrieve the latest node state from DNC & returns the first occurence of returnCode and error with respect to contextFromCNI

cns/nmagentclient/nmagentclient.go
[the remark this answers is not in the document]
jaer-tsun: renamed

cns/configuration/configuration.go
jaer-tsun: if you're referring to cns_config.json then I just added it

cns/service/main.go
ashvindeodhar: I am talking about the error logs. Due to absence of that info in error log / log which says "Node Registered", we wont know what happened after 'Registering node'.

cns/restserver/internalapi.go
```diff
@@ -14,3 +27,123 @@ func (service *HTTPRestService) GetPartitionKey() (dncPartitionKey string) {
...
+		ncid := cns.SwiftPrefix + nc.NetworkContainerid
+		delete(ncsToBeDeleted, ncid)
+		if savedNc, exists := service.state.ContainerStatus[ncid]; !exists || savedNc.CreateNetworkContainerRequest.Version < nc.Version {
+			ncsToBeAdded[ncid] = nc
+		}
+	}
+	service.RUnlock()
+
+	// check if the version is valid and save it to service state
+	for ncid, nc := range ncsToBeAdded {
```
neaggarwMS, line 91: Can we compare the version with the cached version in CNS to avoid idempotent calls to NMagent? This is valid in case there is no change in NC state.

jaer-tsun: it's checked before being added to ncsToBeAdded

cns/restserver/internalapi.go
```diff
@@ -14,3 +27,123 @@ func (service *HTTPRestService) GetPartitionKey() (dncPartitionKey string) {
...
+			errStr = tmpErrStr
+		}
+
+		if tmpReturnCode == UnexpectedError {
+			continue
+		}
+
+		body, _ = json.Marshal(nc)
+		req, _ = http.NewRequest(http.MethodPost, "", bytes.NewBuffer(body))
+		req.Header.Set("Content-Type", "application/json; charset=UTF-8")
```
tamilmani1989, line 114: i see this used in multiple places..can we move this to constant
"Content-Type", "application/json; charset=UTF-8"

cns/restserver/util.go
```diff
@@ -201,7 +206,33 @@ func (service *HTTPRestService) getNetworkContainerResponse(req cns.GetNetworkCo
...
+			if exists {
+				_, getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = isNCWaitingForUpdate(service.state.ContainerStatus[containerID].CreateNetworkContainerRequest.Version, containerID)
+				if getNetworkContainerResponse.Response.ReturnCode == Success {
+					return getNetworkContainerResponse
+				}
+			} else {
+				var (
+					dncEP     = service.GetOption(acn.OptPrivateEndpoint).(string)
+					infraVnet = service.GetOption(acn.OptInfrastructureNetwork).(string)
+					nodeID    = service.GetOption(acn.OptNodeID).(string)
```
tamilmani1989, line 222: are we not using config?

cns/restserver/util.go
```diff
@@ -300,8 +331,33 @@ func (service *HTTPRestService) attachOrDetachHelper(req cns.ConfigureContainerN
...
+				}
+			}
+		} else {
+			var (
+				dncEP     = service.GetOption(acn.OptPrivateEndpoint).(string)
+				infraVnet = service.GetOption(acn.OptInfrastructureNetwork).(string)
+				nodeID    = service.GetOption(acn.OptNodeID).(string)
+			)
+
+			returnCode, msg := service.SyncNodeStatus(dncEP, infraVnet, nodeID, json.RawMessage{})
```
tamilmani1989, line 351: can you add comment why are we passing empty podcontext

cns/restserver/util.go
jaer-tsun: we do use config, those values are passed to the service

cns/restserver/util.go
jaer-tsun: only CNI passes non-empty, there's a func comment about this already

cns/restserver/internalapi.go
```diff
@@ -14,3 +27,123 @@ func (service *HTTPRestService) GetPartitionKey() (dncPartitionKey string) {
...
+		ncVersionURLs.Store(nc.NetworkContainerid, versionURL)
+		waitingForUpdate, tmpReturnCode, tmpErrStr := isNCWaitingForUpdate(nc.Version, nc.NetworkContainerid)
+		if tmpReturnCode != Success && bytes.Compare(nc.OrchestratorContext, contextFromCNI) == 0 {
+			returnCode = tmpReturnCode
+			errStr = tmpErrStr
+		}
+
+		if tmpReturnCode == UnexpectedError {
+			continue
+		}
```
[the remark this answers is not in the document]
jaer-tsun: then that means it's NetworkContainerPendingStatePropagation; in this case, the err gets returned to CNI and CNS will still save the state

cns/restserver/internalapi.go
jaer-tsun: fixed

cns/configuration/configuration.go
```diff
@@ -44,6 +47,13 @@ type TelemetrySettings struct {
 	SnapshotIntervalInMins int
 }
 
+type ManagedSettings struct {
+	PrivateEndpoint           string
+	InfrastructureNetwork     string
```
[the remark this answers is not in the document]
jaer-tsun: added ID, and it takes either IP or DNS, that's why it's just Endpoint

cns/service/main.go
```diff
@@ -200,15 +275,18 @@ func main() {
...
 	startCNM := acn.GetArg(acn.OptStartAzureCNM).(bool)
 	vers := acn.GetArg(acn.OptVersion).(bool)
 	createDefaultExtNetworkType := acn.GetArg(acn.OptCreateDefaultExtNetworkType).(string)
 	telemetryEnabled := acn.GetArg(acn.OptTelemetry).(bool)
 	httpConnectionTimeout := acn.GetArg(acn.OptHttpConnectionTimeout).(int)
 	httpResponseHeaderTimeout := acn.GetArg(acn.OptHttpResponseHeaderTimeout).(int)
 	storeFileLocation := acn.GetArg(acn.OptStoreFileLocation).(string)
+	privateEndpoint := acn.GetArg(acn.OptPrivateEndpoint).(string)
+	infravnet := acn.GetArg(acn.OptInfrastructureNetworkID).(string)
+	nodeID := acn.GetArg(acn.OptNodeID).(string)
```
tamilmani1989, line 289: why do we need this? we are going to get these from config anyway

tamilmani1989: checked with jaeryn..we should not advise customer to use cmdline args..its just for internal testing and it should be removed if not needed

cns/configuration/configuration.go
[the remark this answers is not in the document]
tamilmani1989: yes but can add a comment there

PR #577, opened 2020-06-10
CNS controller for watching and updating networkconfig CRD

pjohnst5:
This is the implementation of watching CRDs for the CNS controller.

Please give me feedback as this is a beginning point and will be much easier to change things now than in the future.

### DIFF
--- a/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
+++ b/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
@@ -25,7 +25,7 @@ import (
 // +kubebuilder:object:root=true
 
 // NodeNetworkConfig is the Schema for the nodenetworkconfigs API
-// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:resource:scope=Namespaced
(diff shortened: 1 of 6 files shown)


nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -25,7 +25,7 @@ import (
 // +kubebuilder:object:root=true
 
 // NodeNetworkConfig is the Schema for the nodenetworkconfigs API
-// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:resource:scope=Namespaced
```
thatmattlong, line 28: just a heads up with this change, if you have a cluster laying around with the crd already loaded in (like me with my local minikube for example) you'll have to `kubectl delete crd acn.azure.com/NodeNetworkConfigs` or similar, since you can't change a CRD from cluster-scope to namespace-scope.

pjohnst5: Thanks, I'm going to review your PR btw soon

PR #578, opened 2020-06-10
Simplify issue template

mainred:
I find people reporting issues in ths repo tend to not follow exactly the template, and I personal fell it's a bit complicated in its structure, so I simply the issue template to make it more report friendly.

**Release note**:

```release-note
```


mainred:
> It looks like the gh issue templating has changed a little bit, perhaps we should adhere to those while we're making this change? Separate templates for CNI, CNS, and NPM issues?

Excellent point, that was also my consideration to make less if-else explanation. But, considering  issue and feature request for different components share the same items, for now, to be maintained

I am looking forward to split different components into different repos :)

mainred:
@matmerr , thanks for your idea. I followed your idea and used issue builder to make one issue more clearly and focused.

matmerr:
I took these changes and refined it a little more in https://github.com/Azure/azure-container-networking/pull/592. Thanks @mainred!

PR #583, opened 2020-06-17
Request controller for CNS

pjohnst5:
CNS Request Controller for the AKS on Swift scenario

 with an example of how to call the methods in cns/examplerequestcontroller/example/main.go


cns/requestcontroller/example/main.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/Azure/azure-container-networking/cns/restserver"
+)
+
+func goRequestController(rc requestcontroller.RequestController) {
+	//Start the RequestController which starts the reconcile loop
+	if err := rc.StartRequestController(); err != nil {
+		logger.Errorf("Error starting requestController: %v", err)
+	}
+	// After calling StartRequestController, there needs to be some pause or, the calling program
+	// must keep running because the reconcile loop is spawned off on a different go-routine inside
```
paulgmiller, line 18: I a real world example we'd want soemthign that would block until a control c signal right?

cns/requestcontroller/kubernetes/k8srequestcontroller.go
```diff
@@ -0,0 +1,175 @@
+package kubernetes
```
paulgmiller, line 1: are you thinking of having request controllers for other orchestrators? or just a test one?

cns/requestcontroller/kubernetes/k8srequestcontroller.go
```diff
@@ -0,0 +1,175 @@
...
+
+const k8sNamespace = "kube-system"
+
+// k8sRequestController watches CRDs for status updates and publishes CRD spec changes
+// mgr acts as the communication between requestController and API server
+// implements the RequestController interface
+type k8sRequestController struct {
+	mgr         manager.Manager             //mgr has method GetClient() to get k8s client
+	Client      K8sClient                   //Relying on the K8sClient interface to more easily test
+	restService *restserver.HTTPRestService //restService is given to nodeNetworkConfigReconciler (the reconcile loop)
```
paulgmiller, line 26: why does the reconcile loop need a server?

cns/requestcontroller/kubernetes/k8srequestcontroller.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Check that HOSTNAME environment variable is set. HOSTNAME is name of node running this program
+	hostName := os.Getenv("HOSTNAME")
+	if hostName == "" {
+		return nil, errors.New("Must declare HOSTNAME environment variable. HOSTNAME is name of node.")
+	}
+
+	//Add CRD scheme to runtime sheme so manager can recognize it
+	var scheme = runtime.NewScheme()
+	_ = clientgoscheme.AddToScheme(scheme)
```
paulgmiller, line 46: are these error we're ignoring?

cns/requestcontroller/example/main.go
paulgmiller: Okay this is doing that already. 
k8sRC.mgr.Start(ctrl.SetupSignalHandler()); err != nil {

Any other reason we'd send a cancel signal outside of termination? Do we want to wait for mgr to finish up before we exit (an exit channel might be useful?

cns/requestcontroller/kubernetes/nodenetworkconfigreconciler.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Reconcile relays status changes in NodeNetworkConfig to CNS
+// Returning non-nil error causes a requeue
+// Returning ctrl.Result{}, nil causes the queue to "forget" the item
+// Other return values are possible, see kubebuilder docs for details
+func (n *NodeNetworkConfigReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	// We are only interested in requests coming for the node that this program is running on
+	// Requeue if it's not for this node
+	if request.Name != n.HostName {
+		return reconcile.Result{}, errors.New("Requeing")
```
paulgmiller, line 30: we don't actually want to requeue here.

cns/requestcontroller/kubernetes/eventfilter.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// If the generations are the same, it means it's status change, and we should return true, so that the
+// reconcile loop is triggered by it.
+// If they're different, it means a spec change, and we should ignore, by returning false, to avoid redundant calls to cns when the
+// status hasn't changed
+// See https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource
+// for more details
+func (NodeNetworkConfigFilter) Update(e event.UpdateEvent) bool {
+	oldGeneration := e.MetaOld.GetGeneration()
```
paulgmiller, line 19: can you filter hostname here?

cns/requestcontroller/kubernetes/nodenetworkconfigreconciler.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	var nodeNetConfig nnc.NodeNetworkConfig
+
+	//Get the CRD object
+	if err := n.K8sClient.Get(context.TODO(), request.NamespacedName, &nodeNetConfig); err != nil {
```
paulgmiller, line 36: is this working off the cached one in the queue or going out to the api server? Prefer the former but not clear.

cns/requestcontroller/kubernetes/k8srequestcontroller.go
```diff
@@ -0,0 +1,175 @@
...
+		if err := k8sRC.mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+			logger.Errorf("[cns-rc] Error starting manager: %v", err)
+		}
+	}()
+
+	return nil
+}
+
+// ReleaseIPsByUUIDs sends release ip request to the API server. Provide the UUIDs of the IP allocations
+func (k8sRC *k8sRequestController) ReleaseIPsByUUIDs(listOfIPUUIDS []string) error {
```
paulgmiller, line 113: plumb context through this and UpdateRequestedIP count so cancelation become easier in the future?

cns/requestcontroller/kubernetes/k8srequestcontroller.go
```diff
@@ -0,0 +1,175 @@
...
+	nodeNetworkConfig, err := k8sRC.getNodeNetConfig(k8sRC.hostName, k8sNamespace)
+	if err != nil {
+		logger.Errorf("[cns-rc] Error getting CRD when releasing IPs by uuid %V", err)
+		return err
+	}
+
+	//Update the CRD IpsNotInUse
+	nodeNetworkConfig.Spec.IPsNotInUse = append(nodeNetworkConfig.Spec.IPsNotInUse, listOfIPUUIDS...)
```
paulgmiller, line 121: do we do seperate calls to deprecate requested ip count at the same time we do this usually? Worth combining?

cns/requestcontroller/kubernetes/k8srequestcontroller.go
```diff
@@ -0,0 +1,175 @@
...
+
+	//Send update to API server
+	if err := k8sRC.updateNodeNetConfig(nodeNetworkConfig); err != nil {
```
weng271190436, line 124: Is the CNS request controller the only one updating the node net config?

cns/requestcontroller/kubernetes/k8srequestcontroller.go
pjohnst5: Yeah, if in the future we ever need to do this same thing with something like private link (not necessarily using CRDs but some other form of message relaying between dnc and cns)

cns/requestcontroller/kubernetes/eventfilter.go
pjohnst5: Yes! Nice catch

cns/requestcontroller/example/main.go
pjohnst5: Adding an exitChan parameter to StartRequestController(exitChan), this exitChan will be notified when the mgr receives SIGINT or SIGTERM. let me know if that's what you mean (usage in example/main.go)

cns/requestcontroller/kubernetes/k8srequestcontroller.go
pjohnst5: This server is actually CNS's state, so reconcile will change the cns's state directly based on the incoming status

cns/requestcontroller/kubernetes/k8srequestcontroller.go
pjohnst5: No let me add those checks

cns/requestcontroller/kubernetes/k8srequestcontroller.go
pjohnst5: Yes

cns/requestcontroller/kubernetes/k8srequestcontroller.go
pjohnst5: Done

cns/requestcontroller/kubernetes/k8srequestcontroller.go
pjohnst5: Yeah we will do deprecate requested ip count at the same time, I'll combine them

cns/requestcontroller/kubernetes/nodenetworkconfigreconciler.go
pjohnst5: Removed it altogether by filtering for this in eventfilter.go

cns/requestcontroller/kubernetes/nodenetworkconfigreconciler.go
pjohnst5: Cached. Way under the hood in controller-runtime code regarding creating a new manager, I found this explanation about the "NewClient" option, I do not provide a "NewClientFunc" when creating the manager so default behavior is using the cache:
// NewClient will create the client to be used by the manager.
	// If not set this will create the default DelegatingClient that will
	// use the cache for reads and the client for writes.
	NewClient NewClientFunc

Makefile
```diff
@@ -378,6 +378,7 @@ test-all:
         ./cnm/network/ \
         ./cni/ipam/ \
         ./cns/ipamclient/ \
+		./cns/requestcontroller/kubecontroller/ \
```
thatmattlong, line 381: (very) nitpicky but curious as to the spacing here. I know `make` is very particular about spacing, so perhaps there is a reason that I'm not seeing

cns/requestcontroller/kubecontroller/cnsinteractor.go
```diff
@@ -0,0 +1,17 @@
+package kubecontroller
+
+import (
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/restserver"
+)
+
+// CNSInteractor implements the CNSClient interface.
```
thatmattlong, line 8: IMO the CNS Client interface should live in the `cns` package or a `cns/client` package as `Client`, and then an implementation, say `HTTPClient`, should live next to it. I think client implementations should live close to the API in which they are a client for. Then, instead of having a `kubecontroller.CNSInteractor`, your controller can just have a `cns.Client` (or `client.Client`, depending on where you put it). Then, at instantiation of your controller, you can choose the implementation; for the `main` function, it would be the `HTTPClient`, and for tests, it could be a `mockClient` so you don't start mixing unit tests and integration tests.

cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -0,0 +1,53 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// CrdReconciler watches for CRD status changes
+type CrdReconciler struct {
+	APIClient APIClient
+	NodeName  string
+	CNSClient requestcontroller.CNSClient
```
thatmattlong, line 18: This would be
```go
CNSClient cns.Client
```
here instead, if you followed the comment I gave above.

cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Reconcile is called on CRD status changes
+func (r *CrdReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	var nodeNetConfig nnc.NodeNetworkConfig
+
+	//Get the CRD object
+	if err := r.APIClient.Get(context.TODO(), request.NamespacedName, &nodeNetConfig); err != nil {
+		logger.Printf("[cns-rc] CRD not found, ignoring %v", err)
```
thatmattlong, line 27: There are other reasons as to why Get could return an error. Instead of this logging message, I would change the logic to something like
```go
if err := r.APIClient.Get(context.TODO(), request.NamespacedName, &nodeNetConfig); err != nil {
    if apierrors.IsNotFound(err) {
        logger.Printf("[cns-rc] CRD not found in cache, ignoring %v", err)
        return reconcile.Result{}, nil
    } else {
        logger.Printf("[cns-rc] error retrieving CRD %v from cache", err)
        return reconcile.Result{}, err
    }
}
```

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -0,0 +1,168 @@
...
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+const nodeNameEnvVar = "NODENAME"
+const k8sNamespace = "kube-system"
+const prometheusAddress = "0"
```
thatmattlong, line 21: maybe a comment here to indicate 0 means disabled (but I do like the name of the constant). I might look at this and think "`0` is pretty close to `0.0.0.0/0` so it must be binding to all addresses", which is the opposite of what it actually does

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	return k8sconfig, nil
+}
+
+//NewCrdRequestController given a reference to CNS's HTTPRestService state, returns a crdRequestController struct
+func NewCrdRequestController(restService *restserver.HTTPRestService, kubeconfig *rest.Config) (*crdRequestController, error) {
+
+	//Check that logger package has been intialized
+	if logger.Log == nil {
+		return nil, errors.New("Must initialize logger before calling")
```
thatmattlong, line 51: Should we fallback to logging to stdout if the logger isn't initialized? Maybe not, but wanted to ask in case

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -0,0 +1,168 @@
...
+// UpdateCRDSpec updates the CRD spec
+func (crdRC *crdRequestController) UpdateCRDSpec(cntxt context.Context, crdSpec *nnc.NodeNetworkConfigSpec) error {
+	nodeNetworkConfig, err := crdRC.getNodeNetConfig(cntxt, crdRC.nodeName, k8sNamespace)
+	if err != nil {
+		logger.Errorf("[cns-rc] Error getting CRD when updating spec %v", err)
+		return err
+	}
+
+	//Update the CRD spec
+	nodeNetworkConfig.Spec = *crdSpec
```
thatmattlong, line 134: I think you should use `crdspec.DeepCopyInto(&nodeNetworkConfig.Spec)` here for a safer assignment https://github.com/Azure/azure-container-networking/blob/master/nodenetworkconfig/api/v1alpha/zz_generated.deepcopy.go#L122

cns/requestcontroller/requestcontrollerintreface.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// RequestController interface for cns to interact with the request controller
+type RequestController interface {
+	StartRequestController(exitChan chan bool) error
+	UpdateCRDSpec(cntxt context.Context, crdSpec *nnc.NodeNetworkConfigSpec) error
+}
+
+// CNSClient interface for request controller to interact with cns
+type CNSClient interface {
```
thatmattlong, line 17: Already mentioned moving this in comments above - but including a comment here as a reminder

Makefile
pjohnst5: You're right, I was using tabs before (because vscode auto-tabbed it) so to fix that, I'm using spaces like the other lines above it are

cns/requestcontroller/kubecontroller/cnsinteractor.go
pjohnst5: Done, made an APIClient interface in the cns/client package, and a Client implementation in the cns/client/httpapi subpackage

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
pjohnst5: Hmm that's a good question, I would say, in order to capture the logs in AI telemetry and stuff then no

cns/requestcontroller/kubecontroller/apiclientinterface.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ObjectKey identifies a Kubernetes Object.
+type ObjectKey = types.NamespacedName
```
thatmattlong, line 12: I don't think this is used anywhere, it can be deleted

pjohnst5: You're right, since checks take so long though, merging this PR and will remove in next iteration

PR #586, opened 2020-06-23
Update Pipeline Image

jaer-tsun:
Updating pipeline image with necessary libs and bumping up base image version to 20.04LTS

### DIFF
--- a/.pipelines/Dockerfile
+++ b/.pipelines/Dockerfile
@@ -1,18 +1,16 @@
-FROM ubuntu:16.04
-RUN apt-get update && apt-get install -y software-properties-common sudo wget apt-transport-https curl lsb-release gnupg jq
-RUN wget -q https://packages.microsoft.com/config/ubuntu/16.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
+FROM ubuntu:20.04
+RUN apt-get update && apt-get install -y software-properties-common sudo wget apt-utils apt-transport-https curl lsb-release gnupg jq
+RUN wget -q https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
 RUN sudo dpkg -i packages-microsoft-prod.deb
-RUN add-apt-repository ppa:longsleep/golang-backports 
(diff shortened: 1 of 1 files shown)


.pipelines/Dockerfile
```diff
@@ -1,11 +1,12 @@
-FROM ubuntu:16.04
+FROM ubuntu:20.04
 RUN apt-get update && apt-get install -y software-properties-common sudo wget apt-transport-https curl lsb-release gnupg jq
 RUN wget -q https://packages.microsoft.com/config/ubuntu/16.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
```
matmerr, line 3: change to `/ubuntu/20.04/packages-microsoft-prod.deb`

.pipelines/Dockerfile
```diff
@@ -1,11 +1,12 @@
...
 RUN sudo dpkg -i packages-microsoft-prod.deb
 RUN add-apt-repository ppa:longsleep/golang-backports 
 RUN curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
 RUN curl -fsSL https://get.docker.com -o get-docker.sh && sh get-docker.sh
 RUN apt-get install -y git golang-go=2:1.13~1longsleep1+xenial iptables ipset iproute2 ebtables python-pip gcc zip dotnet-sdk-2.2 azure-cli
```
matmerr, line 8: let's update this to golang 1.14 while we're at it

.pipelines/Dockerfile
```diff
@@ -1,11 +1,12 @@
...
+RUN apt install -y --no-install-recommends clang cmake zlib1g-dev libboost-dev libboost-thread-dev gdb build-essential libssl-dev
 RUN sudo pip install coverage
 RUN if [ -f Gopkg.toml ]; then curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh ; fi
```
matmerr, line 11: we don't need dep anymore

jaer-tsun: removed

.pipelines/Dockerfile
jaer-tsun: bumped

PR #587, opened 2020-06-23
CRD: Move scaling items into scaler, change scope to namespaced

thatmattlong:
Changes CRD scope to Namespaced and groups together scaling-related items in status. Also, this adds the CRD manifest to the repo for quicker testing iteration.

Note that changing the scope of a CRD is not possible if the CRD already exists. To run this on a cluster that already has this CRD installed, you have to delete the CRD definition first before applying this one.


nodenetworkconfig/manifests/acn.azure.com_nodenetworkconfigs.yaml
```diff
@@ -0,0 +1,104 @@
...
+    controller-gen.kubebuilder.io/version: v0.3.0
+  creationTimestamp: null
+  name: nodenetworkconfigs.acn.azure.com
+spec:
+  group: acn.azure.com
+  names:
+    kind: NodeNetworkConfig
+    listKind: NodeNetworkConfigList
+    plural: nodenetworkconfigs
+    singular: nodenetworkconfig
```
pjohnst5, line 16: Could we add "nnc" as a short name like so?
```
shortNames:
      - nnc
```

thatmattlong: done

PR #599, opened 2020-06-30
Add logic to deal with 0.0.0.0/0 which ipset not support.

csfmomo:
fix: repair 0.0.0.0/0 cidr which ipset not support

**Reason for Change**:
Ipset doesn't support 0.0.0.0/0 cidr. A general approach is convert it to 1.0.0.0/1 and 128.0.0.0/1

- fix: Bug Fixes 🐞


matmerr:
Can we add a test to validate this scenario?

csfmomo:
> Can we add a test to validate this scenario?

Let me figure out a unit test.

matmerr:
Interesting solution, approved

PR #600, opened 2020-07-01
Changes for caching pod ip

neaggarwMS:
Cache the PodIp which is added in the ipset so that it gets cleaned up deterministically as Pod DeleteEvent callback is asynchronous.
 
In case Pod DeleteEvent is received after ip has been reassigning to another POD deployed in the same namespace, then there wont be any ipset change.

### DIFF
--- a/npm/ipsm/ipsm.go
+++ b/npm/ipsm/ipsm.go
@@ -28,7 +28,7 @@ type IpsetManager struct {
 // Ipset represents one ipset entry.
 type Ipset struct {
 	name       string
-	elements   map[string]string // key = ip, value: context associated to the ip like podOwnername
+	elements   map[string]string // key = ip, value: context associated to the ip like podUid
 	referCount int
 }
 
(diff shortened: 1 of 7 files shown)


npm/ipsm/ipsm.go
```diff
@@ -28,14 +28,15 @@ type IpsetManager struct {
 // Ipset represents one ipset entry.
 type Ipset struct {
 	name       string
-	elements   []string
+	elements   map[string]string // key = ip, value: context associated to the ip like podOwnername
```
csfmomo, line 31: Does this podOwnerName mean PodUid mentioned in design doc?

neaggarwMS: Yes, I will fix the comment.

npm/ipsm/ipsm.go
```diff
@@ -291,6 +305,11 @@ func (ipsMgr *IpsetManager) DeleteFromSet(setName, ip string) error {
 		return err
 	}
 
+	// Now cleanup the cache
+	if _, exists := ipsMgr.setMap[setName].elements[ip]; exists {
```
csfmomo, line 309: I think this condition check has already been done in line 285, we can go ahead and cleanup the cache.

PR #603, opened 2020-07-07
Request Controller reconcile CNS state on reboot

pjohnst5:
**Reason for Change**:
This is a feature for CNS request controller to reconcile the CNS state on a reboot.

**What it does**
Added new method InitCNS() to crdRequestController which gets pods on the node, gets nodenetconfig, and passes them to cns

Other minor touches
- Got rid of unused ObjectType
- Updated main.go example for request controller usage


cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -30,30 +34,95 @@ func (r *CrdReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+			return reconcile.Result{}, err
+		}
+	} else {
+		if err := r.markIPs(ipConfigs); err != nil {
+			logger.Errorf("[cns-rc] Error marking ips as allocated when readying CNS: %v", err)
+			//requeue
+			return reconcile.Result{}, err
+		}
+
+		if err := r.CNSClient.InitCNSState(ipConfigs); err != nil {
```
ramiro-gamarra, line 65: Is the intention to lazily initialize CNS? Does it ever go back to being uninitialized?

cns/cnsclient/httpapi/client.go
```diff
@@ -11,7 +11,45 @@ type Client struct {
...
-func (client *Client) UpdateCNSState(createNetworkContainerRequest *cns.CreateNetworkContainerRequest) error {
-	//Mat will pick up from here
-	return nil
+func (client *Client) UpdateCNSState(ipConfigs []*cns.ContainerIPConfigState) error {
+	var (
+		ipConfigsToAdd []*cns.ContainerIPConfigState
+	)
+
+	//Lock to read ipconfigs
+	client.RestService.Lock()
```
ramiro-gamarra, line 20: It feels strange for this calling code to mess with the internals of its dependency (locking and mutating). Ideally RestService should not be exporting its mutex, but that's probably a larger change. Would these methods be better just as part of RestService? In that case, RestService still meets your defined interface, and that would be all you inject where necessary.

cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -30,30 +34,95 @@ func (r *CrdReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	}
+
+	return nil
+}
+
+// GetPods gets the pods running on this node
+func (r *CrdReconciler) getPods() (*corev1.PodList, error) {
+	pods := &corev1.PodList{}
+
+	if err := r.KubeClient.List(context.TODO(), pods, client.MatchingFields{"spec.nodeName": r.NodeName}); err != nil {
```
ramiro-gamarra, line 123: Probably better for this context to be passed into these functions. You should define a parent context at the start of the reconcile method and just pass it as needed.

cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -30,30 +34,95 @@ func (r *CrdReconciler) Reconcile(request reconcile.Request) (reconcile.Result,
...
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&nnc.NodeNetworkConfig{}).
-		WithEventFilter(NodeNetworkConfigFilter{nodeName: r.NodeName}).
+		WithEventFilter(NodeNetworkConfigFilter{nodeName: r.NodeName, namespace: r.Namespace}).
 		Complete(r)
 }
+
+// markIPs marks ips as allocated if it finds a pod using that ip, and available otherwise
+func (r *CrdReconciler) markIPs(ipConfigs []*cns.ContainerIPConfigState) error {
```
ramiro-gamarra, line 92: Thoughts on a different method name, such as `updateIPAvailability`?

cns/cnsclient/httpapi/client.go
matmerr: +1, just pass array of ipconfigs to client

cns/requestcontroller/kubecontroller/crdreconciler.go
pjohnst5: Yes the intention is to lazily initialize CNS, the only time it would go back to being uninitialized is on reboot.

cns/requestcontroller/kubecontroller/crdreconciler.go
pjohnst5: Yeah sure I'll change that

cns/requestcontroller/kubecontroller/crdrequestcontroller_test.go
```diff
@@ -20,35 +21,49 @@ import (
...
 const existingNamespace = k8sNamespace
 const nonexistingNNCName = "nodenetconfig_nonexisting"
+const nonexistingPodName = "pod_nonexisting"
 const nonexistingNamespace = "namespace_nonexisting"
 
 var (
-	mockStore      map[MockKey]*nnc.NodeNetworkConfig
+	mockAPI        *MockAPI
 	mockCNSUpdated bool
+	mockCNSReady   bool
```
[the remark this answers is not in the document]
pjohnst5: Makes sense now, I will remove package level vars

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -5,15 +5,14 @@ import (
 	nnc "github.com/Azure/azure-container-networking/nodenetworkconfig/api/v1alpha"
 )
 
-// CRDStatusToNCRequest translates a crd status to network container request
-func CRDStatusToNCRequest(crdStatus *nnc.NodeNetworkConfigStatus) (*cns.CreateNetworkContainerRequest, error) {
-	//TODO: Translate CRD status into network container request
-	//Mat will pick up from here
+// CRDStatusToCNS translates a crd status to cns recognizable data
+func CRDStatusToCNS(crdStatus *nnc.NodeNetworkConfigStatus) ([]*cns.ContainerIPConfigState, error) {
```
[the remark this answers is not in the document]
pjohnst5: I don't know actually but that is something to consider

cns/cnsclient/httpapi/client.go
pjohnst5: @matmerr and I talked offline and matmerr is going to add a method to the RestService "GetIPConfigsFromState()" which returns inuseIPs. Request controller will then do the filtering by
`inuseIPs := client.RestService.GetIPConfigsFromState()`
client will then use this list of inuseIPs to sift out ipconfigs that cns already has, and thus, all locking/internal RestService things happen in RestService

pjohnst5: I think for now I am just going to merge in this version of UpdateCNSState and fix it later in a later PR after @matmerr has his changes in

cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -16,44 +19,132 @@ import (
...
 	CNSClient  cnsclient.APIClient
 }
 
 // Reconcile is called on CRD status changes
 func (r *CrdReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	var nodeNetConfig nnc.NodeNetworkConfig
+	var (
+		nodeNetConfig nnc.NodeNetworkConfig
+		ipConfigs     []*cns.ContainerIPConfigState
+		cntxt         context.Context
```
neaggarwMS, line 31: What does TODO represent?

cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -16,44 +19,132 @@ import (
...
-	if err := r.KubeClient.Get(context.TODO(), request.NamespacedName, &nodeNetConfig); err != nil {
+	cntxt = context.TODO()
+	if err = r.KubeClient.Get(cntxt, request.NamespacedName, &nodeNetConfig); err != nil {
 		if apierrors.IsNotFound(err) {
 			logger.Printf("[cns-rc] CRD not found, ignoring %v", err)
 			return reconcile.Result{}, client.IgnoreNotFound(err)
 		} else {
 			logger.Errorf("[cns-rc] Error retrieving CRD from cache : %v", err)
+			//requeue
 			return reconcile.Result{}, err
```
neaggarwMS, line 44: If this is the same logger as used by NPM, then lets switch to logF. Ideally we should deprecate printf

cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -16,44 +19,132 @@ import (
...
 		}
 	}
```
neaggarwMS, line 46: Not sure if i understood the change, Request controller will call the NC Request only to CNSClient, no?

cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -16,44 +19,132 @@ import (
...
+		pod := rawObj.(*corev1.Pod)
+		return []string{pod.Spec.NodeName}
+	}); err != nil {
+		return err
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&nnc.NodeNetworkConfig{}).
-		WithEventFilter(NodeNetworkConfigFilter{nodeName: r.NodeName}).
+		WithEventFilter(NodeNetworkConfigFilter{nodeName: r.NodeName, namespace: r.Namespace}).
```
neaggarwMS, line 93: Renme it to ReconcileCNSIpam or ReconcileCNSState, either you should be calling before you initialize Reconcile loop, no? as this is a one time call during initialization. You will list CRD and PODs on a node and reconcileCNS state.

cns/requestcontroller/kubecontroller/crdreconciler.go
```diff
@@ -16,44 +19,132 @@ import (
...
+	// Get current pods running on the node
+	pods, err = r.getPods(cntxt)
+	if err != nil {
+		return err
+	}
+
+	// Index ipConfigs by address to avoid inner loop
+	for _, ipConfig = range ipConfigs {
+		ipConfigByAddress[ipConfig.IPConfig.IPAddress] = ipConfig
+	}
```
neaggarwMS, line 120: skip this block if len(podIP) == 0

cns/cnsclient/httpapi/client.go
```diff
@@ -11,7 +11,45 @@ type Client struct {
...
+
+	//Only add ipconfigs that don't exist in cns state already
+	for _, ipConfig := range ipConfigs {
+		if _, ok := client.RestService.PodIPConfigState[ipConfig.ID]; !ok {
+			ipConfig.State = cns.Available
+			ipConfigsToAdd = append(ipConfigsToAdd, ipConfig)
+		}
```
neaggarwMS, line 27: This block should be called after you set the state

cns/cnsclient/httpapi/client.go
```diff
@@ -11,7 +11,45 @@ type Client struct {
...
+	}
+
+	client.RestService.Unlock()
+
```
neaggarwMS, line 31: How are you initializing/reconciling the NC state? There could be some stale NC/POD configurations as well, no?

cns/cnsclient/httpapi/client.go
neaggarwMS: I agree with Ramiro, I dont think it is right to muck with internal state in CNS in the client code.

cns/cnsclient/httpapi/client.go
```diff
@@ -11,7 +11,45 @@ type Client struct {
...
+	return client.RestService.AddIPConfigsToState(ipConfigsToAdd)
+}
+
+// InitCNSState initializes cns state
```
neaggarwMS, line 35: if it is a get api, then I will rename it to GetIPAMStatus or IsIPAMReady(), this looks like you are going to set CNS IPAM ready

cns/cnsclient/apiclient.go
```diff
@@ -4,5 +4,7 @@ import "github.com/Azure/azure-container-networking/cns"
```
neaggarwMS, line 4: Since this is called in CRD Reconcile, this should always call CreateNetworkContainerRequest)

cns/cnsclient/apiclient.go
```diff
@@ -4,5 +4,7 @@ import "github.com/Azure/azure-container-networking/cns"
 
 // APIClient interface to update cns state
```
neaggarwMS, line 5: This API should be called before we start the APIServer Reconcile loop. 

Also Requestcontroller should pass the ContainerNetworkRequest built from getting a crd (if any, else it will be mull) and map of OrchastratorContext (PodName+Podnamespace)->PodIPs (only the pods which have allocated ips). 

Requestcontroller should not be aware of internal IPAM config state.

cns/cnsclient/apiclient.go
```diff
@@ -4,5 +4,7 @@ import "github.com/Azure/azure-container-networking/cns"
...
 type APIClient interface {
```
neaggarwMS, line 6: You dont need this. you will be able to start reconcile loop only after successful CNS state initialization.

cns/requestcontroller/kubecontroller/crdreconciler.go
pjohnst5: Couldn't find logF in logger package

cns/requestcontroller/requestcontrollerintreface.go
```diff
@@ -3,17 +3,12 @@ package requestcontroller
 import (
 	"context"
 
-	"github.com/Azure/azure-container-networking/cns"
 	nnc "github.com/Azure/azure-container-networking/nodenetworkconfig/api/v1alpha"
 )
 
 // RequestController interface for cns to interact with the request controller
```
neaggarwMS, line 9: Why would CNS will call into RequestController to InitCNS?

pjohnst5: I was thinking CNS would create a new request controller obj, call InitCNS on that obj, and then call StartRequestController on that obj.
How are you envisioning it?

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -115,14 +143,94 @@ func NewCrdRequestController(restService *restserver.HTTPRestService, kubeconfig
...
+		pod           corev1.Pod
+		podInfo       *cns.KubernetesPodInfo
+		nodeNetConfig *nnc.NodeNetworkConfig
+		podInfoByIP   map[string]*cns.KubernetesPodInfo
+		cntxt         context.Context
+		ncRequest     *cns.CreateNetworkContainerRequest
+		err           error
+	)
+
+	cntxt = context.Background()
```
neaggarwMS, line 182: What does this Background() do?

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -115,14 +143,94 @@ func NewCrdRequestController(restService *restserver.HTTPRestService, kubeconfig
...
+		}
+
+		return err
 	}
 
 	return nil
 }
 
+// InitCNS initializes cns by passing pods and a createnetworkcontainerrequest
+func (crdRC *crdRequestController) InitCNS() error {
```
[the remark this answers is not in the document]
pjohnst5: Yes

PR #605, opened 2020-07-08
Updated default cni version in script

tamilmani1989:
**Reason for Change**:

Update default cni plugin version to latest

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


matmerr:
CNI Version v0.4.0 goes up to CNI spec 2.0
CNI Version v0.8.0 goes up to CNI spec 4.0

Do we need to cost out the work to potentially support CNI spec 4.0? There's an added CHECK command, as well as few other changes between spec 2.0 and spec 4.0

PR #608, opened 2020-07-14
Translate CRD <-> CNS

pjohnst5:
**Reason for Change**:
Translation between CRD and CNS


cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -1,19 +1,80 @@
...
+			ipSubnet.PrefixLength = uint8(prefixLength)
+			secondaryIPConfig = cns.SecondaryIPConfig{
+				IPConfig: ipSubnet,
+			}
+			createNCRequest.SecondaryIPConfigs[ipAssignment.Name] = secondaryIPConfig
+		}
+	}
+
+	//Only returning the first network container for now, later we will return a list
+	return &createNCRequest, nil
```
ramiro-gamarra, line 41: Can this be addressed in this PR? Or is there something else that blocks this returning a list?

cns/requestcontroller/example/main.go
```diff
@@ -3,11 +3,11 @@ package main
 import (
```
neaggarwMS, line 3: Why we need this example main.go?

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -1,19 +1,80 @@
...
 }
 
 // CNSToCRDSpec translates CNS's list of Ips to be released and requested ip count into a CRD Spec
-func CNSToCRDSpec() (*nnc.NodeNetworkConfigSpec, error) {
-	//TODO: Translate list of ips to be released and requested ip count to CRD spec
-	return nil, nil
+func CNSToCRDSpec(secondaryIPConfigs []cns.SecondaryIPConfig, ipCount int) (*nnc.NodeNetworkConfigSpec, error) {
```
neaggarwMS, line 45: Can we rename secondaryIPConfigs  to secondaryIpsNotInUse or tobeDeletedSecondaryIpConfigs

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -1,19 +1,80 @@
...
+		secondaryIPConfig cns.SecondaryIPConfig
+	)
+
+	spec.RequestedIPCount = int64(ipCount)
+
+	for _, secondaryIPConfig = range secondaryIPConfigs {
+		spec.IPsNotInUse = append(spec.IPsNotInUse, secondaryIPConfig.IPConfig.IPAddress)
+	}
+
+	return &spec, nil
```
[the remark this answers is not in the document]
neaggarwMS: Agree, not return pointer unless necessary

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -1,19 +1,80 @@
...
+	"strconv"
+	"strings"
+
 	"github.com/Azure/azure-container-networking/cns"
 	nnc "github.com/Azure/azure-container-networking/nodenetworkconfig/api/v1alpha"
 )
 
-// CRDStatusToNCRequest translates a crd status to network container request
+// CRDStatusToNCRequest translates a crd status to createnetworkcontainer request
 func CRDStatusToNCRequest(crdStatus *nnc.NodeNetworkConfigStatus) (*cns.CreateNetworkContainerRequest, error) {
```
neaggarwMS, line 13: Can it return the object instead of a pointer?

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -1,19 +1,80 @@
...
+		ipSubnet          cns.IPSubnet
+		ipAssignment      nnc.IPAssignment
+		prefixLength      int
+	)
+
+	createNCRequest.SecondaryIPConfigs = make(map[string]cns.SecondaryIPConfig)
+
+	for _, nc = range crdStatus.NetworkContainers {
+		createNCRequest.NetworkContainerid = nc.ID
+		createNCRequest.PrimaryInterfaceIdentifier = nc.PrimaryIP
```
neaggarwMS, line 27: This is not correct. NC Primary ip will be saved in createNCRequest.IPConfiguration

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -1,19 +1,80 @@
...
 )
 
-// CRDStatusToNCRequest translates a crd status to network container request
+// CRDStatusToNCRequest translates a crd status to createnetworkcontainer request
 func CRDStatusToNCRequest(crdStatus *nnc.NodeNetworkConfigStatus) (*cns.CreateNetworkContainerRequest, error) {
-	//TODO: Translate CRD status into network container request
-	//Mat will pick up from here
-	return nil, nil
+	var (
+		createNCRequest   cns.CreateNetworkContainerRequest
```
neaggarwMS, line 15: nit: can we rename it to ncRequest? Create is redundant here

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -1,19 +1,80 @@
...
+		secondaryIPConfig cns.SecondaryIPConfig
+		ipSubnet          cns.IPSubnet
+		ipAssignment      nnc.IPAssignment
+		prefixLength      int
+	)
+
+	createNCRequest.SecondaryIPConfigs = make(map[string]cns.SecondaryIPConfig)
+
+	for _, nc = range crdStatus.NetworkContainers {
+		createNCRequest.NetworkContainerid = nc.ID
```
neaggarwMS, line 26: We also need to set NetworkContainerType. Set it to cns.Docker

cns/requestcontroller/example/main.go
pjohnst5: We don't, I'll remove it

cns/requestcontroller/kubecontroller/crdtranslator.go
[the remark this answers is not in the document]
pjohnst5: It's necessary, when updating the CRD spec, I use the "DeepCopyInto()" method of the CRD library to completely overwrite the spec. @thatmattlong for confirmation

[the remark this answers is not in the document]
pjohnst5: Ok I'll get rid of the pointers.

cns/requestcontroller/kubecontroller/crdtranslator.go
pjohnst5: As Neha already implemented her CreateOrUpdateNCRequest it will be a bigger change so I think we should tackle it in a separate pr

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -199,7 +200,9 @@ func (crdRC *crdRequestController) initCNS() error {
 
 	// Convert to CreateNetworkContainerRequest if crd not nill and is populated
 	if nodeNetConfig != nil && len(nodeNetConfig.Status.NetworkContainers) != 0 {
-		if ncRequest, err = CRDStatusToNCRequest(&nodeNetConfig.Status); err != nil {
+		// Assign the reference to the ncRequest, otherwise we'll pass a nil pointer to InitCNSState
```
neaggarwMS, line 203: Its okay to pass the nil pointer. Actually it should be nil pointer in case CRD doesnt exist. Otherwise we wont be able to differiante between a bad NCRequest vs no ncrequest

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -227,20 +230,20 @@ func (crdRC *crdRequestController) initCNS() error {
 	}
 
 	// Call cnsclient init cns passing those two things
-	return crdRC.CNSClient.InitCNSState(ncRequest, podInfoByIP)
+	return crdRC.CNSClient.InitCNSState(ncRequestReference, podInfoByIP)
```
neaggarwMS, line 233: pass &ncRequest only

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -1,19 +1,74 @@
...
 	nnc "github.com/Azure/azure-container-networking/nodenetworkconfig/api/v1alpha"
 )
 
-// CRDStatusToNCRequest translates a crd status to network container request
-func CRDStatusToNCRequest(crdStatus *nnc.NodeNetworkConfigStatus) (*cns.CreateNetworkContainerRequest, error) {
-	//TODO: Translate CRD status into network container request
-	//Mat will pick up from here
-	return nil, nil
+// CRDStatusToNCRequest translates a crd status to createnetworkcontainer request
+func CRDStatusToNCRequest(crdStatus nnc.NodeNetworkConfigStatus) (cns.CreateNetworkContainerRequest, error) {
```
neaggarwMS, line 11: Well, this returns a pointer. As ncRequest can be nil if no CRD exists. Sorry my earlier feedback mislead you. Remove pointers where it is not required.

neaggarwMS: Can we write UT for these functions? Wondering how we will pass nnc.NodeNetworkConfigStatus

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
pjohnst5: Here's the thing. If the crd isn't found, we want to pass a nil pointer right? That's why I added this ncRequestReference variable. ncRequestReference is nil if the crd wasn't found.

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
pjohnst5: Right, that's why I had to add the ncRequestReference variable, see the comment below

pjohnst5: Nvm, saw your other comment, making a pointer

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
pjohnst5: Nvm, saw your other comment, making a pointer

cns/requestcontroller/kubecontroller/crdtranslator.go
pjohnst5: I'll write UTs

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -1,19 +1,79 @@
...
+		ipAssignment      nnc.IPAssignment
+		err               error
+		ip                net.IP
+		ipNet             *net.IPNet
+		bits              int
+	)
+
+	ncRequest.SecondaryIPConfigs = make(map[string]cns.SecondaryIPConfig)
+
+	for _, nc = range crdStatus.NetworkContainers {
```
neaggarwMS, line 27: We should check if crdStatus has any NetworkContainers, if there are none then return nil, else pick the first one (for now we have supported only single nic) and populate the request. We can also throw error if range has > 1 NC

pjohnst5: Added a check that length of NetworkContainers list is 1

PR #615, opened 2020-07-15
CNS Client API interface and skeleton implementation

pjohnst5:
**Reason for Change**:
API template and skeleton implementation to unblock Neha

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -35,7 +35,6 @@ const (
 	Basic                  = "Basic"
 	JobObject              = "JobObject"
 	COW                    = "COW" // Container on Windows
-	AKS                    = "AKS"
 )
 
 // Orchestrator Types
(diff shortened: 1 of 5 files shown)


cns/NetworkContainerContract.go
```diff
@@ -35,6 +35,7 @@ const (
 	Basic                  = "Basic"
 	JobObject              = "JobObject"
 	COW                    = "COW" // Container on Windows
+	AKS                    = "AKS"
```
neaggarwMS, line 38: Discussed with Tamilmani, lets remove it, we will be using ContainerType = "Docker"

cns/cnsclient/apiclient.go
```diff
@@ -4,5 +4,6 @@ import "github.com/Azure/azure-container-networking/cns"
 
 // APIClient interface to update cns state
 type APIClient interface {
-	UpdateCNSState(createNetworkContainerRequest *cns.CreateNetworkContainerRequest) error
+	CreateOrUpdateNC(*cns.CreateNetworkContainerRequest) error
+	InitCNSState(*cns.CreateNetworkContainerRequest, map[string]*cns.KubernetesPodInfo) error
```
neaggarwMS, line 8: nit: can you reverse the order of these APIs. Like InitCnsState before CreateOrUpdateNetworkContainer

PR #617, opened 2020-07-16
Implement CreateOrUpdateNetworkContainerInternal API. 

neaggarwMS:
Added implementation for CreateOrUpdateNetworkContainerInternal. This will be called by the CNS requestController and it will handle SecondaryIpConfigs update which will add state for new Secondary ips and delete the state for any delta.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,9 @@ CNSFILES = \
 	$(wildcard cns/routes/*.go) \
 	$(wildcard cns/service/*.go) \
 	$(wildcard cns/networkcontainers/*.go) \
+	$(wildcard cns/requestcontroller/*.go) \
+	$(wildcard cns/requestcontroller/kubecontroller/*.go) \
+	$(wildcard cns/requestcontroller/example/*.go) \
 	$(COREFILES) \
 	$(CNMFILES)
 
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,6 @@ CNSFILES = \
 	$(wildcard cns/networkcontainers/*.go) \
 	$(wildcard cns/requestcontroller/*.go) \
 	$(wildcard cns/requestcontroller/kubecontroller/*.go) \
-	$(wildcard cns/requestcontroller/example/*.go) \
 	$(COREFILES) \
 	$(CNMFILES)
 
--- a/cns/cnsclient/cnsclient_test.go
+++ b/cns/cnsclient/cnsclient_test.go
@@ -47,10 +47,7 @@ func addTestStateToRestServer(t *testing.T, secondaryIps []string) {
 				PrefixLength: 32,
 			},
 		}
-		ipId, err := uuid.NewUUID()
-		if err != nil {
-			t.Fatalf("Failed to generate UUID for secondaryipconfig, err:%s", err)
-		}
+		ipId := uuid.New()
 		secondaryIPConfigs[ipId.String()] = secIpConfig
 	}
 
--- a/cns/cnsclient/cnsclient_test.go
+++ b/cns/cnsclient/cnsclient_test.go
@@ -91,7 +91,7 @@ func getIPConfigFromGetNetworkContainerResponse(resp *cns.GetIPConfigResponse) (
 func TestMain(m *testing.M) {
 	var (
 		info = &cns.SetOrchestratorTypeRequest{
-			OrchestratorType: cns.Kubernetes}
+			OrchestratorType: cns.KubernetesCRD}
 		body bytes.Buffer
 		res  *http.Response
 	)
@@ -152,13 +152,13 @@ func TestCNSClientRequestAndRelease(t *testing.T) {
 	podNamespace := "testpodnamespace"
 	desiredIpAddress := "10.0.0.5"
 	ip := net.ParseIP(desiredIpAddress)
-	_, ipnet, _ := net.ParseCIDR("10.0.0.5/24")
+	_, ipnet, _ := net.ParseCIDR("10.0.0.5/32")
 	desired := net.IPNet{
 		IP:   ip,
 		Mask: ipnet.Mask,
 	}
 
-	secondaryIps := make([]string, 1)
+	secondaryIps := make([]string, 0)
 	secondaryIps = append(secondaryIps, desiredIpAddress)
 	cnsClient, _ := InitCnsClient("")
 
@@ -170,10 +170,10 @@ func TestCNSClientRequestAndRelease(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// no IP reservation found with that context, expect fail
+	// no IP reservation found with that context, expect no failure.
 	err = cnsClient.ReleaseIPAddress(orchestratorContext)
-	if err == nil {
-		t.Fatalf("Expected failure to release when no IP reservation found with context: %+v", err)
+	if err != nil {
+		t.Fatalf("Release ip idempotent call failed: %+v", err)
 	}
 
 	// request IP address
--- a/cns/cnsclient/cnsclient_test.go
+++ b/cns/cnsclient/cnsclient_test.go
@@ -144,7 +144,8 @@ func TestMain(m *testing.M) {
 	}
 	fmt.Println(res)
 
-	m.Run()
+	exitCode := m.Run()
+	os.Exit(exitCode)
 }
 
 func TestCNSClientRequestAndRelease(t *testing.T) {
--- a/cns/cnsclient/cnsclient_test.go
+++ b/cns/cnsclient/cnsclient_test.go
@@ -1,35 +1,71 @@
 package cnsclient
 
 import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
 	"net"
+	"net/http"
+	"os"
+	"reflect"
 	"strconv"
+	"testing"
 
 	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/common"
+	"github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/Azure/azure-container-networking/cns/restserver"
+	"github.com/Azure/azure-container-networking/log"
+	"github.com/google/uuid"
 )
 
 var (
-	testNCID = "06867cf3-332d-409d-8819-ed70d2c116b0"
-
-	testIP1      = "10.0.0.1"
-	testPod1GUID = "898fb8f1-f93e-4c96-9c31-6b89098949a3"
-	testPod1Info = cns.KubernetesPodInfo{
-		PodName:      "testpod1",
-		PodNamespace: "testpod1namespace",
-	}
+	svc *restserver.HTTPRestService
 )
 
-// func addTestStateToRestServer(svc *restserver.HTTPRestService) {
-// 	// set state as already allocated
-// 	state1, _ := restserver.NewPodStateWithOrchestratorContext(testIP1, 24, testPod1GUID, testNCID, cns.Available, testPod1Info)
-// 	ipconfigs := map[string]cns.SecondaryIPConfig{
-// 		state1.ID: state1,
-// 	}
-// 	nc := cns.CreateNetworkContainerRequest{
-// 		SecondaryIPConfigs: ipconfigs,
-// 	}
+const (
+	primaryIp           = "10.0.0.5"
+	gatewayIp           = "10.0.0.1"
+	dockerContainerType = cns.Docker
+)
 
-// 	svc.CreateOrUpdateNetworkContainerWithSecondaryIPConfigs(nc)
-// }
+func addTestStateToRestServer(t *testing.T, secondaryIps []string) {
+	var ipConfig cns.IPConfiguration
+	ipConfig.DNSServers = []string{"8.8.8.8", "8.8.4.4"}
+	ipConfig.GatewayIPAddress = gatewayIp
+	var ipSubnet cns.IPSubnet
+	ipSubnet.IPAddress = primaryIp
+	ipSubnet.PrefixLength = 32
+	ipConfig.IPSubnet = ipSubnet
+	secondaryIPConfigs := make(map[string]cns.SecondaryIPConfig)
+
+	for _, secIpAddress := range secondaryIps {
+		secIpConfig := cns.SecondaryIPConfig{
+			IPSubnet: cns.IPSubnet{
+				IPAddress:    secIpAddress,
+				PrefixLength: 32,
+			},
+		}
+		ipId, err := uuid.NewUUID()
+		if err != nil {
+			t.Fatalf("Failed to generate UUID for secondaryipconfig, err:%s", err)
+		}
+		secondaryIPConfigs[ipId.String()] = secIpConfig
+	}
+
+	req := cns.CreateNetworkContainerRequest{
+		NetworkContainerType: dockerContainerType,
+		NetworkContainerid:   "testNcId1",
+		IPConfiguration:      ipConfig,
+		SecondaryIPConfigs:   secondaryIPConfigs,
+	}
+
+	returnCode := svc.CreateOrUpdateNetworkContainerInternal(req)
+	if returnCode != 0 {
+		t.Fatalf("Failed to createNetworkContainerRequest, req: %+v, err: %d", req, returnCode)
+	}
+}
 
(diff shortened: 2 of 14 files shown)


cns/restserver/internalapi_test.go
```diff
@@ -0,0 +1,189 @@
...
+			t.Fatal(err)
+		}
+		secondaryIPConfigs[ipId.String()] = secIpConfig
+		startingIndex++
+	}
+
+	createAndValidateNCRequest(t, secondaryIPConfigs)
+
+	// now Scale down, delete 3 ipaddresses from secondaryIpConfig req
+	fmt.Println("Validate Scalesown")
```
ramiro-gamarra, line 78: typo

cns/restserver/internalapi_test.go
```diff
@@ -0,0 +1,189 @@
...
+	var count = 0
+	for ipid, _ := range secondaryIPConfigs {
```
ramiro-gamarra, line 80: if you only need the key, you can just do
```go
for ipid := range secondaryIPConfigs {
```
same for below

cns/restserver/internalapi_test.go
```diff
@@ -0,0 +1,189 @@
...
+		t.Fatal(err)
+	}
+}
+
+func setOrchastratorTypeInternal(orchestratorType string) {
+	fmt.Println("setOrchastratorTypeInternal")
+	svc.state.OrchestratorType = orchestratorType
+}
+
+func createOrUpdateNetworkContainerInternal(t *testing.T, secondaryIpCount int) error {
```
ramiro-gamarra, line 41: does this need to return an error? it seems that at any place where an error would be returned, it instead just fatals.

cns/restserver/internalapi_test.go
```diff
@@ -0,0 +1,189 @@
...
+		t.Fatal()
+	}
+
+	var alreadyValidated = make(map[string]string)
+	for ipid, ipStatus := range svc.PodIPConfigState {
+		if ipaddress, found := alreadyValidated[ipid]; !found {
+			if secondaryIpConfig, ok := req.SecondaryIPConfigs[ipid]; !ok {
+				t.Errorf("PodIpConfigState has stale ipId: %s, config: %+v", ipid, ipStatus)
+				t.Fatal()
+			} else {
```
ramiro-gamarra, line 144: else is not needed. the block above is an early exit from this func

cns/restserver/ipam.go
```diff
@@ -122,132 +92,34 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r
...
-		err := json.Unmarshal(ipconfig.OrchestratorContext, &podInfo)
-
-		// if batch delete failed return
-		if err != nil {
-			return err
+func filterIPConfigMap(toBeAdded map[string]ipConfigurationStatus, f func(*ipConfigurationStatus) bool) []*ipConfigurationStatus {
+	vsf := make([]*ipConfigurationStatus, 0)
+	for _, v := range toBeAdded {
+		if f(&v) {
+			vsf = append(vsf, &v)
```
ramiro-gamarra, line 115: Does this need to return pointers? I believe there is a subtle bug here. The filtered list is being appended a pointer to the loop variable, which points to different values in each iteration. The result is that every item in the filtered list will actually be the same pointer, which will point to the last value added. An example here: https://play.golang.org/p/eE1HIR_KWmF

I would either change the function not to use pointers if possible. Otherwise, just take a copy of the variable and append a pointer to that copy.

cns/restserver/internalapi.go
```diff
@@ -14,3 +19,58 @@ func (service *HTTPRestService) GetPartitionKey() (dncPartitionKey string) {
 	service.RUnlock()
 	return
 }
+
+// This API will be called by CNS RequestController on CRD update.
+func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req cns.CreateNetworkContainerRequest) int {
```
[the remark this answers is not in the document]
neaggarwMS: Good idea. I didnt want to change existing code which uses int to return the error code. Also since some of the APIs are shared , hence I followed the same pattern. eg. saveNetworkContainerGoalState. With above approach I need to make changes to existing API as well.

cns/restserver/internalapi_test.go
```diff
@@ -0,0 +1,189 @@
...
+	fmt.Println("Test: TestCreateOrUpdateNetworkContainerInternal")
+
+	setEnv(t)
+	setOrchastratorTypeInternal(cns.KubernetesCRD)
+
+	err := createOrUpdateNetworkContainerInternal(t, 2)
+	if err != nil {
+		t.Errorf("Failed to save the goal state for network container of type JobObject "+
+			" due to error: %+v", err)
+		t.Fatal(err)
```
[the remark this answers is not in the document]
neaggarwMS: Sure

cns/restserver/internalapi_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+const (
+	primaryIp = "10.0.0.5"
+	gatewayIp = "10.0.0.1"
+	dockerContainerType = cns.Docker
+)
+
+func TestCreateOrUpdateNetworkContainerInternal(t *testing.T) {
+	// requires more than 30 seconds to run
+	fmt.Println("Test: TestCreateOrUpdateNetworkContainerInternal")
```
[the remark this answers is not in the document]
neaggarwMS: Can remove it, i was following existing test conventions

cns/restserver/internalapi_test.go
neaggarwMS: fixed it

cns/restserver/internalapi_test.go
neaggarwMS: yeah I dont need it, will remove it

cns/restserver/ipam.go
matmerr: good catch 👍

cns/restserver/internalapi_test.go
neaggarwMS: it is needed in case entry doesnt exist which is an expected scenario

cns/restserver/internalapi.go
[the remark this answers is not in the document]
ramiro-gamarra: Yeah, I think this is appropriate then if it follows the existing convention 👍

cns/restserver/util.go
```diff
@@ -160,21 +155,157 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(req cns.CreateNetw
...
+// from the inmemory map
+// Note: Also this func is an untransacted API as the caller will take a Service lock
+func (service *HTTPRestService) updateIpConfigsStateUntransacted(req cns.CreateNetworkContainerRequest, existingSecondaryIPConfigs map[string]cns.SecondaryIPConfig) (int, string) {
+	// parse the existingSecondaryIpConfigState to find the deleted Ips
+	newIPConfigs := req.SecondaryIPConfigs
+	var tobeDeletedIpConfigs = make(map[string]cns.SecondaryIPConfig)
+
+	// Populate the ToBeDeleted list, Secondary IPs which doesnt exist in New request anymore.
+	// We will later remove them from the in-memory cache
+	if existingSecondaryIPConfigs != nil {
```
ramiro-gamarra, line 197: FYI There is a difference between an unallocated map (`nil`) and an allocated but empty map. A len check will return 0 for both, thus might be a slightly better check for your purposes. At the same time, `range` will basically be a no-op for both cases.

cns/restserver/util.go
```diff
@@ -160,21 +155,157 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(req cns.CreateNetw
...
+			HostVersion:                   hostVersion}
+
 	service.saveState()
 	return 0, ""
 }
 
+// This func will compute the deltaIpConfigState which needs to be updated (Added or Deleted)
+// from the inmemory map
+// Note: Also this func is an untransacted API as the caller will take a Service lock
+func (service *HTTPRestService) updateIpConfigsStateUntransacted(req cns.CreateNetworkContainerRequest, existingSecondaryIPConfigs map[string]cns.SecondaryIPConfig) (int, string) {
```
ramiro-gamarra, line 190: It's a bit odd for a method receiver to be in a different file than where the type is declared

cns/restserver/util.go
```diff
@@ -160,21 +155,157 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(req cns.CreateNetw
...
+		service.removeToBeDeletedIpsStateUntransacted(ipId, true)
+	}
+
+	// Add the newIpConfigs, ignore if ip state is already in the map
+	service.addIPConfigStateUntransacted(req.NetworkContainerid, newIPConfigs)
+
+	return 0, ""
+}
+
+//AddIPConfigsToState takes a lock on the service object, and will add an array of ipconfigs to the CNS Service.
```
ramiro-gamarra, line 230: The func documentation seems misleading. It has been renamed and is no longer taking locks.

cns/restserver/util.go
```diff
@@ -160,21 +155,157 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(req cns.CreateNetw
...
+		}
+
+		delete(service.PodIPIDByOrchestratorContext, podInfo.GetOrchestratorContextKey())
+	}
+	return nil
+}
+
+//removeUnallocatedIp takes a lock on the service object, and will remove an array of ipconfigs to the CNS Service.
+//Used to add IPConfigs to the CNS pool, specifically in the scenario of rebatching.
+func (service *HTTPRestService) removeToBeDeletedIpsStateUntransacted(ipId string, skipValidation bool) (int, string) {
```
ramiro-gamarra, line 289: Same as above on func documentation

cns/restserver/util.go
```diff
@@ -160,21 +155,157 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(req cns.CreateNetw
...
+			return err
+		}
+
+		delete(service.PodIPIDByOrchestratorContext, podInfo.GetOrchestratorContextKey())
+	}
+	return nil
+}
+
+//removeUnallocatedIp takes a lock on the service object, and will remove an array of ipconfigs to the CNS Service.
+//Used to add IPConfigs to the CNS pool, specifically in the scenario of rebatching.
```
pjohnst5: I think you mean "used to remove IPConfigs.."

cns/restserver/util.go
pjohnst5: It doesn't take a lock

cns/restserver/util.go
neaggarwMS: okay, then i will remove the nil check. Thanks for letting me know.

cns/cnsclient/httpapi/client.go
```diff
@@ -1,6 +1,8 @@
 package httpapi
```
matmerr, line 1: although it's already in master, I think this package name is relatively ambiguous given the two client scenarios of
- External client making requests to CNS (CNI)
- Internal client making function calls to CNS (requestcontroller)

This package is used by the latter, although on first glance I would've assumed the former given that it's the one making http requests

cns/cnsclient/cnsclient_test.go
```diff
@@ -1,35 +1,71 @@
...
+		NetworkContainerid:   "testNcId1",
+		IPConfiguration:      ipConfig,
+		SecondaryIPConfigs:   secondaryIPConfigs,
+	}
 
-// 	svc.CreateOrUpdateNetworkContainerWithSecondaryIPConfigs(nc)
-// }
+	returnCode := svc.CreateOrUpdateNetworkContainerInternal(req)
+	if returnCode != 0 {
+		t.Fatalf("Failed to createNetworkContainerRequest, req: %+v, err: %d", req, returnCode)
```
pjohnst5, line 66: This test is failing right here when called from TestCNSClientRequestAndRelease. The svc.state.OrchestratorType is "kubernetes" for this test and fails on a check in CreateOrUpdateNetworkContainerInternal
image
Run with the "-v" flag and you'll see
`/usr/local/go/bin/go test -v -timeout 30s -tags test github.com/Azure/azure-container-networking/cns/cnsclient -run ^TestCNSClientRequestAndRelease$`

cns/restserver/util.go
```diff
@@ -160,21 +155,157 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(req cns.CreateNetw
...
+			if ipConfigStatus.State == cns.Allocated {
+				errMsg := fmt.Sprintf("Failed to delete an Allocated IP %v", ipConfigStatus)
+				return InconsistentIPConfigState, errMsg
+			}
+		}
+	}
+
+	// now actually remove the deletedIPs
+	for ipId, _ := range tobeDeletedIpConfigs {
+		service.removeToBeDeletedIpsStateUntransacted(ipId, true)
```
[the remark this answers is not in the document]
neaggarwMS: good catch, let me bubble it up.

cns/restserver/util.go
```diff
@@ -160,21 +155,157 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(req cns.CreateNetw
...
+	}
+	if ipSubnet.PrefixLength == 0 {
+		return fmt.Errorf("Failed to add IPConfig to state: %+v, empty IPSubnet.PrefixLength", ipSubnet)
+	}
+	return nil
+}
+
+//cleanupIpConfigState takes a lock on the service object, and will remove an array of ipconfigs to the CNS Service.
+//Used to add IPConfigs to the CNS pool, specifically in the scenario of rebatching.
+func (service *HTTPRestService) cleanupIpConfigState(ipconfigs []ipConfigurationStatus) error {
```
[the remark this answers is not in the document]
neaggarwMS: yeah let me remove it, will improve codecov :)

cns/restserver/util.go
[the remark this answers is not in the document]
neaggarwMS: actually in the remove call, I'm passing skipValidation as true, so that api wont fail. But still good to add the check

cns/cnsclient/httpapi/client.go
neaggarwMS: yes, agreed. Lets fix it in a separate PR.

cns/restserver/util.go
```diff
@@ -160,21 +155,157 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(req cns.CreateNetw
 			service.state.ContainerIDByOrchestratorContext[podInfo.PodName+podInfo.PodNamespace] = req.NetworkContainerid
```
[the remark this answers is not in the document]
neaggarwMS: Dont want to touch an existing code as i dont know what testing it added. So leaving it as it is

cns/restserver/util.go
```diff
@@ -160,21 +155,157 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(req cns.CreateNetw
...
 	return 0, ""
 }
 
+// This func will compute the deltaIpConfigState which needs to be updated (Added or Deleted)
+// from the inmemory map
+// Note: Also this func is an untransacted API as the caller will take a Service lock
+func (service *HTTPRestService) updateIpConfigsStateUntransacted(req cns.CreateNetworkContainerRequest, existingSecondaryIPConfigs map[string]cns.SecondaryIPConfig) (int, string) {
+	// parse the existingSecondaryIpConfigState to find the deleted Ips
+	newIPConfigs := req.SecondaryIPConfigs
+	var tobeDeletedIpConfigs = make(map[string]cns.SecondaryIPConfig)
```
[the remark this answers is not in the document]
neaggarwMS: Lets handle that in rebatching PR if required.

cns/cnsclient/cnsclient_test.go
neaggarwMS: ah good catch, wonder why it passed in my local setup. fixed it

Makefile
```diff
@@ -49,6 +49,9 @@ CNSFILES = \
 	$(wildcard cns/routes/*.go) \
 	$(wildcard cns/service/*.go) \
 	$(wildcard cns/networkcontainers/*.go) \
+	$(wildcard cns/requestcontroller/*.go) \
+	$(wildcard cns/requestcontroller/kubecontroller/*.go) \
+	$(wildcard cns/requestcontroller/example/*.go) \
```
pjohnst5, line 54: I am removing this directory in my pr #608, can you remove this line?

PR #620, opened 2020-07-17
Added check to find accept_ra file exists

tamilmani1989:
**Reason for Change**:

Incase if user disables ipv6 kernel setting, then accept_ra file would not exist and sysctl cmd would fail. This PR addresses that scenario

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/epcommon/endpoint_common.go
+++ b/network/epcommon/endpoint_common.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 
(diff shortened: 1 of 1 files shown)


andyliuliming:
this is a great change,
one of our customer are pending on this issue, they tried to disable the ipv6.
and then the azure cni can not work.

@tamilmani1989 could you please merge this and build a new image? thanks in advance.

PR #623, opened 2020-07-22
Simplify AddToSet logic and fix an inconsistency set type creation.

csfmomo:
fix: an inconsistency set type creation introduced by AddToSet function.

**Reason for Change**:
NPM created 3 types of ipset. AddToSet will check whether set exist before adding entries. However, previous NPM logic only create 1 type of set in AddToSet function which introduced inconsistency between set type. 

**Issue Fixed**:
Simplify logic to create set only when set is not exist for add to set logic.
Fix the inconsistency between set type.

fix: Bug Fixes 🐞
test: Testing 💚
- [X ] adds unit tests

### DIFF
--- a/npm/ipsm/ipsm.go
+++ b/npm/ipsm/ipsm.go
@@ -68,6 +68,20 @@ func (ipsMgr *IpsetManager) Exists(key string, val string, kind string) bool {
 	return true
 }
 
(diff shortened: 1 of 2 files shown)


csfmomo:
Moved to this PR with fork repo
https://github.com/Azure/azure-container-networking/pull/625

PR #625, opened 2020-07-22
Simplify AddToSet logic and fix an inconsistency set type creation.

csfmomo:
fix: an inconsistency set type creation introduced by AddToSet function.

**Reason for Change**:
NPM created 3 types of ipset. AddToSet will check whether set exist before adding entries. However, previous NPM logic only create 1 type of set in AddToSet function which introduced inconsistency between set type.

**Issue Fixed**:
Simplify logic to create set only when set is not exist for add to set logic.
Fix the inconsistency between set type.

  fix: Bug Fixes 🐞X
- [X ] adds unit tests

### DIFF
--- a/npm/ipsm/ipsm.go
+++ b/npm/ipsm/ipsm.go
@@ -68,6 +68,20 @@ func (ipsMgr *IpsetManager) Exists(key string, val string, kind string) bool {
 	return true
 }
 
+// ExistsSet checks whehter an ipset exists.
+func (ipsMgr *IpsetManager) ExistsSet(setName, kind string) bool {
+    m := ipsMgr.setMap
+    if kind == util.IpsetSetListFlag {
+        m = ipsMgr.listMap
+    }
+
+    if _, exists := m[setName]; !exists {
+        return false
+    }
+
+    return true
+}
+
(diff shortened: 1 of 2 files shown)


npm/ipsm/ipsm.go
```diff
@@ -68,6 +68,20 @@ func (ipsMgr *IpsetManager) Exists(key string, val string, kind string) bool {
 	return true
 }
 
+// ExistsSet checks whehter an ipset exists.
+func (ipsMgr *IpsetManager) ExistsSet(setName, kind string) bool {
```
jaer-tsun, line 72: please reverse the order of func name to SetExists

npm/ipsm/ipsm.go
```diff
@@ -68,6 +68,20 @@ func (ipsMgr *IpsetManager) Exists(key string, val string, kind string) bool {
...
+    m := ipsMgr.setMap
+    if kind == util.IpsetSetListFlag {
+        m = ipsMgr.listMap
+    }
+
+    if _, exists := m[setName]; !exists {
```
jaer-tsun, line 78: just leave it as

_, exists := m[setName]

return exists

PR #626, opened 2020-07-23
Implement ReconcileNCState

neaggarwMS:
Implement Init/Reconcile NC state to save NC configuration and save in memory state for Pods which are already allocated with NC's secondary ips.

### DIFF
--- a/cns/cnsclient/httpapi/client.go
+++ b/cns/cnsclient/httpapi/client.go
@@ -24,11 +24,12 @@ func (client *Client) CreateOrUpdateNC(ncRequest cns.CreateNetworkContainerReque
 }
 
 // InitCNSState initializes cns state
-func (client *Client) InitCNSState(ncRequest *cns.CreateNetworkContainerRequest, podInfoByIP map[string]*cns.KubernetesPodInfo) error {
-	// client.RestService.Lock()
-	// client.RestService.ReadyToIPAM = true
-	// client.RestService.Unlock()
+func (client *Client) ReconcileNCState(ncRequest *cns.CreateNetworkContainerRequest, podInfoByIP map[string]cns.KubernetesPodInfo) error {
+	returnCode := client.RestService.ReconcileNCState(ncRequest, podInfoByIP)
 
-	// return client.RestService.AddIPConfigsToState(ipConfigs)
+	if returnCode != 0 {
+		return fmt.Errorf("Failed to Reconcile ncState: ncRequest %+v, podInfoMap: %+v, errorCode: %d", *ncRequest, podInfoByIP, returnCode)
+	}
+
 	return nil
 }
--- a/cns/cnsclient/httpapi/client.go
+++ b/cns/cnsclient/httpapi/client.go
@@ -23,7 +23,7 @@ func (client *Client) CreateOrUpdateNC(ncRequest cns.CreateNetworkContainerReque
 	return nil
 }
 
-// InitCNSState initializes cns state
+// ReconcileNCState initializes cns state
 func (client *Client) ReconcileNCState(ncRequest *cns.CreateNetworkContainerRequest, podInfoByIP map[string]cns.KubernetesPodInfo) error {
 	returnCode := client.RestService.ReconcileNCState(ncRequest, podInfoByIP)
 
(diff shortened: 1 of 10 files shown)


cns/NetworkContainerContract.go
```diff
@@ -98,6 +100,7 @@ type KubernetesPodInfo struct {
 }
 
 // GetOrchestratorContext will return the orchestratorcontext as a string
+// TODO - should use a hashed name or can this be PODUid?
```
thatmattlong, line 103: I think this should be Pod resource ID (GUID) because a Pod can be deleted and recreated with the same name, and we want to know that it's different

cns/cnsclient/httpapi/client.go
```diff
@@ -12,32 +14,22 @@ type Client struct {
...
+	returnCode := client.RestService.CreateOrUpdateNetworkContainerInternal(*ncRequest)
+
+	if returnCode != 0 {
+		return fmt.Errorf("Failed to Create NC request: %+v, errorCode: %d", *ncRequest, returnCode)
+	}
+
+	return nil
 }
 
 // InitCNSState initializes cns state
```
thatmattlong, line 26: Nit: comment will need to reflect new name

cns/cnsclient/httpapi/client.go
```diff
@@ -12,32 +14,22 @@ type Client struct {
...
-func (client *Client) InitCNSState(ncRequest *cns.CreateNetworkContainerRequest, podInfoByIP map[string]*cns.KubernetesPodInfo) error {
```
pjohnst5, line 36: You'll also need to change the name in apiclient.go and in crdrequestconroller_test.go if you change here

pjohnst5: As well as the map type

cns/.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
pjohnst5, line 1: Can you add a line in the .gitignore to ignore this file?

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -231,7 +231,9 @@ func (crdRC *crdRequestController) initCNS() error {
 		}
 	}
 
-	return crdRC.CNSClient.InitCNSState(&ncRequest, podInfoByIP)
+	// Call cnsclient init cns passing those two things
+	return crdRC.CNSClient.ReconcileNCState(ncRequest, podInfoByIP)
```
pjohnst5, line 235: Can you take the address of ncRequest? &ncRequest

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -231,7 +231,9 @@ func (crdRC *crdRequestController) initCNS() error {
 		}
 	}
 
-	return crdRC.CNSClient.InitCNSState(&ncRequest, podInfoByIP)
+	// Call cnsclient init cns passing those two things
```
pjohnst5, line 234: Can you rename the call on line 194 and 204 to "ReconcileNCState" too?

cns/NetworkContainerContract.go
neaggarwMS: Discussed offline, this value is passed to CNI by kubelet. We should discuss this in our syncup that kubelet needs to call cni for first pod deletion before calling add for another.

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
neaggarwMS: Good catch, yes.

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -231,7 +231,9 @@ func (crdRC *crdRequestController) initCNS() error {
...
+	return crdRC.CNSClient.ReconcileNCState(&ncRequest, podInfoByIP)
```
pjohnst5, line 235: Can you also rename the above two calls from InitCNSState to ReconcileNCState? lines 194 and 204

PR #629, opened 2020-07-27
Efficient prometheus

huntergregory:
**Reason for Change**
Number 1 and 2 are important for configuration. 
Number 3 and 4 are important for scraper performance.

Changes: 
1. Moves the port number of the Prometheus endpoint to a high value next to our node port used in CDN
2. Adds a service that forwards a single, random NPM Prometheus endpoint so that a scraper can scrape only one NPM pod for metrics that are redundant across nodes.
3. Removes old IPSet counts when the number of entries is 0 to reduce the amount of lines to scrape.
4. Adds total ipset entries as a Prometheus metric so customers can have this number without having to include the advanced metric - number of entries per ipset.

**Requirements**:
- [X] perf
- [X] includes documentation
- [X] adds unit tests

**Notes**:
Current metrics that are redundant across nodes: 
- npm_num_policies
- npm_num_iptables_rules
- npm_num_ipsets
- npm_num_ipset_entries (new)
- npm_ipset_counts (unlikely, but worst case can have up to 60,000 lines for scraping)

Remaining metrics that are node-specific:
- npm_add_policy_exec_time
- npm_add_iptables_rule_exec_time
- npm_add_ipset_exec_time


npm/azure-npm.yaml
```diff
@@ -114,3 +114,17 @@ spec:
...
+metadata:
+  name: npm-metrics-cluster-service
+  namespace: kube-system
+  labels:
+    app: npm-metrics
+spec:
+  selector:
+    k8s-app: azure-npm
+  ports:
+    - port: 9000
```
csfmomo, line 129: What does this port 9000 do?

npm/metrics/ipset-inventory.go
```diff
@@ -0,0 +1,55 @@
...
+	if exists {
+		return val
+	}
+	return 0
+}
+
+// SetIPSetInventory sets the number of entries in an IPSet and updates a Prometheus metric.
+func SetIPSetInventory(setName string, val int) {
+	_, exists := ipsetInventoryMap[setName]
+	if exists || val != 0 {
```
csfmomo, line 21: Should this be || or && relationship?

npm/azure-npm.yaml
csfmomo, line 129: What does this port 9000 do?

csfmomo:
How did we achieve No.2's goal? What's the service you mentioned in No2?
Does it achieved by ipset_inventory.go file?

npm/azure-npm.yaml
huntergregory: this is just the port for the service (unrelated to the node/host). So you can access this service (which relays a random NPM pod) on port 9000 i.e. `curl npm-metrics-cluster-service.kube-system:9000/metrics`

huntergregory:
> How did we achieve No.2's goal? What's the service you mentioned in No2?
> Does it achieved by ipset_inventory.go file?

Achieved through the new service in azure-npm.yaml: npm-metrics-cluster-service, and by the way you configure the container insights agents, but that's a job for their team. The ipset_inventory.go deletes the metric (line in the scrape file) for every ipset that has no entries/is deleted.

huntergregory:
ran load tests showing that npm with/without prometheus still act the same

huntergregory:
also noting here that having 1 vs 2 HTTP endpoints for an NPM pod seems not to affect performance

PR #635, opened 2020-07-30
Log SecondaryIpConfig in NC snapshot

neaggarwMS:
Log SecondaryIPconfigs detail in NC snapshot event. 

Task: https://msazure.visualstudio.com/One/_workitems/edit/7711831

### DIFF
--- a/cns/restserver/util.go
+++ b/cns/restserver/util.go
@@ -573,7 +573,7 @@ func (service *HTTPRestService) joinNetwork(
 	return joinResponse, joinErr, err
 }
 
-func logNCSnapshot(createNetworkContainerRequest cns.CreateNetworkContainerRequest) {
+func (service *HTTPRestService) logNCSnapshot(createNetworkContainerRequest cns.CreateNetworkContainerRequest) {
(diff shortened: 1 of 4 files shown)


cns/restserver/util.go
```diff
@@ -590,7 +590,9 @@ func logNCSnapshot(createNetworkContainerRequest cns.CreateNetworkContainerReque
 	aiEvent.Properties[logger.NetworkContainerTypeStr] = createNetworkContainerRequest.NetworkContainerType
 	aiEvent.Properties[logger.OrchestratorContextStr] = fmt.Sprintf("%s", createNetworkContainerRequest.OrchestratorContext)
 
-	// TODO - Add for SecondaryIPs (Task: https://msazure.visualstudio.com/One/_workitems/edit/7711831)
+	if service.state.OrchestratorType == cns.KubernetesCRD {
```
ramiro-gamarra, line 593: Does reading `state` need a read lock from `service`? Would it be simpler to just log if len of secondary ip configs > 0?

PR #636, opened 2020-07-30
Dockerfile for cns container

pjohnst5:
**Reason for Change**:
This PR adds a dockerfile for cns. It is an improvement of the previous dockerfile in terms of best practices of containers.
Mainly, the container only has the binary, and builds the binary as part of the dockerfile, instead of how it was doing it before of copying in a pre-built binary into the dockerfile.

Also this new dockerfile only pulls in the necessary dependencies, wheras the old dockerfile brought many unused dependencies into the image.

I'm thinking we can keep both for now, and eventually remove the old one, as well as update the runner yamls to use the new container.

### DIFF
--- a/cns/aks.Dockerfile
+++ b/cns/aks.Dockerfile
@@ -5,7 +5,7 @@ RUN apk --update add ca-certificates
 # Build cns
 FROM golang:1.14-alpine AS builder
 # Build ars
-ARG CNS_VERSION
+ARG VERSION
 ARG CNS_AI_PATH
 ARG CNS_AI_ID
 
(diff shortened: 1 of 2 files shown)


cns/aks.Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Get a trusted CA cert bundle
+FROM alpine:latest as certs
+RUN apk --update add ca-certificates
+
+# Build cns
+FROM golang:1.14-alpine AS builder
+# Build ars
```
thatmattlong, line 7: hyper nit: comment should be args

cns/aks.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ARG VERSION
```
thatmattlong, line 8: maybe set a default like
```
ARG VERSION=dirty
```

cns/aks.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ARG CNS_AI_PATH
```
thatmattlong, line 9: maybe default here as well

pjohnst5: I'm going to leave it mostly because checks take so long to run

cns/aks.Dockerfile
pjohnst5: Same as below

cns/aks.Dockerfile
pjohnst5: Same about checks

PR #637, opened 2020-07-31
CNS yaml update to reflect aks rp helm chart

pjohnst5:
**Reason for Change**:
This is just a dummy fix to update the example cns yaml we have in our repro which mirrors the aks rp.
The yaml doesn't actually do anything other than stand as a reference.

### DIFF
--- a/cns/azure-cns.yaml
+++ b/cns/azure-cns.yaml
@@ -1,3 +1,55 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: azure-cns
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: kube-system
+  name: nodeNetConfigEditor
+rules:
+  - apiGroups: ["acn.azure.com"]
+    resources: ["nodenetworkconfigs"]
+    verbs: ["get", "list", "watch", "patch", "update"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pod-reader-all-namespaces
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "watch", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: nodeNetConfigEditorRoleBinding
+  namespace: kube-system
+subjects:
+- kind: ServiceAccount
+  name: azure-cns
+  namespace: kube-system
+roleRef:
+  kind: Role
+  name: nodeNetConfigEditor
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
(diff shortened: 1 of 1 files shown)


cns/azure-cns.yaml
```diff
@@ -1,3 +1,55 @@
...
+  name: nodeNetConfigEditor
+rules:
+  - apiGroups: ["acn.azure.com"]
+    resources: ["nodenetworkconfigs"]
+    verbs: ["get", "list", "watch", "patch", "update"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pod-reader-all-namespaces
```
thatmattlong, line 20: nit: ClusterRole implies all namespaces, probably doesn't need to be in the name

cns/azure-cns.yaml
```diff
@@ -1,3 +1,55 @@
...
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "watch", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: nodeNetConfigEditorRoleBinding
```
thatmattlong, line 29: nit: same as above, this is a RoleBinding so no need to put it in the name

cns/azure-cns.yaml
```diff
@@ -1,3 +1,55 @@
...
+  namespace: kube-system
+roleRef:
+  kind: Role
+  name: nodeNetConfigEditor
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: pod-reader-all-namespaces-binding
```
thatmattlong, line 43: same as above re: naming

cns/azure-cns.yaml
```diff
@@ -16,47 +68,96 @@ spec:
       annotations:
         cluster-autoscaler.kubernetes.io/daemonset-pod: "true"
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: kubernetes.azure.com/cluster
```
thatmattlong, line 76: I think we should remove this key, I'm not sure we want to require a particular annotation that is not well-known, e.g. `node-role.kubernetes.io/master`)

pjohnst5: This was there already from Wei's change awhile back I'm not sure why though
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/2928590?_a=files&path=%2Fccp%2Fcharts%2Fkube-control-plane%2Fcharts%2Fkube-addons%2Ftemplates%2Faddons%2F_azure-cns.yaml

pjohnst5: Checked with Wei and this label is from aks-engine so it's present on all nodes
https://github.com/Azure/aks-engine/blob/0ca3854c98cc9d1386a0339949c155fe636a6184/pkg/api/types.go#L1740

PR #648, opened 2020-08-12
cns_removePrefixLengthFromSecondaryIpconfig

neaggarwMS:
Removed PrefixLength from SecondaryIPConfig struct. We dont need to populate /32 in this struct. When CNI request for SecondaryIPconfig, then add Subnet range prefix length it the IP CIDR format.


neaggarwMS:
azp /run

PR #649, opened 2020-08-13
fix: pass podinfo to CNS delete

matmerr:
**Reason for Change**:

IPAM delete in CNS was not passing pod info, therefore ID in state was never found

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/util.go
+++ b/cns/restserver/util.go
@@ -622,23 +622,23 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
 	}
 }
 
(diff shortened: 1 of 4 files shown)


cns/restserver/util.go
```diff
@@ -622,22 +622,23 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ncSnapshotIntervalInM
 	}
 }
 
-func (service *HTTPRestService) validateIpConfigRequest(ipConfigRequest cns.GetIPConfigRequest) (int, string) {
+func (service *HTTPRestService) validateIpConfigRequest(ipConfigRequest cns.GetIPConfigRequest) (*cns.KubernetesPodInfo, int, string) {
```
thatmattlong, line 625: I think it would be safer to return a `cns.KubernetesPodInfo` instead of a pointer

PR #656, opened 2020-08-17
Add error log and metrics to AI telemetry.

csfmomo:
feat: add error log and metrics to AI telemetry.

chore: Maintenance 🔧
style: Code Style 🎶

**Notes**:
1. Refactor AI telemetry initialization. Move it from npm.go to ai-utils.go under metrics pakcage.
2. Move constants into constant.go under util package.
3. Add SendErrorMetric function to handle error log / metrics emitting to AI telemtry.
4. Only send error metrics from iptm, ipsm and npm these 3 packages now for incremental delivery purpose.


npm/ipsm/ipsm.go
```diff
@@ -96,6 +96,7 @@ func (ipsMgr *IpsetManager) CreateList(listName string) error {
 	log.Logf("Creating List: %+v", entry)
 	if errCode, err := ipsMgr.Run(entry); err != nil && errCode != 1 {
 		log.Errorf("Error: failed to create ipset list %s.", listName)
+		metrics.SendErrorMetric(util.IpsmID, "Error: failed to create ipset list %s.", listName)
```
matmerr, line 99: Can we use `fmt.Sprintf` to format a single string error message and use that string for both `log.Errorf()` and `metrics.SendErrorMetric()`? I think that would keep consistency between sources

csfmomo: Sure, I can do that.

npm/metrics/ai-utils.go
```diff
@@ -0,0 +1,91 @@
...
+			return
+	}
+	th.TrackMetric(metric)
+}
+
+// SendLog sends log
+func SendLog(report aitelemetry.Report) {
+	if th == nil {
+			log.Logf("AppInsights didn't initialized.")
+			return
```
ramiro-gamarra, line 87: formatting seems off, I think gofmt may not have been run

npm/metrics/ai-utils.go
```diff
@@ -0,0 +1,91 @@
...
+		GetEnvRetryCount:          util.GetEnvRetryCount,
+		GetEnvRetryWaitTimeInSecs: util.GetEnvRetryWaitTimeInSecs,
+	}
+
+	var err error
+	th, err = aitelemetry.NewAITelemetry("", aiMetadata, aiConfig)
+
+	for i := 0; err != nil && i < util.AiInitializeRetryCount; i++ {
+		log.Logf("Failed to init AppInsights with err: %+v for %d time", err, i + 1)
+		time.Sleep(time.Minute * 5)
```
ramiro-gamarra, line 36: This sleeps for 5 minutes on every iteration, blocking the main function if I read correctly. That may be too long to sleep.

npm/ipsm/ipsm.go
ramiro-gamarra: I think we can even go a step further :). In order to keep things DRY, it seems you can just create or wrap an error, bubble it as its already being done, and move all the logging and metrics to the caller. That would avoid having to log and send a metric on every error in this section. Let me know if that makes sense.

npm/util/const.go
```diff
@@ -124,4 +124,24 @@ const (
...
+
+	BatchSize                 int = 32768
+	BatchInterval             int = 30
+	RefreshTimeout            int = 15
+	GetEnvRetryCount          int = 5
+	GetEnvRetryWaitTimeInSecs int = 3
+	AiInitializeRetryCount    int = 5
+	NpmID                     int = 1
+	IpsmID                    int = 2
+	IptmID                    int = 3
```
ramiro-gamarra, line 141: Can we add a comment to some of these? It's hard to know if intervals or timeouts are in secs, mins. Some also seems a bit arbitrary (32768). Are the ids referenced from other places?

csfmomo: For wrapping log and metrics, yes, I do think in same way to avoid duplicated logic.

npm/ipsm/ipsm.go
csfmomo: For wrapping log and metrics, yes, I do think in same way to avoid duplicated logic.

npm/metrics/ai-utils.go
csfmomo: I moved this initialization logic from npm.go from line 117 to 121. It is in a separate go routine previously. How about I keeep this retry logic but move it to a separate go routine?

npm/util/const.go
csfmomo: Sure

csfmomo: image

npm/metrics/ai-utils.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	var err error
+	for i := 0; i < util.AiInitializeRetryCount; i++ {
+		th, err = aitelemetry.NewAITelemetry("", aiMetadata, aiConfig)
+		if err != nil {
+			log.Logf("Failed to init AppInsights with err: %+v for %d time", err, i+1)
+			time.Sleep(time.Minute * time.Duration(util.AiInitializeRetryInMin))
+		} else {
+			i = util.AiInitializeRetryCount
```
ramiro-gamarra, line 38: in Go it's more idiomatic to `break` from the for loop

ramiro-gamarra (commented):
One last comment

npm/metrics/ai-utils.go
csfmomo: Nice! Break makes sense and it's a general practice for other languages too.

npm/util/const.go
csfmomo: 32768 came from this AI telemetry doc
 https://docs.microsoft.com/en-us/azure/azure-monitor/app/pricing
 
image

ramiro-gamarra (approved):
lgtm

matmerr (approved):
lgtm

PR #664, opened 2020-08-27
Improve TestDestroy unit test validation to make it not flaky.

csfmomo:
fix: TestDestroy validation
**Reason for Change**:
TestDestroy unit test is flaky sometimes since it can be influenced by other ipsets exist in the system. If there are ipsets got referenced from kernel, flush or destroy won't delete it but return error. TestDestroy validation should changed to validate whether the testing ipset got deleted or not.

**Issue Fixed**:
Flaky unit test.

test: Testing 💚
- [X ] adds unit tests

### DIFF
--- a/npm/ipsm/ipsm_test.go
+++ b/npm/ipsm/ipsm_test.go
@@ -244,7 +244,7 @@ func TestDeleteSet(t *testing.T) {
 		}
 	}()
 
-	testSetName := "test-set"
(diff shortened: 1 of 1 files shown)


csfmomo:
Moved to this PR
https://github.com/Azure/azure-container-networking/pull/671

csfmomo:
Moved to PR 671, close this one

PR #668, opened 2020-09-04
CNS Scaling Integration (3/3)

matmerr:
**Reason for Change**:

Integrates the IPAM Pool Monitor with existing request controller and CNS.

DO NOT MERGE BEFORE https://github.com/Azure/azure-container-networking/pull/667, this PR is continually rebased off that one.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/ipampoolmonitor/ipampoolmonitor.go
+++ b/cns/ipampoolmonitor/ipampoolmonitor.go
@@ -65,36 +65,37 @@ func (pm *CNSIPAMPoolMonitor) Reconcile() error {
 }
 
 // UpdatePoolLimitsTransacted called by request controller on reconcile to set the batch size limits
-func (pm *CNSIPAMPoolMonitor) UpdatePoolLimitsTransacted(scalarUnits cns.ScalarUnits) {
+func (pm *CNSIPAMPoolMonitor) UpdatePoolLimits(scalarUnits cns.ScalarUnits) {
 	pm.Lock()
 	defer pm.Unlock()
 	pm.scalarUnits = scalarUnits
-
-	if !pm.initialized {
-		pm.goalIPCount = pm.scalarUnits.IPConfigCount
-	}
 
 	// TODO rounding?
 	pm.MinimumFreeIps = int(float64(pm.scalarUnits.BatchSize) * (float64(pm.scalarUnits.RequestThresholdPercent) / 100))
 	pm.MaximumFreeIps = int(float64(pm.scalarUnits.BatchSize) * (float64(pm.scalarUnits.ReleaseThresholdPercent) / 100))
 
-	pm.initialized = true
+	if !pm.initialized && len(pm.cns.GetPodIPConfigState()) > 0 {
+		pm.goalIPCount = len(pm.cns.GetPodIPConfigState())
+		pm.initialized = true
+	}
 }
 
 func (pm *CNSIPAMPoolMonitor) checkForResize() int {
 
-	allocatedIPCount := len(pm.cns.GetAllocatedIPConfigs())
-	freeIPConfigCount := pm.goalIPCount - allocatedIPCount
+	podIPCount := len(pm.cns.GetAllocatedIPConfigs()) + pm.cns.GetPendingAllocationIPCount() // TODO: add pending allocation count to real cns
+	freeIPConfigCount := pm.goalIPCount - podIPCount
 
 	switch {
 	// pod count is increasing
+	case podIPCount == 0:
+		logger.Printf("No pods scheduled")
+		return doNothing
+
 	case freeIPConfigCount < pm.MinimumFreeIps:
-		logger.Printf("Number of free IP's (%d) < minimum free IPs (%d), request batch increase\n", freeIPConfigCount, pm.MinimumFreeIps)
 		return increasePoolSize
 
 	// pod count is decreasing
 	case freeIPConfigCount > pm.MaximumFreeIps:
-		logger.Printf("Number of free IP's (%d) > maximum free IPs (%d), request batch decrease\n", freeIPConfigCount, pm.MaximumFreeIps)
 		return decreasePoolSize
 	}
 	return doNothing
@@ -110,6 +111,7 @@ func (pm *CNSIPAMPoolMonitor) increasePoolSize() error {
 		return err
 	}
 
+	logger.Printf("Increasing pool size, Current Pool Size: %v, Existing Goal IP Count: %v, Pods with IP's:%v, Pods waiting for IP's %v", len(pm.cns.GetPodIPConfigState()), pm.goalIPCount, len(pm.cns.GetAllocatedIPConfigs()), pm.cns.GetPendingAllocationIPCount())
 	return pm.rc.UpdateCRDSpec(context.Background(), pm.cachedSpec)
 }
 
@@ -132,6 +134,7 @@ func (pm *CNSIPAMPoolMonitor) decreasePoolSize() error {
 		return err
 	}
 
+	logger.Printf("Decreasing pool size, Current Pool Size: %v, Existing Goal IP Count: %v, Pods with IP's:%v", len(pm.cns.GetPodIPConfigState()), pm.goalIPCount, pm.cns.GetAllocatedIPConfigs())
 	return pm.rc.UpdateCRDSpec(context.Background(), pm.cachedSpec)
 }
 
--- a/cns/ipampoolmonitor/ipampoolmonitor.go
+++ b/cns/ipampoolmonitor/ipampoolmonitor.go
@@ -3,6 +3,7 @@ package ipampoolmonitor
 import (
 	"context"
 	"log"
+	"math"
 	"sync"
 
 	"github.com/Azure/azure-container-networking/cns"
@@ -11,9 +12,9 @@ import (
 )
 
 var (
-	increasePoolSize = 1
-	decreasePoolSize = -1
-	doNothing        = 0
+	increasePoolSize = int64(1)
+	decreasePoolSize = int64(-1)
+	doNothing        = int64(0)
 )
 
 type CNSIPAMPoolMonitor struct {
@@ -23,10 +24,8 @@ type CNSIPAMPoolMonitor struct {
 	cns            cns.HTTPService
 	rc             requestcontroller.RequestController
 	scalarUnits    cns.ScalarUnits
-	MinimumFreeIps int
-	MaximumFreeIps int
-
-	goalIPCount int
+	MinimumFreeIps int64
+	MaximumFreeIps int64
 
 	sync.RWMutex
 }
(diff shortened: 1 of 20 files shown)


cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
 )
 
 var (
-	increasePoolSize = 1
-	decreasePoolSize = -1
-	doNothing        = 0
+	increasePoolSize = int64(1)
+	decreasePoolSize = int64(-1)
+	doNothing        = int64(0)
```
ramiro-gamarra, line 18: these seem unused

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
...
 
-// UpdatePoolLimitsTransacted called by request controller on reconcile to set the batch size limits
-func (pm *CNSIPAMPoolMonitor) UpdatePoolLimitsTransacted(scalarUnits cns.ScalarUnits) {
-	pm.Lock()
-	defer pm.Unlock()
-	pm.scalarUnits = scalarUnits
+func (pm *CNSIPAMPoolMonitor) Start(poolMonitorRefreshMilliseconds int, exitChan <-chan struct{}) error {
+	logger.Printf("[ipam-pool-monitor] Starting CNS IPAM Pool Monitor")
+	for {
+		if stopReconcile(exitChan) {
```
ramiro-gamarra, line 59: I gather the intent is to have a cancellable loop that does something on an interval. The pattern for that is to use a context and a ticker. Then you can do something like
```go
for {
    select {
    case <-ctx.Done():
        return fmt.Errorf("CNS IPAM Pool Monitor received cancellation signal")
    case <- ticker.C:
        // do things
    }
}
```

The reason that is preferred to `time.Sleep`ing is that there's no way to cancel the goroutine while it sleeps. If `Reconcile` is also a potentially blocking method, it should probably also accept a context.

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
...
+			return fmt.Errorf("CNS IPAM Pool Monitor received cancellation signal")
+		}
 
-	// TODO rounding?
-	pm.MinimumFreeIps = int(pm.scalarUnits.BatchSize * (pm.scalarUnits.RequestThresholdPercent / 100))
-	pm.MaximumFreeIps = int(pm.scalarUnits.BatchSize * (pm.scalarUnits.ReleaseThresholdPercent / 100))
+		err := pm.Reconcile()
+		if err != nil {
+			logger.Printf("[ipam-pool-monitor] CRITICAL %v", err)
```
ramiro-gamarra, line 65: This is probably for offline discussion, but are there any cases where an error from reconciliation should prompt the app to crash and be restarted?

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
...
-	decreasePoolSize = -1
-	doNothing        = 0
+	increasePoolSize = int64(1)
+	decreasePoolSize = int64(-1)
+	doNothing        = int64(0)
 )
 
 type CNSIPAMPoolMonitor struct {
-	initialized bool
+	initialized    bool
```
ramiro-gamarra, line 22: I'm curious about the reason we need an initialized flag. Is there a way that the pool monitor can be hydrated on construction? That simplifies the rest of the methods, where we can assume that it's been initialized properly and that dependencies are there.

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
...
-		logger.Printf("Number of free IP's (%d) < minimum free IPs (%d), request batch increase\n", freeIPConfigCount, pm.MinimumFreeIps)
-		return increasePoolSize
-
-	// pod count is decreasing
-	case freeIPConfigCount > pm.MaximumFreeIps:
-		logger.Printf("Number of free IP's (%d) > maximum free IPs (%d), request batch decrease\n", freeIPConfigCount, pm.MaximumFreeIps)
-		return decreasePoolSize
+func (pm *CNSIPAMPoolMonitor) Reconcile() error {
+	if pm.initialized {
+		cnsPodIPConfigCount := len(pm.cns.GetPodIPConfigState())
```
ramiro-gamarra, line 74: Happy path is usually best unindented, so I would flip the condition and return early. Even cleaner if we can skip the check and assume things have been initialized :)

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
...
+		// if there's a pending change to the spec count, and the pending release state is nonzero,
+		// skip so we don't thrash the UpdateCRD
+		if pm.cachedSpec.RequestedIPCount != int64(len(pm.cns.GetPodIPConfigState())) && len(pm.cns.GetPendingReleaseIPConfigs()) > 0 {
+			return nil
+		}
+
+		switch {
+		// pod count is increasing
+		case freeIPConfigCount < pm.MinimumFreeIps:
+			logger.Printf("[ipam-pool-monitor] Increasing pool size...")
```
ramiro-gamarra, line 91: can some more context be added to logging? like the variables used to trigger the increase (`freeIPConfigCount`, `pm.MinimumFreeIps`). Same for similar cases below

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
...
+		// CRD has reconciled CNS state, and target spec is now the same size as the state
+		// free to remove the IP's from the CRD
+		case pm.pendingRelease && int(pm.cachedSpec.RequestedIPCount) == cnsPodIPConfigCount:
+			logger.Printf("[ipam-pool-monitor] Removing Pending Release IP's from CRD...")
+			return pm.cleanPendingRelease()
+
+		// no pods scheduled
+		case allocatedPodIPCount == 0:
+			logger.Printf("[ipam-pool-monitor] No pods scheduled")
+			return fmt.Errorf("No pods scheduled")
```
ramiro-gamarra, line 108: shouldn't this just be a no-op?

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
...
-	// pod count is increasing
-	case freeIPConfigCount < pm.MinimumFreeIps:
-		logger.Printf("Number of free IP's (%d) < minimum free IPs (%d), request batch increase\n", freeIPConfigCount, pm.MinimumFreeIps)
-		return increasePoolSize
-
-	// pod count is decreasing
-	case freeIPConfigCount > pm.MaximumFreeIps:
-		logger.Printf("Number of free IP's (%d) > maximum free IPs (%d), request batch decrease\n", freeIPConfigCount, pm.MaximumFreeIps)
-		return decreasePoolSize
+func (pm *CNSIPAMPoolMonitor) Reconcile() error {
```
ramiro-gamarra, line 72: This should probably take a lock as well. While the loop above might only run this in a single goroutine, the internals of the pool monitor can also be changed using `UpdatePoolMonitor` which could be executed at the same time.

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
...
 
 	cns            cns.HTTPService
 	rc             requestcontroller.RequestController
 	scalarUnits    cns.ScalarUnits
-	MinimumFreeIps int
-	MaximumFreeIps int
+	MinimumFreeIps int64
+	MaximumFreeIps int64
 
 	sync.RWMutex
```
ramiro-gamarra, line 34: Consider naming this mutex instead of embedding it. Embedding it makes the lock/unlock methods exported as well, which is something calling code should not be aware of.

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -11,121 +13,193 @@ import (
...
+func (pm *CNSIPAMPoolMonitor) UpdatePoolMonitor(scalarUnits cns.ScalarUnits) error {
+	pm.Lock()
+	defer pm.Unlock()
+	pm.scalarUnits = scalarUnits
+
+	pm.MinimumFreeIps = int64(float64(pm.scalarUnits.BatchSize) * (float64(pm.scalarUnits.RequestThresholdPercent) / 100))
+	pm.MaximumFreeIps = int64(float64(pm.scalarUnits.BatchSize) * (float64(pm.scalarUnits.ReleaseThresholdPercent) / 100))
+
+	if pm.cns == nil {
+		return fmt.Errorf("Error Updating Pool Limits, reference to CNS is nil")
```
ramiro-gamarra, line 196: Bit of a defensive check to be done at runtime. It's probably a programming error if we don't have a cns reference at this point.

cns/logger/log.go
```diff
@@ -154,7 +154,7 @@ func Response(tag string, response interface{}, returnCode int, returnStr string
 
 	var msg string
 	if err == nil && returnCode == 0 {
-		msg = fmt.Sprintf("[%s] Sent %T %+v.", tag, response, response)
+		//msg = fmt.Sprintf("[%s] Sent %T %+v.", tag, response, response)
```
ramiro-gamarra, line 157: Are these changes intentional?

ramiro-gamarra (commented):
Need to review further but leaving some initial feedback for ipampoolmonitor

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -2,130 +2,197 @@
...
 
-	// TODO rounding?
-	pm.MinimumFreeIps = int(pm.scalarUnits.BatchSize * (pm.scalarUnits.RequestThresholdPercent / 100))
-	pm.MaximumFreeIps = int(pm.scalarUnits.BatchSize * (pm.scalarUnits.ReleaseThresholdPercent / 100))
+	logger.Printf("[ipam-pool-monitor] Checking pool for resize, Pool Size: %v, Goal Size: %v, BatchSize: %v, MinFree: %v, MaxFree:%v, Allocated: %v, Available: %v, Pending Release: %v, Free: %v", cnsPodIPConfigCount, pm.cachedSpec.RequestedIPCount, pm.scalarUnits.BatchSize, pm.MinimumFreeIps, pm.MaximumFreeIps, allocatedPodIPCount, availableIPConfigCount, pendingReleaseIPCount, freeIPConfigCount)
 
-	pm.initialized = true
-}
+	// if there's a pending change to the spec count, and the pending release state is nonzero,
+	// skip so we don't thrash the UpdateCRD
```
neaggarwMS, line 75: It is okay to updateCRD and request for more from CNS. We wont be able to update the request anyways allocatedCount wont increase until the next batch is ready. We dont need this check

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -2,130 +2,197 @@
...
+	defer pm.mu.Unlock()
+	pm.scalarUnits = scalarUnits
+
+	pm.MinimumFreeIps = int64(float64(pm.scalarUnits.BatchSize) * (float64(pm.scalarUnits.RequestThresholdPercent) / 100))
+	pm.MaximumFreeIps = int64(float64(pm.scalarUnits.BatchSize) * (float64(pm.scalarUnits.ReleaseThresholdPercent) / 100))
+
+	if pm.cns == nil {
+		return fmt.Errorf("Error Updating Pool Limits, reference to CNS is nil")
+	}
+
```
neaggarwMS, line 194: Initial RequestedIpCount should be same as what is specified in the spec. Say you updated the CRP spec RequestedCount to 20 and then CNS died. On reconcile we want to make sure we cache the CRD spec. (Both RequestedCount + IPs which are marked as pending release)

cns/restserver/internalapi.go
```diff
@@ -12,10 +12,8 @@
 	"reflect"
```
neaggarwMS, line 13: This should also pass the CRD spec details. And we should cache the complete spec.

Also as part of reconcile, ips which are listed in ToBeDeleted list must be marked as PendingRelease if they still exists in PodIpConfigState.

cns/ipampoolmonitor/ipampoolmonitor.go
neaggarwMS, line 75: Remove this check, not required.

cns/ipampoolmonitor/ipampoolmonitor.go
neaggarwMS, line 194: RequestedCount should be picked from the crd spec. 
Cachedspec should match the CRD spec. 

1. Created NC, SecIpCount = len(podipConfigState) = 10, Cachedspec.RequestedCount = 10 
2. Updated CRD.spec.RequestedCount = 20
3. CNS crashed
4. on restart, CND wil recreate NC, SecIpCount = len(podIpConfigstate)=10, Cachedspec.RequestedCount = 10

Also we are missing the ToBeDeletedIPs

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -2,130 +2,197 @@
...
-var (
-	increasePoolSize = 1
-	decreasePoolSize = -1
-	doNothing        = 0
-)
-
 type CNSIPAMPoolMonitor struct {
-	initialized bool
+	pendingRelease bool
+
```
neaggarwMS, line 17: Save the whole spec.

neaggarwMS (approved):

neaggarwMS (approved):

neaggarwMS (approved):

PR #670, opened 2020-09-06
Improve formatting according to goimports tool suggestion.

csfmomo:
improve: formatting according to goimports suggestion.

**Reason for Change**:
Part of npm files didn't formatting well. Using formatting tool to make code style consistent. 

style: Code Style 🎶


npm/namespace.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Microsoft. All rights reserved.
+// Package npm use Copyright 2018 Microsoft. All rights reserved.
```
ramiro-gamarra, line 1: What was the linting warning here? The change here seems strange. If there is no simple way to document what the `npm` package does, then better to just leave as is. Same for all similar below.

npm/nwpolicy_test.go
```diff
@@ -78,14 +78,17 @@ func TestAddNetworkPolicy(t *testing.T) {
...
 					From: []networkingv1.NetworkPolicyPeer{
-						networkingv1.NetworkPolicyPeer{
+						// networkingv1.NetworkPolicyPeer
+						{
 							PodSelector: &metav1.LabelSelector{
 								MatchLabels: map[string]string{"app": "test"},
 							},
 						},
-						networkingv1.NetworkPolicyPeer{
+						// networkingv1.NetworkPolicyPeer
```
ramiro-gamarra, line 90: I gather these were just redundant type declarations. Better remove than comment. Same for all below.

ramiro-gamarra (changes requested):
Let me know if we should discuss the comments

PR #672, opened 2020-09-16
Add PendingRelease to CNS on start

matmerr:
**Reason for Change**:

**Issue Fixed**:

Adds the pending release from the CRD to CNS on start

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -5,7 +5,6 @@ package restserver
 
 import (
 	"bytes"
-	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -18,10 +17,6 @@ import (
 	"github.com/Azure/azure-container-networking/common"
 	"github.com/Azure/azure-container-networking/log"
 	nnc "github.com/Azure/azure-container-networking/nodenetworkconfig/api/v1alpha"
-)
-
-var (
-	poolIPAMRefreshRateInMilliseconds = 1000
 )
 
 // This file contains the internal functions called by either HTTP APIs (api.go) or
@@ -196,12 +191,11 @@ func (service *HTTPRestService) ReconcileNCState(ncRequest *cns.CreateNetworkCon
 		}
 	}
 
-	ctx := context.Background()
-	go func() {
-		if err := service.IPAMPoolMonitor.Start(ctx, poolIPAMRefreshRateInMilliseconds); err != nil {
-			logger.Errorf("[Azure CNS] Failed to start pool monitor with err: %v", err)
-		}
-	}()
+	err := service.MarkExistingIPsAsPending(spec.IPsNotInUse)
+	if err != nil {
+		logger.Errorf("[Azure CNS] Error. Failed to mark IP's as pending %v", spec.IPsNotInUse)
+		return UnexpectedError
+	}
 
 	return 0
 }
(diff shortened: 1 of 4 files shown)


cns/restserver/internalapi.go
```diff
@@ -191,6 +196,13 @@ func (service *HTTPRestService) ReconcileNCState(ncRequest *cns.CreateNetworkCon
 		}
 	}
 
+	ctx := context.Background()
```
matmerr, line 199: need a cancellable context, but this function is short lived

ramiro-gamarra: Curious, why is this being moved from main? It seems like in this scope (`ReconcileNCState`) the monitor would get "started" multiple times rather than just once.

cns/restserver/ipam.go
```diff
@@ -194,6 +194,26 @@ func (service *HTTPRestService) releaseIPConfig(podInfo cns.KubernetesPodInfo) e
...
+	for _, id := range pendingIPIDs {
+		if ipconfig, exists := service.PodIPConfigState[id]; exists {
+			if ipconfig.State == cns.Allocated {
+				return fmt.Errorf("Failed to mark IP [%v] as pending, currently allocated", id)
+			}
+
+			ipconfig.State = cns.PendingRelease
+			service.PodIPConfigState[id] = ipconfig
+		} else {
+			return fmt.Errorf("Inconsistent state, ipconfig with ID [%v] marked as pending release, but does not exist in state", id)
```
ramiro-gamarra, line 211: Curious, would it be better to try to fix the spec at this point (i.e. garbage collect the ip id that is not referenced anywhere else)? Maybe that's done somewhere else but I'm not aware.

ramiro-gamarra (commented):
Couple of questions

cns/restserver/internalapi.go
```diff
@@ -191,6 +196,13 @@
...
```
neaggarwMS, line 199: +1 on Ramiro's feedback. This should be called from main only after Reconcile is completed. It could be possible that reconcile api is skipped if CRD is not available but IPAM pool monitor should still start.

cns/restserver/internalapi.go
```diff
@@ -245,12 +257,17 @@
 		logger.Errorf(returnMessage)
 	}
 
+	err = service.MarkExistingIPsAsPending(spec.IPsNotInUse)
```
neaggarwMS, line 260: We dont need to call this during all createOrUpdateNC calls. This only needs to be marked one time during Reconcile. Can we move this to the caller API (ReconcileNCState)

cns/restserver/ipam.go
neaggarwMS: This can be an expected state. say during reconcile, spec had some stale ToBeDeleted ips list which were already cleanup in status. Thus, PodIpConfigState wont have that ip. We can log this but this is not an error.

neaggarwMS (changes requested):

cns/restserver/ipam.go
```diff
@@ -194,6 +194,26 @@ func (service *HTTPRestService) releaseIPConfig(podInfo cns.KubernetesPodInfo) e
...
 
+// called when CNS is starting up and there are existing ipconfigs in the CRD that are marked as pending
+func (service *HTTPRestService) MarkExistingIPsAsPending(pendingIPIDs []string) error {
+	service.Lock()
+	defer service.Unlock()
+
+	for _, id := range pendingIPIDs {
+		if ipconfig, exists := service.PodIPConfigState[id]; exists {
+			if ipconfig.State == cns.Allocated {
+				return fmt.Errorf("Failed to mark IP [%v] as pending, currently allocated", id)
```
weng271190436, line 205: I am not super familiar with this but should we still try to set PendingRelease for subsequent IPs when one of them has allocated state?

neaggarwMS (approved):

PR #673, opened 2020-09-17
CNS CRD nullref fix

matmerr:
**Reason for Change**:

Fixes an issue where if the CRD is nil, then don't pass to IPAM pool monitor

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pjohnst5 (approved):
Looks good to me, talked about this in teams chat, and this is good in case the CRD is defined on the cluster, but the NNC instance of the crd isn't there yet, I missed that

neaggarwMS (approved):

PR #676, opened 2020-09-21
Add logging to record CNI_COMMAND value

ninzavivek:
We have seen cases where our CNI is getting called for COMMAND with commands other ADD/DEL.

CHECK is only supported for 0.4.0 and above, and current logging gives an impression CNI DID nothing for ADD/DEL calls.

### DIFF
--- a/cni/network/plugin/main.go
+++ b/cni/network/plugin/main.go
@@ -225,6 +225,10 @@ func main() {
 		panic("network plugin start fatal error")
 	}
 
+	// Check CNI_COMMAD value for logging purposes.
+	cmd := os.Getenv("CNI_COMMAND")
+	log.Errorf("CNI_COMMAND environment variable set to %s", cmd)
+
 	handled, err := handleIfCniUpdate(netPlugin.Update)
 	if handled == true {
 		log.Printf("CNI UPDATE finished.")
--- a/cni/network/plugin/main.go
+++ b/cni/network/plugin/main.go
@@ -225,9 +225,9 @@ func main() {
 		panic("network plugin start fatal error")
 	}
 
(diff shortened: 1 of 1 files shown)


cni/network/plugin/main.go
```diff
@@ -225,6 +225,10 @@ func main() {
 		panic("network plugin start fatal error")
 	}
 
+	// Check CNI_COMMAD value for logging purposes.
```
ramiro-gamarra, line 228: small typo

cni/network/plugin/main.go
```diff
@@ -225,6 +225,10 @@ func main() {
...
+	cmd := os.Getenv("CNI_COMMAND")
```
ramiro-gamarra, line 229: can we extract to a const?

cni/network/plugin/main.go
```diff
@@ -225,6 +225,10 @@ func main() {
...
+	log.Errorf("CNI_COMMAND environment variable set to %s", cmd)
```
ramiro-gamarra, line 230: `log.Printf`? This seems more akin to INFO level logging

ramiro-gamarra (commented):
few comments

cni/network/plugin/main.go
ninzavivek: fixed.

cni/network/plugin/main.go
ninzavivek: resolved.

cni/network/plugin/main.go
ninzavivek: resolved.

ramiro-gamarra (approved):

PR #679, opened 2020-09-22
test: stop skipping aks-engine restart test

pjohnst5:
**Reason for Change**:
We want to stop skipping the restart test in aks-engine e2e test suite. [code](https://github.com/Azure/aks-engine/blob/c670571cd42c68ecddca9a9379a5a1d3b0cd6a69/test/e2e/kubernetes/kubernetes_test.go#L812)
Can do this by setting `DEBUG_CRASHING_PODS=true` in pipeline


matmerr (approved):
Verified in pipeline, lgtm

PR #680, opened 2020-09-29
mNAT + LB support for swift containers

vakalapa:
**Reason for Change**:
mNAT + LB support changes in CNS

- [x] includes documentation
- [x] adds unit tests

**Notes**:


cns/restserver/internalapi.go
```diff
@@ -254,3 +262,57 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req cns.C
...
+			nodeID, infraVnet, dncEP)
+	}
+
+	nodeRegisterRequest.NmAgentSupportedApis = supportedApis
+	if err := json.NewEncoder(&body).Encode(nodeRegisterRequest); err != nil {
+		log.Errorf("encoding json failed with %v", err)
+		return
+	}
+
+	for sleep := true; err != nil; sleep = true {
```
matmerr, line 294: I know this is just moved code :) but I think we can use ticker to avoid sleeps and returning an error after a certain amount of time
```
ticker := time.NewTicker(time.Second * 5)
for {
  select {
  case <-ticker.C:
  	response, err = httpc.Post(url, "application/json", &body)
		if err == nil {
...
...
  case <-time.After(5 * time.Minute):
  	logger.errorf("no response from nmAgent after an hour") 
        return...
  }
}
```

cns/restserver/api.go
```diff
@@ -1415,3 +1415,45 @@ func (service *HTTPRestService) deleteHostNCApipaEndpoint(w http.ResponseWriter,
...
+	)
+
+	err = service.Listener.Decode(w, r, &req)
+	logger.Request(service.Name, &req, err)
+	if err != nil {
+		return
+	}
+
+	switch r.Method {
+	case "POST":
```
matmerr, line 1438: could use `http.MethodPost` instead of the string

cns/restserver/api.go
```diff
@@ -1415,3 +1415,45 @@ func (service *HTTPRestService) deleteHostNCApipaEndpoint(w http.ResponseWriter,
...
+		supportedApis, returnMessage = nmagentclient.GetNmAgentSupportedApis(
+			req.GetNmAgentSupportedApisURL)
+		if returnMessage != "" {
+			returnCode = NmAgentSupportedApisError
+			logger.Errorf("[Azure-CNS] %s", returnMessage)
+		}
+
+	default:
+		returnMessage = "[Azure-CNS] GetHostLocalIP API expects a GET."
```
matmerr, line 1447: doesn't this API expect a `POST`?

cns/nmagentclient/nmagentclient.go
```diff
@@ -86,3 +96,41 @@ func GetNetworkContainerVersion(
...
+	)
+
+	if getNmAgentSupportedApisURL == "" {
+		getNmAgentSupportedApisURL = fmt.Sprintf(
+			GetNmAgentSupportedApiURLFmt, WireserverIP)
+	}
+
+	response, err := common.GetHttpClient().Get(getNmAgentSupportedApisURL)
+	if err != nil || response.StatusCode != http.StatusOK || response == nil {
+		returnMessage = fmt.Sprintf("Failed to retrieve Supported Apis from NMAgent")
```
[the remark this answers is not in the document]
vakalapa: Agreed!

cns/restserver/api.go
vakalapa: Right, I modeled it as GET first, changed it to POST but forgot to update this message.

cns/nmagentclient/nmagentclient.go
```diff
@@ -86,3 +96,42 @@ func GetNetworkContainerVersion(
 		networkContainerID, response, err)
 	return response, err
 }
+
+// GetNmAgentSupportedApis :- Retrieves Supported Apis from NMAgent
+func GetNmAgentSupportedApis(
+	getNmAgentSupportedApisURL string) ([]string, string) {
```
ramiro-gamarra, line 102: This function should return an error instead of a message. I recommend reading: https://github.com/golang/go/wiki/CodeReviewComments#in-band-errors

cns/nmagentclient/nmagentclient.go
```diff
@@ -86,3 +96,42 @@ func GetNetworkContainerVersion(
...
+	var (
+		returnMessage string
+		supportedApis []string
+	)
```
ramiro-gamarra, line 106: In general, there's no need to declare variables far ahead of their usage in Go. `returnMessage` should really be an error that can be inlined in the returns. You should also consider returning explicit zero values where possible. That makes it easy to understand which return paths are happy path. Example:

```go
return nil, fmt.Errorf("failed to retrieve from nmagent: %v", err)
```
is a concise way to wrap an error as well as return the zero value of a slice in non-happy path. In the same manner
```go
```
also makes it clear that no error is being returned to the calling code.

cns/nmagentclient/nmagentclient.go
```diff
@@ -86,3 +96,42 @@ func GetNetworkContainerVersion(
...
+	if getNmAgentSupportedApisURL == "" {
+		getNmAgentSupportedApisURL = fmt.Sprintf(
+			GetNmAgentSupportedApiURLFmt, WireserverIP)
+	}
+
+	response, err := common.GetHttpClient().Get(getNmAgentSupportedApisURL)
+	if err != nil || response.StatusCode != http.StatusOK || response == nil {
+		returnMessage = fmt.Sprintf(
+			"Failed to retrieve Supported Apis from NMAgent with error %v",
+			err.Error())
```
ramiro-gamarra, line 117: The OR check means the error could still be nil, and this could panic because of nil dereferencing

cns/nmagentclient/nmagentclient.go
```diff
@@ -86,3 +96,42 @@ func GetNetworkContainerVersion(
...
+		logger.Errorf("[Azure-CNS] %s", returnMessage)
```
ramiro-gamarra, line 118: Seems like you can just let the caller code log the errors instead of logging before every return

cns/restserver/internalapi.go
```diff
@@ -254,3 +262,57 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req cns.C
 	return returnCode
 
 }
+
+// Try to register node with DNC when CNS is started in managed DNC mode
+func RegisterNode(httpRestService cns.HTTPService, dncEP, infraVnet, nodeID string) {
```
ramiro-gamarra, line 267: This should return an error

cns/api.go
```diff
@@ -212,3 +213,12 @@ type DeleteHostNCApipaEndpointRequest struct {
 type DeleteHostNCApipaEndpointResponse struct {
 	Response Response
 }
+
+type NmAgentSupportedApisRequest struct {
+	GetNmAgentSupportedApisURL string
```
ramiro-gamarra, line 218: Curious, who will be making this request? Why do they need to pass a URL?

cns/restserver/api.go
```diff
@@ -1415,3 +1415,44 @@ func (service *HTTPRestService) deleteHostNCApipaEndpoint(w http.ResponseWriter,
...
+	)
+
+	err = service.Listener.Decode(w, r, &req)
+	logger.Request(service.Name, &req, err)
+	if err != nil {
+		return
+	}
+
+	switch r.Method {
+	case http.MethodPost:
```
ramiro-gamarra, line 1437: Could this be a GET? After all, the client is getting the supported APIs from CNS, and CNS knows how to communicate to nmagent. Do we need to pass a url in the body?

ramiro-gamarra (commented):
Some comments/questions

cns/restserver/internalapi.go
```diff
@@ -254,3 +262,60 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req cns.C
...
+
+	nodeRegisterRequest.NumCPU = numCPU
+	supportedApis, msg := nmagentclient.GetNmAgentSupportedApis(httpc, "")
+
+	if msg != "" {
+		logger.Printf("[Azure CNS] Failed to retrieve SupportedApis from NMagent of node %s with Infrastructure Network: %s PrivateEndpoint: %s",
+			nodeID, infraVnet, dncEP)
+	}
+
+	nodeRegisterRequest.NmAgentSupportedApis = supportedApis
```
aegal, line 288: Out of curiosity. Why does DNC need this information from nmagent when the node is registered?

cns/restserver/internalapi.go
```diff
@@ -132,6 +135,11 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
 	service.saveState()
 	service.Unlock()
 
+	if nodeInfoResponse.NmAgentApisMissing {
```
aegal, line 138: This doesn't seem like it is needed. CNS would not get past start up if register node did not succeed in the main, so this information would be there if the node was registered with DNC earlier or not?

vakalapa: This will work as kind of a re-try mechanism. If DNC fails to set its properties due to some error, in the next NodeSyncStatus it can solicit for these details again.

cns/restserver/internalapi.go
vakalapa: DNC needs this information to assign GRE key to a NC's primary IP Allocation. 2 conditions are required to be satisfied for DNC to start allocating GRE keys 1. NMAgent needs to have the support, 2. No other NC without greKey allocations to be present in the Node.

cns/restserver/api.go
vakalapa: We had to change from IMDS to Wireserver, what if we want to change this again, albeit long odds.

cns/api.go
vakalapa: DNC will be making this request.

cns/nmagentclient/nmagentclient.go
vakalapa: Agreed, good catch.

cns/nmagentclient/nmagentclient.go
vakalapa: Noted.

cns/NetworkContainerContract.go
```diff
@@ -343,4 +343,11 @@ func (networkContainerRequestPolicy *NetworkContainerRequestPolicies) Validate()
 type NodeInfoResponse struct {
 	NetworkContainers  []CreateNetworkContainerRequest
 	GetNCVersionURLFmt string
+	NmAgentApisMissing bool
```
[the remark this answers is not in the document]
vakalapa: Right now i do not think there are any other requests or actions apart from the one in question. SO kept it simple with a single flag.

cns/restserver/internalapi.go
aegal: The only thing I don't quite understand is if DNC fails to set this properties, then the register node call should fail, so we would never reach here.

vakalapa: Today, register Node happens only once, at the time of initial CNS bringup, main.go L#385. And yes, we do not want to register if DNC sets this flag as false (same as todays behavior)

cns/NetworkContainerContract.go
```diff
@@ -343,4 +343,11 @@ func (networkContainerRequestPolicy *NetworkContainerRequestPolicies) Validate()
...
+}
+
+// NodeRegisterRequest - Struct to hold the node register request.
+type NodeRegisterRequest struct {
+	NumCPU               int
+	NmAgentSupportedApis []string
```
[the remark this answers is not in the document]
vakalapa: ideally first NodeRegister should look like 
NodeRegisterRequest  { NumCPU : 2 , NmAgentSupportedApis : [ "Api1", "Api2"] }

any subsequent Register will be 
NodeRegisterRequest  { NumCPU : 2 , NmAgentSupportedApis : [ ""] }

cns/restserver/internalapi.go
ramiro-gamarra: I don't follow either. If I'm reading main correctly, if node registration failed during start up, that terminates CNS. Registration must succeed before the sync loop is even started. Once it's registered, why do we need to try to register it again here?

cns/restserver/internalapi.go
```diff
@@ -254,3 +265,58 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req cns.C
...
+		body                bytes.Buffer
+		nodeRegisterRequest cns.NodeRegisterRequest
+	)
+
+	nodeRegisterRequest.NumCPU = numCPU
+	supportedApis, retErr := nmagentclient.GetNmAgentSupportedApis(httpc, "")
+
+	if retErr != nil {
+		logger.Printf("[Azure CNS] Failed to retrieve SupportedApis from NMagent of node %s with Infrastructure Network: %s PrivateEndpoint: %s",
+			nodeID, infraVnet, dncEP)
```
ramiro-gamarra, line 287: Bubble error up. I don't think we should attempt to continue if this step failed.

cns/restserver/internalapi.go
ramiro-gamarra: Agree with Mat

ramiro-gamarra (commented):
Few more comments/questions

cns/NetworkContainerContract.go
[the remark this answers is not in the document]
vakalapa: CNS will ask for regular syncNodeStatus, as a response, DNC sends back list of NCs, version and the flag for NMAgentApis list. And DNC will set this flag only if the NodeConfig has GreCapableNotSet property. If the Node is not capable or if the Node has been previously updated with this list then DNC will skip asking for this information.

cns/NetworkContainerContract.go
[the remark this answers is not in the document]
vakalapa: Corrected this behavior, removed the re-register logic. So the requests will be idempotent.

cns/restserver/internalapi.go
vakalapa: After thinking through and testing, i removed the re-register logic. Now if the Node needs to get the ability to add GreKeys, Cx will need to drain NCs, unregister the Node and register again.

cns/restserver/api.go
vakalapa: DNC -> CNS POST message with payload on what CNS needs to send to wireserver. 
CNS-> Wireserver is GET

cns/service/main.go
```diff
@@ -246,43 +248,90 @@ func printVersion() {
...
+			logger.Errorf(err.Error())
+			responseChan <- err
 		}
 
-		if sleep {
-			time.Sleep(acn.FiveSeconds)
-		}
+		response.Body.Close()
+	} else {
+		logger.Errorf("[Azure CNS] Failed to register node with err: %+v", err)
```
ramiro-gamarra, line 333: Is there a reason why this path doesn't send the error? Also, i know this is just adding support of channels, but probably a good opportunity to clean up the function since its kinda hard to follow the code flow. Suggest returning early like:
```go
res, err := post()
if err != nil { /* return */ }

defer res.body.close()

if res.StatusCode != { /* return */ }

if err := decode(req); err != nil { /* return */ }

setNodeOrch(req)
```

ramiro-gamarra (commented):
Few comments

cns/service/main.go
```diff
@@ -246,43 +248,90 @@ func printVersion() {
...
+	supportedApis, retErr := nmagentclient.GetNmAgentSupportedApis(httpc, "")
+
+	if retErr != nil {
+		logger.Errorf("[Azure CNS] Failed to retrieve SupportedApis from NMagent of node %s with Infrastructure Network: %s PrivateEndpoint: %s",
+			nodeID, infraVnet, dncEP)
+		return retErr
+	}
+
+	//To avoid any null-pointer deferencing errors.
+	if supportedApis == nil {
```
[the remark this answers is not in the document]
vakalapa: Okay, I wanted to insure DNC from any null pointer dereferencing.

cns/service/main.go
```diff
@@ -246,43 +248,90 @@ func printVersion() {
...
+	responseChan := make(chan error)
+
+	for {
+		select {
+		case responseErr := <-responseChan:
+			return responseErr
+		case <-nodeRegisterTicker.C:
+			go sendRegisterNodeRequest(httpc, httpRestService, nodeRegisterRequest, url, responseChan)
+		}
+	}
```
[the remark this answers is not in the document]
vakalapa: Will correct the first 5 sec delay. Existing behavior is to retry forever. You are right, we need a better mechanism here. I can add a work item to improve this logic for exponential back off as a separate effort ?

cns/service/main.go
vakalapa: Okay will refactor. Yes, I am assuming this error is retriable error, like timeout or some other transport path error. (which most probably can be due to delays in setting up of azure services). The other errors are specific from DNC side, which i am attributing as not retriable.

ramiro-gamarra (approved):
lgtm!

.pipelines/e2e-step-template.yaml
```diff
@@ -90,6 +90,7 @@ steps:
       export REGIONS=$(AKS_ENGINE_REGION) 
       export IS_JENKINS=false 
       export DEBUG_CRASHING_PODS=true
+      export
```
csfmomo, line 93: Why we need this one?

cns/api.go
csfmomo: Curious if DNC will making this request, why the code is in CNS side? Does that mean CNS will decode the response?

.pipelines/e2e-step-template.yaml
vakalapa: To print out all the env variables. For better debuggability. This will mask any secrets like below:

```
declare -x CLIENT_ID="55990953-d723-433b-a204-01af59561ed8" 
declare -x CLIENT_SECRET="***" 
declare -x CLUSTER_DEFINITION="./cniLinux1604.json"
```

PR #681, opened 2020-10-01
[Windows] Keep the network state cleanup behavior unchanged 

ninzavivek:
A recent change made the operation as no-op.


tamilmani1989 (approved):
lgtm

PR #683, opened 2020-10-01
Throw CNI error - On CNI DEL call, if communication with CNS cannot be established.

ninzavivek:
CNI is some cases is unable to talk to CNS. This can happen, CNS crashing or race between start/CNI getting invoked.

Currently if this communication errrs out on DEL calls, CNI will silently ignore the error.  Container runtime assumes everything went fine and deletes the container. But the HNS endpoints are still there on the host result into them being orphaned. If a container with similar configuration is placed on the host , then ADD call will fail with object already exists error.

Please note, CNI DEL are supposed to be idempotent, and should generally handle the resource not found/exist error elegantely without any error.

### DIFF
--- a/cni/network/multitenancy.go
+++ b/cni/network/multitenancy.go
@@ -211,10 +211,10 @@ func GetMultiTenancyCNIResult(
 	ifName string) (*cniTypesCurr.Result, *cns.GetNetworkContainerResponse, net.IPNet, *cniTypesCurr.Result, error) {
 
 	if nwCfg.MultiTenancy {
-		result, cnsNetworkConfig, subnetPrefix, cnsClienterr := getContainerNetworkConfiguration(nwCfg, k8sPodName, k8sNamespace, ifName)
-		if cnsClienterr != nil {
-			log.Printf("GetContainerNetworkConfiguration failed for podname %v namespace %v with error %+v", k8sPodName, k8sNamespace, cnsClienterr)
-			return nil, nil, net.IPNet{}, nil, cnsClienterr
+		result, cnsNetworkConfig, subnetPrefix, err := getContainerNetworkConfiguration(nwCfg, k8sPodName, k8sNamespace, ifName)
+		if err != nil {
+			log.Printf("GetContainerNetworkConfiguration failed for podname %v namespace %v with error %+v", k8sPodName, k8sNamespace, err)
+			return nil, nil, net.IPNet{}, nil, err
 		}
 
 		log.Printf("PrimaryInterfaceIdentifier :%v", subnetPrefix.IP.String())
@@ -222,7 +222,7 @@ func GetMultiTenancyCNIResult(
 		if checkIfSubnetOverlaps(enableInfraVnet, nwCfg, cnsNetworkConfig) {
 			buf := fmt.Sprintf("InfraVnet %v overlaps with customerVnet %+v", nwCfg.InfraVnetAddressSpace, cnsNetworkConfig.CnetAddressSpace)
 			log.Printf(buf)
-			err := errors.New(buf)
+			err = errors.New(buf)
 			return nil, nil, net.IPNet{}, nil, err
 		}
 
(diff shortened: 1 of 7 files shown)


cni/network/multitenancy.go
```diff
@@ -211,7 +211,7 @@ func GetMultiTenancyCNIResult(
 	ifName string) (*cniTypesCurr.Result, *cns.GetNetworkContainerResponse, net.IPNet, *cniTypesCurr.Result, error) {
 
 	if nwCfg.MultiTenancy {
-		result, cnsNetworkConfig, subnetPrefix, err := getContainerNetworkConfiguration(nwCfg, k8sPodName, k8sNamespace, ifName)
+		result, cnsNetworkConfig, subnetPrefix, _, err := getContainerNetworkConfiguration(nwCfg, k8sPodName, k8sNamespace, ifName)
```
ashvindeodhar, line 214: So here you are basically ignoring the isNotFound flag. Why is that? This function is GetMultiTenancyCNIResult which tries to retrieve the goal state. You are relying only on the err here.
Is the new flag you added really needed? I couldn't understand from the PR as to why exactly we need the new flag and why we cannot use err and cnsNetworkConfig to check if the error occurred during the CNI - CNS call.

cni/network/network.go
```diff
@@ -796,8 +798,16 @@ func (plugin *netPlugin) Delete(args *cniSkel.CmdArgs) error {
...
 
 	// Initialize values from network config.
-	if networkId, err = getNetworkName(k8sPodName, k8sNamespace, args.IfName, nwCfg); err != nil {
+	networkId, isNotFoundErr, err = getNetworkName(k8sPodName, k8sNamespace, args.IfName, nwCfg)
+
+	// If error is not found error, then we ignore it, to comply with CNI SPEC.
+	if err != nil {
 		log.Printf("[cni-net] Failed to extract network name from network config. error: %v", err)
+
+		if !isNotFoundErr {
```
ashvindeodhar, line 807: if error is isNotFoundErr, seems like we don't return err but proceed ahead? What's the networkId we use in that case?

ashvindeodhar:
Ideally we shouldn't need to go to CNS to get the network name for the endpoint that is already created. Regardless of the CNS state we should be able to delete the endpoint in CNI.

ramiro-gamarra (commented):
couple comments

cns/cnsclient/cnsclient.go
```diff
@@ -65,34 +82,37 @@ func (cnsClient *CNSClient) GetNetworkConfiguration(orchestratorContext []byte)
 	err := json.NewEncoder(&body).Encode(payload)
 	if err != nil {
 		log.Errorf("encoding json failed with %v", err)
-		return nil, err
+		return nil, &CNSClientError{restserver.UnexpectedError, err}
```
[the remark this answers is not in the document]
ninzavivek: I think as we discussed, let's make the cnsclient always return cnsclient error.

ashvindeodhar:
looks good to me but lets wait for the pipeline issue to get resolved so that this gets validated with e2e tests

cni/utils/errutils.go
```diff
@@ -0,0 +1,20 @@
...
+
+// TODO : Move to common directory like common, after fixing circular dependencies
+func IsNotFoundError(err error) bool {
+	switch err := err.(type) {
+	case *cnsclient.CNSClientError:
+		var cnsError *cnsclient.CNSClientError
+		// not expected to fail
+		_ = errors.As(err, &cnsError)
+		return (cnsError.Code == restserver.UnknownContainerID)
+	}
```
ramiro-gamarra, line 18: When you do a type switch and assign to a variable, you can use that variable as the concrete type directly within each case block. If you don't want to shadow `err` to make this more evident, you can do something like
```go
switch e := err.(type) {
case *cnsclient.CNSClientError:
    return e.Code == restserver.UnknownContainerID
}
```

ramiro-gamarra (approved):

PR #687, opened 2020-10-07
Add NC version in nnc CRD. Populate NC version info from nnc CRD to N…

csfmomo:
feat: add NC version in nnc CRD and populate it to NC request.

**Reason for Change**:
It's a prerequisite for IP classification feature. In IP classification feature, CNS will validate IPs through comparing NC version between its local file, nnc CRD and NMAgent to determine whether its available or still pending programming.

### DIFF
--- a/cns/requestcontroller/kubecontroller/crdtranslator.go
+++ b/cns/requestcontroller/kubecontroller/crdtranslator.go
@@ -36,6 +36,7 @@ func CRDStatusToNCRequest(crdStatus nnc.NodeNetworkConfigStatus) (cns.CreateNetw
 		ncRequest.SecondaryIPConfigs = make(map[string]cns.SecondaryIPConfig)
 		ncRequest.NetworkContainerid = nc.ID
 		ncRequest.NetworkContainerType = cns.Docker
+		ncRequest.Version = nc.Version
 
 		if ip = net.ParseIP(nc.PrimaryIP); ip == nil {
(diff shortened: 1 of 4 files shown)


cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -36,6 +36,7 @@ func CRDStatusToNCRequest(crdStatus nnc.NodeNetworkConfigStatus) (cns.CreateNetw
 		ncRequest.SecondaryIPConfigs = make(map[string]cns.SecondaryIPConfig)
 		ncRequest.NetworkContainerid = nc.ID
 		ncRequest.NetworkContainerType = cns.Docker
+		ncRequest.Version = nc.Version
```
neaggarwMS, line 39: Does the CRD already has NC version field? Interesting

neaggarwMS: Is it already populated by DNC?

neaggarwMS (dismissed):

cns/requestcontroller/kubecontroller/crdtranslator.go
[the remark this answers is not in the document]
thatmattlong: This is a good use-case for kubernetes api versioning. We should increment the version of this Kind. See https://kubebuilder.io/cronjob-tutorial/gvks.html for a description

neaggarwMS (approved):

PR #695, opened 2020-10-19
fix: Pass host gateway to CNI in swift mode

matmerr:
**Reason for Change**:

In swift mode the route for host to pod on other host needs to go via the default gateway, so we add a route to the bridge for the pod subnet to use the host default gateway

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/bridge_networkclient_linux.go
+++ b/network/bridge_networkclient_linux.go
@@ -50,17 +50,30 @@ func (client *LinuxBridgeClient) CreateBridge() error {
 
 func (client *LinuxBridgeClient) AddRoutes(nwInfo *NetworkInfo, interfaceName string) error {
 	if client.nwInfo.IPAMType == AzureCNS {
+
+		// Add snat Rules
+		gwIP := client.nwInfo.Options[SNATIPKey]
+		if gwIP == nil {
+			return fmt.Errorf("Host gateway IP in Options not set")
+		}
+
+		gatewayIP := net.ParseIP(gwIP.(string))
+		if gatewayIP == nil {
+			return fmt.Errorf("Invalid host gateway IP: %+v", gwIP)
+		}
 		// add pod subnet to host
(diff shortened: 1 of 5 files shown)


network/bridge_networkclient_linux.go
```diff
@@ -50,18 +50,31 @@ func (client *LinuxBridgeClient) CreateBridge() error {
 
 func (client *LinuxBridgeClient) AddRoutes(nwInfo *NetworkInfo, interfaceName string) error {
 	if client.nwInfo.IPAMType == AzureCNS {
+
+		// Add snat Rules
+		gwIP := client.nwInfo.Options[SNATIPKey]
```
vakalapa, line 55: HostGWKey  ? instead of SNATIPKey

network/bridge_networkclient_linux.go
```diff
@@ -50,18 +50,31 @@ func (client *LinuxBridgeClient) CreateBridge() error {
...
 		// add pod subnet to host
 		devIf, _ := net.InterfaceByName(interfaceName)
 		ifIndex := devIf.Index
-		family := netlink.GetIpAddressFamily(Ipv4DefaultRouteDstPrefix.IP)
+		family := netlink.GetIpAddressFamily(gatewayIP)
 
 		nlRoute := &netlink.Route{
 			Family:    family,
 			Dst:       &client.nwInfo.PodSubnet.Prefix,
-			Gw:        Ipv4DefaultRouteDstPrefix.IP,
```
paulgmiller, line 61: Seems like instead of adding a rule we're replacing a rule. Also what is the podsubnet here is it actually the node subnet (what I would think of in aks-rp as vnetsubnetid given to aks?)

neaggarwMS (changes requested):

neaggarwMS (approved):

neaggarwMS (approved):

PR #696, opened 2020-10-19
fix: Add missing CNI invoker delete calls

matmerr:
**Reason for Change**:

All add calls and delete calls should be via the invoker interface, this fix is to change DelegateDel's to invoker.Delete, which is required in swift mode

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


neaggarwMS (dismissed):

matmerr:
> Do we need to add nil check for nwCfg?

there aren't any checks in the existing path for a nil nwCfg

neaggarwMS (approved):

PR #697, opened 2020-10-21
Add nc version in SecondaryIPConfig structure.

csfmomo:
feat: Add NC version in SecondaryIPConfig structure.

**Reason for Change**:
When CNS reconcile, it needs NC version to determine whether an IP should be stay in pending programming or available.

- [X ] adds unit tests

### DIFF
--- a/cns/requestcontroller/kubecontroller/crdtranslator.go
+++ b/cns/requestcontroller/kubecontroller/crdtranslator.go
@@ -57,12 +57,16 @@ func CRDStatusToNCRequest(crdStatus nnc.NodeNetworkConfigStatus) (cns.CreateNetw
 			if ip = net.ParseIP(ipAssignment.IP); ip == nil {
 				return ncRequest, fmt.Errorf("Invalid SecondaryIP %s:", ipAssignment.IP)
 			}
-			ncVersion, _ := strconv.Atoi(ncRequest.Version)
-			secondaryIPConfig = cns.SecondaryIPConfig{
-				IPAddress: ip.String(),
-				NCVersion: ncVersion,
+
+			ipConfig, ok := ncRequest.SecondaryIPConfigs[ipAssignment.Name]
+			if !ok || (ok && ipConfig.IPAddress != ip.String()) {
+				ncVersion, _ := strconv.Atoi(ncRequest.Version)
+				secondaryIPConfig = cns.SecondaryIPConfig{
+					IPAddress: ip.String(),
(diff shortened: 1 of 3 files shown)


cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -57,10 +58,15 @@ func CRDStatusToNCRequest(crdStatus nnc.NodeNetworkConfigStatus) (cns.CreateNetw
 				return ncRequest, fmt.Errorf("Invalid SecondaryIP %s:", ipAssignment.IP)
 			}
 
-			secondaryIPConfig = cns.SecondaryIPConfig{
-				IPAddress: ip.String(),
+			ipConfig, ok := ncRequest.SecondaryIPConfigs[ipAssignment.Name]
+			if !ok || (ok && ipConfig.IPAddress != ip.String()) {
+				ncVersion, _ := strconv.Atoi(ncRequest.Version)
```
neaggarwMS, line 63: We dont need this converstion to happen inside a loop

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -57,10 +58,15 @@ func CRDStatusToNCRequest(crdStatus nnc.NodeNetworkConfigStatus) (cns.CreateNetw
...
+				secondaryIPConfig = cns.SecondaryIPConfig{
+					IPAddress: ip.String(),
+					NCVersion: ncVersion,
```
neaggarwMS, line 66: Is this logic expected? For the existing SecondaryIPs, are we updating the NCversion on CRD update? I thought the whole purpose of adding a version to SecondaryIP was to preserve the version when secondaryip was added to this NC

cns/requestcontroller/kubecontroller/crdtranslator.go
csfmomo: We need this logic. It's the same question as your second comment. Please see explanation there.

cns/requestcontroller/kubecontroller/crdtranslator.go
csfmomo: This logic is expected. We are updating the NCversion on CRD update. CRDStatusToNCRequest is the function will be called when CRD update.
CNS don't want to renew existing IPs in secondary IP.

csfmomo:
Open a new PR with master merged in #701

PR #701, opened 2020-10-23
Add nc version in SecondaryIPConfig stucture.

csfmomo:
feat: Add NC version in SecondaryIPConfig structure.

**Reason for Change**:
When CNS reconcile, it needs NC version to determine whether an IP should be stay in pending programming or available.

- [X ] adds unit tests

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -131,6 +131,7 @@ type IPConfiguration struct {
 // SecondaryIPConfig contains IP info of SecondaryIP
 type SecondaryIPConfig struct {
 	IPAddress string
+	NCVersion int
 }
 
 // IPSubnet contains ip subnet.
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -131,6 +131,7 @@ type IPConfiguration struct {
 // SecondaryIPConfig contains IP info of SecondaryIP
 type SecondaryIPConfig struct {
 	IPAddress string
+	// NCVesion will help in determining whether IP is in pending programming or available when reconciling.
 	NCVersion int
 }
 
--- a/cns/restserver/internalapi_test.go
+++ b/cns/restserver/internalapi_test.go
@@ -61,8 +61,8 @@ func TestCreateAndUpdateNCWithSecondaryIPNCVersion(t *testing.T) {
 	// Build secondaryIPConfig, it will have one item as {IPAddress:"10.0.0.16", NCVersion: 0}
 	ipAddress := "10.0.0.16"
 	secIPConfig := newSecondaryIPConfig(ipAddress, ncVersion)
-	ipID := uuid.New()
-	secondaryIPConfigs[ipID.String()] = secIPConfig
+	ipId := uuid.New()
+	secondaryIPConfigs[ipId.String()] = secIPConfig
 	req := createNCReqInternal(t, secondaryIPConfigs, ncID, strconv.Itoa(ncVersion))
 	validateSecondaryIPsNCVersion(t, req)
 
@@ -72,13 +72,13 @@ func TestCreateAndUpdateNCWithSecondaryIPNCVersion(t *testing.T) {
 	ncVersion++
 	secIPConfig = newSecondaryIPConfig(ipAddress, ncVersion)
 	// "10.0.0.16" will be update to NC version 1 in CRD, reuse the same uuid with it.
-	secondaryIPConfigs[ipID.String()] = secIPConfig
+	secondaryIPConfigs[ipId.String()] = secIPConfig
 
 	// Add {IPAddress:"10.0.0.17", NCVersion: 1} in secondaryIPConfig
 	ipAddress = "10.0.0.17"
 	secIPConfig = newSecondaryIPConfig(ipAddress, ncVersion)
-	ipID = uuid.New()
-	secondaryIPConfigs[ipID.String()] = secIPConfig
+	ipId = uuid.New()
+	secondaryIPConfigs[ipId.String()] = secIPConfig
 	req = createNCReqInternal(t, secondaryIPConfigs, ncID, strconv.Itoa(ncVersion))
 	validateSecondaryIPsNCVersion(t, req)
 }
--- a/cns/restserver/internalapi_test.go
+++ b/cns/restserver/internalapi_test.go
@@ -70,7 +70,7 @@ func TestCreateAndUpdateNCWithSecondaryIPNCVersion(t *testing.T) {
 	receivedSecondaryIPConfigs = containerStatus.CreateNetworkContainerRequest.SecondaryIPConfigs
 	for _, secIPConfig := range receivedSecondaryIPConfigs {
 		// Though "10.0.0.16" IP exists in NC version 1, secodanry IP still keep its original NC version 0
-		if secIPConfig.IPAddress != "10.0.0.16" && secIPConfig.NCVersion != 0 {
+		if secIPConfig.IPAddress != "10.0.0.16" || secIPConfig.NCVersion != 0 {
 			t.Fatalf("nc request version is %d, secondary ip %s nc version is %d, expected nc version is 0",
 				ncVersion, secIPConfig.IPAddress, secIPConfig.NCVersion)
 		}
@@ -90,7 +90,8 @@ func TestCreateAndUpdateNCWithSecondaryIPNCVersion(t *testing.T) {
 	ipId = uuid.New()
 	secondaryIPConfigs[ipId.String()] = secIPConfig
 	req = createNCReqInternal(t, secondaryIPConfigs, ncID, strconv.Itoa(ncVersion))
-
+	// Validate secondary IPs' NC version has been updated by NC request
+	receivedSecondaryIPConfigs = containerStatus.CreateNetworkContainerRequest.SecondaryIPConfigs
 	for _, secIPConfig := range receivedSecondaryIPConfigs {
 		// Though "10.0.0.16" IP exists in NC version 1, secodanry IP still keep its original NC version 0
 		if (secIPConfig.IPAddress == "10.0.0.16" && secIPConfig.NCVersion != 0) ||
--- a/cns/restserver/internalapi_test.go
+++ b/cns/restserver/internalapi_test.go
@@ -424,20 +424,6 @@ func createNCReqInternal(t *testing.T, secondaryIPConfigs map[string]cns.Seconda
 	return req
 }
 
-func validateSecondaryIPsNCVersion(t *testing.T, req cns.CreateNetworkContainerRequest) {
-	containerStatus := svc.state.ContainerStatus[req.NetworkContainerid]
-	// Validate secondary IPs' NC version has been updated by NC request
-	ncVersion, _ := strconv.Atoi(containerStatus.VMVersion)
-	for _, secIPConfig := range containerStatus.CreateNetworkContainerRequest.SecondaryIPConfigs {
-		// Though "10.0.0.16" IP exists in NC version 1, secodanry IP still keep its original NC version 0
-		if (secIPConfig.IPAddress == "10.0.0.16" && secIPConfig.NCVersion != 0) ||
-			(secIPConfig.IPAddress == "10.0.0.17" && secIPConfig.NCVersion != 1) {
-			t.Fatalf("nc request version is %d, secondary ip %s nc version is %d, expected nc version is 0",
-				ncVersion, secIPConfig.IPAddress, secIPConfig.NCVersion)
-		}
-	}
-}
-
 func restartService() {
 	fmt.Println("Restart Service")
 
--- a/cns/restserver/internalapi_test.go
+++ b/cns/restserver/internalapi_test.go
@@ -50,13 +50,11 @@ func TestCreateOrUpdateNCWithLargerVersionComparedToNMAgent(t *testing.T) {
 
 func TestCreateAndUpdateNCWithSecondaryIPNCVersion(t *testing.T) {
 	restartService()
-
 	setEnv(t)
 	setOrchestratorTypeInternal(cns.KubernetesCRD)
 	// NC version set as 0 which is the default initial value.
 	ncVersion := 0
 	secondaryIPConfigs := make(map[string]cns.SecondaryIPConfig)
-	receivedSecondaryIPConfigs := make(map[string]cns.SecondaryIPConfig)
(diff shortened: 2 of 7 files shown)


cns/restserver/util.go
```diff
@@ -248,23 +248,29 @@ func (service *HTTPRestService) updateIpConfigsStateUntransacted(req cns.CreateN
...
 // acquire/release the service lock.
-func (service *HTTPRestService) addIPConfigStateUntransacted(newIPCNSStatus, ncId string, ipconfigs map[string]cns.SecondaryIPConfig) {
+func (service *HTTPRestService) addIPConfigStateUntransacted(newIPCNSStatus, ncId string, ipconfigs, existingSecondaryIPConfigs map[string]cns.SecondaryIPConfig) {
 	// add ipconfigs to state
-	for ipId, ipconfig := range ipconfigs {
-		if _, exists := service.PodIPConfigState[ipId]; exists {
+	for ipID, ipconfig := range ipconfigs {
+		// New secondary IP configs has new NC version however, CNS don't want to override existing IPs'with new NC version
+		// Set it back to previous NC version if IP already exist.
+		if existingIPConfig, existsInPreviousIPConfig := existingSecondaryIPConfigs[ipID]; existsInPreviousIPConfig && existingIPConfig.IPAddress == ipconfig.IPAddress {
```
neaggarwMS, line 256: Instead why even pass the NCVersion. Thus, during saving the state here, only add NCVersion (which came in the Request.Version) to the new secondary IPs added

neaggarwMS:
There is no change in the reconcile funntion? can you also cover it here? That will complete the NC version check for all scenarios .

cns/NetworkContainerContract.go
```diff
@@ -131,6 +131,7 @@ type IPConfiguration struct {
 // SecondaryIPConfig contains IP info of SecondaryIP
 type SecondaryIPConfig struct {
 	IPAddress string
+	NCVersion int
```
neaggarwMS, line 134: Can you add a comment why you are persisting NCVersion with SecondaryIP

neaggarwMS (changes requested):

cns/restserver/util.go
```diff
@@ -248,23 +248,29 @@ func (service *HTTPRestService) updateIpConfigsStateUntransacted(req cns.CreateN
 // addIPConfigStateUntransacted adds the IPConfigs to the PodIpConfigState map with Available state
 // If the IP is already added then it will be an idempotent call. Also note, caller will
 // acquire/release the service lock.
-func (service *HTTPRestService) addIPConfigStateUntransacted(newIPCNSStatus, ncId string, ipconfigs map[string]cns.SecondaryIPConfig) {
+func (service *HTTPRestService) addIPConfigStateUntransacted(newIPCNSStatus, ncId string, ipconfigs, existingSecondaryIPConfigs map[string]cns.SecondaryIPConfig) {
 	// add ipconfigs to state
-	for ipId, ipconfig := range ipconfigs {
-		if _, exists := service.PodIPConfigState[ipId]; exists {
+	for ipID, ipconfig := range ipconfigs {
```
neaggarwMS, line 253: can we rename it to ipId (small d) only?

cns/restserver/util.go
```diff
@@ -248,23 +248,29 @@ func (service *HTTPRestService) updateIpConfigsStateUntransacted(req cns.CreateN
...
+		// New secondary IP configs has new NC version however, CNS don't want to override existing IPs'with new NC version
+		// Set it back to previous NC version if IP already exist.
+		if existingIPConfig, existsInPreviousIPConfig := existingSecondaryIPConfigs[ipID]; existsInPreviousIPConfig && existingIPConfig.IPAddress == ipconfig.IPAddress {
+			ipconfig.NCVersion = existingIPConfig.NCVersion
```
neaggarwMS, line 257: Can you add a log statement here where you are updating the Version back to old one

cns/NetworkContainerContract.go
csfmomo: Sure.

cns/restserver/util.go
csfmomo: Sure.

neaggarwMS: actually lets add a log line for new ips added as well, specially the state and ncversion info

neaggarwMS (changes requested):

cns/restserver/util.go
```diff
@@ -248,23 +248,29 @@ func (service *HTTPRestService) updateIpConfigsStateUntransacted(req cns.CreateN
...
+			ipconfigs[ipID] = ipconfig
+		}
+		if _, exists := service.PodIPConfigState[ipID]; exists {
 			continue
 		}
 		// add the new State
 		ipconfigStatus := cns.IPConfigurationStatus{
 			NCID:                ncId,
-			ID:                  ipId,
+			ID:                  ipID,
```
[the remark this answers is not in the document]
csfmomo: If we want to keep consistency with previous version, I can rename it back to ipId. It's go auto format tool, gofmt suggests me to use ID.

cns/restserver/util.go
csfmomo: Added in line 260.

csfmomo:
> There is no change in the reconcile funntion? can you also cover it here? That will complete the NC version check for all scenarios .

Discussed offline. We'll use a separate PR to address this scenario.

cns/restserver/util.go
csfmomo: Updated.

cns/restserver/internalapi_test.go
```diff
@@ -359,6 +397,29 @@ func validateNCStateAfterReconcile(t *testing.T, ncRequest *cns.CreateNetworkCon
...
+func validateSecondaryIPsNCVersion(t *testing.T, req cns.CreateNetworkContainerRequest) {
+	containerStatus := svc.state.ContainerStatus[req.NetworkContainerid]
+	// Validate secondary IPs' NC version has been updated by NC request
+	ncVersion, _ := strconv.Atoi(containerStatus.VMVersion)
+	for _, secIPConfig := range containerStatus.CreateNetworkContainerRequest.SecondaryIPConfigs {
+		// Though "10.0.0.16" IP exists in NC version 1, secodanry IP still keep its original NC version 0
+		if (secIPConfig.IPAddress == "10.0.0.16" && secIPConfig.NCVersion != 0) ||
+			(secIPConfig.IPAddress == "10.0.0.17" && secIPConfig.NCVersion != 1) {
+			t.Fatalf("nc request version is %d, secondary ip %s nc version is %d, expected nc version is 0",
+				ncVersion, secIPConfig.IPAddress, secIPConfig.NCVersion)
```
ramiro-gamarra, line 418: These assertion values (the expected ip addresses and versions) are very far from the test. If you need to create similar tests with different IP addresses, this would not be very reusable. You should define the expected values on the test, and pass them to this function. The check is also very narrow: if there was only a secondary IP of 10.0.0.15 for example, this would pass.

cns/restserver/util.go
```diff
@@ -248,9 +248,16 @@ func (service *HTTPRestService) updateIpConfigsStateUntransacted(req cns.CreateN
 // addIPConfigStateUntransacted adds the IPConfigs to the PodIpConfigState map with Available state
 // If the IP is already added then it will be an idempotent call. Also note, caller will
 // acquire/release the service lock.
-func (service *HTTPRestService) addIPConfigStateUntransacted(newIPCNSStatus, ncId string, ipconfigs map[string]cns.SecondaryIPConfig) {
+func (service *HTTPRestService) addIPConfigStateUntransacted(newIPCNSStatus, ncId string, ipconfigs, existingSecondaryIPConfigs map[string]cns.SecondaryIPConfig) {
 	// add ipconfigs to state
 	for ipId, ipconfig := range ipconfigs {
+		// New secondary IP configs has new NC version however, CNS don't want to override existing IPs'with new NC version
+		// Set it back to previous NC version if IP already exist.
+		if existingIPConfig, existsInPreviousIPConfig := existingSecondaryIPConfigs[ipId]; existsInPreviousIPConfig && existingIPConfig.IPAddress == ipconfig.IPAddress {
```
ramiro-gamarra, line 256: Is the check on ip address equality necessary?

ramiro-gamarra (commented):
Some comments

cns/restserver/internalapi_test.go
```diff
@@ -359,6 +397,29 @@ func validateNCStateAfterReconcile(t *testing.T, ncRequest *cns.CreateNetworkCon
...
+	req := generateNetworkContainerRequest(secondaryIPConfigs, ncID, ncVersion)
+	returnCode := svc.CreateOrUpdateNetworkContainerInternal(req, fakes.NewFakeScalar(releasePercent, requestPercent, batchSize), fakes.NewFakeNodeNetworkConfigSpec(initPoolSize))
+	if returnCode != 0 {
+		t.Fatalf("Failed to createNetworkContainerRequest, req: %+v, err: %d", req, returnCode)
+	}
+	return req
+}
+
+func validateSecondaryIPsNCVersion(t *testing.T, req cns.CreateNetworkContainerRequest) {
+	containerStatus := svc.state.ContainerStatus[req.NetworkContainerid]
```
neaggarwMS, line 410: This function seems generic but inside it has hardcoded the ipaddress and expected version. Can we pass the expected SecondaryIpConfig as parameter and compare the req with expected list? This will allow to reuse this function in future as well

neaggarwMS (dismissed):

cns/restserver/internalapi_test.go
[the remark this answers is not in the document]
csfmomo: - This is a validation code in test file. I don't think mutex guarded is necessary. No others have access to svc.state.
- Not necessary. If it doesn't exist, test will fail.

cns/restserver/internalapi_test.go
csfmomo: No. This test is designed specifically and is validating specific scenario.
In this specific test, secondary IP address only contains 10.0.0.16 and 10.0.0.17.
10.0.0.16 must be version 0 and 10.0.0.17 must be version 1.

cns/restserver/util.go
csfmomo: I think it's necessary. If uuid is the same, IPAddress must be same. If not the same, then sth wrong. At least we don't want to set the NC version.

cns/restserver/internalapi_test.go
csfmomo: We can but I don't think it's necessary. I don't think any other test should reuse this function. This validation is designed for TestCreateAndUpdateNCWithSecondaryIPNCVersion specifically.

csfmomo: I moved this validation to test function and avoid confusion.

cns/restserver/internalapi_test.go
```diff
@@ -48,6 +48,60 @@ func TestCreateOrUpdateNCWithLargerVersionComparedToNMAgent(t *testing.T) {
...
+
+	// Build secondaryIPConfig, it will have one item as {IPAddress:"10.0.0.16", NCVersion: 0}
+	ipAddress := "10.0.0.16"
+	secIPConfig := newSecondaryIPConfig(ipAddress, ncVersion)
+	ipId := uuid.New()
+	secondaryIPConfigs[ipId.String()] = secIPConfig
+	req := createNCReqInternal(t, secondaryIPConfigs, ncID, strconv.Itoa(ncVersion))
+	containerStatus := svc.state.ContainerStatus[req.NetworkContainerid]
+	// Validate secondary IPs' NC version has been updated by NC request
+	receivedSecondaryIPConfigs = containerStatus.CreateNetworkContainerRequest.SecondaryIPConfigs
```
ramiro-gamarra, line 70: - no need to make a map in line 59, just use `:=` here
- the reason why a check for existence of entry in the map is better is that the test won't fail if there is no match. If you don't get the check for existence, you get the zero value of a struct, including all of it's members recursively, and ranging over a nil map is valid. In case the nc id was not found in the map, the checks below will not happen, and the test will continue as normal. Let me know if that makes sense.

cns/restserver/internalapi_test.go
```diff
@@ -48,6 +48,60 @@ func TestCreateOrUpdateNCWithLargerVersionComparedToNMAgent(t *testing.T) {
...
+	secondaryIPConfigs[ipId.String()] = secIPConfig
+
+	// Add {IPAddress:"10.0.0.17", NCVersion: 1} in secondaryIPConfig
+	ipAddress = "10.0.0.17"
+	secIPConfig = newSecondaryIPConfig(ipAddress, ncVersion)
+	ipId = uuid.New()
+	secondaryIPConfigs[ipId.String()] = secIPConfig
+	req = createNCReqInternal(t, secondaryIPConfigs, ncID, strconv.Itoa(ncVersion))
+	// Validate secondary IPs' NC version has been updated by NC request
+	receivedSecondaryIPConfigs = containerStatus.CreateNetworkContainerRequest.SecondaryIPConfigs
```
ramiro-gamarra, line 94: you need to read from the map again, otherwise the container status goes unchanged in this scope

cns/restserver/internalapi_test.go
```diff
@@ -48,6 +48,60 @@ func TestCreateOrUpdateNCWithLargerVersionComparedToNMAgent(t *testing.T) {
...
+	for _, secIPConfig := range receivedSecondaryIPConfigs {
+		// Though "10.0.0.16" IP exists in NC version 1, secodanry IP still keep its original NC version 0
+		if (secIPConfig.IPAddress == "10.0.0.16" && secIPConfig.NCVersion != 0) ||
+			(secIPConfig.IPAddress == "10.0.0.17" && secIPConfig.NCVersion != 1) {
+			t.Fatalf("nc request version is %d, secondary ip %s nc version is %d, expected nc version is 0",
```
ramiro-gamarra, line 99: prefer a switch statement here, and messages more specific to each. in case the inconsistent version was with 10.0.0.17, the message still tells you that it expected 0, which is not appropriate:
```go
switch secIPConfig:
case "10.0.0.16":
    // check for 0
case "10.0.0.17":
    // check for 1
default:
    // error since this is unexpected

```

ramiro-gamarra (commented):
Few more comments

cns/restserver/internalapi_test.go
csfmomo: I think what you are trying to say is adding one more check in map length. I can do that though it's not the key point in this test.

ramiro-gamarra (approved):

PR #705, opened 2020-10-27
Support for ACL (Hnsv2)

ninzavivek:
Support for ACL (Windows - Hnsv2). This is needed so that access to wireserver can be blocked from inside the container. 

Also added an ACL rule for blocking the access to wireserver by default. 

Testing : Singetenancy - containerd.


ninzavivek:
@ashvindeodhar / @pjohnst5 - Let me if it is ok to put these in the default conf file [single tenant] or not, I think it is a good idea.

network/policy/policy_windows.go
```diff
@@ -343,6 +347,27 @@ func GetHcnPortMappingPolicy(policy Policy) (hcn.EndpointPolicy, error) {
...
+	aclEndpolicySetting := hcn.EndpointPolicy{
+		Type: hcn.ACL,
+	}
+
+	var aclPolicySetting hcn.AclPolicySetting
+	if err := json.Unmarshal(policy.Data, &aclPolicySetting); err != nil {
+		return aclEndpolicySetting, err
+	}
+
+	aclPolicySettingBytes, err := json.Marshal(aclPolicySetting)
```
thatmattlong, line 361: Why marshal into bytes when we already have the bytes in `policy.Data`?

cni/azure-windows.conflist
```diff
@@ -40,6 +40,46 @@
                         "DestinationPrefix": "10.0.0.0/8",
                         "NeedEncap": true
                     }
+                },
+                {
```
ashvindeodhar, line 44: is there any reason you chose to add these ACLs by default in the CNI conflist?

cni/azure-windows.conflist
```diff
@@ -40,6 +40,46 @@
...
+                        "Action":  "Allow",
+                        "Direction":  "Out",
+                        "RemoteAddresses":  "168.63.129.16/32",
+                        "Protocols": "17",
+                        "RemotePorts":  "53",
+                        "Priority":  200
+                    }
+                },
+                {
+                    "Name":  "EndpointPolicy",
```
ashvindeodhar, line 57: Remove 2 spaces after : 
We have one space for all the other settings

network/policy/policy_windows.go
```diff
@@ -176,6 +176,10 @@ func GetPolicyType(policy Policy) CNIPolicyType {
 		}
 	}
 
+	if dataRoute.Type == ACLPolicy {
```
ashvindeodhar, line 179: why is this 'dataRoute' ?

network/policy/policy_windows.go
ashvindeodhar: +1. Here you are unmarshalling and marshalling back. For the other policies such as route, we were unmarshalling, extracting the needed settings, add those in hcn policy settings and then marshalling to []byte.

ashvindeodhar (changes requested):

network/policy/policy_windows.go
ninzavivek: This is to make sure, that it meets ACPolicy construct, rather waiting until the endpoint creation creates.

ninzavivek: Let me know if you wait to until the endpoint creation.

network/policy/policy_windows.go
ninzavivek: dataRoute?

cni/azure-windows.conflist
ninzavivek: This is Azure CNI - The ACL is needed because of the security incident that needs to block access to wireserver API ( Which exist in Azure). My line of thinking is that considering it is security related, every container should get this default rather the partner unknowingly enabling it and creating a security hole for them.

network/policy/policy_windows.go
thatmattlong: I think it's fine, but please add a comment to explain so that later someone doesn't come through cleaning this up and accidentally removes the intended behavior

network/policy/policy_windows.go
ashvindeodhar: if dataRoute.Type == ACLPolicy

network/policy/policy_windows.go
ninzavivek: done.

network/policy/policy_windows.go
ninzavivek: done.

cni/azure-windows.conflist
ninzavivek: done.

ashvindeodhar:
@ninzavivek are done with docker testing? The comment in the description says - 
Testing : Singetenancy - containerd.
docker pending

I think it should be ok to add those ACLs by default. I have resolved the review comment.
The tests are failing though.

ashvindeodhar (approved):

PR #71, opened 2017-07-17
Added ipam support for overlay network

tamilmani1989:
Implemented Ipam driver for overlay network. The pools will be created on demand when a request comes for specific subnet. As of now, ipam driver for azure-vnet and overlay cannot be started concurrently. Instead of starting azure source, the overlay ipam driver starts null source to avoid communication with NM Agent.

PR #714, opened 2020-10-30
Add a go routine to update NC host version from NMAgent periodically.

csfmomo:
feat: 
- Add a go routine to update NC host version from NMAgent periodically.
- If orchestrator type is CRD, update pending programming IPs as well.

**Reason for Change**:
NC version from DNC can be different with NC version from NMAgent when NMAgent didn't keep up the NC version change from DNC side. 
This feature enable CNS sync NC version from NMAgent periodically and update related info.
fix: Bug Fixes 🐞
test: Testing 💚 
- [X ] adds unit tests

### DIFF
--- a/cns/api.go
+++ b/cns/api.go
@@ -38,6 +38,7 @@ type HTTPService interface {
 	SendNCSnapShotPeriodically(int, chan bool)
 	SetNodeOrchestrator(*SetOrchestratorTypeRequest)
 	SyncNodeStatus(string, string, string, json.RawMessage) (int, string)
+	SyncHostNCVersion(string)
 	GetPendingProgramIPConfigs() []IPConfigurationStatus
 	GetAvailableIPConfigs() []IPConfigurationStatus
 	GetAllocatedIPConfigs() []IPConfigurationStatus
--- a/cns/api.go
+++ b/cns/api.go
@@ -38,7 +38,6 @@ type HTTPService interface {
 	SendNCSnapShotPeriodically(int, chan bool)
 	SetNodeOrchestrator(*SetOrchestratorTypeRequest)
 	SyncNodeStatus(string, string, string, json.RawMessage) (int, string)
-	SyncHostNCVersion(string)
 	GetPendingProgramIPConfigs() []IPConfigurationStatus
 	GetAvailableIPConfigs() []IPConfigurationStatus
 	GetAllocatedIPConfigs() []IPConfigurationStatus
--- a/cns/cnsclient/cnsclient_test.go
+++ b/cns/cnsclient/cnsclient_test.go
@@ -63,6 +63,9 @@ func addTestStateToRestServer(t *testing.T, secondaryIps []string) {
 		NetworkContainerid:   "testNcId1",
 		IPConfiguration:      ipConfig,
 		SecondaryIPConfigs:   secondaryIPConfigs,
+		// Set it as -1 to be same as default host version.
+		// It will allow secondary IPs status to be set as available.
+		Version: "-1",
 	}
 
 	returnCode := svc.CreateOrUpdateNetworkContainerInternal(req, fakes.NewFakeScalar(releasePercent, requestPercent, batchSize), fakes.NewFakeNodeNetworkConfigSpec(initPoolSize))
--- a/cns/cnsclient/cnsclient_test.go
+++ b/cns/cnsclient/cnsclient_test.go
@@ -53,6 +53,7 @@ func addTestStateToRestServer(t *testing.T, secondaryIps []string) {
 	for _, secIpAddress := range secondaryIps {
 		secIpConfig := cns.SecondaryIPConfig{
 			IPAddress: secIpAddress,
+			NCVersion: -1,
 		}
 		ipId := uuid.New()
 		secondaryIPConfigs[ipId.String()] = secIpConfig
@@ -125,7 +126,7 @@ func TestMain(m *testing.M) {
 	logger.InitLogger(logName, 0, 0, tmpLogDir+"/")
 	config := common.ServiceConfig{}
 
-	httpRestService, err := restserver.NewHTTPRestService(&config, fakes.NewFakeImdsClient())
+	httpRestService, err := restserver.NewHTTPRestService(&config, fakes.NewFakeImdsClient(), fakes.NewFakeNMAgentClient())
 	svc = httpRestService.(*restserver.HTTPRestService)
 	svc.Name = "cns-test-server"
 	svc.IPAMPoolMonitor = fakes.NewIPAMPoolMonitorFake()
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -30,7 +30,6 @@ type CNSConfig struct {
 	SyncHostNCVersionIntervalSec      int
 	SyncHostNCVersionIntervalMilliSec time.Duration
 	SyncHostNCTimeoutMilliSec         time.Duration
-	ForceMarkIPAvailableTimeRange     time.Duration
 }
 
 type TelemetrySettings struct {
@@ -136,5 +135,4 @@ func SetCNSConfigDefaults(config *CNSConfig) {
 	}
 	config.SyncHostNCVersionIntervalMilliSec = 1000
 	config.SyncHostNCTimeoutMilliSec = 500
-	config.ForceMarkIPAvailableTimeRange = 5000
 }
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -18,18 +18,17 @@ const (
 )
 
 type CNSConfig struct {
-	TelemetrySettings                 TelemetrySettings
-	ManagedSettings                   ManagedSettings
-	ChannelMode                       string
-	UseHTTPS                          bool
-	TLSSubjectName                    string
-	TLSCertificatePath                string
-	TLSPort                           string
-	TLSEndpoint                       string
-	WireserverIP                      string
-	SyncHostNCVersionIntervalSec      int
-	SyncHostNCVersionIntervalMilliSec time.Duration
-	SyncHostNCTimeoutMilliSec         time.Duration
+	TelemetrySettings           TelemetrySettings
+	ManagedSettings             ManagedSettings
+	ChannelMode                 string
+	UseHTTPS                    bool
+	TLSSubjectName              string
+	TLSCertificatePath          string
+	TLSPort                     string
+	TLSEndpoint                 string
+	WireserverIP                string
+	SyncHostNCVersionIntervalMs time.Duration
+	SyncHostNCTimeoutMs         time.Duration
 }
 
 type TelemetrySettings struct {
@@ -133,6 +132,6 @@ func SetCNSConfigDefaults(config *CNSConfig) {
 	if config.ChannelMode == "" {
 		config.ChannelMode = cns.Direct
 	}
-	config.SyncHostNCVersionIntervalMilliSec = 1000
-	config.SyncHostNCTimeoutMilliSec = 500
+	config.SyncHostNCVersionIntervalMs = 1000
+	config.SyncHostNCTimeoutMs = 500
 }
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/cns/logger"
@@ -17,14 +18,18 @@ const (
 )
 
 type CNSConfig struct {
-	TelemetrySettings            TelemetrySettings
-	ManagedSettings              ManagedSettings
-	ChannelMode                  string
-	UseHTTPS                     bool
-	TLSSubjectName               string
-	TLSCertificatePath           string
-	TLSPort                      string
-	SyncHostNCVersionIntervalSec int
+	TelemetrySettings                 TelemetrySettings
+	ManagedSettings                   ManagedSettings
+	ChannelMode                       string
+	UseHTTPS                          bool
+	TLSSubjectName                    string
+	TLSCertificatePath                string
+	TLSPort                           string
+	TLSEndpoint                       string
+	WireserverIP                      string
+	SyncHostNCVersionIntervalSec      int
+	SyncHostNCVersionIntervalMilliSec time.Duration
+	SyncHostNCTimeoutMilliSec         time.Duration
 }
 
 type TelemetrySettings struct {
@@ -128,5 +133,6 @@ func SetCNSConfigDefaults(config *CNSConfig) {
 	if config.ChannelMode == "" {
 		config.ChannelMode = cns.Direct
 	}
-	config.SyncHostNCVersionIntervalSec = 30
+	config.SyncHostNCVersionIntervalMilliSec = 1000
+	config.SyncHostNCTimeoutMilliSec = 500
 }
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -30,6 +30,7 @@ type CNSConfig struct {
 	SyncHostNCVersionIntervalSec      int
 	SyncHostNCVersionIntervalMilliSec time.Duration
 	SyncHostNCTimeoutMilliSec         time.Duration
+	ForceMarkIPAvailableTimeRange     time.Duration
 }
 
 type TelemetrySettings struct {
@@ -135,4 +136,5 @@ func SetCNSConfigDefaults(config *CNSConfig) {
 	}
 	config.SyncHostNCVersionIntervalMilliSec = 1000
 	config.SyncHostNCTimeoutMilliSec = 500
+	config.ForceMarkIPAvailableTimeRange = 5000
 }
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -126,7 +126,7 @@ func setManagedSettingDefaults(managedSettings *ManagedSettings) {
 	}
 }
 
-// Set Default values of CNS config if not specified
+// SetCNSConfigDefaults set default values of CNS config if not specified
 func SetCNSConfigDefaults(config *CNSConfig) {
 	setTelemetrySettingDefaults(&config.TelemetrySettings)
 	setManagedSettingDefaults(&config.ManagedSettings)
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -17,14 +17,14 @@ const (
 )
 
 type CNSConfig struct {
-	TelemetrySettings  TelemetrySettings
-	ManagedSettings    ManagedSettings
-	ChannelMode        string
-	UseHTTPS           bool
-	TLSSubjectName     string
-	TLSCertificatePath string
-	TLSPort            string
-	WireserverIP       string
+	TelemetrySettings            TelemetrySettings
+	ManagedSettings              ManagedSettings
+	ChannelMode                  string
+	UseHTTPS                     bool
+	TLSSubjectName               string
+	TLSCertificatePath           string
+	TLSPort                      string
+	SyncHostNCVersionIntervalSec int
 }
 
 type TelemetrySettings struct {
@@ -128,4 +128,5 @@ func SetCNSConfigDefaults(config *CNSConfig) {
 	if config.ChannelMode == "" {
 		config.ChannelMode = cns.Direct
 	}
+	config.SyncHostNCVersionIntervalSec = 30
 }
--- a/cns/fakes/nmagentclientfake.go
+++ b/cns/fakes/nmagentclientfake.go
@@ -1,0 +1,22 @@
+// Copyright 2017 Microsoft. All rights reserved.
+// MIT License
+
+package fakes
+
+// NMAgentClientTest can be used to query to VM Host info.
+type NMAgentClientTest struct {
+}
+
+// NewFakeNMAgentClient return a mock implemetation of NMAgentClient
+func NewFakeNMAgentClient() *NMAgentClientTest {
+	return &NMAgentClientTest{}
+}
+
+// GetNcVersionListWithOutToken is mock implementation to return nc version list.
+func (nmagentclient *NMAgentClientTest) GetNcVersionListWithOutToken(ncNeedUpdateList []string) map[string]int {
+	ncVersionList := make(map[string]int)
+	for _, ncID := range ncNeedUpdateList {
+		ncVersionList[ncID] = 0
+	}
+	return ncVersionList
+}
--- a/cns/fakes/nmagentclientfake.go
+++ b/cns/fakes/nmagentclientfake.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Microsoft. All rights reserved.
+// Copyright 2020 Microsoft. All rights reserved.
 // MIT License
 
 package fakes
--- a/cns/nmagentclient/nmagentclient.go
+++ b/cns/nmagentclient/nmagentclient.go
@@ -35,7 +35,10 @@ type NMANetworkContainerResponse struct {
 type NMAgentSupportedApisResponseXML struct {
 	SupportedApis []string `xml:"type"`
 }
+<<<<<<< HEAD
 
+=======
+>>>>>>> Resovle merge conflict from master.
 type ContainerInfo struct {
 	NetworkContainerID string `json:"networkContainerId"`
 	Version            string `json:"version"`
--- a/cns/nmagentclient/nmagentclient.go
+++ b/cns/nmagentclient/nmagentclient.go
@@ -191,7 +191,7 @@ func (nmagentclient *NMAgentClient) GetNcVersionListWithOutToken(ncNeedUpdateLis
 	now := time.Now()
 	response, err := http.Get(nmagentclient.connectionURL)
 	latency := time.Since(now)
-	logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken response: %+v. QueryURL is %s, latency is %v", response, nmagentclient.connectionURL, latency)
+	logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken response: %+v, latency is %d", response, latency.Milliseconds())
 
 	if response.StatusCode != http.StatusOK {
 		logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken failed with %d, err is %v", response.StatusCode, err)
@@ -202,7 +202,7 @@ func (nmagentclient *NMAgentClient) GetNcVersionListWithOutToken(ncNeedUpdateLis
 	rBytes, _ := ioutil.ReadAll(response.Body)
 	logger.Printf("Response body is %v", rBytes)
 	json.Unmarshal(rBytes, &nmaNcListResponse)
-	if nmaNcListResponse.ResponseCode != "200" {
+	if nmaNcListResponse.ResponseCode != strconv.Itoa(http.StatusOK) {
 		logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken unmarshal failed with %s", rBytes)
 		return nil
 	}
@@ -212,12 +212,12 @@ func (nmagentclient *NMAgentClient) GetNcVersionListWithOutToken(ncNeedUpdateLis
 		receivedNcVersionListInMap[containers.NetworkContainerID] = containers.Version
 	}
 	for _, ncID := range ncNeedUpdateList {
-		if val, ok := receivedNcVersionListInMap[ncID]; ok {
-			if valInInt, err := strconv.Atoi(val); err != nil {
-				logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken translate version %s to int failed with %s", val, err)
+		if version, ok := receivedNcVersionListInMap[ncID]; ok {
+			if versionInInt, err := strconv.Atoi(version); err != nil {
+				logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken translate version %s to int failed with %s", version, err)
 			} else {
-				ncVersionList[ncID] = valInInt
-				logger.Printf("Containers id is %s, version is %d", ncID, valInInt)
+				ncVersionList[ncID] = versionInInt
+				logger.Printf("Containers id is %s, programmed NC version is %d", ncID, versionInInt)
 			}
 		}
 	}
--- a/cns/nmagentclient/nmagentclient.go
+++ b/cns/nmagentclient/nmagentclient.go
@@ -5,7 +5,10 @@ import (
 	"encoding/json"
 	"encoding/xml"
 	"fmt"
+	"io/ioutil"
 	"net/http"
+	"strconv"
+	"time"
 
 	"github.com/Azure/azure-container-networking/cns/logger"
 	"github.com/Azure/azure-container-networking/common"
@@ -13,12 +16,14 @@ import (
 
 const (
 	//GetNmAgentSupportedApiURLFmt Api endpoint to get supported Apis of NMAgent
-	GetNmAgentSupportedApiURLFmt     = "http://%s/machine/plugins/?comp=nmagent&type=GetSupportedApis"
-	GetNetworkContainerVersionURLFmt = "http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/%s/networkContainers/%s/version/authenticationToken/%s/api-version/1"
+	GetNmAgentSupportedApiURLFmt       = "http://%s/machine/plugins/?comp=nmagent&type=GetSupportedApis"
+	GetNetworkContainerVersionURLFmt   = "http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/%s/networkContainers/%s/version/authenticationToken/%s/api-version/1"
+	GetNcVersionListWithOutTokenURLFmt = "http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/api-version/%s"
 )
 
 //WireServerIP - wire server ip
 var WireserverIP = "168.63.129.16"
+var getNcVersionListWithOutTokenURLVersion = "2"
 
 // NMANetworkContainerResponse - NMAgent response.
 type NMANetworkContainerResponse struct {
@@ -29,6 +34,36 @@ type NMANetworkContainerResponse struct {
 
 type NMAgentSupportedApisResponseXML struct {
 	SupportedApis []string `xml:"type"`
+}
+
+type ContainerInfo struct {
+	NetworkContainerID string `json:"networkContainerId"`
+	Version            string `json:"version"`
+}
+
+type NMANetworkContainerListResponse struct {
+	ResponseCode string          `json:"httpStatusCode"`
+	Containers   []ContainerInfo `json:"networkContainers"`
+}
+
+// NMAgentClient is client to handle queries to nmagent
+type NMAgentClient struct {
+	connectionURL string
+}
+
+// NMAgentClientInterface has interface that nmagent client will handle
+type NMAgentClientInterface interface {
+	GetNcVersionListWithOutToken(ncNeedUpdateList []string) map[string]int
+}
+
+// NewNMAgentClient create a new nmagent client.
+func NewNMAgentClient(url string) (*NMAgentClient, error) {
+	if url == "" {
+		url = fmt.Sprintf(GetNcVersionListWithOutTokenURLFmt, WireserverIP, getNcVersionListWithOutTokenURLVersion)
+	}
+	return &NMAgentClient{
+		connectionURL: url,
+	}, nil
 }
 
 // JoinNetwork joins the given network
@@ -148,4 +183,43 @@ func GetNmAgentSupportedApis(
 
 	logger.Printf("[NMAgentClient][Response] GetNmAgentSupportedApis. Response: %+v.", response)
 	return xmlDoc.SupportedApis, nil
+}
+
+// GetNcVersionListWithOutToken query nmagent for programmed container version.
+func (nmagentclient *NMAgentClient) GetNcVersionListWithOutToken(ncNeedUpdateList []string) map[string]int {
+	ncVersionList := make(map[string]int)
+	now := time.Now()
+	response, err := http.Get(nmagentclient.connectionURL)
+	latency := time.Since(now)
+	logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken response: %+v. QueryURL is %s, latency is %v", response, nmagentclient.connectionURL, latency)
+
+	if response.StatusCode != http.StatusOK {
+		logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken failed with %d, err is %v", response.StatusCode, err)
+		return nil
+	}
+
+	var nmaNcListResponse NMANetworkContainerListResponse
+	rBytes, _ := ioutil.ReadAll(response.Body)
+	logger.Printf("Response body is %v", rBytes)
+	json.Unmarshal(rBytes, &nmaNcListResponse)
+	if nmaNcListResponse.ResponseCode != "200" {
+		logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken unmarshal failed with %s", rBytes)
+		return nil
+	}
+
+	var receivedNcVersionListInMap = make(map[string]string)
+	for _, containers := range nmaNcListResponse.Containers {
+		receivedNcVersionListInMap[containers.NetworkContainerID] = containers.Version
+	}
+	for _, ncID := range ncNeedUpdateList {
+		if val, ok := receivedNcVersionListInMap[ncID]; ok {
+			if valInInt, err := strconv.Atoi(val); err != nil {
+				logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken translate version %s to int failed with %s", val, err)
+			} else {
+				ncVersionList[ncID] = valInInt
+				logger.Printf("Containers id is %s, version is %d", ncID, valInInt)
+			}
+		}
+	}
+	return ncVersionList
 }
--- a/cns/nmagentclient/nmagentclient.go
+++ b/cns/nmagentclient/nmagentclient.go
@@ -35,10 +35,7 @@ type NMANetworkContainerResponse struct {
 type NMAgentSupportedApisResponseXML struct {
 	SupportedApis []string `xml:"type"`
 }
-<<<<<<< HEAD
 
-=======
->>>>>>> Resovle merge conflict from master.
 type ContainerInfo struct {
 	NetworkContainerID string `json:"networkContainerId"`
 	Version            string `json:"version"`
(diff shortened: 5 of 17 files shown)


cns/restserver/internalapi.go
```diff
@@ -151,6 +152,37 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+		if hostVersion >= vmVersion {
+			continue
+		} else {
+			newHostNCVersion := service.imdsClient.GetNetworkContainerInfoFromHostWithoutToken()
+			service.Lock()
+			if channelMode == cns.KubernetesCRD {
+				service.UpdatePendingProgrammingIPs(strconv.Itoa(newHostNCVersion), containerstatus.CreateNetworkContainerRequest)
+			}
+			containerstatus.HostVersion = strconv.Itoa(newHostNCVersion)
+			service.state.ContainerStatus[i] = containerstatus
```
neaggarwMS, line 180: This line is not required. You are updating the object directly. So need to reassign it.

cns/restserver/internalapi.go
```diff
@@ -151,6 +152,37 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+			log.Errorf("Received err when chagne containerstatus.VMVersion %s to int, err msg %v", containerstatus.VMVersion, err)
+			return
+		}
+		// host NC version is the NC version from NMAgent, if it's already keep up with NC version exist in VM, no update needed.
+		if hostVersion >= vmVersion {
+			continue
+		} else {
+			newHostNCVersion := service.imdsClient.GetNetworkContainerInfoFromHostWithoutToken()
+			service.Lock()
+			if channelMode == cns.KubernetesCRD {
```
neaggarwMS, line 176: Instead can you check if this NC has secondaryIps?

cns/restserver/internalapi.go
```diff
@@ -151,6 +152,37 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
 	return
 }
 
+// SyncHostNCVersion will check NC version from NMAgent and save it as host version in container status.
+// If NMAgent updated, CNS will refresh the pending programming IP status.
+func (service *HTTPRestService) SyncHostNCVersion(channelMode string) {
```
neaggarwMS, line 157: Can you move this function out from internalapi.go? InternalApi is used for APIs exposed for CNS service for internal consumption. This is an API invoked for the background go routine like IpamPoolMonitor. Can you move it to a separate file like SyncNCProgrammingVersion.go?

cns/restserver/ipam.go
```diff
@@ -112,6 +113,24 @@ func (service *HTTPRestService) MarkIPsAsPending(numberToMark int) (map[string]c
 	return nil, fmt.Errorf("Failed to mark %d IP's as pending, only marked %d IP's", numberToMark, len(pendingReleaseIPs))
 }
 
+// UpdatePendingProgrammingIPs will update pending programming IPs to available if
+// NMAgent side's programmed NC version keep up with NC version with secondary IP.
+// This function must be called in a service lock.
+func (service *HTTPRestService) UpdatePendingProgrammingIPs(nmagentNCVersion string, req cns.CreateNetworkContainerRequest) {
```
neaggarwMS, line 119: Can you rename it to MarkIpsAsAvailableUntransacted. Reason for untransacted again because caller is taking care of the lock.

neaggarwMS: We also need to log how much time it took for nmagent to actually update the version. Its okay to capture it in a different PR, but please add a WI to track the work.

cns/service/main.go
```diff
@@ -535,6 +535,14 @@ func main() {
 		}
 	}
 
+	go func() {
+		// Periodically poll NC version from NMAgent
+		for {
```
neaggarwMS, line 540: If this thread crashes, how do you restart?

cns/service/main.go
```diff
@@ -535,6 +535,14 @@ func main() {
 		}
 	}
 
+	go func() {
```
neaggarwMS, line 538: Can we initialize this as how IpamPoolMonitor is started?

neaggarwMS: Also can you check with Ashvin, do we need to add this thread for all scenarios? (Both Managed DNC and non-Managed as well?)

cns/api.go
```diff
@@ -37,6 +37,7 @@ type HTTPService interface {
 	SendNCSnapShotPeriodically(int, chan bool)
 	SetNodeOrchestrator(*SetOrchestratorTypeRequest)
 	SyncNodeStatus(string, string, string, json.RawMessage) (int, string)
+	SyncHostNCVersion(string)
```
neaggarwMS, line 40: I dont think this API makes sense here. this interface should only exposed APIs which are exposed by service and be invoked by clients or other components (like CNI)

cns/restserver/internalapi.go
```diff
@@ -151,6 +152,37 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+		if err != nil {
+			log.Errorf("Received err when chagne containerstatus.VMVersion %s to int, err msg %v", containerstatus.VMVersion, err)
+			return
+		}
+		// host NC version is the NC version from NMAgent, if it's already keep up with NC version exist in VM, no update needed.
+		if hostVersion >= vmVersion {
+			continue
+		} else {
+			newHostNCVersion := service.imdsClient.GetNetworkContainerInfoFromHostWithoutToken()
+			service.Lock()
```
neaggarwMS, line 175: Instead of updating it here, I would expose a new Transacted API in HttpService (internalApi.go) like UpdateNetworkContainerProgrammedVersion(). And update ProgrammedVersion + Marking IPs as available in that function.

neaggarwMS (changes requested):

cns/restserver/internalapi.go
```diff
@@ -151,6 +152,37 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
 	return
 }
 
+// SyncHostNCVersion will check NC version from NMAgent and save it as host version in container status.
+// If NMAgent updated, CNS will refresh the pending programming IP status.
+func (service *HTTPRestService) SyncHostNCVersion(channelMode string) {
+	for i, containerstatus := range service.state.ContainerStatus {
```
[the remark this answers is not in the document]
csfmomo: Read lock is needed when reading state, I'll add.
Update nmagent version in container status need write lock.
For checking nmagent version, it's better not in any lock.

cns/restserver/internalapi.go
csfmomo: This line is needed. In golang, when updating map value, reassign is needed. Please refer this link https://stackoverflow.com/questions/42716852/how-to-update-map-values-in-go
image

cns/service/main.go
csfmomo: I think Ashvin mentioned during syncing up meeting that update NC's host version(the version programmed in NMAgent) is necessary for all scenarios. But anyway, I can double check with him.

cns/restserver/internalapi.go
```diff
@@ -151,6 +152,37 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+		vmVersion, err := strconv.Atoi(containerstatus.VMVersion)
+		if err != nil {
+			log.Errorf("Received err when chagne containerstatus.VMVersion %s to int, err msg %v", containerstatus.VMVersion, err)
+			return
+		}
+		// host NC version is the NC version from NMAgent, if it's already keep up with NC version exist in VM, no update needed.
+		if hostVersion >= vmVersion {
+			continue
+		} else {
+			newHostNCVersion := service.imdsClient.GetNetworkContainerInfoFromHostWithoutToken()
```
ashvindeodhar, line 174: just double check if the http client we use to call the nmagent api has timeouts. We don't want to get into a situation where we are indefinitely waiting while nma api.

ashvindeodhar, line 174: I see GetNetworkContainerInfoFromHostWithoutToken has a temp implementation at the moment which returns max int value. Are you planning on changing the implementation as a part of this PR? If not, this code will set the newHostVersion to max int value.
Also, I believe you would need to check if the new API ( without the token ) is supported by the nmagent on this node, if yes, call the api without the token if not, call the api which needs the token. isNCWaitingForUpdate func has the implementation to call with the token.

cns/service/main.go
ashvindeodhar: yes, but you need to check if the NMA on this node supports the API without the token. If that is supported, you would call that API else call the API with the token. I don;t see that code change in this PR. GetNetworkContainerInfoFromHostWithoutToken is missing the correct implementation to call that NMA Api. nmagentclient.go has GetNetworkContainerVersion which is calling the api with tokens.

ashvindeodhar (changes requested):

cns/restserver/internalapi.go
```diff
@@ -151,6 +152,37 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
 }
 
+// SyncHostNCVersion will check NC version from NMAgent and save it as host version in container status.
+// If NMAgent updated, CNS will refresh the pending programming IP status.
+func (service *HTTPRestService) SyncHostNCVersion(channelMode string) {
+	for i, containerstatus := range service.state.ContainerStatus {
+		// Will open a separate PR to convert all the NC version related variable to int. Change from string to int is a pain.
+		hostVersion, err := strconv.Atoi(containerstatus.HostVersion)
+		if err != nil {
+			log.Errorf("Received err when chagne containerstatus.HostVersion %s to int, err msg %v", containerstatus.HostVersion, err)
```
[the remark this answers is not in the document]
csfmomo: Eagle eye. Let me change it.

cns/restserver/internalapi.go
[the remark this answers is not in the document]
csfmomo: OK.

cns/restserver/internalapi_test.go
```diff
@@ -116,6 +116,81 @@ func TestCreateAndUpdateNCWithSecondaryIPNCVersion(t *testing.T) {
 	}
 }
 
+func TestSyncHostNCVersion(t *testing.T) {
+	// cns.KubernetesCRD has one more logic compared to other orchestrator type, so test both of them
+	orchestratorTypes := [6]string{cns.Kubernetes, cns.KubernetesCRD}
```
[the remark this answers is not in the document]
csfmomo: Sorry, previously I tested all 6 different orchestrator types. Then I found out checking these 2 should be good enough.

cns/restserver/internalapi.go
csfmomo: I'm not planning to put implementation in this PR. Yes, it will return max int. Return max int will avoid breaking current logic.

cns/configuration/configuration.go
```diff
@@ -123,4 +124,5 @@ func SetCNSConfigDefaults(config *CNSConfig) {
 	if config.ChannelMode == "" {
 		config.ChannelMode = cns.Direct
 	}
+	config.SyncHostNCVersionIntervalSec = 30
```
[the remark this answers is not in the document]
csfmomo: Update to 1.

cns/restserver/internalapi.go
```diff
@@ -151,6 +152,37 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+			return
+		}
+		vmVersion, err := strconv.Atoi(containerstatus.VMVersion)
+		if err != nil {
+			log.Errorf("Received err when chagne containerstatus.VMVersion %s to int, err msg %v", containerstatus.VMVersion, err)
+			return
+		}
+		// host NC version is the NC version from NMAgent, if it's already keep up with NC version exist in VM, no update needed.
```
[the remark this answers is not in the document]
csfmomo: hostVersion is the name I inheritant from existing CNS code.  I can rename it as programmedNcVersion.

cns/restserver/internalapi.go
```diff
@@ -151,6 +152,37 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+		if hostVersion >= vmVersion {
```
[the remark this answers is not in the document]
csfmomo: Do we really want to log here? This go routine will execute every second and most of time, host nc version should keep up with nc version and no further operation needed.

cns/restserver/internalapi.go
csfmomo: Check whether NC has secondarIps is taken care by the function inside this if statement. I tested it, if no secondaryIps, loop through secondaryIps will return directly.
If we only use SyncHostNCVersion for CRD mode for now, I can remove this check.
If other scenario, e.g. managed DNC want to extend our SyncHostNCVersion function, we can add channelMode back.

cns/restserver/internalapi.go
csfmomo: I can move it to a separate file but not sure whether we should do it. We can sync offline for this, I'm fine in both ways.
All HTTPRestService are listed in that file, include background thread SyncNodeStatus.
In other side, SyncHostNCVersion logic is not complex when compared to IpamPoolMonitor. Ipam pool monitor has 6 functions but SyncHostNCVersion only need get NC version and update it.

cns/restserver/ipam.go
csfmomo: 1. Sure, I'll change the name.
2. I'll add it in this PR, only need 2 more extra lines.

cns/service/main.go
csfmomo: Restart is taken care by the next line time.NewTicker. Ticker will let us do something repeatedly at regular intervals.

cns/service/main.go
csfmomo: The initialization is the same as IpamPoolMonitor except IpamPoolMonitor use context for context.done feature. Context.Done give us visibility(logging) when IpamPoolMonitor go routine received cancellation signal. Do we want this visibility for SyncHostNCVersion? If yes, I can add context feature to it.

cns/api.go
csfmomo: I'll removed it from interface.

cns/cnsclient/cnsclient_test.go
```diff
@@ -63,6 +63,9 @@ func addTestStateToRestServer(t *testing.T, secondaryIps []string) {
 		NetworkContainerid:   "testNcId1",
 		IPConfiguration:      ipConfig,
 		SecondaryIPConfigs:   secondaryIPConfigs,
+		// Set it as -1 to be same as default host version.
+		// It will allow secondary IPs status to be set as available.
```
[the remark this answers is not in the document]
csfmomo: It's a unit testing code. For unit test, I set default NMAgent NC version as 0. I use different NC version to trigger different test cases.

cns/restserver/ipam.go
csfmomo: 1. I'll Updated name to MarkIpsAsAvailableUntransacted
2. I'll add time consuming in this PR. It's a 2 line change.

cns/configuration/configuration.go
```diff
@@ -120,11 +124,14 @@ func setManagedSettingDefaults(managedSettings *ManagedSettings) {
...
+// SetCNSConfigDefaults set default values of CNS config if not specified
 func SetCNSConfigDefaults(config *CNSConfig) {
 	setTelemetrySettingDefaults(&config.TelemetrySettings)
 	setManagedSettingDefaults(&config.ManagedSettings)
 	if config.ChannelMode == "" {
 		config.ChannelMode = cns.Direct
 	}
+	config.SyncHostNCVersionIntervalMilliSec = 1000
+	config.SyncHostNCTimeoutMilliSec = 500
+	config.ForceMarkIPAvailableTimeRange = 5000
```
neaggarwMS, line 138: What does ForceMarkIPAvailableTimeRange  represent?

neaggarwMS (changes requested):

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,68 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
 	return
 }
 
+// SyncHostNCVersion will check NC version from NMAgent and save it as host NC version in container status.
+// If NMAgent NC version got updated, CNS will refresh the pending programming IP status.
+func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMode string, lastUpdatedHostNCTimeStamp time.Time, forceMarkIPAvailableTimeRange time.Duration) {
```
neaggarwMS, line 151: I highly recommend to move this to a separate file (like IPAMPoolMonitor). we only add apis called by internal cns client in this file.

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,68 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+	var hostVersionNeedUpdateNcList []string
+	service.RLock()
+	for _, containerstatus := range service.state.ContainerStatus {
+		// Will open a separate PR to convert all the NC version related variable to int. Change from string to int is a pain.
+		hostNcVersion, err := strconv.Atoi(containerstatus.HostNCVersion)
+		if err != nil {
+			log.Errorf("Received err when change containerstatus.HostNCVersion %s to int, err msg %v", containerstatus.HostNCVersion, err)
+			return
+		}
+		dncNcVersion, err := strconv.Atoi(containerstatus.DncNCVersion)
```
neaggarwMS, line 161: Can you compare it against containerStatus.Req.Version? I really dont know the usage of containerStatus.VMVersion and dont want to take dependency on it.

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,68 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+		go func() {
+			c1 <- service.nmagentClient.GetNcVersionListWithOutToken(hostVersionNeedUpdateNcList)
+		}()
+		select {
+		case newHostNCVersionList := <-c1:
+			service.Lock()
+			if newHostNCVersionList == nil {
+				logger.Errorf("Can't get vfp programmed NC version list from url without token")
+				if channelMode == cns.CRD && time.Since(lastUpdatedHostNCTimeStamp) > forceMarkIPAvailableTimeRange {
+					service.MarkAllPendingProgrammingIpsAsAvailableUntransacted()
```
neaggarwMS, line 183: I dont understand this change. If NMAgent is not returning any list, how does marking an IP to available will help? If we assign this IP to a pod, data path will still be broken, right?

cns/restserver/ipam.go
```diff
@@ -113,6 +114,45 @@ func (service *HTTPRestService) MarkIPsAsPending(numberToMark int) (map[string]c
 	return nil, fmt.Errorf("Failed to mark %d IP's as pending, only marked %d IP's", numberToMark, len(pendingReleaseIPs))
 }
 
+// MarkAllPendingProgrammingIpsAsAvailableUntransacted is the function to update pending programming IPs to available
+// when get NC version failed and we don't want to block IP allocation.
+// Note: this func is an untransacted API as the caller will take a Service lock
+func (service *HTTPRestService) MarkAllPendingProgrammingIpsAsAvailableUntransacted() {
```
neaggarwMS, line 120: This doesnt look right to me, this defeats the purpose that we will ensure IP is programmed before we assign it to the pod

cns/restserver/restserver.go
```diff
@@ -57,8 +59,8 @@ type allocatedIPCount struct {
 // containerstatus is used to save status of an existing container
 type containerstatus struct {
 	ID                            string
-	VMVersion                     string
-	HostVersion                   string
+	DncNCVersion                  string
```
neaggarwMS, line 62: Wont this break backcompat? CNS is already running in production and JSON file stored has previous names, how will you unmarshel it back to this new struct? Also I dont understand what does DncNCVersion means?

neaggarwMS: Also not sure if this VMCersion is NC version. I believe it is the VM Version only. We are not populating it for AKS scenario. Actual NC version is inside the CreateNetworkContainerRequest

cns/restserver/restserver.go
```diff
@@ -84,7 +86,7 @@ type networkInfo struct {
 }
 
 // NewHTTPRestService creates a new HTTP Service object.
-func NewHTTPRestService(config *common.ServiceConfig, imdsClientInterface imdsclient.ImdsClientInterface) (cns.HTTPService, error) {
+func NewHTTPRestService(config *common.ServiceConfig, imdsClientInterface imdsclient.ImdsClientInterface, nmagentClient nmagentclient.NMAgentClientInterface) (cns.HTTPService, error) {
```
neaggarwMS, line 89: Can we please log a WI to remove imdsClient completely.

cns/service/main.go
```diff
@@ -562,6 +567,18 @@ func main() {
...
 		}
 
+		logger.Printf("Starting SyncHostNCVersion")
+		lastUpdatedHostNCTimeStamp := time.Now()
+		rootCxt := context.Background()
+		ctxWithTimeout, _ := context.WithTimeout(rootCxt, cnsconfig.SyncHostNCTimeoutMilliSec)
+		go func() {
+			// Periodically poll vfp programmed NC version from NMAgent
+			for {
+				<-time.NewTicker(cnsconfig.SyncHostNCVersionIntervalMilliSec * time.Second).C
```
neaggarwMS, line 577: As discussed instead of making it run periodically, lets make it signal based and trigger as soon as we got an NC update.

cns/service/main.go
```diff
@@ -562,6 +567,18 @@ func main() {
 			return
 		}
 
+		logger.Printf("Starting SyncHostNCVersion")
```
neaggarwMS, line 570: Can you move this logic to a different file, Follow how IPAMPoolMonitor is initialized

neaggarwMS (dismissed):

cns/configuration/configuration.go
csfmomo: It represents the AKS fall back logic when NMAgent is done,

cns/nmagentclient/nmagentclient.go
```diff
@@ -27,8 +32,39 @@ type NMANetworkContainerResponse struct {
 	Version            string `json:"version"`
 }
 
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
csfmomo: Fixed.

cns/nmagentclient/nmagentclient.go
```diff
@@ -98,6 +134,7 @@ func GetNetworkContainerVersion(
 	return response, err
 }
 
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
csfmomo: Fixed.

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -38,6 +39,8 @@ func CRDStatusToNCRequest(crdStatus nnc.NodeNetworkConfigStatus) (cns.CreateNetw
 		ncRequest.NetworkContainerid = nc.ID
 		ncRequest.NetworkContainerType = cns.Docker
 		ncRequest.Version = strconv.FormatInt(nc.Version, 10)
+		log.Printf("Set nc request info with SecondaryIPConfigs %v, NetworkContainerid %s, NetworkContainerType %s, NC Version %s",
```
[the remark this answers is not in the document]
csfmomo: I think it should stay here but removing the part of printing SecondaryIPConfigs and change the word Set to Trying set.  SecondaryIPConfigs dosn't have any value here. SecondaryIPConfigs  start getting value from line #71. However, the other 3 variable already finished getting value so that's why I want to print this log here.

cns/restserver/internalapi.go
csfmomo: This file is not for internal cns call from begining. It's AKS start using it for internal CNS call. There are 6 functions in internalapi.go, only 2 of them are called by cns client. CreateOrUpdateNetworkContainerInternal and ReconcileNCState are called by cns client. The rest of 4 are not called by cns client.

cns/restserver/restserver.go
csfmomo: 1. It's backward compatible. No unmarshal happens in containerstatus.
2. DNCNCVersion means NC version which coming from DNC. This PR is resolving the issue when discrepancy happens in NC version from DNC and from NMAgent.
3. VMVersion is NC version. It get value from nc request version and compared with nc request version. Code reference 
https://github.com/Azure/azure-container-networking/blob/master/cns/restserver/util.go#L135
https://github.com/Azure/azure-container-networking/blob/master/cns/restserver/api.go#L789

cns/restserver/restserver.go
csfmomo: Sure, WI 8874138 has been created.

cns/service/main.go
csfmomo: Need to sync up for it.

cns/service/main.go
csfmomo: Not sure what's the benefit of moving to a different file. Keep it here we can also leverage the advantage of using context with timeout and retry every second.

cns/service/main.go
csfmomo: Will use a new PR to address this feedback.

cns/restserver/restserver.go
csfmomo: After discussion, we'll remain the VMVersion variable name here.

cns/restserver/internalapi.go
csfmomo: OK

cns/restserver/restserver.go
csfmomo: Remained it as VMVersion here.

cns/restserver/internalapi.go
csfmomo: Could I remain the function here for this version? 
I can do moving logic to separate file for next PR and we can move all the related logic, including what managed DNC use in a separate file.

cns/configuration/configuration.go
csfmomo: Discuessed offline, moved this part.

cns/restserver/const.go
```diff
@@ -38,6 +38,7 @@ const (
 	NetworkContainerVfpProgramComplete     = 35
 	NetworkContainerVfpProgramCheckSkipped = 36
 	NmAgentSupportedApisError              = 37
+	UnsupportedNCVersion                   = 37
```
matmerr, line 41: new const?

csfmomo: Yep, it's a mistake during resolving merge conflicts. Let me fix it.

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+		ctxWithTimeout, _ := context.WithTimeout(ctx, syncHostNCTimeoutMilliSec*time.Millisecond)
+		go func() {
+			c1 <- service.nmagentClient.GetNcVersionListWithOutToken(hostVersionNeedUpdateNcList)
+			close(c1)
+		}()
+		select {
+		case newHostNCVersionList := <-c1:
+			service.Lock()
+			if newHostNCVersionList == nil {
+				logger.Errorf("Can't get vfp programmed NC version list from url without token")
```
ashvindeodhar, line 183: if we return error from GetNcVersionListWithOutToken, we can print it here

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+			return
+		}
+		// host NC version is the NC version from NMAgent, if it's smaller than
+		if hostVersion < dncNcVersion {
+			hostVersionNeedUpdateNcList = append(hostVersionNeedUpdateNcList, containerstatus.ID)
+		}
+	}
+	service.RUnlock()
+	if len(hostVersionNeedUpdateNcList) > 0 {
+		c1 := make(chan map[string]int)
```
ashvindeodhar, line 173: can you rename the channel to something indicative of rather than c1 ?

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+			c1 <- service.nmagentClient.GetNcVersionListWithOutToken(hostVersionNeedUpdateNcList)
+			close(c1)
+		}()
+		select {
+		case newHostNCVersionList := <-c1:
+			service.Lock()
+			if newHostNCVersionList == nil {
+				logger.Errorf("Can't get vfp programmed NC version list from url without token")
+			} else {
+				for ncID, newHostNCVersion := range newHostNCVersionList {
```
ashvindeodhar, line 185: nitpick: you can take the lock inside else because lock is not needed if if newHostNCVersionList == nil {

cns/service/main.go
```diff
@@ -453,8 +453,13 @@ func main() {
 		return
 	}
 
+	nmaclient, err := nmagentclient.NewNMAgentClient("")
+	if err != nil {
+		logger.Errorf("Failed to start nmagent client due to error %v\n", err)
```
ashvindeodhar, line 458: I don't think \n is needed

cns/service/main.go
```diff
@@ -453,8 +453,13 @@ func main() {
 		return
 	}
 
+	nmaclient, err := nmagentclient.NewNMAgentClient("")
```
ashvindeodhar, line 456: what's the use of having empty param to this?

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
 	return
 }
 
+// SyncHostNCVersion will check NC version from NMAgent and save it as host NC version in container status.
+// If NMAgent NC version got updated, CNS will refresh the pending programming IP status.
+func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMode string, syncHostNCTimeoutMilliSec time.Duration) {
+	var hostVersionNeedUpdateNcList []string
+	service.RLock()
+	for _, containerstatus := range service.state.ContainerStatus {
```
ashvindeodhar, line 154: If you want, you can extract this earlier portion in a separate function - which returns hostVersionNeedUpdateNcList  which this function uses to call to nma. It's ok if you want to keep it as is

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+		if hostVersion < dncNcVersion {
+			hostVersionNeedUpdateNcList = append(hostVersionNeedUpdateNcList, containerstatus.ID)
+		}
+	}
+	service.RUnlock()
+	if len(hostVersionNeedUpdateNcList) > 0 {
+		c1 := make(chan map[string]int)
+		ctxWithTimeout, _ := context.WithTimeout(ctx, syncHostNCTimeoutMilliSec*time.Millisecond)
+		go func() {
+			c1 <- service.nmagentClient.GetNcVersionListWithOutToken(hostVersionNeedUpdateNcList)
```
ashvindeodhar, line 176: How will this work for the NMA which don't support the without token?

cns/configuration/configuration.go
```diff
@@ -17,14 +18,18 @@ const (
...
+	ManagedSettings                   ManagedSettings
+	ChannelMode                       string
+	UseHTTPS                          bool
+	TLSSubjectName                    string
+	TLSCertificatePath                string
+	TLSPort                           string
+	TLSEndpoint                       string
+	WireserverIP                      string
+	SyncHostNCVersionIntervalSec      int
+	SyncHostNCVersionIntervalMilliSec time.Duration
```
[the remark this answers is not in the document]
csfmomo: We only need one. Another one is the unnecessary info when resolve merge conflicts. Let me remove it.

cns/configuration/configuration.go
```diff
@@ -17,14 +18,18 @@ const (
...
+	SyncHostNCTimeoutMilliSec         time.Duration
```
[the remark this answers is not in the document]
csfmomo: Sure, let me update it.

cns/fakes/nmagentclientfake.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2017 Microsoft. All rights reserved.
```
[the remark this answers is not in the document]
csfmomo: Sure.

cns/nmagentclient/nmagentclient.go
```diff
@@ -149,3 +184,42 @@ func GetNmAgentSupportedApis(
...
 	return xmlDoc.SupportedApis, nil
 }
+
+// GetNcVersionListWithOutToken query nmagent for programmed container version.
+func (nmagentclient *NMAgentClient) GetNcVersionListWithOutToken(ncNeedUpdateList []string) map[string]int {
+	ncVersionList := make(map[string]int)
+	now := time.Now()
+	response, err := http.Get(nmagentclient.connectionURL)
+	latency := time.Since(now)
+	logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken response: %+v. QueryURL is %s, latency is %v", response, nmagentclient.connectionURL, latency)
```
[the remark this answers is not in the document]
csfmomo: Sure.

cns/nmagentclient/nmagentclient.go
```diff
@@ -149,3 +184,42 @@ func GetNmAgentSupportedApis(
...
+	if response.StatusCode != http.StatusOK {
+		logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken failed with %d, err is %v", response.StatusCode, err)
+		return nil
+	}
+
+	var nmaNcListResponse NMANetworkContainerListResponse
+	rBytes, _ := ioutil.ReadAll(response.Body)
+	logger.Printf("Response body is %v", rBytes)
+	json.Unmarshal(rBytes, &nmaNcListResponse)
+	if nmaNcListResponse.ResponseCode != "200" {
```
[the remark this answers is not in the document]
csfmomo: Sure.

cns/nmagentclient/nmagentclient.go
```diff
@@ -149,3 +184,42 @@ func GetNmAgentSupportedApis(
...
+		logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken unmarshal failed with %s", rBytes)
+		return nil
+	}
+
+	var receivedNcVersionListInMap = make(map[string]string)
+	for _, containers := range nmaNcListResponse.Containers {
+		receivedNcVersionListInMap[containers.NetworkContainerID] = containers.Version
+	}
+	for _, ncID := range ncNeedUpdateList {
+		if val, ok := receivedNcVersionListInMap[ncID]; ok {
```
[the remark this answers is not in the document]
csfmomo: yes.

cns/nmagentclient/nmagentclient.go
```diff
@@ -149,3 +184,42 @@ func GetNmAgentSupportedApis(
...
+			if valInInt, err := strconv.Atoi(val); err != nil {
+				logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken translate version %s to int failed with %s", val, err)
+			} else {
+				ncVersionList[ncID] = valInInt
+				logger.Printf("Containers id is %s, version is %d", ncID, valInInt)
```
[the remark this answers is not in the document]
csfmomo: OK.

cns/requestcontroller/kubecontroller/crdtranslator.go
```diff
@@ -66,7 +67,10 @@ func CRDStatusToNCRequest(crdStatus nnc.NodeNetworkConfigStatus) (cns.CreateNetw
 				NCVersion: ncVersion,
 			}
 			ncRequest.SecondaryIPConfigs[ipAssignment.Name] = secondaryIPConfig
+			log.Debugf("Seconday IP Configs got set, name is %s, config is %v", ipAssignment.Name, secondaryIPConfig)
 		}
+		log.Printf("Set nc request info with NetworkContainerid %s, NetworkContainerType %s, NC Version %s",
```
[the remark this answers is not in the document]
csfmomo: Sure.

cns/restserver/internalapi.go
csfmomo: Let me keep as it is for now. I can have another PR to make structure better.

cns/restserver/internalapi.go
csfmomo: Synced up offline. This logic is for AKS only so we'll keep as it is. When we want to extend it to other scenarios, we'll add a logic to fall back to with token API.

cns/service/main.go
csfmomo: Updated.

cns/restserver/internalapi.go
csfmomo: Updated.

cns/service/main.go
csfmomo: In case we want to override nmagent client url.

cns/restserver/internalapi.go
csfmomo: Rename it to ncVersionChannel

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+			log.Errorf("Received err when change containerstatus.HostVersion %s to int, err msg %v", containerstatus.HostVersion, err)
+			return
+		}
+		dncNcVersion, err := strconv.Atoi(containerstatus.CreateNetworkContainerRequest.Version)
+		if err != nil {
+			log.Errorf("Received err when change nc version %s in containerstatusto int, err msg %v", containerstatus.CreateNetworkContainerRequest.Version, err)
+			return
+		}
+		// host NC version is the NC version from NMAgent, if it's smaller than
+		if hostVersion < dncNcVersion {
```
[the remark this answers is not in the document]
csfmomo: Added
image

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+		if err != nil {
+			log.Errorf("Received err when change containerstatus.HostVersion %s to int, err msg %v", containerstatus.HostVersion, err)
+			return
+		}
+		dncNcVersion, err := strconv.Atoi(containerstatus.CreateNetworkContainerRequest.Version)
+		if err != nil {
+			log.Errorf("Received err when change nc version %s in containerstatusto int, err msg %v", containerstatus.CreateNetworkContainerRequest.Version, err)
+			return
+		}
+		// host NC version is the NC version from NMAgent, if it's smaller than
```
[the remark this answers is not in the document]
csfmomo: Complete it as "host NC version is the NC version from NMAgent, if it's smaller than NC version from DNC, then append it to indicate it needs update."

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+	for _, containerstatus := range service.state.ContainerStatus {
+		// Will open a separate PR to convert all the NC version related variable to int. Change from string to int is a pain.
+		hostVersion, err := strconv.Atoi(containerstatus.HostVersion)
+		if err != nil {
+			log.Errorf("Received err when change containerstatus.HostVersion %s to int, err msg %v", containerstatus.HostVersion, err)
+			return
+		}
+		dncNcVersion, err := strconv.Atoi(containerstatus.CreateNetworkContainerRequest.Version)
+		if err != nil {
+			log.Errorf("Received err when change nc version %s in containerstatusto int, err msg %v", containerstatus.CreateNetworkContainerRequest.Version, err)
```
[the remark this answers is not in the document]
csfmomo: Separate it to "containerstatus to"

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+			return
```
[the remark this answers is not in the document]
csfmomo: Sure. Let me change return to continue.

cns/restserver/internalapi.go
```diff
@@ -143,6 +146,62 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+// If NMAgent NC version got updated, CNS will refresh the pending programming IP status.
+func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMode string, syncHostNCTimeoutMilliSec time.Duration) {
+	var hostVersionNeedUpdateNcList []string
+	service.RLock()
+	for _, containerstatus := range service.state.ContainerStatus {
+		// Will open a separate PR to convert all the NC version related variable to int. Change from string to int is a pain.
+		hostVersion, err := strconv.Atoi(containerstatus.HostVersion)
+		if err != nil {
+			log.Errorf("Received err when change containerstatus.HostVersion %s to int, err msg %v", containerstatus.HostVersion, err)
+			return
```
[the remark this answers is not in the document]
csfmomo: Sure.

cns/nmagentclient/nmagentclient.go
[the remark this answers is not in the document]
csfmomo: Update it to "logger.Printf("Containers id is %s, programmed NC version is %d", ncID, versionInInt)"

cns/nmagentclient/nmagentclient.go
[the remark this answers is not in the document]
csfmomo: Update it to version.

cns/nmagentclient/nmagentclient.go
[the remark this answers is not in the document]
csfmomo: Updated it to nmaNcListResponse.ResponseCode != strconv.Itoa(http.StatusOK)

cns/nmagentclient/nmagentclient.go
```diff
@@ -149,3 +184,42 @@ func GetNmAgentSupportedApis(
...
+func (nmagentclient *NMAgentClient) GetNcVersionListWithOutToken(ncNeedUpdateList []string) map[string]int {
+	ncVersionList := make(map[string]int)
+	now := time.Now()
+	response, err := http.Get(nmagentclient.connectionURL)
+	latency := time.Since(now)
+	logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken response: %+v. QueryURL is %s, latency is %v", response, nmagentclient.connectionURL, latency)
+
+	if response.StatusCode != http.StatusOK {
+		logger.Printf("[NMAgentClient][Response] GetNcVersionListWithOutToken failed with %d, err is %v", response.StatusCode, err)
+		return nil
```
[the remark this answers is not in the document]
csfmomo: Could I keep as it is for current version? No matter what error we got, we'll keep retrying. Let me change the return value in next PR when we extend this logic to other CNS scenario. Then we need to choose different behavior when different error occur, e.g. whether keep retrying or fall back to token API

cns/restserver/internalapi.go
csfmomo: Yep.

ashvindeodhar (approved):

PR #720, opened 2020-11-04
log what ips we got and make defer pickup when no ips are found

paulgmiller:
**Reason for Change**:

We had an incident where it seems like nmagent returned an empty list of ip address.
Improve logging so we know next time

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/ipam/azure.go
+++ b/ipam/azure.go
@@ -158,7 +158,7 @@ func (s *azureSource) refresh() error {
 				continue
 			}
 
-			addesscount := 0
+			addressCount := 0
 			// For each address in the subnet...
 			for _, a := range s.IPAddress {
 				// Primary addresses are reserved for the host.
@@ -173,9 +173,9 @@ func (s *azureSource) refresh() error {
 					log.Printf("[ipam] Failed to create address:%v err:%v.", address, err)
 					continue
 				}
-				addesscount++
+				addressCount++
 			}
-			log.Printf("[ipam] got %d addresses from interface %s, subnet %s", ifName, subnet)
+			log.Printf("[ipam] got %d addresses from interface %s, subnet %s", addressCount, ifName, subnet)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


ipam/azure.go
```diff
@@ -172,7 +173,9 @@ func (s *azureSource) refresh() error {
 					log.Printf("[ipam] Failed to create address:%v err:%v.", address, err)
 					continue
 				}
+				addesscount++
 			}
+			log.Printf("[ipam] got %d addresses from interface %s, subnet %s", ifName, subnet)
```
paulgmiller, line 178: Could also reject pools with empty address lists in setaddressspace below.

ipam/azure.go
```diff
@@ -172,7 +173,9 @@ func (s *azureSource) refresh() error {
 					log.Printf("[ipam] Failed to create address:%v err:%v.", address, err)
 					continue
 				}
+				addesscount++
```
weng271190436, line 176: misspelled

ipam/pool.go
```diff
@@ -515,7 +515,8 @@ func (ap *addressPool) requestAddress(address string, options map[string]string)
 		}
 
 		if ar == nil {
-			return "", errNoAvailableAddresses
+			err = errNoAvailableAddresses
+			return "", err
```
thatmattlong, line 519: What is the purpose of this change? style?

thatmattlong (dismissed):

ipam/pool.go
paulgmiller: Not a style thing. It's only so the deferd log statement above picks it up. It was confusing looking at the logs and seeing the function returned nil, nil when that wasn't true

ipam/azure.go
weng271190436: maybe log error

weng271190436 (approved):

ipam/azure.go
paulgmiller: Punting on this since nobody asked for it.

matmerr (approved):
/azp run

matmerr (approved):

PR #725, opened 2020-11-06
Update CNS to handle updated DNC API in managed mode

ashvindeodhar:
fix: Update CNS to handle updated DNC API in managed mode

This change updates CNS code to consume the updated DNC API urls for RegisterNode and
SyncNodeNcStatus. This change also refactors the calls to DNC in dncclient package.


cns/dncclient/dncclient.go
```diff
@@ -0,0 +1,96 @@
...
+
+			response.Body.Close()
+		} else {
+			logger.Errorf("[dncclient] Failed to register node with err: %+v", err)
+		}
+
+		if sleep {
+			time.Sleep(acn.FiveSeconds)
+		}
+	}
```
ramiro-gamarra, line 69: I understand that this is code that was moved, but perhaps it's a good chance to clean it up since it is a bit hard to follow. My main recommendation would be to extract the body of the loop to a function such as:
```go
func registerNode(httpCl *http.Client, url string, body io.Reader) (cns.SetOrchestratorTypeRequest, error) {
	response, err := httpCl.Post(url, acn.JsonContent, body)
	if err != nil {...}
	defer response.Body.Close()

	if response.StatusCode != http.StatusOK {...}

	var req cns.SetOrchestratorTypeRequest
	_ = json.NewDecoder(response.Body).Decode(&req)
	return req, nil
}
```
 which then makes the loop much simpler:
```go
for {
	req, err := registerNode(httpc, registerNodeURL, &body)
	if err != nil {
		logger.Errorf(err.Error())
		time.Sleep(acn.FiveSeconds)
		continue
	}
	httpRestService.SetNodeOrchestrator(&req)
	break
}
```
To go a step further, I would consider making a proper `DNCClient` struct, which embeds reusable dependencies such as an http client. The main benefit of this is that the calling code can then use an interface and stub out the actual implementation when writing tests. That can be a later refactor though. See cnsclient in the dnc repo as an example.

cns/dncclient/dncclient.go
```diff
@@ -0,0 +1,96 @@
...
+		if response.StatusCode == http.StatusOK {
+			err = json.NewDecoder(response.Body).Decode(&nodeInfoResponse)
+		} else {
+			err = fmt.Errorf("%d", response.StatusCode)
+		}
+
+		response.Body.Close()
+	}
+
+	return nodeInfoResponse, err
```
ramiro-gamarra, line 95: prefer to return early. it's easier then to read the happy path of the code without much indentation.

ramiro-gamarra (commented):
Couple comments

cns/dncclient/dncclient.go
ashvindeodhar: done

cns/dncclient/dncclient.go
ashvindeodhar: Thanks, I have updated the logic. I will create the DNCClient struct as a part of next PR where I'll handle the msi in the dncclient.

ramiro-gamarra (approved):

cns/dncclient/dncclient.go
vakalapa: I refactored this section in #680 , can you please cross verify.

PR #732, opened 2020-11-17
NPM: bugfix: Adding port_ prefix for named ports and changing ns- to ns_

vakalapa:
Changing naming convention for IPSETs in case of namespaces and namedports

### DIFF
--- a/npm/pod_test.go
+++ b/npm/pod_test.go
@@ -70,6 +70,18 @@ func TestAddPod(t *testing.T) {
 			Phase: "Running",
 			PodIP: "1.2.3.4",
 		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				corev1.Container{
+					Ports: []corev1.ContainerPort{
+						corev1.ContainerPort{
+							Name:          "app:test-pod",
+							ContainerPort: 8080,
+						},
+					},
+				},
+			},
+		},
 	}
 
 	npMgr.Lock()
--- a/npm/translatePolicy.go
+++ b/npm/translatePolicy.go
@@ -28,10 +28,11 @@ func craftPartialIptEntrySpecFromPort(portRule networkingv1.NetworkPolicyPort, s
 	}
 
 	if portRule.Port != nil {
+		portName := "port_" + portRule.Port.String()
 		partialSpec = append(
 			partialSpec,
 			sPortOrDPortFlag,
-			portRule.Port.String(),
+			portName,
 		)
 	}
 
@@ -48,7 +49,7 @@ func craftPartialIptablesCommentFromPort(portRule networkingv1.NetworkPolicyPort
 	}
 
 	if portRule.Port != nil {
-		partialComment += "PORT-"
+		partialComment += "PORT_"
 		partialComment += portRule.Port.String()
 	}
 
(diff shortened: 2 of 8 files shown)


npm/pod_test.go
```diff
@@ -70,6 +70,18 @@ func TestAddPod(t *testing.T) {
 			Phase: "Running",
 			PodIP: "1.2.3.4",
 		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				corev1.Container{
+					Ports: []corev1.ContainerPort{
+						corev1.ContainerPort{
+							Name:          "app:test-pod",
+							ContainerPort: 8080,
```
paulgmiller, line 79: should something assert the output of this? I didn't see it.

npm/translatePolicy.go
```diff
@@ -28,10 +28,11 @@ func craftPartialIptEntrySpecFromPort(portRule networkingv1.NetworkPolicyPort, s
 	}
 
 	if portRule.Port != nil {
+		portName := "port_" + portRule.Port.String()
 		partialSpec = append(
 			partialSpec,
 			sPortOrDPortFlag,
-			portRule.Port.String(),
+			portName,
```
neaggarwMS, line 35: We can have the portRule without the Named port as well, right? @csfmomo/ @matmerr

npm/translatePolicy.go
```diff
@@ -48,7 +49,7 @@ func craftPartialIptablesCommentFromPort(portRule networkingv1.NetworkPolicyPort
 	}
 
 	if portRule.Port != nil {
-		partialComment += "PORT-"
+		partialComment += "PORT_"
```
neaggarwMS, line 52: What is this change?.

paulgmiller:
Since we're considering this for a hotfix.
1) Can we get a better description of what this is clashing with? I assume since the custoemr can name the port anything it clashes with some other standard name in ipset but an example would do wonders for anyone coming back to this. 
2) Since npm is delivered as a daemonset not as a fresh vm like cni do we need to worry about back compat? WIll the new npm code miss all the previous rules with ns- and no port_ preifix and just leave them in or is there a mechansim to clear them out? 

Less important.
Consts and helper functions might helf if we have to rename again.
Why _ vs - ? If backcompat (#2)  is important then aesthetic reasons have to take a back seat

npm/translatePolicy.go
```diff
@@ -232,7 +233,7 @@ func translateIngress(ns string, policyName string, targetSelector metav1.LabelS
 		if portRuleExists && !fromRuleExists && !allowExternal {
 			for _, portRule := range rule.Ports {
 				if portRule.Port != nil && portRule.Port.IntValue() == 0 {
-					portName := portRule.Port.String()
+					portName := "port_" + portRule.Port.String()
```
neaggarwMS, line 236: Not all Port rules have named ports. can we confirm before adding prefix to all port rules

npm/translatePolicy.go
vakalapa: To update comments in Iptable rules like below

"ALLOW-app:backend-IN-ns-testnamespace-AND-PORT_8000-TO-app:frontend-IN-ns-testnamespace"
before:
"ALLOW-app:backend-IN-ns-testnamespace-AND-PORT-8000-TO-app:frontend-IN-ns-testnamespace"

npm/translatePolicy.go
vakalapa: >portRule.Port.IntValue() == 0

This check above is identifying that port has no int value and is alphanumeric. This suggests we do not support numerical name for namedPorts.

Issue #733, opened 2020-11-17
[NPM] issue with using same string for namedPort and labelKey


vakalapa:
**What happened:**  

In some cluster, customer are seeing this below errors in azure-npm.log 
> Stderr: [exit status 1, ipset v6.34: Syntax error: Elem separator in <IP_ADDR>,<PORT_NUM>, but settype hash:net supports none.]

This error occurs when 2 different k8s objs for which NPM creates Ipsets are named the same. One such example is illustrated below. Here both the port is named as "http" and the label's key is "http". This results in NPM creating the same ipset name.
```
apiVersion: v1
kind: Pod
metadata:
  name: hello
  labels:
    http: redis
spec:
      containers:
      - name: hello
        image: busybox
        command: ['sh', '-c', 'echo "Hello, Kubernetes!" && sleep 3600']
        ports:
        - containerPort: 80
          name: http
      restartPolicy: OnFailure
```

**What you expected to happen:**  
Expected behavior is two different ipsets being created without issues.

**How to reproduce it:**  
1. Add any network policy
2. Apply above yaml and monitor NPM log.

**Kubernetes Version:**  
all

**Kernel (e.g. `uname -a`):**  
Ubuntu 18.04
Ubuntu 16.04

PR #734, opened 2020-11-18
[NPM] 🐞 {fix} error creating IPsets when same string for ports and labels is used

vakalapa:
**Reason for Change**:

NPM creates IPsets for the following today:

- NameSpace
- Label
- LabelKey:Value
- NamedPorts
- CIDR blocks

Namespace and namespace labels are getting a `ns-` prefix helping NPM to distinguish between PodLabels and NameSpaceLabels. But there are several instances where the IPSet names for Labels and NamedPorts are surfacing. This occurs when a podLabel `http:test` and a namedPort `http` is created. An example can be seen in #733 . NPM hashes `http` from label and creates a IpSet with this hash, then when it is adding namedPort `http` it would hash to the same value and it logs this below error.

>Stderr: [exit status 1, ipset v6.34: Syntax error: Elem separator in <IP_ADDR>,<PORT_NUM>, but settype hash:net supports none.]

Using ":" as prefix delimiter as k8s does not allow using a ":" in label names.

**Issue Fixed**:
#733 

**Further Enhancements**
Name clash probability is still high. Work item is added to have a robust logic while adding IpSets

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/npm/ipsm/ipsm.go
+++ b/npm/ipsm/ipsm.go
@@ -6,6 +6,7 @@ package ipsm
 import (
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
 	"syscall"
 
@@ -70,12 +71,12 @@ func (ipsMgr *IpsetManager) Exists(key string, val string, kind string) bool {
 
 // SetExists checks whehter an ipset exists.
 func (ipsMgr *IpsetManager) SetExists(setName, kind string) bool {
-    m := ipsMgr.setMap
-    if kind == util.IpsetSetListFlag {
-        m = ipsMgr.listMap
-    }
-    _, exists := m[setName]
-    return exists
+	m := ipsMgr.setMap
+	if kind == util.IpsetSetListFlag {
+		m = ipsMgr.listMap
+	}
+	_, exists := m[setName]
+	return exists
 }
 
 func isNsSet(setName string) bool {
@@ -457,6 +458,78 @@ func (ipsMgr *IpsetManager) Restore(configFile string) error {
 	cmd.Wait()
 
 	//TODO based on the set name and number of entries in the config file, update IPSetInventory
+
+	return nil
+}
+
+// DestroyNpmIpsets destroys only ipsets created by NPM
+func (ipsMgr *IpsetManager) DestroyNpmIpsets() error {
+
+	cmdName := util.Ipset
+	cmdArgs := util.IPsetCheckListFlag
+
+	reply, err := exec.Command(cmdName, cmdArgs).Output()
+	if msg, failed := err.(*exec.ExitError); failed {
+		errCode := msg.Sys().(syscall.WaitStatus).ExitStatus()
+		if errCode > 0 {
+			log.Logf("Error: There was an error running command: [%s] Stderr: [%v, %s]", cmdName, err, strings.TrimSuffix(string(msg.Stderr), "\n"))
+			metrics.SendErrorMetric(util.IpsmID, "Error: There was an error running command: [%s] Stderr: [%v, %s]", cmdName, err, strings.TrimSuffix(string(msg.Stderr), "\n"))
+		}
+
+		return err
+	}
+	if reply == nil {
+		log.Logf("Received empty string from ipset list while destroying azure-npm ipsets")
+		return nil
+	}
+
+	log.Logf("Reply from command executed is %s", reply)
+	re := regexp.MustCompile("Name: (azure-npm-\\d+)")
+	ipsetRegexSlice := re.FindAllSubmatch(reply, -1)
+
+	if len(ipsetRegexSlice) == 0 {
+		log.Logf("No Azure-NPM IPsets are found in the Node.")
+		return nil
+	}
+
+	ipsetLists := make([]string, 3)
+	for _, matchedItem := range ipsetRegexSlice {
+		if len(matchedItem) == 2 {
+			itemString := string(matchedItem[1])
+			if strings.Contains(itemString, "azure-npm") {
+				ipsetLists = append(ipsetLists, itemString)
+			}
+		}
+	}
+
+	if len(ipsetLists) == 0 {
+		return nil
+	}
+
+	entry := &ipsEntry{
+		operationFlag: util.IpsetFlushFlag,
+	}
+
+	for _, ipsetName := range ipsetLists {
+		entry := &ipsEntry{
+			operationFlag: util.IpsetFlushFlag,
+			set:           ipsetName,
+		}
+
+		if _, err := ipsMgr.Run(entry); err != nil {
+			metrics.SendErrorMetric(util.IpsmID, "Error: failed to flush ipset %s", ipsetName)
+			log.Logf("Error: failed to flush ipset %s", ipsetName)
+		}
+	}
+
+	for _, ipsetName := range ipsetLists {
+		entry.operationFlag = util.IpsetDestroyFlag
+		entry.set = ipsetName
+		if _, err := ipsMgr.Run(entry); err != nil {
+			metrics.SendErrorMetric(util.IpsmID, "Error: failed to destroy ipset %s", ipsetName)
+			log.Logf("Error: failed to destroy ipset %s", ipsetName)
+		}
+	}
 
 	return nil
 }
--- a/npm/ipsm/ipsm.go
+++ b/npm/ipsm/ipsm.go
@@ -495,7 +495,7 @@ func (ipsMgr *IpsetManager) DestroyNpmIpsets() error {
 	for _, matchedItem := range ipsetRegexSlice {
 		if len(matchedItem) == 2 {
 			itemString := string(matchedItem[1])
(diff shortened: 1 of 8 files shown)


npm/pod_test.go
```diff
@@ -70,12 +70,27 @@ func TestAddPod(t *testing.T) {
...
+				},
+			},
+		},
 	}
 
 	npMgr.Lock()
 	if err := npMgr.AddPod(podObj); err != nil {
 		t.Errorf("TestAddPod failed @ AddPod")
 	}
+	if !ipsMgr.Exists(util.GetHashedName("app:test-pod"), "1.2.3.4,8080", "") {
```
csfmomo, line 91: Will it be port:app:test-pod?

vakalapa: Yes thank you.

npm/npm.go
```diff
@@ -188,6 +189,12 @@ func NewNetworkPolicyManager(clientset *kubernetes.Clientset, informerFactory in
 	iptMgr := iptm.NewIptablesManager()
 	iptMgr.UninitNpmChains()
 
+	log.Logf("Azure-NPM creating, cleaning existing IPSets")
+	destroyErr := ipsm.NewIpsetManager().Destroy()
```
neaggarwMS, line 193: Wont this cleanup all the ipsets from that node? We should only clean the ones created by Azure-NPM. Also ignore the error as this is best effort to clean the old ipsets.

npm/util/const.go
```diff
@@ -109,6 +109,9 @@ const (
 	IpsetMaxelemNum  string = "4294967295"
 
 	IpsetNomatch string = "nomatch"
+
+	//Prefixes for ipsets
+	NamedPortIPSetPrefix string = "port:"
```
neaggarwMS, line 114: Can we rename it to portname (or namedport):, port seems very common

npm/translatePolicy.go
```diff
@@ -936,7 +936,7 @@ func translateEgress(ns string, policyName string, targetSelector metav1.LabelSe
 					if portRuleExists {
 						for _, portRule := range rule.Ports {
 							if portRule.Port != nil && portRule.Port.IntValue() == 0 {
-								portName := portRule.Port.String()
+								portName := util.NamedPortIPSetPrefix + portRule.Port.String()
```
neaggarwMS, line 939: Can we confirm this? We can still have a port rule without a named port, right?

neaggarwMS: Okay checked the logic, all port strings are getting appending to Namedports and in the end we call util.DropEmptyFields(namedPorts) (Line: 1438). This is weird. But with the updated logic empty port strings will not be dropped. We should only add if portRule.port.String() is not empty. This should have been the logic at first place instead of double parsing :-/

neaggarwMS: We should still add a test with policy which has port rules (without named ports) and one without

npm/npm.go
vakalapa: As we discussed offline, NPM is not saving state any state while updating, to avoid any leak of ipsets, we will destroy everything. I logged a work item to add state fullness to this logic.

npm/translatePolicy.go
vakalapa: NPM is allowing this below network policy, corrected the behavior now.

```
  ingress:
    - from:
        - podSelector:
            matchLabels:
              app: backend
      ports:
        - port: ""
```

npm/npm.go
matmerr: Agree with Neha here, an ipset flush wipes out all ipsets, including ones NPM may have not created, this is where we need to iterate through all ipsets that contain azure-npm

vakalapa: Added support to just delete azure-npm-(numerics) ipsets and lists

npm/ipsm/ipsm.go
```diff
@@ -459,4 +460,76 @@ func (ipsMgr *IpsetManager) Restore(configFile string) error {
...
+			metrics.SendErrorMetric(util.IpsmID, "Error: failed to flush ipset %s", ipsetName)
+			log.Logf("Error: failed to flush ipset %s", ipsetName)
+		}
+	}
+
+	for _, ipsetName := range ipsetLists {
+		entry.operationFlag = util.IpsetDestroyFlag
+		entry.set = ipsetName
+		if _, err := ipsMgr.Run(entry); err != nil {
+			metrics.SendErrorMetric(util.IpsmID, "Error: failed to destroy ipset %s", ipsetName)
```
neaggarwMS, line 529: Can you add the FunctionName in the log statement for better debugability: DestroyNpmIpsets (We dont print the FileName/Line# in the Geneva output, although we should.)

npm/ipsm/ipsm.go
```diff
@@ -459,4 +460,76 @@ func (ipsMgr *IpsetManager) Restore(configFile string) error {
...
+
+	for _, ipsetName := range ipsetLists {
+		entry := &ipsEntry{
+			operationFlag: util.IpsetFlushFlag,
+			set:           ipsetName,
+		}
+
+		if _, err := ipsMgr.Run(entry); err != nil {
+			metrics.SendErrorMetric(util.IpsmID, "Error: failed to flush ipset %s", ipsetName)
+			log.Logf("Error: failed to flush ipset %s", ipsetName)
```
neaggarwMS, line 521: SendErrorMetric also logs the error. Actually that is an incorrect functionName it should be SendErrorLog or SendLog

npm/ipsm/ipsm.go
```diff
@@ -459,4 +460,76 @@ func (ipsMgr *IpsetManager) Restore(configFile string) error {
...
+		}
+	}
+
+	for _, ipsetName := range ipsetLists {
+		entry.operationFlag = util.IpsetDestroyFlag
+		entry.set = ipsetName
+		if _, err := ipsMgr.Run(entry); err != nil {
+			metrics.SendErrorMetric(util.IpsmID, "Error: failed to destroy ipset %s", ipsetName)
+			log.Logf("Error: failed to destroy ipset %s", ipsetName)
```
neaggarwMS, line 530: Again remove the duplicate logF line

npm/translatePolicy.go
```diff
@@ -38,6 +38,19 @@ func craftPartialIptEntrySpecFromPort(portRule networkingv1.NetworkPolicyPort, s
 	return partialSpec
 }
 
+func getPortType(portRule networkingv1.NetworkPolicyPort) string {
+	if portRule.Port == nil {
```
neaggarwMS, line 42: nit: We can combine 1st and last check 
if portRule.Port == nil || portRule.Port.IntValue() != 0 
{ 
    return "validport" 
} else ...

neaggarwMS (commented):
Left minor comments, rest looks good.

npm/npm.go
```diff
@@ -188,6 +189,12 @@ func NewNetworkPolicyManager(clientset *kubernetes.Clientset, informerFactory in
 	iptMgr := iptm.NewIptablesManager()
 	iptMgr.UninitNpmChains()
 
+	log.Logf("Azure-NPM creating, cleaning existing Azure NPM IPSets")
+	destroyErr := ipsm.NewIpsetManager().DestroyNpmIpsets()
+	if destroyErr != nil {
+		log.Logf("Azure-NPM error occurred while destroying existing IPSets err: %s", destroyErr.Error())
```
csfmomo, line 195: nit, wondering, if DestroyNpmIpsets did metrics.SendErrorMetric inside its function, whether do we want to log here again.

npm/ipsm/ipsm.go
```diff
@@ -459,4 +460,73 @@ func (ipsMgr *IpsetManager) Restore(configFile string) error {
...
+	if len(ipsetRegexSlice) == 0 {
+		log.Logf("No Azure-NPM IPsets are found in the Node.")
+		return nil
+	}
+
+	ipsetLists := make([]string, 3)
+	for _, matchedItem := range ipsetRegexSlice {
+		if len(matchedItem) == 2 {
+			itemString := string(matchedItem[1])
+			if strings.Contains(itemString, "azure-npm") {
```
csfmomo, line 498: Do we want to use a const for "azure-npm"

csfmomo (approved):

csfmomo (approved):

npm/translatePolicy.go
```diff
@@ -38,6 +38,15 @@ func craftPartialIptEntrySpecFromPort(portRule networkingv1.NetworkPolicyPort, s
 	return partialSpec
 }
 
+func getPortType(portRule networkingv1.NetworkPolicyPort) string {
+	if portRule.Port == nil || portRule.Port.IntValue() != 0 {
+		return "validport"
```
matmerr, line 43: can we make these consts in case we end up using this elsewhere?

neaggarwMS (dismissed):

csfmomo (approved):

PR #738, opened 2020-11-24
Configurable wireserver endpoint

ramiro-gamarra:
**Reason for Change**:

This PR adds a configuration option for wireserver ip, helpful in cases where we are not in a live environment and need to use mocks for nmagent. It also adds a check in the publish container endpoint to possibly extract the wireserver endpoint from the incoming request; given that the rest of the payload contains urls used to communicate to wireserver, it follows that we should try to use that instead of the internal cns configuration.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/configuration/cns_config.json
+++ b/cns/configuration/cns_config.json
@@ -18,5 +18,6 @@
     "UseHTTPS" : false,
     "TLSSubjectName" : "",
     "TLSCertificatePath" : "",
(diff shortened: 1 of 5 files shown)


thatmattlong (approved):
lgtm

PR #750, opened 2020-12-12
Ensure pending programming IPs will be released before available IPs when scale down.

csfmomo:
Improvement: Ensure pending programming IPs will be released first when scale down.
**Reason for Change**:
When scale down, we want to ensure pending programming IPs can be released before available IPs to improve performance. Pending programming IPs need to check programmed NC version before available. Release pending programming IPs will save one step.

**Issue Fixed**:
Improve performance. 

- [X ] uses [conventional commit messages](https://www.conventionalcommits.org/)
        perf: Performance Improvements 🚀
- [X ] includes documentation
- [X] adds unit tests

### DIFF
--- a/cns/restserver/ipam.go
+++ b/cns/restserver/ipam.go
@@ -92,33 +92,35 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r 
 	return
 }
 
-// MarkIPsAsPending will mark IPs to pending release state.
-func (service *HTTPRestService) MarkIPsAsPending(numberToMark int) (map[string]cns.IPConfigurationStatus, error) {
-	pendingReleaseIPs := make(map[string]cns.IPConfigurationStatus)
+// MarkIPAsPendingRelease will mark IPs to pending release state.
+func (service *HTTPRestService) MarkIPAsPendingRelease(numberToMark int) (map[string]cns.IPConfigurationStatus, error) {
+	allReleasedIPs := make(map[string]cns.IPConfigurationStatus)
 	// Ensure PendingProgramming IPs will be release before Available ones.
 	ipStateTypes := [2]string{cns.PendingProgramming, cns.Available}
 
 	service.Lock()
 	defer service.Unlock()
 	for _, ipStateType := range ipStateTypes {
-		pendingReleaseIPs := service.markSepcificIPTypeAsPending(numberToMark, len(pendingReleaseIPs), ipStateType, pendingReleaseIPs)
-		alreadyMarkedNum := len(pendingReleaseIPs)
-		if alreadyMarkedNum == numberToMark {
-			return pendingReleaseIPs, nil
+		pendingReleaseIPs := service.markSpecificIPTypeAsPending(numberToMark, ipStateType)
+		for uuid, pependingReleaseIP := range pendingReleaseIPs {
+			allReleasedIPs[uuid] = pependingReleaseIP
+		}
+		numberToMark -= len(pendingReleaseIPs)
+		if numberToMark == 0 {
+			return allReleasedIPs, nil
 		}
 	}
-	return nil, fmt.Errorf("Failed to mark %d IP's as pending, only marked %d IP's", numberToMark, len(pendingReleaseIPs))
+	return nil, fmt.Errorf("Failed to mark %d IP's as pending, only marked %d IP's", numberToMark, len(allReleasedIPs))
 }
 
-func (service *HTTPRestService) markSepcificIPTypeAsPending(numberToMark, markedIPCount int, ipStateType string, pendingReleaseIPs map[string]cns.IPConfigurationStatus) map[string]cns.IPConfigurationStatus {
-	for uuid := range service.PodIPConfigState {
-		mutableIPConfig := service.PodIPConfigState[uuid]
+func (service *HTTPRestService) markSpecificIPTypeAsPending(numberToMark int, ipStateType string) map[string]cns.IPConfigurationStatus {
+	pendingReleaseIPs := make(map[string]cns.IPConfigurationStatus)
+	for uuid, mutableIPConfig := range service.PodIPConfigState {
 		if mutableIPConfig.State == ipStateType {
 			mutableIPConfig.State = cns.PendingRelease
 			service.PodIPConfigState[uuid] = mutableIPConfig
 			pendingReleaseIPs[uuid] = mutableIPConfig
-			markedIPCount++
-			if markedIPCount == numberToMark {
+			if len(pendingReleaseIPs) == numberToMark {
 				return pendingReleaseIPs
 			}
 		}
--- a/cns/restserver/ipam.go
+++ b/cns/restserver/ipam.go
@@ -92,26 +92,38 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r 
 	return
 }
 
+// MarkIPsAsPending will mark IPs to pending release state.
 func (service *HTTPRestService) MarkIPsAsPending(numberToMark int) (map[string]cns.IPConfigurationStatus, error) {
 	pendingReleaseIPs := make(map[string]cns.IPConfigurationStatus)
-	markedIPCount := 0
+	// Ensure PendingProgramming IPs will be release before Available ones.
+	ipStateTypes := [2]string{cns.PendingProgramming, cns.Available}
 
 	service.Lock()
 	defer service.Unlock()
-	for uuid, _ := range service.PodIPConfigState {
+	for _, ipStateType := range ipStateTypes {
+		pendingReleaseIPs := service.markSepcificIPTypeAsPending(numberToMark, len(pendingReleaseIPs), ipStateType, pendingReleaseIPs)
+		alreadyMarkedNum := len(pendingReleaseIPs)
+		if alreadyMarkedNum == numberToMark {
+			return pendingReleaseIPs, nil
+		}
+	}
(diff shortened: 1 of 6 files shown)


neaggarwMS (changes requested):

cns/restserver/ipam.go
```diff
@@ -92,26 +92,38 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r
...
 	pendingReleaseIPs := make(map[string]cns.IPConfigurationStatus)
-	markedIPCount := 0
+	// Ensure PendingProgramming IPs will be release before Available ones.
+	ipStateTypes := [2]string{cns.PendingProgramming, cns.Available}
 
 	service.Lock()
 	defer service.Unlock()
-	for uuid, _ := range service.PodIPConfigState {
+	for _, ipStateType := range ipStateTypes {
+		pendingReleaseIPs := service.markSepcificIPTypeAsPending(numberToMark, len(pendingReleaseIPs), ipStateType, pendingReleaseIPs)
```
ramiro-gamarra, line 104: there's a possible bug here. the `pendingReleaseIPs` variable is redeclared in this scope, so what executes on every iteration is equivalent to
```go
service.markSpecificIPTypeAsPending(numberToMark, 0, ipStateType, /*empty map*/)
```

cns/restserver/ipam.go
```diff
@@ -92,26 +92,38 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r
...
+		alreadyMarkedNum := len(pendingReleaseIPs)
+		if alreadyMarkedNum == numberToMark {
+			return pendingReleaseIPs, nil
+		}
+	}
+	return nil, fmt.Errorf("Failed to mark %d IP's as pending, only marked %d IP's", numberToMark, len(pendingReleaseIPs))
+}
+
+func (service *HTTPRestService) markSepcificIPTypeAsPending(numberToMark, markedIPCount int, ipStateType string, pendingReleaseIPs map[string]cns.IPConfigurationStatus) map[string]cns.IPConfigurationStatus {
```
ramiro-gamarra, line 113: - typo on function name
- function is a bit confusing. I think it would be simpler to just take in a count (number to mark) and ip state type. The returned map can be initialized locally, no need to accept it as a parameter.

cns/restserver/ipam.go
```diff
@@ -92,26 +92,38 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r
...
+	for uuid := range service.PodIPConfigState {
 		mutableIPConfig := service.PodIPConfigState[uuid]
```
ramiro-gamarra, line 115: no need to fetch value again, you can just range over the map like 
```go
for uuid, mutableIPConfig := range service.PodIPConfigState
```

ramiro-gamarra (commented):
Some comments

cns/restserver/ipam.go
```diff
@@ -92,26 +92,38 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r
 	return
 }
 
+// MarkIPsAsPending will mark IPs to pending release state.
 func (service *HTTPRestService) MarkIPsAsPending(numberToMark int) (map[string]cns.IPConfigurationStatus, error) {
```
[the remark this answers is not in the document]
csfmomo: Sure

cns/restserver/ipam.go
```diff
@@ -92,26 +92,38 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r
...
 	defer service.Unlock()
-	for uuid, _ := range service.PodIPConfigState {
+	for _, ipStateType := range ipStateTypes {
+		pendingReleaseIPs := service.markSepcificIPTypeAsPending(numberToMark, len(pendingReleaseIPs), ipStateType, pendingReleaseIPs)
+		alreadyMarkedNum := len(pendingReleaseIPs)
+		if alreadyMarkedNum == numberToMark {
+			return pendingReleaseIPs, nil
+		}
+	}
+	return nil, fmt.Errorf("Failed to mark %d IP's as pending, only marked %d IP's", numberToMark, len(pendingReleaseIPs))
```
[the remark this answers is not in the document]
csfmomo: Sure, test added.
	// Try to release IP when no IP can be released. It should return error and ips will be nil
	ips, err = svc.MarkIPAsPendingRelease(1)
	if err == nil || ips != nil {
		t.Fatalf("Unexpected failure releasing IP: %+v", err)
	}

cns/restserver/util.go
```diff
@@ -276,6 +276,8 @@ func (service *HTTPRestService) addIPConfigStateUntransacted(ncId string, hostVe
 		// When reconcile, service.PodIPConfigState doens't exist, rebuild it with the help of NC version attached with IP.
 		if hostVersion < ipconfig.NCVersion {
 			newIPCNSStatus = cns.PendingProgramming
+		} else {
+			newIPCNSStatus = cns.Available
```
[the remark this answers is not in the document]
csfmomo: Sure

cns/restserver/ipam.go
csfmomo: Modified to avoid redeclaration.

cns/restserver/ipam.go
csfmomo: Fixed typo.
Removed pendingReleaseIPs  as parameter.

cns/restserver/ipam.go
csfmomo: Sure.

cns/restserver/util.go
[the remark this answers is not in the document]
ramiro-gamarra: the newIPCNSStatus variable usage is very short lived, it can be declared right above line 277, no need for it to live outside the loop

cns/restserver/ipam.go
```diff
@@ -92,26 +92,38 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r
...
 	service.Lock()
 	defer service.Unlock()
-	for uuid, _ := range service.PodIPConfigState {
-		mutableIPConfig := service.PodIPConfigState[uuid]
-		if mutableIPConfig.State == cns.Available || mutableIPConfig.State == cns.PendingProgramming {
+	for _, ipStateType := range ipStateTypes {
+		pendingReleaseIPs = service.markSpecificIPTypeAsPending(numberToMark, len(pendingReleaseIPs), ipStateType)
+		alreadyMarkedNum := len(pendingReleaseIPs)
+		if alreadyMarkedNum == numberToMark {
+			return pendingReleaseIPs, nil
```
ramiro-gamarra, line 107: I think this is still not correct. on the second iteration you are just returning the ips that went from available to pending. the previous ones got discarded. I would suggest the following structure for clarity

```go
allIPs := make(map[string]cns.IPConfigurationStatus)

for _, ipStateType := range ipStateTypes {
    releasedIPs := service.markSpecificIPTypeAsPending(numberToMark, ipStateType) // no need to send the second argument, just determine what the function needs from this scope
    /* 
        add releasedIPs to allIPs here
    */
    numberToMark -= len(releasedIPs) // determine the count for next iteration
    if numberToMark == 0 {
        return allIPs 
    }
}
```

csfmomo: It works when we only have 2 ipStateTypes but not work when state type is over 2. Sure, I can change to calculate the number to mark independently.

ramiro-gamarra: the problem is that the function call to `service.markSpecificIPTypeAsPending` returns a **new** map, so each assignment to `pendingReleaseIPs` overwrites the previous one. the second iteration will return the second allocated map.

cns/restserver/ipam.go
```diff
@@ -92,26 +92,39 @@ func (service *HTTPRestService) releaseIPConfigHandler(w http.ResponseWriter, r
...
+		if numberToMark == 0 {
+			return pendingReleaseIPs, nil
+		}
+	}
+	return nil, fmt.Errorf("Failed to mark %d IP's as pending, only marked %d IP's", numberToMark, len(pendingReleaseIPs))
+}
+
+func (service *HTTPRestService) markSpecificIPTypeAsPending(numberToMark int, ipStateType string) map[string]cns.IPConfigurationStatus {
+	pendingReleaseIPs := make(map[string]cns.IPConfigurationStatus)
+	markedIPCount := 0
```
ramiro-gamarra, line 115: no need for this anymore, just use the length of the map

cns/restserver/ipam.go
csfmomo: Yep I know but I think the current implementation already apvid it.

cns/restserver/ipam_test.go
```diff
@@ -575,6 +579,93 @@ func TestIPAMMarkIPCountAsPending(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unexpected failure releasing IP: %+v", err)
 	}
+
+	// Try to release IP when no IP can be released. It should return error and ips will be nil
+	ips, err = svc.MarkIPAsPendingRelease(1)
+	if err == nil || ips != nil {
+		t.Fatalf("Unexpected failure releasing IP: %+v", err)
```
ramiro-gamarra, line 586: I think the message should be different. If the error was nil, we should print that an error was expected. If there were ips returned, we should print that 0 ips were expected

ramiro-gamarra (approved):
lgtm

PR #755, opened 2020-12-16
Write to intermediate file before moving to state file

tamilmani1989:
**Reason for Change**:

Write to intermediate file before moving to state file

**Issue Fixed**:

IP leak fixes
Ipam pool fixes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


store/json.go
```diff
@@ -109,20 +112,42 @@ func (kvs *jsonFileStore) Flush() error {
...
+
+	defer func(){
+		if err != nil {
+			// remove temp file after job is done
+			_ = os.Remove(tmpFileName)
+			f.Close()
+		}
+	}()
+
+	if err = ioutil.WriteFile(tmpFileName, buf, 0); err != nil {
```
ramiro-gamarra, line 140: the temp file handle should already be good for writing, so you can just do 

```go
if err := f.Write(buf); err != nil {
   //...
}
```

store/json.go
```diff
@@ -109,20 +112,42 @@ func (kvs *jsonFileStore) Flush() error {
...
+		return fmt.Errorf("ioutil.WriteFile failed with: %v", err)
 	}
+
+	f.Close()
```
ramiro-gamarra, line 144: check error?

store/json.go
```diff
@@ -109,20 +112,42 @@ func (kvs *jsonFileStore) Flush() error {
...
+
+	log.Printf("renaming tmp file %v to state file", tmpFileName)
+	if err = os.Rename(tmpFileName, kvs.fileName); err != nil {
```
ramiro-gamarra, line 147: The current stdlib implementation seems to have improved since the time the `atomic` package was created, but I took a closer look and one difference is the [usage of the MOVEFILE_WRITE_THROUGH flag on the syscall](https://github.com/natefinch/atomic/blob/master/file_windows.go#L29). It's [documentation](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexw#parameters) states:


MOVEFILE_WRITE_THROUGH8 (0x8) | The function does not return until the file is actually moved on the disk.Setting this value guarantees that a move performed as a copy and delete operation is flushed to disk before the function returns. The flush occurs at the end of the copy operation.
-- | --

So I'm wondering if this is really atomic. Thoughts?

ramiro-gamarra (commented):
Some comments on the file flush, I'm not familiar with the other changes

store/json.go
tamilmani1989: let me check what os.rename implementation is

tamilmani1989: looks like os.replace uses MOVEFILE_REPLACE_EXISTING

platform/os.go
```diff
@@ -0,0 +1,70 @@
...
+	_, err := os.Stat(filepath)
+	if err == nil {
+		return true, nil
+	}
+
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+
+	return true, err
```
ramiro-gamarra, line 53: I know this is code moved from somewhere else, but should this be `false, err` instead? Although the client should be checking the error first, its common to return the zero value for all other returns other than the error.

platform/os.go
```diff
@@ -0,0 +1,70 @@
...
+
+func CreateDirectory(dirPath string) error {
+	var err error
+
+	if dirPath == "" {
+		log.Printf("dirPath is empty, nothing to create.")
+		return nil
+	}
+
+	isExist, _ := CheckIfFileExists(dirPath)
```
ramiro-gamarra, line 64: check error?

platform/os.go
tamilmani1989: hmm may be what you saying is correct but i dont want to break anything by changing this.

ramiro-gamarra: fair enough :)

platform/os.go
tamilmani1989: same here...i can print err but i cannot validate error

tamilmani1989: checked this link: https://stackoverflow.com/questions/37932551/mkdir-if-not-exists-using-golang . let me validate error also and createdirectory

matmerr (approved):
followed up offline, approving invoker changes

ramiro-gamarra (approved):

AbelHu:
@tamilmani1989 @ramiro-gamarra @matmerr When will the new Azure CNI version be released? One AKS Windows user hit the issue 3 times in one week that the Azure CNI file is invalid.

zhiweiv:
@AbelHu 
Today 1.2.2 released with this fix, can you merge this to AKS asap, the **invalid azure cni state file** issue still occur frequently in our clusters, ip leak issue also occur several times.

We'd like to verify if this fix works.

AbelHu:
@zhiweiv We are working on this. We will try our best to update it in AKS ASAP.

AbelHu:
@zhiweiv azure cni v1.2.2 should be available in Azure global regions now. You can upgrade it by upgrading cluster or updating node images of Windows agent pools.

zhiweiv:
Upgraded the problem AKS to lastest node image, it is azure cni 1.2.2 now. I will keep monitoring this issue.

PR #758, opened 2020-12-18
Makefile windows

pjohnst5:
**Reason for Change**:
Cross-platform makefile capabilities, to allow the possibility of running make targets on windows

Also cleaning up the Makefile and getting rid of some unused scripts


rbtr (approved):

pjohnst5:
Had to fix something, I wrote a specific clause for windows platforms, and set the `EXT` variable there, but that `EXT` variable should only be set if the `GOOS` is windows

rbtr (dismissed):

pjohnst5:
Something's wrong with this (zipping up windows files is corrupted) working on it

rbtr (approved):

PR #761, opened 2020-12-25
write the files in one transaction.

andyliuliming:
our customer(AKS) sometimes meet the issue that the json file is incomplete.
and we need to restart the services in both linux/windows to mitigate the issue.
I think put the file writing into one "transaction" can help our customer resolve that issue.


andyliuliming:
ok.. looks like dup with
https://github.com/Azure/azure-container-networking/pull/755/files

Issue #713, opened 2020-10-29
[CNI] Worker nodes is fetching only first DNS server IP from custom DNS list


vakalapa:
**What happened:** 
There are 2 kinds of custom DNS server settings in Azure, 
1.	From the Azure VNET custom DNS servers list. This will show up as Global DNS name servers in system-resolv like below and CNI is NOT reading this information.
```root@k8s-agentpool1-111111-0:/# systemd-resolve --status
Global
     DNS Servers: 168.63.129.16
                  8.8.8.8
                  1.1.1.1
     DNS Domain: reddog.microsoft.com
     DNSSEC NTA: 10.in-addr.arpa 
          .....
```

2.	From the Azure VM Network interface custom DNS servers list. This will show up in eth0 systemd-resolve, even though CNI is reading this info, it is expecting in a certain format than what system-resolv is printing.
```
Link 2 (eth0)
       Current Scopes: none
       LLMNR setting: yes
       MulticastDNS setting: no
       DNSSEC setting: no
       DNSSEC supported: no
       DNS Servers: 168.63.129.16
                    8.8.8.8
                    1.1.1.1
        DNS Domain: reddog.microsoft.com
```
CNI is expecting the DNS servers to be printed as below:
```
DNS Servers:   168.63.129.16
DNS Servers:   8.8.8.8
DNS Servers:   1.1.1.1
```
This section needs to be updated:
https://github.com/Azure/azure-container-networking/blob/6fad74d0152b6489755bcd1329b1bb0bfc45c6d5/network/network_linux.go#L213-L226

**What you expected to happen:**  
Expected behavior is for Azure CNI to read Global DNS servers list & eth0 DNS servers list and configure azure0 with them.

**How to reproduce it:**  
1. Go to worker node Azure VNET ->Settings -> DNS Servers:
    Edit the option from "Default (Azure-provided)" to "Custom" and add additional DNS servers.
    Worker node will need to be restarted for this change to get applied.

    After the reboot, "systemd-resolve --status|grep 'DNS Servers' -A4 -B4" will show only the first DNS server from above change being applied

**Orchestrator and Version (e.g. Kubernetes, Docker):**  
Kubernetes

**Operating System (Linux/Windows):**  
Linux

**Kernel (e.g. `uanme -a` for Linux or `$(Get-ItemProperty -Path "C:\windows\system32\hal.dll").VersionInfo.FileVersion` for Windows):**
v5.4

Kenneth-Abrams:
[...]

PR #769, opened 2021-01-27
fix for configuring multiple dns server on azure0 interface

tamilmani1989:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -231,14 +231,25 @@ func readDnsInfo(ifName string) (DNSInfo, error) {
 		return dnsInfo, fmt.Errorf("[net] Console output doesn't have any lines")
 	}
 
+	dnsServerFound := false
 	for _, line := range lineArr {
 		if strings.Contains(line, dnsServersStr) {
 			dnsServerSplit := strings.Split(line, colonDelimiter)
 			if len(dnsServerSplit) > 1 {
+				dnsServerFound = true
 				dnsServerSplit[1] = strings.TrimSpace(dnsServerSplit[1])
 				dnsInfo.Servers = append(dnsInfo.Servers, dnsServerSplit[1])
 			}
-		} else if strings.Contains(line, dnsDomainStr) {
+		} else if !strings.Contains(line, colonDelimiter) && dnsServerFound {
+			dnsServer := strings.TrimSpace(line)
+			dnsInfo.Servers = append(dnsInfo.Servers, dnsServer)
+		} else {
+			dnsServerFound = false
+		}
+	}
+
+	for _, line := range lineArr {
+		if strings.Contains(line, dnsDomainStr) {
 			dnsDomainSplit := strings.Split(line, colonDelimiter)
 			if len(dnsDomainSplit) > 1 {
 				dnsInfo.Suffix = strings.TrimSpace(dnsDomainSplit[1])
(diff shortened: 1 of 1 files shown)


network/network_linux.go
```diff
@@ -292,10 +303,23 @@ func (nm *networkManager) applyIPConfig(extIf *externalInterface, targetIf *net.
...
-	var err error
+	var (
+		setDnsList string
+		err        error
+	)
+
+	if extIf != nil {
+		for _, server := range extIf.DNSInfo.Servers {
+			if net.ParseIP(server).To4() == nil {
+				log.Errorf("invalid dns ip %s. Shouldn't happen", server)
```
thatmattlong, line 314: "Shouldn't happen" 😆

thatmattlong (approved):
I would like to see a test that fails on the prev code and passes on the new code, but this looks good to me otherwise

tamilmani1989:
> I would like to see a test that fails on the prev code and passes on the new code, but this looks good to me otherwise

I would love to but the problem is that api relies on system cmd "systemd-resolve" which takes multiple dns addresses. The platform we unit test may not have this cmd installed. if we redesign the code to define interfaces, then we can unit test atleast the data is in expected format. I think paul @pjohnst5  fundamental work would cover that.

network/network_linux.go
tamilmani1989: i will update the log

network/network_linux.go
```diff
@@ -231,14 +231,25 @@ func readDnsInfo(ifName string) (DNSInfo, error) {
 		return dnsInfo, fmt.Errorf("[net] Console output doesn't have any lines")
 	}
 
+	dnsServerFound := false
 	for _, line := range lineArr {
 		if strings.Contains(line, dnsServersStr) {
```
vakalapa, line 236: There are two different kind of custom DNS server list which users can add in Azure refer #713 .  I believe we are getting only the eth0 (or whichever interface) and adding them to azure0, do we expect global DNS servers to still work fine ? since they Global would not worry about change in interface properties,

network/network_linux.go
```diff
@@ -292,10 +303,23 @@ func (nm *networkManager) applyIPConfig(extIf *externalInterface, targetIf *net.
...
 func applyDnsConfig(extIf *externalInterface, ifName string) error {
-	var err error
+	var (
+		setDnsList string
+		err        error
+	)
+
+	if extIf != nil {
+		for _, server := range extIf.DNSInfo.Servers {
+			if net.ParseIP(server).To4() == nil {
```
vakalapa, line 313: I think Azure does support ipv6 DNS servers, is that true for VM DNS servers ? [Azure-DNS-Ipv6](https://docs.microsoft.com/en-us/azure/dns/dns-faq#do-azure-dns-name-servers-resolve-over-ipv6)

network/network_linux.go
tamilmani1989: not sure how to configure global dnsserver in portal. If it's not tied to nic, then we dont need to worry about that.

network/network_linux.go
tamilmani1989: imo can't add ipv6 dns server in v4 vnet. we can consider this when we support dualstack

network/network_linux.go
tamilmani1989: i veriifed updating vnet custom dnsserver list and it worked

vakalapa: Can you add a comment on the missing scenario, where if the NIC DNS server list is updated after the VM and CNI is setup then the VM will need a reboot for them to apply. VNET level (global) DNS servers should get updated even after the VM and CNI are setup i guess (but it will be good to verify).

tamilmani1989: if dns server is updated, vm needs a reboot in any case. It doesn't matter if cni is setup or not. I will add a comment though

vakalapa (approved):
lgtm

PR #770, opened 2021-01-27
[Draft] tag ip tables rules with the network policy that created them

paulgmiller:
**Reason for Change**:
IP Tables are hard to debug!
Tell us what policy resulted in the iptable line we're looking at. Will let us debug easier.

Could also drop a count entries added per policy. 

Only done for ingress and this function could use some refactoring

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/translatePolicy.go
+++ b/npm/translatePolicy.go
@@ -189,6 +189,7 @@ func translateIngress(ns string, policyName string, targetSelector metav1.LabelS
 		fromRuleExists := false
 		addedPortEntry = addedPortEntry || portRuleExists
 		ipCidrs[i] = make([]string, len(rule.From))
+		comment := ns + "/" + policyName
 
 		if rule.From != nil {
 			if len(rule.From) == 0 {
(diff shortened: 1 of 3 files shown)


npm/translatePolicy.go
```diff
@@ -189,6 +189,7 @@ func translateIngress(ns string, policyName string, targetSelector metav1.LabelS
 		fromRuleExists := false
 		addedPortEntry = addedPortEntry || portRuleExists
 		ipCidrs[i] = make([]string, len(rule.From))
+		comment := ns + "/" + policyName
```
matmerr, line 192: not aware of a delimiter after policyName, we should potentially add some separation for legibility

comment might read as "ns/policynameALLOW-ALL-TO-targetselectorComment"

npm/translatePolicy.go
```diff
@@ -221,6 +222,7 @@ func translateIngress(ns string, policyName string, targetSelector metav1.LabelS
 				util.IptablesSrcFlag,
 			)
 			entry.Specs = append(entry.Specs, targetSelectorIptEntrySpec...)
+			comment += "ALLOW-ALL-TO-" + targetSelectorComment + "-FROM-" + util.KubeAllNamespacesFlag
```
matmerr, line 225: I believe the max length for an iptables comment is 256, kubernetes objects have a name limit of 253, we should probably have some bounds checking when concatenating more strings that could be 253 chars long

PR #773, opened 2021-01-29
[NPM] Supporting Namespace label updates

vakalapa:
NPM currently deletes all the ipsets when an existing namespace is updated. This fix selectively deletes or adds only the changed labels for an existing Namespace.

### DIFF
--- a/npm/namespace_test.go
+++ b/npm/namespace_test.go
@@ -135,6 +135,114 @@ func TestUpdateNamespace(t *testing.T) {
 	npMgr.Unlock()
 }
 
+func TestAddNamespaceLabel(t *testing.T) {
+	npMgr := &NetworkPolicyManager{
+		nsMap:            make(map[string]*namespace),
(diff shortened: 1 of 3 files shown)


npm/namespace_test.go
```diff
@@ -135,6 +135,114 @@ func TestUpdateNamespace(t *testing.T) {
...
+			Labels: map[string]string{
+				"app":    "old-test-namespace",
+				"update": "true",
+			},
+		},
+	}
+
+	npMgr.Lock()
+	if err := npMgr.AddNamespace(oldNsObj); err != nil {
+		t.Errorf("TestAddNamespaceLabel failed @ npMgr.AddNamespace")
```
csfmomo, line 182: PR looks good to me. For this test, is there anyway that we check the new namespace label has been added besides no error happens?

vakalapa: Fixed this.

csfmomo (approved):

PR #775, opened 2021-01-29
Fix for deleting hostvethpair in transparent mode

tamilmani1989:
…pace is deleted

**Reason for Change**:

The issue is that in transparent mode cni generates the veth pair names based on pod name and namespace. In this case, a new container with the same pod name and namespace called CNI ADD before the old container with the same pod name and namespace call  CNI DEL. Due to this, when the old container issued CNI DEL, it deleted the veth host interface, in turn deleting the eth0(other veth) of the new container.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/transparent_endpointclient_linux.go
+++ b/network/transparent_endpointclient_linux.go
@@ -12,8 +12,8 @@ import (
 
 const (
 	virtualGwIPString = "169.254.1.1/32"
-	defaultGwCidr = "0.0.0.0/0"
-	defaultGw = "0.0.0.0"
+	defaultGwCidr     = "0.0.0.0/0"
+	defaultGw         = "0.0.0.0"
 )
 
 type TransparentEndpointClient struct {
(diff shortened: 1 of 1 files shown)


network/transparent_endpointclient_linux.go
```diff
@@ -189,12 +189,5 @@ func (client *TransparentEndpointClient) ConfigureContainerInterfacesAndRoutes(e
 }
 
 func (client *TransparentEndpointClient) DeleteEndpoints(ep *endpoint) error {
```
ashvindeodhar, line 191: should we log indicating skipping endpoint deletion? It will help debugging any issues with transparent mode.

tamilmani1989: imo, i dont think its necessary because we skip deleting interface in all cases. If we are not deleting based on some condition then logs will be useful

matmerr (approved):

PR #776, opened 2021-02-01
[NPM] Ignoring hostnetwork pods from being added into Ipsets

vakalapa:
This fix involves ignoring any Add, Update or Delete Pod events for Hostnetwork Pods.

### DIFF
--- a/npm/pod.go
+++ b/npm/pod.go
@@ -126,6 +126,17 @@ func (npMgr *NetworkPolicyManager) UpdatePod(oldPodObj, newPodObj *corev1.Pod) e
 		return nil
 	}
 
+	// today K8s does not allow updating HostNetwork flag for an existing Pod. So NPM can safely
+	// check on the oldPodObj for hostNework value
+	if isHostNetworkPod(oldPodObj) {
+		log.Logf(
+			"POD UPDATING ignored for HostNetwork Pod:\n old pod: [%s/%s/%+v/%s/%s]\n new pod: [%s/%s/%+v/%s/%s]",
+			oldPodObj.ObjectMeta.Namespace, oldPodObj.ObjectMeta.Name, oldPodObj.Status.PodIP,
+			newPodObj.ObjectMeta.Namespace, newPodObj.ObjectMeta.Name, newPodObj.Status.PodIP,
+		)
+		return nil
+	}
+
 	if isInvalidPodUpdate(oldPodObj, newPodObj) {
 		return nil
 	}
@@ -149,17 +160,6 @@ func (npMgr *NetworkPolicyManager) UpdatePod(oldPodObj, newPodObj *corev1.Pod) e
 		oldPodObjNs, oldPodObjName, oldPodObjLabel, oldPodObjPhase, oldPodObjIP,
 		newPodObjNs, newPodObjName, newPodObjLabel, newPodObjPhase, newPodObjIP,
 	)
-
-	// today K8s does not allow updating HostNetwork flag for an existing Pod. So NPM can safely
-	// check on the oldPodObj for hostNework value
-	if isHostNetworkPod(oldPodObj) {
-		log.Logf(
-			"POD UPDATING ignored for HostNetwork Pod:\n old pod: [%s/%s/%+v/%s/%s]\n new pod: [%s/%s/%+v/%s/%s]",
-			oldPodObjNs, oldPodObjName, oldPodObjLabel, oldPodObjPhase, oldPodObjIP,
-			newPodObjNs, newPodObjName, newPodObjLabel, newPodObjPhase, newPodObjIP,
-		)
-		return nil
-	}
 
 	// Todo: Update if cached ip and podip changed and it is not a delete event
 
--- a/npm/pod.go
+++ b/npm/pod.go
@@ -21,14 +21,17 @@ func isSystemPod(podObj *corev1.Pod) bool {
 	return podObj.ObjectMeta.Namespace == util.KubeSystemFlag
 }
 
+func isHostNetworkPod(podObj *corev1.Pod) bool {
+	return podObj.Spec.HostNetwork
+}
+
 func isInvalidPodUpdate(oldPodObj, newPodObj *corev1.Pod) (isInvalidUpdate bool) {
 	isInvalidUpdate = oldPodObj.ObjectMeta.Namespace == newPodObj.ObjectMeta.Namespace &&
 		oldPodObj.ObjectMeta.Name == newPodObj.ObjectMeta.Name &&
 		oldPodObj.Status.Phase == newPodObj.Status.Phase &&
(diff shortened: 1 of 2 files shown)


npm/pod.go
```diff
@@ -27,7 +27,8 @@ func isInvalidPodUpdate(oldPodObj, newPodObj *corev1.Pod) (isInvalidUpdate bool)
 		oldPodObj.Status.Phase == newPodObj.Status.Phase &&
 		oldPodObj.Status.PodIP == newPodObj.Status.PodIP &&
 		newPodObj.ObjectMeta.DeletionTimestamp == nil &&
-		newPodObj.ObjectMeta.DeletionGracePeriodSeconds == nil
+		newPodObj.ObjectMeta.DeletionGracePeriodSeconds == nil &&
+		newPodObj.Spec.HostNetwork //Ignore if HostNetwork pod
```
csfmomo, line 31: Should we use && here? In my understanding, it's an OR relationship since if it use a host network, then we don't need to check other condition, we can just ignore it and let it be an invalid update.

vakalapa: Thank you.

npm/pod.go
```diff
@@ -140,6 +150,17 @@ func (npMgr *NetworkPolicyManager) UpdatePod(oldPodObj, newPodObj *corev1.Pod) e
 		newPodObjNs, newPodObjName, newPodObjLabel, newPodObjPhase, newPodObjIP,
 	)
 
+	// today K8s does not allow updating HostNetwork flag for an existing Pod. So NPM can safely
+	// check on the oldPodObj for hostNework value
+	if isHostNetworkPod(oldPodObj) {
```
matmerr, line 155: pr lgtm, but can we have add a UT to hit these lines for coverage and sanity :)

csfmomo, line 155: It's a nit pick. Do we want to move this logic before line 129, before isInvalidPodUpdate logic? Check whether isHostNetworkPod is quick and we can return early without compare oldPodObj with newPodObj.

vakalapa: Done, moved it above

npm/pod_test.go
```diff
@@ -202,3 +202,183 @@ func TestDeletePod(t *testing.T) {
...
+		t.Errorf("TestDeleteHostNetworkPod failed @ AddPod")
+	}
+
+	if len(npMgr.podMap) >= 1 {
+		t.Errorf("TestDeleteHostNetworkPod failed @ podMap length check")
+	}
+
+	if err := npMgr.DeletePod(podObj); err != nil {
+		t.Errorf("TestDeleteHostNetworkPod failed @ DeletePod")
+	}
```
csfmomo, line 382: Do we want to check the length of pod map now is 0?

npm/pod_test.go
```diff
@@ -202,3 +202,183 @@ func TestDeletePod(t *testing.T) {
...
+			Phase: "Running",
+			PodIP: "4.3.2.1",
+		},
+	}
+
+	npMgr.Lock()
+	if err := npMgr.AddPod(oldPodObj); err != nil {
+		t.Errorf("TestUpdateHostNetworkPod failed @ AddPod")
+	}
+
```
csfmomo, line 319: Do we want to check the ipset situation? When adding oldPodObj, ipset should have no item for this pod but after adding new pod, it will have 1 ipset. An example, https://github.com/Azure/azure-container-networking/blob/master/npm/nwpolicy_test.go#L116
However, I didn't evaluate whether we can start a ipset manager here. If we can't start ipset mgr, then we don't have a way to check it. The current test is good enough.

npm/pod_test.go
vakalapa: We are already checking that above, and in DeletePod, we are not going to add any pods to PodMap, length check on podmap may be redundant.

npm/pod_test.go
vakalapa: We will return early in the Update func after checking the hostnetwork true flag.

npm/pod_test.go
csfmomo: Does that mean after deleting pod, the len(npMgr.podMap) is still 1?

npm/pod_test.go
csfmomo: OK I see.

csfmomo (approved):

npm/pod_test.go
```diff
@@ -202,3 +202,183 @@ func TestDeletePod(t *testing.T) {
...
+	allNs, err := newNs(util.KubeAllNamespacesFlag)
+	if err != nil {
+		panic(err.Error)
+	}
+	npMgr.nsMap[util.KubeAllNamespacesFlag] = allNs
+
+	ipsMgr := ipsm.NewIpsetManager()
+	if err := ipsMgr.Save(util.IpsetTestConfigFile); err != nil {
+		t.Errorf("TestAddHostNetworkPod failed @ ipsMgr.Save")
```
ramiro-gamarra, line 221: add error to test logging. gives more context as to why something failed. same for below

npm/pod_test.go
```diff
@@ -202,3 +202,183 @@ func TestDeletePod(t *testing.T) {
...
+		Status: corev1.PodStatus{
+			Phase: "Running",
+			PodIP: "1.2.3.4",
+		},
+		Spec: corev1.PodSpec{
+			HostNetwork: true,
+		},
+	}
+
+	npMgr.Lock()
```
ramiro-gamarra, line 247: curious about locks being taken, since this variable is scoped to the test. what else could be modifying this?

vakalapa: These are all compartmentalized tests, in the current state they are in, no two tests (i have to double check) share the npMgr. So the lock should be scop is limited to a given testcase.

npm/pod_test.go
```diff
@@ -202,3 +202,183 @@ func TestDeletePod(t *testing.T) {
...
+func TestAddHostNetworkPod(t *testing.T) {
+	npMgr := &NetworkPolicyManager{
+		nsMap:            make(map[string]*namespace),
+		podMap:           make(map[string]string),
+		TelemetryEnabled: false,
+	}
+
+	allNs, err := newNs(util.KubeAllNamespacesFlag)
+	if err != nil {
+		panic(err.Error)
```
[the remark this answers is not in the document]
vakalapa: Logged a work item to fix this for all NPM tests.

npm/pod_test.go
vakalapa: Logged a work item to fix this for all NPM tests.

matmerr (approved):
approved, the npm test suite needs some help for sure :)

npm/pod_test.go
ramiro-gamarra: I suppose my question is, is it necessary to take locks?

PR #782, opened 2021-02-09
fix: Remove unreadable CNS state file upon CNS service start

ashvindeodhar:
This change removes unreadable state file ( azure-cns.json ) while starting the CNS service

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/util.go
+++ b/cns/restserver/util.go
@@ -74,13 +74,13 @@ func (service *HTTPRestService) saveState() error {
 }
 
 // restoreState restores CNS state from persistent store.
-func (service *HTTPRestService) restoreState() error {
+func (service *HTTPRestService) restoreState() {
 	logger.Printf("[Azure CNS] restoreState")
 
 	// Skip if a store is not provided.
 	if service.store == nil {
 		logger.Printf("[Azure CNS]  store not initialized.")
-		return nil
+		return
 	}
 
 	// Read any persisted state.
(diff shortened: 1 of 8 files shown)


cns/restserver/util.go
```diff
@@ -89,11 +89,12 @@ func (service *HTTPRestService) restoreState() error {
...
 			logger.Printf("[Azure CNS]  No state to restore.\n")
-			return nil
+		} else {
+			logger.Errorf("[Azure CNS]  Failed to restore state, err:%v. Removing azure-cns.json", err)
+			service.store.Remove()
 		}
 
-		logger.Errorf("[Azure CNS]  Failed to restore state, err:%v\n", err)
-		return err
+		return nil
```
thatmattlong, line 97: There is no codepath for this function to return an error now, so we should remove the error return type:
```go
func (service *HTTPRestService) restoreState() {
    ....
}
```

cns/restserver/util.go
```diff
@@ -89,15 +89,16 @@ func (service *HTTPRestService) restoreState() error {
 		if err == store.ErrKeyNotFound {
 			// Nothing to restore.
 			logger.Printf("[Azure CNS]  No state to restore.\n")
-			return nil
+		} else {
```
tamilmani1989, line 92: we should remove only oif its empty.. if not can we log and return error

tamilmani1989 (changes requested):

cns/restserver/util.go
tamilmani1989: i think this should be fine. as dnc will reconcile state even if its lost

tamilmani1989 (approved):
lgtm

PR #787, opened 2021-02-17
[NPM] Periodic Reconciliation of NPM Chain ordering in FORWARD table

vakalapa:
Today, NPM is in a configure and forget state for adding AZURE-NPM chain in FORWARD iptable chain. This change will include running a go routine to periodically (every 5 mins) check the order of AZURE-NPM chain in FORWARD. We can later extend this go routine to check other chains too.

### DIFF
--- a/npm/iptm/iptm.go
+++ b/npm/iptm/iptm.go
@@ -54,44 +54,14 @@ func NewIptablesManager() *IptablesManager {
 func (iptMgr *IptablesManager) InitNpmChains() error {
 	log.Logf("Initializing AZURE-NPM chains.")
 
-	if err := iptMgr.AddChain(util.IptablesAzureChain); err != nil {
-		return err
-	}
-
-	// Insert AZURE-NPM chain to FORWARD chain.
-	entry := &IptEntry{
-		Chain: util.IptablesForwardChain,
-		Specs: []string{
-			util.IptablesJumpFlag,
-			util.IptablesAzureChain,
-		},
-	}
-	exists, err := iptMgr.Exists(entry)
+	err := iptMgr.CheckAndAddForwardChain()
 	if err != nil {
-		return err
+		metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to add AZURE-NPM chain to FORWARD chain.")
 	}
 
-	if !exists {
-		// retrieve KUBE-SERVICES index
-		index := "1"
-		iptFilterEntries := exec.Command(util.Iptables, "-t", "filter", "-n", "--list", "FORWARD", "--line-numbers")
-		grep := exec.Command("grep", "KUBE-SERVICES")
-		pipe, _ := iptFilterEntries.StdoutPipe()
-		grep.Stdin = pipe
-		iptFilterEntries.Start()
-		output, err := grep.CombinedOutput()
-		if err == nil && len(output) > 2 {
-			tmpIndex, _ := strconv.Atoi(string(output[0]))
-			index = strconv.Itoa(tmpIndex + 1)
-		}
-		pipe.Close()
-		// position Azure-NPM chain after Kube-Forward and Kube-Service chains if it exists
-		iptMgr.OperationFlag = util.IptablesInsertionFlag
-		entry.Specs = append([]string{index}, entry.Specs...)
-		if _, err = iptMgr.Run(entry); err != nil {
-			metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to add AZURE-NPM chain to FORWARD chain.")
-			return err
-		}
+	entry := &IptEntry{
+		Chain: util.IptablesAzureChain,
+		Specs: []string{},
 	}
 
 	// Create AZURE-NPM-INGRESS-PORT chain.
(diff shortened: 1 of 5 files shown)


npm/pod.go
```diff
@@ -126,21 +126,21 @@ func (npMgr *NetworkPolicyManager) UpdatePod(oldPodObj, newPodObj *corev1.Pod) e
 		return nil
 	}
 
+	if isInvalidPodUpdate(oldPodObj, newPodObj) {
```
neaggarwMS, line 129: Isnt this already checked in?

npm/iptm/iptm.go
```diff
@@ -437,6 +407,83 @@
...
+		return nil
+	}
+
+	npmChainLine, err := iptMgr.GetChainLineNumber(util.IptablesAzureChain, util.IptablesForwardChain)
+	if err != nil {
+		metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to get index of AZURE-NPM in FORWARD chain with error: %s", err.Error())
+	}
+
+	// Kube-services line number is less than npm chain line number then all good
+	if kubeServicesLine > 0 && npmChainLine == index {
```
neaggarwMS, line 459: >npmChainLine == index [](start = 28, length = 21)

Instead should we simply check if npmChainLine > kubeServicesLane?

npm/iptm/iptm.go
```diff
@@ -437,6 +407,83 @@
...
+		return nil
+	} else if kubeServicesLine <= 0 {
+		return nil
+	}
+
+	// NPM Chain number is less than KUBE-SERVICES then
+	// delete existing NPM chain and add it in the right order
+	iptMgr.OperationFlag = util.IptablesDeletionFlag
```
neaggarwMS, line 467: Can you log in case the order got swaped. Also add a metric as we would like to get the data for how many times we did the reconcilation

neaggarwMS (changes requested):
:clock1:

npm/npm.go
```diff
@@ -313,3 +315,16 @@ func NewNetworkPolicyManager(clientset *kubernetes.Clientset, informerFactory in
 
 	return npMgr
 }
+
+// reconcileChains checks for ordering of AZURE-NPM chain in FORWARD chain periodically.
+func (npMgr *NetworkPolicyManager) reconcileChains() {
+	iptMgr := iptm.NewIptablesManager()
+	var err error
```
ramiro-gamarra, line 322: no need to lift this variable out of the scope where it's used

npm/npm.go
```diff
@@ -313,3 +315,16 @@ func NewNetworkPolicyManager(clientset *kubernetes.Clientset, informerFactory in
...
+	for {
+		time.Sleep(reconcileChainTimeInMinutes * time.Minute)
+
+		if err = iptMgr.CheckAndAddForwardChain(); err != nil {
+			metrics.SendErrorLogAndMetric(util.NpmID, "Error: failed to add AZURE-NPM chain to FORWARD chain.")
```
ramiro-gamarra, line 327: the specific error context is lost here, better to add it to the message?

npm/npm.go
```diff
@@ -313,3 +315,16 @@ func NewNetworkPolicyManager(clientset *kubernetes.Clientset, informerFactory in
 
 	return npMgr
 }
+
+// reconcileChains checks for ordering of AZURE-NPM chain in FORWARD chain periodically.
+func (npMgr *NetworkPolicyManager) reconcileChains() {
+	iptMgr := iptm.NewIptablesManager()
+	var err error
+	for {
+		time.Sleep(reconcileChainTimeInMinutes * time.Minute)
```
ramiro-gamarra, line 324: Probably better to use `time.After` in a `for select` if you ever intend to do handle termination gracefully, since that gives you a hook for a context.

npm/npm.go
```diff
@@ -177,6 +178,7 @@ func (npMgr *NetworkPolicyManager) Start(stopCh <-chan struct{}) error {
 		return fmt.Errorf("Network policy informer failed to sync")
 	}
 
+	go npMgr.reconcileChains()
```
ramiro-gamarra, line 181: My main question about this is, are there any other goroutines calling iptables? If so, what is the consequence of having iptables commands being run simultaneously? Not necessarily asking about iptables being thread safe since each invocation might have its own process, but interleaved invocations (goroutine A and B both call `CheckAndAddForwardChain` at the same time, which in turn calls iptables multiple times). Do those blocks need to perform atomically?

ramiro-gamarra (commented):
Some feedback

npm/pod.go
vakalapa: Yes it did get checked in but this might be stale code from all the branch changes we did. since this was a single commit, I could not cut it down

npm/npm.go
vakalapa: Removing and returning this error as the log message will be redundant

npm/npm.go
vakalapa: Iptables have inbuilt locking mechanism and NPM has a 60 sec wait to acquire the lock and then perform iptables action, so we do not have to worry about concurrency in issuing iptable commands

npm/iptm/iptm.go
```diff
@@ -437,6 +407,83 @@
...
+	// retrieve KUBE-SERVICES index
+	kubeServicesLine, err := iptMgr.GetChainLineNumber(util.IptablesKubeServersChain, util.IptablesForwardChain)
+	if kubeServicesLine > 0 {
+		index = kubeServicesLine + 1
+	}
+	if err != nil {
+		metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to get index of KUBE-SERVICES in FORWARD chain with error: %s", err.Error())
+	}
+
+	if !exists {
```
matmerr, line 441: imo we should call iptMgr.Exists closer to where the result exists is used

matmerr (approved):
approved, taking minor fixes in subsequent PR

neaggarwMS (approved):
:shipit:

PR #788, opened 2021-02-17
CNI: Modified CNI to be pass through for Azure baremetal scenario

msvik:
**Reason for Change**:
Adding support in CNI for "baremetal scenario" where when CNI is executed on baremetal nodes, it is pass through and invokes Azure's node service to configure container networking. 

**Issue Fixed**:
Added 

**Requirements**:

- [ ] includes documentation
- [ x] adds unit tests


cni/network/network.go
```diff
@@ -791,6 +814,31 @@ func (plugin *netPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+		telemetry.SendCNIMetric(&cniMetric, plugin.tb)
+	}()
+
+	log.Printf("Execution mode :%s", nwCfg.ExecutionMode)
+	if nwCfg.ExecutionMode == string(Baremetal) {
+
+		log.Printf("Baremetal mode. Calling vnet agent for delete container")
+
+		// set the flag to true for attempt. This will be used by deferred telemetry method to emit metrics
+		deleteTried = true
```
ramiro-gamarra, line 837: Flags in defers are hard to follow (you need to scan the entire method to know where it could have changed). Shows better intent ("only send metrics after this code path") if we extract the deferred anonymous function out, and defer in places where needed, like

```go
func sendMetric(...) {...}

func delete() { 
    if nwCfg.ExecutionMode == string(Baremetal) {
        defer sendMetric(...)
        return plugin.nnsClient.DeleteContainerNetworking()
    }

    defer sendMetric(...)
    return plugin.nm.DeleteEndpoint
}
```

cni/network/plugin/main.go
```diff
@@ -173,7 +174,7 @@ func main() {
 
 	cniReport.GetReport(pluginName, version, ipamQueryURL)
 
-	netPlugin, err := network.NewPlugin(name, &config)
+	netPlugin, err := network.NewPlugin(name, &config, &nnsgrpcclient.NnsGrpcClient{})
```
ramiro-gamarra, line 177: Just a thought on package naming, it's very common in Go to have packages reflect domain, such that types within can have succinct names, without losing context from a callers perspective. You could have an `nns` package, and then types such as `nns.Client`

cni/nnsgrpcclient/nnsgrpcclient_test.go
```diff
@@ -0,0 +1,117 @@
...
+	setup()
+	code := m.Run()
+	teardown()
+	os.Exit(code)
+}
+
+func setup() {
+	mockserver  = nnsmockserver.NewNnsMockServer()
+	go  mockserver.StartGrpcServer(nnsPort)
+	fmt.Printf("mock nns server started\n")
```
ramiro-gamarra, line 25: `fmt.Println` if you don't need formatting directives, also saves you from having to write the newline. Same for below.

cni/nnsgrpcclient/nnsgrpcclient_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+// CNI ADD to add container to network
+func TestAddContainerNetworking(t *testing.T) {
+
+	client := &NnsGrpcClient{}
+	if err := client.AddContainerNetworking(
+		 context.Background(),
+		"sf_8e9961f4-5b4f-4b3c-a9ae-c3294b0d9681",
+		"testnwspace"); err != nil {
+       t.Fatalf("TestAddContainerNetworking failed: %v", err)
```
ramiro-gamarra, line 41: formatting seems off. if you are using goland, you can enable file watchers and goimports (like gofmt but with added functionality)

cni/nnsgrpcclient/nnsgrpcclient_test.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	fmt.Printf("TestAddContainerNetworking success\n")
```
ramiro-gamarra, line 44: - you can use `t.Log` or `t.Logf` for logging within the test
- running the tests in verbose mode will tell you which pass and fail, so this seems a bit redundant. same for below

cni/nnsgrpcclient/nnsgrpcclient_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+func teardown() {
+	mockserver.StopGrpcServer()
+	fmt.Printf("mock nns server stopped\n")
+}
+
+// CNI ADD to add container to network
+func TestAddContainerNetworking(t *testing.T) {
+
+	client := &NnsGrpcClient{}
```
ramiro-gamarra, line 36: how does the client know to hit the mock server?

cni/nnsgrpcclient/nnsgrpcclient_test.go
```diff
@@ -0,0 +1,117 @@
...
+func TestAddContainerNetworkingFailure(t *testing.T) {
+
+	client := &NnsGrpcClient{}
+
+	var err error
+	if err = client.AddContainerNetworking(context.Background(), "testpod", "testnwspace"); err == nil {
+		t.Fatalf("TestAddContainerNetworkingFailure failed. Expected error but none returned\n")
+	}
+
+	if !strings.Contains(err.Error(), "Setup") {
```
ramiro-gamarra, line 71: Checking the string representation of errors is brittle. It's better to have error types. In this case, does it matter what specific error was returned? Same for below

cni/nnsgrpcclient/nnsgrpcclient_test.go
```diff
@@ -0,0 +1,117 @@
...
+func TestAddContainerNetworkingGrpcServerDown(t *testing.T) {
+
+	// shutdown server to simulate connection error
+	teardown()
+
+	// bring up the server again after 25 seconds
+	go func() {
+		time.Sleep(25 * time.Second)
+		setup()
+	}()
```
ramiro-gamarra, line 105: how does this test work? does the client retry indefinitely? if it doesn't, then since the mock server is shared in many tests, this can leave set up broken for future tests. If it does, 25 seconds seems very long for this test to run (and for the client to retry). let me know if im missing something

cni/nnsgrpcclient/nnsgrpclient.go
```diff
@@ -0,0 +1,89 @@
...
+	"context"
+	"fmt"
+	"github.com/Azure/azure-container-networking/log"
+	nns "github.com/Azure/azure-container-networking/proto/nodenetworkservice/3.302.0.744"
+	"google.golang.org/grpc"
+	"time"
+)
+
+const (
+	nnsPort = "6668"
```
ramiro-gamarra, line 14: how set in stone is this? should they ever change, code would need to as well. should this be config instead?

cni/nnsgrpcclient/nnsgrpclient.go
```diff
@@ -0,0 +1,89 @@
...
+	apiTimeout = 5 * time.Minute // recommended timeout from node service
```
ramiro-gamarra, line 15: how intensive are these operations that they may require up to 5 minutes? 😁

ramiro-gamarra (commented):
Some early feedback

cni/network/network.go
```diff
@@ -791,6 +814,31 @@ func (plugin *netPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+		err = plugin.nnsClient.DeleteContainerNetworking(context.Background(), k8sPodName, args.Netns)
+		return err
```
tamilmani1989, line 839: do we return error for all cases even if container is already removed ? adding brief desc will be helpful

cni/nnsgrpcclient/nnsgrpclient.go
```diff
@@ -0,0 +1,89 @@
...
+		ContainerId: podName,
+		NetworkNamespaceId: nwNamespace,
+	}
+
+	localCtx, cancel := context.WithTimeout(ctx, apiTimeout)
+	defer cancel()
+
+	res, err := client.ConfigureContainerNetworking(localCtx, req)
+	if err != nil {
+		log.Debugf("ConfigureContainerNetworking for container %s for nw namespace %s failed with error: %s",
```
tamilmani1989, line 66: one suggestion, for baremetal specific logs you can add prefix [baremetal]

cni/nnsgrpcclient/nnsgrpclient.go
```diff
@@ -0,0 +1,89 @@
...
+    return err
+}
+
+func newNnsGrpcClient (ctx context.Context) (nns.NodeNetworkServiceClient, *grpc.ClientConn,  error)  {
+
+	localCtx, cancel := context.WithTimeout(ctx, connectionTimeout)
+	defer cancel()
+
+    nnsEndpoint := fmt.Sprintf(":%s", nnsPort)
+	conn, err := grpc.DialContext(
```
tamilmani1989, line 82: add log to print nnsendpoint cni connecting to

cni/nnsgrpcclient/nnsgrpclient.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func configureContainerNetworking (
+	ctx context.Context,
+	reqtype nns.RequestType,
+	podName, nwNamespace string) error {
+
+	client, conn, err := newNnsGrpcClient(ctx)
+	if err != nil {
+		return err
```
tamilmani1989, line 46: hope we log this error somewhere

cni/nnsgrpcclient/nnsgrpclient.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/Azure/azure-container-networking/log"
+	nns "github.com/Azure/azure-container-networking/proto/nodenetworkservice/3.302.0.744"
+	"google.golang.org/grpc"
+	"time"
+)
+
+const (
+	nnsPort = "6668"
+	apiTimeout = 5 * time.Minute // recommended timeout from node service
+	connectionTimeout = 2 * time.Minute
```
tamilmani1989, line 16: is 2 mins isn't it longer timeout?

proto/nodenetworkservice/3.302.0.744/NodeNetworkService.proto
```diff
@@ -0,0 +1,53 @@
+syntax = "proto3";
+package azure.nodenetworkservice;
```
tamilmani1989, line 2: can it be azurenodenetworkservice or azurenns

proto/nodenetworkservice/3.302.0.744/NodeNetworkService.proto
```diff
@@ -0,0 +1,53 @@
...
+option go_package = ".;nodenetworkservice";
+
+service NodeNetworkService {
+    rpc ConfigureNetworking(ConfigureNetworkingRequest) returns (ConfigureNetworkingResponse);
```
tamilmani1989, line 6: can you add short description on what these api does?

also whats the diff between ConfigureNetworking and ConfigureContainerNetworking

cni/nnsgrpcclient/nnsgrpclient.go
```diff
@@ -0,0 +1,89 @@
...
+			podName, nwNamespace, err)
+	} else {
+		log.Debugf("ConfigureContainerNetworking for container %s for nw namespace %s succeeded with result: %v",
+			podName, nwNamespace, res)
+	}
+
+    return err
+}
+
+func newNnsGrpcClient (ctx context.Context) (nns.NodeNetworkServiceClient, *grpc.ClientConn,  error)  {
```
tamilmani1989, line 76: short desc on what this function does would help

cni/nnsgrpcclient/nnsgrpclient.go
```diff
@@ -0,0 +1,89 @@
...
+	req := &nns.ConfigureContainerNetworkingRequest{
+		RequestType: reqtype,
+		ContainerId: podName,
+		NetworkNamespaceId: nwNamespace,
+	}
+
+	localCtx, cancel := context.WithTimeout(ctx, apiTimeout)
+	defer cancel()
+
+	res, err := client.ConfigureContainerNetworking(localCtx, req)
```
tamilmani1989, line 64: can you add a comment on what this function does

tamilmani1989 (changes requested):

cni/network/network.go
```diff
@@ -60,13 +61,26 @@ const (
...
+)
+
 // NetPlugin represents the CNI network plugin.
 type netPlugin struct {
 	*cni.Plugin
 	nm          network.NetworkManager
 	ipamInvoker IPAMInvoker
 	report      *telemetry.CNIReport
 	tb          *telemetry.TelemetryBuffer
+	nnsClient   NnsClient
```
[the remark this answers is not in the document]
msvik: Agree, this is an overhead. Having said that no one has brought up performance as an issue yet, so we took the path of least resistance for now. If performance become an issue due to this connection setup overhead, CNS can be considered an option.

cni/network/plugin/main.go
msvik: Moved it outside of plugin and renamed the package.

cni/nnsgrpcclient/nnsgrpcclient_test.go
msvik: ran gofmt on this file.

cni/nnsgrpcclient/nnsgrpcclient_test.go
msvik: fixed

cni/network/network.go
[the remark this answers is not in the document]
msvik: I have moved the nns grpclient out of plugin. It is at the root level now just like other client packages. The reason it is part of netPlugin struct is that because we are doing a dependency injection here via an interface. And it is passed into plugin struct from main.go. This way we can also pass a mock implementation for testing. For example - we can add a parameter in CNI config saying - "UseMockNns" and when it is true, we initialize and pass a mock client all the way to plugin. That is the intention here.

cni/network/network.go
```diff
@@ -60,13 +61,26 @@ const (
...
+}
+
+// client for node network service
+type NnsClient interface {
+	AddContainerNetworking(ctx context.Context, podName, nwNamespace string) error
+	DeleteContainerNetworking(ctx context.Context, podName, nwNamespace string) error
```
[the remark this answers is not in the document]
msvik: Added documentation

cni/network/network.go
msvik: CNI is pass through. All cases are handled by node network service.

cni/nnsgrpcclient/nnsgrpcclient_test.go
msvik: Since this client is meant to be used on a host, it is written to reach out to a specific port on localhost. Test actually starts a mock nns server (which starts at the expected port), instantiates a client and invokes APIs

cni/nnsgrpcclient/nnsgrpcclient_test.go
msvik: make sense. resolved.

cni/nnsgrpcclient/nnsgrpcclient_test.go
msvik: client's connection timeout is 2 minutes, so it will retry for up 2 mins, so 25 seconds chosen here well within that range. If we run tests in parallel in this package, this could be a problem but in general we run it sequentially, so it shouldn't impact future tests.

cni/nnsgrpcclient/nnsgrpclient.go
msvik: My understanding is that once it is set, these things don't really change since it is a host level component and will require deployment across the fleet for any client depending on it. Having it in the config is definitely better, but regardless we will need to produce a new package. I'll ask node service team if it can be changed in future.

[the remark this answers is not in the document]
msvik: done

cni/nnsgrpcclient/nnsgrpclient.go
msvik: This was recommended by nns team to us. I don't think API is intensive or takes long time. When it works, whole thing gets done in matter of seconds. But so far we have only tested it on a single node with single pod being created. At a scale, it is possible they NNS might take longer. Scale testing hasn't be done yet.

cni/nnsgrpcclient/nnsgrpclient.go
msvik: This is returned to plugin and plugin logs it just like any other error.

cni/nnsgrpcclient/nnsgrpclient.go
msvik: default http connection timeout is 2 mins, so I kept it same.

cni/nnsgrpcclient/nnsgrpclient.go
msvik: Added documentation in this package.

cni/nnsgrpcclient/nnsgrpclient.go
msvik: done

cni/nnsgrpcclient/nnsgrpclient.go
msvik: added the baremetal prefix in logs

proto/nodenetworkservice/3.302.0.744/NodeNetworkService.proto
msvik: This comes from nns team, do I didn't want to rename it. When we generate go contracts from it, we override this package name to nodenetworkservice.

proto/nodenetworkservice/3.302.0.744/NodeNetworkService.proto
msvik: Again, this file comes from nns team. in future it could come via a nuget too. So we don't want to edit it. I have already added description of ConfigureContainerNetwoking in grpc client. Other APIs for other services that consume this contract. We only consume ConfigureContainerNetworking API.

cni/nnsgrpcclient/nnsgrpcclient_test.go
msvik: This is not needed as such. I added it because same API in mock server returns result of both Add/Delete. So this extra check here makes sure of right operation was done on mock server.

cni/network/network.go
msvik: good suggestion. fixed

cni/network/network.go
```diff
@@ -1143,3 +1192,42 @@ func determineSnat() (bool, bool, error) {
...
+					log.Printf("convertNnsToCniResult error: %s \n", err)
+					continue
+				}
+
+				gateway := net.ParseIP(ip.DefaultGateway)
+				ipConfig := &cniTypesCurr.IPConfig{
+					Address: *ipNet,
+					Gateway: gateway,
+					Version: ip.Version,
+					Interface: &intIndex,
```
ramiro-gamarra, line 1222: do we own this struct? pointer to int is interesting

nns/grpcclient_test.go
```diff
@@ -0,0 +1,107 @@
...
+	setup()
+	code := m.Run()
+	teardown()
+	os.Exit(code)
+}
+
+func setup() {
+	mockserver = nnsmockserver.NewNnsMockServer()
+	go mockserver.StartGrpcServer(nnsPort)
+	fmt.Println("mock nns server started\n")
```
ramiro-gamarra, line 25: `fmt.Println` doesn't need line break. same for below

nns/grpcclient_test.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// CNI ADD to add container to network - failure case
+func TestAddContainerNetworkingFailure(t *testing.T) {
+
+	client := &GrpcClient{}
+
+	var err error
+	if err, _ = client.AddContainerNetworking(context.Background(), "testpod", "testnwspace"); err == nil {
+		t.Fatalf("TestAddContainerNetworkingFailure failed. Expected error but none returned\n")
```
ramiro-gamarra, line 64: line break not needed, same for below

ramiro-gamarra (commented):
few more comments

cni/network/network.go
```diff
@@ -60,13 +62,33 @@ const (
 	jsonFileExtension = ".json"
 )
 
+type  ExecutionModes string
```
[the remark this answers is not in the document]
msvik: fixed

cni/network/network.go
```diff
@@ -361,6 +379,24 @@ func (plugin *netPlugin) Add(args *cniSkel.CmdArgs) error {
 		return plugin.Errorf(errMsg)
 	}
 
+	log.Printf("Execution mode :%s", nwCfg.ExecutionMode)
+	if nwCfg.ExecutionMode == string(Baremetal) {
+		var  res *nnscontracts.ConfigureContainerNetworkingResponse
+        log.Printf("Baremetal mode. Calling vnet agent for ADD")
```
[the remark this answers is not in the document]
msvik: I did gofmt on entire file

cni/network/network.go
```diff
@@ -1143,3 +1192,42 @@ func determineSnat() (bool, bool, error) {
...
+	result := &cniTypesCurr.Result{}
+	var resultIpconfigs []*cniTypesCurr.IPConfig
+
+    if netRes.Interfaces != nil {
+		for i, ni :=  range netRes.Interfaces {
+
+			intIndex := i
+			for _, ip := range ni.Ipaddresses {
+
+				ipWithPrefix := strings.Join([]string { ip.Ip, ip.PrefixLength }, "/")
```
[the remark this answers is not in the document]
msvik: fixed

cni/network/network.go
```diff
@@ -1143,3 +1192,42 @@ func determineSnat() (bool, bool, error) {
...
+				_, ipNet, err := net.ParseCIDR(ipWithPrefix)
+				if err != nil {
+					log.Printf("convertNnsToCniResult error: %s \n", err)
```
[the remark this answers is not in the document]
msvik: Added more info in the logs. Removed \n

cni/network/network.go
msvik: This is vendored from container networking repo.

cni/network/network.go
```diff
@@ -1143,3 +1193,47 @@ func determineSnat() (bool, bool, error) {
...
+	// This function does not add interfaces to CNI result. Reason being CRI (containerD in baremetal case)
+	// only looks for default interface named "eth0" and this default interface is added in the defer
+	// method of ADD method
+	result := &cniTypesCurr.Result{}
+	var resultIpconfigs []*cniTypesCurr.IPConfig
+
+	if netRes.Interfaces != nil {
+		for i, ni := range netRes.Interfaces {
+
+			intIndex := i
```
tamilmani1989, line 1212: why do we need copy of this variable?

tamilmani1989 (changes requested):

cni/network/network.go
msvik: I think it being done when creating result object. That's why.

tamilmani1989: i think this is the place you were mentioning:` Interface: &intIndex,` . can't we use &i instead of &intIndex?

tamilmani1989 (approved):

PR #796, opened 2021-02-22
[NPM] Consolidating default chains and rules for easier readability

vakalapa:
Right now all the basic chain and rules are hard to comprehend with redundant code. Cleaning that up

### DIFF
--- a/npm/iptm/helper.go
+++ b/npm/iptm/helper.go
@@ -1,0 +1,202 @@
+package iptm
+
+import (
+	"fmt"
+
+	"github.com/Azure/azure-container-networking/npm/util"
+)
+
+// GetAllChainsAndRules returns all NPM chains and rules
+func getAllChainsAndRules() [][]string {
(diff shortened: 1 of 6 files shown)


npm/iptm/helper.go
```diff
@@ -0,0 +1,202 @@
...
+
+	return chainsAndRules
+}
+
+// getAzureNPMChainRules returns all rules for AZURE-NPM chain
+func getAzureNPMChainRules() [][]string {
+	// Note: make sure 0th index is prent chain for logging
+	return [][]string{
+		{
+			util.IptablesAzureChain,
```
matmerr, line 33: let's make a note to potentially wrap these string slices into structs with validation, the potential for ordering mishaps is pretty high

vakalapa: Noted!

matmerr (approved):
approved, not sure why the diff shows changes in the pipeline yamls, master should be up to date

PR #799, opened 2021-02-24
[NPM] Default-Deny all bug fix with design change

vakalapa:
With the new MARK'ing design change, default deny all scenario with an egress allow is failing. This PR helps correcting this bug


matmerr (approved):
reviewed on call

PR #807, opened 2021-03-01
[CNI] Bypassing POSTROUTING table for Swift POD traffic

vakalapa:
Adding a new rule in POSTROUTING table to accept pod traffic going to internet

Chain POSTROUTING (policy ACCEPT 37 packets, 2743 bytes)
    0     0 ACCEPT     all  --  *      *       10.241.0.0/16        0.0.0.0/0            destination IP range ! 168.63.129.16-168.63.129.16

### DIFF
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -150,6 +150,9 @@ func setHostOptions(nwCfg *cni.NetworkConfig, hostSubnetPrefix *net.IPNet, ncSub
 	}
 
 	azureDNSMatch := fmt.Sprintf(" -m addrtype ! --dst-type local -s %s -d %s -p %s --dport %d", ncSubnetPrefix.String(), iptables.AzureDNS, iptables.UDP, iptables.DNSPort)
+
+	// TODO remove this rule once we remove adding MASQUEARDE from AgentBaker, check below PR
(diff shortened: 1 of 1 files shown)


cni/network/invoker_cns.go
```diff
@@ -150,9 +150,14 @@ func setHostOptions(nwCfg *cni.NetworkConfig, hostSubnetPrefix *net.IPNet, ncSub
 	}
 
 	azureDNSMatch := fmt.Sprintf(" -m addrtype ! --dst-type local -s %s -d %s -p %s --dport %d", ncSubnetPrefix.String(), iptables.AzureDNS, iptables.UDP, iptables.DNSPort)
+
+	// TODO remove this rule once we remove adding MASQUEARDE from AgentBaker, check below PR
+	// https://github.com/Azure/AgentBaker/pull/367/files
+	podTrafficAccept := fmt.Sprintf(" -m iprange  ! --dst-range 168.63.129.16-168.63.129.16  -s %s ", ncSubnetPrefix.String())
```
matmerr, line 156: since dst is a single ip, can we say `--dst` instead of `--dst-range` ?

neaggarwMS (approved):
:shipit:

PR #809, opened 2021-03-02
Azure CNS fix reconcile bug

neaggarwMS:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/ipampoolmonitor/ipampoolmonitor.go
+++ b/cns/ipampoolmonitor/ipampoolmonitor.go
@@ -27,6 +27,7 @@ type CNSIPAMPoolMonitor struct {
 }
 
 func NewCNSIPAMPoolMonitor(cns cns.HTTPService, rc requestcontroller.RequestController) *CNSIPAMPoolMonitor {
+	logger.Printf("NewCNSIPAMPoolMonitor: Create IPAM Pool Monitor")
 	return &CNSIPAMPoolMonitor{
 		pendingRelease: false,
 		cns:            cns,
@@ -52,7 +53,7 @@ func (pm *CNSIPAMPoolMonitor) Start(ctx context.Context, poolMonitorRefreshMilli
 	for {
 		select {
 		case <-ctx.Done():
-			return fmt.Errorf("CNS IPAM Pool Monitor received cancellation signal")
+			return fmt.Errorf("[ipam-pool-monitor] CNS IPAM Pool Monitor received cancellation signal")
 		case <-ticker.C:
 			err := pm.Reconcile()
 			if err != nil {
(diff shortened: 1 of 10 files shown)


cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -139,20 +143,38 @@
...
-		err error
-	)
+	logger.Printf("StartRequestController")
 
-	logger.Printf("Initializing CNS state")
-	if err = crdRC.initCNS(); err != nil {
-		logger.Errorf("[cns-rc] Error initializing cns state: %v", err)
-		return err
+	crdRC.lock.Lock()
+	if crdRC.initialized != true {
```
thatmattlong, line 169: Could simplify with `if !crdRC.initialized {}`

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -242,7 +271,7 @@
 }
 
 // UpdateCRDSpec updates the CRD spec
-func (crdRC *crdRequestController) UpdateCRDSpec(cntxt context.Context, crdSpec nnc.NodeNetworkConfigSpec) error {
+func (crdRC *crdRequestController) UpdateCRDSpec(cntxt context.Context, crdSpec nnc.NodeNetworkConfigSpec)  error {
```
thatmattlong, line 274: I don't think anything changed here, just an extra space snuck in before the return type

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -100,8 +102,8 @@ func (pm *CNSIPAMPoolMonitor) Reconcile() error {
 }
 
 func (pm *CNSIPAMPoolMonitor) increasePoolSize() error {
-	pm.mu.Lock()
 	defer pm.mu.Unlock()
+	pm.mu.Lock()
```
ramiro-gamarra, line 106: Curious, why the flip?

cns/service/main.go
```diff
@@ -709,3 +665,94 @@
...
+		if requestController.IsStarted() {
+			logger.Printf("RequestController is started")
+			break
+		}
+
+		logger.Printf("Waiting for requestController to start...")
+		time.Sleep(time.Millisecond * 500)
+	}
+
+	ctx := context.Background()
```
thatmattlong, line 737: At some point we need to use contexts properly, such as adding a cancel func or timeout. However, this PR doesn't need to address this, since there are `context.Background()`s and `context.TODO()`s all throughout the CNS code, so we probably need one PR just to address context.

thatmattlong (approved):

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -42,6 +43,9 @@ type crdRequestController struct {
 	CNSClient       cnsclient.APIClient
 	nodeName        string //name of node running this program
 	Reconciler      *CrdReconciler
+	initialized     bool
+	Started         bool
+	lock            sync.Mutex
```
ramiro-gamarra, line 48: After looking at this and its usage for a bit, I understand the intent, but I think this is a sign that this struct needs to be split into different abstractions. Ideally, this shouldn't need to carry state about initialization, we should just have ready-to-go dependencies injected, or document proper usage. I would suggest adding a todo to clean up further because the intention of these is hard to follow. Also, since there is a getter, `Started` need not be exported.

cns/ipampoolmonitor/ipampoolmonitor.go
neaggarwMS: No reason in particular, just coding practice to add a defer unlock API before Lock. 

---
In reply to: [585721003](https://github.com/Azure/azure-container-networking/pull/809#discussion_r585721003) [](ancestors = 585721003)

cns/service/main.go
neaggarwMS: Agreed

---
In reply to: [585724212](https://github.com/Azure/azure-container-networking/pull/809#discussion_r585724212) [](ancestors = 585724212)

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
neaggarwMS: This piece of code requires adding more tests. Created a PBI to incorporate feedback and add more tests

---
In reply to: [585726500](https://github.com/Azure/azure-container-networking/pull/809#discussion_r585726500) [](ancestors = 585726500)

cns/restserver/restserver.go
```diff
@@ -209,6 +209,19 @@ func (service *HTTPRestService) Start(config *common.ServiceConfig) error {
...
+func (service *HTTPRestService) Start(config *common.ServiceConfig) error {
+
+	// Start the listener.
+	// continue to listen on the normal endpoint for http traffic, this will be supported
+	// for sometime until partners migrate fully to https
+	if err := service.StartListener(config); err != nil {
+		return err
+	}
+
+	return nil
```
ramiro-gamarra, line 222: Can just 
```go
return service.StartListener(config)
```

ramiro-gamarra (approved):

PR #815, opened 2021-03-08
fix: ACLPolicy check for hnsv2

ashvindeodhar:
Current code is incorrectly checking against EndpointPolicyType.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


aegal (approved):
lgtm

aegal (commented):
Is it valid to rely on Json.unmarshal to detect this is a type of ACL policy, as far as I am aware, if the data defintion is different this doesn't return error as long as you feed it properly formed JSON. So I believe in this case, we will always return ACL policy if it gets there.

aegal (approved):
lgtm

PR #819, opened 2021-03-11
[fix] IPset fail if set doesn't exist when attempting to add to list

matmerr:
**Reason for Change**:

This change is to make sure that an ipset exists in npm state, so we don't waste calls to ipset which will fail. Also add check to make sure that when IP being added to set it is nonempty

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/ipsm/ipsm.go
+++ b/npm/ipsm/ipsm.go
@@ -4,6 +4,7 @@
 package ipsm
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"regexp"
@@ -133,8 +134,9 @@ func (ipsMgr *IpsetManager) AddToList(listName string, setName string) error {
 		return nil
 	}
 
+	// ensure set being added to list exists
 	if !ipsMgr.SetExists(setName, util.IpsetSetListFlag) {
-		return nil
+		return fmt.Errorf("Set does not exist [%s] when attempting to add to list [%s]", setName, listName)
 	}
 
 	if ipsMgr.Exists(listName, setName, util.IpsetSetListFlag) {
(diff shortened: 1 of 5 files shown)


npm/ipsm/ipsm.go
```diff
@@ -267,11 +291,19 @@ func (ipsMgr *IpsetManager) AddToSet(setName, ip, spec, podUid string) error {
 		return nil
 	}
 
-	if !ipsMgr.SetExists(setName, spec) {
+	if ip == "" {
```
vakalapa, line 294: This alone check may not work, since in case of hash:ip,port, the ip will be "1.1.1.1,tcp:80", but if the Ip is empty then at this stage we will get ",tcp:80"

vakalapa (changes requested):

npm/ipsm/ipsm.go
```diff
@@ -267,11 +291,25 @@ func (ipsMgr *IpsetManager) AddToSet(setName, ip, spec, podUid string) error {
...
 	}
 
-	if !ipsMgr.SetExists(setName, spec) {
+	// possible formats
+	//192.168.0.1
+	//192.168.0.1,tcp:25227
+	// todo: handle ip and port with protocol, plus just ip
+	// always guaranteed to have ip, not guaranteed to have port + protocol
+	ipDetails := strings.Split(ip, ",")
+	if len(ipDetails) > 0 && ipDetails[0] == "" {
```
vakalapa, line 300: Are we handling the case where len(ipDetails) == 0 ?

PR #821, opened 2021-03-12
Change version comparison from string to int and add more log in SyncHostNCVersion.

csfmomo:
fix: Change version comparison from string to int and as well as the following 3 improvements.
    1. Add more log in SyncHostNCVersion.
    2. Remove unnecessary check for MarkIpsAsAvailable.
    3. Auto formatting.

**Reason for Change**:
Logs are not sufficient enough to debug. 

**Issue Fixed**:
After these logs got added, it will unsure whether host version got updated.
In another word, after log got added, we are clear whether select sentence after channel got executed porperly.

### DIFF
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -171,6 +171,7 @@ func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMo
 	}
 	service.RUnlock()
 	if len(hostVersionNeedUpdateNcList) > 0 {
+		logger.Printf("Updating version of the following NC IDs: %v", hostVersionNeedUpdateNcList)
 		ncVersionChannel := make(chan map[string]int)
 		ctxWithTimeout, _ := context.WithTimeout(ctx, syncHostNCTimeoutMilliSec*time.Millisecond)
 		go func() {
@@ -191,8 +192,10 @@ func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMo
 						if channelMode == cns.CRD {
 							service.MarkIpsAsAvailableUntransacted(ncInfo.ID, newHostNCVersion)
 						}
+						oldHostNCVersion := ncInfo.HostVersion
 						ncInfo.HostVersion = strconv.Itoa(newHostNCVersion)
 						service.state.ContainerStatus[ncID] = ncInfo
+						logger.Printf("Updated NC %s host version from %s to %s", ncID, oldHostNCVersion, ncInfo.HostVersion)
 					}
 				}
 				service.Unlock()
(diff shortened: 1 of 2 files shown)


cns/restserver/internalapi.go
```diff
@@ -171,6 +171,7 @@ func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMo
 	}
 	service.RUnlock()
 	if len(hostVersionNeedUpdateNcList) > 0 {
+		logger.Printf("Going to update these NC version list %v", hostVersionNeedUpdateNcList)
```
ramiro-gamarra, line 174: Perhaps a clearer message
```
Updating version of the following NC ids: %v
```

cns/restserver/internalapi.go
```diff
@@ -191,8 +192,10 @@ func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMo
 						if channelMode == cns.CRD {
 							service.MarkIpsAsAvailableUntransacted(ncInfo.ID, newHostNCVersion)
 						}
+						oldHostNCVersion := ncInfo.HostVersion
 						ncInfo.HostVersion = strconv.Itoa(newHostNCVersion)
 						service.state.ContainerStatus[ncID] = ncInfo
+						logger.Printf("Update NC %s host version from %s to %s", ncID, oldHostNCVersion, ncInfo.HostVersion)
```
ramiro-gamarra, line 198: `Updated`

cns/restserver/ipam.go
```diff
@@ -152,9 +152,12 @@ func (service *HTTPRestService) MarkIpsAsAvailableUntransacted(ncID string, newH
 	if ncInfo, exist := service.state.ContainerStatus[ncID]; !exist {
 		logger.Errorf("Can't find NC with ID %s in service state, stop updating its pending programming IP status", ncID)
 	} else {
-		previousHostNCVersion := ncInfo.HostVersion
+		previousHostNCVersion, err := strconv.Atoi(ncInfo.HostVersion)
+		if err != nil {
+			logger.Printf("[MarkIpsAsAvailableUntransacted] Get int value from ncInfo.HostVersion %s failed for %v", ncInfo.HostVersion, err)
```
ramiro-gamarra, line 157: nit: better to use colon than `for`
```
[MarkIpsAsAvailableUntransacted] Get int value from ncInfo.HostVersion %s failed: %v
```
also, if it fails, the rest of the block should not execute, since it would be acting on `previousHostNCVersion` being `0`. Probably should `continue` here

ramiro-gamarra (changes requested):
A test should be added to the function that failed previously

cns/restserver/ipam.go
csfmomo: I'll return here.

ramiro-gamarra (approved):

PR #823, opened 2021-03-18
[NPM] Clearing NPM MARKs on packets before accepting them

vakalapa:
This PR includes:

1. A new chain where NPM will remove the added MARKs and will ACCEPT packets
2. Bailing out when error occurs


npm/pod.go
```diff
@@ -215,36 +215,42 @@
 		npMgr.NsMap[podNs], err = newNs(podNs)
 		if err != nil {
 			metrics.SendErrorLogAndMetric(util.PodID, "[AddPod] Error: failed to create namespace %s with err: %v", podNs, err)
+			return err
```
neaggarwMS, line 218: >return err [](start = 3, length = 10)

I think we need a workerQueue model before we take this to production. Currently for intermittent errors, we can miss events during resync as well. With workQueue controller, we will not mark the entry done until the call succeeds.

neaggarwMS (approved):
:shipit:

PR #825, opened 2021-03-19
[CNS] Debug - Expose In-memory data from HTTPRestService

kmurudi:
Feat:  [CNS] Debug API to expose In-Memory Data HTTPRestService 
This commit has following changes - 
1. Removed AllocatedIPCount field from HTTPRestService struct in restserver.go as it is not used in project.
2. Added debug command getInMemory and API to expose fields-HTTPRestService and 2 fields from IPAMPoolMonitor.
3. Added Test function to test the new debug api response.

**Reason for Change**: Helping in debug commands for CNS service.

**Notes**: Please review and suggest the naming of the command, handler and end point.

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -24,7 +24,7 @@ const (
 	ReleaseIPConfig                          = "/network/releaseipconfig"
 	GetIPAddresses                           = "/debug/getipaddresses"
 	GetPodIPOrchestratorContext              = "/debug/getpodcontext"
-	GetHTTPRestStruct                        = "/debug/gethttpreststruct"
+	GetHTTPRestData                          = "/debug/getrestdata"
 )
 
 // NetworkContainer Prefixes
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -24,6 +24,7 @@ const (
 	ReleaseIPConfig                          = "/network/releaseipconfig"
 	GetIPAddresses                           = "/debug/getipaddresses"
 	GetPodIPOrchestratorContext              = "/debug/getpodcontext"
+	GetHTTPRestStruct                        = "/debug/gethttpreststruct"
 )
 
 // NetworkContainer Prefixes
--- a/cns/api.go
+++ b/cns/api.go
@@ -180,8 +180,10 @@ type IPAMPoolMonitor interface {
 
 //struct to expose state values for IPAMPoolMonitor struct
 type IpamPoolMonitorStateSnapshot struct {
-	MinimumFreeIps int64
-	MaximumFreeIps int64
+	MinimumFreeIps           int64
+	MaximumFreeIps           int64
+	UpdatingIpsNotInUseCount int
+	CachedNNC                nnc.NodeNetworkConfig
 }
 
 // Response describes generic response from CNS.
--- a/cns/api.go
+++ b/cns/api.go
@@ -175,6 +175,13 @@ type NodeConfiguration struct {
 type IPAMPoolMonitor interface {
 	Start(ctx context.Context, poolMonitorRefreshMilliseconds int) error
 	Update(scalar nnc.Scaler, spec nnc.NodeNetworkConfigSpec) error
+	GetStateSnapshot() IpamPoolMonitorStateSnapshot
+}
+
+//struct to expose state values for IPAMPoolMonitor struct
+type IpamPoolMonitorStateSnapshot struct {
+	MinimumFreeIps int64
+	MaximumFreeIps int64
 }
 
 // Response describes generic response from CNS.
--- a/cns/cnsclient/cli.go
+++ b/cns/cnsclient/cli.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/restserver"
 )
 
 const (
@@ -15,7 +16,11 @@ const (
 	getAllocatedArg      = "Allocated"
 	getAllArg            = "All"
 	getPendingReleaseArg = "PendingRelease"
+<<<<<<< HEAD
 	getPodCmdArg         = "getPodContexts"
+=======
+	getInMemoryData      = "getInMemory"
+>>>>>>> da5071c ([CNS] Debug API to expose In-Memory Data HTTPRestService)
 
 	releaseArg = "release"
 
@@ -29,7 +34,11 @@ const (
 var (
 	availableCmds = []string{
 		getCmdArg,
+<<<<<<< HEAD
 		getPodCmdArg,
+=======
+		getInMemoryData,
+>>>>>>> da5071c ([CNS] Debug API to expose In-Memory Data HTTPRestService)
 	}
 
 	getFlags = []string{
@@ -51,8 +60,13 @@ func HandleCNSClientCommands(cmd, arg string) error {
 	switch {
 	case strings.EqualFold(getCmdArg, cmd):
 		return getCmd(cnsClient, arg)
+<<<<<<< HEAD
 	case strings.EqualFold(getPodCmdArg, cmd):
 		return getPodCmd(cnsClient)
+=======
+	case strings.EqualFold(getInMemoryData, cmd):
+		return getInMemory(cnsClient)
+>>>>>>> da5071c ([CNS] Debug API to expose In-Memory Data HTTPRestService)
 	default:
 		return fmt.Errorf("No debug cmd supplied, options are: %v", getCmdArg)
 	}
@@ -101,13 +115,20 @@ func printIPAddresses(addrSlice []cns.IPConfigurationStatus) {
 	}
 }
 
+<<<<<<< HEAD
 func getPodCmd(client *CNSClient) error {
 
 	resp, err := client.GetPodOrchestratorContext()
+=======
+func getInMemory(client *CNSClient) error {
+
+	inmemoryData, err := client.GetHTTPServiceData()
+>>>>>>> da5071c ([CNS] Debug API to expose In-Memory Data HTTPRestService)
 	if err != nil {
 		return err
 	}
 
+<<<<<<< HEAD
 	printPodContext(resp)
 	return nil
 }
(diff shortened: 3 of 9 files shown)


cns/NetworkContainerContract.go
```diff
@@ -23,6 +23,7 @@ const (
 	RequestIPConfig                          = "/network/requestipconfig"
 	ReleaseIPConfig                          = "/network/releaseipconfig"
 	GetIPAddresses                           = "/debug/getipaddresses"
+	GetHTTPRestData                        = "/debug/getrestdata"
```
ramiro-gamarra, line 26: The name `getrestdata` is a bit generic. If the purpose of this endpoint is to return ip state, maybe the endpoint can be named `getipstate` or if its for a more generic placeholder, just `state`. Thoughts @neaggarwMS? Formatting is also off. You can enable goimports on save if you are using Goland.

cns/api.go
```diff
@@ -45,6 +45,7 @@ type HTTPService interface {
 	GetPendingReleaseIPConfigs() []IPConfigurationStatus
 	GetPodIPConfigState() map[string]IPConfigurationStatus
 	MarkIPAsPendingRelease(numberToMark int) (map[string]IPConfigurationStatus, error)
+	//GetIPAMPoolMonitorVars() IPAMPoolMonitor
```
ramiro-gamarra, line 48: remove dead code

cns/api.go
```diff
@@ -175,6 +176,13 @@ type NodeConfiguration struct {
 type IPAMPoolMonitor interface {
 	Start(ctx context.Context, poolMonitorRefreshMilliseconds int) error
 	Update(scalar nnc.Scaler, spec nnc.NodeNetworkConfigSpec) error
+	IPAMPoolMonitorFreeIps() IPAMPoolMonitorFreeIps
```
ramiro-gamarra, line 179: I think there will be more state that can be surfaced through here. I think to convey proper intent, this could be renamed to `GetStateSnapshot()` and the struct below to `IpamPoolMonitorStateSnapshot`. This way, more data can be added without making the intention of the struct ambiguous.

cns/cnsclient/cli.go
```diff
@@ -15,6 +16,7 @@ const (
 	getAllocatedArg      = "Allocated"
 	getAllArg            = "All"
 	getPendingReleaseArg = "PendingRelease"
+    getInMemoryData      = "getInMemory"
```
ramiro-gamarra, line 19: formatting

cns/cnsclient/cli.go
```diff
@@ -96,3 +101,20 @@ func printIPAddresses(addrSlice []cns.IPConfigurationStatus) {
...
+	}
+
+	printInMemoryStruct(inmemoryData.HttpRestServiceData)
+	return nil
+}
+
+func printInMemoryStruct(data restserver.HttpRestServiceData){
+	 fmt.Println("PodIPIDByOrchestratorContext : ", data.PodIPIDByOrchestratorContext)
+	 fmt.Println("PodIPConfigState : ", data.PodIPConfigState)
+	 fmt.Println("IPAMPoolMonitor : ", data.IPAMPoolMonitor)
```
ramiro-gamarra, line 119: nit: no space before colon

cns/cnsclient/cnsclient.go
```diff
@@ -371,3 +371,51 @@ func (cnsClient *CNSClient) GetIPAddressesMatchingStates(StateFilter ...string)
...
+
+	payload := ""
+
+	err = json.NewEncoder(&body).Encode(payload)
+	if err != nil {
+		log.Errorf("encoding json failed with %v", err)
+		return resp, err
+	}
+
+	res, err = http.Post(url, contentTypeJSON, &body)
```
ramiro-gamarra, line 395: Shouldn't this be a GET endpoint?

cns/cnsclient/cnsclient_test.go
```diff
@@ -242,8 +236,47 @@ func TestCNSClientRequestAndRelease(t *testing.T) {
...
+	if err1 != nil {
+		t.Fatalf("get IP from CNS failed with %+v", err1)
+	}
+
+	//test for debug api/cmd to get inmemory data from HTTPRestService
+	inmemory, err := cnsClient.GetHTTPServiceData()
+	if err != nil {
+		t.Fatalf("Get in-memory http REST Struct failed %+v", err)
+	}
+	fmt.Println(inmemory.HttpRestServiceData)
```
ramiro-gamarra, line 280: Should make some assertions on the response. What specific data do you expect in `inmemory`?

cns/restserver/ipam.go
```diff
@@ -190,6 +190,49 @@ func (service *HTTPRestService) GetPodIPConfigState() map[string]cns.IPConfigura
...
+func (service *HTTPRestService) GetHTTPRestDataHandler(w http.ResponseWriter, r *http.Request) {
+	var (
+		req           string
+		resp          GetHTTPServiceDataResponse
+		statusCode    int
+		returnMessage string
+		err           error
+	)
+
+	statusCode = UnexpectedError
```
ramiro-gamarra, line 202: this is only used when there is an error from the response. no need to keep an extra variable, just 
```go
if err != nil {
    resp.Response.ReturnCode = UnexpectedError
    ....
}
```

cns/restserver/ipam.go
```diff
@@ -264,7 +307,12 @@ func filterIPConfigs(toBeAdded map[string]cns.IPConfigurationStatus, f func(cns.
 	}
 	return vsf
 }
-
+/*
+func (service *HTTPRestService) GetIPAMPoolMonitorVars() cns.IPAMPoolMonitor{
+	  service.RLock()
+	  defer service.Unlock()
+	  return
+}*/
```
ramiro-gamarra, line 315: remove dead code

cns/cnsclient/cli.go
kmurudi: I thought giving the space would make the CLI o/p more readable, but makes sense

cns/cnsclient/cli.go
kmurudi: resolved

cns/api.go
kmurudi: makes sense, I was confused as to what to name so this was temp, but the StateSnapshot seems right to convey that. Have refactored.

cns/restserver/ipam.go
kmurudi: ack, addressed this

cns/cnsclient/cnsclient_test.go
kmurudi: added few checks on this function

cns/NetworkContainerContract.go
kmurudi: Yes, for the CLI command I have used `getInMemory` after we decide on the end-point name, I will change this too.

cns/cnsclient/cnsclient.go
kmurudi: Made the changes in recent commit

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -249,3 +248,14 @@ func (pm *CNSIPAMPoolMonitor) Update(scalar nnc.Scaler, spec nnc.NodeNetworkConf
 
 	return nil
 }
+
+//this function sets the values for state in IPAMPoolMonitor Struct
+func (pm *CNSIPAMPoolMonitor) GetStateSnapshot() cns.IpamPoolMonitorStateSnapshot{
+	defer pm.mu.Unlock()
+	pm.mu.Lock()
```
ramiro-gamarra, line 255: This works as intended because scope is limited, and I can see its the style in the rest of the file, but for future reference, it is not a good practice to defer a closing/unlocking action before a lock has been taken/something has been opened.

cns/restserver/ipam.go
```diff
@@ -190,6 +190,38 @@ func (service *HTTPRestService) GetPodIPConfigState() map[string]cns.IPConfigura
...
 
+func (service *HTTPRestService) GetHTTPRestDataHandler(w http.ResponseWriter, r *http.Request) {
+	var (
+		resp          GetHTTPServiceDataResponse
+		returnMessage string
+		err           error
+	)
+
+	defer func() {
+		if err != nil {
```
ramiro-gamarra, line 201: i just made a similar comment in another file (apologies for not catching and discussing earlier), but this will never be nil in this line. function can be made simpler

ramiro-gamarra (approved):
I think this is fine to merge, but let's discuss the comments left first.

cns/ipampoolmonitor/ipampoolmonitor.go
kmurudi: ack, this makes sense, I can put the defer after the lock is called.

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -249,3 +248,14 @@
...
+
+	return cns.IpamPoolMonitorStateSnapshot {
+		MinimumFreeIps: pm.MinimumFreeIps,
```
neaggarwMS, line 258: Can  you also return

 cachedNNC   nnc.NodeNetworkConfig
 updatingIpsNotInUseCount int

neaggarwMS (changes requested):
:clock1:

cns/ipampoolmonitor/ipampoolmonitor.go
kmurudi: @neaggarwMS  yes I have added them too to IPAMPoolMonitor in the latest commit, please review if that looks good and the CLI command/end-point we should name for this debug API.

neaggarwMS (approved):
:shipit:

cns/cnsclient/cnsclient_test.go
```diff
@@ -282,4 +316,80 @@ func TestCNSClientPodContextApi(t *testing.T) {
...
+	}
+}
+
+func TestCNSClientDebugAPI(t *testing.T) {
+	podName := "testpodname"
+	podNamespace := "testpodnamespace"
+	desiredIpAddress := "10.0.0.5"
+
+	secondaryIps := make([]string, 0)
+	secondaryIps = append(secondaryIps, desiredIpAddress)
```
ramiro-gamarra, line 333: can just 
```go
secondaryIps := []string{desiredIpAddress}
```

ramiro-gamarra (approved):

ramiro-gamarra (approved):

PR #826, opened 2021-03-19
[NPM] pushing Netpol cache update below errors

vakalapa:
Right now, network policy cache is updated even if there are errors below, this PR will make sure that cache is not updated if error occurs


neaggarwMS (dismissed):
:shipit:

neaggarwMS (approved):
:shipit:

matmerr (approved):

PR #836, opened 2021-03-27
[NPM] Support pod controller and unit test to improve reliability 

JungukCho:
This PR is to support pod controller in npm to improve reliability according to pod events.
For example, when event handling was failed, we need to retry them.

* code layout to support pod controller in npm

* filter events if they do not need to handle and clean up business logic

* put lock when it needs to access shared resource
use namespace/name as key
clean up functions

* - use pod key instead of uid.
- remove unnecessary error check

* use namespace prefix in pod namespace. add log messages to know what events happens.

* move event logs with more contexts in needsync func

* Put returning an error in a right place

* Return if the RV of both pod obj are the same. Proactively start cleaning up pod when the pod is deleted.

* first version of podController UT

* add ipset management in pod controller unit test

* Make methods flexible for ipset store and restore operation

* clean up functions and variables

Co-authored-by: Junguk Cho <jungukcho@microsoft.com>

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


npm/podController.go
```diff
@@ -0,0 +1,634 @@
...
+		}
+		return err
+	}
+
+	if pod.DeletionTimestamp != nil || pod.DeletionGracePeriodSeconds != nil {
+		err = c.cleanUpDeletedPod(pod)
+		if err != nil {
+			metrics.SendErrorLogAndMetric(util.PodID, "Error: cannot delete ipset due to %v when DeletionTimestamp or DeletionGracePeriodSeconds is not nil", err)
+			return fmt.Errorf("Failed to clean up pod due to %s when DeletionTimestamp or DeletionGracePeriodSeconds is not nil\n", err.Error())
+		}
```
vakalapa, line 339: We will need to return here, or else we will be cleaning up and adding again

JungukCho: applied

npm/podController_test.go
```diff
@@ -0,0 +1,372 @@
...
+	f.kubeobjects = append(f.kubeobjects, podObj)
+
+	f.newPodController()
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+	f.kubeInformer.Start(stopCh)
+
+	// simulate pod add event and add pod object to sharedInformer cache
+	f.podController.addPod(podObj)
+	if podObj.Spec.HostNetwork {
```
vakalapa, line 160: Instead of doing checks on the hostnetpod and returning, can we check the workqueue length like below 
f.podController.workqueue.Len()

npm/podController_test.go
```diff
@@ -0,0 +1,372 @@
...
+func deletePod(t *testing.T, f *podFixture, podObj *corev1.Pod) {
+	addPod(t, f, podObj)
+	t.Logf("Complete add pod event")
+
+	// simulate pod delete event and delete pod object from sharedInformer cache
+	f.kubeInformer.Core().V1().Pods().Informer().GetIndexer().Delete(podObj)
+
+	// call deletePod event
+	f.podController.deletePod(podObj)
+	if !podObj.Spec.HostNetwork {
```
vakalapa, line 202: same as above check len of WQ

npm/podController_test.go
```diff
@@ -0,0 +1,372 @@
...
+	f.podController.deletePod(oldPodObj)
+	if !oldPodObj.Spec.HostNetwork {
+		err := f.podController.syncPod(getKey(oldPodObj, t))
+		if err != nil {
+			f.t.Errorf("error syncing pod: %v", err)
+		}
+	}
+
+	// TODO: add more checking list and clean up with tables with expected values.
+	if newPodObj.Spec.HostNetwork {
```
vakalapa, line 256: Since this is being check multiple times, we can have check function like below

func checkCachemapStatus(maplen = len(Pod/NS map), newPodObj.Spec.HostNetwork, a=1, b=2) {
switch newPodObj.Spec.HostNetwork {
case true:
  if maplen != a
 return
case false:
 if maplen != b
retrun
}
}

npm/podController_test.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+	status = newPodStatus("Running", "4.3.2.1")
+	newPodObj := newPod("new-test-pod", "test-namespace", newRV, labels, status)
+	newPodObj.Spec.HostNetwork = isHostNetwork
+
+	updatePod(t, f, oldPodObj, newPodObj)
+}
+
+func TestUpdateHostNetworkPod(t *testing.T) {
+}
```
vakalapa, line 360: Is there a reason to leave this blank ?

npm/podController.go
```diff
@@ -0,0 +1,627 @@
...
+	defer c.npMgr.Unlock()
+
+	cachedNpmPodObj, npmPodExists := c.npMgr.PodMap[key]
+	if !npmPodExists {
+		return
+	}
+
+	// if the podIp exists, it must match the cachedIp
+	// (TODO): when does this situation happen? Is it ok to add it to workqueue?
+	if len(podObj.Status.PodIP) > 0 && cachedNpmPodObj.PodIP != podObj.Status.PodIP {
```
vakalapa, line 231: we should add it to the WQ, but we will need to have a similar check, where we are deleting the cachedNpmPodobj and not the new PodIP.

this will happen if a Pod NSA/A is deleted but NPM missed the event, and then NSA/A is created with a new IP and is deleted again. This may not be an issue in WQ's scenario, but is a good check to have where we are cleaning up the cachedPod objs and not the state given to us

npm/podController.go
```diff
@@ -0,0 +1,627 @@
...
+	npmPodObj := newNpmPod(podObj)
+	podNs := util.GetNSNameWithPrefix(podObj.Namespace)
+	podKey, _ := cache.MetaNamespaceKeyFunc(podObj)
+	ipsMgr := c.npMgr.NsMap[util.KubeAllNamespacesFlag].IpsMgr
+	log.Logf("POD CREATING: [%s%s/%s/%s%+v%s]", npmPodObj.PodUID, podNs, npmPodObj.Name, npmPodObj.NodeName, npmPodObj.Labels, npmPodObj.PodIP)
+
+	// Add pod namespace if it doesn't exist
+	var err error
+	if _, exists := c.npMgr.NsMap[podNs]; !exists {
+		// (TODO): Any chance to return error?
```
vakalapa, line 360: we should handle the error here

npm/podController.go
```diff
@@ -0,0 +1,627 @@
...
+
+	podKey, _ := cache.MetaNamespaceKeyFunc(newPodObj)
+	newPodObjNs := util.GetNSNameWithPrefix(newPodObj.ObjectMeta.Namespace)
+	ipsMgr := c.npMgr.NsMap[util.KubeAllNamespacesFlag].IpsMgr
+
+	// Add pod namespace if it doesn't exist
+	var err error
+	if _, exists := c.npMgr.NsMap[newPodObjNs]; !exists {
+		// (TODO): Any chance to return error?
+		c.npMgr.NsMap[newPodObjNs], _ = newNs(newPodObjNs)
```
vakalapa, line 405: same should handle here too

npm/podController_test.go
```diff
@@ -0,0 +1,372 @@
...
+		}
+	}
+
+	// simulate pod delete event and delete pod object from sharedInformer cache
+	f.kubeInformer.Core().V1().Pods().Informer().GetIndexer().Delete(oldPodObj)
+
+	// call deletePod event
+	f.podController.deletePod(oldPodObj)
+	if !oldPodObj.Spec.HostNetwork {
+		err := f.podController.syncPod(getKey(oldPodObj, t))
```
vakalapa, line 249: Call processNextitem instead of syncPod. that way we are checking WQ logic

npm/podController_test.go
```diff
@@ -0,0 +1,372 @@
...
+			f.t.Errorf("error syncing pod: %v", err)
+		}
+	}
+
+	// TODO: add more checking list and clean up with tables with expected values.
+	if newPodObj.Spec.HostNetwork {
+		if len(f.npMgr.PodMap) != 0 {
+			t.Errorf("updatePod failed @ PodMap length check -current len %d %v", len(f.npMgr.PodMap), f.npMgr.PodMap)
+		}
+	} else {
```
vakalapa, line 260: Move these checks to specific testcase.

npm/podController_test.go
```diff
@@ -0,0 +1,372 @@
...
+			f.t.Errorf("error syncing pod: %v", err)
+		}
+	}
+
+	// simulate pod delete event and delete pod object from sharedInformer cache
+	f.kubeInformer.Core().V1().Pods().Informer().GetIndexer().Delete(oldPodObj)
+
+	// call deletePod event
+	f.podController.deletePod(oldPodObj)
+	if !oldPodObj.Spec.HostNetwork {
```
vakalapa, line 248: instead of this check the length of WQ and call processNextItem

npm/podController.go
JungukCho: This comment is about "newNs()", not c.npMgr.NsMap. newNs function always return nil. So it may be better to remove "return error" in newNs() function.

npm/podController_test.go
JungukCho: Good idea. applied.

npm/podController_test.go
JungukCho: Applied with the ideas.

npm/podController_test.go
JungukCho: applied.

npm/podController_test.go
JungukCho: applied.

npm/podController_test.go
JungukCho: applied.

npm/podController.go
```diff
@@ -0,0 +1,623 @@
...
+		return fmt.Errorf("Failed to sync pod due to %v\n", err)
+	}
+
+	return nil
+}
+
+func (c *podController) syncAddedPod(podObj *corev1.Pod) error {
+	npmPodObj := newNpmPod(podObj)
+	podNs := util.GetNSNameWithPrefix(podObj.Namespace)
+	podKey, _ := cache.MetaNamespaceKeyFunc(podObj)
```
matmerr, line 371: utilruntime.HandleError ?

npm/podController.go
```diff
@@ -0,0 +1,623 @@
...
+// syncAddAndUpdatePod handles updating pod ip in its label's ipset.
+func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
+	log.Logf("[syncAddAndUpdatePod]")
+
+	podKey, _ := cache.MetaNamespaceKeyFunc(newPodObj)
+	newPodObjNs := util.GetNSNameWithPrefix(newPodObj.ObjectMeta.Namespace)
+	ipsMgr := c.npMgr.NsMap[util.KubeAllNamespacesFlag].IpsMgr
+
+	// Add pod namespace if it doesn't exist
+	var err error
```
matmerr, line 420: this paradigm will need to change in the future when using NPMError

npm/podController.go
```diff
@@ -0,0 +1,623 @@
...
+	needSync := false
+	var key string
+
+	podObj, ok := obj.(*corev1.Pod)
+	if !ok {
+		metrics.SendErrorLogAndMetric(util.NpmID, "ADD Pod: Received unexpected object type: %v", obj)
+		return key, needSync
+	}
+
+	log.Logf("[POD %s EVENT for %s in %s", eventType, podObj.Name, podObj.Namespace)
```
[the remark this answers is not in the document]
JungukCho: ack

npm/podController.go
JungukCho: Yes. It will be handled with another PR.

vakalapa (approved):
lgtm

vakalapa (approved):

npm/podController.go
JungukCho: This seems a convention in K8s controller implementation. So we also follow it. In high level, this method is preferable to logging the error (https://pkg.go.dev/k8s.io/apimachinery/pkg/util/runtime#HandleError)

PR #838, opened 2021-03-29
Aks swift test update

pjohnst5:
**Reason for Change**:
Updating the AKS-Swift tests to use a custom underlay in our runners sub
Also makes cni-manager changes like making transparent mode default


matmerr (approved):
lgtm!

PR #842, opened 2021-03-31
Fixing a regression with integration of two controllers

vakalapa:
After adding two controller and resolving conflicts through web editor, it left some dangling imports resulting in failures.


JungukCho (approved):
looks good to me.

PR #843, opened 2021-04-01
feature: don't let CNS ask for more IPS than node limit

pjohnst5:
**Reason for Change**:
CNS would ask for more IPS than the node limit, so making CNS stop asking once the node limit has been reached
Also ensures CNS asks for a multiple of the batch size on scale down after reaching the node limit


cns/fakes/requestcontrollerfake.go
```diff
@@ -7,15 +7,18 @@ import (
...
 
 type RequestControllerFake struct {
 	fakecns   *HTTPServiceFake
 	cachedCRD nnc.NodeNetworkConfig
 	ip        net.IP
+	node corev1.Node
 }
 
 func NewRequestControllerFake(cnsService *HTTPServiceFake, scalar nnc.Scaler, subnetAddressSpace string, numberOfIPConfigs int) *RequestControllerFake {
+	// TODO: use maxPodIPCount to populate CRD max ip count
```
neaggarwMS, line 21: Can you add it or create a task to finish this?

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -72,12 +72,17 @@ func (pm *CNSIPAMPoolMonitor) Reconcile() error {
...
-		cnsPodIPConfigCount, pm.cachedNNC.Spec.RequestedIPCount, pm.scalarUnits.BatchSize, pm.MinimumFreeIps, pm.MaximumFreeIps, allocatedPodIPCount, availableIPConfigCount, pendingReleaseIPCount, freeIPConfigCount, pendingProgramCount)
+	msg := fmt.Sprintf("[ipam-pool-monitor] Pool Size: %v, Goal Size: %v, BatchSize: %v, MaxIPCount: %v, MinFree: %v, MaxFree:%v, Allocated: %v, Available: %v, Pending Release: %v, Free: %v, Pending Program: %v",
+		cnsPodIPConfigCount, pm.cachedNNC.Spec.RequestedIPCount, pm.scalarUnits.BatchSize, pm.scalarUnits.MaxIPCount, pm.MinimumFreeIps, pm.MaximumFreeIps, allocatedPodIPCount, availableIPConfigCount, pendingReleaseIPCount, freeIPConfigCount, pendingProgramCount)
 
 	switch {
 	// pod count is increasing
 	case freeIPConfigCount < pm.MinimumFreeIps:
+		if pm.cachedNNC.Spec.RequestedIPCount == pm.scalarUnits.MaxIPCount {
+			// If we're already at the maxIpCount, don't try to increase
+			return nil
```
neaggarwMS, line 83: Can you add a log line here?

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -112,7 +117,23 @@ func (pm *CNSIPAMPoolMonitor) increasePoolSize() error {
 		return err
 	}
 
+	// Query the max ip count
+	maxIpCount:= pm.scalarUnits.MaxIPCount
```
neaggarwMS, line 121: How will handle you existing clusters where NNC is not updated. @thatmattlong  ?

cns/ipampoolmonitor/ipampoolmonitor_test.go
```diff
@@ -395,3 +434,51 @@ func TestPoolSizeDecreaseToReallyLow(t *testing.T) {
 		t.Fatalf("Expected IP's not in use to be 0 after reconcile, expected %v, actual %v", (initialIPConfigCount - batchSize), len(poolmonitor.cachedNNC.Spec.IPsNotInUse))
 	}
 }
+
+func TestDecreaseAfterNodeLimitReached(t *testing.T) {
```
neaggarwMS, line 438: Awesome, for adding right test coverage

neaggarwMS, line 438: Can we also add a test for increase where maxPodIPCount is not a multiple of batchSize?

cns/ipampoolmonitor/ipampoolmonitor.go
thatmattlong: For cases where max pods is not set it will unmarshal as 0, so @pjohnst5 let's take max to be 250 if `pm.scalarUnits.MaxIPCount == 0`

cns/ipampoolmonitor/ipampoolmonitor.go
pjohnst5: If I add a log line here, it clutters the logs very much so, every couple seconds, the line would be printed, and I think a lot of info would be lost through it

cns/ipampoolmonitor/ipampoolmonitor.go
pjohnst5: Ok

cns/ipampoolmonitor/ipampoolmonitor_test.go
pjohnst5: Thanks

pjohnst5: 1. We already have that, check `TestPoolIncreasePastNodeLimit`, BatchSize = 16, MaxPodIpCount = 30
2. I think you mean where BatchSize > MaxPodIpCount? If so I'll add it in

pjohnst5: Added

cns/ipampoolmonitor/ipampoolmonitor.go
pjohnst5: I did add a log statment though once we hit the max for the first time, so that will be in the logs

neaggarwMS (approved):
:shipit:

PR #848, opened 2021-04-02
[NPM] better error handling and cache building

vakalapa:
Today, during a resource event handling, NPM will return immediately if it encounters an error and local NPM cache will not be updated. This results in leaked ipset/iptables state in kernel which will never be cleaned up since the local NPM cache is not aware of it. 

Solution:
Build cache after each successful operation. Return on first error.

This PR, incorporates the above logic, NPM now keeps updating the local cache after every successful operation. NPM also ignores the notexists errors in Delete set operations.
Logging is updated to use klog instead of using acn/log.
if the Cached Pod IP changes, now NPM will delete the existing cached pod and adds a new pod.
CleanUpDeletedPod will now only require the key of the pod which needs to be deleted.

### DIFF
--- a/npm/ipsm/ipsm.go
+++ b/npm/ipsm/ipsm.go
@@ -188,8 +188,10 @@ func (ipsMgr *IpsetManager) DeleteFromList(listName string, setName string) erro
 	exists, _ := ipsMgr.SetExists(setName)
 
 	// if set does not exist, then return because the ipset call will fail due to set not existing
+	// TODO make sure these are info and not errors, use NPmErr
 	if !exists {
-		return fmt.Errorf("Set [%s] does not exist when attempting to delete from list [%s]", setName, listName)
+		metrics.SendErrorLogAndMetric(util.IpsmID, "Set [%s] does not exist when attempting to delete from list [%s]", setName, listName)
+		return nil
 	}
 
 	//Check if list being added exists in the listmap, if it exists we don't care about the set type
@@ -197,11 +199,13 @@ func (ipsMgr *IpsetManager) DeleteFromList(listName string, setName string) erro
 
 	// if set does not exist, then return because the ipset call will fail due to set not existing
 	if !exists {
-		return fmt.Errorf("Set [%s] does not exist when attempting to add to list [%s]", setName, listName)
+		metrics.SendErrorLogAndMetric(util.IpsmID, "Set [%s] does not exist when attempting to add to list [%s]", setName, listName)
+		return nil
 	}
 
 	if listtype != util.IpsetSetListFlag {
-		return fmt.Errorf("Set [%s] is of the wrong type when attempting to delete list [%s], actual type [%s]", setName, listName, listtype)
+		metrics.SendErrorLogAndMetric(util.IpsmID, "Set [%s] is of the wrong type when attempting to delete list [%s], actual type [%s]", setName, listName, listtype)
+		return nil
 	}
 
 	if _, exists := ipsMgr.ListMap[listName]; !exists {
@@ -552,7 +556,6 @@ func (ipsMgr *IpsetManager) DestroyNpmIpsets() error {
 		return nil
 	}
 
-	log.Logf("{DestroyNpmIpsets} Reply from command %s executed is %s", cmdName+" "+cmdArgs, reply)
 	re := regexp.MustCompile("Name: (" + util.AzureNpmPrefix + "\\d+)")
 	ipsetRegexSlice := re.FindAllSubmatch(reply, -1)
 
--- a/npm/nameSpaceController.go
+++ b/npm/nameSpaceController.go
@@ -7,7 +7,6 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/Azure/azure-container-networking/log"
 	"github.com/Azure/azure-container-networking/npm/ipsm"
 	"github.com/Azure/azure-container-networking/npm/iptm"
 	"github.com/Azure/azure-container-networking/npm/metrics"
@@ -23,22 +22,22 @@ import (
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
 )
 
 type LabelAppendOperation bool
 
 const (
-	ClearExitingLabels    LabelAppendOperation = true
-	AppendToExitingLabels LabelAppendOperation = false
+	ClearExistingLabels    LabelAppendOperation = true
+	AppendToExistingLabels LabelAppendOperation = false
 )
 
 type Namespace struct {
-	name            string
-	LabelsMap       map[string]string // NameSpace labels
-	SetMap          map[string]string
-	IpsMgr          *ipsm.IpsetManager
-	iptMgr          *iptm.IptablesManager
-	resourceVersion uint64 // NameSpace ResourceVersion
+	name      string
+	LabelsMap map[string]string // NameSpace labels
+	SetMap    map[string]string
+	IpsMgr    *ipsm.IpsetManager
+	iptMgr    *iptm.IptablesManager
 }
 
 // newNS constructs a new namespace object.
@@ -49,9 +48,6 @@ func newNs(name string) (*Namespace, error) {
 		SetMap:    make(map[string]string),
 		IpsMgr:    ipsm.NewIpsetManager(),
 		iptMgr:    iptm.NewIptablesManager(),
-		// resource version is converted to uint64
-		// so make sure it is initialized to "0"
-		resourceVersion: 0,
 	}
 
 	return ns, nil
@@ -64,11 +60,6 @@ func (nsObj *Namespace) getNamespaceObjFromNsObj() *corev1.Namespace {
 			Labels: nsObj.LabelsMap,
 		},
 	}
-}
-
-// setResourceVersion setter func for RV
-func (nsObj *Namespace) setResourceVersion(rv string) {
-	nsObj.resourceVersion = util.ParseResourceVersion(rv)
 }
 
 func (nsObj *Namespace) appendLabels(new map[string]string, clear LabelAppendOperation) {
@@ -134,7 +125,7 @@ func (nsc *nameSpaceController) needSync(obj interface{}, event string) (string,
 		return key, needSync
 	}
 
-	log.Logf("[NAMESPACE %s EVENT] for namespace [%s]", event, key)
+	klog.Infof("[NAMESPACE %s EVENT] for namespace [%s]", event, key)
 
 	needSync = true
 	return key, needSync
@@ -143,7 +134,7 @@ func (nsc *nameSpaceController) needSync(obj interface{}, event string) (string,
 func (nsc *nameSpaceController) addNamespace(obj interface{}) {
 	key, needSync := nsc.needSync(obj, "ADD")
 	if !needSync {
-		log.Logf("[NAMESPACE ADD EVENT] No need to sync this namespace [%s]", key)
+		klog.Infof("[NAMESPACE ADD EVENT] No need to sync this namespace [%s]", key)
 		return
 	}
 	nsc.workqueue.Add(key)
@@ -152,7 +143,7 @@ func (nsc *nameSpaceController) addNamespace(obj interface{}) {
 func (nsc *nameSpaceController) updateNamespace(old, new interface{}) {
 	key, needSync := nsc.needSync(new, "UPDATE")
 	if !needSync {
-		log.Logf("[NAMESPACE UPDATE EVENT] No need to sync this namespace [%s]", key)
+		klog.Infof("[NAMESPACE UPDATE EVENT] No need to sync this namespace [%s]", key)
 		return
 	}
 
@@ -160,7 +151,7 @@ func (nsc *nameSpaceController) updateNamespace(old, new interface{}) {
 	oldNsObj, ok := old.(*corev1.Namespace)
 	if ok {
 		if oldNsObj.ResourceVersion == nsObj.ResourceVersion {
-			log.Logf("[NAMESPACE UPDATE EVENT] Resourceversion is same for this namespace [%s]", key)
+			klog.Infof("[NAMESPACE UPDATE EVENT] Resourceversion is same for this namespace [%s]", key)
 			return
 		}
 	}
@@ -172,7 +163,7 @@ func (nsc *nameSpaceController) updateNamespace(old, new interface{}) {
 	cachedNsObj, nsExists := nsc.npMgr.NsMap[nsKey]
 	if nsExists {
 		if reflect.DeepEqual(cachedNsObj.LabelsMap, nsObj.ObjectMeta.Labels) {
-			log.Logf("[NAMESPACE UPDATE EVENT] Namespace [%s] labels did not change", key)
+			klog.Infof("[NAMESPACE UPDATE EVENT] Namespace [%s] labels did not change", key)
 			return
 		}
 	}
@@ -213,7 +204,7 @@ func (nsc *nameSpaceController) deleteNamespace(obj interface{}) {
 	nsKey := util.GetNSNameWithPrefix(key)
 	_, nsExists := nsc.npMgr.NsMap[nsKey]
 	if !nsExists {
-		log.Logf("[NAMESPACE DELETE EVENT] Namespace [%s] does not exist in case, so returning", key)
+		klog.Infof("[NAMESPACE DELETE EVENT] Namespace [%s] does not exist in case, so returning", key)
 		return
 	}
 
@@ -224,16 +215,16 @@ func (nsc *nameSpaceController) Run(threadiness int, stopCh <-chan struct{}) err
 	defer utilruntime.HandleCrash()
 	defer nsc.workqueue.ShutDown()
 
-	log.Logf("Starting Namespace controller\n")
-	log.Logf("Starting workers")
+	klog.Info("Starting Namespace controller\n")
+	klog.Info("Starting workers")
 	// Launch workers to process namespace resources
 	for i := 0; i < threadiness; i++ {
 		go wait.Until(nsc.runWorker, time.Second, stopCh)
 	}
 
-	log.Logf("Started workers")
+	klog.Info("Started workers")
 	<-stopCh
-	log.Logf("Shutting down workers")
+	klog.Info("Shutting down workers")
 
 	return nil
 }
@@ -274,7 +265,7 @@ func (nsc *nameSpaceController) processNextWorkItem() bool {
 		// Finally, if no error occurs we Forget this item so it does not
 		// get queued again until another change happens.
 		nsc.workqueue.Forget(obj)
-		log.Logf("Successfully synced '%s'", key)
+		klog.Infof("Successfully synced '%s'", key)
 		return nil
 	}(obj)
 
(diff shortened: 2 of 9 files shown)


npm/ipsm/ipsm.go
```diff
@@ -188,20 +188,24 @@ func (ipsMgr *IpsetManager) DeleteFromList(listName string, setName string) erro
 	exists, _ := ipsMgr.SetExists(setName)
 
 	// if set does not exist, then return because the ipset call will fail due to set not existing
+	// TODO make sure these are info and not errors, use NPmErr
 	if !exists {
-		return fmt.Errorf("Set [%s] does not exist when attempting to delete from list [%s]", setName, listName)
+		metrics.SendErrorLogAndMetric(util.IpsmID, "Set [%s] does not exist when attempting to delete from list [%s]", setName, listName)
```
JungukCho, line 193: (Log collection question) Do we need to collect these information? If it is ignorable case (especially, it returns "nil"), are there special reason to send this log?

npm/nameSpaceController.go
```diff
@@ -385,6 +410,13 @@ func (nsc *nameSpaceController) syncUpdateNameSpace(newNsObj *corev1.Namespace)
 			metrics.SendErrorLogAndMetric(util.NSID, "[UpdateNamespace] Error: failed to delete namespace %s from ipset list %s with err: %v", newNsName, labelKey, err)
 			return err
 		}
+		// {IMPORTANT} The order of compared list will be key and then key+val. NPM should only append after both key
+		// key + val ipsets are worked on. 0th index will be key and 1st index will be value of the label
+		// (TODO) need to remove this ordering dependency
```
JungukCho, line 415: Can we resolve this issue in this PR? As I understood this issue and we discussed this before, but if possible, it is better to resolve this issue to avoid future possible issue.

npm/nameSpaceController.go
```diff
@@ -395,10 +427,16 @@ func (nsc *nameSpaceController) syncUpdateNameSpace(newNsObj *corev1.Namespace)
 			metrics.SendErrorLogAndMetric(util.NSID, "[UpdateNamespace] Error: failed to add namespace %s to ipset list %s with err: %v", newNsName, labelKey, err)
 			return err
 		}
+		// {IMPORTANT} Same as above order is assumed to be key and then key+val. NPM should only append to existing labels
```
JungukCho, line 430: Same. Can we resolve this dependency?

npm/nameSpaceController.go
JungukCho: current GetLabelKVFromSet returns key and value. So better to returns (string, string) in the GetLabelKVFromSet 
 function and use "removedKey" and "removedValue" in this function

npm/nameSpaceController.go
```diff
@@ -395,10 +427,16 @@ func (nsc *nameSpaceController) syncUpdateNameSpace(newNsObj *corev1.Namespace)
...
+		// only after both ipsets for a given label's key value pair are added successfully
+		addedLabel := util.GetLabelKVFromSet(nsLabelVal)
+		if len(addedLabel) > 1 {
+			curNsObj.appendLabels(map[string]string{addedLabel[0]: addedLabel[1]}, AppendToExitingLabels)
+		}
 	}
 
 	// Append all labels to the cache NS obj
-	curNsObj.LabelsMap = util.ClearAndAppendMap(curNsObj.LabelsMap, newNsLabel)
+	curNsObj.appendLabels(newNsLabel, ClearExitingLabels)
```
JungukCho, line 439: Does it overwrite items(434 line)? When do we need ClearExitingLabels?

npm/nameSpaceController.go
```diff
@@ -395,10 +427,16 @@ func (nsc *nameSpaceController) syncUpdateNameSpace(newNsObj *corev1.Namespace)
...
 	curNsObj.setResourceVersion(newNsObj.GetObjectMeta().GetResourceVersion())
```
JungukCho, line 440: Do we use resource version?

npm/nameSpaceController.go
```diff
@@ -418,15 +456,24 @@ func (nsc *nameSpaceController) cleanDeletedNamespace(nsName string, nsLabel map
...
 	ipsMgr := nsc.npMgr.NsMap[util.KubeAllNamespacesFlag].IpsMgr
-	nsLabels := util.GetIPSetListFromLabels(cachedNsObj.LabelsMap)
 	// Delete the namespace from its label's ipset list.
-	for _, nsLabelKey := range nsLabels {
-		labelKey := util.GetNSNameWithPrefix(nsLabelKey)
-		log.Logf("Deleting namespace %s from ipset list %s", nsName, labelKey)
-		if err = ipsMgr.DeleteFromList(labelKey, nsName); err != nil {
-			metrics.SendErrorLogAndMetric(util.NSID, "[DeleteNamespace] Error: failed to delete namespace %s from ipset list %s with err: %v", nsName, labelKey, err)
+	for nsLabelKey, nsLabelVal := range cachedNsObj.LabelsMap {
+		labelIpsetname := util.GetNSNameWithPrefix(nsLabelKey)
```
JungukCho, line 461: nitpick: labelIpsetname -> labelIpsetName?

npm/podController.go
```diff
@@ -489,16 +517,15 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
...
 			return fmt.Errorf("[syncAddAndUpdatePod] Error: failed to delete pod from label ipset with err: %v", err)
 		}
-	}
-
-	// Add the pod to its label's ipset.
-	for _, addIPSetName := range addToIPSets {
-		log.Logf("Adding pod %s to ipset %s", newPodObj.Status.PodIP, addIPSetName)
-		if err = ipsMgr.AddToSet(addIPSetName, newPodObj.Status.PodIP, util.IpsetNetHashFlag, podKey); err != nil {
-			return fmt.Errorf("[syncAddAndUpdatePod] Error: failed to add pod to label ipset with err: %v", err)
+		// {IMPORTANT} The order of compared list will be key and then key+val. NPM should only append after both key
```
JungukCho, line 520: (Same as namespace comment) Can we resolve this issue in this PR? As I understood this issue and we discussed this before, but if possible, it is better to resolve this issue to avoid future possible issue.
Current GetLabelKVFromSet returns key and value. So better to returns (string, string) in the GetLabelKVFromSet
function and use "removedKey" and "removedValue" in this function

npm/podController.go
```diff
@@ -508,14 +535,36 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
...
+	// Updating pod cache with new npmPod information
+	c.npMgr.PodMap[podKey] = newNpmPod(newPodObj)
+
+	// Add the pod to its label's ipset.
+	for _, addIPSetName := range addToIPSets {
+		log.Logf("Adding pod %s to ipset %s", newPodObj.Status.PodIP, addIPSetName)
+		if err = ipsMgr.AddToSet(addIPSetName, newPodObj.Status.PodIP, util.IpsetNetHashFlag, podKey); err != nil {
+			return fmt.Errorf("[syncAddAndUpdatePod] Error: failed to add pod to label ipset with err: %v", err)
+		}
+		// {IMPORTANT} Same as above order is assumed to be key and then key+val. NPM should only append to existing labels
```
JungukCho, line 550: Can we resolve this issue in this PR?

npm/podController.go
```diff
@@ -508,14 +535,36 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
...
+		// only after both ipsets for a given label's key value pair are added successfully
+		// (TODO) will need to remove this ordering dependency
+		addedLabel := util.GetLabelKVFromSet(addIPSetName)
+		if len(addedLabel) > 1 {
+			c.npMgr.PodMap[podKey].appendLabels(map[string]string{addedLabel[0]: addedLabel[1]}, AppendToExitingLabels)
+		}
+	}
+	// This will ensure after all labels are worked on to overwrite. This way will reduce any bugs introduced above
```
JungukCho, line 558: What are possible bugs? It would be nice to update bugs case as comment.

npm/nameSpaceController.go
```diff
@@ -25,6 +25,13 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
+type LabelAppendOperation bool
+
+const (
+	ClearExitingLabels    LabelAppendOperation = true
```
JungukCho, line 31: ClearExitingLables -> ClearExistingLabels?

npm/nameSpaceController.go
```diff
@@ -25,6 +25,13 @@ import (
...
+	AppendToExitingLabels LabelAppendOperation = false
```
JungukCho, line 32: AppendToExitingLabels -> AppendToExistingLabels ?

JungukCho (changes requested):

npm/ipsm/ipsm.go
vakalapa: For debugging these messages can be very useful. It gives a full picture of why the deleteFromList failed or was a no-op.  Because NPM cases are sometimes hard to debug, any additional log message is useful

npm/podController.go
vakalapa: I added a comment. Due to the ordering issues if we not delete and add a label, this appedlabels will help override it.

npm/nameSpaceController.go
vakalapa: As we discussed offline, keeping this as it is for now.

npm/podController.go
```diff
@@ -396,7 +396,7 @@ func (c *podController) syncAddedPod(podObj *corev1.Pod) error {
 	podNs := util.GetNSNameWithPrefix(podObj.Namespace)
 	podKey, _ := cache.MetaNamespaceKeyFunc(podObj)
 	ipsMgr := c.npMgr.NsMap[util.KubeAllNamespacesFlag].IpsMgr
-	klog.Infof("POD CREATING: [%s%s/%s/%s%+v%s]", npmPodObj.PodUID, podNs, npmPodObj.Name, npmPodObj.NodeName, npmPodObj.Labels, npmPodObj.PodIP)
+	klog.Infof("POD CREATING: [%s%s/%s/%s%+v%s]", npmPodObj.PodUID, podNs, npmPodObj.Name, npmPodObj.NodeName, podObj.Labels, npmPodObj.PodIP)
```
JungukCho, line 399: No, message. suggest creating npmPodObj  just before c.npMgr.PodMap[podKey] = npmPodObj if possible

npm/podController.go
```diff
@@ -465,40 +498,36 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
 		// new PodIP needs to be added to all newLabels
 		addToIPSets = util.GetIPSetListFromLabels(newPodObj.Labels)
 
-		log.Logf("Deleting pod %s %s from ipset %s and adding pod %s to ipset %s",
+		klog.Infof("Deleting pod %s %s from ipset %s and adding pod %s to ipset %s",
 			cachedNpmPodObj.PodUID, cachedNpmPodObj.PodIP, cachedNpmPodObj.Namespace, newPodObj.Status.PodIP, newPodObjNs,
 		)
 
 		// Delete the pod from its namespace's ipset.
 		if err = ipsMgr.DeleteFromSet(cachedNpmPodObj.Namespace, cachedNpmPodObj.PodIP, podKey); err != nil {
```
JungukCho, line 506: This looks a bug. We need to append "ns-".
if err = ipsMgr.DeleteFromSet(cachedNpmPodObj.Namespace, cachedNpmPodObj.PodIP, podKey); err != nil 
-> if err = ipsMgr.DeleteFromSet(util.GetNSNameWithPrefix(cachedNpmPodObj.Namespace) , cachedNpmPodObj.PodIP, podKey); err != nil 

We need to improve UT to check real ipset as well to catch this bug.

npm/podController.go
```diff
@@ -472,47 +472,36 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
...
-
 	// Dealing with "updatePod" event - Compare last applied states against current Pod states
 	// There are two possiblities for npmPodObj and newPodObj
 	// #1 case The same object with the same UID and the same key (namespace + name)
 	// #2 case Different objects with different UID, but the same key (namespace + name) due to missing some events for the old object
-	// Start processing three update cases - "ip address", label", and "portlist" change by comparing last applied states against current Pod states
 
-	deleteFromIPSets := []string{}
-	addToIPSets := []string{}
 	// compare cached NPMPod's IP address against newPod's IP address
```
JungukCho, line 480: Better to add comment - "Deal with #2 case." to clarify code

npm/podController.go
```diff
@@ -472,47 +472,36 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
...
+
+		klog.Infof("Adding back Pod with key:%s after IP Mistmatch", podKey)
+		if err = c.syncAddedPod(newPodObj); err != nil {
+			return err
+		}
+
+		return nil
 	}
 
+	// the IP addresses of the cached npmPod and newPodObj is the same
```
JungukCho, line 500: Better to add comment to clarify code - "Deal with #first(1) case"

npm/podController.go
```diff
@@ -472,47 +472,36 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
...
+	// If no change in labels, then GetIPSetListCompareLabels will return empty list.
+	// Otherwise it returns list of deleted PodIP from cached pod's labels and list of added PodIp from new pod's labels
+	addToIPSets, deleteFromIPSets := util.GetIPSetListCompareLabels(cachedNpmPodObj.Labels, newPodObj.Labels)
```
JungukCho, line 503: From here, do we deal with the first case? If so, can we keep using cachedNpmPodObj and update the  cachedNpmPodObj if needed? Are there specific reasons to create newNpmPod() and update cache, etc?

npm/podController.go
```diff
@@ -532,21 +521,14 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
...
 
-	if cachedPobObjIpChanged {
-		// Add the pod to its namespace's ipset.
-		if err = ipsMgr.AddToSet(newPodObjNs, newPodObj.Status.PodIP, util.IpsetNetHashFlag, podKey); err != nil {
-			return fmt.Errorf("[syncAddAndUpdatePod] Error: failed to add pod to namespace ipset with err: %v", err)
-		}
-	}
-
 	// Updating pod cache with new npmPod information
 	newNpmPodObj := newNpmPod(newPodObj)
```
JungukCho, line 533: Need to remove unused members in NpmPod (e.g., rv, NodeName, PodUID, etc)

npm/podController.go
```diff
@@ -472,47 +472,36 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
...
-		if err = ipsMgr.DeleteFromSet(util.GetNSNameWithPrefix(cachedNpmPodObj.Namespace), cachedNpmPodObj.PodIP, podKey); err != nil {
-			return fmt.Errorf("[syncAddAndUpdatePod] Error: failed to delete pod from namespace ipset with err: %v", err)
+	// NPM should clean up existing references of cached pod obj and its IP.
+	// then, re-add new pod obj.
+	if cachedNpmPodObj.PodIP != newPodObj.Status.PodIP {
+		metrics.SendErrorLogAndMetric(util.PodID, "[syncAddAndUpdatePod] Info: Unexpected state. Pod (Namespace:%s, Name:%s, uid:%s, newUid:%s) , has cachedPodIp:%s which is different from PodIp:%s",
+			newPodObjNs, newPodObj.Name, cachedNpmPodObj.PodUID, string(newPodObj.UID), cachedNpmPodObj.PodIP, newPodObj.Status.PodIP)
+
+		klog.Infof("Deleting cached Pod with key:%s first due to IP Mistmatch", podKey)
+		if err = c.cleanUpDeletedPod(cachedNpmPodObj.getPodObjFromNpmPodObj()); err != nil {
```
JungukCho, line 488: Would like to take the same approach from namespace controller (https://github.com/Azure/azure-container-networking/blob/2c74675e182a1608b6ef606a8e171d25d7e91813/npm/nameSpaceController.go#L438) - use "cachedNpmPodKey" as parameter for cleanUpDeletedPod function, not corev1.Pod object. So, we can remove unnecessary creation of PodObj  and getPodObjFromNpmPodObj function.

npm/podController.go
vakalapa: the reason we are creating new obj is because we have fields like Phase, PodIPs, Rv etc which needs to be updated to reflect the latest newPodobj ... we can add them individually too if it makes sense

npm/podController.go
vakalapa: Done makes sense

npm/podController.go
vakalapa: Good improvement, adding this change

npm/podController.go
```diff
@@ -35,62 +33,51 @@ const (
...
-		PodIPs:          podObj.Status.PodIPs,
-		IsHostNetwork:   podObj.Spec.HostNetwork,
-		Labels:          make(map[string]string),
-		ContainerPorts:  []corev1.ContainerPort{},
-		ResourceVersion: rv,
-		Phase:           podObj.Status.Phase,
+		Name:           podObj.ObjectMeta.Name,
+		Namespace:      podObj.ObjectMeta.Namespace,
+		PodIP:          podObj.Status.PodIP,
+		IsHostNetwork:  podObj.Spec.HostNetwork,
```
JungukCho, line 50: Do we use IsHostNetwork?

npm/podController.go
```diff
@@ -35,62 +33,51 @@ const (
...
 		},
 		Status: corev1.PodStatus{
-			Phase:  nPod.Phase,
-			PodIP:  nPod.PodIP,
-			PodIPs: nPod.PodIPs,
+			Phase: nPod.Phase,
+			PodIP: nPod.PodIP,
 		},
 		Spec: corev1.PodSpec{
 			HostNetwork: nPod.IsHostNetwork,
```
JungukCho, line 72: seems we do not use HostNetwork in npmPod, so it is ok to remove this field.

npm/podController.go
```diff
@@ -106,7 +93,23 @@ func (nPod *NpmPod) removeLabelsWithKey(key string) {
 }
 
 func (nPod *NpmPod) appendContainerPorts(podObj *corev1.Pod) {
-	nPod.ContainerPorts = getContainerPortList(podObj)
+	portList := getContainerPortList(podObj)
+	// To remove any references to objects sent to events by sharedinformer,
+	// NPM will need to deepcopy data from the provided objects
+	copiedPortList := make([]corev1.ContainerPort, len(portList))
```
JungukCho, line 99: getContainerPortList returns new slices with copy. Why do we need the same operation again?

npm/podController.go
```diff
@@ -159,7 +172,7 @@ func (c *podController) needSync(eventType string, obj interface{}) (string, boo
 func (c *podController) addPod(obj interface{}) {
 	key, needSync := c.needSync("ADD", obj)
 	if !needSync {
-		log.Logf("[POD ADD EVENT] No need to sync this pod")
+		klog.Infof("[POD ADD EVENT] No need to sync this pod")
 		return
 	}
 	// K8s categorizes Succeeded and Failed pods as a terminated pod and will not restart them
```
JungukCho, line 178: Remove this comment? This comment is not needed.

npm/podController.go
```diff
@@ -336,15 +350,10 @@ func (c *podController) syncPod(key string) error {
...
-
-			// Found the npmPod object from PodMap local cache and start cleaning up processes
-			err = c.cleanUpDeletedPod(cachedNpmPodObj.getPodObjFromNpmPodObj())
+			klog.Infof("pod %s not found, may be it is deleted", key)
+			// cleanUpDeletedPod will check if the pod exists in cache, if it does then proceeds with deletion
+			// if it does not exists, then event will be no-op
+			err = c.cleanUpDeletedPod(key)
 			if err != nil {
 				// need to retry this cleaning-up process
 				metrics.SendErrorLogAndMetric(util.PodID, "Error: %v when pod is not found", err)
```
JungukCho, line 359: Redundant? In processNextWorkItem() func, it sends this information.

npm/podController.go
```diff
@@ -358,7 +367,7 @@ func (c *podController) syncPod(key string) error {
 
 	// If newPodObj status is either corev1.PodSucceeded or corev1.PodFailed or DeletionTimestamp is set, start clean-up the lastly applied states.
 	if isCompletePod(pod) {
-		if err = c.cleanUpDeletedPod(pod); err != nil {
+		if err = c.cleanUpDeletedPod(key); err != nil {
 			metrics.SendErrorLogAndMetric(util.PodID, "Error: %v when pod is in completed state.", err)
```
JungukCho, line 371: Redundant? In processNextWorkItem() func, it sends this information.

npm/podController.go
```diff
@@ -449,55 +468,64 @@ func (c *podController) syncAddAndUpdatePod(newPodObj *corev1.Pod) error {
...
-		addToIPSets = util.GetIPSetListFromLabels(newPodObj.Labels)
-
-		log.Logf("Deleting pod %s %s from ipset %s and adding pod %s to ipset %s",
-			cachedNpmPodObj.PodUID, cachedNpmPodObj.PodIP, cachedNpmPodObj.Namespace, newPodObj.Status.PodIP, newPodObjNs,
-		)
-
-		// Delete the pod from its namespace's ipset.
-		if err = ipsMgr.DeleteFromSet(cachedNpmPodObj.Namespace, cachedNpmPodObj.PodIP, podKey); err != nil {
-			return fmt.Errorf("[syncAddAndUpdatePod] Error: failed to delete pod from namespace ipset with err: %v", err)
+		metrics.SendErrorLogAndMetric(util.PodID, "[syncAddAndUpdatePod] Info: Unexpected state. Pod (Namespace:%s, Name:%s, newUid:%s) , has cachedPodIp:%s which is different from PodIp:%s",
```
JungukCho, line 476: I guess we do not need this message since it can happen.

npm/podController.go
```diff
@@ -132,14 +145,14 @@ func (c *podController) needSync(eventType string, obj interface{}) (string, boo
 		return key, needSync
```
JungukCho, line 145: About metrics.SEndErrorLogAndMetric(). I think we do not need send this information since it can be ignorable. If want to keep it, need to change "util.NpmID" -> util.PodID"

npm/podController.go
vakalapa: this line was very useful in Genesys debugging, we woudl want to see why the delete and add happened for any new folks.

npm/podController.go
vakalapa: good catch changed this in 3 locations.

JungukCho (approved):
Thank you for applying comments! Looks good to me.

PR #849, opened 2021-04-06
[NPM] support network policy controller and its unit tests

JungukCho:
**Reason for Change**:

This PR is mainly to add network policy controller to make dealing with network policy event reliable.
Add unit tests for network policy controller to test various network policy events.
Cleans up previous codes (e.g., not used logics).
Put comments of existing and new codes as many as possible.

**Issue Fixed**:

1. Retry reconcile loop if there is failed
2. Fix incorrect return error.
 

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


npm/networkPolicyController.go
```diff
@@ -0,0 +1,491 @@
...
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+
+	netPolCachedKey := util.GetNSNameWithPrefix(key)
+
+	// (TODO): Reduce scope of lock later
+	c.npMgr.Lock()
+	defer c.npMgr.Unlock()
```
vakalapa, line 143: We can optimize this lock from defer to explicitly mentioning below RawNpMap exists check. thta way we are not locking when Add is being performed. Probably we can expand this to other controllers later.

npm/networkPolicyController.go
```diff
@@ -0,0 +1,491 @@
...
+	// Get the network policy resource with this namespace/name
+	netPolObj, err := c.netPolLister.NetworkPolicies(namespace).Get(name)
+
+	// (TODO): Reduce scope of lock later
+	c.npMgr.Lock()
+	defer c.npMgr.Unlock()
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			log.Logf("Network Policy %s is not found, may be it is deleted", key)
```
vakalapa, line 239: Even in these log messages we should have the [<funcName>] that way it will be easier while reading logs

npm/networkPolicyController.go
```diff
@@ -0,0 +1,491 @@
...
+	if err = c.removeCidrsRule("out", netPolObj.ObjectMeta.Name, netPolObj.ObjectMeta.Namespace, egressIPCidrs, ipsMgr); err != nil {
+		return fmt.Errorf("[deleteNetworkPolicy] Error: removeCidrsRule out due to %v", err)
+	}
+
+	// (TODO): Can we decouple below logic from cachedNetPolObj since if all deletions for cachedNetPolObj are successful, but UnititNpmChains function is failed,
+	// deleteNetworkPolicy function will be executed again.
+
+	// if there is only one cached network policy in RawNPMap and no immediate network policy to process,
+	// clean up default azure npm chains since the cached network policy from RawNPMap is removed.
+	if isSafeCleanUpAzureNpmChain && len(c.npMgr.RawNpMap) == 1 {
```
vakalapa, line 416: Instead of this one check, why not move this cleanup operation below delete(c.npMgr.RawNpMap, cachedNetPolKey)

removal of the cache element is critical than the ChainCleanup. Any error in chaincleanup should not block the clean up of cache. Because that will result in more errors.

npm/networkPolicyController_test.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func (f *netPolFixture) ipSetSave(ipsetConfigFile string) {
+	//  call /sbin/ipset save -file /var/log/ipset-test.conf
+	f.t.Logf("Start storing ipset to %s", ipsetConfigFile)
+	if err := f.ipsMgr.Save(ipsetConfigFile); err != nil {
+		f.t.Errorf("ipSetSave failed @ ipsMgr.Save")
+	}
+}
+func (f *netPolFixture) ipSetRestore(ipsetConfigFile string) {
```
vakalapa, line 73: We can reuse ipsetsave and restore from other tests right ?

npm/networkPolicyController_test.go
```diff
@@ -0,0 +1,281 @@
...
+	f.netPolController.netPolListerSynced = alwaysReady
+
+	for _, netPol := range f.netPolLister {
+		f.kubeInformer.Networking().V1().NetworkPolicies().Informer().GetIndexer().Add(netPol)
+	}
+
+	f.kubeInformer.Start(stopCh)
+}
+
+func (f *netPolFixture) ipSetSave(ipsetConfigFile string) {
```
vakalapa, line 66: with this, we will need IpTableSave and IpTableRestore, since each test will be manipulating iptable rules, we should clean them up once done

npm/networkPolicyController.go
```diff
@@ -0,0 +1,491 @@
...
+			return fmt.Errorf("[removeCidrsRule] deleting ipset %s with err: %v", ipCidrSet, err)
+		}
+	}
+
+	return nil
+}
+
+// GetProcessedNPKey will return netpolKey
+// (TODO): will use this function when optimizing management of multiple network policies with merging and deducting multiple network policies.
+func (c *networkPolicyController) getProcessedNPKey(netPolObj *networkingv1.NetworkPolicy) string {
```
vakalapa, line 481: Can you add a testcase for this function.

npm/networkPolicyController.go
```diff
@@ -0,0 +1,491 @@
...
+	iptMgr := c.npMgr.NsMap[util.KubeAllNamespacesFlag].iptMgr
+	sets, namedPorts, lists, ingressIPCidrs, egressIPCidrs, iptEntries := translatePolicy(netPolObj)
+	for _, set := range sets {
+		log.Logf("Creating set: %v, hashedSet: %v", set, util.GetHashedName(set))
+		if err = ipsMgr.CreateSet(set, append([]string{util.IpsetNetHashFlag})); err != nil {
+			return fmt.Errorf("[syncAddAndUpdateNetPol] Error: creating ipset %s with err: %v", set, err)
+		}
+	}
+	for _, set := range namedPorts {
+		log.Logf("Creating set: %v, hashedSet: %v", set, util.GetHashedName(set))
```
vakalapa, line 340: Also one more test case with NamedPort network policy, there should be an example in npm/testpolicies/ folder.

npm/networkPolicyController.go
```diff
@@ -0,0 +1,491 @@
...
+	// flag to indicate default Azure NPM chain is created or not
+	isAzureNpmChainCreated bool
+}
+
+func NewNetworkPolicyController(npInformer networkinginformers.NetworkPolicyInformer, clientset kubernetes.Interface, npMgr *NetworkPolicyManager) *networkPolicyController {
+	netPolController := &networkPolicyController{
+		clientset:              clientset,
+		netPolLister:           npInformer.Lister(),
+		netPolListerSynced:     npInformer.Informer().HasSynced,
+		workqueue:              workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "NetPols"),
```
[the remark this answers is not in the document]
JungukCho: ack

npm/networkPolicyController.go
```diff
@@ -0,0 +1,491 @@
...
+		npMgr:                  npMgr,
+		isAzureNpmChainCreated: false,
+	}
+
+	npInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc:    netPolController.addNetPol,
```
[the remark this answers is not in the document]
JungukCho: ack. For consistency, I also changed deleteNetPol to deleteNetworkPolicy. deleteNetworkPolicy -> cleanUpNetworkPolicy

npm/networkPolicyController.go
```diff
@@ -0,0 +1,491 @@
...
+	// needSync already checked validation of casting new network policy.
+	newNetPol, _ := new.(*networkingv1.NetworkPolicy)
+	oldNetPol, ok := old.(*networkingv1.NetworkPolicy)
+	if ok {
+		if oldNetPol.ResourceVersion == newNetPol.ResourceVersion {
+			// Periodic resync will send update events for all known network plicies.
+			// Two different versions of the same network policy will always have different RVs.
+			return
+		}
+	}
```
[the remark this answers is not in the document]
JungukCho: Make sense. Will update IsSamePolicy since it does not consider name. Let me check it.

npm/networkPolicyController.go
JungukCho: Good point. Another important thing is we need to decouple all shared maps (e.g., RawNpMap, etc) from npm.go to each controller and then carefully walk through them again. Even we may encapsulate each map with lock as a struct. Let's understand each maps in details and start this process as another PR.

npm/networkPolicyController.go
JungukCho: Good point. By the way why don't we use klog? It will save our time and more consistent since file names and line comes as free. Are there special reasons?

npm/networkPolicyController.go
JungukCho: I agree we need to decouple chaincleanup with deletion event, but not sure clean way. It does not cause any more errors. It needs extra computation when retrying happens. If it is ok to leave default Azure NPM chain even though there is no active network policy, we can reverse order. With changing order, we do not have chance to clean up default Azure NPM chain if UninitNpmchains function.

npm/networkPolicyController.go
JungukCho: It would be better when we start using this function with network policy management (e.g., addpolicy and deductpolicy). I will comment it out to remove coverage.

npm/networkPolicyController.go
JungukCho: Do you mean we need tp add another yaml file in npm/testpolicies folder?

npm/networkPolicyController.go
JungukCho: One more what overhead is if we leave default Azure chain in data plane side?

npm/networkPolicyController_test.go
JungukCho: Good point. Added two methods for iptables management. Now the methods are not used in networkPolicyController_test.go due to npm_test.go. We will use them later.

npm/networkPolicyController_test.go
JungukCho: Definitely. we can. I would suggest we left them separately now and see them after reviewing and thinking overall structure of our codes since as we discussed, we may decouple each controller into each package later.

npm/networkPolicyController.go
vakalapa: this log lib is generic one we use for other services in this repo. If it makes sense we can move, or add such support to acn/log so our other services can reap the benefits

npm/networkPolicyController.go
vakalapa: either use the existing yaml to have a named port UT or for consistency purposes you can create a namedPort policy with createPolicy

JungukCho: Good points. applied to test different type of ports.

npm/networkPolicyController.go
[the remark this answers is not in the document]
JungukCho: Applied and also add UT to test this.

npm/networkPolicyController.go
JungukCho: Concensus opinion is ok to use klog. Replaced log with klog in networkPolicyController, which adds extra information to help reading logs.

vakalapa (approved):
lgtm :shipit:   🚀

npm/networkPolicyController.go
```diff
@@ -0,0 +1,498 @@
...
+	defer c.npMgr.Unlock()
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.Infof("Network Policy %s is not found, may be it is deleted", key)
+			// netPolObj is not found, but should need to check the RawNpMap cache with cachedNetPolKey
+			// #1 No item in RawNpMap, which means network policy with the cachedNetPolKey is not applied. Thus, do not need to clean up process.
+			// #2 item in RawNpMap exists, which means network policy with the cachedNetPolKey is applied . Thus, Need to clean up process.
+			cachedNetPolKey := util.GetNSNameWithPrefix(key)
+			_, cachedNetPolObjExists := c.npMgr.RawNpMap[cachedNetPolKey]
```
vakalapa, line 254: Do we have to check if the cahedObj exists here ? Because the first check we do in cleanUpnetworkPolicy is if the cachedObj exists, i think we can safely remove this check. Any reason for leaving this check here ?

npm/networkPolicyController.go
```diff
@@ -0,0 +1,498 @@
...
+			utilruntime.HandleError(fmt.Errorf("expected string in workqueue but got %#v", obj))
+			return nil
+		}
+		// Run the syncNetPol, passing it the namespace/name string of the
+		// network policy resource to be synced.
+		// TODO : may consider using "c.queue.AddAfter(key, *requeueAfter)" according to error type later
+		if err := c.syncNetPol(key); err != nil {
+			// Put the item back on the workqueue to handle any transient errors.
+			c.workqueue.AddRateLimited(key)
+			return fmt.Errorf("error syncing '%s': %s, requeuing", key, err.Error())
```
vakalapa, line 213: [Important] We are missing a metrics.SendErrorLogAndMetric(util.NSID log event here! if this is not present, then our kusto will not have these failed netpol events

vakalapa (changes requested):

npm/networkPolicyController.go
JungukCho: Good catch. Removed the redundant check.

npm/networkPolicyController.go
JungukCho: Have the logic to send error message in https://github.com/Azure/azure-container-networking/blob/761a9740a51dc0ccb80fd20c9d98487604c25c3d/npm/networkPolicyController.go#L224

npm/networkPolicyController.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+// syncAddAndUpdateNetPol handles a new network policy or an updated network policy object triggered by add and update events
+func (c *networkPolicyController) syncAddAndUpdateNetPol(netPolObj *networkingv1.NetworkPolicy) error {
+	// This timer measures execution time to run this function regardless of success or failure cases
+	timer := metrics.StartNewTimer()
+	defer timer.StopAndRecord(metrics.AddPolicyExecTime)
+
+	var err error
+	netpolKey, err := cache.MetaNamespaceKeyFunc(netPolObj)
```
vakalapa, line 306: NetPoljKey right now always needs the ns- prefix right ? Why dont we create a function which gets metacachekey and adds the prefix, that way we will reduce the complexity and any changes needed will be in a single function

vakalapa: if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
		utilruntime.HandleError(err)
		return
	}

	netPolCachedKey := util.GetNSNameWithPrefix(key)
this can be a single func

npm/networkPolicyController.go
```diff
@@ -0,0 +1,492 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.Infof("Network Policy %s is not found, may be it is deleted", key)
+			// netPolObj is not found, but should need to check the RawNpMap cache with cachedNetPolKey.
+			// cleanUpNetworkPolicy method will take care of the deletion of a cached network policy if the cached network policy exists with cachedNetPolKey in our RawNpMap cache.
+			cachedNetPolKey := util.GetNSNameWithPrefix(key)
+			err = c.cleanUpNetworkPolicy(cachedNetPolKey, SafeToCleanUpAzureNpmChain)
+			if err != nil {
+				return fmt.Errorf("[syncNetPol] Error: %v when network policy is not found\n", err)
+			}
```
vakalapa, line 256: missing another return err ? i do not think we will need it. But checking

npm/networkPolicyController.go
```diff
@@ -0,0 +1,492 @@
...
+	// Install this default rules for kube-system and azure-npm chains if they are not initilized.
+	// Execute initializeDefaultAzureNpmChain function first before actually starting processing network policy object.
+	if err = c.initializeDefaultAzureNpmChain(); err != nil {
+		return fmt.Errorf("[syncNetPol] Error: due to %v", err)
+	}
+
+	// Cache network object first before applying ipsets and iptables.
+	// If error happens while applying ipsets and iptables,
+	// the key is re-queued in workqueue and process this function again, which eventually meets desired states of network policy
+	c.npMgr.RawNpMap[cachedNetPolKey] = netPolObj
```
vakalapa, line 340: We will need a deepcopy of this here ! we are not removing references from the sharedinformer cache

npm/networkPolicyController.go
```diff
@@ -0,0 +1,492 @@
...
+
+	netPolCachedKey := util.GetNSNameWithPrefix(key)
+	c.npMgr.Lock()
+	cachedNetPolObj, netPolExists := c.npMgr.RawNpMap[netPolCachedKey]
+	c.npMgr.Unlock()
+	if netPolExists {
+		// if network policy does not have different states against lastly applied states stored in cachedNetPolObj,
+		// netPolController does not need to reconcile this update.
+		// in this updateNetworkPolicy event, newNetPol was updated with states which netPolController does not need to reconcile.
+		if isSameNetworkPolicy(cachedNetPolObj, newNetPol) {
```
neaggarwMS, line 116: This can also be called as needSync, but I'm fine with isSameNetworkPolicy

npm/networkPolicyController.go
```diff
@@ -0,0 +1,492 @@
...
+	var err error
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+
+	netPolCachedKey := util.GetNSNameWithPrefix(key)
+
+	// (TODO): need to decouple this lock from npMgr if possible
+	c.npMgr.Lock()
```
neaggarwMS, line 155: Shouldnt we take a ReadLock? Same for above

npm/networkPolicyController.go
```diff
@@ -0,0 +1,492 @@
...
+			c.workqueue.Forget(obj)
+			utilruntime.HandleError(fmt.Errorf("expected string in workqueue but got %#v", obj))
+			return nil
+		}
+		// Run the syncNetPol, passing it the namespace/name string of the
+		// network policy resource to be synced.
+		// TODO : may consider using "c.queue.AddAfter(key, *requeueAfter)" according to error type later
+		if err := c.syncNetPol(key); err != nil {
+			// Put the item back on the workqueue to handle any transient errors.
+			c.workqueue.AddRateLimited(key)
```
neaggarwMS, line 212: >c.workqueue.AddRateLimited(key) [](start = 3, length = 31)

Why do we need to do that? Simply dont call c.WorkQueue.Done in linie#196  (You should call that only if err!=nil)

npm/networkPolicyController.go
```diff
@@ -0,0 +1,492 @@
...
+			return fmt.Errorf("error syncing '%s': %s, requeuing", key, err.Error())
+		}
+		// Finally, if no error occurs we Forget this item so it does not
+		// get queued again until another change happens.
+		c.workqueue.Forget(obj)
```
neaggarwMS, line 217: How it is different from c.Workqueue.Done?

npm/npm.go
```diff
@@ -271,53 +269,8 @@ func NewNetworkPolicyManager(clientset *kubernetes.Clientset, informerFactory in
...
-					}
-				}
-				npMgr.Lock()
-				npMgr.DeleteNetworkPolicy(networkPolicyObj)
-				npMgr.Unlock()
-			},
-		},
-	)
+	// create network policy controller
+	npMgr.netPolController = NewNetworkPolicyController(informerFactory.Networking().V1().NetworkPolicies(), clientset, npMgr)
```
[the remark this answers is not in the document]
JungukCho: I would like to remove npInformer from NetworkPolicyManager struct in npm.go. Let me first check whether I can remove npInformer member. If not, I can use npInfomer.

npm/networkPolicyController_test.go
```diff
@@ -0,0 +1,337 @@
...
+	defer close(stopCh)
+	f.newNetPolController(stopCh)
+
+	addNetPol(t, f, netPolObj1)
+	addNetPol(t, f, netPolObj2)
+
+	testCases := []expectedNetPolValues{
+		{1, 2, 0, true, true},
+	}
+	checkNetPolTestResult("TestAddMultipleNetPols", f, testCases)
```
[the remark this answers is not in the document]
JungukCho: Good catch. Will add them.

npm/networkPolicyController.go
```diff
@@ -0,0 +1,492 @@
...
+
+func (c *networkPolicyController) processNextWorkItem() bool {
+	obj, shutdown := c.workqueue.Get()
+
+	if shutdown {
+		return false
+	}
+
+	err := func(obj interface{}) error {
+		defer c.workqueue.Done(obj)
```
neaggarwMS, line 196: >defer c.workqueue.Done(obj) [](start = 2, length = 27)

                   c.workqueue.Done(obj)
                   if err != nil {
                      c.workqueue.AddRateLimited(key)
                  }

npm/networkPolicyController.go
JungukCho: "nil" is return in next line, but good to add "return err" to improve readability.

npm/networkPolicyController.go
JungukCho: I think current code is better in terms of memory and computations. If we use deepcopy network policyhere, we will always have two network policy objects and increase computations for deepcopy operation. But, if we use reference of network policy from sharedinformer cache, it will be either one or two objects eventually and no computation overhead for deepcopy.

npm/networkPolicyController.go
JungukCho: Good catch. We can use readlock here, but current NPM code base is using global Mutex. When we decouple global Mutex from our code, we can take your suggestion. It will be different PR.

npm/networkPolicyController.go
JungukCho: We want to put ratelimit when reconcile is failed not to try it again so fast.	https://pkg.go.dev/k8s.io/client-go/util/workqueue#RateLimitingInterface

npm/networkPolicyController.go
JungukCho: // We call Done here so the workqueue knows we have finished processing this item. We also must remember to call Forget if we do not want this work item being re-queued. For example, we do not call Forget if a transient error occurs, instead the item is put back on the workqueue and attempted again after a back-off period. - Refer to 
https://github.com/kubernetes/sample-controller/blob/master/controller.go#L196

npm/networkPolicyController.go
JungukCho: I think we agree to stick to current code.

npm/networkPolicyController.go
JungukCho: Network policy controller does not need "ns-" prefix. So, we agree to not to use "ns-" prefix as key for network policy cache.

npm/networkPolicyController.go
```diff
@@ -0,0 +1,492 @@
...
+			AddFunc:    netPolController.addNetworkPolicy,
+			UpdateFunc: netPolController.updateNetworkPolicy,
+			DeleteFunc: netPolController.deleteNetworkPolicy,
+		},
+	)
+	return netPolController
+}
+
+// needSync checks whether the obj is valid network policy object.
+func (c *networkPolicyController) needSync(obj interface{}) (string, error) {
```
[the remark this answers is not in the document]
JungukCho: Good suggestion. applied.

npm/networkPolicyController.go
JungukCho: Now I would like to stick to isSameNetworkPolicy name since it is related to isSamePolicy function in parsePolicy.go and  ProcessedNpMap data structure. So, when revisiting this issue, I will consider suggestion.

npm/npm.go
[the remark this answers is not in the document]
JungukCho: npInformer member in NetworkPolicyManager struct  is also used in Start method while it may be not necessary to keep the member. I changed it to "npInformer".

npm/networkPolicyController.go
vakalapa: Makes sense, we can remove the prefix in chachedNetpolKey

vakalapa (approved):

PR #852, opened 2021-04-20
[NPM] Add ipset package from k/k package

matmerr:
**Reason for Change**:

Use the well used utils/ipset package from k/k. Bringing it in this way since vendoring k/k is not recommended. Added nethash and set list

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


JungukCho:
Suggest further exploring how utils/ipset package from k/k is align with our NPM code base (e.g., in each controller unit tests) to understand what changes are required in our NPM code base in at least high level. Then, we will make the decision whether we take this direction or build our own extension for mocking ipset/iptables.

PR #856, opened 2021-04-22
[NPM] fix management of DeletedFinalStateUnknown object on deletion event in each controller

JungukCho:
**Reason for Change**:

Current code does not enqueue a key of DeletedFinalStateUnknown object on deletion event, which may cause leak of applied states (e.g., ipset, iptables) and keep an item in a local cache (e.g., podMap, nsMap, nsMap).
This PR is to properly clean up applied states (e.g., ipset, iptables, and local cache (e.g., podMap, nsMap, nsMap) in case of DeletedFinalStateUnknown object on deletion event in each controller.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/npm/nameSpaceController_test.go
+++ b/npm/nameSpaceController_test.go
@@ -16,6 +16,7 @@ import (
 	kubeinformers "k8s.io/client-go/informers"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
 	core "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/cache"
 )
 
 var (
@@ -124,7 +125,7 @@ func updateNamespace(t *testing.T, f *nameSpaceFixture, oldNsObj, newNsObj *core
 	f.nsController.processNextWorkItem()
 }
 
-func deleteNamespace(t *testing.T, f *nameSpaceFixture, nsObj *corev1.Namespace) {
+func deleteNamespace(t *testing.T, f *nameSpaceFixture, nsObj *corev1.Namespace, isDeletedFinalStateUnknownObject IsDeletedFinalStateUnknownObject) {
 	addNamespace(t, f, nsObj)
 	t.Logf("Complete add namespace event")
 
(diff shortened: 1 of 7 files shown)


npm/networkPolicyController_test.go
```diff
@@ -305,13 +315,54 @@ func TestDeleteNetworkPolicy(t *testing.T) {
...
+	}
+
+	f.netPolController.deleteNetworkPolicy(tombstone)
+	testCases := []expectedNetPolValues{
+		{1, 0, 0, false, false, 0, nil, 0, nil},
+	}
+	checkNetPolTestResult("TestDeleteNetworkPolicyWithTombstone", f, testCases)
+}
+
+func TestDeleteNetworkPolicyWithTombstoneAfterAddingNetworkPolicy(t *testing.T) {
```
vakalapa, line 349: <Nit> I feel like there is no major difference in functionality between TestDeleteNetworkPolicyWithTombstone and TestDeleteNetworkPolicyWithTombstoneAfterAddingNetworkPolicy as AddingNS does not care about tombstone and in deleting the way we contruct tombstone is same. So we can reduce the complexity these test files by having only one tescase and not send delefinalstate in deleteNetPol(t *testing.T, ...) func bu

JungukCho7: They are different.  First test case only check "delete" event function in each controller.
Second test case is to try to mimic the realistic situation which walks through "delete" event -> workqueue -> reconcile. Thus, second test case first adds resources and then we emulate the realistic story (i.e., missing event and relist happens). Then, it proceeds all required steps like cleaning up the applied states from "delete" event.
If we want to keep one test case, I would like to keep second one which has more comprehensive test.

I am not sure this increase complexity. It has one extra boolean variable for only delete event in UT and later we can do more sophisticated tests. But, if there is better way to deal with this, let me know.

JungukCho: They are different. 
First test case only check "delete" event function in each controller. 
Second test case is to try to mimic the realistic situation which walks through "delete" event -> workqueue -> reconcile. Thus, second test case first adds resources and then we emulate the realistic story (i.e., missing event and relist happens). Then, it proceeds all required steps like cleaning up the applied states from "delete" event. 
If we want to keep one test case, I would like to keep second one which has more comprehensive test.

I am not sure this increase complexity. It has one extra boolean variable for only delete event in UT and later we can do more sophisticated tests. But, if there is better way to deal with this, let me know.

vakalapa (approved):
lgtm

matmerr (approved):

npm/nameSpaceController_test.go
```diff
@@ -124,15 +125,24 @@ func updateNamespace(t *testing.T, f *nameSpaceFixture, oldNsObj, newNsObj *core
 	f.nsController.processNextWorkItem()
 }
 
-func deleteNamespace(t *testing.T, f *nameSpaceFixture, nsObj *corev1.Namespace) {
+func deleteNamespace(t *testing.T, f *nameSpaceFixture, nsObj *corev1.Namespace, isDeletedFinalStateUnknownObject IsDeletedFinalStateUnknownObject) {
```
matmerr, line 128: Potential function name ambiguity, can clarify between this helper deleteNamespace and f.nsController.deleteNamespace()

JungukCho: Good point. I think last time Vamsi also mentioned this. We can discuss and refactor this point when we add more UTs with ipset and iptables.

PR #861, opened 2021-04-28
[NPM] Use utilexec for IPSet calls and fakeexec in podcontroller tests

matmerr:
**Reason for Change**:

Our current test framework for NPM is heavily dependent on making actual syscalls to ipset/iptables. By interfacing out the calls to ipset, we can have actual UT coverage, without having to run tests as root, but we can also verify call ordering and have better visibility to the syscalls we're making.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


npm/npm_test.go
```diff
@@ -28,31 +30,29 @@ func getKey(obj interface{}, t *testing.T) string {
...
+	allNs, _ := newNs(util.KubeAllNamespacesFlag, npMgr.Exec)
 	npMgr.NsMap[util.KubeAllNamespacesFlag] = allNs
 	return npMgr
 }
 
 func TestMain(m *testing.M) {
 	metrics.InitializeAll()
 	iptMgr := iptm.NewIptablesManager()
-	iptMgr.Save(util.IptablesConfigFile)
+	iptMgr.UninitNpmChains()
```
matmerr, line 50: if a test fails, subsequent tests acn fail in the restore after test execution, in my opinion uniniting iptm and ipsm is better for the interim until everything is fakeexec'd out

npm/networkPolicyController_test.go
```diff
@@ -271,7 +272,8 @@ func TestAddMultipleNetworkPolicies(t *testing.T) {
 	// namedPort
 	netPolObj2.Spec.Ingress[0].Ports[0].Port = &intstr.IntOrString{StrVal: fmt.Sprintf("%s", netPolObj2.Name)}
 
-	f := newNetPolFixture(t)
+	fexec := exec.New()
```
vakalapa, line 275: why cant we move this into newNetPolFixture itself, reducing the diff

matmerr: This is a temporary diff until this test suite is updated with fake exec. Until the suites are migrated, some tests are run with a real exec and some with a fake, so imo it's more pluggable to have the method of exec at test level

npm/ipsm/ipsm.go
```diff
@@ -496,9 +503,10 @@ func (ipsMgr *IpsetManager) Save(configFile string) error {
...
 	}
 
-	cmd := exec.Command(util.Ipset, util.IpsetSaveFlag, util.IpsetFileFlag, configFile)
-	if err := cmd.Start(); err != nil {
-		metrics.SendErrorLogAndMetric(util.IpsmID, "Error: failed to save ipset to file.")
+	cmd := ipsMgr.Exec.Command(util.Ipset, util.IpsetSaveFlag, util.IpsetFileFlag, configFile)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		metrics.SendErrorLogAndMetric(util.IpsmID, "Error: failed to save ipset to file with err %v, stderr: %v", err, string(output))
 		return err
```
JungukCho, line 510: Is it better to return error with the same format you did in "Run" function to provide more info (e.g., stderr)? 
Same for Restore and DestroyNpmIpsets functions.

npm/ipsm/ipsm.go
```diff
@@ -25,6 +26,7 @@ type ipsEntry struct {
 
 // IpsetManager stores ipset states.
 type IpsetManager struct {
+	Exec    utilexec.Interface
```
JungukCho, line 29: Can we define "exec utilexec.Interface" if Exec is only used in this package?

npm/ipsm/ipsm_test.go
```diff
@@ -90,16 +96,23 @@ func TestAddToList(t *testing.T) {
 }
 
 func TestDeleteFromList(t *testing.T) {
-	ipsMgr := NewIpsetManager()
-	if err := ipsMgr.Save(util.IpsetTestConfigFile); err != nil {
-		t.Errorf("TestDeleteFromList failed @ ipsMgr.Save")
+	var calls = []testutils.TestCmd{
```
JungukCho, line 99: better to use string variables to define "test-set" and "test-list" just in case we want to use other values later?

npm/ipsm/ipsm_test.go
```diff
@@ -90,16 +96,23 @@ func TestAddToList(t *testing.T) {
...
 	}
 
-	defer func() {
-		if err := ipsMgr.Restore(util.IpsetTestConfigFile); err != nil {
-			t.Errorf("TestDeleteFromList failed @ ipsMgr.Restore")
-		}
-	}()
+	fexec, fcmd := testutils.GetFakeExecWithScripts(calls)
+
+	ipsMgr := NewIpsetManager(fexec)
```
JungukCho, line 115: Based on "exec.New()" and "fexec", ipsMgr decides which one is used. Some of the unit tests are using "exec.New()" and some of them use "fexec". What are the rules for them in unit tests?

npm/ipsm/ipsm_test.go
```diff
@@ -818,12 +855,15 @@ func TestIPSetCannotBeAddedAsElementDoesNotExist(t *testing.T) {
 */
 func TestMain(m *testing.M) {
 	metrics.InitializeAll()
-	ipsMgr := NewIpsetManager()
-	ipsMgr.Save(util.IpsetConfigFile)
```
JungukCho, line 822: Do we need "Save" and "Restore" anymore to store existing ipsets and restore them after completion of UTs when UTs do not use "fexec"?

npm/nameSpaceController.go
```diff
@@ -41,12 +42,12 @@ type Namespace struct {
...
 
 // newNS constructs a new namespace object.
-func newNs(name string) (*Namespace, error) {
+func newNs(name string, exec utilexec.Interface) (*Namespace, error) {
 	ns := &Namespace{
 		name:      name,
 		LabelsMap: make(map[string]string),
 		SetMap:    make(map[string]string),
-		IpsMgr:    ipsm.NewIpsetManager(),
+		IpsMgr:    ipsm.NewIpsetManager(exec),
```
JungukCho, line 50: As far as I understood our code, there is no reason to tie IpsMgr and IptMgr in Namespace struct. In my opinion, it is better to locate them to other places (e.g., IpsMgr in npMgr, IptMgr in networkpolicyController). So, it will be easier to use our main code and UTs. I felt like there are multiple places to pass exec, etc now. However, I think that is beyond of this PR. We will fix it in other PRs. Any thoughts?

npm/npm_test.go
JungukCho: Can you elaborate this since I guess you will work on pluggable approach to run real-exec and fake-exec? Here, it sounds like that you use fakeexe in all UTs.

npm/podController_test.go
```diff
@@ -211,7 +214,25 @@ func TestAddMultiplePods(t *testing.T) {
 	podObj1 := createPod("test-pod-1", "test-namespace", "0", "1.2.3.4", labels, NonHostNetwork, corev1.PodRunning)
 	podObj2 := createPod("test-pod-2", "test-namespace", "0", "1.2.3.5", labels, NonHostNetwork, corev1.PodRunning)
 
-	f := newFixture(t)
+	var calls = []testutils.TestCmd{
```
JungukCho, line 217: nitpick: better to use string variables for them?

JungukCho: Wondering what the best practice is in UTs? It seems many elements in TestCmd are duplicated in multiple places. Can we remove the redundancy?

npm/ipsm/ipsm_test.go
matmerr: In this PR, only podcontroller_test and some tests in ipsm_test are using fexec. In subsequent pr's, tests will be migrated to use fexec. The use of exec.New() is to not break existing tests that use ipset

npm/ipsm/ipsm_test.go
matmerr: when using fexec we don't need to save/restore because no state is created

npm/ipsm/ipsm_test.go
```diff
@@ -818,12 +855,15 @@ func TestIPSetCannotBeAddedAsElementDoesNotExist(t *testing.T) {
...
 
-	exitCode := m.Run()
+	log.Printf("Uniniting iptables")
+	iptm := iptm.NewIptablesManager()
+	iptm.UninitNpmChains()
```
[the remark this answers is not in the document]
matmerr: Yes, if a previous test panics then iptable state will be left behind and break next tests. The goal is to remove the iptables/ipset dependency from UT's

npm/npm_test.go
matmerr: eventually fakeexec will be used in all UT's but ns and netpol controller are still using a realexec

npm/ipsm/ipsm_test.go
matmerr: discussed offline, will optimize in subsequent pr's

npm/podController_test.go
matmerr: discussed offline, will add helper methods in subsequent pr

JungukCho:
lgtm and look forward to subsequent PR.

JungukCho (dismissed):
lgtm and look forward to subsequent PR.

JungukCho (approved):
lgtm and look forward to subsequent PR again.

PR #863, opened 2021-04-30
[NPM] [BUG] Supporting multiple values under label selector MatchExpr

vakalapa:
This PR is fix the bug of supporting multiple values in a LabelSelector match expressions. 
```yaml
 - key: pod
    operator: In
    values:
      - b
      - c
```
Today NPM considers the values of the above selector to be b AND c, but it should be b OR c

NPM now follows different design for podSelector and nameSpaceSelector. 

For PodSelector, NPM has ipsets with immutable members like ips or ip,proto:port. Ipset allows creating 2nd level ipset ype list:set to be created with pod label ipsets as members and NPM will  use this 2nd level ipset as the matchset criteria in rules.

For NameSpaceSelector, NPM uses ipsets with mutable members like other ipsets. Ipset does not allow adding list:set as members to list:set ipset, so 2nd level ipset cannot be used in case of nameSelectors. For this reason NPM will create combinations of nameSpaceSelectors and add individual rule for each combination.


npm/ipsm/ipsm.go
```diff
@@ -70,6 +87,21 @@ func (ipsMgr *IpsetManager) Exists(listName string, setName string, kind string)
 	return true
 }
 
+// IpSetReferIncOrDec checks if an element exists in setMap/listMap and then increases or decreases tis refer count.
```
JungukCho, line 90: tis refer count -> referCount in the Ipset

npm/ipsm/ipsm.go
```diff
@@ -36,11 +44,20 @@ type Ipset struct {
 	referCount int
 }
 
+func (ipset *Ipset) IncReferCount() {
```
JungukCho, line 47: I think it is better not to expose this function to other packages. IncReferCount() -> incReferCount

npm/ipsm/ipsm.go
```diff
@@ -36,11 +44,20 @@ type Ipset struct {
...
+	ipset.referCount++
+}
+
+func (ipset *Ipset) DecReferCount() {
```
JungukCho, line 51: I think it is better not to expose this function to other packages. DecReferCount() -> decReferCount

npm/ipsm/ipsm.go
```diff
@@ -70,6 +87,21 @@ func (ipsMgr *IpsetManager) Exists(listName string, setName string, kind string)
 	return true
 }
 
+// IpSetReferIncOrDec checks if an element exists in setMap/listMap and then increases or decreases tis refer count.
+func (ipsMgr *IpsetManager) IpSetReferIncOrDec(ipsetName string, kind string, countOperation ReferCountOperation) {
```
JungukCho, line 91: It is better not to expose this function to other packages. IpSetReferIncOrDec() -> ipSetReferIncOrDec.

May Improve readability by making separate functions for increasing and decreasing reference count instead of using flag if possible, but, do not have strong opinion.

npm/networkPolicyController.go
```diff
@@ -351,10 +351,24 @@ func (c *networkPolicyController) syncAddAndUpdateNetPol(netPolObj *networkingv1
...
+	}
+	// Then NPM will add members to the above list, this is to avoid members being added
+	// to lists before they are created.
+	for listKey, listLabelsMembers := range lists {
+		for _, listMember := range listLabelsMembers {
+			if err = ipsMgr.AddToList(listKey, listMember); err != nil {
+				return fmt.Errorf("[syncAddAndUpdateNetPol] Error: Adding ipset member %s to ipset list %s with err: %v", listMember, listKey, err)
+			}
 		}
+		ipsMgr.IpSetReferIncOrDec(listKey, util.IpsetSetListFlag, ipsm.IncrementOp)
```
JungukCho, line 371: better to include this call in AddToList(). So, we do not need to expose those functions to other packages and encapsulate information in ipsMgr.

npm/networkPolicyController.go
```diff
@@ -351,10 +351,24 @@ func (c *networkPolicyController) syncAddAndUpdateNetPol(netPolObj *networkingv1
...
-		if err = ipsMgr.CreateList(list); err != nil {
-			return fmt.Errorf("[syncAddAndUpdateNetPol] Error: creating ipset list %s with err: %v", list, err)
+
+	// lists is a map with list name and members as value
+	// NPM will create the list first and increments the refer count
+	for listKey, _ := range lists {
+		if err = ipsMgr.CreateList(listKey); err != nil {
+			return fmt.Errorf("[syncAddAndUpdateNetPol] Error: creating ipset list %s with err: %v", listKey, err)
+		}
+		ipsMgr.IpSetReferIncOrDec(listKey, util.IpsetSetListFlag, ipsm.IncrementOp)
```
JungukCho, line 361: better to include this call in CreateList(). So, we do not need to expose this function to other packages and encapsulate information in ipsMgr.

npm/parseSelector.go
```diff
@@ -14,18 +14,18 @@ import (
 // An ReqHeap is a min-heap of labelSelectorRequirements.
 type ReqHeap []metav1.LabelSelectorRequirement
 
-func (h ReqHeap) Len() int {
```
JungukCho, line 17: Where does NPM use heap?

npm/parseSelector.go
```diff
@@ -121,73 +121,170 @@ func sortSelector(selector *metav1.LabelSelector) {
 	selector.MatchExpressions = sortedReqs
 }
 
+// getSetNameForMultiValueSelector takes in label with multiple values without operator
+// and returns a new 2nd level ipset name
+func getSetNameForMultiValueSelector(key string, vals []string) string {
+	rtStr := key
```
JungukCho, line 127: What is "rtStr"?

npm/parseSelector.go
```diff
@@ -121,73 +121,170 @@ func sortSelector(selector *metav1.LabelSelector) {
...
-func parseSelector(selector *metav1.LabelSelector) ([]string, []string, []string) {
+// flattenNameSpaceSelector will help flatten multiple NameSpace selector match Expressions values
+// into multiple label selectors helping with the OR condition.
+func FlattenNameSpaceSelector(nsSelector *metav1.LabelSelector) []metav1.LabelSelector {
+	//egress section
+	// for _, egressRule := range nsSelector {
+	// 	log.Logf("%+v", egressRule)
+	// }
+	if nsSelector == nil {
+		return []metav1.LabelSelector{*nsSelector}
```
JungukCho, line 148: Not sure why it includes *nsSelector. Any reasons? Will be the same to just return nil or empty slices.

npm/parseSelector.go
```diff
@@ -121,73 +121,170 @@ func sortSelector(selector *metav1.LabelSelector) {
...
+	}
+
+	if len(nsSelector.MatchExpressions) == 0 {
+		return []metav1.LabelSelector{*nsSelector}
```
JungukCho, line 152: Is it different to just return nsSelector?

npm/parseSelector.go
```diff
@@ -121,73 +121,170 @@ func sortSelector(selector *metav1.LabelSelector) {
...
+
+	baseSelector := &metav1.LabelSelector{
+		MatchLabels:      nsSelector.MatchLabels,
+		MatchExpressions: []metav1.LabelSelectorRequirement{},
+	}
+
+	multiValuePresent := false
+	multiValueMatchExprs := []metav1.LabelSelectorRequirement{}
+	for _, req := range nsSelector.MatchExpressions {
+		switch op := req.Operator; op {
```
JungukCho, line 163: The code are exactly same in the first two cases  and in the last two cases. So, we can combine them. May be useful to use if instead switch.

npm/parseSelector.go
vakalapa: rtrStr = returnString. Corrected the variableName

npm/networkPolicyController.go
vakalapa: The reason I am not incrementing or decrementing a List count on creation is that, unless a network policy uses a IPSet, it is not considered be referred by others. So i chose to expose the IpSetReferIncOrDec to NetPol package, to increment the refer only when a NetworkPolicy uses this IPset, same when we delete a network policy is when we decrement the value

npm/networkPolicyController.go
vakalapa: As i explained above, this should not be used in createList, but only when netowrkPolicy or a 2nd level List uses an IPset

npm/parseSelector.go
vakalapa: this was earlier used in creating the single podselector for processed NPMap. this is just a lint change, and i do not want to touch this part in this PR, as the scope is different.

npm/parseSelector.go
vakalapa: The way we designed, FlattenNSSelector will return a list of selectors, so it can be looped over, thats why for an nill or empty selector, i am just returning the original selector as a list.

npm/parseSelector.go
vakalapa: See previous comment.

npm/parseSelector.go
rbtr: @vakalapa `returnString` is not clearer; in Go it is better to name the variable by what it holds, not by its type or use - "~returnString~ is the ... (`IPSetName` maybe?)"
maybe that is what @JungukCho was getting at

npm/parseSelector.go
JungukCho: Is "[]metav1.LabelSelector{*nsSelector} =>  []metav1.LabelSelector{}" different if nsSelector is nil?

npm/parseSelector_test.go
```diff
@@ -488,41 +446,113 @@ func TestParseSelector(t *testing.T) {
...
+				},
+			},
+		},
+		MatchLabels: map[string]string{
+			"c": "d",
+			"a": "b",
+		},
+	}
+
+	testSelectors := FlattenNameSpaceSelector(firstSelector)
```
JungukCho, line 511: Will be very helpful if there are comparison between expected values and testSelectors to easily understand what expected values are.

npm/parseSelector_test.go
```diff
@@ -488,41 +446,113 @@ func TestParseSelector(t *testing.T) {
...
+				},
+			},
+		},
+		MatchLabels: map[string]string{
+			"c": "d",
+			"a": "b",
+		},
+	}
+
+	testSelectors = FlattenNameSpaceSelector(secondSelector)
```
JungukCho, line 554: Will be very helpful if there are comparison between expected values and testSelectors to easily understand what expected values are from FlattenNameSpaceSelector and improve debugging capability of UTs.

npm/parseSelector.go
```diff
@@ -121,73 +121,162 @@ func sortSelector(selector *metav1.LabelSelector) {
...
+	}
+
+	multiValuePresent := false
+	multiValueMatchExprs := []metav1.LabelSelectorRequirement{}
+	for _, req := range nsSelector.MatchExpressions {
+
+		if (req.Operator == metav1.LabelSelectorOpIn) || (req.Operator == metav1.LabelSelectorOpNotIn) {
+			if len(req.Values) == 1 {
+				baseSelector.MatchExpressions = append(baseSelector.MatchExpressions, req)
+				continue
```
JungukCho, line 167: May be better readability to use "else" instead of "continue"

npm/parseSelector.go
```diff
@@ -121,73 +121,162 @@ func sortSelector(selector *metav1.LabelSelector) {
...
+		}
+	}
+
+	// If there are no multiValue NS selector match expressions
+	// return the original NsSelector
+	if !multiValuePresent {
+		return []metav1.LabelSelector{*nsSelector}
+	}
+
+	// Now use the baseSelector and loop over multiValueMatchExprs to create all
```
JungukCho, line 185: Highly suggest moving codes from 185-192 to "zipMatchExprs". It will improve readability. Now, I need to track for loop in 190 with zipMatchExprs function which uses returned values as input again.

npm/parseSelector.go
```diff
@@ -121,73 +121,162 @@ func sortSelector(selector *metav1.LabelSelector) {
...
+	for _, req := range multiValueMatchExprs {
+		flatNsSelectors = zipMatchExprs(flatNsSelectors, req)
+	}
+
+	return flatNsSelectors
+}
+
+func zipMatchExprs(baseSelectors []metav1.LabelSelector, matchExpr metav1.LabelSelectorRequirement) []metav1.LabelSelector {
+	zippedLabelSelectors := []metav1.LabelSelector{}
+	for _, selector := range baseSelectors {
```
JungukCho, line 199: Can you check whether this work or not when nsSelector does not have "MatchLabels" fields and have only multiple values in "MatchExpressions" field? Seems it does not iterate for loop at all. Also add this case in UTs to confirm this function.

npm/parseSelector.go
JungukCho: Or moving codes from "zipMatchExprs" to here. The function name of "zipMatchExprs" is not very intuitive at least to me.  I think "FlattenNameSpaceSelector" function has a good name and combining all codes in the function is good to understand and not deep.

npm/translatePolicy.go
```diff
@@ -113,17 +114,44 @@ func craftPartialIptEntrySpecFromOpsAndLabels(ns string, ops, labels []string, s
...
+func craftPartialIptEntrySpecFromSelector(ns string, selector *metav1.LabelSelector, srcOrDstFlag string, isNamespaceSelector bool) ([]string, []string, map[string][]string) {
+	// parse the sector into labels and maps of multiVal match Exprs
+	labelsWithOps, nsLabelListKVs := parseSelector(selector)
 	ops, labels := GetOperatorsAndLabels(labelsWithOps)
-	return craftPartialIptEntrySpecFromOpsAndLabels(ns, ops, labels, srcOrDstFlag, isNamespaceSelector)
+
+	valueLabels := []string{}
+	listLabelsWithMembers := make(map[string][]string)
+	labelsForSpec := labels
+	// TODO add comments very confusing for sure
```
JungukCho, line 134: Can you add the detailed comments in this PR? It is good to have comments now when you have contexts in details and helps reviwers to understand the codes as well.

npm/translatePolicy.go
```diff
@@ -113,17 +114,44 @@ func craftPartialIptEntrySpecFromOpsAndLabels(ns string, ops, labels []string, s
...
 	}
 
 	return spec
 }
 
 // craftPartialIptEntrySpecFromSelector :- ns must be "" for namespace selectors
-func craftPartialIptEntrySpecFromSelector(ns string, selector *metav1.LabelSelector, srcOrDstFlag string, isNamespaceSelector bool) []string {
-	labelsWithOps, _, _ := parseSelector(selector)
+// TODO check all references of this func
+func craftPartialIptEntrySpecFromSelector(ns string, selector *metav1.LabelSelector, srcOrDstFlag string, isNamespaceSelector bool) ([]string, []string, map[string][]string) {
```
JungukCho, line 126: Will be very helpful to add comment of this function

npm/translatePolicy.go
```diff
@@ -160,11 +195,25 @@ func craftPartialIptablesCommentFromSelector(ns string, selector *metav1.LabelSe
...
+		}
+	}
+}
+
+func translateIngress(ns string, policyName string, targetSelector metav1.LabelSelector, rules []networkingv1.NetworkPolicyIngressRule) ([]string, []string, map[string][]string, [][]string, []*iptm.IptEntry) {
 	var (
-		sets                                  []string // ipsets with type: net:hash
-		namedPorts                            []string // ipsets with type: hash:ip,port
-		lists                                 []string // ipsets with type: list:set
+		sets                                  []string            // ipsets with type: net:hash
```
JungukCho, line 214: nitPick: better to name netHashIPsets?

npm/translatePolicy.go
```diff
@@ -160,11 +195,25 @@ func craftPartialIptablesCommentFromSelector(ns string, selector *metav1.LabelSe
...
+		namedPorts                            []string            // ipsets with type: hash:ip,port
+		lists                                 map[string][]string // ipsets with type: list:set
```
JungukCho, line 216: nitPick: better to name listIPsets?

npm/translatePolicy.go
```diff
@@ -423,21 +473,129 @@ func translateIngress(ns string, policyName string, targetSelector metav1.LabelS
...
-				nsLabelsWithOps, _, _ := parseSelector(fromRule.NamespaceSelector)
-				_, nsLabelsWithoutOps := GetOperatorsAndLabels(nsLabelsWithOps)
-				if len(nsLabelsWithoutOps) == 1 && nsLabelsWithoutOps[0] == "" {
-					// Empty namespaceSelector. This selects all namespaces
-					nsLabelsWithoutOps[0] = util.KubeAllNamespacesFlag
-				} else {
-					for i, _ := range nsLabelsWithoutOps {
-						// Add namespaces prefix to distinguish namespace ipset lists and pod ipsets
-						nsLabelsWithoutOps[i] = "ns-" + nsLabelsWithoutOps[i]
+				for _, nsSelector := range FlattenNameSpaceSelector(fromRule.NamespaceSelector) {
```
JungukCho, line 476: Prefer to splitting this one line to two lines to understand what it returns. Same to all in this file.

JungukCho:
It would be nice to leave comments to describe high-level behaviors (not very details) of your updates in this PR.

npm/parseSelector.go
```diff
@@ -121,73 +121,162 @@ func sortSelector(selector *metav1.LabelSelector) {
...
-	vals = append(vals, sortedVals...)
 
 	for _, req := range selector.MatchExpressions {
 		var k string
 		switch op := req.Operator; op {
 		case metav1.LabelSelectorOpIn:
+			k = req.Key
+			if len(req.Values) == 1 {
+				labels = append(labels, k+":"+req.Values[0])
+				continue
```
JungukCho, line 251: better to use else instead of continue?

npm/parseSelector.go
```diff
@@ -121,73 +121,162 @@ func sortSelector(selector *metav1.LabelSelector) {
...
-				keys = append(keys, k)
-				vals = append(vals, v)
-				labels = append(labels, k+":"+v)
+				vals[k] = append(vals[k], v)
 			}
 		case metav1.LabelSelectorOpNotIn:
+			k = util.IptablesNotFlag + req.Key
+			if len(req.Values) == 1 {
+				labels = append(labels, k+":"+req.Values[0])
+				continue
```
JungukCho, line 263: better to use else instead of continue?

npm/parseSelector.go
vakalapa: Added the UT to confirm and it works as expected. Without MatchLabel it just loops over the matchExprs

npm/parseSelector_test.go
vakalapa: Very good idea, i added more deepEqual checks to make sure the output is as expected.

npm/parseSelector_test.go
vakalapa: Agreed, added comparison values

npm/parseSelector.go
vakalapa: I modelled it as a recursive function, where we take in a slice and expand it using each expr.

npm/parseSelector_test.go
```diff
@@ -513,6 +513,10 @@ func TestFlattenNameSpaceSelector(t *testing.T) {
 		t.Errorf("TestFlattenNameSpaceSelector failed @ 1st selector length check %+v", testSelectors)
 	}
 
+	if !reflect.DeepEqual(testSelectors[0], *firstSelector) {
```
JungukCho, line 516: There are multiple time to check length of returned values from FlattenNameSpaceSelector and deepEqual in this function and TestFlattenNameSpaceSelectorWoMatchLabels function. So, I think it is better to make one helper function to check them together.

npm/parseSelector_test.go
```diff
@@ -555,4 +559,389 @@ func TestFlattenNameSpaceSelector(t *testing.T) {
...
+		},
+	}
+
+	if !reflect.DeepEqual(expectedSelectors, testSelectors) {
+		t.Errorf("TestFlattenNameSpaceSelector failed @ 2nd selector deepEqual check.\n Expected: %+v \n Actual: %+v", expectedSelectors, testSelectors)
+	}
+}
+
+func TestFlattenNameSpaceSelectorWoMatchLabels(t *testing.T) {
+	firstSelector := &metav1.LabelSelector{
```
JungukCho, line 844: Same here. You can make it concise and better readability by using variables to avoid unnecessary repeat.

npm/networkPolicyController.go
JungukCho: In this case, I prefer to having another function which functions "CreateList and IPSetReferIncOrDec". It will be easy to remember and help to encapsulate details. 
I assume that your next PR is about those reference counts, we can revisit this in the PR.

npm/ipsm/ipsm.go
```diff
@@ -462,6 +462,33 @@ func (ipsMgr *IpsetManager) DeleteFromSet(setName, ip, podKey string) error {
 	return nil
 }
 
+// Clean removes all the empty sets & lists under the namespace.
```
JungukCho: When do we use this function?

npm/ipsm/ipsm.go
```diff
@@ -87,7 +87,7 @@ func (ipsMgr *IpsetManager) Exists(listName string, setName string, kind string)
 	return true
 }
 
-// IpSetReferIncOrDec checks if an element exists in setMap/listMap and then increases or decreases tis refer count.
+// IpSetReferIncOrDec checks if an element exists in setMap/listMap and then increases or decreases this referCount.
 func (ipsMgr *IpsetManager) IpSetReferIncOrDec(ipsetName string, kind string, countOperation ReferCountOperation) {
```
JungukCho, line 91: Do you have plan to extend this function for other types in addition to "list" as well?
In this PR, kind is always "list".

npm/networkPolicyController.go
vakalapa: I agree, in refer count work item, we can look at different scenarios and better organize these functions.

npm/ipsm/ipsm.go
vakalapa: Yes in the next work item(refer coutn) we will extend support of this IncDec to SetMap where we will track the use of iptable rules and their references to a given ipset. we will track both kinds of ipsets.

npm/ipsm/ipsm.go
vakalapa: We will need to expose this to network policy controller as netpol is the one referencing out ipsetmaps

npm/parseSelector.go
```diff
@@ -140,10 +140,41 @@ func HashSelector(selector *metav1.LabelSelector) string {
 // flattenNameSpaceSelector will help flatten multiple NameSpace selector match Expressions values
 // into multiple label selectors helping with the OR condition.
 func FlattenNameSpaceSelector(nsSelector *metav1.LabelSelector) []metav1.LabelSelector {
-	//egress section
-	// for _, egressRule := range nsSelector {
-	// 	log.Logf("%+v", egressRule)
-	// }
+	/*
```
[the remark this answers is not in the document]
vakalapa: This comes back to what we discussed in our meeting today, having a visual representation of what the YAML looks like and what the transformation looks like will help other developers to quickly understand. I would like to keep this as is if its ok.

npm/parseSelector_test.go
vakalapa: Makes sense.

npm/ipsm/ipsm.go
vakalapa: Clean() function is not being currently, this is for future work items.

npm/parseSelector_test.go
```diff
@@ -555,4 +559,389 @@ func TestFlattenNameSpaceSelector(t *testing.T) {
 	if len(testSelectors) != 8 {
 		t.Errorf("TestFlattenNameSpaceSelector failed @ 2nd selector length check %+v", testSelectors)
 	}
+
+	expectedSelectors := []metav1.LabelSelector{
```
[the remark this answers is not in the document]
vakalapa: added a common matchLabel to help with readability

npm/parseSelector.go
JungukCho: Luckily, no one from translate.go calls this FlattenNameSpaceSelector function with **nil nsSelector** value in current NPM. 
However, we are not sure who calls this function in the future. So, it is good to check nil value at the first place.

However, in that case, it should be below code if "nsSelector" is nil. 
Otherwise, it will be panic. We cannot dereference nil value.
```go
return []metav1.LabelSelector{}
```

JungukCho: I think this is a good lesson. We should add various and edge cases in UTs. 
Suggest adding UT for nil test for this function.

npm/parseSelector.go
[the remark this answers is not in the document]
JungukCho: To be clear, I suggested putting comprehensive information (e.g., pods yaml and netpol yaml which shows the problem in our NPM) to reproduce the problem, which helps to understand big picture at the first. T
his code is very straightforward. So, I don't think we need this, but if you want to keep it, I don't have strong opinion.

npm/ipsm/ipsm.go
JungukCho: I would suggest not to add function which is not used in this PR. I think this is a good practice. 
There are multiple reasons. 
1. Reviewer do not need to spend time. 
2. Add the function with the PR for your future work items. So, it will be well aligned and easy to understand PR. 
3. There are possibility that this function will not be used in the future. At that time, this function just stays for long time. 

I think you have future work item to use this function. So, add this function in the PR.

JungukCho (approved):
Thank you for applying comments. Look forward to your next PR.

JungukCho (approved):

vakalapa:
Merging, we got a green in Cyclonus.

PR #864, opened 2021-05-03
[NPM] Filter pod having ipv6

JungukCho:
**Reason for Change**:

This PR is to filter a pod in pod event handler functions when the pod has ipv6 address since NPM does not support ipv6 yet.
In addition, it also checks a valid ipv4 address format of the pod.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation (in NPM/IPTables and IPSet error handling dir in acndoc)
- [x] adds unit tests

**Notes**:
We need to revisit error code handling in ipsm.go and iptm.go.
One example is that ipsm caches the item even though invalid IPv4 or IPv6 address are inserted (returned error code is 1).
Further it causes controllers to process the rest of operations.
While this PR filters invalid IPv4 and IPv6 address in early stage to avoid the situation, we need to revisit all codes thoroughly.
We may need to use this code #840

### DIFF
--- a/npm/ipsm/ipsm.go
+++ b/npm/ipsm/ipsm.go
@@ -350,7 +350,8 @@ func (ipsMgr *IpsetManager) AddToSet(setName, ip, spec, podKey string) error {
 	}
 
 	// TODO(jungukcho): revisit all error code handling codes in ipsm.go and iptm.go.
-	// For example, in below code, ipsm caches the item even though invalid IPv4 or IPv6 address are inserted in case returned error code is 1.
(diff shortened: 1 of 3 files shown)


vakalapa (dismissed):

JungukCho:
I realized observed error message related to IPv6 is from network policy. 
So, I need to update networkPolicyController.go or ipsm.go to handle the issue correctly. 
I may create another PR to resolve it.

JungukCho:
Close now. I will link a new PR which will resolve the root cause of this issue (netpol with ipv6 cidr value) in this PR later.

PR #867, opened 2021-05-10
Fix to bubble up the error and restart in case it fails to create CRD…

neaggarwMS:
Crash Azure-CNS in case it fails to start the CRD Request Controller. 

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -504,7 +504,11 @@ func main() {
 	if config.ChannelMode == cns.CRD {
 		requestControllerStopChannel := make(chan struct{})
 		defer close(requestControllerStopChannel)
-		IniitalizeCRDState(httpRestService, cnsconfig, requestControllerStopChannel)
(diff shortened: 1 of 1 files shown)


cns/service/main.go
```diff
@@ -667,22 +671,22 @@ func main() {
...
-		return
+		return err
 	}
 
 	//convert interface type to implementation type
 	httpRestServiceImplementation, ok := httpRestService.(*restserver.HTTPRestService)
 	if !ok {
 		logger.Errorf("[Azure CNS] Failed to convert interface httpRestService to implementation: %v", httpRestService)
-		return
+		return err
```
ramiro-gamarra, line 689: this returns a nil error, which I don't think is intended

neaggarwMS: Thanks for the catch, fixed it

neaggarwMS:
azp /run

pjohnst5 (approved):
LGTM

PR #868, opened 2021-05-13
[NPM] Use fakeexec for ipsm and iptm tests

matmerr:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


npm/ipsm/ipsm_test.go
```diff
@@ -481,19 +473,16 @@ func TestDeleteFromSetWithPodCache(t *testing.T) {
...
-			t.Errorf("TestClean failed @ ipsMgr.Restore")
-		}
-	}()
+	fexec := testutils.GetFakeExecWithScripts(calls)
+	ipsMgr := NewIpsetManager(fexec)
+	defer testingutils.VerifyCalls(t, fexec, calls)
 
-	if err := ipsMgr.CreateSet("test-set", append([]string{util.IpsetNetHashFlag})); err != nil {
-		t.Errorf("TestClean failed @ ipsMgr.CreateSet with err %+v", err)
+	if err := ipsMgr.Save(util.IpsetTestConfigFile); err != nil {
```
matmerr, line 484: remove this and update TestClean test, make actual calls in Clean step

JungukCho (commented):
First, I am sorry for late review.
I walked through the changes once and have few suggestions before walking though PR in details since the first suggestion may need some changes if you are ok.

1. I am not sure it is good idea to have api and fakes directories (i.e., different packages) and files in the directory. Especially they are very specific for iptables (e.g., one method name is "GrabIptablesLocks").
Instead, I would prefer to putting files from api and fakes directories into iptm directory together and use fakes/iptm.go -> iptm/iptm_fakes.go and then encapsulate all methods since they are in the same package.

Do you have specific reasons to have this structures which I missed?

2.  I am not sure, but it would be good to separate "cyclonous-netpol-test.yaml" and "repo-hygiene.yaml", and "cyclonus-test.txt" into another PR and merge them if it is beneficial to merge them early.

3. I think it is ok to remove workqueue related commit since we resolved it.

npm/ipsm/ipsm_test.go
```diff
@@ -294,45 +283,42 @@ func TestDeleteSet(t *testing.T) {
...
 	}
 
-	if err := ipsMgr.AddToSet(testSetName, "1.2.3.4/nomatch", util.IpsetNetHashFlag, ""); err != nil {
-		t.Fatalf("TestAddToSet with nomatch failed @ ipsMgr.AddToSet %v", err)
-	}
+	fexec := testutils.GetFakeExecWithScripts(calls)
+	ipsMgr := NewIpsetManager(fexec)
+	defer testingutils.VerifyCalls(t, fexec, calls)
+
+	err := ipsMgr.AddToSet(testSetName, "1.2.3.4", util.IpsetNetHashFlag, "")
```
JungukCho, line 301: Can we leverage table driven tests to maintain and understand testsets better?
For example, we use below codes to have the same tests. Since there are no dependencies between calls, I think below logics will be easier to maintain testsets later (e.g., adding more testsets) and understand testsets as well.
```go
	testSetName := "test-set"
	var calls = []testutils.TestCmd{
		{Cmd: []string{"ipset", "-N", "-exist", util.GetHashedName(testSetName), "nethash"}},
		{Cmd: []string{"ipset", "-A", "-exist", util.GetHashedName(testSetName), "1.2.3.4"}},
		{Cmd: []string{"ipset", "-A", "-exist", util.GetHashedName(testSetName), "1.2.3.4/", "nomatch"}},
		{Cmd: []string{"ipset", "-A", "-exist", util.GetHashedName(testSetName), "1.1.1.1,:"}}, // todo: verify this is proper formatting
		{Cmd: []string{"ipset", "-A", "-exist", util.GetHashedName(testSetName), "1.1.1.1"}},
		{Cmd: []string{"ipset", "-A", "-exist", util.GetHashedName(testSetName), "1.1.1.1,tcp:8080"}},
	}

	fexec := testutils.GetFakeExecWithScripts(calls)
	ipsMgr := NewIpsetManager(fexec)
	defer testingutils.VerifyCalls(t, fexec, calls)

	ipsets := [][]string{
		{testSetName, "1.2.3.4", util.IpsetNetHashFlag, ""},
		{testSetName, "1.2.3.4/nomatch", util.IpsetNetHashFlag, ""},
		{testSetName, fmt.Sprintf("%s,:", "1.1.1.1"), util.IpsetIPPortHashFlag, "0"},
		{testSetName, fmt.Sprintf("%s", "1.1.1.1"), util.IpsetIPPortHashFlag, "0"},
		{testSetName, fmt.Sprintf("%s,%s:%d", "1.1.1.1", "tcp", 8080), util.IpsetIPPortHashFlag, "0"},
	}

	for _, ipset := range ipsets {
		err := ipsMgr.AddToSet(ipset[0], ipset[1], ipset[2], ipset[3])
		require.NoError(t, err)
	}

	ipsets = [][]string{
		{testSetName, fmt.Sprintf("%s,%s:%d", "", "tcp", 8080), util.IpsetIPPortHashFlag, "0"},
		{testSetName, fmt.Sprintf(""), util.IpsetIPPortHashFlag, "0"},
	}

	for _, ipset := range ipsets {
		err := ipsMgr.AddToSet(ipset[0], ipset[1], ipset[2], ipset[3])
		require.Error(t, err)
	}
```

If it looks reasonable, would you please walk through ipsm_test.go to take same approaches?
I think there are multiple places where we can leverage table driven tests (e.g., TestCreateSet, TestDeleteFromList, etc).

In addition, in all testsets, "test-list" and "test-set" are used. So, I think it is ok to declare them as global variables in this file  and use them. So, when we want to change them to other values, we can change only one line.

npm/ipsm/ipsm_test.go
```diff
@@ -568,28 +556,35 @@ func TestRun(t *testing.T) {
...
 
 	var calls = []testutils.TestCmd{
 		{Cmd: []string{"ipset", "-N", "-exist", util.GetHashedName(setname), "nethash"}, Stdout: "test failure", ExitCode: 2},
 	}
 
-	fexec, fcmd := testutils.GetFakeExecWithScripts(calls)
-
+	fexec := testutils.GetFakeExecWithScripts(calls)
 	ipsMgr := NewIpsetManager(fexec)
+	defer testingutils.VerifyCalls(t, fexec, calls)
```
JungukCho, line 567: This testset has failed in azure pipeline. 
I think the reason is below code.
```go
# VerifyCalls in utils.go.
require.Nil(t, err) 
```  
It seems using "require.NotNil(t, err)" is right for this testset. 
Please check it.

npm/iptm/iptm_test.go
```diff
@@ -7,77 +7,145 @@ import (
...
-	iptMgr := &IptablesManager{}
+	var calls = []testutils.TestCmd{
+		{Cmd: []string{"iptables-save"}},
+	}
+
+	fexec := testutils.GetFakeExecWithScripts(calls)
+	defer testutils.VerifyCalls(t, fexec, calls)
+	iptMgr := NewIptablesManager(fexec, NewFakeIptOperationShim())
+
 	if err := iptMgr.Save(util.IptablesTestConfigFile); err != nil {
```
JungukCho, line 98: This causes "permission denied" when I ran it locally.  
So, only for testing Save function, I think using temporal file is better which you used in ipsm_test.go.

npm/ipsm/ipsm_test.go
rbtr: since we're importing testify anyway, the convenience methods read better than checking nil:
`assert.Error(t, err)` or `assert.NoError(t, err)`, whichever it should be

npm/iptm/iptm_test.go
```diff
@@ -7,77 +7,145 @@ import (
...
-	if err := iptMgr.Save(util.IptablesTestConfigFile); err != nil {
-		t.Errorf("TestUninitNpmChains failed @ iptMgr.Save")
-	}
-
-	defer func() {
-		if err := iptMgr.Restore(util.IptablesTestConfigFile); err != nil {
-			t.Errorf("TestUninitNpmChains failed @ iptMgr.Restore")
-		}
-	}()
+	var calls = unInitCalls
```
JungukCho, line 130: nitpick: I think just using unInitCalls is ok. Are there any issues?

npm/iptm/iptm_test.go
```diff
@@ -7,77 +7,145 @@ import (
...
+
 	if err := iptMgr.Restore(util.IptablesTestConfigFile); err != nil {
-		t.Errorf("TestRestore failed @ iptMgr.Restore")
+		t.Errorf("TestRestore failed @ iptMgr.Restore with err %v", err)
 	}
 }
 
 func TestInitNpmChains(t *testing.T) {
-	iptMgr := &IptablesManager{}
+	var calls = initCalls
```
JungukCho, line 118: nitpick: I think just using initCalls is ok. Are there any issues?

npm/ipsm/ipsm_test.go
matmerr: I typically like `require.NoError` for failfast behavior, I used it elsewhere in this PR so I'll update for consistency

npm/iptm/iptm_test.go
```diff
@@ -7,77 +7,145 @@ import (
 	"github.com/Azure/azure-container-networking/npm/metrics"
 	"github.com/Azure/azure-container-networking/npm/metrics/promutil"
 	"github.com/Azure/azure-container-networking/npm/util"
+	testutils "github.com/Azure/azure-container-networking/test/utils"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	initCalls = []testutils.TestCmd{
```
JungukCho, line 15: I think it is really good to divide this initCalls slide into three slices according to "AddAllChains()", "CheckAndAddForwardChain()", and "AddAllRulesToChains()". Then merging them into initCalls. It will help better readability and maintenance and help reviewers to understand codes.

npm/iptm/iptm_test.go
```diff
@@ -7,77 +7,145 @@ import (
...
+		{Cmd: []string{"iptables", "-w", "60", "-C", "AZURE-NPM-EGRESS", "-j", "RETURN", "-m", "mark", "--mark", "0x1000", "-m", "comment", "--comment", "RETURN-on-EGRESS-mark-0x1000"}},
+		{Cmd: []string{"iptables", "-w", "60", "-C", "AZURE-NPM-EGRESS", "-j", "AZURE-NPM-EGRESS-DROPS"}},
+		{Cmd: []string{"iptables", "-w", "60", "-C", "AZURE-NPM-EGRESS-PORT", "-j", "RETURN", "-m", "mark", "--mark", "0x3000", "-m", "comment", "--comment", "RETURN-on-EGRESS-and-INGRESS-mark-0x3000"}},
+		{Cmd: []string{"iptables", "-w", "60", "-C", "AZURE-NPM-EGRESS-PORT", "-j", "RETURN", "-m", "mark", "--mark", "0x1000", "-m", "comment", "--comment", "RETURN-on-EGRESS-mark-0x1000"}},
+		{Cmd: []string{"iptables", "-w", "60", "-C", "AZURE-NPM-EGRESS-TO", "-j", "RETURN", "-m", "mark", "--mark", "0x3000", "-m", "comment", "--comment", "RETURN-on-EGRESS-and-INGRESS-mark-0x3000"}},
+		{Cmd: []string{"iptables", "-w", "60", "-C", "AZURE-NPM-EGRESS-TO", "-j", "RETURN", "-m", "mark", "--mark", "0x1000", "-m", "comment", "--comment", "RETURN-on-EGRESS-mark-0x1000"}},
+	}
+
+	unInitCalls = []testutils.TestCmd{
```
JungukCho, line 60: Same as initCalls if it clear division is possible.

npm/iptm/ioshim.go
```diff
@@ -0,0 +1,13 @@
+package iptm
+
+import (
+	"io"
+	"os"
+)
+
+type IOShim interface {
+	GrabIptablesLocks() (*os.File, error)
```
JungukCho, line 9: I am still thinking all of the methods in this interface can be isolated (i.e., non-capital). 
So, corresponding structs will be also isolated from other packages. 
Do I misunderstand something?

npm/iptm/iptm_test.go
```diff
@@ -7,77 +7,145 @@ import (
...
+		{Cmd: []string{"iptables", "-w", "60", "-N", "AZURE-NPM-EGRESS"}},
+		{Cmd: []string{"iptables", "-w", "60", "-N", "AZURE-NPM-INGRESS-PORT"}},
+		{Cmd: []string{"iptables", "-w", "60", "-N", "AZURE-NPM-INGRESS-FROM"}},
+		{Cmd: []string{"iptables", "-w", "60", "-N", "AZURE-NPM-EGRESS-PORT"}},
+		{Cmd: []string{"iptables", "-w", "60", "-N", "AZURE-NPM-EGRESS-TO"}},
+		{Cmd: []string{"iptables", "-w", "60", "-N", "AZURE-NPM-INGRESS-DROPS"}},
+		{Cmd: []string{"iptables", "-w", "60", "-N", "AZURE-NPM-EGRESS-DROPS"}},
+		{Cmd: []string{"iptables", "-w", "60", "-N", "AZURE-NPM"}},
+
+		{Cmd: []string{"iptables", "-t", "filter", "-n", "--list", "FORWARD", "--line-numbers"}, Stdout: "3  "}, // THIS IS THE GREP CALL
```
JungukCho, line 28: What does "3" mean? I think it is better to use a meaningful variable since it is used in multiple places in this code.

npm/iptm/iptm_test.go
```diff
@@ -7,77 +7,145 @@ import (
...
+		{Cmd: []string{"grep", "KUBE-SERVICES"}, Stdout: "4  "},
```
JungukCho, line 29: Same here. What does "4" mean in this code?

npm/iptm/iptm_test.go
```diff
@@ -223,123 +280,29 @@ func TestRun(t *testing.T) {
...
-	case !npmExists:
-		t.Errorf("TestGetChainLineNumber @ after Init chains line number check iptMgr.GetChainLineNumber with failed to Add chain ")
-	}
-
-	if err = iptMgr.UninitNpmChains(); err != nil {
-		t.Errorf("TestGetChainLineNumber @ iptMgr.UninitNpmChains")
-	}
+	lineNum, err := iptMgr.GetChainLineNumber(util.IptablesAzureChain, util.IptablesForwardChain)
+	require.NoError(t, err)
+	require.Equal(t, lineNum, 3)
```
JungukCho, line 298: Beneficial to use variable or put comment. Based on current code base, it should be located in the third line of forward chain.

test/utils/utils.go
```diff
@@ -14,38 +16,39 @@ import (
...
 
 	fcmd := &fakeexec.FakeCmd{}
 
 	for _, call := range calls {
 		stdout := call.Stdout
+		ccmd := call.Cmd
 		if call.ExitCode != 0 {
 			err := &fakeexec.FakeExitError{Status: call.ExitCode}
-			fcmd.CombinedOutputScript = append(fcmd.CombinedOutputScript, func() ([]byte, []byte, error) { return []byte(stdout), nil, err })
+			fcmd.CombinedOutputScript = append(fcmd.CombinedOutputScript, func() ([]byte, []byte, error) { return []byte(stdout), []byte{}, err })
```
JungukCho, line 33: Curious - where []byte{} is used and what is purpose of it? 
It seems fakeexec does not even use this return value ([]byte{}).

npm/iptm/iptm_test.go
```diff
@@ -223,123 +280,29 @@ func TestRun(t *testing.T) {
...
 }
 
 func TestMain(m *testing.M) {
 	metrics.InitializeAll()
-	iptMgr := NewIptablesManager()
-	iptMgr.Save(util.IptablesConfigFile)
+	metrics.CreateFakeTelemetryHandle()
```
JungukCho, line 303: I think this is to avoid sending logs from functions (e.g., SendErrorLogAndMetric()) in UTs. Is it right? 

It is not scope of this PR and I am not familiar with metrics package code yet, but I feel like it would be nice to remove this dependency (e.g., metrics.InitializeAll()) to run UTs and revisit metrics package code within next few PRs to better structures if needed. Any thoughts?

JungukCho (commented):
In next PR, I think we can add more UTs (AddAllChains(), CheckAndAddForwardChain(), AddAllRulesToChains(), etc) for iptm_test.go even though they are called from InitNpmChains().

test/utils/utils.go
matmerr: this is true, nil is set on line 35 and it should be nil here too

npm/ipsm/ipsm_test.go
```diff
@@ -4,95 +4,97 @@ package ipsm
...
+	fexec := testutils.GetFakeExecWithScripts(calls)
+	ipsMgr := NewIpsetManager(fexec)
+	defer testingutils.VerifyCalls(t, fexec, calls)
+	err := ipsMgr.Save("ipset.conf")
+	require.NoError(t, err)
 }
 
 func TestRestore(t *testing.T) {
-	ipsMgr := NewIpsetManager(exec.New())
+	tmpFile, err := ioutil.TempFile(os.TempDir(), filepath.Base(util.IpsetTestConfigFile))
```
JungukCho, line 33: Prefer to having a brief comment about filepath.Base like only using last element in path, but no strong opinion.

npm/iptm/ioshim.go
```diff
@@ -0,0 +1,93 @@
...
+
+type ioshim interface {
+	lockIptables() error
+	unlockIptables() error
+	openConfigFile(configFile string) (io.Reader, error)
+	createConfigFile(configFile string) (io.Writer, error)
+	closeConfigFile() error
+}
+
+type IptOperationShim struct {
```
JungukCho, line 22: We missed this one. IptOperationShim can be **i**ptOperationShim.

npm/iptm/ioshim.go
```diff
@@ -0,0 +1,93 @@
...
+	configFile   *os.File
+	iptablesLock *os.File
```
JungukCho, line 24: Prefer to using iptablesLockFile.

npm/iptm/ioshim.go
```diff
@@ -0,0 +1,93 @@
...
+	// Grab 1.6.x style lock.
+	var err error
+	i.iptablesLock, err = os.OpenFile(util.IptablesLockFile, os.O_CREATE, 0600)
+	if err != nil {
+		metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to open iptables lock file %s.", util.IptablesLockFile)
+		return err
+	}
+
+	if err := wait.PollImmediate(200*time.Millisecond, 2*time.Second, func() (bool, error) {
+		if err := grabIptablesFileLock(i.iptablesLock); err != nil {
```
JungukCho, line 74: I am not sure that we need separate a file (i.e., iptm_linux.go) for only this function ("grabIptablesFileLock"). 
Prefer to combining the function into this file (i.e., ioshim.go) together.

JungukCho: This is not the scope of this PR and good to discuss it. However, we may need to walk through and explore locking for iptables.   As far as I understood running iptables automatically holds lock in a kernel. So, I am wondering we need to manage lock in our code as well. It seems it is redundant and dangerous. For example,  **Save** function in iptm.go is dangerous since if lock (xtables.lock) is hold and if unlockIptables() are failed, the other operation cannot work.

npm/iptm/ioshim_fake.go
```diff
@@ -0,0 +1,77 @@
...
+	"io"
+	"io/fs"
+	"testing/fstest"
+
+	testingutils "github.com/Azure/azure-container-networking/test/utils"
+)
+
+type fakeIptOperationShim struct {
+	configname string
+	m          fstest.MapFS
```
JungukCho, line 13: I am wondering whether m and fo are necessary? Seems we can take the same approaches (e.g., just returning nil or etc). Are there any reason to have them?

npm/npm_test.go
```diff
@@ -46,10 +46,11 @@ func newNPMgr(t *testing.T, exec utilexec.Interface) *NetworkPolicyManager {
 
 func TestMain(m *testing.M) {
 	metrics.InitializeAll()
-	iptMgr := iptm.NewIptablesManager()
```
JungukCho, line 49: Do we still need iptMgr and ipsMgr and methods of them in npm_test.go since we do not use any real management of iptables and ipset in our UTs with this PR? I assume we do not need them here anymore. If so, it is safe to remove them to avoid uncertain situation in test environments from managements of real iptables and ipset.

npm/iptm/ioshim.go
```diff
@@ -0,0 +1,93 @@
...
+	}); err != nil {
+		metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to acquire new iptables lock: %v.", err)
+		return err
+	}
+
+	success = true
+	return nil
+}
+
+func (i *IptOperationShim) unlockIptables() error {
```
JungukCho, line 88: Useful to put comment like closing file automatically releases a lock for the file.

npm/iptm/ioshim_fake.go
matmerr: they are used to create an in memory file structure for tests to behave as closest as a real fs as possible

.pipelines/pipeline.yaml
```diff
@@ -37,7 +41,7 @@ stages:
         pool:
           name: "$(BUILD_POOL_NAME_DEFAULT)"
         steps:
-
+
```
matmerr, line 44: not needed

JungukCho (approved):
lgtm!

PR #879, opened 2021-05-26
disable cleaning up hns network and state files from windows cni

tamilmani1989:
**Reason for Change**:

Disable cleaning up hns network and state files from windows cni. Since state and hns network is cleaned up by CRI startup script, no need for cni to cleanup those. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This PR is workaround for the issue: https://github.com/Azure/azure-container-networking/issues/947


cnm/ipam/ipam.go
```diff
@@ -21,6 +21,7 @@ const (
 	requiresRequestReplay = false
 	returnCode            = 0
 	returnStr             = "Success"
+	rehydrateIpamInfoOnReboot = true
```
pjohnst5, line 24: I thought this should be "false"?

ipam/manager.go
```diff
@@ -70,13 +70,13 @@ func NewAddressManager() (AddressManager, error) {
 }
 
 // Initialize configures address manager.
-func (am *addressManager) Initialize(config *common.PluginConfig, options map[string]interface{}) error {
+func (am *addressManager) Initialize(config *common.PluginConfig, rehydrateIpamInfoOnReboot bool, options map[string]interface{}) error {
```
pjohnst5, line 73: Can you make `rehydrateIpamInfoOnReboot` a new member of `options` instead?

pjohnst5 (changes requested):

cnm/ipam/ipam.go
tamilmani1989: this is not cni..true for cnm

AbelHu:
```Since state and hns network is cleaned up by CRI startup script, no need for cni to cleanup those.``` @tamilmani1989 what is the CRI startup script? Do you mean that [kubeletstart.ps1](https://github.com/Azure/aks-engine/blob/master/staging/provisioning/windows/kubeletstart.ps1) in aks-engine?

**Comment by tamilmani1989** [comment]
> `Since state and hns network is cleaned up by CRI startup script, no need for cni to cleanup those.` @tamilmani1989 what is the CRI startup script? Do you mean that [kubeletstart.ps1](https://github.com/Azure/aks-engine/blob/master/staging/provisioning/windows/kubeletstart.ps1) in aks-engine?

Nope this one:  https://github.com/Azure/aks-engine/blob/master/staging/provisioning/windows/cleanupnetwork.ps1

File: cni/network/network.go [hunk 3]
```diff
@@ -72,6 +72,10 @@ const (
 	Baremetal ExecutionMode = "baremetal"
 )
 
+const (
+	rehydrateNetworkInfoOnReboot = false
+)
```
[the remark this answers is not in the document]
rbtr: why do we need a constant `false`? this just hides whatever the real test is from me when I read the code where it is checked (like, is this actually a "is running on windows" bit?)
I think it would be much clearer if, where ever we use this conditional, we were checking the actual underlying condition that makes this switch

rbtr (changes requested):
blocking on gofmt

cni/network/network.go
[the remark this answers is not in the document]
tamilmani1989: nope this is not running on windows.. cnm and cni uses same code path. have to rehydrate for cnm but not for cni

rbtr (approved):

tamilmani1989:
auto trigger pipeline getting failed due to some infra issue. Manually triggered pipeline for this PR: https://msazure.visualstudio.com/One/_build/results?buildId=44757594&view=results

tamilmani1989:
pipeline succeeded https://msazure.visualstudio.com/One/_build/results?buildId=44757594&view=results.

Issue #870, opened 2021-05-17
[NPM] TO/FROM rule and port rule on same PodSelector in multiple policies is having issues.


vakalapa:
**What happened:**  

If there are 2((or more) different policies being applied on the same PodSelector and these policies contain TO/FROM rule and PORT rule separately, then TO/FROM rule traffic is being dropped. This does not happen if any of the policies has  both To/FROM rules and PORT rules, then this issue does not occur.

This happens because RETURN rule for a MARKED packet is not added after EGRESS-TO or INGRESS-FROM chains. 

**How to reproduce it:**  

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-backend-to-frontend-on-port-8000-policy
  namespace: testnamespace
spec:
  policyTypes:
    - Ingress
  podSelector:
    matchLabels:
      app: frontend
  ingress:
    - from:
        - podSelector:
            matchLabels:
              app: backend
      ports:
        - port: 8000
---
 apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-from-all-ns
  namespace: testnamespace
spec:
  policyTypes:
    - Ingress
  podSelector:
    matchLabels:
      app: "frontend"
  ingress:
    - from:
        - namespaceSelector: {}
```

Apply the above policy and try to reach Pod with "app: frontend" from other namespaces' Pod, requests/ping will timeout

**How to mitigate **

Combining the Port rules and TO/FROM rules into a single policy will help mitigate the issue. We can combine the above two policies as follow:

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-backend-to-frontend-on-port-8000-policy
  namespace: testnamespace
spec:
  policyTypes:
    - Ingress
  podSelector:
    matchLabels:
      app: frontend
  ingress:
    - from:
        - podSelector:
            matchLabels:
              app: backend
      ports:
        - port: 8000
    - from:
        - namespaceSelector: {}
```

This does not change the way policy behaves and mitigates from hitting the issue.

justbert:
Thanks for the fix!

PR #881, opened 2021-05-26
[NPM] RETURN early on MARK in DROP chains

vakalapa:
If there are multiple policies applied on a cluster with TO rules in one and PORT rules in another selecting the same namespace and pod selector, then some TO traffic is being dropped. The reason for that is in PORT chains we are adding TO/FROM chain rules and DROP rules, event though TO/FORM chain MARKs the packet and returns it, since there are no RETURNs in PORT after the TO/FROM is called, we are sending the packet to DROP chains and dropping the traffic.

This fix adds a RETURN rule at the top of both DROP chains to return, in future we should check if we can reduce the TO/FROM rules and DROP rules in PORT chain

Along with that, with this PR, we will now be only adding a single JUMP to either TO/FROM chain in PORT chain for all traffic. We will not be adding any DROP rules in PORT or TO/FROM chains ! This will reduce the number jump rules drastically, keeps a simple logic on where the MARK rules are hit and where the DROP rules are hit.


JungukCho (commented):
Can you please add a detailed explanation to describe what was an issue and how "early on Mark" resolves the issue with one example yaml? Since I only understood what code does (e.g., control order of rules (e.g., append or insert) based on type of chain), I cannot quite map what it solved. It will help me to understand this PR. Since this PR is linked with #870 , it would be nice to use the example. Will be very appreciate to show iptables changes as well.  

Also are there valid unit test for this?

JungukCho (commented):
I would like to point out exact codes, but not sure where they are.
However, after playing with this PR, in high-level, this PR removes "AZURE-NPM-EGRESS(or INGRESS)-DROPS" chains from "AZURE-NPM-EGRESS (or INGRESS)-PORT" chain to resolve packet drop issues. 

One thing that I am not sure is to put "RETURN" in each "AZURE-NPM-EGRESS (or INGRESS)-DROPS" chains is necessary.
IPtables output from example yaml which enables reproducing packet drop issues.  Italic parts seems redundant. 

Or are there something which I missed?

Chain AZURE-NPM-EGRESS (1 references)
    0     0 AZURE-NPM-EGRESS-PORT  all  --  *      *       0.0.0.0/0            0.0.0.0/0
    _**0     0 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x3000 /* RETURN-on-EGRESS-and-INGRESS-mark-0x3000 */
    0     0 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x1000 /* RETURN-on-EGRESS-mark-0x1000 */**_
    0     0 AZURE-NPM-EGRESS-DROPS  all  --  *      *       0.0.0.0/0            0.0.0.0/0

Chain AZURE-NPM-EGRESS-DROPS (1 references)
    **0     0 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x3000 /* _RETURN-on-EGRESS-and-INGRESS-mark-0x3000 */
    0     0 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x1000 /* RETURN-on-EGRESS-mark-0x1000 */**
    0     0 DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            match-set azure-npm-784554818 src /* DROP-ALL-FROM-ns-default */_
    0     0 DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            match-set azure-npm-784554818 src /* DROP-ALL-FROM-ns-default */

npm/iptm/iptm.go
```diff
@@ -56,6 +62,15 @@ type IptablesManager struct {
 	OperationFlag string
 }
 
+func isDropsChain(chainName string) bool {
+	for _, chain := range IptablesAzureDropsChainList {
```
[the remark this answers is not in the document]
vakalapa: ok let me fix it

npm/iptm/helper.go
```diff
@@ -13,10 +13,10 @@ func getAllChainsAndRules() [][]string {
 		getAzureNPMAcceptChainRules,
 		getAzureNPMIngressChainRules,
 		getAzureNPMIngressPortChainRules,
-		getAzureNPMIngressFromChainRules,
```
neaggarwMS, line 16: We still need these chains no?

vakalapa: yeah we will need the chains, this is just removing the default RETURN rules added in these chains

vakalapa:
> I would like to point out exact codes, but not sure where they are.
> However, after playing with this PR, in high-level, this PR removes "AZURE-NPM-EGRESS(or INGRESS)-DROPS" chains from "AZURE-NPM-EGRESS (or INGRESS)-PORT" chain to resolve packet drop issues.
> 
> One thing that I am not sure is to put "RETURN" in each "AZURE-NPM-EGRESS (or INGRESS)-DROPS" chains is necessary.
> IPtables output from example yaml which enables reproducing packet drop issues. Italic parts seems redundant.
> 
> Or are there something which I missed?
> 
> Chain AZURE-NPM-EGRESS (1 references)
> pkts bytes target prot opt in out source destination
> 0 0 AZURE-NPM-EGRESS-PORT all -- * * 0.0.0.0/0 0.0.0.0/0
> ___0 0 RETURN all -- * * 0.0.0.0/0 0.0.0.0/0 mark match 0x3000 /_ RETURN-on-EGRESS-and-INGRESS-mark-0x3000 _/ 0 0 RETURN all -- * * 0.0.0.0/0 0.0.0.0/0 mark match 0x1000 /_ RETURN-on-EGRESS-mark-0x1000 */_*_
> 0 0 AZURE-NPM-EGRESS-DROPS all -- * * 0.0.0.0/0 0.0.0.0/0
> 
> Chain AZURE-NPM-EGRESS-DROPS (1 references)
> pkts bytes target prot opt in out source destination
> __0 0 RETURN all -- * * 0.0.0.0/0 0.0.0.0/0 mark match 0x3000 /_ _RETURN-on-EGRESS-and-INGRESS-mark-0x3000 _/ 0 0 RETURN all -- * * 0.0.0.0/0 0.0.0.0/0 mark match 0x1000 /_ RETURN-on-EGRESS-mark-0x1000 _/__
> 0 0 DROP all -- * * 0.0.0.0/0 0.0.0.0/0 match-set azure-npm-784554818 src /* DROP-ALL-FROM-ns-default _/_ 0 0 DROP all -- * * 0.0.0.0/0 0.0.0.0/0 match-set azure-npm-784554818 src /_ DROP-ALL-FROM-ns-default */

The redundant RETURN rules in DROP chains are a form contingency for future chain ordering when we call DROP chains directly unlike today. These rules ensure any future changes will insure us from dropping relevant MARK'ed packets (whcih has caused this issue in the first place)

npm/iptm/iptm.go
```diff
@@ -78,6 +88,11 @@ func (iptMgr *IptablesManager) InitNpmChains() error {
 		metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to add AZURE-NPM chain to FORWARD chain. %s", err.Error())
 	}
 
+	// (TODO) need a review of all chains to make sure we are not adding duplicate rules.
+	// NPM should not add multiple TO/FROM rules along with DROPs in PORT chains,
```
JungukCho, line 92: Still need these comments? Is it resolved in this PR?

npm/iptm/helper.go
neaggarwMS: Shall we change the function name then? GetAllChainsAndRules is misleading

neaggarwMS:
> // getAzureNPMEgressPortChainRules returns rules for AZURE-NPM-INGRESS-PORT

nit: typo

---
Refers to: npm/iptm/helper.go:244 in c41f87a. [](commit_id = c41f87a72f66ab0e52e79a93fea507bb361c4135, deletion_comment = False)

npm/iptm/iptm.go
```diff
@@ -359,7 +374,9 @@ func (iptMgr *IptablesManager) Add(entry *IptEntry) error {
 
 	log.Logf("Adding iptables entry: %+v.", entry)
 
-	if entry.IsJumpEntry {
+	// Since there is a RETURN statement added to each DROP chain, we need to make sure
+	// any new DROP rule added to ingress or egress DROPS chain is added at the BOTTOM
+	if entry.IsJumpEntry || isDropsChain(entry.Chain) {
```
JungukCho, line 379: NPM does not need entry.IsJumpEntry anymore with this PR. 
So, please confirm in NPM code.  If we confirm this, we can remove this if condition.
Also, review and update translatePolicy_test.go with this change accordingly. If other codes also impacted, please check their UTs as well.

As I put comment about this before, RETURNs in "AZURE-NPM-EGRESS (or INGRESS)-DROPS" chains  are redundant.

If we confirm this, it seems we only need below block in iptm.go with this PR.
```go
iptMgr.OperationFlag = util.IptablesInsertionFlag.
```

Let me know if I miss something.

JungukCho (commented):
One concern in this PR is I am not 100% sure that whether it does allow invalid packets or not allow valid packets when various multiple network policies are applied in AZURE-NPM-INGRESS(or EGRESS)-DROPS chain while I think this PR does not change those behavior from previous code base and getDefaultDropEntries() guarantees that one.

Do you have any thoughts?

npm/iptm/iptm.go
vakalapa: Yes we will not need this comment as we are taking care of these duplicate rules in this PR.

npm/iptm/iptm.go
vakalapa: You are right, we do not need to append when the jumpentry is true, since we are now removing all jump entries and are relying on the default rules to help with that jump transitions to next chains.

As far as the redundant RETURN rule in DROP chains are concerned, i feel strongly to keep those rules in place as a contingency for any later changes. With these redundant rules, we are ensuring that no matter what we will NEVER DROP a MARK'ed packet and this assurance helps are from hitting any other bugs in the future.

vakalapa: > One concern in this PR is I am not 100% sure that whether it does allow invalid packets or not allow valid packets when various multiple network policies are applied in AZURE-NPM-INGRESS(or EGRESS)-DROPS chain while I think this PR does not change those behavior from previous code base and getDefaultDropEntries() guarantees that one.
> 
> Do you have any thoughts?

I think overall this is a better design because, keeping the translation logic aside, while evaluating rule we are making sure  that a MARKED packet will never be dropped and a unmarked packet will be dropped on certain conditions. This simplifies  the logic from before where we have multiple avenue to DROPs chain before we evaluated all MARK rules in a given direction (which was wrong).

JungukCho:
Thank you for applying comments! 

One more suggestion is to remove "IsJumpEntry" member from "IptEntry" struct (https://github.com/Azure/azure-container-networking/blob/master/npm/iptm/iptm.go#L50) with this PR since the "IsJumpEntry" member is not needed anymore.
Are there reasons to keep it which I missed?

Except for above suggestion, it looks good to me.

JungukCho (approved):
LGTM. 

I think this PR has functional improvement, but I assume it also has data plane latency improvement. So, while it is not urgent now, it would be good to have some measurements to see data plane improvements with this PR later.

PR #886, opened 2021-05-28
[NPM] Make each controller independent

JungukCho:
**Reason for Change**:

This PR is the first step to make each controller independent as much as possible.
Current NPM uses a global lock to avoid race condition among controllers even though they can be independent. It results in slow processing events and increases memory usages by queuing k8s resources for long time in client-go.

1. Remove a global lock in NetworkPolicyManager 
2. Make event handling functions (e.g., add, update, delete) for each resource (e.g., Pod, Namespace and NetworkPolicy) lock-free to keep processing event handler logics in each controller independently
3. Concurrently processing between event handling functions and actual reconcile function in each controller
4. Minimize a lock scope as much as possible between podController and Namespace controller
5. Cleaning up codes (e.g., encapsulation, removing redundant codes, etc)

**Issue Fixed**:

**Requirements**:

- [x] includes documentation (in npm doc directory)
- [ ] adds unit tests

**Notes**:
ipsm.go requires more works (e.g., reducing lock scope, error handling, etc) since now it is a synchronization point among all controllers. That will be next PR after completing this PR throughly.

### DIFF
--- a/npm/http/server/server.go
+++ b/npm/http/server/server.go
@@ -65,7 +65,7 @@ func NewNpmRestServer(listeningAddress string) *NPMRestServer {
 
 func (n *NPMRestServer) GetNpmMgr(npMgr *npm.NetworkPolicyManager) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// (Question?) What information does it need from npMgr?
+		// QUESTION(jungukcho) What information does it need from npMgr?
 		// Is this function required to hold lock for npMgr?
 		// npMgr.Lock()
 		err := json.NewEncoder(w).Encode(npMgr)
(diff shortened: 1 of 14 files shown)


npm/http/server/server.go
```diff
@@ -65,12 +65,14 @@ func NewNpmRestServer(listeningAddress string) *NPMRestServer {
 
 func (n *NPMRestServer) GetNpmMgr(npMgr *npm.NetworkPolicyManager) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		npMgr.Lock()
+		// QUESTION(jungukcho) What information does it need from npMgr?
+		// Is this function required to hold lock for npMgr?
+		// npMgr.Lock()
```
vakalapa, line 70: This function is used by Dataplane parser Quang uses, to get the whole internal cache for PodMap, Nsmap and netpolmap, since this is a read operation, i will let your design drive if lock is required.

But please do make sure that this function will return the 3 maps along with listmap and setmap

JungukCho:
Close this PR and open another PR #949 which is based on a recent master branch.

PR #891, opened 2021-06-04
[CNI] Add GET_STATE command to dump CNI state to stdout

matmerr:
**Reason for Change**:

Instead of CNS rehydrating from API server, CNS can use CNI to rehydrate from state stored on disk.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/api/api.go
```diff
@@ -0,0 +1,15 @@
+package api
+
+import "net"
+
+type NetworkInterfaceInfo struct {
```
matmerr, line 5: This struct is very similar to IPConfigRequest, perhaps we could use the same, but `Request` suffix on that struct would be confusing in this usecase

cni/network/invoker_cns.go
```diff
@@ -57,8 +58,15 @@ func (invoker *CNSIPAMInvoker) Add(nwCfg *cni.NetworkConfig, hostSubnetPrefix *n
...
+	endpointId := GetEndpointID(args)
+	ipconfig := cns.IPConfigRequest{
+		OrchestratorContext: orchestratorContext,
+		PodInterfaceID:      endpointId,
+		InfraContainerID:    args.ContainerID,
+	}
+
 	log.Printf("Requesting IP for pod %v", podInfo)
-	response, err := invoker.cnsClient.RequestIPAddress(orchestratorContext)
+	response, err := invoker.cnsClient.RequestIPAddress(&ipconfig)
```
matmerr, line 69: log endpoint ID here as well

cni/network/plugin/main.go
```diff
@@ -226,40 +227,58 @@ func main() {
...
+		netPlugin.Stop()
+
+		// release cni lock
+		if errUninit := netPlugin.Plugin.UninitializeKeyValueStore(false); errUninit != nil {
+			log.Errorf("Failed to uninitialize key-value store of network plugin, err:%v.\n", errUninit)
+		}
+
+		executionTimeMs := time.Since(startTime).Milliseconds()
+
+		if err != nil {
```
rbtr, line 240: i worry about this err getting shadowed - it's already not immediately clear where it comes from in the outer scope since this is deferred.

cni/network/plugin/main.go
```diff
@@ -226,40 +227,58 @@ func main() {
...
+			reportPluginError(reportManager, tb, err)
+			panic("network plugin execute fatal error")
```
rbtr, line 242: we're double panicking if the `netPlugin.Start` fails? 
it doesn't look like we're actually attempting to [recover from a panic](https://github.com/Azure/azure-container-networking/pull/891/files/364dc84771ed95dbe317573c940b33a4e9fda125#diff-1ac75250e969b454bbdfa034ee14417f209c876f78412186f62ae1a5be58bad0L207-L209) and it seems like this could all be accomplished by returning errors and exiting dirty

cni/network/plugin/main.go
```diff
@@ -226,40 +227,58 @@ func main() {
...
+		simpleState, err := netPlugin.GetSimpleState()
+		if err != nil {
+			log.Errorf("Failed to get Azure CNI state, err:%v.\n", err)
+		}
+		b, err := json.Marshal(simpleState)
+		if err != nil {
+			log.Errorf("Failed to unmarshall Azure CNI state, err:%v.\n", err)
+		}
+		log.Printf("State: %s", string(b))
+		fmt.Printf("%s\n", string(b))
```
rbtr, line 274: extra log?

cni/client/client_integration_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"os/user"
+	"testing"
+
+	"github.com/Azure/azure-container-networking/cni/api"
+	testutils "github.com/Azure/azure-container-networking/test/utils"
+	"github.com/stretchr/testify/require"
+	"k8s.io/utils/exec"
+)
+
+func isRoot() bool {
```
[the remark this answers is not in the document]
rbtr: this is duped in the testing utils below

rbtr (changes requested):
confusing that the lint step didn't catch the error casing, because it does locally for me

cni/api/api.go
neaggarwMS: It is okay to use separate one, this contract is owned by CNI and later if we want to extend it then we dont need to touch other one

neaggarwMS, line 5: Can we call it PodNetworkInterfaceInfo or ContainerNetworkInterfaceInfo?

cni/network/invoker_azure.go
```diff
@@ -50,7 +51,9 @@ func (invoker *AzureIPAMInvoker) Add(nwCfg *cni.NetworkConfig, subnetPrefix *net
 	defer func() {
 		if err != nil {
 			if len(result.IPs) > 0 {
-				invoker.plugin.ipamInvoker.Delete(&result.IPs[0].Address, nwCfg, nil, options)
+				if er := invoker.plugin.ipamInvoker.Delete(&result.IPs[0].Address, nwCfg, nil, options); er != nil {
```
neaggarwMS, line 54: Why change in this file?

cni/network/invoker_cns.go
```diff
@@ -77,7 +85,7 @@ func (invoker *CNSIPAMInvoker) Add(nwCfg *cni.NetworkConfig, hostSubnetPrefix *n
 	// set the NC Primary IP in options
 	options[network.SNATIPKey] = info.ncPrimaryIP
 
-	log.Printf("[cni-invoker-cns] Received info %v for pod %v", info, podInfo)
+	log.Printf("[cni-invoker-cns] Received info %+v	v for pod %v", info, podInfo)
```
neaggarwMS, line 88: >v [](http://example.com/codeflow?start=49&length=1)

typo, remove extra v

cni/network/invoker_cns.go
```diff
@@ -57,8 +58,15 @@ func (invoker *CNSIPAMInvoker) Add(nwCfg *cni.NetworkConfig, hostSubnetPrefix *n
...
 	}
 
+	endpointId := GetEndpointID(args)
+	ipconfig := cns.IPConfigRequest{
+		OrchestratorContext: orchestratorContext,
+		PodInterfaceID:      endpointId,
+		InfraContainerID:    args.ContainerID,
+	}
+
 	log.Printf("Requesting IP for pod %v", podInfo)
```
neaggarwMS, line 68: Log the whole ipConfig struct

cni/network/plugin/main.go
```diff
@@ -226,40 +227,58 @@ func main() {
...
+			}
+		}
+	}()
+
 	// Check CNI_COMMAND value for logging purposes.
 	cniCmd := os.Getenv(cni.Cmd)
 	log.Printf("CNI_COMMAND environment variable set to %s", cniCmd)
 
+	// used to dump state
+	if cniCmd == cni.CmdState {
```
neaggarwMS, line 263: Where are we acquiring the file lock?

cns/cnsclient/cnsclient.go
```diff
@@ -211,7 +211,7 @@ func (cnsClient *CNSClient) DeleteHostNCApipaEndpoint(networkContainerID string)
 }
 
 // RequestIPAddress calls the requestIPAddress in CNS
-func (cnsClient *CNSClient) RequestIPAddress(orchestratorContext []byte) (*cns.IPConfigResponse, error) {
+func (cnsClient *CNSClient) RequestIPAddress(ipconfig *cns.IPConfigRequest) (*cns.IPConfigResponse, error) {
```
neaggarwMS, line 214: >RequestIPAddress [](http://example.com/codeflow?start=28&length=16)

Is this used for a debug API?

cni/network/plugin/main.go
matmerr: https://github.com/Azure/azure-container-networking/blob/a01039a3125cf2e849228c01136b2e74e15d72a2/network/manager.go#L384

lock is taken here, should be changed to RLock

cni/network/invoker_azure.go
rbtr: `invoker.plugin.ipamInvoker.Delete` can return an error and we were previously ignoring it completely
because parts of this file were changed, golangci-lint flagged this problem.

cni/network/plugin/main.go
neaggarwMS: For now, it is okay for now as we dont want any other CNI instance to update this state while reconcile is going on

cni/network/plugin/main.go
matmerr: the first line logs to the file in `/var/log/azure-vnet.log`, the second is the result to be read by the caller capturing stdout

cni/network/plugin/main.go
```diff
@@ -226,40 +227,58 @@ func main() {
 		panic("network plugin start fatal error")
 	}
 
+	defer func() {
+		netPlugin.Stop()
+
+		// release cni lock
+		if errUninit := netPlugin.Plugin.UninitializeKeyValueStore(false); errUninit != nil {
```
matmerr, line 234: dupe uninitialize keyvalue store in defer above

cns/cnsclient/cnsclient.go
matmerr: This is used in the CNI call to CNS for ipam

cni/api/api.go
```diff
@@ -0,0 +1,37 @@
...
+	"net"
+	"os"
+
+	"github.com/Azure/azure-container-networking/log"
+)
+
+type PodNetworkInterfaceInfo struct {
+	PodName        string
+	PodNamespace   string
+	PodInterfaceID string
```
tamilmani1989, line 14: can it be PodEndpointId?

cni/network/plugin/main.go
```diff
@@ -204,6 +206,10 @@ func main() {
 			log.Errorf("Failed to uninitialize key-value store of network plugin, err:%v.\n", errUninit)
 		}
 
+		if err != nil {
```
tamilmani1989, line 209: why do we need this?

cni/network/plugin/main.go
```diff
@@ -170,6 +171,7 @@ func main() {
 	upTime, err := platform.GetLastRebootTime()
 	if err == nil {
 		cniReport.VMUptime = upTime.Format("2006-01-02 15:04:05")
+		return
```
tamilmani1989, line 174: why we return here. cni shouldnt exit here if it fails to get last reboot time. recently we removed cni dependency on reboot time.

tamilmani1989 (changes requested):

cni/network/invoker_azure.go
matmerr: this is using a tightly scoped variable `er` and wrapping the original error with `%w` in the return so it's not lost

cni/network/invoker_cns.go
```diff
@@ -57,8 +58,15 @@ func (invoker *CNSIPAMInvoker) Add(nwCfg *cni.NetworkConfig, hostSubnetPrefix *n
 		return nil, nil, err
 	}
 
-	log.Printf("Requesting IP for pod %v", podInfo)
-	response, err := invoker.cnsClient.RequestIPAddress(orchestratorContext)
+	endpointId := GetEndpointID(args)
+	ipconfig := cns.IPConfigRequest{
+		OrchestratorContext: orchestratorContext,
```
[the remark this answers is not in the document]
matmerr: this is for debuggability and clarity from CNS

cni/network/invoker_cns.go
matmerr: updated log line on 68 to print both podInfo and ipconfig, which contains this information

cni/api/api.go
```diff
@@ -0,0 +1,41 @@
...
+	"net"
+	"os"
+
+	"github.com/Azure/azure-container-networking/log"
+)
+
+type PodNetworkInterfaceInfo struct {
+	PodName        string
+	PodNamespace   string
+	PodEndpointId string
```
neaggarwMS, line 14: Do we need to update the contract also to call it PodEndpointId?

neaggarwMS (approved):
:shipit:

PR #897, opened 2021-06-11
Avoid creating timer each time SyncHostVersion is called, instead use…

neaggarwMS:
… timer channel

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -567,9 +567,12 @@ func main() {
 		}
 		go func(ep, vnet, node string) {
 			// Periodically poll DNC for node updates
+			tickerChannel := time.Tick(time.Duration(cnsconfig.ManagedSettings.NodeSyncIntervalInSeconds) * time.Second)
 			for {
-				<-time.NewTicker(time.Duration(cnsconfig.ManagedSettings.NodeSyncIntervalInSeconds) * time.Second).C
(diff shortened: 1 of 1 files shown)


cns/service/main.go
```diff
@@ -743,9 +743,12 @@ func InitializeMultiTenantController(httpRestService cns.HTTPService, cnsconfig
 	rootCxt := context.Background()
 	go func() {
 		// Periodically poll vfp programmed NC version from NMAgent
+		timerChannel := time.NewTicker(cnsconfig.SyncHostNCVersionIntervalMs * time.Millisecond).C
```
rbtr, line 746: usually we would not pull the channel out of the ticker, we would use a ticker var:
```
ticker := time.NewTicker(t)
defer ticker.Stop()
for {
        select {
        case <-ctx.Done():
            return
        case <- ticker.C:
            // work
}
```
although, since we have no need to ever stop the ticker here, we can abbreviate this using `time.Tick`:
```suggestion
		tick:= time.Tick(cnsconfig.SyncHostNCVersionIntervalMs * time.Millisecond)
```

rbtr (changes requested):

rbtr (approved):

PR #900, opened 2021-06-14
chore: migrate from exit/err channels to context

rbtr:

**Reason for Change**:

Migrates from a variety of exit and err channels to a runtime context. In this PR, the root context is passed directly to the child contexts (since the previous exit channels where never closed independently, only when the parent scope was exiting), but this could serve as a basis to derive cancellable or timeout child contexts in the future.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/service/main.go
```diff
@@ -246,6 +246,27 @@ var args = acn.ArgumentList{
 	},
 }
 
+func init() {
```
thatmattlong, line 249: why `init()` function instead of main? I tend to avoid init when possible since it can cause some confusion, but if there is a compelling reason, I'm not necessarily against it

thatmattlong (changes requested):

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -17,18 +17,16 @@ const (
 )
 
 type CNSIPAMPoolMonitor struct {
-	pendingRelease bool
-
+	MaximumFreeIps           int64
+	MinimumFreeIps           int64
 	cachedNNC                nnc.NodeNetworkConfig
-	updatingIpsNotInUseCount int
+	ctx                      context.Context
```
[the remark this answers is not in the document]
rbtr: agree - i didn't read the ipampoolmonitor closely enough and thought i shouldn't change the signature on the `Reconcile()` because it was being called externally...but fixed it now
i put contexts in structs when that object needs to be able to spawn child contexts and manage them without injecting its parent's context

cns/service/main.go
rbtr: doing this type of work in an init is more canonical than main.
initializing globals that need to exist before any code in the package runs is the stated intent of init functions. this stuff could be at the top of main, but that does less to communicate that it _must_ be done before anything in the package runs, and there could be other entry points in to code in this package than the main func (obviously there _aren't_, but theoretically).

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -144,40 +147,38 @@ func NewCrdRequestController(restService *restserver.HTTPRestService, kubeconfig
 }
 
 // InitRequestController will initialize/reconcile the CNS state
-func (crdRC *crdRequestController) InitRequestController() error {
+func (rc *requestController) InitRequestController(ctx context.Context) error {
```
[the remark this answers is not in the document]
rbtr: as long as it's [short, clear, and consistent](https://github.com/golang/go/wiki/CodeReviewComments#receiver-names), anything is acceptable. imo "rc" is as short as we can abbreviate "request controller" without losing information 😉

cns/service/main.go
neaggarwMS: Where is this called? Is it called before main()? I agree with Matt, it gets confusing (at least for the readers coming from non-go world). We havent used this pattern in other services. May be a detailed comment in both init and main function will future developers to follow this standard

cns/service/main.go
```diff
@@ -847,20 +834,17 @@ func InitializeCRDState(httpRestService cns.HTTPService, cnsconfig configuration
...
-				httpRestServiceImplementation.SyncHostNCVersion(rootCxt, cnsconfig.ChannelMode, cnsconfig.SyncHostNCTimeoutMs)
-            case <-rootCxt.Done():
-                return
+				httpRestServiceImplementation.SyncHostNCVersion(rootCtx, cnsconfig.ChannelMode, cnsconfig.SyncHostNCTimeoutMs)
+			case <-rootCtx.Done():
+				return
 			}
 		}
-
-		logger.Printf("[Azure CNS] Exiting SyncHostNCVersion")
```
neaggarwMS, line 863: Why did you remvoe this log line? (same comment on another PR)

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -32,10 +33,12 @@ const (
 	prometheusAddress = "0" //0 means disabled
 )
 
-// crdRequestController
+var _ requestcontroller.RequestController = (*requestController)(nil)
+
+// requestController
 // - watches CRD status changes
 // - updates CRD spec
-type crdRequestController struct {
```
neaggarwMS, line 38: crdRequestController will make it specific to single tenant crd scenarios...for multi tenant crds, we are using multitenantCrdRequestController

neaggarwMS (approved):
:shipit:

neaggarwMS (changes requested):
:clock1:

cns/service/main.go
rbtr: `init()`s are called automatically when the package is imported - before any code runs in them runs. they are used for things like building pre-run state that can't be done in package level declarations. for example, here, I want to set up the signal handling before any code runs so that the program can exit gracefully even before `main()` is invoked.
more info: https://golang.org/doc/effective_go#init
spec: https://golang.org/ref/spec#Package_initialization

cns/service/main.go
rbtr: it is unreachable code

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
rbtr: since the multitenant request controller is in a different package, it's unambiguous and having the name be simpler is better - the package is part of the name when it is imported and used so no information is lost

cns/service/main.go
thatmattlong: Looks like this is the intended usage of `init()` according to the official style guide, so I'm happy with it this way (in fact I prefer to follow the style guide, so wouldn't be happy any other way)

thatmattlong (approved):

cns/requestcontroller/kubecontroller/crdrequestcontroller.go
neaggarwMS: Agreed with Matt above, if we are fixing it, then lets fix it completely. Either we rename this file to SingleTenantRequestController and similarly the structure or what Matt suggested above

cns/service/main.go
neaggarwMS: Sounds good, Evan can we add a function comment to Init() and probably link the go guideline

neaggarwMS (changes requested):
:clock1:

neaggarwMS (approved):
:shipit:

PR #904, opened 2021-06-15
[NPM] Add a protocol flag to namedport 

vakalapa:
Today NPM does not check protocol flag for a given named port. Take below example:

```yaml
    - port: serve-81-udp
      protocol: TCP
```
-------------
Below is the named port yaml snippet from the pod:
```yaml
    ports:
    - containerPort: 81
      name: serve-81-udp
      protocol: UDP
```

In this above example, NPM should not send either UDP traffic or TCP traffic on port 81, since the namedport is UDP and the protocol mentioned is TCP, as per "AND" property it should block all traffic.

but due to a bug, we allow UDP traffic mentioned in the named port. This change makes sure to add a protocol flag mentioned with the named port to make sure that the correct protocol is allowed.


JungukCho (dismissed):
LGTM! Within next few PRs, I hope we have chance to refactor translatePolicy.go since there are a lot of repeat codes.

JungukCho (approved):

PR #907, opened 2021-06-23
[NPM] Dataplane Parser + Dataplane Converter + Traffic Analyzer

nddq:
This pull request contains:
- Dataplane Parser: Parses data from iptables-save into a Go object.
- Dataplane Converter: Parses rules within iptables-save's Go object to a list of JSON or Protobuf objects.
- Traffic Analyzer: Take in a source and a destination and evaluate whether network traffic is allowed between them based on the rules from the converter. Returns a list of hit rule tuples.

This PR will be test on real iptables and ipset in the next CLI PR. Functionalities confirmed with file option.


hack/dataplaneConverter/converter.go
```diff
@@ -0,0 +1,140 @@
...
+	SPort         string            `json:"sport"`
+	Allowed       bool              `json:"allowed"`
+	Direction     string            `json:"direction"`
+	UnsortedIpset map[string]string `json:"unsortedIpset"` //key: ipset name, value: src,dst or dst,dst
+}
+
+type Direction int
+
+func (d Direction) String() string {
+	return [...]string{"EGRESS", "INGRESS"}[d]
```
vakalapa, line 32: this is a little loosely coupled, any change in this order by other contributors will break all subsequent packages using this direction information.

We might not want to use integer as static but use the string it self as static, that way we can control the string alloc and not worry about the ordering of int to string

vakalapa: Also move this to iptables.go

hack/dataplaneParser/iptable/iptable.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type Direction string
+
+func grabIptablesFileLock(f *os.File) error {
+	return unix.Flock(int(f.Fd()), unix.LOCK_EX|unix.LOCK_NB)
+}
+
+// grabs iptables v1.6 xtable lock
+func GrabIptablesLocks() (*os.File, error) {
```
vakalapa, line 46: Do we need to re-write this function here ? I think we already have a grabiptableslock funciton in pkg iptm, we should probably use that.

hack/dataplaneParser/iptable/iptable.go
```diff
@@ -0,0 +1,114 @@
...
+	}); err != nil {
+		// metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to acquire new iptables lock: %v.", err)
+		return nil, err
+	}
+
+	success = true
+	return l, nil
+}
+
+// func (t *Iptables) PrintIptable() {
```
vakalapa, line 79: Lets put these functions back, these are good for debugging.

hack/dataplaneParser/iptable/iptable.go
```diff
@@ -0,0 +1,114 @@
...
+	"os"
+	"time"
+
+	"github.com/Azure/azure-container-networking/npm/util"
+	"golang.org/x/sys/unix"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+type Iptables struct {
+	Name   string
```
vakalapa, line 13: lets make all the variables inside start with lowercase, we want the variables to be references through helper functions on top of the defined structs.

hack/dataplaneParser/iptable/iptable.go
```diff
@@ -0,0 +1,114 @@
...
+	Rules []*IptablesRule
+}
+
+type IptablesRule struct {
+	Protocol string
+	Target   *Target
+	Modules  []*Module
+}
+
+type Module struct {
```
vakalapa, line 29: module and target will not be referred alone from outside, so lets make them start with lower case too, that way we can be sure that they will be always be used through the IptableRule

npm/debugTools/networkTupleProcessor/tupleProcessor_test.go
```diff
@@ -0,0 +1,80 @@
...
+		sortedExpectedTupleList := hashTheSortTupleList(test.expected)
+		_, actualTupleList := p.GetNetworkTuple(test.input.src, test.input.dst, "../testFiles/npmCache.json", "../testFiles/clusterIptableSave")
+		sortedActualTupleList := hashTheSortTupleList(actualTupleList)
+		if !reflect.DeepEqual(sortedExpectedTupleList, sortedActualTupleList) {
+			t.Errorf("expected '%+v', got '%+v'", sortedExpectedTupleList, sortedActualTupleList)
+		}
+	}
+}
+
+func TestEvaluateSetInfo(t *testing.T) {
```
vakalapa, line 78: Can we fill this test, if not used lets remove this.

npm/debugTools/networkTupleProcessor/tupleProcessor.go
```diff
@@ -0,0 +1,367 @@
...
+
+const (
+	IPADDRS  InputType = 0
+	PODNAME  InputType = 1
+	INTERNET InputType = 2
+)
+
+// GetNetworkTuple returns a list of hit rules between the source and the destination in JSON format and a list of tuples from those rules. Filenames following the format, cacheFile first and then iptable-save file
+// optional for debugging
+func (p *Processor) GetNetworkTuple(src, dst *Input, filenames ...string) ([][]byte, []*Tuple) {
```
vakalapa, line 47: Lets expand filenames to take the exact input, say cacheFile and iptablesOutFile. then that gives you more combinations of either to run only one of the file or both

npm/debugTools/networkTupleProcessor/tupleProcessor.go
```diff
@@ -0,0 +1,367 @@
...
+		cmd := exec.Command(util.IptablesSave, cmdArgs...)
+
+		cmd.Stdout = iptableBuffer
+		stderrBuffer := bytes.NewBuffer(nil)
+		cmd.Stderr = stderrBuffer
+
+		err := cmd.Run()
+
+		if err != nil {
+			stderrBuffer.WriteTo(iptableBuffer) // ignore error, since we need to return the original error
```
vakalapa, line 84: lets not ignore this error, at least for lint's sake, catch the error and print it, as a warning.

npm/debugTools/networkTupleProcessor/tupleProcessor.go
```diff
@@ -0,0 +1,367 @@
...
+		for _, b := range byteArray {
+			iptableBuffer.WriteByte(b)
+		}
+		allRules = c.GetProtobufRulesFromIptable(tableName, iptableBuffer, filenames[0])
+
+	} else {
+		cacheObj = c.GetNpmCache()
+		// need to add iptable locks
+		cmdArgs := []string{"-t", string(tableName)}
+		cmd := exec.Command(util.IptablesSave, cmdArgs...)
```
vakalapa, line 75: lets move this logic to its own function, better yet, reuse our existing code functions.

npm/debugTools/dataplaneConverter/converter.go
```diff
@@ -0,0 +1,297 @@
...
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+type Converter struct {
+	ListMap           map[string]string
+	SetMap            map[string]string
+	RequiredChainsMap map[string]bool
+}
+
+func (c *Converter) GetNpmCache(filename ...string) *NPMCache {
```
vakalapa, line 25: This is a crucial function, if this fails we have nothing right, so lets return error and continue based on this error. You are doing that i am sure, but bubbling up error is always a good idea, so tomorrow if we want to have a mitigation in place in the caller function we can add something

npm/debugTools/dataplaneConverter/converter.go
```diff
@@ -0,0 +1,297 @@
...
+	cachObj := &NPMCache{}
+
+	if len(filename) > 0 {
```
vakalapa, line 28: Same lets expand on this filename

npm/debugTools/networkTupleProcessor/tupleProcessor.go
```diff
@@ -0,0 +1,367 @@
...
+		allRules      []*pb.RuleResponse
+		srcPod        *npm.NpmPod
+		dstPod        *npm.NpmPod
+		iptableBuffer = bytes.NewBuffer(nil)
+		tableName     = "filter"
+	)
+
+	// hacky way to make it works for testing
+	if len(filenames) > 1 {
+		cacheObj = c.GetNpmCache(filenames[0])
```
vakalapa, line 60: it is very easy for us to rearrange that order in the caller function, so lets have specific filenames

vakalapa: This GetNPmCahce might a duplicate  call, since below in GetProtobufRulesFromIptable -> inti container also we are getting it

npm/debugTools/networkTupleProcessor/tupleProcessor.go
```diff
@@ -0,0 +1,367 @@
...
+		}
+		allRules = c.GetProtobufRulesFromIptable(tableName, iptableBuffer)
+
+	}
+	switch src.Type {
+	case PODNAME:
+		srcPod = cacheObj.PodMap[src.Content]
+	case IPADDRS:
+		for _, pod := range cacheObj.PodMap {
+			if pod.PodIP == src.Content {
```
vakalapa, line 94: Can we add some sanities here of the src.Content is nil or even an IP ?

npm/debugTools/networkTupleProcessor/tupleProcessor.go
```diff
@@ -0,0 +1,367 @@
...
+		srcPod = &npm.NpmPod{}
+	default:
+		panic("Invalid source type")
+	}
+	switch dst.Type {
+	case PODNAME:
+		dstPod = cacheObj.PodMap[dst.Content]
+	case IPADDRS:
+		for _, pod := range cacheObj.PodMap {
+			if pod.PodIP == dst.Content {
```
vakalapa, line 112: same here

npm/debugTools/networkTupleProcessor/tupleProcessor.go
nddq: This should be resolve in the CLI PR, where I did the input type check before this step

JungukCho (commented):
This is great progress. It would be nice to have better naming for variables, remove redundancy, and split one function to multiples if it helps overall. More unit tests with even simple testset is also helpful.

I did not walk through evaluation part in tupleProcess.go yet.  After you revise your PR, I would like to walk through it again. Please let me know.

npm/debugTools/dataplaneParser/parser/parser.go
```diff
@@ -0,0 +1,339 @@
...
+			rules := val.Rules()
+			rules = append(rules, p.parseRuleFromLine(line[ruleStartIndex:]))
+			val.SetRules(rules)
+		}
+	}
+	return chainMap
+}
+
+// parseLine parse each line of the read-in byteArray of iptables-save
+func (p *Parser) parseLine(readIndex int, byteArray []byte) ([]byte, int) {
```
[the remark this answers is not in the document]
rbtr: this method is complex, it would benefit from being generously commented (explain the why of all of these loops and checks).
to @JungukCho's point, at a first read, it feels like you could pre-condition the input and then `bufio.ReadLine()`

npm/debugTools/dataplane/iptable.go
```diff
@@ -0,0 +1,211 @@
...
+	t.name = name
+}
+
+func (t *Iptables) Chains() map[string]*IptablesChain {
+	return t.chains
+}
+
+func (t *Iptables) SetChains(chain_m map[string]*IptablesChain) {
+	t.chains = chain_m
+}
```
rbtr, line 35: [I left this feedback on your other PR](https://github.com/Azure/azure-container-networking/pull/951#discussion_r675148161); don't do this Java style constructor/getter/setter fluff in Go. If the struct fields need to be directly accessed outside of the package, make them public and initialize the struct directly.

rbtr (changes requested):
no Java only Go pls

npm/debugTools/dataplane/iptable.go
vakalapa: this bit of code was my doing ! Thanks for the suggestion.

npm/debugTools/networkTupleProcessor/tupleProcessor.go
```diff
@@ -0,0 +1,369 @@
...
+	Content string
+	Type    InputType
+}
+
+type InputType int32
+
+const (
+	IPADDRS  InputType = 0
+	PODNAME  InputType = 1
+	INTERNET InputType = 2
```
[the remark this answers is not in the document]
JungukCho: In Kubernetes or network policy, is it called normally External (ExternalTraffic)?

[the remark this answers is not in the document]
nddq: If that's the case, I could change this to that.

vakalapa:
Removing the work-in-progress flag as PR is ready to be review again, @rbtr  @JungukCho

npm/debugTools/dataplane/parser.go
```diff
@@ -0,0 +1,338 @@
...
+)
+
+// Parser struct
+type Parser struct{}
+
+// ParseIptablesObject create a Go object from specified iptable by calling iptables-save within node
+func (p *Parser) ParseIptablesObject(tableName string) *Iptables {
+	iptableBuffer := bytes.NewBuffer(nil)
+	// TODO: need to get iptable's lock
+	cmdArgs := []string{"-t", string(tableName)}
```
vakalapa, line 19: Good point lets add the locking mechanism for iptables here, 

```suggestion
	cmdArgs := []string{util.IptablesWaitFlag, entry.LockWaitTimeInSeconds, iptMgr.OperationFlag, entry.Chain,"-t", string(tableName)}
```

npm/debugTools/dataplane/parser.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	for _, b := range byteArray {
+		iptableBuffer.WriteByte(b)
+	}
+	chains := p.parseIptablesChainObject(tableName, iptableBuffer.Bytes())
+	return &Iptables{Name: tableName, Chains: chains}
+}
+
+// parseIptablesChainObject create a map of iptable chain name and iptable chain object
+func (p *Parser) parseIptablesChainObject(tableName string, iptableBuffer []byte) map[string]*IptablesChain {
```
vakalapa, line 52: Can you document the limitations we have with this parsing logic, or document what it can do today.

npm/debugTools/dataplane/converter.go
```diff
@@ -0,0 +1,403 @@
...
+	return nil
+}
+
+// NpmCache initialize NPM cache from node
+func (c *Converter) NpmCache() error {
+	c.NPMCache = &NPMCache{}
+	req, err := http.NewRequestWithContext(
+		context.Background(),
+		http.MethodGet,
+		"http://localhost:10091/npm/v1/debug/manager",
```
vakalapa, line 63: Instead of hard coding this, can you take this value from https://github.com/Azure/azure-container-networking/blob/a68b7fd5318d65378a46f4d2bc83d5492ef6b763/npm/http/api/api.go#L8

npm/debugTools/dataplane/parser.go
```diff
@@ -0,0 +1,338 @@
...
+			lastNonWhiteSpaceIndex = curReadIndex // update index of last non-white space character
+			rightLineIndex = -1                   // reset right index of line
+		}
+	}
+	// line with right spaces (unlikely to encounter)
+	return iptableBuffer[leftLineIndex : lastNonWhiteSpaceIndex+1], curReadIndex
+}
+
+// parseChainNameFromRuleLine gets the chain name from given rule line
+func (p *Parser) parseChainNameFromRuleLine(line []byte) (string, int) {
```
JungukCho, line 142: line -> ruleLine for consistency

npm/debugTools/dataplane/parser.go
```diff
@@ -0,0 +1,338 @@
...
+	spaceIndex1 := bytes.Index(line, util.SpaceBytes)
```
JungukCho, line 143: It seems ok to use only spaceIndex instead of using both spaceIndex1 and spaceIndex2.
Also it is ok to use only start and end. 
Slightly better to use chainStart and chainEnd?

npm/debugTools/dataplane/parser.go
```diff
@@ -0,0 +1,338 @@
...
+			if !ok {
+				val = &IptablesChain{chainName, []byte{}, make([]*IptablesRule, 0)}
+			}
+			val.Rules = append(val.Rules, p.parseRuleFromLine(line[ruleStartIndex:]))
+		}
+	}
+	return chainMap
+}
+
+// parseLine parse the line starting from the given readIndex of the iptableBuffer.
```
JungukCho, line 100: typo : parse -> parses

npm/debugTools/dataplane/parser.go
```diff
@@ -0,0 +1,338 @@
...
+	module.Verb = verb
+	return p.parseModuleOptionAndValue(nextIndex+spaceIndex+1, module, "", ruleLine, true)
+}
+
+func (p *Parser) parseModuleOptionAndValue(
+	nextIndex int,
+	module *Module,
+	curOption string,
+	ruleLine []byte,
+	included bool,
```
JungukCho, line 287: Curious: Is it necessary comma(,)?

npm/debugTools/dataplane/parser.go
```diff
@@ -0,0 +1,338 @@
...
+		default:
+			lastNonWhiteSpaceIndex = curReadIndex // update index of last non-white space character
+			rightLineIndex = -1                   // reset right index of line
+		}
+	}
+	// line with right spaces (unlikely to encounter)
+	return iptableBuffer[leftLineIndex : lastNonWhiteSpaceIndex+1], curReadIndex
+}
+
+// parseChainNameFromRuleLine gets the chain name from given rule line
```
JungukCho, line 141: Can we have UT for this function?

npm/debugTools/dataplane/parser_test.go
```diff
@@ -0,0 +1,208 @@
...
+	"bytes"
+	"reflect"
+	"testing"
+
+	"github.com/Azure/azure-container-networking/npm/util"
+)
+
+func TestParseIptablesObjectFile(t *testing.T) {
+	p := &Parser{}
+	p.ParseIptablesObjectFile(util.IptablesFilterTable, "../testFiles/iptableSave")
```
JungukCho, line 13: Any comparable values after running function?

npm/debugTools/dataplane/parser_test.go
```diff
@@ -0,0 +1,208 @@
...
+			input:    testL4,
+			expected: expectByteArray,
+		},
+	}
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.input, func(t *testing.T) {
+			actualLine, _ := p.parseLine(0, []byte(tc.input))
+			if equal := bytes.Compare(expectByteArray, actualLine); equal != 0 {
+				t.Errorf("expected '%+v', got '%+v'", tc.expected, actualLine)
```
JungukCho, line 63: nitpick: it seems idiomatic of go is to print failed message with the order of "got" and "expected".
Refer to https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

npm/debugTools/dataplane/converter.go
```diff
@@ -0,0 +1,403 @@
...
+	if strings.HasPrefix(name, util.NamespacePrefix) {
+		return pb.SetType_NAMESPACE
+	}
+	if strings.HasPrefix(name, util.NamedPortIPSetPrefix) {
+		return pb.SetType_NAMEDPORTS
+	}
+	if strings.Contains(name, util.IpsetLabelDelimter) {
+		return pb.SetType_KEYVALUELABELOFPOD
+	}
+	if matched, _ := regexp.MatchString(`(?i)[^ ]+-in-ns-[^ ]+-\d(out|in)\b`, name); matched {
```
rbtr, line 279: it might be more efficient to pre-allocate this static matcher at the package scope with ex: `var matcher = regexp.MustCompile(...)` but I haven't confirmed this

npm/debugTools/dataplane/parser.go
rbtr: yes, go needs trailing commas

npm/debugTools/dataplane/parser.go
```diff
@@ -0,0 +1,338 @@
...
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os/exec"
+
+	"github.com/Azure/azure-container-networking/npm/util"
+)
+
+// Parser struct
+type Parser struct{}
```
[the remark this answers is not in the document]
nddq: At first, my thoughts were making all parsing functions can only be call from a parser struct would be more organized and easier to understand but now I think you and Evan are right, I will make the changes.

npm/debugTools/dataplane/parser.go
nddq: I tried running the -w flag with the iptables-save command but it returns an invalid option error. Is there a function similar to GrabIptableLock I saw sometimes ago?

npm/debugTools/dataplane/converter.go
```diff
@@ -0,0 +1,403 @@
...
+	if strings.Contains(iptableChainObjName, "EGRESS") {
+		return pb.Direction_EGRESS
+	} else if strings.Contains(iptableChainObjName, "INGRESS") {
+		return pb.Direction_INGRESS
+	}
+	return pb.Direction_UNDEFINED
+}
+
+func (c *Converter) getSetType(name string, m string) pb.SetType {
+	// TODO: Handle CIDR blocks
```
JungukCho, line 260: If it seems now it support IPblcok. If so, please delete todo.

npm/debugTools/dataplane/parser.go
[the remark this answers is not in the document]
JungukCho: After reading some more codes, it seems there is advantage to use parser struct. 
It explicitly lets us know it is related to parsing logics while the methods already have "parse-" prefix.

I am not sure which one (with or without parser struct) is better  while dropping parser struct is more natural.
It would be nice to drop all parse- prefix in all methods in parser.go if we stick to use parser struct.

What do you think?

npm/debugTools/dataplane/parser.go
vakalapa: https://github.com/Azure/azure-container-networking/blob/3bd237cbf11514c60656f82d21581a3cd47e6c5e/npm/iptm/iptm_linux.go#L10

npm/debugTools/dataplane/parser.go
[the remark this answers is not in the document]
rbtr: drop the struct, put in own `parse` package. then the usage will be `parse.IptablesObject`, pure functions without needing to hold an object ref for access

npm/debugTools/dataplane/converter.go
```diff
@@ -0,0 +1,403 @@
...
+		ruleRes.DstList = append(ruleRes.DstList, infoObj)
+	}
+	return nil
+}
+
+// populate CIDRBlock set's content with ip addresses
+func populateCIDRBlockSet(setInfoObj *pb.RuleResponse_SetInfo) {
+	ipsetBuffer := bytes.NewBuffer(nil)
+	cmdArgs := []string{"list", setInfoObj.HashedSetName}
+	cmd := exec.Command(util.Ipset, cmdArgs...) //nolint:gosec
```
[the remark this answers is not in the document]
nddq: https://stackoverflow.com/questions/58860612/go-exec-command-security-linting-messages-using-golangci

It's says here that the linter wants hardcoded command call, so this comment is used to bypass it.

npm/debugTools/dataplane/parser.go
nddq: Now that I think about it, do we really need to get the iptables' lock here? Since I'm only doing a read operation and not modify anything.

npm/debugTools/dataplane/tupleProcessor.go
```diff
@@ -0,0 +1,431 @@
...
+			matchedSource, err := evaluateSetInfo("src", setInfo, src, rule, cacheObj)
+			if err != nil {
+				return nil, fmt.Errorf("error occurred during evaluating source's set info : %w", err)
+			}
+			if !matchedSource {
+				matched = false
+				break
+			}
+		}
+		for _, setInfo := range rule.DstList {
```
JungukCho, line 231: Can we skip this iteration if matched is false after the previous for loop?

npm/debugTools/dataplane/tupleProcessor.go
```diff
@@ -0,0 +1,431 @@
...
+		return nil, nil, fmt.Errorf("error occurred during get network tuple : %w", err)
+	}
+
+	return getNetworkTupleCommon(src, dst, c.NPMCache, allRules)
+}
+
+// Common function.
+func getNetworkTupleCommon(
+	src, dst *Input,
+	npmCache *NPMCache,
```
JungukCho, line 78: seems good to use npmCache instead cacheObj in all functions.

npm/debugTools/dataplane/tupleProcessor.go
```diff
@@ -0,0 +1,431 @@
...
+	// 	ruleJson, err := json.MarshalIndent(rule, "", "  ")
+	// 	if err != nil {
+	// 		log.Fatalf("Error occurred during marshaling. Error: %s", err.Error())
+	// 	}
+	// 	tupleResListJson = append(tupleResListJson, ruleJson)
+	// }
+	return ruleResListJSON, resTupleList, nil
+}
+
+func getCorrespondPod(origin *Input, cacheObj *NPMCache) (*npm.NpmPod, error) {
```
JungukCho, line 131: suggestion: getNPMPod instead of getCorrespondPod

npm/debugTools/dataplane/tupleProcessor.go
```diff
@@ -0,0 +1,431 @@
...
+	switch origin.Type {
+	case PODNAME:
+		return cacheObj.PodMap[origin.Content], nil
+	case IPADDRS:
+		for _, pod := range cacheObj.PodMap {
```
JungukCho, line 136: Possible optimization: Good to pre-process to maintain map like map[ip]PodName. 
This operation is heavy to iterate it everytime.

npm/debugTools/dataplane/tupleProcessor.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+	if rule.Protocol != "" {
+		tuple.Protocol = rule.Protocol
+	} else {
+		tuple.Protocol = ANY
+	}
+	return tuple
+}
+
+func getHitRules(
```
JungukCho, line 206: This function will become a bottleneck if there are a lot of rules. 
It seems that it needs some optimizations with other data structures.
While we do not implement it in this PR, any cool suggestions (may pre-process while building rules)?

npm/debugTools/dataplane/tupleProcessor.go
```diff
@@ -0,0 +1,431 @@
...
+		}
+		if matched {
+			res = append(res, rule)
+		}
+	}
+	return res, nil
+}
+
+// evalute an ipset to find out whether the pod's attributes match with the set
+func evaluateSetInfo(
```
JungukCho, line 255: suggestion: matchedSetInfo or hasSameSetInfo?
Do not strong opinion.

npm/debugTools/dataplane/tupleProcessor.go
JungukCho, line 206: suggestion: matchedRules instead getHitRules.
Do not strong opinion.

npm/debugTools/dataplane/tupleProcessor.go
```diff
@@ -0,0 +1,431 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("error occurred during get destination pod : %w", err)
+	}
+
+	hitRules, err := getHitRules(srcPod, dstPod, allRules, npmCache)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error occurred during get hit rules : %w", err)
+	}
+
+	if len(hitRules) == 0 {
```
JungukCho, line 97: Is it better to move this code into getHitRules() function?
```go
      if len(res) == 0 {
		// either no hit rules or no rules at all. Both cases allow all traffic
		res= append(res, &pb.RuleResponse{Allowed: true})
	}
	return res, nil


```

npm/debugTools/dataplane/tupleProcessor.go
```diff
@@ -0,0 +1,431 @@
...
+		return nil, nil, fmt.Errorf("error occurred during get source pod : %w", err)
+	}
+
+	dstPod, err := getCorrespondPod(dst, npmCache)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error occurred during get destination pod : %w", err)
+	}
+
+	hitRules, err := getHitRules(srcPod, dstPod, allRules, npmCache)
+	if err != nil {
```
JungukCho, line 93: error message will be duplicated if err is not nil.
In getHitRules, it also returns
```go
	if err != nil {
				return nil, fmt.Errorf("error occurred during evaluating source's set info : %w", err)
			}
```
Can you please walk through them to clean-up duplicated error messages?

JungukCho (commented):
Thank you for great work and I think this work will be so useful for debugging NPM.
Due to a large PR, there are many comments. Hope they are useful.
Please take them according to your timeline!

npm/debugTools/dataplane/tupleProcessor.go
nddq: I was about to ask you for suggestions 🙂

JungukCho: While we do not include these issues in this PR, but just for brainstorming. 
There are two main interesting issues as future works. 
1. How frequently npm analyzer needs to re-build all data structures since iptables and npm cache information are very dynamic and some information in all created data structures frequently become stale. Thus, it needs a good algorithm to partially update or delete created data structure to avoid creating all of them from scratch. 
2. How to (subsequent) query fast.  I think it is related to this function. For them, we may need to have another data structures by using existing data structures information for fast lookup. 

I am not sure how to deal with it now and If you have any thoughts, please share them.  

Leave it as future works. You can contribute them after internship if you want.  This is open-source and your are the main contributor for this part 👍

JungukCho (approved):
Thank you for applying comments. Great work!

PR #908, opened 2021-06-24
fix: only use rootCtx in main()

rbtr:
**Reason for Change**:

Tidy up the use of rootCtx so that it's only used directly in `main()` and pass child contexts to other functions.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -552,7 +552,7 @@ func main() {
 		}
 		logger.Printf("Set GlobalPodInfoScheme %v", cns.GlobalPodInfoScheme)
 
-		err = InitializeCRDState(httpRestService, cnsconfig)
+		err = InitializeCRDState(rootCtx, httpRestService, cnsconfig)
 		if err != nil {
 			logger.Errorf("Failed to start CRD Controller, err:%v.\n", err)
 			return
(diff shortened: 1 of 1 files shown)


cns/service/main.go
```diff
@@ -526,7 +526,7 @@ func main() {
 	// Initialze state in if CNS is running in CRD mode
 	// State must be initialized before we start HTTPRestService
 	if config.ChannelMode == cns.CRD {
-		err = InitializeCRDState(httpRestService, cnsconfig)
+		err = InitializeCRDState(rootCtx, httpRestService, cnsconfig)
```
ramiro-gamarra, line 529: If, as per PR title, `rootCtx` is only used in main, does it still need to be a package level variable?

rbtr: it does, because it is created in the `init()`, before `main()` executes. 
this change is groundwork for a refactor of this `main.go` to remove all the helper functions to more appropriate locations in their target packages, leaving only init/main and some minimal global state (rootCtx/cancel) in this file/in the main package.

ramiro-gamarra (approved):

ramiro-gamarra (approved):

PR #912, opened 2021-06-26
Drop arp request for snatbridge apipa range

tamilmani1989:
added drop rule for arp request for snat bridge ip

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/ovssnat/ovssnat.go
+++ b/network/ovssnat/ovssnat.go
@@ -327,10 +327,10 @@ func setBridgeMac(hostPrimaryMac string) error {
 	return err
 }
 
-func dropArpForSnatBridgeIP(snatBridgeIP string) error {
+func dropArpForSnatBridgeIP(snatBridgeIP, azSnatVethIfName string) error {
 	var err error
-	ip, _, _ := net.ParseCIDR(snatBridgeIP)
-	if err = ebtables.SetArpDropRuleForIp(ip.String()); err != nil {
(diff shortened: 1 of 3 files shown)


network/ovssnat/ovssnat.go
```diff
@@ -311,26 +314,58 @@ func (client *OVSSnatClient) DeleteSnatEndpoint() error {
...
+	log.Printf("Setting snat bridge mac: %s", hostPrimaryMac)
+	if err := setBridgeMac(hostPrimaryMac); err != nil {
+		return err
 	}
 
-	if err := netlink.AddLink(&link); err != nil {
+	log.Printf("Drop ARP for snat bridge ip: %s", snatBridgeIP)
+	if err := dropArpForSnatBridgeIP(snatBridgeIP, azureSnatVeth0); err != nil {
 		return err
 	}
```
chandanAggarwal, line 370: are these calls idempotent? previously we were just returning if snatbridge exsited?

tamilmani1989: @chandanAggarwal  yes these calls are idempotent...for ebtable we check if rule exists and add it..the other one where we set mac to bridge is idempotent..i tested this in runner setup which creates multuple swift pods and each pod invoke this codepath

chandanAggarwal (approved):
:shipit:

PR #916, opened 2021-06-30
fix: [CNI] handle getting endpoints when state file is empty

matmerr:
**Reason for Change**:

When reconciling with CNI, it may be the case that CNI state hasn't been created yet. Don't block consumers of that state when empty.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/client/client_integration_test.go
+++ b/cni/client/client_integration_test.go
@@ -16,45 +16,47 @@ import (
 	"k8s.io/utils/exec"
 )
 
-func TestMain(m *testing.M) {
-	testutils.RequireRootforTestMain()
+const (
+	stateFilePath = "/var/run/azure-vnet.json"
+)
+
+func setTestFile() error {
 	var err error
 	// copy test state file to /var/run/azure-vnet.json
 	in, err := os.Open("./testresources/azure-vnet-test.json")
 	if err != nil {
-		return
+		return err
 	}
 
-	out, err := os.Create("/var/run/azure-vnet.json")
+	out, err := os.Create(stateFilePath)
 	if err != nil {
-		return
+		return err
 	}
-
-	exit := 0
-	defer func() {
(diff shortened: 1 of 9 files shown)


neaggarwMS (dismissed):
:shipit:

store/json.go
```diff
@@ -71,8 +71,18 @@ func (kvs *jsonFileStore) Read(key string, value interface{}) error {
...
 		defer file.Close()
 
+		b, err := ioutil.ReadAll(file)
+		if err != nil {
+			return err
+		}
+
+		if len(b) == 0 {
+			log.Printf("Unable to read file %s, was empty", kvs.fileName)
+			return nil
```
tamilmani1989, line 81: i would suggest return ErrEmptyFile and handle this in restore function (network/manager.go and ipam/manager.go) as like ErrKeyNotFound

cni/client/client_integration_test.go
```diff
@@ -59,8 +59,7 @@ func TestMain(m *testing.M) {
 
 // todo: enable this test in CI, requires built azure vnet
 func TestGetStateFromAzureCNI(t *testing.T) {
-
-	c := AzureCNIClient{exec: exec.New()}
+	c := New(exec.New())
```
tamilmani1989, line 62: add a test case with empty file and check it returns appropriate error

tamilmani1989 (changes requested):

tamilmani1989 (dismissed):
lgtm

store/json.go
neaggarwMS: @tamilmani1989 , this store file is also read by CNS to save CNS state, if are returning error then we need to handle it in there as well.

neaggarwMS (approved):
:shipit:

PR #922, opened 2021-07-07
chore: appease the linter, pt 1 of ?

rbtr:
**Reason for Change**:

A batch of cosmetic/non-functional changes to fix existing syntactic/style issues discovered by the linter.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cnm/plugin/main.go
+++ b/cnm/plugin/main.go
@@ -207,7 +207,7 @@ func main() {
 
 	// Relay these incoming signals to OS signal channel.
 	osSignalChannel := make(chan os.Signal, 1)
-	signal.Notify(osSignalChannel, os.Interrupt, os.Kill, syscall.SIGTERM)
+	signal.Notify(osSignalChannel, os.Interrupt, syscall.SIGTERM)
 
 	// Wait until receiving a signal.
 	select {
--- a/cnms/service/networkmonitor.go
+++ b/cnms/service/networkmonitor.go
@@ -139,7 +139,7 @@ func main() {
 	tb.ConnectToTelemetryService(telemetryNumRetries, telemetryWaitTimeInMilliseconds)
 	defer tb.Close()
 
-	for true {
+	for {
 		config.Store, err = store.NewJsonFileStore(platform.CNIRuntimePath + pluginName + ".json")
 		if err != nil {
 			fmt.Printf("[monitor] Failed to create store: %v\n", err)
@@ -178,6 +178,4 @@ func main() {
 		time.Sleep(time.Duration(timeout) * time.Second)
 		nm = nil
 	}
-
-	log.Close()
 }
--- a/network/epcommon/endpoint_common.go
+++ b/network/epcommon/endpoint_common.go
@@ -189,7 +189,7 @@ func BlockIPAddresses(bridgeName string, action string) error {
 func EnableIPForwarding(ifName string) error {
 	// Enable ip forwading on linux vm.
 	// sysctl -w net.ipv4.ip_forward=1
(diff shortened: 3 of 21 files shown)


cnms/service/networkmonitor.go
```diff
@@ -178,6 +178,4 @@ func main() {
 		time.Sleep(time.Duration(timeout) * time.Second)
 		nm = nil
 	}
-
-	log.Close()
```
vakalapa, line 182: Dont we have to defer log.close() this ?

rbtr: since the above `for` loop in has no `break` this is unreachable code and we don't currently ever calling `Close()`. 
so i don't know if we need to, but if it hasn't been a problem i would guess no.

tamilmani1989: its good practice to have defer log.Close()..if someone changes this behaviour tomorrow, then they may fail to add log.close()

cnm/plugin/main.go
```diff
@@ -207,7 +207,7 @@ func main() {
 
 	// Relay these incoming signals to OS signal channel.
 	osSignalChannel := make(chan os.Signal, 1)
-	signal.Notify(osSignalChannel, os.Interrupt, os.Kill, syscall.SIGTERM)
+	signal.Notify(osSignalChannel, os.Interrupt, syscall.SIGTERM)
```
tamilmani1989, line 210: did we test this change?

network/epcommon/endpoint_common.go
```diff
@@ -189,7 +189,7 @@ func BlockIPAddresses(bridgeName string, action string) error {
 func EnableIPForwarding(ifName string) error {
 	// Enable ip forwading on linux vm.
 	// sysctl -w net.ipv4.ip_forward=1
-	cmd := fmt.Sprintf(enableIPForwardCmd)
+	cmd := fmt.Sprint(enableIPForwardCmd)
```
tamilmani1989, line 192: why do we change this? what's the use?

telemetry/telemetrybuffer.go
```diff
@@ -254,15 +256,13 @@ func push(x interface{}) {
 		kvs.Unlock(true)
 	}
 
-	switch x.(type) {
+	switch x := x.(type) {
```
tamilmani1989, line 259: may be diff variable to avoid confusion y := x.(type)?

tamilmani1989 (changes requested):

network/epcommon/endpoint_common.go
rbtr: we are not doing a string interpolation so it's unnecessary to call `Sprintf`, and using the minimal `Sprint` is more correct

cnm/plugin/main.go
rbtr: `os.Kill` cannot be trapped, so including it in the signal list is currently ineffective.

cnms/service/networkmonitor.go
rbtr: @tamilmani1989 I read through the log package and I don't think it's safe to call `log.Close()` (even deferred) the way that it's written right now. We could accidentally close stdout or stderr and prevent the go runtime from writing the crash logs during a panic. That needs some work that out of scope for this linting fix

tamilmani1989 (approved):
cni changes lgtm... leaving to vamsi/mat to sign off npm changes

vakalapa (approved):
lgtm, please wait for all the tests to complete and then you can merge.

PR #923, opened 2021-07-07
[NPM] [Bug] Fixing the looping issue in backup and reconcile chains

vakalapa:
Due to this bug, reconcile loop is dying after one 5 minute loop and not continuing.

### DIFF
--- a/npm/npm.go
+++ b/npm/npm.go
@@ -288,7 +288,7 @@ func NewNetworkPolicyManager(clientset *kubernetes.Clientset, informerFactory in
 }
 
 // reconcileChains checks for ordering of AZURE-NPM chain in FORWARD chain periodically.
-func (npMgr *NetworkPolicyManager) reconcileChains(stopCh <-chan struct{}) error {
+func (npMgr *NetworkPolicyManager) reconcileChains(stopCh <-chan struct{}) {
 	iptMgr := iptm.NewIptablesManager()
 
 	ticker := time.NewTicker(time.Minute * time.Duration(reconcileChainTimeInMinutes))
@@ -296,10 +296,10 @@ func (npMgr *NetworkPolicyManager) reconcileChains(stopCh <-chan struct{}) error
 	for {
 		select {
 		case <-stopCh:
-			return nil
+			return
 		case <-ticker.C:
 			if err := iptMgr.CheckAndAddForwardChain(); err != nil {
-				return err
+				metrics.SendErrorLogAndMetric(util.NpmID, "Error: failed to reconcileChains Azure-NPM due to %s", err.Error())
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


npm/npm.go
```diff
@@ -157,14 +157,22 @@ func (npMgr *NetworkPolicyManager) restore() {
...
-	var err error
-	for {
-		time.Sleep(backupWaitTimeInSeconds * time.Second)
+func (npMgr *NetworkPolicyManager) backup(stopCh <-chan struct{}) {
+	var (
+		err    error
+		iptMgr = iptm.NewIptablesManager()
+		ticker = time.NewTicker(time.Second * time.Duration(backupWaitTimeInSeconds))
+	)
+	defer ticker.Stop()
```
rbtr, line 166: don't declare variables like this, it's preferred to use the one-line declare and initialize whenever you can (and you can, here)
```suggestion
	iptMgr := iptm.NewIptablesManager()
	ticker := time.NewTicker(time.Second * time.Duration(backupWaitTimeInSeconds))
	defer ticker.Stop()
```

and do the same thing for the error in the inner for-select scope, you don't need it at the function level

npm/npm.go
```diff
@@ -157,14 +157,22 @@ func (npMgr *NetworkPolicyManager) restore() {
...
 
-		if err = iptMgr.Save(util.IptablesConfigFile); err != nil {
-			metrics.SendErrorLogAndMetric(util.NpmID, "Error: failed to back up Azure-NPM states %s", err.Error())
+	for {
+		select {
+		case <-stopCh:
+			return
+		case <-ticker.C:
+			if err = iptMgr.Save(util.IptablesConfigFile); err != nil {
```
rbtr, line 173: this err does not need to be function scoped, you can scope it just to this conditional:
```suggestion
			if err := iptMgr.Save(util.IptablesConfigFile); err != nil {
```

npm/npm.go
```diff
@@ -280,13 +288,19 @@ func NewNetworkPolicyManager(clientset *kubernetes.Clientset, informerFactory in
...
 
 // reconcileChains checks for ordering of AZURE-NPM chain in FORWARD chain periodically.
-func (npMgr *NetworkPolicyManager) reconcileChains() error {
+func (npMgr *NetworkPolicyManager) reconcileChains(stopCh <-chan struct{}) {
 	iptMgr := iptm.NewIptablesManager()
-	select {
-	case <-time.After(reconcileChainTimeInMinutes * time.Minute):
-		if err := iptMgr.CheckAndAddForwardChain(); err != nil {
-			return err
+
```
rbtr, line 293: you can use newlines to separate logical blocks, otherwise don't use extra vertical whitespace

rbtr (changes requested):
the fix lgtm! but i have some feedback in other changed code

rbtr (approved):

PR #924, opened 2021-07-07
[NPM] [bug] Adding sort in iptable comment for deterministic behavior in Flake UT

vakalapa:
TestAllowMultiplePodSelectors has a flake UT result due to ordering of ipset name in the comment. This PR fixes it

### DIFF
--- a/npm/translatePolicy.go
+++ b/npm/translatePolicy.go
@@ -208,8 +208,7 @@ func craftPartialIptablesCommentFromSelector(ns string, selector *metav1.LabelSe
 	}
 
 	sort.Strings(comments)
-	comment := strings.Join(comments, "-AND-") + "-AND-"
-	return comment[:len(comment)-len("-AND-")] + postfix
+	return strings.Join(comments, "-AND-") + postfix
 }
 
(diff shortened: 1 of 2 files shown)


npm/translatePolicy.go
```diff
@@ -200,11 +202,13 @@ func craftPartialIptablesCommentFromSelector(ns string, selector *metav1.LabelSe
...
-	for i, _ := range labelsWithoutOps {
-		comment += prefix + ops[i] + labelsWithoutOps[i]
-		comment += "-AND-"
+	comments := []string{}
+	for i := range labelsWithoutOps {
+		comments = append(comments, prefix+ops[i]+labelsWithoutOps[i])
 	}
 
+	sort.Strings(comments)
+	comment := strings.Join(comments, "-AND-") + "-AND-"
```
rbtr, line 211: are you appending `-AND-` but then immediately chomping it off below when returning `[:len - len("-AND-")]`?

rbtr (commented):
lgtm with one question/change

npm/translatePolicy.go
vakalapa: Correcting that.

rbtr (approved):
lgtm

PR #925, opened 2021-07-08
chore: appease the linter, pt 2 of ?

rbtr:
**Reason for Change**:

A batch of cosmetic/non-functional changes to fix existing syntactic/style issues discovered by the linter.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/networkcontainers/networkcontainers.go
+++ b/cns/networkcontainers/networkcontainers.go
@@ -22,20 +22,18 @@ import (
 )
 
 const (
-	versionStr              = "cniVersion"
-	pluginsStr              = "plugins"
-	nameStr                 = "name"
-	k8sPodNamespaceStr      = "K8S_POD_NAMESPACE"
-	k8sPodNameStr           = "K8S_POD_NAME"
-	k8sPodInfraContainerStr = "K8S_POD_INFRA_CONTAINER_ID"
-	cniAdd                  = "ADD"
-	cniDelete               = "DEL"
-	cniUpdate               = "UPDATE"
+	versionStr         = "cniVersion"
+	pluginsStr         = "plugins"
+	nameStr            = "name"
+	k8sPodNamespaceStr = "K8S_POD_NAMESPACE"
+	k8sPodNameStr      = "K8S_POD_NAME"
+	cniAdd             = "ADD"
+	cniDelete          = "DEL"
+	cniUpdate          = "UPDATE"
 )
 
 // NetworkContainers can be used to perform operations on network containers.
 type NetworkContainers struct {
-	logpath string
 }
 
 // NetPluginConfiguration represent network plugin configuration that is used during CNI ADD/DELETE/UPDATE operation
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -117,7 +117,7 @@ func (service *HTTPRestService) SyncNodeStatus(
 			var resp cns.CreateNetworkContainerResponse
 			if err = json.Unmarshal(w.Body.Bytes(), &resp); err == nil && resp.Response.ReturnCode == types.Success {
 				service.Lock()
-				ncstatus, _ := service.state.ContainerStatus[ncid]
+				ncstatus := service.state.ContainerStatus[ncid]
 				ncstatus.VfpUpdateComplete = !waitingForUpdate
 				service.state.ContainerStatus[ncid] = ncstatus
 				service.Unlock()
@@ -337,7 +337,7 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(
 	existingNCInfo, ok := service.getNetworkContainerDetails(req.NetworkContainerid)
 	if ok {
 		existingReq := existingNCInfo.CreateNetworkContainerRequest
-		if reflect.DeepEqual(existingReq.IPConfiguration, req.IPConfiguration) != true {
+		if !reflect.DeepEqual(existingReq.IPConfiguration, req.IPConfiguration) {
 			logger.Errorf("[Azure CNS] Error. PrimaryCA is not same, NCId %s, old CA %s, new CA %s", req.NetworkContainerid, existingReq.PrimaryInterfaceIdentifier, req.PrimaryInterfaceIdentifier)
 			return types.PrimaryCANotSame
 		}
--- a/cns/restserver/internalapi_test.go
+++ b/cns/restserver/internalapi_test.go
@@ -30,16 +30,7 @@ const (
 )
 
 var (
-	dnsservers        = []string{"8.8.8.8", "8.8.4.4"}
-	hostSupportedApis = `<SupportedRequestTypes>
-							<type>GetSupportedApis</type>
-							<type>GetIpRangesV1</type>
-							<type>GetIpRangesV2</type>
-							<type>GetInterfaceInfoV1</type>
-							<type>PortContainerIOVInformationV1</type>
-							<type>NetworkManagement</type>
-							<type>NetworkManagementDNSSupport</type>
-						</SupportedRequestTypes>`
+	dnsservers = []string{"8.8.8.8", "8.8.4.4"}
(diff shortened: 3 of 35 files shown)


cns/restserver/ipam.go
```diff
@@ -442,28 +438,25 @@ func (service *HTTPRestService) AllocateDesiredIPConfig(podInfo cns.PodInfo, des
...
+			if err := service.setIPConfigAsAllocated(ipState, podInfo); err != nil {
+				return cns.PodIpInfo{}, err
 			}
 
-			err = service.populateIpConfigInfoUntransacted(ipState, &podIpInfo)
-			if err != nil {
-				return podIpInfo, err
+			podIPInfo := cns.PodIpInfo{}
+			if err := service.populateIpConfigInfoUntransacted(ipState, &podIPInfo); err != nil {
+				return cns.PodIpInfo{}, err
```
JungukCho, line 452: Is it normal to return this instance even with error case? 
I feel like to have *cns.PodIpInfo as return value and return nil when it has error.

cns/restserver/util.go
```diff
@@ -387,13 +383,13 @@ func (service *HTTPRestService) getNetworkContainerResponse(
 			}
 
 			vfpUpdateComplete := !waitingForUpdate
-			ncstatus, _ := service.state.ContainerStatus[containerID]
+			ncstatus := service.state.ContainerStatus[containerID]
```
JungukCho, line 386: assume map always has ipconfig Or it is ok to use zero value of the return type.

cns/restserver/internalapi.go
```diff
@@ -337,7 +337,7 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(
 	existingNCInfo, ok := service.getNetworkContainerDetails(req.NetworkContainerid)
 	if ok {
 		existingReq := existingNCInfo.CreateNetworkContainerRequest
-		if reflect.DeepEqual(existingReq.IPConfiguration, req.IPConfiguration) != true {
+		if !reflect.DeepEqual(existingReq.IPConfiguration, req.IPConfiguration) {
 			logger.Errorf("[Azure CNS] Error. PrimaryCA is not same, NCId %s, old CA %s, new CA %s", req.NetworkContainerid, existingReq.PrimaryInterfaceIdentifier, req.PrimaryInterfaceIdentifier)
```
JungukCho, line 341: I saw a lot of [Azure CNS], etc. I am not sure, but I think the purpose of this label is to find where the error is from. Why not moving to `klog`. In NPM, we are gradually moving to `klog`. It gives file name and line numbers for free. So it is much easy to maintain the code.

network/network_linux.go
```diff
@@ -21,19 +21,21 @@ const (
...
-	systemdResolvConfFile = "/run/systemd/resolve/resolv.conf"
-	SnatBridgeIPKey       = "snatBridgeIP"
-	LocalIPKey            = "localIP"
-	InfraVnetIPKey        = "infraVnetIP"
-	OptVethName           = "vethname"
+	virtualMacAddress = "12:34:56:78:9a:bc"
+	versionID         = "VERSION_ID"
+	distroID          = "ID"
+	ubuntuStr         = "ubuntu"
+	dnsServersStr     = "DNS Servers"
```
JungukCho, line 28: Minor : is it necessary to put "Str" as suffix in several consts?

testutils/store_mock.go
```diff
@@ -2,8 +2,12 @@ package testutils
 
 import (
 	"time"
+
+	"github.com/Azure/azure-container-networking/store"
 )
 
+var _ store.KeyValueStore = (*KeyValueStoreMock)(nil)
```
JungukCho, line 9: curious: what is the purpose of this anonymous variable?

JungukCho (commented):
Most of comments are just two cents and my curiosity.
I think there are no functional changes in this PR and good to go.
However, I want to listen to your feedback about comments before approving this PR if you are ok.

Btw, thank you for improving code quality!

testutils/store_mock.go
rbtr: this coerces `nil` in to type `*KeyValueStoreMock` and assigns it an anonymous interface type `store.KeyValueStore`. it has no run-time footprint, since it's a nil assigned to no variable, and it tells us at compile-time that the *Struct is an implementation of the Interface.

cns/networkcontainers/networkcontainers.go
```diff
@@ -22,20 +22,18 @@ import (
...
-	versionStr              = "cniVersion"
-	pluginsStr              = "plugins"
-	nameStr                 = "name"
-	k8sPodNamespaceStr      = "K8S_POD_NAMESPACE"
-	k8sPodNameStr           = "K8S_POD_NAME"
-	k8sPodInfraContainerStr = "K8S_POD_INFRA_CONTAINER_ID"
-	cniAdd                  = "ADD"
-	cniDelete               = "DEL"
-	cniUpdate               = "UPDATE"
+	versionStr         = "cniVersion"
```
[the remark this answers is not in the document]
rbtr: no, that is discouraged (but out of scope on this PR)

cns/networkcontainers/networkcontainers.go
```diff
@@ -22,20 +22,18 @@ import (
...
+	pluginsStr         = "plugins"
+	nameStr            = "name"
+	k8sPodNamespaceStr = "K8S_POD_NAMESPACE"
+	k8sPodNameStr      = "K8S_POD_NAME"
+	cniAdd             = "ADD"
```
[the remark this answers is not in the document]
rbtr: yes, I try to make types for this stuff also (but it is out of scope here)

cns/restserver/internalapi.go
```diff
@@ -117,7 +117,7 @@ func (service *HTTPRestService) SyncNodeStatus(
 			var resp cns.CreateNetworkContainerResponse
 			if err = json.Unmarshal(w.Body.Bytes(), &resp); err == nil && resp.Response.ReturnCode == types.Success {
 				service.Lock()
-				ncstatus, _ := service.state.ContainerStatus[ncid]
+				ncstatus := service.state.ContainerStatus[ncid]
```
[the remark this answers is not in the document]
rbtr: it is something to consider, but I'm not changing functionality yet

cns/restserver/internalapi_test.go
```diff
@@ -164,7 +155,7 @@ func TestPendingIPsGotUpdatedWhenSyncHostNCVersion(t *testing.T) {
 	if len(receivedSecondaryIPConfigs) != 1 {
 		t.Errorf("Unexpected receivedSecondaryIPConfigs length %d, expeted length is 1", len(receivedSecondaryIPConfigs))
 	}
-	for i, _ := range receivedSecondaryIPConfigs {
+	for i := range receivedSecondaryIPConfigs {
```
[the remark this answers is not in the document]
rbtr: IMO if you are simply iterating the whole list and doing something with the indexes, ranging is simpler and easier to read/understand. I would only do the C-style `for` when the start/end conditions are not 0/len(slice), or if I did not have a range-able data structure (like if I just knew I needed to loop 10 times).

cns/restserver/internalapi.go
rbtr: yes, we should standardize logging and get rid of our custom logging packages (I would actually prefer we move all our projects to [`zap`](https://github.com/uber-go/zap)). out of scope here, though.

cns/restserver/ipam.go
rbtr: it's okay to return the zero-val with an error if it doesn't make sense to be using pointers. the caller should always check for errors first.

JungukCho (approved):
lgtm

rbtr:
finally 🎉 ty @JungukCho

PR #928, opened 2021-07-09
move cni statefile to subdir and add migration functionality in json store initialization

rbtr:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -6,6 +6,7 @@ package network
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -135,7 +136,7 @@ func (plugin *netPlugin) Start(config *common.PluginConfig) error {
 	// Initialize base plugin.
 	err := plugin.Initialize(config)
 	if err != nil {
-		log.Printf("[cni-net] Failed to initialize base plugin, err:%v.", err)
+		log.Printf("[cni-net] Failed to initialize base plugin, err: %v.", err)
 		return err
 	}
 
@@ -295,13 +296,13 @@ func addNatIPV6SubnetInfo(nwCfg *cni.NetworkConfig,
 	}
 }
 
-func acquireLockForStore(config *common.PluginConfig, plugin *netPlugin) error {
+func (plugin *netPlugin) acquireLockForStore() error {
 	var err error
 	if err = plugin.Store.Lock(true); err != nil {
 		log.Printf("[CNI] Failed to lock store: %v. check if process running", err)
 		if isSafe, _ := plugin.IsSafeToRemoveLock(azureCniName); isSafe {
 			log.Printf("[CNI] Removing lock file as process holding lock exited")
-			if err = releaseLockForStore(plugin); err != nil {
+			if err = plugin.releaseLockForStore(); err != nil {
 				log.Errorf("Failed to release lock file, err:%v.\n", err)
 			}
 		}
@@ -310,7 +311,7 @@ func acquireLockForStore(config *common.PluginConfig, plugin *netPlugin) error {
 	return err
 }
 
-func releaseLockForStore(plugin *netPlugin) error {
+func (plugin *netPlugin) releaseLockForStore() error {
 	if plugin.Store != nil {
 		err := plugin.Store.Unlock(false)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


cni/plugin.go
```diff
@@ -153,7 +153,10 @@ func (plugin *Plugin) InitializeKeyValueStore(config *common.PluginConfig) error
 	// Create the key value store.
 	if plugin.Store == nil {
 		var err error
-		plugin.Store, err = store.NewJsonFileStore(platform.CNIRuntimePath + plugin.Name + ".json")
```
neaggarwMS, line 156: can you remove the main.exe that got added accidently?

store/json.go
```diff
@@ -55,9 +57,115 @@
...
+		if errors.Is(err, os.ErrNotExist) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+// NewJsonFileStore creates a new jsonFileStore object, accessed as a KeyValueStore.
+func NewJsonFileStoreOrMigrate(legacyFileName, fileName string) (KeyValueStore, error) {
```
neaggarwMS, line 81: Can you add UT for this?

store/json.go
```diff
@@ -55,9 +57,115 @@ func NewJsonFileStore(fileName string) (KeyValueStore, error) {
...
+	}
+
+	return kvs, nil
+}
+
+// fileExists states the passed filename and returns if it exists.
+// if any error except for os.ErrNotExists is encountered, it is returned.
+func fileExists(filename string) (bool, error) {
+	if _, err := os.Stat(filename); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
```
matmerr, line 72: can probably use `os.IsNotExist(err)` in lieu of manually checking error type

store/json.go
```diff
@@ -55,9 +57,115 @@ func NewJsonFileStore(fileName string) (KeyValueStore, error) {
...
+	// the statefile does not exist, so we need to lock and check if there is
+	// a legacy statefile that we need to migrate
+	if err := kvs.Lock(true); err != nil {
+		return nil, err
+	}
+	defer kvs.Unlock(false)
+
+	// check if the statefile exists again, now that we have acquired the lock,
+	// just in case someone else wrote it since we checked but before we locked
+	ok, err = fileExists(fileName)
```
matmerr, line 132: potentially wrap the error with context, calling `fileExists(fileName)` multiple times in this func on the same filepath and returning the error could be difficult to trace

store/json.go
matmerr: offline, apparently os.IsNotExist(err) is deprecated

rbtr: full context: errors.Is/As should be preferred to the built-in error type functions because they correctly handle wrapped errors and legacy built-ins like os.IsNotExist are not guaranteed to

tamilmani1989 (changes requested):
will review the code further once you change it as per we discussed

cni/network/network.go
```diff
@@ -322,6 +323,63 @@
 	log.Printf("Released lock file")
 	return nil
 }
+
+// MigrateKeyValueStore attempts to idempotently migrate a kvstore at the
+// legacy location in to the new location as a one-time operation. If a
+// migration is not necessary, returns nil. If a migration is required but
+// does not succeed for any reason, returns an error.
+func (plugin *netPlugin) MigrateKeyValueStore() error {
```
neaggarwMS, line 331: >MigrateKeyValueStore [](http://example.com/codeflow?start=25&length=20)

Can you add the test for this API? Feel free to add a work item to add it later if this requires more changes

platform/os_linux.go
```diff
@@ -19,7 +19,9 @@ const (
 	// CNMRuntimePath is the path where CNM state files are stored.
 	CNMRuntimePath = "/var/lib/azure-network/"
 	// CNIRuntimePath is the path where CNI state files are stored.
-	CNIRuntimePath = "/var/run/"
+	CNIRuntimePath = "/var/run/azure-vnet/"
```
neaggarwMS, line 22: Are we changing this for Multitenancy too? @tamilmani1989 ?

cni/network/network.go
```diff
@@ -322,6 +323,63 @@
...
+	ok, err = platform.CheckIfFileExists(filename)
+	if err != nil {
+		return fmt.Errorf("error checking current file store location %s after lock: %w", filename, err)
+	}
+	if ok {
+		return nil
+	}
+
+	// we have the lock and still require a migration
+	return platform.Copy(legacyFilename, filename)
```
neaggarwMS, line 380: @tamilmani1989, this wont impact the temp file created, right?

tamilmani1989: nope it wont

cni/network/plugin/main.go
```diff
@@ -188,7 +188,17 @@ func main() {
 			reportPluginError(reportManager, tb, err)
 			tb.Close()
 		}
+		return
+	}
 
+	// try and migrate the kvstore
+	if err = netPlugin.MigrateKeyValueStore(); err != nil {
```
tamilmani1989, line 195: can you move this call after line 224 .so you dont need to create telemetry buffer and connect .
you can use reportPluginError

platform/os.go
```diff
@@ -71,3 +72,24 @@ func CreateDirectory(dirPath string) error {
...
+	dest, err := os.Create(destinationFilename)
+	if err != nil {
+		return err
+	}
+	defer dest.Close()
+
+	if _, err := io.Copy(dest, src); err != nil {
+		return err
+	}
+	return dest.Sync()
```
tamilmani1989, line 94: can you add comment what this sync does

tamilmani1989 (changes requested):

cni/network/network_test.go
```diff
@@ -138,3 +142,27 @@
...
+
+	legacy, err := os.Create(legacyFilename)
+	assert.NoError(t, err)
+	defer os.Remove(legacy.Name())
+	defer legacy.Close()
+
+	_, err = legacy.WriteString("test")
+	assert.NoError(t, err)
+
+	err = plugin.MigrateKeyValueStore(filename, legacyFilename)
```
neaggarwMS, line 162: Can you add a test case where in first case legacyFileName doesnt exist

neaggarwMS (changes requested):
:clock1:

PR #929, opened 2021-07-10
removed lock for version command

tamilmani1989:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -369,7 +369,11 @@ func (plugin *netPlugin) Add(args *cniSkel.CmdArgs) error {
 
 	// acquire cni lock file
 	config := &common.PluginConfig{}
-	acquireLockForStore(config, plugin)
+	if err := acquireLockForStore(config, plugin); err != nil {
+		log.Errorf("Couldn't acquire lock: %+v", err)
+		return err
+	}
+
 	defer releaseLockForStore(plugin)
 
 	// restore network state
@@ -805,7 +809,11 @@ func (plugin *netPlugin) Get(args *cniSkel.CmdArgs) error {
 	iptables.DisableIPTableLock = nwCfg.DisableIPTableLock
 
 	config := &common.PluginConfig{}
-	acquireLockForStore(config, plugin)
+	if err := acquireLockForStore(config, plugin); err != nil {
+		log.Errorf("Couldn't acquire lock: %+v", err)
+		return err
+	}
+
 	defer releaseLockForStore(plugin)
 
 	// restore network state
@@ -902,7 +910,11 @@ func (plugin *netPlugin) Delete(args *cniSkel.CmdArgs) error {
 
 	// acquire cni lock file
 	config := &common.PluginConfig{}
-	acquireLockForStore(config, plugin)
+	if err := acquireLockForStore(config, plugin); err != nil {
+		log.Errorf("Couldn't acquire lock: %+v", err)
+		return err
+	}
+
 	defer releaseLockForStore(plugin)
 
 	// restore network state
@@ -1076,7 +1088,11 @@ func (plugin *netPlugin) Update(args *cniSkel.CmdArgs) error {
 
 	// acquire cni lock file
 	config := &common.PluginConfig{}
-	acquireLockForStore(config, plugin)
+	if err := acquireLockForStore(config, plugin); err != nil {
+		log.Errorf("Couldn't acquire lock: %+v", err)
+		return err
+	}
+
 	defer releaseLockForStore(plugin)
 
 	// restore network state
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -297,13 +297,14 @@ func addNatIPV6SubnetInfo(nwCfg *cni.NetworkConfig,
 
 func acquireLockForStore(config *common.PluginConfig, plugin *netPlugin) error {
 	var err error
-	if err = plugin.InitializeKeyValueStore(config); err != nil {
-		log.Errorf("Failed to initialize key-value store of network plugin, err:%v.\n", err)
-
+	if err = plugin.Store.Lock(true); err != nil {
+		log.Printf("[CNI] Failed to lock store: %v. check if process running", err)
 		if isSafe, _ := plugin.IsSafeToRemoveLock(azureCniName); isSafe {
 			log.Printf("[CNI] Removing lock file as process holding lock exited")
-			if err = plugin.UninitializeKeyValueStore(true); err != nil {
-				log.Errorf("Failed to uninitialize key-value store of network plugin, err:%v.\n", err)
+			if errUninit := releaseLockForStore(plugin); errUninit != nil {
+				log.Errorf("Failed to release lock file, err:%v.\n", errUninit)
+			} else {
+				err = nil
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


cni/network/network.go
```diff
@@ -349,6 +367,18 @@
 	iptables.DisableIPTableLock = nwCfg.DisableIPTableLock
 	plugin.setCNIReportDetails(nwCfg, CNI_ADD, "")
 
+	// acquire cni lock file
+	config := &common.PluginConfig{}
+	acquireLockForStore(config, plugin)
```
neaggarwMS, line 372: Can you add err check and return on failure

cni/network/network.go
```diff
@@ -349,6 +367,18 @@
...
+	defer releaseLockForStore(plugin)
```
neaggarwMS, line 373: >defer releaseLockForStore(plugin) [](http://example.com/codeflow?start=1&length=33)

Shouldnt this be above acquire, always safe to release in case it got acquired above and failed after

cni/network/network.go
```diff
@@ -145,12 +146,6 @@
 	common.LogNetworkInterfaces()
 
 	// Initialize network manager.
-	err = plugin.nm.Initialize(config, rehydrateNetworkInfoOnReboot)
```
neaggarwMS, line 148: Why did we move this?

tamilmani1989: we restore state inside this call which we have to do after lock

cni/network/network.go
tamilmani1989: nope not needed, if this fails no need to unlock

cni/network/network.go
neaggarwMS: same for other two calls

neaggarwMS (changes requested):
:clock1:

cni/network/network.go
```diff
@@ -301,6 +295,30 @@
...
+			if err = plugin.UninitializeKeyValueStore(true); err != nil {
+				log.Errorf("Failed to uninitialize key-value store of network plugin, err:%v.\n", err)
+			}
+		}
+	}
+
+	return err
+}
+
+func releaseLockForStore(plugin *netPlugin) error {
```
neaggarwMS, line 314: Can we add a comment that we are releasing the lock with pid?

cni/network/network.go
```diff
@@ -301,6 +295,30 @@
 	}
 }
 
+func acquireLockForStore(config *common.PluginConfig, plugin *netPlugin) error {
```
neaggarwMS, line 298: similary can we log, acquiring the lock file

cni/network/network.go
tamilmani1989: added log..pid will be part of every log..so we can figure out from that

neaggarwMS (dismissed):
:shipit:

cni/network/network.go
```diff
@@ -301,6 +295,35 @@ func addNatIPV6SubnetInfo(nwCfg *cni.NetworkConfig,
...
+
+	return err
+}
+
+func releaseLockForStore(plugin *netPlugin) error {
+	if plugin.Store != nil {
+		err := plugin.Store.Unlock(false)
+		if err != nil {
+			log.Printf("[cni] Failed to unlock store: %v.", err)
+			return err
```
rbtr, line 320: it doesn't look like you're ever checking the returned error from this function - you're only calling this in defer - so maybe it should not return an error and the log is sufficient.

cni/network/network.go
```diff
@@ -301,6 +295,35 @@ func addNatIPV6SubnetInfo(nwCfg *cni.NetworkConfig,
 	}
 }
 
+func acquireLockForStore(config *common.PluginConfig, plugin *netPlugin) error {
+	var err error
+	if err = plugin.Store.Lock(true); err != nil {
```
rbtr, line 300: following this err is hard. it might be better to make this a one-line declaration/initialization (`if err := ...; err != nil`) and in the nested `if`s below, explicitly return `err` or `nil` early instead of sometimes setting `err = nil` and returning `err` at the end of the function.

cni/network/network.go
```diff
@@ -349,6 +372,23 @@ func (plugin *netPlugin) Add(args *cniSkel.CmdArgs) error {
...
+	if err := acquireLockForStore(config, plugin); err != nil {
+		log.Errorf("Couldn't acquire lock: %+v", err)
+		return err
+	}
+
+	defer releaseLockForStore(plugin)
+
+	// restore network state
+	err = plugin.nm.Initialize(config, rehydrateNetworkInfoOnReboot)
+	if err != nil {
```
rbtr, line 387: can be one-line declaration/initialization `if err := ...; err != nil {`

cni/network/network.go
```diff
@@ -774,6 +814,21 @@ func (plugin *netPlugin) Get(args *cniSkel.CmdArgs) error {
...
+	config := &common.PluginConfig{Store: plugin.Store}
+	if err := acquireLockForStore(config, plugin); err != nil {
+		log.Errorf("Couldn't acquire lock: %+v", err)
+		return err
+	}
+
+	defer releaseLockForStore(plugin)
+
+	// restore network state
+	err = plugin.nm.Initialize(config, rehydrateNetworkInfoOnReboot)
```
rbtr, line 826: same, can be one-line decl/init

cni/network/network.go
```diff
@@ -349,6 +372,23 @@ func (plugin *netPlugin) Add(args *cniSkel.CmdArgs) error {
...
+	if err != nil {
+		log.Printf("[cni-net] Failed to initialize network manager, err:%v.", err)
```
rbtr, line 388: for readability i would like to see space in `err:%v.` -> `err: %v.`, but that applies to the whole codebase so i won't block for that

cni/network/network.go
```diff
@@ -859,6 +914,22 @@ func (plugin *netPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+	config := &common.PluginConfig{Store: plugin.Store}
+	if err := acquireLockForStore(config, plugin); err != nil {
+		log.Errorf("Couldn't acquire lock: %+v", err)
+		return err
+	}
+
+	defer releaseLockForStore(plugin)
+
+	// restore network state
+	err = plugin.nm.Initialize(config, rehydrateNetworkInfoOnReboot)
```
rbtr, line 927: one line

cni/network/network.go
```diff
@@ -1021,6 +1092,22 @@ func (plugin *netPlugin) Update(args *cniSkel.CmdArgs) error {
...
```
rbtr, line 1105: one line

rbtr (changes requested):

cni/network/network.go
```diff
@@ -349,6 +372,23 @@
...
```
neaggarwMS, line 386: @rbtr , dont we need this also during startup?

cni/network/network.go
tamilmani1989: i wanted to avoid multiple returns

cni/network/network.go
tamilmani1989: im checking it in another place (inside acquireLockForStore)

cni/network/network.go
rbtr: multiple returns are okay, in go it's encouraged to return early when you can. returning early on errors lets us keep the code linear and not need to nest it as deeply.

rbtr (approved):

PR #930, opened 2021-07-12
[NPM] Support endPort feature

JungukCho:
**Reason for Change**:

This PR is to support [endPort feature ](https://kubernetes.io/docs/concepts/services-networking/network-policies/#targeting-a-range-of-ports)in NPM which is supported by Kubernetes v1.21 [alpha]
Only "npm/translatePolicy.go" and "npm/translatePolicy_test.go" files are updated and the rest of them are module updates.

For example, below yaml snippet from example network policy from [endPort feature ](https://kubernetes.io/docs/concepts/services-networking/network-policies/#targeting-a-range-of-ports), the target port is between the range 32000 and 32768.
```yaml
    ports:
    - protocol: TCP
      port: 32000
      endPort: 32768
```
Since iptables supports "--destination-port port[:port]", this PR leverages the capability.
Below is data plane representation of above yaml snippet with this PR and without this PR
```shell
# With this PR (dpts:32000:32768")
Chain AZURE-NPM-INGRESS-PORT (1 references)
    0     0 MARK       tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            match-set azure-npm-784554818 dst match-set azure-npm-1547420863 dst match-set azure-npm-784554818 src match-set azure-npm-3036499794 src tcp **dpts:32000:32768** /* ALLOW-role:client-IN-ns-default-AND-TCP-PORT-32000:32768-TO-role:db-IN-ns-default */ MARK set 0x2000

# Current NPM without this PR ("dpt:32000")
Chain AZURE-NPM-INGRESS-PORT (1 references)
    0     0 MARK       tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            match-set azure-npm-784554818 dst match-set azure-npm-1547420863 dst match-set azure-npm-784554818 src match-set azure-npm-3036499794 src **tcp dpt:32000** /* ALLOW-role:client-IN-ns-default-AND-TCP-PORT-32000-TO-role:db-IN-ns-default */ MARK set 0x2000
```

**Issue Fixed**:

**Requirements**:

Based on [endPort feature ](https://kubernetes.io/docs/concepts/services-networking/network-policies/#targeting-a-range-of-ports), this feature as an alpha feature is disabled by default. To enable the endPort field at a cluster level, you (or your cluster administrator) need to enable the NetworkPolicyEndPort feature gate for the API server with --feature-gates=NetworkPolicyEndPort=true,….

In addition, it needs to update multiple modules (e.g., client-go, etc).
Thus, it may need to discuss how to proceed PR.

- [X] includes documentation ("npm/Network policy examples to test NPM new features or bug-fix/end-port" directory in acndoc: describe how to reproduce this PR and you can play with endport)
- [X] adds unit tests

**Notes**:
Since some duplicated rules can be installed in iptables based on submitted network policies, we will need to optimize them later.
For example, when two yaml files (below A.yaml and B.yaml) are the same except for endPort field, the two rules in iptables are installed.
```yaml
# A.yaml
    ports:
    - protocol: TCP
      port: 32000
      endPort: 32768

# B.yaml
    ports:
    - protocol: TCP
      port: 32000
```


npm/translatePolicy.go
```diff
@@ -19,6 +20,19 @@ type portsInfo struct {
...
 }
 
+// getPortRange returns a specific port (only portRule.Port exist) or port ranges (when EndPort exists as well) based on portRule
+func getPortRange(portRule networkingv1.NetworkPolicyPort) (string, bool) {
+	if portRule.Port == nil {
+		return "", false
+	}
+
+	portRange := portRule.Port.String()
+	if portRule.EndPort != nil {
```
vakalapa, line 30: I know API would have done validations, but can we do some validations too to make sure endport is always greater than start port.

JungukCho: This is a valid point.  I also considered this and decided not to check it in the NPM code since API server validates the input and end-port is a spec in network policy since I think this is a string contract.
In addition, even though we check the validation in NPM and "end port" is less than "start port" (while it does not happen without our code bug), what is the right action in NPM? I think there are no right actions for this.   

Please let me know you have other opinions (or string opinion to add the logic).

Btw, I am curious how iptables react when I insert "[endport]:[start port]" order in iptables. I will check it.

JungukCho:
Confirmed it with cyclonus test. 
I will check what components are broken with client-go version update.

npm/translatePolicy.go
vakalapa: Any update on the check you mentioned above? you are right, API server will maintain that sanity checks and we can trust it.

JungukCho:
Close this PR since master already has required package version (e.g., apiserver, client-go). Raise another PR: https://github.com/Azure/azure-container-networking/pull/1125

PR #933, opened 2021-07-13
write empty json object to empty cni statefile if exists

rbtr:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/cnireconciler/statefile.go
+++ b/cns/cnireconciler/statefile.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"os"
+
+	"github.com/Azure/azure-container-networking/cns/logger"
 )
 
 // WriteObjectToCNIStatefile checks for a file at the CNI statefile path,
@@ -23,7 +25,7 @@ func WriteObjectToCNIStatefile() error {
 func writeObjectToFile(filename string) error {
 	fi, err := os.Stat(filename)
 	if err != nil {
-		if errors.Is(os.ErrNotExist, err) {
+		if errors.Is(err, os.ErrNotExist) {
 			return nil
 		}
 		return err
@@ -33,6 +35,7 @@ func writeObjectToFile(filename string) error {
 		return nil
 	}
 
+	logger.Printf("Writing {} to CNI statefile")
 	b, _ := json.Marshal(map[string]string{})
 	return os.WriteFile(filename, b, os.FileMode(0666))
 }
--- a/cns/cnireconciler/statefile.go
+++ b/cns/cnireconciler/statefile.go
@@ -1,0 +1,38 @@
+package cnireconciler
+
+import (
+	"encoding/json"
+	"errors"
+	"os"
+)
+
+// WriteObjectToCNIStatefile checks for a file at the CNI statefile path,
+// and checks if it is empty. If it is empty, writes an empty JSON object to
+// it so older CNI can execute. Does nothing and returns no error if the
+// file does not exist.
+//
+// This is a hack to get older CNI to run when CNS has mounted the statefile
+// path, but the statefile wasn't written by CNI yet. Kubelet will stub an
+// empty file on the host filesystem, crashing older CNI because it doesn't know
+// how to handle empty statefiles.
+func WriteObjectToCNIStatefile() error {
+	filename := "/var/run/azure-vnet.json"
+	return writeObjectToFile(filename)
+}
+
+func writeObjectToFile(filename string) error {
+	fi, err := os.Stat(filename)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


cns/cnireconciler/statefile.go
```diff
@@ -0,0 +1,38 @@
...
+			return nil
+		}
+		return err
+	}
+
+	if fi.Size() != 0 {
+		return nil
+	}
+
+	b, _ := json.Marshal(map[string]string{})
```
tamilmani1989, line 36: log before writing empty json to file?

tamilmani1989 (changes requested):

cns/cnireconciler/statefile.go
neaggarwMS, line 36: Shouldnt we acquire the lock? Can there be a race where it will accidently update the correct file?

neaggarwMS: can you also validate reconcile works fine, say cns crashes write after updating this

tamilmani1989: i think that's not possible. I guess this code is at start of cns. cni dependent on cns to get ip and create endpoint. cni would be in error state unless it get ip from cns

neaggarwMS: This is a race, 

T1. CNS starts up file
T2.  CNI gets an IP but hasnt updated the file yet
T3. CNS crashes again
T4. CNS starts checks File size is still 0
T5. After CNS file size check, CNI updates the IP it got at T2
T6. CNS overwrites with empty json

Seems a very rare scenario but it could still happen

rbtr: we can't acquire the lock because in old versions of CNI that path is just `/var/run/azure-vnet.lock` and we're back at the same problem with the hostpath mount

tamilmani1989: @neaggarwMS cns should not return ip to cni unless it creates empty json file. so T4 won't occur.

neaggarwMS: @tamilmani1989, it will during T1 it already created an empty json and became active.

rbtr: if CNS has written `{}` to the file, the file size will be > 0, and it won't re-write `{}` to the file

tamilmani1989: then file size won't be 0 at T4

neaggarwMS: ah, makes sense.

neaggarwMS (dismissed):

tamilmani1989 (dismissed):
lgtm

tamilmani1989 (approved):

cns/cnireconciler/statefile.go
```diff
@@ -0,0 +1,41 @@
...
+func WriteObjectToCNIStatefile() error {
+	filename := "/var/run/azure-vnet.json"
+	return writeObjectToFile(filename)
+}
+
+func writeObjectToFile(filename string) error {
+	fi, err := os.Stat(filename)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
```
neaggarwMS, line 29: Add a log line?

neaggarwMS (approved):

cns/cnireconciler/statefile.go
```diff
@@ -0,0 +1,41 @@
...
+// and checks if it is empty. If it is empty, writes an empty JSON object to
+// it so older CNI can execute. Does nothing and returns no error if the
+// file does not exist.
+//
+// This is a hack to get older CNI to run when CNS has mounted the statefile
+// path, but the statefile wasn't written by CNI yet. Kubelet will stub an
+// empty file on the host filesystem, crashing older CNI because it doesn't know
+// how to handle empty statefiles.
+func WriteObjectToCNIStatefile() error {
+	filename := "/var/run/azure-vnet.json"
```
pjohnst5, line 21: Hi @rbtr 👋 what's the best way I can make this different for OS windows? 

I was thinking about making a `statefile_windows.go` with this same code but different filename, that seems to copy alot however.

Any ideas? I just want to defer to you since you know alot more about go than I do, I can only think about an if statement if the OS is windows, just wondering if that's enough

rbtr: this was a hack for older CNIs, pre 1.4.7 or so; so if you're adding support to Windows now then this might not even be necessary.
if it is necessary, I would expect the only thing that's different on Windows is the filepath, right? So that should be injected from an early OS check all the way down in to this function. OR, probable more correctly, this shouldn't be a hardcoded const and should come from env or config.

PR #941, opened 2021-07-16
[NPM]Make order of iptables entries in UT determinisitc in case of MultipleLabelsToMultipleLabels

JungukCho:
**Reason for Change**:

TestAllowMultiplePodSelectors has a flake UT result due to the ordering of ipset in Specs since the order of all ipset values in Specs is not guaranteed in MultiplePodSelectors case.
Thus, in this UT, before checking the iptables entries, sorting Specs is necessary before comparing them.

You can break this UT with this script without this PR.
```shell

max=1000
idx=0
while [[ $idx -ne $max ]]
do
/usr/local/go/bin/go test -timeout 30s -v --count=1 -run ^TestAllowMultiplePodSelectors$ github.com/Azure/azure-container-networking/npm
if [[ $? -ne 0 ]]
then
	exit
fi

idx=$((idx+1))
echo $idx
done
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
While we have one UT to be sorted in current UTs, we need to generalize this approach in all tests to avoid undeterministic behavior in the added UTs in the future.  In addition, I feel like we need to clean up translatePolicy_test.go since it is unnecessarily too lengthy.

### DIFF
--- a/npm/testpolicies/allow-ns-y-z-pod-b-c-with-namedPort.yaml
+++ b/npm/testpolicies/allow-ns-y-z-pod-b-c-with-namedPort.yaml
@@ -1,7 +1,7 @@
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: allow-ns-y-z-pod-b-c
+  name: allow-ns-y-z-pod-b-c-with-namedPort
   namespace: netpol-4537-x
 spec:
   ingress:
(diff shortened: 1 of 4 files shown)


vakalapa (approved):
lgtm

JungukCho:
There is another issue in this PR (sorted one can cause incorrect results as well). I will update the issue in details and codes according to the issue.

JungukCho:
Close now. I am working on fixing this issue in another PR (refactoring  translatePolicy.go) which fundamentally fixes this issue while there are reusable commits here. I will revisit them in the other PR.

PR #949, opened 2021-07-21
[NPM] Make each controller in npm independent 

JungukCho:
**Reason for Change**:

This PR is the first step to make each controller independent as much as possible.
Current NPM uses a global lock to avoid race condition among controllers even though they can be independent. It results in slow processing events and increases memory usages by queuing k8s resources for long time in client-go.

1. Remove a global lock in NetworkPolicyManager 
2. Make event handling functions (e.g., add, update, delete) for each resource (e.g., Pod, Namespace and NetworkPolicy) lock-free to keep processing event handler logics in each controller independently
3. Concurrently processing between event handling functions and actual reconcile function in each controller
4. Minimize a lock scope as much as possible between podController and Namespace controller
5. Cleaning up codes (e.g., encapsulation, removing redundant codes, etc)

**Issue Fixed**:

**Requirements**:

- [x] includes documentation (in npm doc directory)
- [ ] adds unit tests

**Notes**:
ipsm.go requires more works (e.g., reducing lock scope, error handling, etc) since now it is a synchronization point among all controllers. That will be next PR after completing this PR throughly.

The merging Quang's PR was verified in K8s clusters with Quang's testset, but it was not fully verified with stored file for `TestGetNetworkTuple` function. In local setup, it shows inconsistent results while it passed in pipeline test.
I will dive into the file option in next PR.


npm/http/server/server.go
```diff
@@ -65,12 +65,14 @@ func NewNpmRestServer(listeningAddress string) *NPMRestServer {
 
 func (n *NPMRestServer) GetNpmMgr(npMgr *npm.NetworkPolicyManager) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		npMgr.Lock()
+		// QUESTION(jungukcho) What information does it need from npMgr?
+		// Is this function required to hold lock for npMgr?
```
vakalapa, line 69: Copying the message from old PR:

This function is used by Dataplane parser Quang uses, to get the whole internal cache for PodMap, Nsmap and netpolmap, since this is a read operation, i will let your design drive if lock is required.

But please do make sure that this function will return the 3 maps along with listmap and setmap

vakalapa (commented):
First pass review.

npm/podController.go
```diff
@@ -425,26 +407,30 @@ func (c *podController) syncAddedPod(podObj *corev1.Pod) error {
...
+	c.npmNamespaceCache.Lock()
+	if _, exists := c.npmNamespaceCache.nsMap[newPodObjNs]; !exists {
+		// Create ipset related to namespace which this pod belong to if it does not exist.
+		if err = c.ipsMgr.CreateSet(newPodObjNs, []string{util.IpsetNetHashFlag}); err != nil {
+			c.npmNamespaceCache.Unlock()
 			return fmt.Errorf("[syncAddAndUpdatePod] Error: failed to create ipset for namespace %s with err: %v", newPodObjNs, err)
 		}
 
-		if err = ipsMgr.AddToList(util.KubeAllNamespacesFlag, newPodObjNs); err != nil {
+		if err = c.ipsMgr.AddToList(util.KubeAllNamespacesFlag, newPodObjNs); err != nil {
```
vakalapa, line 421: I did not see where we are creating KubeAllNamespacesFlag ipset, can you point me to that .. 
Same with kube-system ns ipset

vakalapa: ignore: NPM creates ipset while adding to set

npm/nameSpaceController.go
```diff
@@ -304,12 +268,18 @@ func (nsc *nameSpaceController) syncNameSpace(key string) error {
 	}
 
 	if nsObj.DeletionTimestamp != nil || nsObj.DeletionGracePeriodSeconds != nil {
-		return nsc.cleanDeletedNamespace(util.GetNSNameWithPrefix(nsObj.Name))
+		return nsc.cleanDeletedNamespace(cachedNsKey)
+	}
 
+	cachedNsObj, nsExists := nsc.npmNamespaceCache.nsMap[cachedNsKey]
```
vakalapa, line 274: Since we are checking the nsObj existence here? why not just call from syncAddNameSpace, instead of going through syncUpdateNameSpace and re-checking the same conditions again on the obj which wont change. 

I like the original approach here, of filtering these kind of events in updateNamespace or addNamespace, that way we do not burden our worker go routine to have to filter events.

npm/nameSpaceController.go
```diff
@@ -199,35 +179,21 @@ func (nsc *nameSpaceController) deleteNamespace(obj interface{}) {
 		return
 	}
 
-	nsc.npMgr.Lock()
```
vakalapa, line 202: We should filter events in these functions itself. reducing the burden on worker go routine.

npm/nameSpaceController.go
```diff
@@ -157,18 +149,6 @@ func (nsc *nameSpaceController) updateNamespace(old, new interface{}) {
 		}
 	}
 
-	nsKey := util.GetNSNameWithPrefix(key)
```
vakalapa, line 160: Same keeping these filtering mechanisms in here would help churn through duplicates events faster with less code involved.

npm/ipsm/ipsm.go
```diff
@@ -124,9 +128,64 @@ func isNsSet(setName string) bool {
...
+		}
+
+		return exitCode, errfmt
+	}
+
+	return 0, nil
+}
+
+func (ipsMgr *IpsetManager) createList(listName string) error {
+	if _, exists := ipsMgr.listMap[listName]; exists {
```
vakalapa, line 188: Can we move this check to Public caller ` CreateList(listName string)`, since when we call createList from with the package we are already checking this conditions

npm/ipsm/ipsm.go
```diff
@@ -137,71 +196,137 @@ func (ipsMgr *IpsetManager) CreateList(listName string) error {
...
 	return nil
 }
 
-// DeleteList removes an ipset list.
-func (ipsMgr *IpsetManager) DeleteList(listName string) error {
+// createSet creates an ipset.
+func (ipsMgr *IpsetManager) createSet(setName string, spec []string) error {
+	timer := metrics.StartNewTimer()
+
+	if _, exists := ipsMgr.setMap[setName]; exists {
```
vakalapa, line 232: sam with this check, i think we can safely move this to the public CreateSet function

npm/ipsm/ipsm.go
vakalapa: Can you double check if this change causes an issue in UTs,

npm/iptm/iptm.go
```diff
@@ -80,27 +83,125 @@ func NewIptablesManager(exec utilexec.Interface, io ioshim) *IptablesManager {
...
+		if err := iptMgr.deleteChain(chain); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Add adds a rule in iptables.
+func (iptMgr *IptablesManager) Add(entry *IptEntry) error {
```
vakalapa, line 147: Okay, so if my understanding is correct, we wont be adding a lock to IPTM but instead leverage the locking mechanism we have in place for IPTables to help add rules, i am a bit uncomfortable with this strategy, i believe this is going to put more pressure on the locking mechanism of Iptables. NPM uses os/exec package to spin up child processes to call iptable cmds and we know that NPM's control plane is lightening fast compared to dataplane, if we do not lock in NPM code and let iptables lock take care of the items, we might end up with a LOT of child processes waiting to acquire the lock from iptables, wdyt ?

vakalapa (commented):
Can you confirm if my understanding of iptm lock is correct.

npm/npm.go
```diff
@@ -106,6 +155,7 @@ func GetAIMetadata() string {
 }
 
 // SendClusterMetrics :- send NPM cluster metrics using AppInsights
+// QUESTION(jungukcho): Where is this function called?
```
vakalapa, line 158: This function is important for us to send periodic metrics about cluster size to Prometheus. Looks like it got removed in one of the previous PRs which i think is a mistake and we should be calling a go routine on this for periodic updates.

https://github.com/Azure/azure-container-networking/commit/d1699290486d3b2faa7e79bbae7aa5bca2830857#diff-22ef58b44b664ba24dd77fa2feeb41115c5b9b290f7406c4c642c38ba44cc9feL78

vakalapa: These metrics are being sent back to our local AITelemetry to log information on the cluster. I think we can safely remove these functions.

npm/http/server/server.go
JungukCho: Thank you for copying it. 

Sure. After reading Quang's PR, I realize it relies on NPM cache information.
I will see first how nicely expose NPM cache information to external components to minimize costs at controller side (e.g., lock cost)

npm/npm.go
```diff
@@ -127,178 +177,52 @@ func (npMgr *NetworkPolicyManager) SendClusterMetrics() {
...
+		nwPolicyCount.Value += float64(lenOfRawNpMap)
+
 		podCount.Value = 0
-		//Reducing one to remove all-namespaces ns obj
-		nsCount.Value = float64(len(npMgr.NsMap) - 1)
-		nwPolicyCount.Value += float64(len(npMgr.RawNpMap))
-		podCount.Value += float64(len(npMgr.PodMap))
-		npMgr.Unlock()
+		lenOfPodMap := npMgr.podController.lengthOfPodMap()
+		podCount.Value += float64(lenOfPodMap)
```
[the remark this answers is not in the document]
JungukCho: Good catch. Just for consistency with others.
```go
		lenOfPodMap := npMgr.podController.lengthOfPodMap()
		podCount.Value += float64(lenOfPodMap)
```

npm/npm.go
```diff
@@ -127,178 +177,52 @@ func (npMgr *NetworkPolicyManager) SendClusterMetrics() {
 
 	for {
 		<-heartbeat
-		npMgr.Lock()
+
+		// Reducing one to remove all-namespaces ns obj
+		// TODO(jungukcho) Check this is safe or not?
```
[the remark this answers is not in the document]
JungukCho: It only uses this place.  So, I don't think it needs method for this and it is better to directly use the variable.
I used method for two other cache (pod, netpol) to indicate they can be isolated later while all of them are in the same package now.

Let me know if you have another opinion.

npm/networkPolicyController.go
```diff
@@ -61,6 +71,35 @@ func NewNetworkPolicyController(npInformer networkinginformers.NetworkPolicyInfo
 	return netPolController
 }
 
+// initializeDataPlane do all initialization tasks for data plane
+func (c *networkPolicyController) initializeDataPlane() error {
```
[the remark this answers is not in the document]
JungukCho: Yep. "resetDataPlane" is more accurate.

npm/networkPolicyController.go
```diff
@@ -61,6 +71,35 @@ func NewNetworkPolicyController(npInformer networkinginformers.NetworkPolicyInfo
...
+	klog.Infof("Initiailize data plane. Clean up Azure-NPM chians and start reconcile iptables")
+
+	// TODO(jungukcho): will clean-up error handling codes to initialize iptables and ipset in a separate PR
+	// It is important to keep order to clean-up iptables and ipset.
+	// IPtables should be cleaned first to avoid failures to clean-up iptables due to "ipset is using in kernel" error
+	// 1. clean-up NPM-related iptables information and then running periodic processes to keep iptables correct
+	c.iptMgr.UninitNpmChains()
```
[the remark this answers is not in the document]
JungukCho: Good suggestion, but I would like to use **utilruntime.HandleError** function for log instead of **SendErrorLogAndMetric**.
```go
        if err := c.iptMgr.UninitNpmChains(); err != nil {
		utilruntime.HandleError(fmt.Errorf("Failed to UninitNpmChains with err: %s", err))
	}

	// 2. then clean-up all NPM ipsets states
	if err := c.ipsMgr.DestroyNpmIpsets(); err != nil {
		utilruntime.HandleError(fmt.Errorf("Failed to DestroyNpmIpsets with err: %s", err))
	}
```

I think using "utilruntime.HandleError" is fine since the error message is sent in "UninitNpmChains()" function already. 
Further, I do not want to use "metric.SendErrorLogAndMetric" function more in NPM code except for really critical error,  especially when the message is duplicated since it is an expensive operation (it uses network to send messages).

We need to re-organize error messages and error handling parts in concise and simple way.
Let me know if you have other opinion.

npm/networkPolicyController.go
```diff
@@ -61,6 +71,35 @@ func NewNetworkPolicyController(npInformer networkinginformers.NetworkPolicyInfo
...
+
+	// 2. then clean-up all NPM ipsets states
+	c.ipsMgr.DestroyNpmIpsets()
```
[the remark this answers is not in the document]
JungukCho: Please check above comment

npm/podController.go
```diff
@@ -252,34 +246,19 @@ func (c *podController) deletePod(obj interface{}) {
...
+func (c *podController) Run(stopCh <-chan struct{}) {
 	defer utilruntime.HandleCrash()
 	defer c.workqueue.ShutDown()
 
-	klog.Infof("Starting Pod %d worker(s)", threadiness)
-	for i := 0; i < threadiness; i++ {
-		go wait.Until(c.runWorker, time.Second, stopCh)
-	}
+	klog.Infof("Starting Pod worker")
+	go wait.Until(c.runWorker, time.Second, stopCh)
```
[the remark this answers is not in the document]
JungukCho: I think that was what I want sine it is a well-defined pattern in controller, but in the our meeting, this was suggested to prevent unexpected situation.

npm/npm.go
```diff
@@ -31,21 +29,22 @@ import (
...
-	restoreRetryWaitTimeInSeconds = 5
-	restoreMaxRetries             = 10
-	backupWaitTimeInSeconds       = 60
-	telemetryRetryTimeInSeconds   = 60
-	heartbeatIntervalInMinutes    = 30
-	reconcileChainTimeInMinutes   = 5
-	// TODO: consider increasing thread number later when logics are correct
-	threadness = 1
+	// QUESTION(jungukcho): telemetryRetryTimeInSeconds is not used. Where can we use this const telemetryRetryTimeInSeconds?
+	telemetryRetryTimeInSeconds = 60
```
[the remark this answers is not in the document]
JungukCho: Removed in another PR #952

npm/nameSpaceController.go
JungukCho: As we discussed, conceptually this approach has multiple advantages while we need measured number.
1. Can concurrently run event handlers (e.g., add, update, clean) and reconcile function since event handler does not need a lock.
2. Duplicated events will be collapsed into one in workqueue. So, it will get advantages when bulky events come since anyway, it will check last event object.

npm/nameSpaceController.go
JungukCho: same with above comment.

npm/ipsm/ipsm.go
JungukCho: **createList** is called in multiple places (e.g., . AddToList) So, it is neccessary to keep it in **createList** function. 
I also updated all UTs with this change and all passed, but I will check it again.

npm/iptm/iptm.go
JungukCho: We do not need a lock to program iptables since only networkPolicyController use iptables.
So, all of them are sequential and it is safe.

Let me know if you miss something.

npm/npm.go
JungukCho: Good catch.  
I think it is good to keep those numbers. Interval is 30min and I don't think it is heavy operations, but we can get useful metrics. I will add the removed code again. 
Let me know  if you have another opinion.

Later, we need to consolidate all metrics-related codes in metric packages in a nice way. 
Let's put it as future item.

npm/networkPolicyController.go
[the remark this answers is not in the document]
JungukCho: The **UninitNpmChains** assumes there are already AZURE-NPM chains and if it does not exist, it returns error.
```go
   if err := c.iptMgr.UninitNpmChains(); err != nil {
		utilruntime.HandleError(fmt.Errorf("Failed to UninitNpmChains with err: %s", err))
	}

	// 2. then clean-up all NPM ipsets states
	if err := c.ipsMgr.DestroyNpmIpsets(); err != nil {
		utilruntime.HandleError(fmt.Errorf("Failed to DestroyNpmIpsets with err: %s", err))
	}
```

I want to keep this way in this PR and resolve this issue in a next PR since there are multiple places where two functions and used and it needs to walk through them carefully. In addition, considering the size of this PR, I do not want to increase review burden in this PR, which will reduce productivity.
Let me know if you have another opinion.

npm/http/server/server.go
vakalapa: agreed, this is an optimization we can do later, at this moment i am okay with a primitive approach.

npm/networkPolicyController.go
```diff
@@ -61,6 +69,36 @@ func NewNetworkPolicyController(npInformer networkinginformers.NetworkPolicyInfo
 	return netPolController
 }
 
+// initializeDataPlane do all initialization tasks for data plane
+// TODO(jungukcho) Need to refactor UninitNpmChains since it assumes it has already AZURE-NPM chains
+func (c *networkPolicyController) resetDataPlane() error {
+	klog.Infof("Initiailize data plane. Clean up Azure-NPM chians and start reconcile iptables")
```
vakalapa, line 75: <nit> chains spelling

npm/networkPolicyController.go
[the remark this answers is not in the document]
vakalapa: Okay, for time being this is good, I think it is okay to ignore the error in this stage, i do not see a error case which can be fatal in NPMs operation later on.

npm/podController.go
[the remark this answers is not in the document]
vakalapa: Lets put a pin on this and discuss in next meeting.

npm/nameSpaceController.go
```diff
@@ -85,17 +76,18 @@ type nameSpaceController struct {
 	nameSpaceLister       corelisters.NamespaceLister
 	nameSpaceListerSynced cache.InformerSynced
```
vakalapa, line 77: I do not see a place where we are using this has synced variable, can we remove it if not required.

npm/nameSpaceController.go
```diff
@@ -367,14 +335,11 @@ func (nsc *nameSpaceController) syncAddNameSpace(nsObj *corev1.Namespace) error
...
-	klog.Infof(
-		"NAMESPACE UPDATING:\n namespace: [%s/%v]",
-		newNsName, newNsLabel,
-	)
+	klog.Infof("NAMESPACE UPDATING:\n namespace: [%s/%v]", newNsName, newNsLabel)
 
-	// If orignal AddNamespace failed for some reason, then NS will not be found
-	// in nsMap, resulting in retry of ADD.
-	curNsObj, exists := nsc.npMgr.NsMap[newNsName]
+	// If previous syncAddNameSpace failed for some reasons before caching npm namespace object or syncUpdateNameSpace is called due to namespace creation event,
```
vakalapa, line 340: Please apply this golint

vakalapa (commented):
Still going through controllers, no major issues until now, can you please apply all linting comments

rbtr:
@JungukCho (and @vakalapa) i wonder what you think of the `goerr113` and `wrapcheck` lints (since you have several here)?

npm/nameSpaceController.go
JungukCho: Good catch.
NPM checks sync of all controller in npm.go - https://github.com/Azure/azure-container-networking/blob/master/npm/npm.go#L151.  So good to delete this in all controllers.

JungukCho: Oh, it is used in each controller UT, but I am not sure it is necessary yet. 
I will check it.  If it does not have issue, I will delete them.

JungukCho:
> @JungukCho (and @vakalapa) i wonder what you think of the `goerr113` and `wrapcheck` lints (since you have several here)?

@rbtr I think it is good to gradually remove lints warning if they make senses. I think they follows best practice. However, it is impossible to move it as one step. 

1. I like `wrapcheck`  warning 
- As far as I understood, it helps point out where the error comes from in log message!

2. err123 - It seems two advantages. 
- Can compare error values directly. So it seems it is useful when we need to do something based on error values.
- A little fast (?) by avoiding create error every time
However, I am not sure, in case we do not need value check and it used only one place, it is good to create static (package-level) error.

What do you think?

rbtr:
I agree with all your points. I will add that `wrapcheck` requires _all_ external errors be wrapped and that might be tedious if we are not used to it.

PR #952, opened 2021-07-23
[NPM] Remove unused codes to save and restore ipset and iptables

JungukCho:
**Reason for Change**:

There are a bug in Save and Restore functions in iptm.go. It does not store iptables information into file with the error message "exec: Stdout already set"

As the similar issue exists in Restore function in iptm.go. 
The first commit fixed the issues (in save and restore of iptables) in this PR.

However, we all agree to delete those codes since it is not necessary to save iptables information now and saving iptables information has high overhead.
So, I clean-up related save and restore codes for iptables and ipset in following commits.

**.Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


JungukCho:
\azp run

vakalapa (approved):

PR #953, opened 2021-07-23
[NPM] Remove duplicate namedports in slices to avoid unnecessary CreateSet calls in network policy controller

JungukCho:
**Reason for Change**:

translatePolicy function returns duplicated ipsets for namedPort if duplicated ipsets exist, which causes unnecessary call in network policy controller. While the duplicated ipset is filtered in network policy controller, it is the best to avoid the call.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/translatePolicy.go
+++ b/npm/translatePolicy.go
@@ -1700,10 +1700,9 @@ func translatePolicy(npObj *networkingv1.NetworkPolicy) ([]string, []string, map
 	}
 
 	entries = append(entries, getDefaultDropEntries(npNs, npObj.Spec.PodSelector, hasIngress, hasEgress)...)
-	resultSets = util.UniqueStrSlice(resultSets)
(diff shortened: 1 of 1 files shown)


vakalapa (approved):
Lgtm, not sure how we missed this earlier. Worst case scenario we will wait a couple of ipset create calls i think. good to catch

matmerr (approved):
approving with the caveat that this change is a bandaid fix, the parent issue of adding duplicates requires a much larger refactor and is actively being discussed how to fix in translatepolicy

JungukCho:
> approving with the caveat that this change is a bandaid fix, the parent issue of adding duplicates requires a much larger refactor and is actively being discussed how to fix in translatepolicy

Correct. I am working on refactoring translatepolicy.go with determinisitic UTs together in #941.

PR #954, opened 2021-07-27
chore: typed response codes

rbtr:
**Reason for Change**:

Upgrade the response codes from raw int to a first-class `ResponseCode` type. Implicitly guarantees that the codes being returned are a known response code, since returns must be correctly typed to compile.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/cnsclient/error_test.go
```diff
@@ -0,0 +1,46 @@
...
+			err: &CNSClientError{
+				Code: types.UnexpectedError,
+				Err:  fmt.Errorf("unexpected err"),
+			},
+			want: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
ramiro-gamarra, line 40: If you are running tests in parallel, you shouldn't close over the loop variable

cns/types/codes.go
```diff
@@ -0,0 +1,82 @@
+package types
```
ramiro-gamarra, line 1: `types` seems like a very generic package name. These seem like part of the API contract, is there a better home for them?

rbtr: i original called the package `response` and that stutter wasn't good either. i think it wants to be with the code in the `restserver` package, but actually putting it in that package has some potential cyclical dependency problems.

cns/cnsclient/error_test.go
rbtr: should close over the loop variable, will fix. 
@ramiro-gamarra ty, this led me to discover [this](https://github.com/Azure/azure-container-networking/pull/956)

cns/types/codes.go
ramiro-gamarra: I figured cyclical dependencies had something to do with the choice to extract this out. As a CNS client, DNC also consumes these response codes, so this change will require refactoring from that end as well. From that perspective, the `cns` package itself already contains the symbols used to work with the http api, is that an option?

rbtr: it really doesn't work in any of the existing packages, import cycles 🙃
i'm on board with the root `cns/` package holding all of the contract, but it has more than just that contract at the moment, and that causes a big mess.

neaggarwMS (dismissed):
:clock1:

cns/service.go
```diff
@@ -57,7 +57,9 @@ func (service *Service) Initialize(config *common.ServiceConfig) error {
 	log.Debugf("[Azure CNS] Going to initialize a service with config: %+v", config)
 
 	// Initialize the base service.
-	service.Service.Initialize(config)
+	if err := service.Service.Initialize(config); err != nil {
```
neaggarwMS, line 60: This is a good change.

cni/network/network.go
```diff
@@ -912,7 +911,7 @@ func (plugin *netPlugin) Delete(args *cniSkel.CmdArgs) error {
 	if err != nil {
 		log.Printf("[cni-net] Failed to extract network name from network config. error: %v", err)
 
-		if !utils.IsNotFoundError(err) {
+		if !cnsclient.IsNotFound(err) {
```
neaggarwMS, line 914: >cnsclient [](http://example.com/codeflow?start=6&length=9)

Not every scenario has cnsClient. It is not used for AKS single tenancy.

cni/utils/errutils.go
```diff
@@ -1,15 +0,0 @@
-package utils
-
-import (
-	"github.com/Azure/azure-container-networking/cns/cnsclient"
-	"github.com/Azure/azure-container-networking/cns/restserver"
-)
-
-// TODO : Move to common directory like common, after fixing circular dependencies
```
neaggarwMS, line 8: >ove to common directory like common, after fixing circular dependencies [](http://example.com/codeflow?start=11&length=71)

what is the circular dependency? @tamilmani1989 do you know?

cns/service.go
tamilmani1989: add log also?

cni/utils/errutils.go
tamilmani1989: @rbtr correct me if im wrong. I think cns restserver package has dependency on cni common and this code cannot be moved to common because of that.  imo, this code shouldn't be in common and should be in cnsclient package. why didn't we put it there?

tamilmani1989 (dismissed):

cni/network/network.go
tamilmani1989: the right way is to have shim (one more layer) here for getnetworkname

cni/utils/errutils.go
rbtr: @tamilmani1989, I deleted this file, I didn't write the comment so I don't know what the original reason for putting this here was. I have moved this functionality to the `cnsclient` package because that's where the `CNSClientError` struct is

cni/network/network.go
ramiro-gamarra: there is no further dependency taken here (see imports). the check is appropriate because the only error handled seems to have always been a CNS specific error.

cns/types/codes.go
ramiro-gamarra: In that case, maybe call this package `api` and leave a todo to move the rest of the public contract here? I think the underlying problem really is the pervasiveness of these return codes in different layers where proper errors should be used, but that is a larger fix for another day.

rbtr: I agree and I can move it, but the whole public contract should eventually get moved _back_ to the root `cns` package, imo

ramiro-gamarra: I can agree with that if the goal is to have the `cns` package be very lean and not have extraneous imports for consumers :).

cni/network/network.go
tamilmani1989: cnsclient package is specific to swift singletenancy and this function looks like common for both swift and traditional. if so, then it should be moved to common package imo.

rbtr: broad common or util packages are an antipattern in go, this function deals only with the `CNSClientError` so it should live in the same file as the `CNSClientError` (which should live next to the `CNSClient` since that's the only thing that returns it).
that cnsclient is a poor choice of package name, and that that client is initialized as a global with package side-effects are both bad things we can argue, but are out of scope on this change.

tamilmani1989: i went through `getnetworkname` function and looks like it returns error only in windows multitenancy scenario. we can ignore this error in getNetworkName function itself instead of checking here.. also i found even if we ignore error at this step, it will still fail in `if nwInfo, err = plugin.nm.GetNetworkInfo(networkId); err != nil {` and return without any useful action. may be its not scoped to this PR, but we should invoke cns here to cleanup based on containerid as like traditional case.

tamilmani1989: i will take back my comment on checking in getNetworkName..that doesn't sound right but can we add a check in isnotfound function to do this only in case of multitenancy

thatmattlong: I think it's safer to just check the error type regardless of the scenario. No need to do a scenario check to try to avoid a single if statement that will evaluate to false. If we started doing scenario checks, the code will get really complicated when we add more and more scenarios.

cns/types/codes.go
thatmattlong: +1 on this eventually going back to `cns`. Types and interfaces should be defined in `cns` root, implementations of those interfaces should exist in subpackages that are named according to the implementation. For now an `api` package is probably fine to alleviate the cyclical imports, but usually cyclical imports means the package design is all wrong, which probably requires a much larger PR to fix.

thatmattlong (approved):

rbtr:
failing test is unrelated and was introduced in https://github.com/Azure/azure-container-networking/pull/960#issuecomment-894521249

PR #958, opened 2021-07-28
[NPM] Print running NPM version in stdout.

JungukCho:
**Reason for Change**:

This PR is to print running NPM version in stdout which can help diagnose errors or know tested version (e.g., which version is used for testing).
In addition, this PR replace log with klog for better stdout format in npm.go and main.go

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/plugin/main.go
+++ b/npm/plugin/main.go
@@ -73,7 +73,7 @@ func main() {
 	factor := rand.Float64() + 1
 	resyncPeriod := time.Duration(float64(minResyncPeriod.Nanoseconds()) * factor)
 
-	klog.Info("Resync period for NPM pod is set to %d.", int(resyncPeriod/time.Minute))
+	klog.Infof("Resync period for NPM pod is set to %d.", int(resyncPeriod/time.Minute))
 	factory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
 
 	npMgr := npm.NewNetworkPolicyManager(clientset, factory, exec.New(), version)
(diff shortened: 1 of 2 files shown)


matmerr:
maybe with the cli tools we should add a `version` command

JungukCho:
> maybe with the cli tools we should add a `version` command

That's a good idea.

npm/plugin/main.go
```diff
@@ -71,17 +73,17 @@ func main() {
 	factor := rand.Float64() + 1
 	resyncPeriod := time.Duration(float64(minResyncPeriod.Nanoseconds()) * factor)
 
-	log.Logf("[INFO] Resync period for NPM pod is set to %d.", int(resyncPeriod/time.Minute))
+	klog.Infof("Resync period for NPM pod is set to %d.", int(resyncPeriod/time.Minute))
 	factory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
 
 	npMgr := npm.NewNetworkPolicyManager(clientset, factory, exec.New(), version)
-	metrics.CreateTelemetryHandle(npMgr.GetAppVersion(), npm.GetAIMetadata())
+	metrics.CreateTelemetryHandle(version, npm.GetAIMetadata())
```
vakalapa, line 80: Can you fix this lint please, apart from that the PR looks good.

vakalapa (approved):
lgtm

npm/plugin/main.go
JungukCho: As we agreed, it will come with another PR about metric.

PR #962, opened 2021-08-06
Fix CNS Scale Up/down fix to avoid cleanup pending ToBeDeleted ips

neaggarwMS:
There is a current bug in IPAM pool monitor where if a scale up happens right after scale down then we were accidently deleting the IpsNotInUse list in case Scale up Requested count is same as Total current ip count in CNS. 

As a result DNC was not honoring the scale down as well not allocating new IPs because the existing secondary ip count matches the updated requested count.


cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -244,28 +241,22 @@ func (pm *CNSIPAMPoolMonitor) cleanPendingRelease(ctx context.Context) error {
 
 	// save the updated state to cachedSpec
 	pm.cachedNNC.Spec = tempNNCSpec
-	pm.pendingRelease = false
 	return nil
 }
 
 // CNSToCRDSpec translates CNS's map of Ips to be released and requested ip count into a CRD Spec
-func (pm *CNSIPAMPoolMonitor) createNNCSpecForCRD(resetNotInUseList bool) (nnc.NodeNetworkConfigSpec, error) {
+func (pm *CNSIPAMPoolMonitor) createNNCSpecForCRD() (nnc.NodeNetworkConfigSpec, error) {
```
thatmattlong, line 248: Can you change this function signature to remove the error: `func (pm *CNSIPAMPoolMonitor) createNNCSpecForCRD() nnc.NodeNetworkConfigSpec { ... }`?

thatmattlong (changes requested):
@neaggarwMS can you fix the linter errors?

PR #964, opened 2021-08-06
ipv6 dualstack support transparent mode

tamilmani1989:
**Reason for Change**:

This PR adds support for dualstack (ipv4, ipv6) in transparent mode. The pods will have both v4 and v6 addresses on their interface. They can communicate with other pods/other VMs in vnet with both ipv4 and ipv6. All k8s services expected to work with ipv6 also (like lb, nodeport)

As like v4, this PR adds
 - routes for ipv6 pod ips in VM
 - sets up virtual ipv6 gateway in container
 - add neigh entry for virtual ipv6 gateway in container
 - v6 ips are not vnet ips so it adds snat rule to snat traffic leaving VM with secondary VM v6 ip
 - Enables ipv6 settings

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Tests are not added. We have to stub out iptables and netlink dependency and then only we can add tests for these.


ipam/ipv6Ipam.go
```diff
@@ -211,7 +212,9 @@ func (source *ipv6IpamSource) refresh() error {
 		return err
 	}
 
+	source.isLoaded = true
```
rbtr, line 215: this is subject to a race, are the side effects to be concerned about if this fn executes more than once?

tamilmani1989: this function is getting called within mutex lock.  is it still a concern?

PR #972, opened 2021-08-12
enable prometheus metrics in cns

rbtr:
**Reason for Change**:

Enables prometheus metrics in CNS via the controller-runtime, adds initial histovecs on the IP request API endpoints

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/ipampoolmonitor/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	ipamAllocatedIPCount = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "ipam_allocated_ip",
```
ramiro-gamarra, line 11: Might be a good idea to adhere to Prometheus naming conventions [here](https://prometheus.io/docs/practices/naming/). If we do, units should be plural where appropriate.

cns/ipampoolmonitor/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+	)
+	ipamAvailableIPCount = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "ipam_available_ip",
+			Help: "Available IP count.",
+		},
+	)
+	ipamBatchSize = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "ipam_batch",
```
ramiro-gamarra, line 23: thoughts on `size` as a suffix? Same for `pool` below.

cns/metric/alloc.go
```diff
@@ -0,0 +1,111 @@
...
+// and we queue one right before receiving an NNC status update, we have no
+// way to decorrelate that update with the in-flight requsets and will record
+// a very short response latency.
+
+var allocLatency = prometheus.NewHistogramVec(
+	prometheus.HistogramOpts{
+		//nolint:gomnd
+		Buckets: prometheus.ExponentialBuckets(0.05, 2, 15), // 50 ms to ~800 seconds
+		Help:    "IP allocation latency in seconds by batch size",
+		Name:    "ip_alloc_latency",
```
ramiro-gamarra, line 38: Per naming conventions, should probably include the unit as suffix (`seconds`). Same for below.

cns/metric/alloc.go
```diff
@@ -0,0 +1,111 @@
...
+// request via the NNC, and then popping any alloc/dealloc start times that
+// we have saved as soon we get an updated NNC.
+// If we have not received a new NNC, multiple requests to start an
+// alloc/dealloc timer will noop - we only record the longest span between
+// pushing an NNC spec and receiving an updated NNC status.
+//
+// This data will be left-skewed: if there is no in-flight alloc or dealloc,
+// and we queue one right before receiving an NNC status update, we have no
+// way to decorrelate that update with the in-flight requsets and will record
+// a very short response latency.
```
ramiro-gamarra, line 31: This is amazing documentation, thank you.

cns/metric/alloc.go
```diff
@@ -0,0 +1,111 @@
...
+// We get this data by recording a start time when we push an alloc/dealloc
+// request via the NNC, and then popping any alloc/dealloc start times that
+// we have saved as soon we get an updated NNC.
+// If we have not received a new NNC, multiple requests to start an
+// alloc/dealloc timer will noop - we only record the longest span between
+// pushing an NNC spec and receiving an updated NNC status.
+//
+// This data will be left-skewed: if there is no in-flight alloc or dealloc,
+// and we queue one right before receiving an NNC status update, we have no
+// way to decorrelate that update with the in-flight requsets and will record
```
ramiro-gamarra, line 30: typo on requests

cns/restserver/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var requestLatency = prometheus.NewHistogramVec(
+	prometheus.HistogramOpts{
+		Name: "request_latency",
```
ramiro-gamarra, line 13: same as above re unit suffix

cns/restserver/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+	metrics.Registry.MustRegister(
+		requestLatency,
+	)
+}
+
+func newHandlerFuncWithHistogram(handler http.HandlerFunc, histogram *prometheus.HistogramVec) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		start := time.Now()
+		defer func() {
+			histogram.WithLabelValues(req.URL.RequestURI(), req.Method, "0").Observe(time.Since(start).Seconds())
```
ramiro-gamarra, line 31: What is the `"0"` here?

cns/singletenantcontroller/kubecontroller/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+	)
+	requestedIPs = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "requested_ips",
+			Help: "Requested IP count.",
+		},
+	)
+	unusedIPs = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "unused_ip",
```
ramiro-gamarra, line 23: `ips`

cns/restserver/restserver.go
```diff
@@ -57,7 +57,7 @@ type GetHTTPServiceDataResponse struct {
 	Response            Response
 }
 
-//struct to return in-memory httprest data in debug api
+// HttpRestServiceData struct to return in-memory httprest data in debug api
```
ramiro-gamarra, line 60: This removes lint warnings but we can do better :). How about
```
HttpRestServiceData is a response from the debug API. It represents in-memory CNS data.
```

ramiro-gamarra (commented):
This looks really good, just some feedback around metric naming.

cns/ipampoolmonitor/metrics.go
rbtr: works for me

cns/ipampoolmonitor/metrics.go
rbtr: `ip`->`ips` everywhere, since that seems to by my unit here 😆

cns/metric/alloc.go
rbtr: updated

cns/restserver/metrics.go
rbtr: dropped. it was supposed to represent response code, but the current router setup doesn't let me easily extract that in a middleware func like this

cns/singletenantcontroller/kubecontroller/crdrequestcontroller.go
```diff
@@ -116,20 +114,20 @@ func New(cfg Config) (*requestController, error) {
 	// for serving prometheus metrics, set to "0" to disable
 	mgr, err := ctrl.NewManager(cfg.KubeConfig, ctrl.Options{
 		Scheme:             scheme,
-		MetricsBindAddress: prometheusAddress,
+		MetricsBindAddress: ":9090",
```
ramiro-gamarra, line 117: missed this before, but should this come from config?

rbtr: maybe. `:9090` is the standard prometheus scrape port, and i don't think we will have any reason to change it, but i'm not against making that a knob with `:9090` as an overridable default

ramiro-gamarra: I think it would be better for orchestrator teams other than AKS (which may or may not run CNS as a container), such that they don't end up with any port collisions.

rbtr: i will add this as default so that it can be disabled with a config change.
just to note: as currently implemented, I'm using controller-runtime's prometheus integration so this codepath only gets exercised if we're running in k8s

cns/configuration/configuration.go
```diff
@@ -133,6 +134,7 @@ func SetCNSConfigDefaults(config *CNSConfig) {
 	if config.ChannelMode == "" {
 		config.ChannelMode = cns.Direct
 	}
+	config.MetricsBindAddress = ":9090"
```
ramiro-gamarra, line 137: It seems like this would always overwrite the address even if provided

rbtr: embarrassing

ramiro-gamarra (dismissed):
lgtm. I know that @neaggarwMS also wanted to review so I'd wait until she approves to merge.

PR #975, opened 2021-08-16
Consolidate CRD packages

rbtr:
**Reason for Change**:

Relocates the NNC and multitenant CRD in to a `crd/` package to consolidate current (and future) shared CRD related code and components.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


feiskyer:
I think it's better to keep the root package name as "crds" instead of "crd" because multiple CRDs are added there. What do you think?

rbtr:
@feiskyer I went back and forth on this. At first I agree that it reads better as `crds`, but the [consensus in the community](https://github.com/golang-standards/project-layout/issues/7) seems to be that packages of "collections", several of the same type of thing like CRDs, should be singular `crd`. 
Seems like the only reason we have `bytes` or `errors` in the stdlib is because those would conflict with reserved keywords `byte` and `error`.
@ramiro-gamarra wdyt?

ramiro-gamarra:
@rbtr It looks like in this new structure `crd` is indeed a package and not just a directory so I would also favor the singular form.

crd/nodenetworkconfig/api/v1alpha/groupversion_info.go
```diff
@@ -1,19 +1,3 @@
-/*
```
thatmattlong, line 1: I am by no means a legal expert but do we want to keep the licenses of generated files as the license that they were generated under?

thatmattlong (approved):
Approved, but with a question about removing some of the license headers

crd/nodenetworkconfig/api/v1alpha/groupversion_info.go
rbtr: These files are generated by kubebuilder as apache2 by default, but that's customizable and we should have overriden it when we scaffolded the CRDs. Since currently that's the default and these are the only files not covered by the root MIT license, and that seemed accidental. Removing the headers covers them under MIT with everything else.

PR #977, opened 2021-08-20
fix: Check if systemd-resolved is running before copying dns servers

jaer-tsun:
**Reason for Change**:

Today for multitenancy on Ubuntu 18, we assume systemd-resolved is running, query the dns servers on eth0 and move them to azure0 when we create the bridge. The problem for privatecluster is that they are using ubuntu 18 but disabling systemd-resolved. So instead of assuming it is running for Ubuntu 18, we should instead check the service status, and if it is not running, don't copy the dns servers from eth0 to azure0 (basically just do what we do for ubuntu 16 today).

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -407,11 +407,16 @@ func (nm *networkManager) connectExternalInterface(extIf *externalInterface, nwI
 		If custom dns server is updated, VM needs reboot for the change to take effect.
 	*/
 	isGreaterOrEqualUbuntu17 := isGreaterOrEqaulUbuntuVersion(ubuntuVersion17)
+	isSystemdResolvedActive := false
 	if isGreaterOrEqualUbuntu17 {
-		log.Printf("[net] Saving dns config from %v", extIf.Name)
-		if err = saveDnsConfig(extIf); err != nil {
-			log.Printf("[net] Failed to save dns config: %v", err)
-			return err
+		// Don't copy dns servers if systemd-resolved isn't available
+		if _, cmderr := platform.ExecuteCommand("systemctl status systemd-resolved"); cmderr == nil {
+			isSystemdResolvedActive = true
+			log.Printf("[net] Saving dns config from %v", extIf.Name)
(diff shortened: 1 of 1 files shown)


network/network_linux.go
```diff
@@ -408,10 +408,13 @@ func (nm *networkManager) connectExternalInterface(extIf *externalInterface, nwI
 	*/
 	isGreaterOrEqualUbuntu17 := isGreaterOrEqaulUbuntuVersion(ubuntuVersion17)
 	if isGreaterOrEqualUbuntu17 {
-		log.Printf("[net] Saving dns config from %v", extIf.Name)
-		if err = saveDnsConfig(extIf); err != nil {
-			log.Printf("[net] Failed to save dns config: %v", err)
-			return err
+		// Don't copy dns servers if systemd-resolved isn't available
+		if _, cmderr := platform.ExecuteCommand("systemctl status systemd-resolved"); cmderr == nil {
```
rbtr, line 412: instead of shelling out to systemctl, should we use [dbus](https://github.com/coreos/go-systemd/blob/main/dbus/methods.go#L469)?

tamilmani1989 (changes requested):
we have to add a check where we apply dns config also.

network/network_linux.go
tamilmani1989: what happens if systemd-resolved is available but not running? should we do systemctl check systemd-resolved and verify if its active?

rbtr: dbus would give us that granularity of information :)

jaer-tsun: [this](http://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic.html#INISCRPTACT) should be good enough in our case (checking whether systemd-resolved is active or not)

tamilmani1989 (approved):

PR #979, opened 2021-08-24
feat: [NPM] Add config file and feature toggles

matmerr:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


npm/cmd/main.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func main() {
+	klog.Infof("Start NPM version: %s", version)
+	cobra.CheckErr(rootCmd.Execute())
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+	rootCmd.Flags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.azure-npm-debug-cli.yaml)")
```
rbtr, line 58: help string is misleading, maybe:
```suggestion
	rootCmd.Flags().StringVar(&cfgFile, "config", "", fmt.Sprintf("config file (default is %s)", npmconfig.GetConfigPath()))
```

npm/cmd/main.go
```diff
@@ -0,0 +1,80 @@
...
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "azure-npm",
+	Short: "Collection of functions related to Azure NPM's debugging tools",
+
+	RunE: func(cmd *cobra.Command, args []string) error {
+		config := &npmconfig.Config{}
+		err := viper.Unmarshal(config)
+		if err != nil {
+			fmt.Printf("unable to decode into config struct, %v", err)
```
rbtr, line 32: since this is a `RunE`, probably better to wrap and return

npm/cmd/main.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		Start(*config)
```
rbtr, line 35: similarly, this func should block or return an error, and we should `return Start(...)`

npm/cmd/main.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/spf13/viper"
+	"k8s.io/klog"
+)
+
+// Version is populated by make during build.
+var version string
+var cfgFile string
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
```
rbtr, line 24: very subjective, but i prefer to put the root command in a `root.go`, and keep `main.go` very lean, with only a `main()` and the min cobra imports

npm/cmd/debugcmd.go
```diff
@@ -1,4 +1,4 @@
-package cmd
+package main
```
rbtr, line 1: the package is cmd, isn't it? and we can drop the "cmd" from this filename.

npm/util/errors/errors.go
```diff
@@ -1,12 +1,18 @@
...
 import (
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/Azure/azure-container-networking/npm/util"
 )
 
+var (
+	SrcNotSpecified = errors.New("source not specified")
```
rbtr: should these be in the package where they get used instead of a `util`?

rbtr (commented):
few organizational nitpicks, ❤️ consolidating on cobra and a proper config pkg

vakalapa (approved):
Lgtm 🚀

PR #984, opened 2021-08-30
[NPM] Generic Dataplane interface for windows and linux

vakalapa:
**Reason for Change**:

This PR helps outline the DataPlane interface which acts as an abstraction layer for similar functionalities in Windows and Linux.

a. Addition of new dataplane APIs for control plane to use
b. specific build target functions to convert ipsets and acl policies accoridngly

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


npm/pkg/dataplane/dataplane.go
```diff
@@ -0,0 +1,134 @@
...
+
+type OsType string
+
+const (
+	Windows OsType = "windows"
+	Linux   OsType = "linux"
+)
+
+type DataPlaneInterface interface {
+	NewDataplane() (*DataPlane, error)
```
huntergregory, line 36: would this not be in the interface because it's a function instead of a method?

vakalapa: You are right, will correct that

rbtr: @vakalapa that you accidentally included the `New` in the interface kinda points at a problem with this interface: it's very big but also very specific. This hasn't _really_ decoupled anything, you've just moved where the coupling happens.
I noticed that you have naturally grouped the methods in the interface, which might indicate that those groups are the true interfaces. You could break this up in to smaller and easier to swap out interfaces and more well isolated implementations.

vakalapa: Agreed, there is not strict reason to use this interface apart from being a easily way to discover dataplane APIs for some body working on the control plane side. I will remove it as soon as this PR hits some maturity.

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,191 @@
...
+func (set *IPSet) AddNetPolReference(netPolName string) {
+	set.NetPolReference[netPolName] = struct{}{}
+}
+
+func (set *IPSet) DeleteNetPolReference(netPolName string) {
+	delete(set.NetPolReference, netPolName)
+}
+
+func (set *IPSet) CanBeDeleted() bool {
+	if len(set.SelectorReference) > 0 {
```
huntergregory, line 165: this could be written as 
`return len(a) == 0 && len(b) ==0 && ...`
Go will shortcircuit (evaluate from left to right and stop on the first false)

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (set *IPSet) IncIpsetReferCount() {
+	set.IpsetReferCount++
+}
+
+func (set *IPSet) DecIpsetReferCount() {
+	set.IpsetReferCount--
+}
+
+func (set *IPSet) AddSelectorReference(netPolName string) {
```
huntergregory, line 148: different arg name (below func too)?

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,191 @@
...
+		return false
+	}
+	if len(set.IPPodKey) > 0 {
+		return false
+	}
+	return true
+}
+
+func (set *IPSet) UsedByNetPol() bool {
+	if len(set.SelectorReference) <= 0 {
```
huntergregory, line 184: length can never be negative, so maybe == 0?

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,191 @@
...
+	ListSet SetKind = "list"
+	HashSet SetKind = "set"
+)
+
+func NewIPSet(name string, setType SetType) *IPSet {
+	set := &IPSet{
+		Name:              name,
+		HashedName:        util.GetHashedName(name),
+		Type:              setType,
+		SelectorReference: make(map[string]struct{}),
```
huntergregory, line 82: is the empty struct a placeholder? Or what does it mean?

npm/pkg/dataplane/dataplane.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (dp *DataPlane) AddToList(listName string, setNames []string) error {
+	return dp.ipsetMgr.AddToList(listName, setNames)
+}
+
+func (dp *DataPlane) RemoveFromList(listName string, setNames []string) error {
+	return dp.ipsetMgr.RemoveFromList(listName, setNames)
+}
+
+func (dp *DataPlane) UpdatePod(pod *npm.NpmPod) error {
```
huntergregory, line 91: will there be AddPod or DeletePod?

npm/pkg/dataplane/ipsets/ipset.go
rbtr: you should make the same suggestion as above! `return len(set.SelectorReference) > 0 || len(set.NetPolReference) > 0`

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,191 @@
...
+
+func GetSetType(x string) SetType {
+	return SetType(SetTypeValue[x])
+}
+
+type SetKind string
+
+const (
+	ListSet SetKind = "list"
+	HashSet SetKind = "set"
```
matmerr, line 74: tbh I think there's room for what defines a `set` here, taking linux vernacular could be misleading in windows, but I think in terms of platform agnostic, what about something like hashset is "set" and list is "superset" since it contains only "sets"(?)

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,191 @@
...
+			contents = append(contents, memberSet.HashedName)
+		}
+		return contents, nil
+	default:
+		return contents, fmt.Errorf("Unknown set type %s", setType)
+	}
+}
+
+func getSetKind(set *IPSet) SetKind {
+	switch set.Type {
```
matmerr, line 114: would it be possible to add set kind as a property of the set itself instead of this switch dictionary?

matmerr (commented):
some comments

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	return set
+}
+
+func (set *IPSet) GetSetContents() ([]string, error) {
+	contents := make([]string, 0)
+	setType := getSetKind(set)
+	switch setType {
+	case HashSet:
+		for podIP := range set.IPPodKey {
```
matmerr, line 99: also should allocate slice here with make([]string, len(set.MemberIPSets)), same below to avoid slice resizing, heh @rbtr

npm/pkg/dataplane/dataplane.go
```diff
@@ -0,0 +1,109 @@
...
+	OsType    OsType
+	networkID string
+	// key is PodKey
+	endpointCache map[string]*NPMEndpoint
+}
+
+type NPMEndpoint struct {
+	Name      string
+	ID        string
+	NetpolRef []string
```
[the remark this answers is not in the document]
vakalapa: Makes sense, corrected it. Keeping it same in both places.

npm/pkg/dataplane/dataplane.go
vakalapa: No UpdatePod, is like a trigger by controller to say, hey please update the references for this newly created pod and also check to apply any applicable network policies

npm/pkg/dataplane/ipsets/ipset.go
vakalapa: My intension was to emulate a "set" behavior, https://www.davidkaya.com/sets-in-golang/ .. so that we do not have to check for duplicates everytime.

npm/pkg/dataplane/ipsets/ipset.go
vakalapa: essentially what we are adding is the network policy name if its a selector ref or a normal ref.

npm/pkg/dataplane/dataplane.go
```diff
@@ -0,0 +1,109 @@
...
+
+type OsType string
+
+const (
+	Windows OsType = "windows"
+	Linux   OsType = "linux"
+)
+
+// Detects the OS type
+func detectOsType() OsType {
```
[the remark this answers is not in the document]
vakalapa: For sure, thats my intension too, but added this function initially thinking we may need it. Will clean it up.

npm/pkg/dataplane/ipsets/ipset.go
vakalapa: I like this "superset" since it is intuitive. But keeping list since we have been used to calling it as list. We can use "nestedset" which windows is using ?

npm/pkg/dataplane/ipsets/ipset.go
vakalapa: I was afraid that the complexity of maintaining to values which mean almost the same thing, and we will need to take care in making sure both the attributes are set always. Keeping one makes the code a but simpler to understand. Wdyt ?

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,185 @@
...
+		len(set.MemberIPSets) > 0 ||
+		len(set.IPPodKey) > 0 {
+		// return false if this IPSet has any references and memebrs.
+		return false
+	}
+	return true
+}
+
+// UsedByNetPol check if an IPSet is referred in network policies.
+func (set *IPSet) UsedByNetPol() bool {
```
huntergregory, line 179: this and the above func could still be cleaner without an if-statement e.g. 
```
func (set *IPSet) UsedByNetPol() bool {
    return len(set.SelectorReference) > 0 || len(set.NetPolReference) > 0
}
```

for the other function:
```
func (set *IPSet) CanBeDeleted() bool {
  return len(selectorref) == 0 && len(netpolref) == 0 && refcount <= 0 & len(memberipsets) == 0 && len(ippodkey) == 0
}
```

npm/pkg/dataplane/ipsets/ipset.go
huntergregory: What do we think of encapsulating set behavior in a new struct like they suggest in that article? This may be natural to experienced GoLangers, but I think it would improve readability, perhaps be less bug prone, and wouldn't be wasted energy since we use the data structure in 4 different fields. Thoughts @rbtr @matmerr @JungukCho ?
```
type set struct {
    m map[string]struct{}
}

func NewSet() *set {
    s := &set{}
    s.m = make(map[string]struct{})
}

func (s *set) Add(value string) {
    s.m[value] = exists
}

func (s *set) Remove(value string) {
    delete(s.m, value)
}

func (s *set) Contains(value string) bool {
    _, c := s.m[value]
    return c
}
```

rbtr: I'm lukewarm on a Set type,  I don't think it would simplify the code in reality and is unusual enough in go that I would then have to go find the custom type definition to see how it had been implemented.
BUT, I'm very used to using maps for this, so 🤷🏼‍♂️

npm/pkg/dataplane/policies/policymanager.go
```diff
@@ -0,0 +1,58 @@
...
+func NewPolicyManager() PolicyManager {
+	return PolicyManager{
+		policyMap: &PolicyMap{
+			cache: make(map[string]*NPMNetworkPolicy),
+		},
+	}
+}
+
+func (pMgr *PolicyManager) GetPolicy(name string) (*NPMNetworkPolicy, error) {
+	pMgr.policyMap.Lock()
```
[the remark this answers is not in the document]
vakalapa: I have to read about sync.RWMutex, i do not think we have enough leverage to move to a RWmutex, since most of our operations needs to have write lock. GetPolicy may not even end up in our fnial code. At this moment i anticipate it to be useful in network policy controller.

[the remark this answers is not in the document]
rbtr: you only need to change the embedded mutex to `sync.RWMutex` and start using `RLock` and `RUnlock` on reads. it's drop-in, no other changes need to be made to the code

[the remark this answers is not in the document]
vakalapa: Okay makes sense, let me switch it up.

npm/pkg/dataplane/ipsets/ipset.go
vakalapa: i am okay with either approach, no major preference, but i felt the Map approach keeps it simple and less cognitive load for reviewer. I understand this question of why struct may arise again, i will leave a comment explaining the same in code.

vakalapa: If you look at our vendor files apimachinery or grpc lib uses this same notion.

npm/pkg/dataplane/ipsets/ipset.go
vakalapa: Interesting, golint now complains "append to slice `contents` with non-zero initialized length (makezero)"

npm/pkg/dataplane/ipsets/ipset.go
rbtr: like @huntergregory alluded to, this should make sense to gophers immediately, it's common to see. but it's not obvious the first time you see it - a comment saying "using a set" or something would help 👍🏼

npm/pkg/dataplane/ipsets/ipset.go
rbtr: yes, it's 👎🏼 to append to a slice that you've sized already.
to be maximally efficient _and_ satisfy the linter, you should add an idx and put the key values in to their spot directly:
```go
i := 0
keys := make([]string, len(set.IPPodKey)
for podIP := range set.IPPodKey {
	keys[i] = podIP
	i++
}
```

vakalapa: for some reason this makes me uncomfortable, but yeah i see the perf improvement.

vakalapa: Now i am thinking if golint will be okay with below 
> make([]string, 0, len(set.IPPodKey)

and then appending

npm/pkg/dataplane/ipsets/ipset.go
matmerr:
```go
type SetProperties struct {
	Type string //list/set
	Kind string 
}
```

vakalapa: Pushed new changes

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,198 @@
...
+			contents[i] = memberSet.HashedName
+			i++
+		}
+		return contents, nil
+	default:
+		return []string{}, fmt.Errorf("Unknown set type %s", set.Properties.Type.String())
+	}
+}
+
+func getSetKind(setType SetType) SetKind {
```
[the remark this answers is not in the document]
vakalapa: Yes we derive the kind from type of set input.

JungukCho (commented):
Before I put other comments, there are structural (i.e., architectural) comments of this PR.
I may misunderstand the codes and intentions, but some of them are different from what I expected.

1. I thought interface will be used for `ipsetmanager` and `dataplane` to abstract linux and windows. But it seems you use composition with concrete struct instead of interface. I think both `ipsetmanager` and `dataplane` are nicely abstract with interface (Example in npm - https://github.com/Azure/azure-container-networking/blob/master/npm/iptm/ioshim.go). 
Do I know why you decide this direction? 

2. There are many almost inline method (just one or two lines) for exposed members. In this case, if you want to keep methods, I think non-capitalizing members if possible since it is good to maintain code as our code base becomes bigger and bigger.  In addition, some of them can be removed with interface direction instead of concrete struct. 
I may be wrong, but I think function or method having one line is not very useful. wdyt?

3.  It would be nice to have UTs for `ipsetmanager_windows.go` and `dataplane_windows.go` in this PR to guarantee their functionality, but as you said, it will be next PR. So, I think it is ok in this PR. 

4. While I read this PR, there are many common variables and const with `debugTools`. I think it is good to remove redundant declarations by checking `debugTools`  code carefully. One examaple, is `ipset.go` and `rule.pb.go` generated from `rule.proto`.

npm/pkg/dataplane/ipsets/ipset.go
JungukCho: I don't think you repeat below comment whenever using set. 
```go
// Using a map to emulate set and value as struct{} for
// minimal memory consumption
```
Instead, for example, shortly. It will be better comments to improve readability. Especially, more important information is to specify what key and values are when map and set are used in the code.
```go
set using map, key: (describe what key is)
```

vakalapa:
@JungukCho thanks foe the comments

1. I felt there is not much abstraction to be made with interfaces. As part of our design exercise we did find that most of the core logic  is same for both dataplanes, only the dataplane specific artifact conversions and application is different. So i took the https://golang.org/src/os/executable.go approach. The main executable package contains a publicly exposed function Executable() and each dataplane specific file will implement the executable() like https://golang.org/src/os/executable_windows.go. I felt this is much easier to read as the flow of code is pretty straight forward.  Please let me know if you think there is any issue with this approach
2. Sure, most of the dataplane.go function may stay inline like above executable approach but some will get filled up with abstractions needed for both dataplanes. Also a single dataplane call will make our lifes easier in terms extending dataplane with a gprc call in between later on for operator model. Like you mentioned i will try to see if i can non-capitalize any members.
3. Agreed, tests are important and will definitely be one focus area when adding windows dataplane support.
4. I did not want to increase the scope of this PR. We do have a debug tools work item in pipeline and as part of that work item and also as part of dataplane integration we should be able to clean pkg/dataplane all together. Will be a incremental process. Idea is to not touch any existing code as much possible.

npm/pkg/dataplane/policies/policy.go
```diff
@@ -0,0 +1,77 @@
...
+	// Making this a podKey instead should be
+	// use NPMPod obj
+	Pods  []string
+	RawNP *networkingv1.NetworkPolicy
+}
+
+type ACLPolicy struct { // Iptable rules
+	// PolicyID is the rules name with a given network policy
+	PolicyID string
+	// Comment is the string attached to rule to identity its representation
```
JungukCho, line 25: Does windows stack also support `comment`? 
Even I think comment will be generated in each dataplane?

npm/pkg/dataplane/policies/policy_windows.go
```diff
@@ -0,0 +1,58 @@
...
+	return policySettings, nil
+}
+
+func getHCNDirection(direction Direction) hcn.DirectionType {
+	switch direction {
+	case Ingress:
+		return hcn.DirectionTypeIn
+	case Egress:
+		return hcn.DirectionTypeOut
+	case Both:
```
JungukCho, line 27: seems this case is not necessary.

npm/pkg/dataplane/dataplane.go
```diff
@@ -0,0 +1,118 @@
...
+func NewDataPlane() *DataPlane {
+	return &DataPlane{
+		policyMgr:     policies.NewPolicyManager(),
+		ipsetMgr:      ipsets.NewIPSetManager(),
+		endpointCache: make(map[string]*NPMEndpoint),
+	}
+}
+
+// InitializeDataPlane helps in setting up dataplane for NPM
+// in linux this function should be adding required chains and rules
```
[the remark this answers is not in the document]
vakalapa: Good point. Sure will rearrange these comments.

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,218 @@
...
+type SetProperties struct {
+	// Stores type of ip grouping
+	Type SetType
+	// Stores kind of ipset in dataplane
+	Kind SetKind
+}
+
+type SetType int32
+
+const (
```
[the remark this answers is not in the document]
vakalapa: I will be doing this as part of DebugTools for dataplane, as this will touch existing working code and needs a total rehaul of our debugtools.

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -0,0 +1,300 @@
...
+	"net"
+	"sync"
+
+	"github.com/Azure/azure-container-networking/log"
+	"github.com/Azure/azure-container-networking/npm/metrics"
+	"github.com/Azure/azure-container-networking/npm/util/errors"
+)
+
+type IPSetManager struct {
+	setMap map[string]*IPSet
```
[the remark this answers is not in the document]
vakalapa: https://github.com/Azure/azure-container-networking/blob/7a4dbf84b6b1fd362deb0a130de495d9fd042538/npm/pkg/dataplane/ipsets/ipsetmanager.go#L14 

I did that originally, tried to keep listmap and setmap separate but then quickly realized there is not much perf gain we will have and as you said it is MUCH simpler to maintain one Map, easy to read and follow.

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,218 @@
...
+	KeyValueLabelOfPod SetType = 5
+	// NamedPorts IPSets contains a given namedport
+	NamedPorts SetType = 6
+	// NestedLabelOfPod is derived for multivalue matchexpressions
+	NestedLabelOfPod SetType = 7
+	// CIDRBlocks holds CIDR blocks
+	CIDRBlocks SetType = 8
+)
+
+var setTypeName = map[int32]string{
```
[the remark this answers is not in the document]
rbtr: `SetType` shouldn't be `int32`, either. just use `int`

npm/pkg/dataplane/ipsets/ipsetmanager_windows.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return err
+	}
+
+	policyNetworkRequest := hcn.PolicyNetworkRequest{
+		Policies: []hcn.NetworkPolicy{},
+	}
+
+	for _, policy := range network.Policies {
+		if policy.Type != "SetPolicy" {
```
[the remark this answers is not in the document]
vakalapa: Yes there is a constant in HCN but there is no HCN release with this SetPolicy changes, i can write up a TODO here to change this to constant when we update HCSshim

npm/pkg/dataplane/ipsets/ipset.go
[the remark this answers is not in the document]
vakalapa: Any particular i am actually thinking to use int8, @JungukCho  i am saving it as int32 to save member for each IPSet object, as you know we create at most about 64000 ipsets and on an average we will save about 1/4th the size when using int32. Int8 would use only 1 byte, compared to 16 of string. I am inclined to use int8 wdyt ?

npm/pkg/dataplane/policies/policy.go
```diff
@@ -0,0 +1,77 @@
...
+	networkingv1 "k8s.io/api/networking/v1"
+)
+
+type NPMNetworkPolicy struct {
+	Name string
+	// PodSelectorIPSets holds all the IPSets generated from Pod Selector
+	PodSelectorIPSets []*ipsets.IPSet
+	// OtherIPSets holds all IPSets generated from policy
+	// except for pod selector IPSets
+	OtherIPSets []*ipsets.IPSet
```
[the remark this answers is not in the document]
vakalapa: I was thinking of other names, but do you think peeripsets is the right word here ? i can use ruleIpsets ? 

SelectorIpSets -- from pod selector
RuleIPSets -- from rule ?

npm/pkg/dataplane/ipsets/ipset.go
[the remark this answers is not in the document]
rbtr: it seems like a premature optimization to me, but if you know you're holding 64000 of them that's immediately .5 gb of memory saved so...maybe

JungukCho:
> @JungukCho thanks foe the comments
> 
> 1. I felt there is not much abstraction to be made with interfaces. As part of our design exercise we did find that most of the core logic  is same for both dataplanes, only the dataplane specific artifact conversions and application is different. So i took the https://golang.org/src/os/executable.go approach. The main executable package contains a publicly exposed function Executable() and each dataplane specific file will implement the executable() like https://golang.org/src/os/executable_windows.go. I felt this is much easier to read as the flow of code is pretty straight forward. 
 Please let me know if you think there is any issue with this approach

If they are almost identical, this approach seems reasonable. How about `Dataplane`? I think they are pretty different implementations. I will take a look at your reference. I may find better patterns instead of interface.  I will let you know if I still prefer to interface and we can discuss more.

> 2. Sure, most of the dataplane.go function may stay inline like above executable approach but some will get filled up with abstractions needed for both dataplanes. Also a single dataplane call will make our lifes easier in terms extending dataplane with a gprc call in between later on for operator model. Like you mentioned i will try to see if i can non-capitalize any members.

Still I think those inline functions will be nicely removed with interface. Further more dataplane-specific functions will be added in next PRs.  So, it will increase more these inline functions. 
I think in this stage thinking a grpc call in design is not very helpful since it is too future. wdyt?

> 3. Agreed, tests are important and will definitely be one focus area when adding windows dataplane support.
> 4. I did not want to increase the scope of this PR. We do have a debug tools work item in pipeline and as part of that work item and also as part of dataplane integration we should be able to clean pkg/dataplane all together. Will be a incremental process. Idea is to not touch any existing code as much possible.

I am not sure it increases scope of the PR. I meant checking the existing variables and then reusing them to avoid redundant codes in this PR without touching existing logics.

npm/pkg/dataplane/policies/policy.go
vakalapa: Comment is not useful today in windows dataplane and comment is generated in translate policy. I left this in for debugging purposes in case of windows.

npm/pkg/dataplane/policies/policy_windows.go
vakalapa: Windows dataplane does accept rules on both directions, in case of network policies with both ingress and egress rules i can use a DENY all with BOTH, that way we get away with single rule than 2

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -0,0 +1,218 @@
...
+import (
+	"fmt"
+
+	"github.com/Azure/azure-container-networking/npm/util"
+)
+
+type IPSet struct {
+	Name       string
+	HashedName string
+	Properties SetProperties
```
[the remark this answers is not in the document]
vakalapa: Makes sense.

npm/util/util.go
```diff
@@ -341,3 +341,17 @@ func CompareSlices(list1, list2 []string) bool {
 	}
 	return true
 }
+
+func SliceToString(list []string) string {
```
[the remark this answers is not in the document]
huntergregory: nitpick but why not use `list` instead of `list[:]`

npm/pkg/dataplane/dataplane.go
```diff
@@ -0,0 +1,118 @@
...
+// they do not change apply changes into dataplane. This function needs to be called at the
+// end of IPSet operations of a given controller event, it will check for the dirty ipset list
+// and accordingly makes changes in dataplane. This function helps emulate a single call to
+// dataplane instead of multiple ipset operations calls ipset operations calls to dataplane
+func (dp *DataPlane) ApplyDataPlane() error {
+	return dp.ipsetMgr.ApplyIPSets(dp.networkID)
+}
+
+// AddPolicies takes in a translated NPMNetworkPolicy object and applies on dataplane
+func (dp *DataPlane) AddPolicies(policies *policies.NPMNetworkPolicy) error {
```
huntergregory, line 105: Can we make a consistent naming convention for function names and args around policies (with respect to plural vs. singular)?
In this file we have:
- AddPolicies(policies *NPMNetworkPolicy)  [plural, plural]
- RemovePolicies(policyName string)    [plural, singular]
and in policymanager.go we have:
-  AddPolicies(policy *NPMNetworkPolicy)   [plural, singular]
- GetPolicy(name string)   [singular, singular]

Even though a *NPMNetworkPolicy object can have multiple ACLs, I'd prefer using singular name and arg because plural usually implies a slice/array of the base object (string or *NPMNetworkPolicy).

vakalapa: That makes sense. Changed the functions to reflect singular CRUD operations.

npm/pkg/dataplane/policies/policy.go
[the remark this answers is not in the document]
JungukCho: By the way, based on your comments in `OtherIPSets`, will you store all pod selectors from rule and target information in `PodSelectorIPSet`? If so, how to differentiate them? Please check below example yaml.

I understood `PodSelectorIPSets` from `target pod` and others are from `egress` and `ingress` which will be peers to `target pod`. So, I suggested `peersIPSets`.  However, I think `RuleIPSets` is more reasonable since networkpolicy uses `NetworkPolicyIngressRule` and `NetworkPolicyEgressRule`.

```yaml
  podSelector: # target pod
    matchLabels:
      role: db
  policyTypes:
  - Ingress
  - Egress
  ingress:  # peers
  - from:
    - ipBlock:
        cidr: 172.17.0.0/16
        except:
        - 172.17.1.0/24
    - namespaceSelector:
        matchLabels:
          project: myproject
    - podSelector: # another podSelector in rules
        matchLabels:
          role: frontend
    ports:
    - protocol: TCP
      port: 6379
  egress:
  - to:
    - ipBlock:
        cidr: 10.0.0.0/24
    ports:
    - protocol: TCP
      port: 5978

```

npm/pkg/dataplane/ipsets/ipset.go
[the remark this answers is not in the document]
vakalapa: It is not a premature optimization, we know from current user behavior that each NPM holds at least 3-4k of Ipsets for a relatively small cluster. Which is still a high number.

npm/pkg/dataplane/policies/policy_windows.go
JungukCho: I meant `case Both` is not necessary based on logic. Anyway, by default it returns ""

npm/pkg/dataplane/policies/policy_windows.go
```diff
@@ -0,0 +1,58 @@
+package policies
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Microsoft/hcsshim/hcn"
```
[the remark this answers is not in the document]
vakalapa: Thanks for checking with Sravanth, below is his response. 
[10:10] Sravanth Bangari
HCSSHIM is legacy and has API's for compute + networking. I followed up with someone on the HCS team and this is only in use by docker today. HCN has the latest API's and is used by containerD, kube-proxy etc

JungukCho (dismissed):
LGTM. Thank you for applying comments. 
This PR will be foundation of next NPM!

matmerr (approved):
🚀

huntergregory (dismissed):

matmerr (approved):
🚀

huntergregory (approved):

JungukCho (approved):
lgtm

PR #988, opened 2021-09-02
[Swift] Add snat rule to host IP for IMDS for CNS IPAM

thatmattlong:
**Reason for Change**:

Let Pods query IMDS when CNS is doing the IPAM (when IPs come from secondary IPs of NC)

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -161,16 +161,15 @@ func setHostOptions(nwCfg *cni.NetworkConfig, hostSubnetPrefix *net.IPNet, ncSub
 	}
 
 	azureDNSMatch := fmt.Sprintf(" -m addrtype ! --dst-type local -s %s -d %s -p %s --dport %d", ncSubnetPrefix.String(), iptables.AzureDNS, iptables.UDP, iptables.DNSPort)
+	azureIMDSMatch := fmt.Sprintf(" -m addrtype ! --dst-type local -s %s -d %s -p %s --dport %d", ncSubnetPrefix.String(), iptables.AzureIMDS, iptables.TCP, iptables.HTTPPort)
 
-	// TODO remove this rule once we remove adding MASQUEARDE from AgentBaker, check below PR
-	// https://github.com/Azure/AgentBaker/pull/367/files
-	podTrafficAccept := fmt.Sprintf(" -m iprange  ! --dst-range 168.63.129.16-168.63.129.16  -s %s ", ncSubnetPrefix.String())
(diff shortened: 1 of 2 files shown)


iptables/iptables.go
```diff
@@ -78,7 +79,8 @@ const (
 
 // known IP's
 const (
-	AzureDNS = "168.63.129.16"
+	AzureDNS  = "168.63.129.16"
+	AzureIMDS = "169.254.169.254"
```
rbtr, line 83: idk what this is for, specifically, but should we do it for the whole link-local block?

thatmattlong: This is specifically to allow http traffic to the metadata service from pods. Since NCs are considered "secondary interfaces" and there is an explicit deny rule on the host to drop http traffic from secondary interfaces. So really we only want to snat traffic to this one IP against the host IP , since it gets dropped otherwise. For the rest of link-local there is no explicit deny rule so no need to snat.

cni/network/invoker_cns.go
```diff
@@ -161,16 +161,15 @@ func setHostOptions(nwCfg *cni.NetworkConfig, hostSubnetPrefix *net.IPNet, ncSub
 	}
 
 	azureDNSMatch := fmt.Sprintf(" -m addrtype ! --dst-type local -s %s -d %s -p %s --dport %d", ncSubnetPrefix.String(), iptables.AzureDNS, iptables.UDP, iptables.DNSPort)
+	azureIMDSMatch := fmt.Sprintf(" -m addrtype ! --dst-type local -s %s -d %s -p %s --dport %d", ncSubnetPrefix.String(), iptables.AzureIMDS, iptables.TCP, iptables.HTTPPort)
 
-	// TODO remove this rule once we remove adding MASQUEARDE from AgentBaker, check below PR
-	// https://github.com/Azure/AgentBaker/pull/367/files
-	podTrafficAccept := fmt.Sprintf(" -m iprange  ! --dst-range 168.63.129.16-168.63.129.16  -s %s ", ncSubnetPrefix.String())
 	snatPrimaryIPJump := fmt.Sprintf("%s --to %s", iptables.Snat, info.ncPrimaryIP)
+	snatHostIPJump := fmt.Sprintf("%s --to %s", iptables.Snat, info.hostPrimaryIP)
```
tamilmani1989, line 167: add a comment what the rule is?

thatmattlong: done

tamilmani1989 (approved):
lgtm

PR #990, opened 2021-09-02
Adding a build stage for arm64 cni

pjohnst5:
**Reason for Change**:
We need to build an ARM64 version of CNI too now.
Validation on an ARM64 machine:
```
azureuser@test-vm:~$ lscpu
Architecture:        aarch64
Byte Order:          Little Endian
CPU(s):              32
On-line CPU(s) list: 0-31
Thread(s) per core:  1
Core(s) per socket:  32
Socket(s):           1
NUMA node(s):        1
Vendor ID:           Cavium
Model:               2
Model name:          ThunderX2 99xx
Stepping:            0x1
BogoMIPS:            400.00
NUMA node0 CPU(s):   0-31
Flags:               fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics cpuid asimdrdm
azureuser@test-vm:~$ sudo ./azure-vnet
v1.4.11
```

And when I try running an amd64`cni binary on an arm64 machine, it fails
```
azureuser@test-vm:~$ sudo amd64/azure-vnet
amd64/azure-vnet: 3: amd64/azure-vnet: Syntax error: end of file unexpected (expecting ")")
```

### DIFF
--- a/.pipelines/pipeline.yaml
+++ b/.pipelines/pipeline.yaml
@@ -51,6 +51,7 @@ stages:
               cd ./output
               find . -name '*.tgz' -print -exec mv -t ./bins/ {} +
               find . -name '*.zip' -print -exec mv -t ./bins/ {} +
+              shopt -s extglob
               rm -rf !("bins")
             name: "PrepareArtifacts"
             displayName: "Prepare Artifacts"
--- a/.pipelines/pipeline.yaml
+++ b/.pipelines/pipeline.yaml
@@ -49,8 +49,8 @@ stages:
           - script: |
               mkdir -p ./output/bins
               cd ./output
-              find . -name '*.tgz' -print -exec mv -t ./bins/ {} +;
-              find . -name '*.zip' -print -exec mv -t ./bins/ {} +;
+              find . -name '*.tgz' -print -exec mv -t ./bins/ {} +
+              find . -name '*.zip' -print -exec mv -t ./bins/ {} +
               rm -rf !("bins")
             name: "PrepareArtifacts"
             displayName: "Prepare Artifacts"
--- a/.pipelines/pipeline.yaml
+++ b/.pipelines/pipeline.yaml
@@ -42,35 +42,16 @@ stages:
           name: "$(BUILD_POOL_NAME_DEFAULT)"
         steps:
           - script: |
-              GOOS=windows make all-binaries VERSION=$(TAG)
-            name: "BuildWindows"
-            displayName: "Build Windows"
-
-          - script: |
-              GOOS=linux make all-binaries VERSION=$(TAG)
-            name: "BuildLinux"
-            displayName: "Build Linux"
-
-          - script: |
-              GOOS=linux GOARCH=arm64 make azure-cni-plugin VERSION=$(TAG)
-            name: "BuildCNIArm64"
-            displayName: "Build CNI ARM64"
+              make all-binaries-platforms GOOS=linux,windows GOARCH=amd64,arm64 VERSION=$(TAG)
+            name: "BuildAllPlatformBinaries"
+            displayName: "Build all platform binaries"
 
           - script: |
               mkdir -p ./output/bins
-              cd ./output/linux_amd64
-              sudo find . -mindepth 2 -type f -regextype posix-extended ! -iregex '.*\.(zip|tgz)$' -delete
-              sudo find . -mindepth 2 -type f -print -exec mv {} ../bins \;
-              sudo rm -R -- */ && cd ..
-              cd ./output/linux_arm64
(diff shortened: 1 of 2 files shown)


rbtr:
rather than new build steps, could the existing build steps be updated to take a matrix of build args? it seems unlikely that this is the last GOOS or GOARCH change we will ever have to make, and that would give us a head start on having arm64 builds for all the rest of our components.

pjohnst5:
Hi Evan, what would that look like? I like the idea, just don't really know what it would look like

rbtr:
I think this could be as easy as adding a make target to the effect of `make all-binaries-multiplatform` that takes arrays of GOOS and GOARCH values, something like
```bash
make all-binaries-multiplatform GOOS=linux,windows GOARCH=arm64,amd64
```
then that make target could iterate those arrays to build the matrix of build targets, and just call the existing make targets with update build args:
```make
all-binaries-multiplatform:
	for goos in GOOS:
		for goarch in GOARCH:
			$(MAKE) all-binaries GOOS=goos GOARCH=goarch
```
but...in actual shell script😁

rbtr (changes requested):
blocking on discussed multiplatform change

pjohnst5:
Trying out the for loop way, where we pass in GOOS=linux,windows GOARCH=amd64,arm64

Makefile
```diff
@@ -269,11 +271,14 @@ all-containerized:
 	docker rm $(BUILD_CONTAINER_NAME)
 	docker rmi $(BUILD_CONTAINER_IMAGE):$(VERSION)
 
-# Make both linux and windows binaries
+# Make all platform binaries
```
rbtr, line 274: please move the comment to after the target, on the same line (see the tools targets at L500+) so that it will be picked up as a docstring when I add the help target like in aquarius

Makefile
```diff
@@ -269,11 +271,14 @@ all-containerized:
...
 .PHONY: all-binaries-platforms
 all-binaries-platforms:
-	export GOOS=linux; make all-binaries
-	export GOOS=windows; make all-binaries
+	for goos in $$(echo $(GOOS) | sed "s/,/ /g"); do \
+		for goarch in $$(echo $(GOARCH) | sed "s/,/ /g"); do \
+			make all-binaries GOOS=$$goos GOARCH=$$goarch; \
+		done \
+	done
```
rbtr, line 281: if the env vars are already arrays, we can simplify the array iteration significantly:
```suggestion
	@for goos in "$(GOOSES)"; do \
		for goarch in "$(GOARCHES)"; do \
			make all-binaries GOOS=$$goos GOARCH=$$goarch; \
		done \
	done
```

rbtr: we can get this done by setting these default vars in the makefile: 
```
GOOSES ?= "linux windows"
GOARCHES ?= "amd64 arm64"
```

that makes the default invocation:
`make all-binaries-multiplatform`.

to override it you need to set those as _Make_ variables (not shell variables), to different strings: 
`make all-binaries-multiplatform GOOSES="\"solaris bsd\"" ...`

rbtr (changes requested):

Makefile
rbtr: worth leaving a comment that those strings need to be escaped to override correctly

pjohnst5: Thanks, attempted in the latest commit, let me know if that's what you mean or if I should change something

.pipelines/pipeline.yaml
```diff
@@ -42,26 +42,17 @@ stages:
...
           - script: |
-              GOOS=windows make all-binaries VERSION=$(TAG)
-            name: "BuildWindows"
-            displayName: "Build Windows"
-
-          - script: |
-              GOOS=linux make all-binaries VERSION=$(TAG)
-            name: "BuildLinux"
-            displayName: "Build Linux"
+              make all-binaries-platforms GOOS=linux,windows GOARCH=amd64,arm64 VERSION=$(TAG)
```
rbtr, line 45: drop the GOOS and GOARCH so that we use the Makefile defaults in the pipeline runs

Makefile
```diff
@@ -74,6 +76,8 @@ NPMFILES = \
 # Build defaults.
 GOOS ?= linux
 GOARCH ?= amd64
+GOOSES ?= "linux windows"
+GOARCHES ?= "amd64 arm64"
```
rbtr, line 80: can you add comments by both of these that say basically >to override at the cli do `GOOSES="\"darwin bsd\""`
I think the need to double/escape quote will otherwise be confusing

rbtr (changes requested):
multiplat target looks good 💯 just a couple cleanup things around it

Makefile
pjohnst5: Thank you, adding these comments too

rbtr (approved):
lgtm

PR #992, opened 2021-09-08
feat: update cns client

rbtr:

**Reason for Change**:

Started off only trying to make the Debug API work, but it required further changes to the client.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/cnsclient/cnsclient.go
```diff
@@ -14,86 +16,74 @@ import (
...
+		return nil, errors.Wrap(err, "http request failed")
 	}
-
 	defer res.Body.Close()
 
 	if res.StatusCode != http.StatusOK {
-		errMsg := fmt.Sprintf("[Azure CNSClient] GetNetworkConfiguration invalid http status code: %v", res.StatusCode)
-		log.Errorf(errMsg)
-		return nil, &CNSClientError{types.UnexpectedError, errors.New(errMsg)}
+		return nil, &CNSClientError{types.UnexpectedError, errors.Errorf("http response %d", res.StatusCode)}
```
ramiro-gamarra, line 83: Can we add field names to struct literals?

cns/cnsclient/cnsclient.go
```diff
@@ -14,86 +16,74 @@ import (
...
 }
 
 // GetNetworkConfiguration Request to get network config.
-func (cnsClient *CNSClient) GetNetworkConfiguration(orchestratorContext []byte) (
-	*cns.GetNetworkContainerResponse, error) {
-	var body bytes.Buffer
-
-	url := cnsClient.connectionURL + cns.GetNetworkContainerByOrchestratorContext
-	log.Printf("GetNetworkConfiguration url %v", url)
+func (c *Client) GetNetworkConfiguration(orchestratorContext []byte) (*cns.GetNetworkContainerResponse, error) {
```
ramiro-gamarra, line 52: I see that context has been added to some methods below. Let's take this chance and add them to all client's methods.

ramiro-gamarra (commented):
Some initial comments, will review further tomorrow

tamilmani1989 (commented):
passing cnscli interface should be part of networkmanager object instead of argument in createendpoint . As discussed lets wait for  @vakalapa PR (which changes method signatures for createendpoint,etc to take nm as object) to get merged

rbtr:
@tamilmani1989 this shouldn't be blocked by that, they can go in any order.
this is blocking some changes in CNS currently though.

network/manager.go
```diff
@@ -67,18 +67,18 @@ type NetworkManager interface {
...
-	GetEndpointInfoBasedOnPODDetails(networkId string, podName string, podNameSpace string, doExactMatchForPodName bool) (*EndpointInfo, error)
-	AttachEndpoint(networkId string, endpointId string, sandboxKey string) (*endpoint, error)
-	DetachEndpoint(networkId string, endpointId string) error
-	UpdateEndpoint(networkId string, existingEpInfo *EndpointInfo, targetEpInfo *EndpointInfo) error
-	GetNumberOfEndpoints(ifName string, networkId string) int
+	DeleteNetwork(networkID string) error
+	GetNetworkInfo(networkID string) (NetworkInfo, error)
+
+	CreateEndpoint(client apipaClient, networkID string, epInfo *EndpointInfo) error
+	DeleteEndpoint(cli apipaClient, networkID string, endpointID string) error
```
ramiro-gamarra, line 74: Do clients change on different invocations? Or are they tied to the lifetime of the manager? If the latter, this should probably just be an internal field of the manager.

cns/client/client_test.go
```diff
@@ -391,3 +392,39 @@ func TestCNSClientDebugAPI(t *testing.T) {
...
+		want    *Client
+		wantErr bool
+	}{
+		{
+			name: "empty url",
+			url:  "",
+			want: &Client{
+				routes: emptyRoutes,
+				client: http.Client{
+					Timeout: 0,
```
ramiro-gamarra, line 411: is this line needed?

cns/client/client_test.go
```diff
@@ -391,3 +392,39 @@ func TestCNSClientDebugAPI(t *testing.T) {
...
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := New(tt.url, tt.timeout)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("New() error = %v, wantErr %v", err, tt.wantErr)
+				return
```
ramiro-gamarra, line 423: Thoughts on
```go
if tt.wantErr {
    require.Error(t, err)
} else {
    require.NoError(t, err)
}
```
? (I'm a big fan of the testify assertion packages :))

cns/client/client_test.go
```diff
@@ -391,3 +392,39 @@ func TestCNSClientDebugAPI(t *testing.T) {
 	t.Logf("PodIPConfigState: %+v", inmemory.HTTPRestServiceData.PodIPConfigState)
 	t.Logf("IPAMPoolMonitor: %+v", inmemory.HTTPRestServiceData.IPAMPoolMonitor)
 }
+
+func TestNew(t *testing.T) {
+	emptyRoutes, _ := buildRoutes(defaultBaseURL, clientPaths)
+	tests := []struct {
```
ramiro-gamarra, line 398: Will more cases be added? Otherwise, are table tests needed?

cns/client/client.go
```diff
@@ -0,0 +1,402 @@
...
+}
+
+// Client specifies a client to connect to Ipam Plugin.
+type Client struct {
+	client http.Client
+	routes map[string]url.URL
+}
+
+// New returns a new CNS client configured with the passed URL and timeout.
+func New(baseURL string, requestTimeout time.Duration) (*Client, error) {
```
ramiro-gamarra, line 43: This may be a larger change, but what we really need to pass into this client is an `*http.Client`. Users of this code should have complete control on all the knobs available on the transport layer. If this is too big to address now, can you add a todo?

ramiro-gamarra (commented):
Some more comments

network/manager.go
rbtr: agree but refactoring the manager is wayyyy out of scope here imo and there is ongoing work for that happening elsewhere

cns/client/client_test.go
rbtr: strictly speaking no, that's the zero val

cns/client/client_test.go
rbtr: more tests to follow, just need to get this base refactor out

cns/client/client.go
rbtr: it's coming, movement in that direction: https://github.com/Azure/azure-container-networking/pull/1001

cns/client/client_test.go
rbtr: I also prefer testify, will incorporate this in to the follow up changes

cns/client/client.go
```diff
@@ -0,0 +1,402 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), &body)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to build request")
+	}
+	req.Header.Set(headerContentType, contentTypeJSON)
+	res, err := c.client.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "http request failed")
+	}
+	defer res.Body.Close()
```
ramiro-gamarra, line 105: One thing that should also be added to all of these network calls is making sure the entire response body is read before closing, otherwise tcp connections won't be reused efficiently:

```go
defer func() {
    _, _ = io.Copy(ioutil.Discard, res.Body)
    res.Body.Close()
}()
```

ramiro-gamarra (approved):
looks headed in the right direction to me

tamilmani1989:
> @tamilmani1989 this shouldn't be blocked by that, they can go in any order.
> this is blocking some changes in CNS currently though.

sure. would you be changing the code to initialize cnsclient in networkmanager object later?

rbtr:
@tamilmani1989 yes, I have started discussing with @vakalapa already

cns/client/client_test.go
pjohnst5: I'll pick this up

PR #993, opened 2021-09-10
[NPM] refactor: npmcache json marshaling

JungukCho:
**Reason for Change**:

Do a right way of encoding and decoding npmCache to achieve performance and json compatibility.
All credits to @rbtr to refactor codes.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:


npm/pkg/dataplane/testfiles/npmgr.json
```diff
@@ -0,0 +1,1579 @@
+{
```
rbtr, line 1: instead of "testfiles", call test folders "testdata"
https://pkg.go.dev/cmd/go/internal/test
> The go tool will ignore a directory named "testdata", making it available
to hold ancillary data needed by the tests.

rbtr (changes requested):

rbtr (approved):

PR #996, opened 2021-09-13
Netlink package interfacing and adding a fake

vakalapa:
1. Creating a interface for netlink and calling interface instead of the package
2. Changing all netlink pkg calls to take in this interface. This includes adding a new interface for epcommon.


network/endpoint_linux.go
```diff
@@ -437,12 +437,12 @@ func updateRoutes(existingEp *EndpointInfo, targetEp *EndpointInfo) error {
...
 	}
 
-	err := deleteRoutes(existingEp.IfName, tobeDeletedRoutes)
+	err := deleteRoutes(nl, existingEp.IfName, tobeDeletedRoutes)
 	if err != nil {
 		return err
 	}
 
-	err = addRoutes(existingEp.IfName, tobeAddedRoutes)
+	err = addRoutes(nl, existingEp.IfName, tobeAddedRoutes)
```
tamilmani1989, line 445: addroutes and deleteroutes can have networkmanager as objects

network/epcommon/endpoint_common.go
```diff
@@ -35,22 +36,17 @@ const (
...
 
-func getPrivateIPSpace() []string {
-	privateIPAddresses := []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "169.254.0.0/16"}
-	return privateIPAddresses
-}
-
-func getFilterChains() []string {
-	chains := []string{"FORWARD", "INPUT", "OUTPUT"}
-	return chains
+type NetlinkRedirection struct {
```
tamilmani1989, line 39: EpCommon?

network/epcommon/endpoint_common.go
```diff
@@ -35,22 +36,17 @@ const (
...
-	actions := []string{"ACCEPT", "DROP"}
-	return actions
+func NewNetlinkRedirection(netlink netlink.Netlink) NetlinkRedirection {
+	return NetlinkRedirection{
+		netlink: netlink,
+	}
 }
 
-func CreateEndpoint(hostVethName string, containerVethName string) error {
+func (nr NetlinkRedirection) CreateEndpoint(hostVethName string, containerVethName string) error {
```
tamilmani1989, line 49: func (epc *EpCommon)

vakalapa:
Working on fixing some GoLints for PR is ready for review now.

vakalapa:
Also, i did a merge master to include @rbtr 's CNSClient PR, and this PR shows that diff too for some reason

network/endpoint_linux.go
vakalapa: Keeping this simple by passing in the netlink object and not the whole NM object

cni/network/network.go
```diff
@@ -108,8 +109,9 @@ func NewPlugin(name string, config *common.PluginConfig, client NnsClient) (*net
 		return nil, err
 	}
 
+	netlink := netlink.NewNetlink()
```
ramiro-gamarra, line 112: this var shadows the package name. i guess the linter is not run on the target branch?

cnms/service/networkmonitor.go
```diff
@@ -146,7 +147,8 @@ func main() {
 			return
 		}
 
-		nm, err := network.NewNetworkManager()
+		netlink := netlink.NewNetlink()
```
ramiro-gamarra, line 150: same as above

netlink/ip.go
```diff
@@ -1,6 +1,7 @@
 // Copyright 2017 Microsoft. All rights reserved.
 // MIT License
 
+//go:build linux
```
ramiro-gamarra, line 4: better to just add the suffix to file name?

netlink/link.go
```diff
@@ -1,6 +1,7 @@
 // Copyright 2017 Microsoft. All rights reserved.
 // MIT License
 
+//go:build linux
```
ramiro-gamarra, line 4: same as above

netlink/netlink.go
```diff
@@ -1,32 +1,7 @@
...
-	req := newRequest(unix.NLMSG_NOOP, unix.NLM_F_ECHO|unix.NLM_F_ACK)
-	if req == nil {
-		return unix.ENOMEM
-	}
-
-	req.addPayload(newAttributeString(0, text))
+type Netlink struct{}
 
-	return s.sendAndWaitForAck(req)
+func NewNetlink() Netlink {
```
ramiro-gamarra, line 5: Convention is that a `New` prefix in constructors returns pointers

network/bridge_networkclient_linux.go
```diff
@@ -1,29 +1,45 @@
...
 )
 
 const (
 	multicastSolicitPrefix = "ff02::1:ff00:0/104"
 )
 
+var errorLinuxBridgeClient = errors.New("LinuxBridgeClient Error")
+
+func newErrorLinuxBridgeClient(errStr string) error {
+	return fmt.Errorf("ErrorLinuxBridgeClient %w : %s", errorLinuxBridgeClient, errStr)
```
ramiro-gamarra, line 22: Perhaps a new type that implements `error` would be better here. This way produces an error message that's a bit odd:
```
ErrorLinuxBridgeClient LinuxBridgeClient Error : <...>
```

network/epcommon/endpoint_common.go
```diff
@@ -35,22 +38,23 @@ const (
 	acceptRAV6File       = "/proc/sys/net/ipv6/conf/%s/accept_ra"
 )
 
-func getPrivateIPSpace() []string {
-	privateIPAddresses := []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "169.254.0.0/16"}
-	return privateIPAddresses
+var errorEPCommon = errors.New("ErrorEPCommon Error")
+
+func newErrorEPCommon(errStr string) error {
+	return fmt.Errorf("ErrorEPCommon %w : %s", errorEPCommon, errStr)
```
ramiro-gamarra, line 44: Same feedback as above

network/epcommon/endpoint_common.go
```diff
@@ -35,22 +38,23 @@ const (
...
 }
 
-func getFilterChains() []string {
-	chains := []string{"FORWARD", "INPUT", "OUTPUT"}
-	return chains
+type EPCommon struct {
```
ramiro-gamarra, line 47: What is the meaning of `Common` in this context? Seems like this file is specific to linux. What is this common to?

netlink/mocknetlink.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return nil
+}
+
+func (f *MockNetlink) AddLink(Link) error {
+	return f.error()
+}
+
+func (f *MockNetlink) DeleteLink(string) error {
+	return f.error()
```
ramiro-gamarra, line 39: How is the interface that this mock implements used? The problem with having a single flag for all returns is that it restricts different permutations on the behavior this type displays. As a silly example, if a client has 
```go
func foo(m NetlinkInterface) error {
    if err := m.AddLink(...); err != nil {
        return err
    }

    return m.DeleteLink(...)
}
```

Then I wouldn't be able to cover the last line with a test.

ramiro-gamarra (commented):
Some initial comments/questions

cni/network/network.go
vakalapa: Linter did run on the target branch. Local lint is fine with it too. But do you think this is not a good approach ?

netlink/mocknetlink.go
vakalapa: I am too worried about how the mock works underneath, we can change it as we start writing tests, this PR is to build the infra. In consecutive PRs for adding tests with Mocknetlink we can improve it based on requirement

cni/network/network.go
ramiro-gamarra: @rbtr is this not a warning in the current CI lint setup? @vakalapa yeah, this causes problems. Say you want a second instance of this type in this block using `NewNetlink()`. You won't be able to because the package is shadowed by the variable. Same for any other exported symbol from that package.

network/epcommon/endpoint_common.go
vakalapa: We decided to rename the file to networkutility_linux but will be part of a 2nd PR with other changes. Because i am blocking some changes from Shriroop and Ali.

network/bridge_networkclient_linux.go
vakalapa: Noted, will fix in 2nd PR

cnm/network/network_test.go
```diff
@@ -59,9 +59,10 @@ func TestMain(m *testing.M) {
 		fmt.Printf("Failed to start network plugin %v\n", err)
 		os.Exit(2)
 	}
+	nl := netlink.NewNetlink()
```
tamilmani1989, line 62: is this getting compiled in windows? i see netlink specific to linux

vakalapa: Yes you can see checks ran on this PR, i created a netlink_windows to help implement the interface for windows too.

cni/network/network.go
matmerr: yeah this is an issue, I don't see any warnings coming from lint but import naming collisions are raised by the ide, ex:

image

rbtr: current lint config was not flagging shadows of imports, update to include that here: https://github.com/Azure/azure-container-networking/pull/1009

network/bridge_endpointclient_linux.go
```diff
@@ -50,7 +54,8 @@ func NewLinuxBridgeEndpointClient(
 }
 
 func (client *LinuxBridgeEndpointClient) AddEndpoints(epInfo *EndpointInfo) error {
-	if err := epcommon.CreateEndpoint(client.hostVethName, client.containerVethName); err != nil {
+	epc := epcommon.NewEPCommon(client.netlink)
```
matmerr, line 57: can we move the epc to the LinuxBridgeEndpointClient struct and reuse in `ConfigureContainerInterfacesAndRoutes` , `SetupContainerInterfaces` etc?

cni/network/network.go
vakalapa: Fixed in places where there is a standalone netlink var. This sure will cause confusion later on so keeping it simple.

tamilmani1989 (approved):

netlink/netlink_windows.go
```diff
@@ -2,3 +2,82 @@
...
+	Type string
+	Name string
+}
+
+func (linkInfo *LinkInfo) Info() *LinkInfo {
+	return linkInfo
+}
+
+func (Netlink) AddLink(link Link) error {
+	return nil
```
rbtr, line 26: that this whole file is a stub in the os-specific build makes me think that this abstraction is wrong

PR #1000, opened 2021-09-13
chore: Remove dead codes related to telemetry files.

JungukCho:
**Reason for Change**:

Remove dead codes from telemetry.go and telemetry_<suffix>.go files.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/telemetry/telemetrybuffer.go
+++ b/telemetry/telemetrybuffer.go
@@ -327,17 +327,6 @@ func (tb *TelemetryBuffer) ConnectToTelemetryService(telemetryNumRetries, teleme
 	}
 }
 
-// TryToConnectToTelemetryService - Attempt to connect telemetry process without spawning it if it's not already running.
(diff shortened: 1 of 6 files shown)


telemetry/telemetrybuffer.go
```diff
@@ -327,17 +327,6 @@ func (tb *TelemetryBuffer) ConnectToTelemetryService(telemetryNumRetries, teleme
 	}
 }
 
-// TryToConnectToTelemetryService - Attempt to connect telemetry process without spawning it if it's not already running.
```
tamilmani1989, line 330: why was this removed?

JungukCho: @tamilmani1989  It was not called from any places. So, I think this is dead code. Do I miss something?

JungukCho:
Close it to raise the another PR.

PR #1006, opened 2021-09-15
Create NNC scoped client and isolate ipampoolmonitor and NNC reconciler

rbtr:
**Reason for Change**:

- Creates an NNC-scoped client and moves NNC interactions in CNS to this client interface (simplifies NNC manipulations).
- Updates crd/client to support NNC scoped client requirements.
- Breaks the circular dependency between the NNC reconciler and the ipampoolmonitor by using scoped client.
- Pulls dependency initialization out of these sub-components in to main where they can be injected and shared/reused.
- Refactors NNC reconciler initialization and extracts the CNS state initialization from it so that the reconciler doesn't block main.
- Removes lots of old code and adds lots of tests.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/service/main.go
```diff
@@ -773,18 +779,81 @@ func InitializeMultiTenantController(ctx context.Context, httpRestService cns.HT
...
+// InitCNS initializes cns by passing pods and a createnetworkcontainerrequest
+func initCNS(ctx context.Context, cli nodeNetworkConfigGetter, ncReconciler ncStateReconciler) error {
+	// Get nnc using direct client
+	nnc, err := cli.Get(ctx)
 	if err != nil {
-		logger.Errorf("[Azure CNS] Failed to get kubeconfig for request controller: %v", err)
+		// If the CRD is not defined, exit
+		if crd.IsNotDefined(err) {
+			logger.Errorf("CRD is not defined on cluster: %v", err)
+			os.Exit(1)
```
ramiro-gamarra, line 799: let `main()` be the one to exit the process? maybe this logic belongs in caller, or above

cns/service/main.go
```diff
@@ -773,18 +779,81 @@ func InitializeMultiTenantController(ctx context.Context, httpRestService cns.HT
...
+			return nil
+		}
+
+		// If instance of crd is not found, pass nil to CNSClient
+		if client.IgnoreNotFound(err) == nil {
+			err = ncReconciler.ReconcileNCState(nil, nil, nnc.Status.Scaler, nnc.Spec)
+			if err != nil {
+				return errors.Wrap(err, "failed to reconcile NC state")
+			}
+			return nil
```
ramiro-gamarra, line 813: Any reason not to just 
```go
return ncReconciler.ReconcileNCState(nil, nil, nnc.Status.Scaler, nnc.Spec)
```

Otherwise, if you must wrap the error, `errors.Wrap` returns nil if `err` is nil. You could just
```go
err = ncReconciler.ReconcileNCState(nil, nil, nnc.Status.Scaler, nnc.Spec)
return errors.Wrap(err, "failed to reconcile NC state")
```

crd/nodenetworkconfig/client.go
```diff
@@ -5,36 +5,66 @@ import (
...
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to create nnc crd")
 	}
 	return res, nil
 }
 
+// Get returns the NodeNetworkConfig identified by the NamespacedName.
+func (c *Client) Get(ctx context.Context, key types.NamespacedName) (*v1alpha.NodeNetworkConfig, error) {
+	nodeNetworkConfig := &v1alpha.NodeNetworkConfig{}
+	return nodeNetworkConfig, errors.Wrapf(c.nnccli.Get(ctx, key, nodeNetworkConfig), "failed to get nnc %v", key)
```
ramiro-gamarra, line 65: have the statements in separate lines? bit hard to read on first pass

crd/nodenetworkconfig/client.go
```diff
@@ -54,18 +85,53 @@ func (c *Client) InstallOrUpdate(ctx context.Context) (*v1.CustomResourceDefinit
...
+
+	return obj, nil
+}
+
+// UpdateSpec does a fetch, deepcopy, and update of the NodeNetworkConfig with the passed spec.
+// Deprecated: UpdateSpec is deprecated and usage should migrate to PatchSpec.
+func (c *Client) UpdateSpec(ctx context.Context, key types.NamespacedName, spec *v1alpha.NodeNetworkConfigSpec) (*v1alpha.NodeNetworkConfig, error) {
+	nnc, err := c.Get(ctx, key)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to update nnc")
```
ramiro-gamarra, line 130: `failed to get nnc`?

ramiro-gamarra (commented):
a few initial comments

cns/service/main.go
rbtr: > errors.Wrap returns nil if err is nil
I have (ab)used this in other places, wasn't sure how others felt about it though. I prefer it and will update this

cns/service/main.go
rbtr: changed to just return the error, and main is already set up to exit there 👍🏼

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -9,28 +9,31 @@ import (
...
 	mu                       sync.RWMutex
-	rc                       singletenantcontroller.RequestController
 	scalarUnits              v1alpha.Scaler
 	updatingIpsNotInUseCount int
+	nnccli                   nodeNetworkConfigSpecUpdater
 }
 
-func NewCNSIPAMPoolMonitor(httpService cns.HTTPService, rc singletenantcontroller.RequestController) *CNSIPAMPoolMonitor {
+func NewCNSIPAMPoolMonitor(httpService cns.HTTPService, nnccli nodeNetworkConfigSpecUpdater) *CNSIPAMPoolMonitor {
 	logger.Printf("NewCNSIPAMPoolMonitor: Create IPAM Pool Monitor")
```
thatmattlong, line 33: I wonder how much value this adds for us. Maybe we can move this log line to the caller and get rid of this constructor func completely

cns/singletenantcontroller/reconciler_test.go
```diff
@@ -0,0 +1,161 @@
...
+	spec   v1alpha.NodeNetworkConfigSpec
+}
+
+type mockCNSClient struct {
+	state                 cnsClientState
+	createOrUpdateNC      func(cns.CreateNetworkContainerRequest) error
+	updateIPAMPoolMonitor func(v1alpha.Scaler, v1alpha.NodeNetworkConfigSpec)
+}
+
+//nolint:gocritic // ignore hugeParam pls
```
thatmattlong, line 31: 😆

crd/nodenetworkconfig/client.go
```diff
@@ -54,18 +86,53 @@ func (c *Client) InstallOrUpdate(ctx context.Context) (*v1.CustomResourceDefinit
...
+	}
+
+	if err := c.nnccli.Patch(ctx, obj, ctrlcli.RawPatch(types.ApplyPatchType, patch)); err != nil {
+		return nil, errors.Wrap(err, "failed to patch nnc")
+	}
+
+	return obj, nil
+}
+
+// UpdateSpec does a fetch, deepcopy, and update of the NodeNetworkConfig with the passed spec.
```
thatmattlong, line 126: Update does a resourceVersion check that Patch does not do, right? In which case we may not want to deprecate?

rbtr: the problem is that this update process operates client-side. patch can be server-side. so we don't run in to race/stale-data issues

cns/service/main.go
```diff
@@ -773,18 +779,78 @@ func InitializeMultiTenantController(ctx context.Context, httpRestService cns.HT
...
 	if err != nil {
-		logger.Errorf("[Azure CNS] Failed to get kubeconfig for request controller: %v", err)
+
+		if crd.IsNotDefined(err) {
+			return errors.Wrap(err, "failed to get NNC during init CNS state")
+		}
+
+		if nnc == nil {
+			logger.Errorf("NodeNetworkConfig is not present on cluster")
+			return nil
```
ramiro-gamarra, line 803: is this just a sanity check or is it expected to happen? seems like the client code would not return nil if I'm reading correctly

ramiro-gamarra (commented):
Just some style nits, but one question in `initCNS`

cns/service/main.go
rbtr: you're right, this will never be nil with the new client and the subsequent `IgnoreNotFound(err)` check covers this case anyway

ramiro-gamarra (approved):
lgtm

PR #1008, opened 2021-09-15
Add UT coverage ovs_network_client

shriroopjoshi:
**Reason for Change**:

This PR attempts to improve code coverage for acn/network directory by:
1. Adding interfaces for `ovs-ctl`
2. Adding tests to cover more code


network/endpoint_linux.go
```diff
@@ -223,7 +225,7 @@ func (nw *network) deleteEndpointImpl(_ apipaClient, nl netlinkinterface.Netlink
 	// entering the container netns and hence works both for CNI and CNM.
 	if ep.VlanID != 0 {
 		epInfo := ep.getInfo()
-		epClient = NewOVSEndpointClient(nw, epInfo, ep.HostIfName, "", ep.VlanID, ep.LocalIP, nl)
+		epClient = NewOVSEndpointClient(nw, epInfo, ep.HostIfName, "", ep.VlanID, ep.LocalIP, nl, ovsctl.NewOvsctl())
```
aegal, line 228: As is, this change doesn't seem to allow ways to inject a mock ovs into this code path, one would be required to pass it down from the platform agnostic code. I think we need to discuss how we will inject the mock you add in this P.R to this client so we can have tests around this.

vakalapa: I am working on platform mock. Platform is the pkg being called inside ovsctl. I will update Ovsctl to take the mocked platform or real platform. That will alleviate issues with injection from upper levels

vakalapa (approved):
lgtm, we will need to update ovsctl to take in platform

PR #1011, opened 2021-09-15
chore: Refactor UTs in telemetry packages

JungukCho:
**Reason for Change**:

Refactor UTs to cleanup UTs and increase UT coverages in telemetry package

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/telemetry/aiwrapper_test.go
+++ b/telemetry/aiwrapper_test.go
@@ -1,4 +1,3 @@
-// Copyright Microsoft. All rights reserved.
 package telemetry
 
 import (
--- a/telemetry/aiwrapper_test.go
+++ b/telemetry/aiwrapper_test.go
@@ -5,7 +5,7 @@ import (
 	"testing"
 
 	"github.com/Azure/azure-container-networking/aitelemetry"
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestCreateAITelemetryHandle(t *testing.T) {
@@ -35,15 +35,15 @@ func TestCreateAITelemetryHandle(t *testing.T) {
 		},
 	}
 
-	for _, test := range tests {
-		tt := test
+	for _, tt := range tests {
+		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			err := CreateAITelemetryHandle(tt.aiConfig, tt.disableAll, tt.disableMetric, tt.disableTrace)
 			if tt.wantErr {
-				assert.Error(t, err)
+				require.Error(t, err)
 				return
 			}
-			assert.NoError(t, err)
+			require.NoError(t, err)
 		})
 	}
 }
--- a/telemetry/aiwrapper_test.go
+++ b/telemetry/aiwrapper_test.go
@@ -1,0 +1,49 @@
+// Copyright Microsoft. All rights reserved.
+package telemetry
+
+import (
+	"testing"
+
+	"github.com/Azure/azure-container-networking/aitelemetry"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateAITelemetryHandle(t *testing.T) {
+	tests := []struct {
+		name          string
+		aiConfig      aitelemetry.AIConfig
+		disableAll    bool
+		disableMetric bool
+		disableTrace  bool
+		wantErr       bool
+	}{
+		{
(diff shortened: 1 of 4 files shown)


telemetry/aiwrapper_test.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		tt := test
+		t.Run(tt.name, func(t *testing.T) {
+			err := CreateAITelemetryHandle(tt.aiConfig, tt.disableAll, tt.disableMetric, tt.disableTrace)
+			if tt.wantErr {
+				assert.Error(t, err)
+				return
```
rbtr, line 44: use require, it exits the test on a failed assertion
```suggestion
				require.Error(t, err)
```

telemetry/aiwrapper_test.go
```diff
@@ -0,0 +1,49 @@
...
+			aiConfig:      aitelemetry.AIConfig{},
+			disableAll:    true,
+			disableMetric: true,
+			disableTrace:  true,
+			wantErr:       true,
+		},
+	}
+
+	for _, test := range tests {
+		tt := test
```
rbtr, line 39: for test, just do `tt := tt`

telemetry/telemetrybuffer_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	for _, test := range tests {
+		tt := test
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tbClient.Write(tt.data)
+			if tt.wantErr {
+				assert.Error(t, err)
+				return
+			}
+			assert.NoError(t, err)
```
rbtr, line 108:
```suggestion
			if tt.wantErr {
				require.Error(t, err)
			}
			require.NoError(t, err)
```

rbtr (dismissed):
you could make a couple changes to be more strict about require/assert but otherwise lgtm 🚀

telemetry/aiwrapper_test.go
JungukCho: @rbtr  Initially I did it, but it looks a little confused. Any critical reasons to stick `tt := tt` ?

```go
for _, tt := range tests {
		tt := tt
```

telemetry/aiwrapper_test.go
JungukCho: @rbtr  Used `assert` to keep iterating the rest of them even though one is failed.

telemetry/aiwrapper_test.go
rbtr: it's just conventional to assign to the same name, since you're not going to use the var in the outer scope
https://github.com/golang/go/wiki/CommonMistakes

telemetry/aiwrapper_test.go
rbtr: with table driven tests, the other scenarios will still be executed, this will just stop the current scenario from progressing. i think it makes sense to use it on the error/noerror because testing anything after that if you don't have what you expect doesn't make sense.

telemetry/telemetrybuffer_test.go
JungukCho: While still it still needs `return` in `if` statement. It not, it runs `require.NoError(t, err)`

tamilmani1989 (approved):
lgtm. ubuntu18 e2e failed. please check

telemetry/aiwrapper_test.go
```diff
@@ -0,0 +1,49 @@
+// Copyright Microsoft. All rights reserved.
```
rbtr, line 1: don't need license headers, the license.txt in the root covers everything

rbtr (approved):

telemetry/aiwrapper_test.go
JungukCho: You meant `license/cla — All CLA requirements met.` in pipeline covers all? So, we do not need to add this line anymore for new files, but leave it in existing files?

rbtr: unless the file has a _different_ license, all files in this repo are covered by https://github.com/Azure/azure-container-networking/blob/master/LICENSE

rbtr (approved):

PR #1012, opened 2021-09-16
wrap hnsv2 calls and add a test with mocks

aegal:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


network/endpoint_windows.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/Microsoft/hcsshim/hcn"
 )
 
+var hnsv2 hnswrapper.HnsV2WrapperInterface = hnswrapper.Hnsv2wrapper{}
```
vakalapa, line 21: Instead of initializing it here, can you pass this in for each needed call. And let this be part of NetworkManager. We can initialize this in NewnetworkManager. That way we can control what gets passed on early on.

network/endpoint_windows_test.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2017 Microsoft. All rights reserved.
+// MIT License
+
+// +build !windows
```
vakalapa, line 4: Is this build tag right ? this will NOT build for windows right ?

network/endpoint_windows_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/Azure/azure-container-networking/network/hnswrapper"
+	"net"
+	"testing"
+)
+
+func TestNewAndDeleteEndpointImplHnsV2(t *testing.T){
+	nw := &network{
+		Endpoints: map[string]*endpoint{},
+	}
+	hnsv2 = hnswrapper.Hnsv2wrapperFake{}
```
vakalapa, line 19: Where is this hnsv2 being used ?

vakalapa (commented):
Some initial questions

network/endpoint_windows_test.go
aegal: This is package level, I do this to avoid passing around os specific code such as HNS in platform agnostic code, unfortunately its not the best but given our setup, I think its a better approach.

network/endpoint_windows.go
aegal: I want to avoid passing around os specific code such as HNS in platform agnostic code. By passing from the top down, its my opinion that it muddies the code. Happy to chat more about it though.

shriroopjoshi (approved):

shriroopjoshi (approved):

PR #1019, opened 2021-09-17
Remove cns http client which did nothing

pjohnst5:
**Reason for Change**:
removing cns httpapi client which did nothing essentially


pjohnst5:
Hi @feiskyer, can you take a look at this change which touches the multitenant scenario? I believe you had written tests for it, which are now slightly different.

The gist of it is, there used to be a `cns/cnsclient/apiclient.go` file which was moved to `cns/multitenantcontroller/multitenantoperator/cnsclient.go` which now I am getting rid of, since the `cns/client/httpapi/client.go` file did nothing essentially, just passed arguments through.

feiskyer:
+@thatmattlong could you confirm this change won't break things?

@pjohnst5  please ensure the unit tests are fixed when making changes

cns/restserver/errors.go
```diff
@@ -0,0 +1,18 @@
+package restserver
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/Azure/azure-container-networking/cns/types"
+)
+
+var errResponseCode = errors.New("Response code is error")
```
feiskyer, line 10: this is not clear, could you convert the error code to real error string?

cns/multitenantcontroller/multitenantoperator/multitenantcrdreconciler.go
```diff
@@ -29,11 +30,17 @@ const (
 	NCStateTerminated = "Terminated"
 )
 
+type cnsrestservice interface {
```
thatmattlong, line 33: Let's fix this name to match go style guide, `cnsRESTService`

cns/restserver/errors.go
```diff
@@ -0,0 +1,18 @@
...
+
+// ResponseCodeToError converts a cns response code to error type. If the response code is OK, then return value is nil
+func ResponseCodeToError(responseCode types.ResponseCode) error {
```
thatmattlong, line 13: I would suggest creating a new type that implmenets `error`, such as `CNSRESTError`, which might look like:
```
type CNSRESTError struct {
    ResponseCode types.ResponseCode
}
func (c *CNSRESTError) Error() string {
    return fmt.Errorf("response code: %d", c.ResponseCode)
}
```
This allows errors to hold data better, rather than just storing all the data in the error string. If we start using the `errors.Is` and `errors.As` functions, this will be a big help.

thatmattlong: Of course, then this `ResponseCodeToError` function becomes
```
return &CNSRESTError{
    ResponseCode: responseCode,
}
```

pjohnst5: Thanks @thatmattlong , wouldn't we still want an if statement in `ResponseCodeToError` like so?
```
func ResponseCodeToError(responseCode types.ResponseCode) error {
	if responseCode == types.Success {
		return nil
	}

	return &CNSRESTError{
		ResponseCode: responseCode,
	}
}

```

rbtr: _really_ we would want to go the other way, and return the error struct or nil where we are currently returning the response code, but that might snowball this change considerably

pjohnst5: Yeah I took a peek and it did seem to unravel a bit

pjohnst5: I will resolve this unless anyone has more comments on it

rbtr: ack

pjohnst5:
Could I get approval @thatmattlong ?

thatmattlong (approved):

PR #1027, opened 2021-09-20
add histo metric for IP allocation latency per unique pod

rbtr:

**Reason for Change**:

Adds a data structure and metric for keeping track of IP allocation requests per pod and recording the total time taken to return an IP for that pod.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/restserver/restserver.go
```diff
@@ -34,6 +34,28 @@ var (
 	ncVersionURLs sync.Map
 )
 
+type lockedDurationSet struct {
+	sync.Mutex
+	waits map[string]time.Time
+}
+
+func (m *lockedDurationSet) Push(key string) {
+	m.Lock()
```
neaggarwMS, line 43: Why do we need separate lock? Can we update this while acquiring the httpRestService lock only?

cns/restserver/restserver.go
```diff
@@ -34,6 +34,28 @@ var (
...
+	defer m.Unlock()
+	_, ok := m.waits[key]
+	if !ok {
+		m.waits[key] = time.Now()
+	}
+}
+
+func (m *lockedDurationSet) Pop(key string) time.Duration {
+	m.Lock()
+	start := m.waits[key]
```
neaggarwMS, line 53: check if key is not found then ignore

cns/restserver/restserver.go
rbtr: then we increase lock contention, and block unrelated things trying to access other synchronized state. I think it's better to scope the mutex tightly to the datastructure that it is protecting access to.

cns/restserver/ipam.go
```diff
@@ -12,76 +12,97 @@ import (
...
+		logger.ResponseEx(service.Name, req, resp, resp.ReturnCode, err)
 		return
 	}
 
-	var podInfo cns.PodInfo
-	podInfo, resp.ReturnCode, resp.Message = service.validateIPConfigRequest(req)
+	podInfo, returnCode, message := service.validateIPConfigRequest(req)
+
+	// remove this pod from the waiting for IP allocation set, if it is present
+	_ = service.podsPendingIPAllocation.Pop(podInfo.Key())
```
neaggarwMS, line 95: Actually with data deleting during release wont give us complete data. AFAIK, ContainerD calls ADD and if ADD fails to provision then it calls DEL followed by another ADD. If that is true, then we will never be able to truely evaluate the actual time it took to allocate an IP

rbtr: I have updated the PR to use a custom queue-style structure for this:
- We will hold up to the last 250 (maxpods) unique pod IDs that we have seen, recording the timestamp that that pod was first seen.
- When an IP is allocated, we will pop the entry for that pod and record the duration since it was first seen.
- When the queue is full and we see a new pod, we will drop the oldest pod record by first-seen-timestamp and record the new pod.

PR #1030, opened 2021-09-22
Arm64 docker

pjohnst5:
**Reason for Change**:
This adds arm64 docker images for Azure CNS and Azure NPM.
Uses docker buildx to do this, which makes it possible for the clients to pull which arch image they need at runtime dynically, and keep the same tag for both amd64 and arm64. Read more about it here

https://wiki.onap.org/display/DW/Building+a+Platform-Agnostic+Container+Image
https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

### DIFF
--- a/.pipelines/npm/npm-conformance-tests.yaml
+++ b/.pipelines/npm/npm-conformance-tests.yaml
@@ -72,8 +72,8 @@ jobs:
   - script: |
       echo Tag: $(TAG)
       echo ResourceGroup: $(RESOURCE_GROUP)
+      docker run --privileged --rm tonistiigi/binfmt --install arm64
       VERSION=$(TAG) make azure-npm-image
-      docker push $IMAGE_REGISTRY/azure-npm:$(TAG)
     displayName: 'Build and Push NPM Image'
 
   - task: Docker@2
--- a/Makefile
+++ b/Makefile
@@ -250,8 +250,8 @@ ifeq ($(GOOS),linux)
 	--build-arg VERSION=$(VERSION) \
 	--build-arg CNS_AI_PATH=$(CNS_AI_PATH) \
 	--build-arg CNS_AI_ID=$(CNS_AI_ID) \
-	--platform=linux/amd64,linux/arm64 \
-	--push \
+	--platform=$(IMAGE_PLATFORM_ARCHES) \
+	--$(IMAGE_ACTION) \
 	.
 
 	echo $(AZURE_CNS_IMAGE):$(VERSION) > $(IMAGE_DIR)/$(CNS_IMAGE_INFO_FILE)
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,7 @@ CNI_IPAM_DIR = $(REPO_ROOT)/cni/ipam/plugin
 CNI_IPAMV6_DIR = $(REPO_ROOT)/cni/ipam/pluginv6
 CNI_TELEMETRY_DIR = $(REPO_ROOT)/cni/telemetry/service
 ACNCLI_DIR = $(REPO_ROOT)/tools/acncli
-TELEMETRY_CONF_DIR = $(REPO_ROOT)/telemetry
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -233,58 +219,42 @@ azure-cnm-plugin-image: azure-cnm-plugin
...
+	docker buildx create --use
+	docker buildx build \
 	--no-cache \
 	-f cns/Dockerfile \
 	-t $(AZURE_CNS_IMAGE):$(VERSION) \
 	--build-arg VERSION=$(VERSION) \
-	--build-arg CNS_AI_PATH=$(cnsaipath) \
+	--build-arg CNS_AI_PATH=$(CNS_AI_PATH) \
 	--build-arg CNS_AI_ID=$(CNS_AI_ID) \
+	--platform=linux/amd64,linux/arm64 \
```
rbtr, line 253: `--platform=$(IMAGE_PLATFORM_ARCHES)`?

npm/Dockerfile
```diff
@@ -1,6 +1,24 @@
-# Use a minimal image as a parent image
+# Build npm
+FROM golang:1.16-alpine AS builder
```
rbtr, line 2: up to 1.17, drop the -alpine

npm/Dockerfile
```diff
@@ -1,6 +1,24 @@
...
+WORKDIR /usr/local/src/npm
+
+# Copy the source
+COPY . .
+
+# Build npm
+RUN CGO_ENABLED=0 go build -v -o /usr/local/bin/azure-npm -ldflags "-X main.version="$VERSION" -X "$NPM_AI_PATH"="$NPM_AI_ID"" -gcflags="-dwarflocationlists=true" npm/cmd/*.go
+
+# Use a minimal image as a final image base
 FROM ubuntu:focal
```
rbtr, line 17: probably should be @vakalapa:
a whole ubuntu image seems a little heavy to just get net-utils, any reason this couldn't be alpine or smaller?

.pipelines/npm/npm-conformance-tests.yaml
```diff
@@ -72,8 +72,8 @@ jobs:
   - script: |
       echo Tag: $(TAG)
       echo ResourceGroup: $(RESOURCE_GROUP)
+      docker run --privileged --rm tonistiigi/binfmt --install arm64
```
rbtr, line 75: is this installing arm64 emu support on the host that's executing this pipeline step? that doesn't feel right

rbtr (commented):
some immediate comments

npm/Dockerfile
pjohnst5: I didn't try it, just copied what was there before, I can try alpine

.pipelines/npm/npm-conformance-tests.yaml
pjohnst5: I believe it is, I ran into a bug otherwise, here's the thread I followed
https://github.com/moby/buildkit/issues/1929#issuecomment-758269704

npm/Dockerfile
pjohnst5: Alpine failed the cyclonus tests

pjohnst5: We can look at this in another PR but for this scope I'm going to leave it with `ubuntu:focal`
alpine results:
image

.pipelines/npm/npm-conformance-tests.yaml
rbtr: i mean, if it works, it works, but it doesn't seem right to install things on the host during a pipeline run

pjohnst5: Mat's oof on leave, I will check with him once he's back on how to get this installed on the images pre-pipeline, in the meantime, I'll resolve this though to get our arm64 images to AKS

rbtr (approved):

PR #1031, opened 2021-09-22
Set MTU for veths based on VM interface

tamilmani1989:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/transparent_endpoint_linux_test.go
+++ b/network/transparent_endpoint_linux_test.go
@@ -1,0 +1,111 @@
+//+build linux
+
+package network
+
+import (
+	"testing"
+
+	"github.com/Azure/azure-container-networking/netio"
+	"github.com/Azure/azure-container-networking/netlink"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddEndpoints(t *testing.T) {
+	tests := []struct {
+		name       string
+		client     *TransparentEndpointClient
+		epInfo     *EndpointInfo
+		wantErr    bool
+		wantErrMsg string
+	}{
+		{
+			name: "Add endpoints",
+			client: &TransparentEndpointClient{
+				hostPrimaryIfName: "eth0",
+				hostVethName:      "azvhost",
+				containerVethName: "azvcontainer",
+				netlink:           netlink.NewMockNetlink(false, ""),
+				netioshim:         netio.NewMockNetIO(false, 0),
+			},
+			epInfo:  &EndpointInfo{},
+			wantErr: false,
+		},
+		{
+			name: "Add endpoints netlink fail",
+			client: &TransparentEndpointClient{
+				hostPrimaryIfName: "eth0",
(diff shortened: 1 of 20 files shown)


network/transparent_endpoint_linux_test.go
```diff
@@ -0,0 +1,111 @@
...
+				containerVethName: "azvcontainer",
+				netlink:           netlink.NewMockNetlink(false, ""),
+				netioshim:         netio.NewMockNetIO(true, 2),
+			},
+			epInfo:     &EndpointInfo{},
+			wantErr:    true,
+			wantErrMsg: "TransparentEndpointClient Error : " + netio.ErrMockNetIOFail.Error() + ":eth0",
+		},
+		{
+			name: "Add endpoints get interface fail for host veth",
```
tamilmani1989, line 72: have to updat description here to container veth

network/transparent_endpoint_linux_test.go
```diff
@@ -0,0 +1,111 @@
...
+				containerVethName: "azvcontainer",
+				netlink:           netlink.NewMockNetlink(false, ""),
+				netioshim:         netio.NewMockNetIO(true, 3),
+			},
+			epInfo:     &EndpointInfo{},
+			wantErr:    true,
+			wantErrMsg: "TransparentEndpointClient Error : " + netio.ErrMockNetIOFail.Error() + ":azvcontainer",
+		},
+		{
+			name: "get interface fail for container veth",
```
tamilmani1989, line 85: same as above

network/transparent_endpointclient_linux.go
```diff
@@ -64,33 +66,57 @@ func setArpProxy(ifName string) error {
...
 		log.Printf("Deleting old host veth %v", client.hostVethName)
 		if err = client.netlink.DeleteLink(client.hostVethName); err != nil {
 			log.Printf("[net] Failed to delete old hostveth %v: %v.", client.hostVethName, err)
 			return newErrorTransparentEndpointClient(err.Error())
 		}
 	}
 
+	primaryIf, err := client.netioshim.GetNetworkInterfaceByName(client.hostPrimaryIfName)
+	if err != nil {
+		return newErrorTransparentEndpointClient(err.Error())
```
vakalapa, line 79: Can we not continue if this fails ? we can just ignore setting MTU with a warning message. I am thinking setting MTU is a luxury than necessity and we should not fail the veth pair setup because we could not setup MTU, please let me know if my understanding is wrong ?

tamilmani1989: ya i agree. this feature not needed for all customers. also want to make sure we are not breaking existing behaviour. let me just log it and not return error

vakalapa (dismissed):
lgtm

vakalapa (approved):

PR #1032, opened 2021-09-22
[NPM] Modifying destroy ipsets logic

vakalapa:
Current logic of destroy IPSets flushes and destroy each ipset individually, this results in some ipsets being deleted before their references have been wiped out. This PR will now flush all the Ipsets and then will delete all the ipsets making sure the references are removed while deleting,

### DIFF
--- a/npm/metrics/ipsets.go
+++ b/npm/metrics/ipsets.go
@@ -63,7 +63,13 @@ func RemoveEntryFromIPSet(setName string) {
 	}
 }
 
-// RemoveAllEntriesFromIPSet sets the number of entries for ipset setName to 0.
+// ResetEntriesInIPSet sets the number of entries for ipset setName to 0.
+// It doesn't ever update the number of IPSets.
+func ResetEntriesInIPSet(setName string) {
+	numIPSetEntries.Add(-getEntryCountForIPSet(setName))
+}
+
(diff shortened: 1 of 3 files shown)


npm/metrics/ipsets.go
```diff
@@ -63,7 +63,13 @@ func RemoveEntryFromIPSet(setName string) {
 	}
 }
 
-// RemoveAllEntriesFromIPSet sets the number of entries for ipset setName to 0.
+// ResetEntriesInIPSet sets the number of entries for ipset setName to 0.
+// It doesn't ever update the number of IPSets.
+func ResetEntriesInIPSet(setName string) {
+	numIPSetEntries.Add(-getEntryCountForIPSet(setName))
```
huntergregory, line 69: We shouldn't have to update prometheus. For memory consumption and scraping considerations, we currently don't track set names which have entry counts of 0. If we decide to change this later, we should bake that change into the existing functions (RemoveAllEntriesFromIPSet and RemoveEntryFromSet)

huntergregory (changes requested):
IPSM change looks good. Added a comment about prometheus

npm/metrics/ipsets.go
vakalapa: Sounds good, let me remove metrics changes.

huntergregory (approved):

PR #1034, opened 2021-09-27
Ipsetmanager-update

huntergregory:
Adds logic for adding/removing sets to kernel in ipsetmanger and changes the dataplane API to not use IPSets


npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -23,261 +25,346 @@ func (iMgr *IPSetManager) exists(name string) bool {
...
+	return nil
+}
 
-		// delete IPSet from the list
-		delete(list.MemberIPSets, setName)
-		set.DecIpsetReferCount()
-		// Update metrics of the IpSet
-		metrics.RemoveEntryFromIPSet(listName)
+func (iMgr *IPSetManager) addMemberIPSet(listName, memberName string) {
+	list := iMgr.setMap[listName]
```
vakalapa, line 323: can we send in a pointer of the list obj, that way we are using up memory and waiting for GC to  clean it up ?

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -23,261 +25,346 @@ func (iMgr *IPSetManager) exists(name string) bool {
...
+	listIsInKernel := list.shouldBeInKernel()
+	if listIsInKernel {
+		iMgr.incKernelReferCountAndModifyCache(member)
 	}
-	iMgr.updateDirtyCache(listName)
+}
 
-	return nil
+func (iMgr *IPSetManager) removeMemberIPSet(listName, memberName string) {
+	list := iMgr.setMap[listName]
```
vakalapa, line 336: same here send in a pointer of the obj

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -23,261 +25,346 @@ func (iMgr *IPSetManager) exists(name string) bool {
...
+			return npmerrors.Errorf(npmerrors.AppendIPSet, false, fmt.Sprintf("ipset %s is already a member of ipset %s", setName, listName))
 		}
+	}
+	return nil
+}
 
-		list, exists := iMgr.setMap[listName] // check if the Set exists
-		if !exists {
-			return npmerrors.Errorf(npmerrors.DeleteIPSet, false, fmt.Sprintf("ipset %s does not exist", listName))
+func (iMgr *IPSetManager) checkForRemoveFromListErrors(listName string, setNames []string) error {
```
vakalapa, line 280: lets remove this function and also checkForAddToListErrors, then can be part of their respective callers

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -23,261 +25,346 @@ func (iMgr *IPSetManager) exists(name string) bool {
...
+		return npmerrors.Errorf(npmerrors.DeleteIPSet, false, fmt.Sprintf("ipset %s cannot be deleted", name))
 	}
 
+	// the set will not be in the kernel since there are no references, so there's no need to update the dirty cache
 	delete(iMgr.setMap, name)
+	metrics.DecNumIPSets()
 	return nil
 }
 
 func (iMgr *IPSetManager) DeleteSet(name string) error {
```
vakalapa, line 366: I think we can have a single delete function like createIPSet we can have deleteIPSet

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -23,261 +25,346 @@ func (iMgr *IPSetManager) exists(name string) bool {
...
+	return iMgr.createIPSet(setName, setType)
 }
 
-func (iMgr *IPSetManager) createIPSet(set *IPSet) error {
-	// Check if the Set already exists
-	if iMgr.exists(set.Name) {
-		// ipset already exists
-		// we should calculate a diff if the members are different
-		return nil
+func (iMgr *IPSetManager) createIPSet(setName string, setType SetType) error {
```
vakalapa, line 153: We had this smaller createIPSet so we can call it from other functions, if we are going to return an error when an ipset is missing, lets remove this function and consolidate into CreateIPSet

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -163,47 +175,77 @@ func getSetKind(setType SetType) SetKind {
...
 }
 
-func (set *IPSet) CanBeDeleted() bool {
-	return len(set.SelectorReference) == 0 &&
-		len(set.NetPolReference) == 0 &&
-		set.IpsetReferCount == 0 &&
+func (set *IPSet) canBeDeleted() bool {
+	return !set.usedByNetPol() &&
+		!set.referencedInList() &&
 		len(set.MemberIPSets) == 0 &&
```
vakalapa, line 229: check for null pointer here for MemberIPSets and IPPodKey in case of List, IPPodKey  may be nil and same with other way around

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -23,261 +25,346 @@ func (iMgr *IPSetManager) exists(name string) bool {
...
+		metrics.SetNumEntriesForKernelIPSet(setName, numEntries)
+	*/
+}
+
+// for the following two functions, we don't need to check the deletionDirtyCache
+// if a set is in that cache, then it had no references, and if it should be in the kernel later,
+// then one of the following functions will have been called for it:
+// - AddReference
+// - addMemberIPSet
+func (iMgr *IPSetManager) modifyCacheForKernelMemberUpdate(setName string, isAdd bool) {
```
vakalapa, line 65: Lets loose isAdd if we are not using, or put a log message for timebeing

vakalapa (changes requested):
We will need think about bulk inputs, i think bulk inputs will complicate the controllers. We can take it up as a separate work item.

npm/pkg/dataplane/ipsets/ipset.go
huntergregory: len(nil) is actually 0

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -23,261 +25,322 @@ func (iMgr *IPSetManager) exists(name string) bool {
...
+		additionOrUpdateDirtyCache: make(map[string]struct{}),
+		deletionDirtyCache:         make(map[string]struct{}),
 	}
 }
 
-func (iMgr *IPSetManager) updateDirtyCache(setName string) {
-	set, exists := iMgr.setMap[setName] // check if the Set exists
-	if !exists {
-		return
+func (iMgr *IPSetManager) modifyCacheForKernelRemoval(setName string) {
```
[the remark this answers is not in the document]
huntergregory: new order of external functions
- create set
- delete set
- add reference
- delete reference
- add to set
- remove from set
- add to list
- remove from list
- apply ipsets

npm/pkg/dataplane/ipsets/ipsetmanager.go
huntergregory: this change would help 2 functions and for the other 2 it would use up memory i.e. add in the code `set := iMgr.setMap[setName]`

Left the change out for now to simplify changes

vakalapa: addMemberIPSet this function seems to be used only in one place, so we should be able to send in the pointer right ?

huntergregory: oh I was actually looking at `modifyCacheForKernelMemberUpdate()`

this and the function referenced in the comment below are called from functions that don't have the pointers, so we would need to add the code `set := iMgr.setMap[setName]` to them, which adds more for GC to clean up?

vakalapa (commented):
we do not have to fix these now, just for note keeping purposes

npm/pkg/dataplane/ipsets/ipsetmanager.go
vakalapa: that would be good too, since we are calling that function multiple times, a set obj pointer would need to be created multiple times we can get away with not duplicating by passing pointer

huntergregory: we are going to remove the for loop in a followup PR though?

vakalapa (approved):
more iterations to come!

PR #1035, opened 2021-09-28
Reduce lock timeout and check pid same before cleaning up

tamilmani1989:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/plugin.go
```diff
@@ -20,8 +20,11 @@ import (
 	cniTypes "github.com/containernetworking/cni/pkg/types"
 	cniTypesCurr "github.com/containernetworking/cni/pkg/types/current"
 	cniVers "github.com/containernetworking/cni/pkg/version"
+	"github.com/pkg/errors"
 )
 
+var errEmptyContent = errors.New("Read content is zero bytes")
```
rbtr, line 26: error strings should begin lowercase
```suggestion
var errEmptyContent = errors.New("read content is zero bytes")
```

store/json.go
```diff
@@ -25,7 +25,7 @@ const (
 	lockExtension = ".lock"
 
 	// Maximum number of retries before failing a lock call.
-	lockMaxRetries = 200
+	lockMaxRetries = 100
```
rbtr, line 28: how quickly do we retry? this could go by very fast. should there be some exponential back-off, maybe?

tamilmani1989: Isn't 10 seconds long enough? 100 retries with 100ms delay

rbtr: I don't know. Usually (like with http calls), we would do an exponential back-off where we wait maybe twice as long every time instead of a fixed interval (ex: 50ms, 100, 200, 400, 800, ...). But if 100x100ms is what we want, that's fine with me

cni/plugin.go
```diff
@@ -197,26 +200,30 @@ func (plugin *Plugin) IsSafeToRemoveLock(processName string) (bool, error) {
...
+			return false, errors.Wrap(err, "IsSafeToRemoveLock lockfile read failed")
 		}
 
-		log.Printf("Read from Lock file:%s", content)
+		log.Printf("Read from Lock file:%s", lockFilePid)
 		// Get the process name if running and
 		// check if that matches with our expected process
-		pName, err := platform.GetProcessNameByID(string(content))
+		pName, err := platform.GetProcessNameByID(lockFilePid)
 		if err != nil {
```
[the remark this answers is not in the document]
tamilmani1989: will add comment but i dont want to change logic now since i need to verify both linux and windows. anyway this code will go away once we move to filelock api.

cni/plugin.go
```diff
@@ -197,26 +200,33 @@ func (plugin *Plugin) IsSafeToRemoveLock(processName string) (bool, error) {
...
-			return false, fmt.Errorf("Num bytes read from lock file is 0")
+			return false, errors.Wrap(err, "IsSafeToRemoveLock lockfile read failed")
 		}
 
-		log.Printf("Read from Lock file:%s", content)
+		log.Printf("Read from lockfile:%s", lockFilePid)
 		// Get the process name if running and
 		// check if that matches with our expected process
-		pName, err := platform.GetProcessNameByID(string(content))
+		// if it runs non-nil error then process is not running
```
ashvindeodhar, line 213: nit: returns

ashvindeodhar (approved):

cni/plugin.go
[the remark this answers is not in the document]
rbtr: @ashvindeodhar I disagree. Semantically, `GetProcessNameByID` should only *not* return an error if and only if a process name is found and returned. If a process name in not found, it will return an error, because that's the contract implied by the method name, so the most correct thing to check is if we received an error or not.
This is pretty standard in Go and you can see examples of it in e.g. the controller-runtime client, where the `Get` method will return an error of type ErrNotFound

PR #1036, opened 2021-09-28
fix: Add DNS NAT Policy for Windows AKS-Swift Scenario

jaer-tsun:
**Reason for Change**:
Adding DNS NAT Policy for Windows AKS-Swift Scenario

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pjohnst5:
@jaer-tsun Thanks for the PR, do you know what endpoints we use of the `169.254.169.254` (IMDS) ip?

I just know that anything other than `169.254.169.254/metadata` is blocked now

pjohnst5:
@tamilmani1989 can you take a peek at this PR? 

Basically we're saying now, that if CNI windows is Aks-Swift (aka, not multitenancy, and CNS as Ipam), then add two snat rules for IMDS and DNS Ips to be routed through the host

tamilmani1989:
@neaggarwMS as we discussed long back, can we set execution mode to determine if cni running in aks-swift or not.. instead of checking checking multitenancy and ipam type as azure-cns?

PR #1037, opened 2021-09-28
chore: moving order of GOOS and GOARCh env sets in makefile

pjohnst5:
**Reason for Change**:
Fixes the order of the makefile's GOOS and GOARCH setting, if you're just running this on a windows host for example it's necessary, but like we were before with `make all-binaries-platforms` it wasn't necessary/failing, just if only on windows


pjohnst5:
@rbtr this is a small change, just for a windows only case, the Go env vars were set incorrectly, so fixing that

rbtr (approved):

PR #1040, opened 2021-09-30
refactor: move existing controllers to v1

matmerr:
**Reason for Change**:

persist existing behavior but leave room for v2 DPI

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


npm/pkg/controlplane/controllers/v2/namespacecontroller.go
```diff
@@ -0,0 +1,459 @@
...
+		klog.Infof("Adding namespace %s to ipset list %s", corev1NsName, labelIpsetName)
+		if err = nsc.dp.AddToList(labelIpsetName, []string{corev1NsName}); err != nil {
+			metrics.SendErrorLogAndMetric(util.NSID, "[AddNamespace] Error: failed to add namespace %s to ipset list %s with err: %v", corev1NsName, labelIpsetName, err)
+			return err
+		}
+
+		// Append succeeded labels to the cache NS obj
+		npmNs.appendLabels(map[string]string{nsLabelKey: nsLabelVal}, AppendToExistingLabels)
+	}
+
```
vakalapa, line 345: At the end of all these updates to IpSets, in NS controller we will need to call "dp.ApplyDataPlane()" this calculates needed updates and applies them, until then all the changes will be in only staged in cache.

vakalapa: This ApplyDataplane call should be done at the end of all 3 events for both controllers

npm/pkg/controlplane/controllers/v2/podcontroller.go
```diff
@@ -0,0 +1,633 @@
...
+		cachedNpmPod.removeContainerPorts()
+
+		// Add new pod's named ports from its ipset.
+		if err = c.manageNamedPortIpsets(newPodPorts, podKey, newPodObj.Status.PodIP, addNamedPort); err != nil {
+			return fmt.Errorf("[syncAddAndUpdatePod] Error: failed to add pod to named port ipset with err: %v", err)
+		}
+		cachedNpmPod.appendContainerPorts(newPodObj)
+	}
+	cachedNpmPod.updateNpmPodAttributes(newPodObj)
+
```
vakalapa, line 516: Along with ApplyDataplane, we should be calling "UpdatePod" for both Add and Update pod, delete we do not care as the rules will be removed

vakalapa (commented):
initial comments

npm/pkg/controlplane/controllers/v2/podcontroller.go
```diff
@@ -0,0 +1,633 @@
...
+func (c *PodController) syncAddedPod(podObj *corev1.Pod) error {
+	klog.Infof("POD CREATING: [%s%s/%s/%s%+v%s]", string(podObj.GetUID()), podObj.Namespace,
+		podObj.Name, podObj.Spec.NodeName, podObj.Labels, podObj.Status.PodIP)
+
+	var err error
+	podNs := util.GetNSNameWithPrefix(podObj.Namespace)
+	podKey, _ := cache.MetaNamespaceKeyFunc(podObj)
+	// Add the pod ip information into namespace's ipset.
+	klog.Infof("Adding pod %s to ipset %s", podObj.Status.PodIP, podNs)
+	if err = c.dp.AddToSet([]string{podNs}, podObj.Status.PodIP, podKey); err != nil {
```
vakalapa, line 363: Earlier addtoSet or any of theother functions would create ipset if it is missing, but the behavior now is DP expects a createSet call before any of these helper functions,

Issue #774, opened 2021-01-29
FailedCreatePodSandBox: Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox


jsturtevant:
**What happened:** 

CNI failed to set up network:

```
Jan 29 06:41:25.617: INFO: At 2021-01-29 06:36:50 +0000 UTC - event for webserver-deployment-5575796f8f-k92dd: {kubelet 2884k8s001} Started: Started container httpd
Jan 29 06:41:25.617: INFO: At 2021-01-29 06:37:08 +0000 UTC - event for webserver-deployment-5575796f8f-9jj9z: {kubelet 2884k8s000} FailedCreatePodSandBox: Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "678fa65619f0dd4c0434d2b4e510bd99930bc62e8ea4b64ea8a5f79a73b20f6a": unexpected end of JSON input
Jan 29 06:41:25.617: INFO: At 2021-01-29 06:37:10 +0000 UTC - event for webserver-deployment-5575796f8f-6rbcl: {kubelet 2884k8s000} FailedCreatePodSandBox: Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "cb4c4d0e705d57ac9da1eae9a42e2bb421dca03484fa8e0260672425feb55f15": unexpected end of JSON input
Jan 29 06:41:25.617: INFO: At 2021-01-29 06:37:10 +0000 UTC - event for webserver-deployment-5575796f8f-dtc4x: {kubelet 2884k8s000} FailedCreatePodSandBox: Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "a20980651ba231f652be02984d594aa661c613d440b8dd740d4073f2bfa4d50b": unexpected end of JSON input
Jan 29 06:41:25.617: INFO: At 2021-01-29 06:37:10 +0000 UTC - event for webserver-deployment-5575796f8f-jjdlr: {kubelet 2884k8s000} FailedCreatePodSandBox: Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "ba151015cc98dd643d424bf2eafaa71855fb4fb6b75fbb24bb4ea2616f964faf": unexpected end of JSON input
[...]

PR #1041, opened 2021-09-30
Remove lock files on reboot not needed

tamilmani1989:
**Reason for Change**:

No need to remove lock files on reboot as like state files. This call causing unnecessary open to lock file.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/plugin.go
```diff
@@ -158,9 +158,6 @@ func (plugin *Plugin) InitializeKeyValueStore(config *common.PluginConfig) error
 			log.Printf("[cni] Failed to create store: %v.", err)
 			return err
 		}
-
-		// Force unlock the json store if the lock file is left on the node after reboot
-		removeLockFileAfterReboot(plugin)
```
ashvindeodhar, line 163: can you remove the func definition for this as we don't need this anymore? You can remove it from linux as well. ( the one in linux is just printing the reboot time )

func removeLockFileAfterReboot(plugin *Plugin) {
	rebootTime, _ := platform.GetLastRebootTime()
	log.Printf("[cni] reboot time %v", rebootTime)
}

ashvindeodhar (changes requested):

store/json.go
```diff
@@ -176,16 +176,20 @@ func (kvs *jsonFileStore) flush() error {
...
 	kvs.Mutex.Lock()
 	defer kvs.Mutex.Unlock()
 
 	if kvs.locked {
 		return ErrStoreLocked
 	}
 
-	var lockFile *os.File
-	var err error
-	lockPerm := os.FileMode(0o664) + os.FileMode(os.ModeExclusive)
```
jsturtevant, line 188: was this value `0o664` valid?

jsturtevant:
xref: https://github.com/Azure/azure-container-networking/issues/774 this should resolve that issue partially

store/json.go
tamilmani1989: i changed to 0664 which is valid.

cni/plugin.go
tamilmani1989: > xref: #774 this should resolve that issue partially

not sure if this fixes that issue.

tamilmani1989:
> xref: #774 this should resolve that issue partially

not sure if this fixes that issue

cni/plugin.go
tamilmani1989: lets have it for a while and will remove this once we move to filelock api

store/json.go
rbtr: @jsturtevant `0o644` is a legal (and unambiguous) representation of 0644, where the little `o` indicates octal

rbtr: @tamilmani1989 you uncorrected it, 0o644 is less ambiguous and therefore preferred

tamilmani1989: @rbtr oops. thanks for letting me know. i will revert this back

ashvindeodhar:
lgtm, you can merge once you get approval from Evan for filemode changes

ashvindeodhar (approved):

store/json.go
jsturtevant: @rbtr thanks for the explanation!

rbtr (approved):

PR #1044, opened 2021-10-01
fix: prevent ipampoolmonitor from reconciling before it receives a nodenetworkconfig

rbtr:

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -15,57 +15,63 @@ import (
 
 const defaultMaxIPCount = int64(250)
 
-type CNSIPAMPoolMonitor struct {
-	MaximumFreeIps           int64
-	MinimumFreeIps           int64
-	cachedNNC                v1alpha.NodeNetworkConfig
+type PoolMon struct {
+	nnc                      *v1alpha.NodeNetworkConfig
```
neaggarwMS, line 19: Can we rename it back to cachednnc?

cns/ipampoolmonitor/ipampoolmonitor.go
```diff
@@ -261,53 +258,61 @@ func (pm *CNSIPAMPoolMonitor) createNNCSpecForCRD() v1alpha.NodeNetworkConfigSpe
...
+	if int(nnc.Spec.RequestedIPCount) == allocatedIPs {
 		// observe elapsed duration for IP pool scaling
 		metric.ObserverPoolScaleLatency()
 	}
 
-	logger.Printf("[ipam-pool-monitor] Update spec %+v, pm.MinimumFreeIps %d, pm.MaximumFreeIps %d",
-		pm.cachedNNC.Spec, pm.MinimumFreeIps, pm.MaximumFreeIps)
+	// defer closing the init channel to signify that we have received at least one NodeNetworkConfig.
+	defer pm.once.Do(func() { close(pm.initialized) })
+	pm.nncSource <- *nnc
```
neaggarwMS, line 288: Can you add a log line with updated nnc?

cns/ipampoolmonitor/ipampoolmonitor.go
neaggarwMS, line 19: Why do we need to persist the whole nnc object (which can include upto 250 ips)?

neaggarwMS: Just Scaler and Spec is enough, no?

cns/service/main.go
```diff
@@ -813,7 +813,7 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 	}
 
 	// initialize the ipam pool monitor
-	httpRestServiceImplementation.IPAMPoolMonitor = ipampoolmonitor.NewCNSIPAMPoolMonitor(httpRestServiceImplementation, requestController)
+	httpRestServiceImplementation.IPAMPoolMonitor = ipampoolmonitor.NewPoolMon(httpRestServiceImplementation, requestController)
```
neaggarwMS, line 816: nit: Can we rename it to NewAutoScalerMonitor (NewPoolMon is too generic, IMO)

neaggarwMS (changes requested):
:clock1:

cns/service/main.go
rbtr: It can't be used without its package qualifier `ipampoolmonitor.` so it's already unambiguous. It's arguably better Go style to go the other way and make it just `ipampool.Monitor`.
See[ `http.Client`](https://pkg.go.dev/net/http#Client) for similar convention.

PR #1045, opened 2021-10-04
[NPM] generic Policy Manager and some windows specific policy bits

vakalapa:


npm/podController.go
```diff
@@ -38,6 +38,7 @@ type NpmPod struct {
 	Name           string
 	Namespace      string
 	PodIP          string
+	NodeName       string
```
vakalapa, line 41: Remove nodeName not needed

npm/pkg/dataplane/dataplane.go
```diff
@@ -3,32 +3,55 @@ package dataplane
...
 	Name string
 	ID   string
+	IP   string
 	// Map with Key as Network Policy name to to emulate set
 	// and value as struct{} for minimal memory consumption
 	NetPolReference map[string]struct{}
 }
 
-func NewDataPlane() *DataPlane {
+// UpdateNPMPod pod controller will populate and send this datastructure to dataplane
```
vakalapa, line 37: Move this to dataplane, to avoid circular dependency

npm/pkg/dataplane/dataplane.go
```diff
@@ -3,32 +3,55 @@ package dataplane
...
+	PolicyActionAdd    PolicyAction = "add"
+	PolicyActionRemove PolicyAction = "remove"
+	AzureNetworkName                = "azure"
 )
 
 type DataPlane struct {
 	policyMgr *policies.PolicyManager
 	ipsetMgr  *ipsets.IPSetManager
 	networkID string
+	nodeName  string
```
vakalapa, line 23: remove nodename

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -1,33 +1,164 @@
...
+			toRemovePolicies[policyName] = struct{}{}
+		}
+	}
 
+	// Now check if any of these network policies are applied on this endpoint.
+	// If yes then proceed to delete the network policy
+	for policyName := range toRemovePolicies {
+		if _, ok := endpoint.NetPolReference[policyName]; ok {
+			// Delete the network policy
+			err := dp.policyMgr.RemovePolicy(policyName, []string{endpoint.ID})
```
vakalapa, line 76: delete this netpol reference from endpoint, do we need a separate loop for this ?

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -150,6 +151,53 @@ func (set *IPSet) GetSetContents() ([]string, error) {
 	}
 }
 
+// ShallowCompare check if the properties of IPSets are same
+func (set *IPSet) ShallowCompare(newSet *IPSet) bool {
```
huntergregory, line 155: Where do we plan to use this? I removed similar code since we don't take in ipsets as parameters now (have to do calls like addreference, addmember, etc.). 

For reference, I wrote this beforehand. Your checks for ippodkey and memberipsets are a little off, and Evan and I prefer not having if-bool-return-bool code. 

```
func ShallowCompare(set1 *IPSet, set2 *IPSet) bool {
	return set1.Name == set2.Name &&
		set1.Type == set2.Type &&
		set1.Kind == set2.Kind
}

func AreSameSet(set1 *IPSet, set2 *IPSet) bool {
	return ShallowCompare(set1, set2) && 
		haveSameSetContents(set1, set2)
}

func haveSameSetContents(set1 *IPSet, set2 *IPSet) bool {
	return haveSameCounts(set1, set2) ||
		haveSameIPPodKeys(set1, set2) ||
		haveSameMemberIPsets(set1, set2)
}

func haveSameCounts(set1 *IPSet, set2 *IPSet) bool {
	return set1.ipsetReferCount == set2.ipsetReferCount && // remove refer count or add kernel refer count too?
		len(set1.NetPolReference) == len(set2.NetPolReference) &&
		len(set1.SelectorReference) == len(set2.SelectorReference) &&
		len(set1.IPPodKey) == len(set2.IPPodKey) &&
		len(set1.MemberIPSets) == len(set2.MemberIPSets)
}

// assumes sets have same number of IPs
func haveSameIPPodKeys(set1 *IPSet, set2 *IPSet) bool {
	for ip, podKey1 := range set1.IPPodKey {
		podKey2, exists := set2.IPPodKey[ip]
		if !exists || podKey1 != podKey2 {
			return false
		}
	}
	return true
}

// assumes sets have same number of members
func haveSameMemberIPsets(set1 *IPSet, set2 *IPSet) bool {
	for memberName := range set1.MemberIPSets {
		_, exists := set2.MemberIPSets[memberName]
		if !exists {
			return false
		}
	}
	return true
}
```

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -208,6 +232,46 @@ func (iMgr *IPSetManager) ApplyIPSets(networkID string) error {
...
+}
+
+func (iMgr *IPSetManager) GetNetworkPolicyReferencesBySet(setName string) (map[string]struct{}, error) {
+	iMgr.Lock()
+	defer iMgr.Unlock()
+	if !iMgr.exists(setName) {
+		return nil, fmt.Errorf("[ipset manager] selector ipset %s does not exist.", setName)
+	}
+	set := iMgr.setMap[setName]
+	return set.SelectorReference, nil
```
huntergregory, line 272: return set.NetPolReference?

npm/pkg/dataplane/dataplane.go
```diff
@@ -135,19 +183,126 @@ func (dp *DataPlane) AddPolicy(policies *policies.NPMNetworkPolicy) error {
...
+	for _, set := range sets {
+		// Check if any 2nd level IPSets are generated by Controller with members
+		// Apply members to the list set
+		if set.Kind == ipsets.ListSet {
+			if len(set.MemberIPSets) > 0 {
+				memberList := []string{}
+				for _, memberSet := range set.MemberIPSets {
+					memberList = append(memberList, memberSet.Name)
+				}
+				err := dp.ipsetMgr.AddToList(set.Name, memberList)
```
huntergregory, line 263: I'm confused to the purpose of this code? Is this code adding the members of a list to the list itself?

npm/pkg/dataplane/dataplane.go
```diff
@@ -135,19 +183,126 @@ func (dp *DataPlane) AddPolicy(policies *policies.NPMNetworkPolicy) error {
...
+	// Create IPSets first along with reference updates
+	for _, set := range sets {
+		dp.ipsetMgr.CreateIPSet(set.Name, set.Type)
+		err := dp.ipsetMgr.AddReference(set.Name, netpolName, referenceType)
+		if err != nil {
+			return err
+		}
+	}
+
+	// TODO is there a possibility for a list set of selector referencing rule ipset?
```
huntergregory, line 251: I'm not sure what this refers to

npm/pkg/dataplane/dataplane.go
huntergregory: We don't allow the controller's to modify the IPSet objects themselves so I don't think we need this loop

npm/pkg/dataplane/dataplane.go
```diff
@@ -135,19 +183,126 @@ func (dp *DataPlane) AddPolicy(policies *policies.NPMNetworkPolicy) error {
...
+		// TODO ignore set does not exist error
+		// TODO add delete ipset after removing members
+		err := dp.ipsetMgr.DeleteReference(set.Name, netpolName, referenceType)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Check if any list sets are provided with members to add
+	for _, set := range sets {
```
huntergregory, line 285: see note above about addIPSetReferences

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -1,33 +1,164 @@
...
+}
+
+func (dp *DataPlane) shouldUpdatePod() bool {
+	return true
+}
+
+// updatePod has two responsibilities in windows
+// 1. Will call into dataplane and updates endpoint references of this pod.
+// 2. Will check for existing applicable network policies and applies it on endpoint
+func (dp *DataPlane) updatePod(pod *UpdateNPMPod) error {
```
huntergregory, line 35: what is this struct?

huntergregory, line 35: I think this function is related to the efficient algorithm I designed. I haven't fully walked through the policy removal code here, but the algorithm I designed is the most efficient for adding policies to an endpoint

huntergregory (commented):
updatePod() can be more efficient using the algorithm I designed, and I think we might have some disconnect with AddIPSetReferences()

npm/pkg/dataplane/ipsets/ipset.go
vakalapa: Yes i was thinking in the same lines, we could calculate the HASH of a given struct to determine if they are same. But this overly complicates existing scenario of UPDATE policy. So i tweaked the design to now Remove existing cached policy and Apply the new updated policy. This keeps things super simple for us while determining what dependencies to take care of.

npm/pkg/dataplane/dataplane.go
```diff
@@ -44,21 +67,14 @@ func (dp *DataPlane) ResetDataPlane() error {
...
-		return fmt.Errorf("[DataPlane] error while creating set: %w", err)
-	}
-	return nil
+func (dp *DataPlane) CreateIPSet(setName string, setType ipsets.SetType) {
+	dp.ipsetMgr.CreateIPSet(setName, setType)
 }
 
 // DeleteSet checks for members and references of the given "set" type ipset
 // if not used then will delete it from cache
 func (dp *DataPlane) DeleteIPSet(name string) error {
```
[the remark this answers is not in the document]
vakalapa: Done

npm/pkg/dataplane/ipsets/ipsetmanager.go
vakalapa: Sorry i corrected this, we need to return selectorreferences here, i corrected all variables in the caller function.

npm/pkg/dataplane/dataplane.go
vakalapa: IPSets in NPMnetworkPolicy object are what is TRANSLATED from the policy and not what is the actual state of an ipset in cache or dataplane. In Multivalue scenarios translate policy will return a NESTED IPSet with some labels as members, in that we will have to create this list and add relevant members to this new list.

npm/pkg/dataplane/dataplane_windows.go
vakalapa: When a pod add/update event happens, a PodIP gets added to multiple IPsets and can be removed from multiple Ips, this struct helps provide all those changes in one go so dataplane can determine if any network policies should be removed or applied on this pods endpoints.

vakalapa: As discussed, that algo we discussed was before we had IPset -> selector netpol reference. Now that we have ipset to selector reference, code is much simpler.

npm/pkg/dataplane/dataplane_test.go
```diff
@@ -9,14 +9,145 @@ import (
...
 
 	if dp == nil {
 		t.Error("NewDataPlane() returned nil")
 	}
 
-	err := dp.CreateIPSet("test", ipsets.NameSpace)
+	dp.CreateIPSet("test", ipsets.NameSpace)
+}
+
+func TestInitializeDataPlane(t *testing.T) {
```
huntergregory, line 21: minor code style note: we can write the following code in a TestMain() function (called before all testing begins) and then reset the dataplane at the top of each test to prevent having this in every test:
```
	metrics.InitializeAll()
	dp := NewDataPlane("testnode")

	if dp == nil {
		t.Error("NewDataPlane() returned nil")
	}

	err := dp.InitializeDataPlane()
	if err != nil {
		t.Errorf("InitializeDataPlane() returned error %v", err)
	}
```

PR #1069, opened 2021-10-22
NPM: Linux Policy Manager

huntergregory:
redesign iptables flow for NPM and add/remove ACLs in batch OS calls

TODO in next iteration: 
- remove dead policy chains in background thread
- retry logic (abort adding a chain if there are errors)


npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+	return 0, nil
+}
+
+func (pMgr *PolicyManager) getCreatorForInitChains() *ioutil.FileCreator {
+	creator := pMgr.getNewCreatorWithChains(iptablesAzureChains)
+
+	// add AZURE-NPM chain rules
+	creator.AddLine("", nil, util.IptablesAppendFlag, util.IptablesAzureChain, util.IptablesJumpFlag, util.IptablesAzureIngressChain)
+
```
JungukCho, line 177: nitPick: remove unneeded newlines in this function if there is no specific reason.

JungukCho (approved):
Put minor comments and they can be revised in your next PR if they make sense.

PR #1072, opened 2021-10-28
FIX: upload higher resolution npm v2 chain design image

huntergregory:


JungukCho (approved):
Thank you for applying request.
lgtm!

PR #1078, opened 2021-11-02
[NPM] Netpol Controller v2 and v2 controller UTs with dpmock

vakalapa:
TBD


JungukCho:
Hi, @vakalapa 

If it is not complex to raise PR again, would you please separate this PR into multiple PRs?
I may be wrong, but with quick skimming through this PR, by separating this PR helps to understand PR and reduce review times. 
1) One PR to only move v1 networkpolicy controller and translation, and parse files.
2) One PR to only have v2 networkpolicy controller and its corresponding UTs
3) Not sure what they are now,b but it looks like they may be other controller's UTs.

vakalapa:
@JungukCho  Agreed, it is definitely convoluted, but that is a very valid suggestion in terms of readability and CR best practices. Let me cut this down into multiple PRs.

PR #1083, opened 2021-11-04
feat: [NPM] Support graceful shutdown in pod

### LINKED ISSUE #1066: NPM blocks graceful shutdown connections for pods in Terminating state
**What happened:**  
When a Pod is deleted, NPM does not allow any grace period and removes pod ip from all relevant IPSets resulting in blocking of graceful shutdown connections.

NPM Should not delete pod as soon as an update event comes with deletionTimeStamp set, it should wait until the delete event is called or wait for the  terminationGracePeriodSeconds to expire.

**How to reproduce it:**  
Check the linked issue for repro details.

JungukCho:
**Reason for Change**:
PodController in NPM removes ipset information as soon as Pod's status becomes “Terminating”. 
So, even though terminated Pod has some activities (e.g., sending traffic to other pods to notify something) during "Terminating" status, NPM does not allow this connection from the Pod.
Thus, NPM needs to delay cleanup process until Pod completes graceful shutdown.

Plus : started trying parallel UTs.

**Issue Fixed**:

Fixes #1066

**Requirements**:

- [x] includes documentation in acndoc
- [x] adds unit tests

**Notes**:
Only updated podController in v1 now.

### DIFF
--- a/npm/pkg/controlplane/controllers/v1/podController.go
+++ b/npm/pkg/controlplane/controllers/v1/podController.go
@@ -324,7 +324,9 @@ func (c *PodController) syncPod(key string) error {
 		return err
 	}
 
-	// If newPodObj status is either corev1.PodSucceeded or corev1.PodFailed or DeletionTimestamp is set, start clean-up the lastly applied states.
+	// TODO(jungukcho): update..
+	// If newPodObj status is either corev1.PodSucceeded or corev1.PodFailed or DeletionTimestamp is set,
+	// start clean-up the lastly applied states.
 	if isCompletePod(pod) {
 		if err = c.cleanUpDeletedPod(key); err != nil {
 			return fmt.Errorf("Error: %v when when pod is in completed state.\n", err)
@@ -591,7 +593,7 @@ func (c *PodController) manageNamedPortIpsets(portList []corev1.ContainerPort, p
 }
 
 func isCompletePod(podObj *corev1.Pod) bool {
-	if podObj.DeletionTimestamp != nil {
+	if podObj.DeletionTimestamp != nil && podObj.DeletionGracePeriodSeconds != nil && *podObj.DeletionGracePeriodSeconds == 0 {
 		return true
 	}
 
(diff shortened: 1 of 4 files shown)


vakalapa (approved):

npm/pkg/controlplane/controllers/v1/podController_test.go
```diff
@@ -721,3 +804,71 @@ func TestWorkQueue(t *testing.T) {
...
 }
+
+func TestNPMPodNoUpdate(t *testing.T) {
+	type podInfo struct {
+		podName        string
+		ns             string
+		rv             string
+		podIP          string
+		labels         map[string]string
+		isHostNewtwork bool
```
huntergregory, line 815: nitpick typo: extra "w" in name

npm/pkg/controlplane/controllers/v1/podController_test.go
```diff
@@ -690,6 +691,88 @@ func TestHasValidPodIP(t *testing.T) {
 	}
 }
 
+func TestIsCompletePod(t *testing.T) {
+	var zeroGracePeriod int64
+	var defaultGracePeriod int64 = 30
+
+	type podState struct {
+		Phase                      corev1.PodPhase
```
huntergregory, line 699: nitpick: doesn't matter, but is it Go convention to lowercase local struct fields?

huntergregory (commented):
lgtm

npm/pkg/controlplane/controllers/v1/podController_test.go
JungukCho: Good catch

npm/pkg/controlplane/controllers/v1/podController_test.go
JungukCho: I am not sure, but I think it is good to have lowercase.

rbtr: it's not convention, the casing of struct fields has the same syntactic meaning as public/private keywords in other languages. 
this is a field in a private type, it should be private (lowercase).

JungukCho: So, I addressed it in a right way!

npm/pkg/controlplane/controllers/v1/podController.go
```diff
@@ -82,6 +82,16 @@ func (nPod *NpmPod) updateNpmPodAttributes(podObj *corev1.Pod) {
 	}
 }
 
+// noUpdate evaluates whether NpmPod is required to be update given podObj.
+func (nPod *NpmPod) noUpdate(podObj *corev1.Pod) bool {
+	return nPod.Namespace == podObj.ObjectMeta.Namespace &&
+		nPod.Name == podObj.ObjectMeta.Name &&
+		nPod.Phase == podObj.Status.Phase &&
+		nPod.PodIP == podObj.Status.PodIP &&
+		reflect.DeepEqual(nPod.Labels, podObj.ObjectMeta.Labels) &&
```
rbtr, line 91: Please don't import `reflect` outside of tests unless you're doing meta-programming. It's expensive to use, and the result of `DeepEqual` specifically is sometimes (at the worst times) not what you would expect.
NPM can't spare this performance 🥲

rbtr (changes requested):

npm/pkg/controlplane/controllers/v1/podController.go
JungukCho: @rbtr Good catch. We also recently found performance issues using `reflect`. I think this is good time to revise it in this PR.

rbtr (approved):

JungukCho:
Will replace `IsSameLabels` with [Equals](https://github.com/kubernetes/apimachinery/blob/460d10991a520527026863efae34f49f89c2f4e1/pkg/labels/labels.go#L129) from official K8s in a next PR

PR #1084, opened 2021-11-04
stupid simple retry

paulgmiller:
**Reason for Change**:
apiserver nnc might not be registered or api server might be down and crashloop backoff puts us outside of 5-10 minutes we have for aks addons to come up so retry a bit more aggresively here.

We could also exponential backoff here but wouldn't want ot go above 1 minute or we'd be in same place. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


paulgmiller:
replaced with avas #1087

PR #1085, opened 2021-11-04
feat: [NPM] ipset save before restoring and fixing grep UTs

huntergregory:
# Changes for Applying IPSets
- use ipset save file to add/delete members
- update error handling logic to skip previously run lines in a restore file

## Design Decision
Instead of calculating which members should be removed from ipsets, ipset manager will use the current kernel state via ipset save and apply the cache as the expected state. 

### Pros
- for the dirty sets, the kernel will always become exactly as we desire (delete everything not desired from the kernel i.e. no leftover set members that aren't in the cache)
- we are effectively reconciling and mitigating security concerns for dirty sets
- no need to modify cache management at this time

### Cons
- increased compute and latency 
- added complexity and perhaps harder to maintain

### Looking Forward
We will eventually have an infrequent reconcile routine which will reconcile the whole cache with the kernel state.

We will assess the extra latency of this design choice later.

# Other changes
- Allow proper exit codes and stdouts for piped commands in UTs.

TODO: 
- move some UT code into smaller-scoped functions for less complexity


npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -29,153 +67,409 @@ func (iMgr *IPSetManager) resetIPSets() error {
...
-			methodFlag = util.IpsetSetListFlag
-		} else if set.Type == NamedPorts {
-			methodFlag = util.IpsetIPPortHashFlag
-		}
+// updates the creator (adds/deletes members) for dirty sets already in the kernel
+// updates the toAddOrUpdateCache: after calling this function, the cache will only consist of sets to create
+// error handling principal:
+// - if contract with ipset save (or grep) is breaking, salvage what we can, take a snapshot without grep, and log the failure
+// - have a background process for sending/removing snapshots intermittently
+func (iMgr *IPSetManager) updateDirtyKernelSets(saveFile []byte, creator *ioutil.FileCreator) {
```
JungukCho, line 205: Does this function walk through all ipset information stored in kernel?
If so, is it necessary?
If there are large number of ipsets are programmed in kernel, this cost will be high?

Do we leverage `-exist` option for `add` and `del` to make it simple?
```shell
# from IPSet man page
add SETNAME ADD-ENTRY [ ADD-OPTIONS ]
Add a given entry to the set. If the -exist option is specified, ipset ignores if the entry already added to the set.
del SETNAME DEL-ENTRY [ DEL-OPTIONS ]
Delete an entry from a set. If the -exist option is specified and the entry is not in the set (maybe already expired), then the command is ignored.
```

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -29,153 +67,409 @@ func (iMgr *IPSetManager) resetIPSets() error {
...
+		createMatches := nameForCreateRegex.FindSubmatch(createLine)
+		if len(createMatches) == lengthForSuccessfulMatch {
+			return
+		}
+	}
+	return
+}
+
+func haveTypeProblem(set *IPSet, restOfCreateLine []byte) bool {
+	// TODO check maxelem for hash sets? CIDR blocks have a different maxelem
```
JungukCho, line 314: I am not sure about `TODO`, but just context of maxelem for CIDR is the maximum number of ip address for the hashset -> 2^32 -1 -> 4,294,967,295 based on my understanding. This only works for `ipv4` though

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -90,9 +90,8 @@ func (pMgr *PolicyManager) initializeNPMChains() error {
 // removeNPMChains removes the jump rule from FORWARD chain to AZURE-NPM chain
 // and flushes and deletes all NPM Chains.
 func (pMgr *PolicyManager) removeNPMChains() error {
-	deleteErrCode, deleteErr := pMgr.runIPTablesCommand(util.IptablesDeletionFlag, jumpFromForwardToAzureChainArgs...)
-	hadDeleteError := deleteErr != nil && deleteErrCode != couldntLoadTargetErrorCode
-	if hadDeleteError {
+	deleteErrCode, deleteErr := pMgr.runIPTablesCommandAndIgnoreErrorCode(couldntLoadTargetErrorCode, util.IptablesDeletionFlag, jumpFromForwardToAzureChainArgs...)
```
JungukCho, line 93: In deletion operation, is it ok not to get error code about `doesNotExistErrorCode` or `couldntLoadTargetErrorCode`, etc?
Anyway, the goal is to clean up the tables.

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -161,10 +164,11 @@ func (pMgr *PolicyManager) runIPTablesCommand(operationFlag string, args ...stri
 		errCode := exitError.ExitStatus()
 		allArgsString := strings.Join(allArgs, " ")
 		msgStr := strings.TrimSuffix(string(output), "\n")
-		if errCode > 0 && operationFlag != util.IptablesCheckFlag {
+		if errCode > 0 && errCode != errCodeToIgnore {
```
JungukCho, line 167: Can you check `removeNPMChains` case with `errCodeToIgnore`?
In case the chain does not exists, it is fine to proceed.

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -146,6 +145,10 @@ func (pMgr *PolicyManager) reconcileChains(stopChannel <-chan struct{}) {
 
 // this function has a direct comparison in NPM v1 iptables manager (iptm.go)
 func (pMgr *PolicyManager) runIPTablesCommand(operationFlag string, args ...string) (int, error) {
+	return pMgr.runIPTablesCommandAndIgnoreErrorCode(-1, operationFlag, args...)
+}
+
+func (pMgr *PolicyManager) runIPTablesCommandAndIgnoreErrorCode(errCodeToIgnore int, operationFlag string, args ...string) (int, error) {
```
JungukCho, line 151: It is a little hard to follow which operations are ok to ignore which errors.
Do we delegate error handling in a caller side?

We just have one `run` function and a caller will do right actions bases on return values.
It will be easier to follow up codes.

JungukCho:
One comment: naming (file-creator.go and its contents) is misleading if I correctly understand the code. IPset and iptables use `stdout`  to program them. Initially, I am looking for where is file.

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
huntergregory: we grep for all npm sets, so the work is linear in the number of sets in NPM. For non-dirty sets, we do skip reading all add lines in the save file, so this won't be too intensive. We wouldn't know what sets to delete without walking through the ipset save file, so `del --exist` couldn't work, and since we're already determining the members to delete for dirty sets, determining the exact members to add is equal work, and we write a smaller restore file compared to looping through all members and running `add --exist`

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -29,153 +67,409 @@ func (iMgr *IPSetManager) resetIPSets() error {
...
+-A new-set-2 2.3.4.5
+-A new-set-1 2.3.4.5
+-A new-set-1 1.2.3.4
+-A new-set-3 1.2.3.4
+-A new-set-3 2.3.4.5
+
+save file for example above:
+	create set-in-kernel-1 net:hash ...
+	add set-in-kernel-1 1.2.3.4
+	add set-in-kernel-1 8.8.8.8
```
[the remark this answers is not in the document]
huntergregory: rearranging this for clarity

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
huntergregory: clarified

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -2,21 +2,59 @@ package ipsets
...
+	// constants for parsing ipset save
+	ipsetAddString           = "add"
+	ipsetCreateString        = "create"
+	ipsetSetListString       = "list:set"
+	ipsetNetHashString       = "hash:net"
+	ipsetIPPortHashString    = ipsetIPPortHashFlag
+	lengthForSuccessfulMatch = 2
+
+	// creator constants
+	maxTryCount                    = 3
```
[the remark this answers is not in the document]
huntergregory: to my knowledge, there are only line-based errors in ipset restore. all lines before an error do get applied so we handle the error at a given line by restoring from the line after the failure (and removing any subsequent, related lines). So if an IP was malformed and we missed it, then the add for it will fail, we'll continue on, and the everything else will be applied as it should be

npm/pkg/dataplane/ioutil/file-creator.go
```diff
@@ -245,26 +261,24 @@ func (creator *FileCreator) handleLineError(lineNum int, commandString, stdErr s
...
-			if !exists {
-				log.Errorf("can't abort section because line references section %d which doesn't exist, so skipping the line instead", line.sectionID)
-				creator.lineNumbersToOmit[lineNumIndex] = struct{}{}
-			} else {
-				for _, lineNum := range section.lineNums {
-					creator.lineNumbersToOmit[lineNum] = struct{}{}
-				}
+		case Continue:
+			klog.Errorf("continuing after line %d for command [%s]", lineNum, commandString)
+			for k := 0; k <= lineNumIndex; k++ {
```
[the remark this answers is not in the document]
huntergregory: just habit haha I think I used to avoid i because it's the symbol for imaginary numbers. will change to i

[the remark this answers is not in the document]
huntergregory: just habit haha I think I used to avoid i because it's the symbol for imaginary numbers. will change to i

npm/pkg/dataplane/ioutil/grep.go
```diff
@@ -0,0 +1,35 @@
...
+	commandError = command.Start()
+	if commandError != nil {
+		return
+	}
+
+	// Without this wait, defunct iptable child process are created
+	wait := func() { _ = command.Wait() } // appease go lint
+	defer wait()
+
+	output, err := grepCommand.CombinedOutput()
```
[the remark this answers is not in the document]
huntergregory: no actually since grep returns an error if we don't find anything. I'd like to return commandError nil for this case

npm/pkg/dataplane/policies/chain-management_linux.go
huntergregory: The main consideration for ignoring an error code in this function is that this function logs when there's an error, but I think it shouldn't log an error if we ignore the error later. We previously had some of that code baked into run() for check operations, but I think we should also do it for deleting a rule.

npm/pkg/dataplane/policies/chain-management_linux.go
huntergregory: sorry, could you clarify this?

npm/pkg/dataplane/policies/chain-management_linux.go
huntergregory: if the AZURE-NPM chain doesn't exist, we'll get couldntLoadTargetErrorCode and if the chain exists but the rule doesn't exist, we'll get doesNotExistErrorCode. So this should ignore both of these codes actually, since this is called before initializing DP.

huntergregory:
> One comment: naming (file-creator.go and its contents) is misleading if I correctly understand the code. IPset and iptables use `stdout` to program them. Initially, I am looking for where is file.

Open to suggestions. How about restore-file-creator.go?

npm/pkg/dataplane/policies/chain-management_linux.go
huntergregory: if there's an error besides those two, something's wrong, but it could be several things, so ya maybe we just log and continue.

For instance, we get `couldntLoadTargetErrorCode` for "unknown option" and for a malformed IP address.

npm/pkg/dataplane/ioutil/file-creator.go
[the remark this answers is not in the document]
JungukCho: You are mathematician!

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
JungukCho: I may misunderstand the code, but in the code, it also  has string match. So the work is not linear to the number of sets. It is linear to the number of sets and its string in each set.

My fundamental question in the code is why we need to walk through all ipsets (read from kernel and store in `saveFile` variable) in every `applyIPSets`  call and derive some actions (add or deletion) even though we maintain a lot of data structures in  `ipsetmanager_linux.go`. Can we leverage those information and apply what we need?

npm/pkg/dataplane/policies/chain-management_linux.go
JungukCho: Seems you resolved it.

JungukCho:
> > One comment: naming (file-creator.go and its contents) is misleading if I correctly understand the code. IPset and iptables use `stdout` to program them. Initially, I am looking for where is file.
> 
> Open to suggestions. How about restore-file-creator.go?
May be restoreLinuxDP.go since both iptables and ipset uses this for mainly restore, but your suggestion is ok. 
Just having `file` makes me confused, but I may be so sensitive.

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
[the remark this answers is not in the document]
JungukCho: I am not sure, but failure will be cascading (For example, if `create` failed, adding elements will be failed). 

The other concern is that I am not sure it is easy to resolve failed one in a next reconcile.
If you simply handle only failed one in a next reconcile, I think it will be fine.

huntergregory:
> > > One comment: naming (file-creator.go and its contents) is misleading if I correctly understand the code. IPset and iptables use `stdout` to program them. Initially, I am looking for where is file.
> > 
> > 
> > Open to suggestions. How about restore-file-creator.go?
> > May be restoreLinuxDP.go since both iptables and ipset uses this for mainly restore, but your suggestion is ok.
> > Just having `file` makes me confused, but I may be so sensitive.

ok I'll call it restore_linux.go. Are you ok with the name FileCreator for the struct?

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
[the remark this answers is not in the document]
huntergregory: the current logic will handle cascading errors (e.g. if a create fails, then abandon all adds/deletes for that set)

eventually we will save these failures in a reconcile cache and handle separately

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
huntergregory: So I guess the decision is between these options:

1. read in save file for dirty sets
2. maintain copy of ipsets object from before they became dirty

We're doing the first here. The second would require extra memory (worst case, doubling the memory for the ipset cache before applying) and would require a redesign of the whole ipset manager. 

Thoughts @vakalapa ?

vakalapa: This operation is definitely very intensive wrt string matching, even if we skip over the section of some ipsets we will be doing first of line matching. 

We chose not to save deleted members of IPSets to save memory because we are already very constrained on memory in linux and data structures we are maintaining in Ipset manager is just source of truth. This is modelled against k8s philosophy, we have a source of truth (our internal ipset cache), we read the state on the machine and try to reconcile to that state.

vakalapa: Also, this approach is robust in the sense that kernel state will hardly deviate from expected state since NPM takes a snapshot of sets to be updated and then applies changes based on that, current IPSM's draw back is that as errors occur and as NPM runs for longer duration, kernel state heavily deviates from what is expected and this results in a cascading affect of issues, 

Hopefully in the future, when we figure out all the errors and have low-probability-for-error controller, then we can tone this approach down. But for initial purposes i feel, even with this compute intensive update, we are VERY fast compared to previous generation.

Wdygt?

JungukCho: If you guys think this is a way, I am ok with this approach.

I just wrote my two cents.
Based on my understanding, the goal of this enhancements in ipset in high-level is
1) Compared to v1 NPM, in v2 we do kind of batch process (one shot for one controller event) unlike in v1 NPM (sequentially called exec.run call for each ipset operation) to improve performance.
2) More fined-grained ipset management with reference counts to avoid known issues (e.g., `used in kernel`)
Do I miss something?

I may be wrong and miss some details,  but to me, it looks like that we took a kind of reverse engineering way, which is hard and complex way while we have enough information to derive all information since data plane is based on all information in control plane.
1) We know when and which operations we need to program (e.g., add, deletion, update) based on controller event) 
2) We know what information are programmed from ipset manger data-structures
3) All the references counts for ipsets
So, I think we can achieve the same functionality with those information in much simple way.

About performance, with a lot of ipsets, I am not sure how much we gains since we do string match O(# of ipsets * # of characters per ipset) in every event which needs applyIPset.

JungukCho:
> > > > One comment: naming (file-creator.go and its contents) is misleading if I correctly understand the code. IPset and iptables use `stdout` to program them. Initially, I am looking for where is file.
> > > 
> > > 
> > > Open to suggestions. How about restore-file-creator.go?
> > > May be restoreLinuxDP.go since both iptables and ipset uses this for mainly restore, but your suggestion is ok.
> > > Just having `file` makes me confused, but I may be so sensitive.
> 
> ok I'll call it restore_linux.go. Are you ok with the name FileCreator for the struct?

No strong opinion.

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
vakalapa: As we discussed offline, this current design is more secure than alternatives because even if some malicious actor changes the ipsets, NPM will try to reconcile to the desired state and overwrite manual entries. Granted NPM will only do this for "dirty" sets. A backlog item is added to have a background thread "at leisure" should try to reconcile all ipsets to the desired state, this design will make NPM more secure in the long run. 

Also, perf evaluation is planned on this current design to understand at scale what latency this compute intensive task is going to induce in rule programming.

npm/pkg/dataplane/policies/chain-management_linux.go
huntergregory: these changes will be moved to the next PR which updates policy/chain management

npm/pkg/dataplane/policies/chain-management_linux.go
huntergregory: these changes will be moved to the next PR which updates policy/chain management

npm/pkg/dataplane/policies/chain-management_linux.go
huntergregory: these changes will be moved to the next PR which updates policy/chain management

npm/config/config.go
```diff
@@ -17,7 +17,7 @@ var DefaultConfig = Config{
 		EnablePrometheusMetrics: true,
 		EnablePprof:             true,
```
JungukCho: Just check-in:
Based on our discussion, it would be better to set `EnablePprof` as `false`?

JungukCho: Oops. Reverted! nevermind.

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -29,153 +67,414 @@ func (iMgr *IPSetManager) resetIPSets() error {
...
-	creator := iMgr.getFileCreator(maxTryCount, toDeleteSetNames, toAddOrUpdateSetNames)
-	err := creator.RunCommandWithFile(util.Ipset, util.IpsetRestoreFlag)
-	if err != nil {
-		return fmt.Errorf("%w", err)
+	var saveFile []byte
+	var saveError error
+	if len(iMgr.toAddOrUpdateCache) > 0 {
+		saveFile, saveError = iMgr.ipsetSave()
+		if saveError != nil {
+			return fmt.Errorf("%w", saveError)
```
matmerr, line 144: can we wrap with context so we know the save error came from this stack

JungukCho (approved):
LGTM! Thank you for applying comments. 
We can walk about translation and linux dataplane before integration test.

vakalapa (approved):

PR #1086, opened 2021-11-04
fix: [NPM] reposition iptables jump to AZURE-NPM chain

### LINKED ISSUE #1088: Defalut deny in Azure network policy is not working with Loadbalancer service (internal/external)
**What happened:**  

Create a network policy to block all ingress and egress traffic to a namespace, it is not working for lb IP's, means lb IP's are able to access the resource of that namespace, though it has default deny policy.

I have tested same scenario with calcio, if we apply default deny policy it should block all traffic to that namespace whether its internal or external loadbalancer ingress

**How to reproduce it:**  

Create a namespace 

` kubectl create ns elf`

create a deployment in ns elf

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  namespace: elf
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2 # tells deployment to run 2 pods matching the template
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```

create internal lb

```
apiVersion: v1
kind: Service
metadata:
  name: nginx-svc
  namespace: elf
  annotations:
          service.beta.kubernetes.io/azure-load-balancer-internal: "true"
spec:
  type: LoadBalancer
  ports:
  - port: 80
  selector:
    app: nginx
```
create a default deny network policy (allowed port 53 for DNS)

```
kind: NetworkPolicy
metadata:
  name: default-deny-all
  namespace: elf
spec:
  egress:
  - ports:
    - port: 53
      protocol: TCP
    - port: 53
      protocol: UDP
  podSelector: {}
  policyTypes:
  - Ingress
  - Egress
```

create debug pod to test the traffic and traffic is allowed to that ns using lb IP.

```

kubectl exec -it ubuntu -- bash
curl <internall-lb-ip>
```

**Kubernetes Version:**  
AKS v1.20.9

**Kernel (e.g. `uname -a`):**  

```
#uname -a
Linux aks-agentpool-23187946-vmss000004 5.4.0-1061-azure #64~18.04.1-Ubuntu SMP Thu Oct 7 21:00:58 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

The same setup working fine with calico network policy, which means it should also block loadbalancer traffic.

huntergregory:
Fix #1088 

# Problem
In the iptables nat table, under certain conditions, KUBE-SERVICES will mark a packet sent to an LB service before DNAT to a pod's IP. Then, in iptables filter table the KUBE-FORWARD table will accept packets with this mark before the packet reaches AZURE-NPM chain.

As a result, we might not get a chance to drop packets that we should.

#### FORWARD chain (as it is now)
As it is now, NPM comes after KUBE-FORWARD.
```
Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
17386 1819K KUBE-FORWARD  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes forwarding rules */
17129 1806K KUBE-SERVICES  all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate NEW /* kubernetes service portals */
  845 43940 AZURE-NPM  all  --  *      *       0.0.0.0/0            0.0.0.0/0
17129 1806K KUBE-EXTERNAL-SERVICES  all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate NEW /* kubernetes externally-visible service portals */
    0     0 DROP       tcp  --  *      *       0.0.0.0/0            168.63.129.16        tcp dpt:80
```

#### KUBE-FORWARD chain 
Notice the accept on masquerade mark (0x4000).
```
Chain KUBE-FORWARD (1 references)
    0     0 DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate INVALID
    0     0 ACCEPT     all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes forwarding rules */ mark match 0x4000/0x4000
    0     0 ACCEPT     all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes forwarding conntrack pod source rule */ ctstate RELATED,ESTABLISHED
    0     0 ACCEPT     all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes forwarding conntrack pod destination rule */ ctstate RELATED,ESTABLISHED
```

### nat table (example with an ILB service named elf/nginx-svc)
For this example, elf/nginx-svc delegates traffic to two pods for incoming traffic on tcp port 80 or nodeport 30525.

KUBE-SERVICES is Kubernetes' nat chain. It is referenced in the OUTPUT AND PREROUTING chains.

Depending on the src, we may mark for masquerading if port 80 is used. Depending on the IP/port used for the svc, we either jump to the svc chain, fwd chain, or nodeports chain.
```
Chain KUBE-SERVICES (2 references)
      ...
    0     0 KUBE-MARK-MASQ  tcp  --  *      *      !10.240.0.0/12        ILB_CLUSTER_IP          /* elf/nginx-svc cluster IP */ tcp dpt:80
    0     0 KUBE-SVC-TQTTMDXT2ILLOFTL  tcp  --  *      *       0.0.0.0/0            ILB_CLUSTER_IP          /* elf/nginx-svc cluster IP */ tcp dpt:80
    0     0 KUBE-FW-TQTTMDXT2ILLOFTL  tcp  --  *      *       0.0.0.0/0            ILB_EXTERNAL_IP          /* elf/nginx-svc loadbalancer IP */ tcp dpt:80
      ...
 1453 79972 KUBE-NODEPORTS  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes service nodeports; NOTE: this must be the last rule in this chain */ ADDRTYPE match dst-type LOCAL
```

Here's the mark-for-masquerade chain:
```
Chain KUBE-MARK-MASQ (91 references)
  890 46280 MARK       all  --  *      *       0.0.0.0/0            0.0.0.0/0            MARK or 0x4000
```

Here's the LB's svc chain, which DNATs to one of the pods randomly. The packet is marked for masquerade (0x4000) if the src IP is the same as the dst pod's IP.
```
Chain KUBE-SVC-TQTTMDXT2ILLOFTL (3 references)
  449 23348 KUBE-SEP-7NIBX3P3TLQVYL3E  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* elf/nginx-svc */ statistic mode random probability 0.50000000000
  453 23556 KUBE-SEP-KJY2B5OMZRWCDXG5  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* elf/nginx-svc */

Chain KUBE-SEP-7NIBX3P3TLQVYL3E (1 references)
    0     0 KUBE-MARK-MASQ  all  --  *      *       POD2_IP          0.0.0.0/0            /* elf/nginx-svc */
  449 23348 DNAT       tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            /* elf/nginx-svc */ tcp to:POD2_IP:80

Chain KUBE-SEP-KJY2B5OMZRWCDXG5 (1 references)
    0     0 KUBE-MARK-MASQ  all  --  *      *       POD1_IP          0.0.0.0/0            /* elf/nginx-svc */
  453 23556 DNAT       tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            /* elf/nginx-svc */ tcp to:POD1_IP:80
```

If the traffic was sent to the LB's external IP, it gets forwarded (marked for masquerade before sent to above svc chain for DNAT).
```
Chain KUBE-FW-TQTTMDXT2ILLOFTL (1 references)
    0     0 KUBE-MARK-MASQ  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* elf/nginx-svc loadbalancer IP */
    0     0 KUBE-SVC-TQTTMDXT2ILLOFTL  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* elf/nginx-svc loadbalancer IP */
    0     0 KUBE-MARK-DROP  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* elf/nginx-svc loadbalancer IP */
```

If the nodeport was used, mark for masquerade and send to the svc chain)
```
Chain KUBE-NODEPORTS (1 references)
  902 46904 KUBE-MARK-MASQ  tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            /* elf/nginx-svc */ tcp dpt:30525
  902 46904 KUBE-SVC-TQTTMDXT2ILLOFTL  tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            /* elf/nginx-svc */ tcp dpt:30525
```

# Solution
When a toggle is turned on (default off), move the jump from FORWARD to AZURE-NPM chain above the jump to KUBE-FORWARD. 

As a result, packets DNAT to a pod from the ILB will pass through NPM instead of being accepted beforehand. 

## new FORWARD chain in filter table
We add a `ctstate NEW` requirement for the jump to Azure chain, and position the jump depending on the toggle. This follows KUBE-FORWARD's practice and is necessary so for example, we don't deny an HTTP response for an HTTP request that we allow.
 
When the toggle is set to place the azure chain first: 
```
Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
  845 43940 AZURE-NPM  all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate NEW
17386 1819K KUBE-FORWARD  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubernetes forwarding rules */
17129 1806K KUBE-SERVICES  all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate NEW /* kubernetes service portals */
17129 1806K KUBE-EXTERNAL-SERVICES  all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate NEW /* kubernetes externally-visible service portals */
    0     0 DROP       tcp  --  *      *       0.0.0.0/0            168.63.129.16        tcp dpt:80
```

Keep the rest the same, including the redundant check for state RELATED/ESTABLISHED in the final ACCEPT in Azure chain.
```
Chain AZURE-NPM (1 references)
  845 43940 AZURE-NPM-INGRESS  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
    0     0 AZURE-NPM-EGRESS  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
    0     0 AZURE-NPM-ACCEPT  all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x3000 /* ACCEPT-on-INGRESS-and-EGRESS-mark-0x3000 */
    0     0 AZURE-NPM-ACCEPT  all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x2000 /* ACCEPT-on-INGRESS-mark-0x2000 */
    0     0 AZURE-NPM-ACCEPT  all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match 0x1000 /* ACCEPT-on-EGRESS-mark-0x1000 */
    0     0 ACCEPT     all  --  *      *       0.0.0.0/0            0.0.0.0/0            state RELATED,ESTABLISHED /* ACCEPT-on-connection-state */
```

### DIFF
--- a/npm/config/config.go
+++ b/npm/config/config.go
@@ -18,7 +18,7 @@ var DefaultConfig = Config{
 		EnablePprof:                true,
 		EnableHTTPDebugAPI:         true,
 		EnableV2Controllers:        false,
-		ShouldPlaceAzureChainFirst: true, // FIXME! default to false after validating
+		ShouldPlaceAzureChainFirst: false,
 	},
 }
 
--- a/npm/config/config.go
+++ b/npm/config/config.go
@@ -18,7 +18,7 @@ var DefaultConfig = Config{
 		EnablePprof:                true,
 		EnableHTTPDebugAPI:         true,
 		EnableV2Controllers:        false,
-		ShouldPlaceAzureChainFirst: false,
+		ShouldPlaceAzureChainFirst: true, // FIXME! default to false after validating
 	},
 }
 
--- a/npm/config/config.go
+++ b/npm/config/config.go
@@ -14,11 +14,11 @@ var DefaultConfig = Config{
 	ListeningPort:         defaultListeningPort,
 	ListeningAddress:      "0.0.0.0",
 	Toggles: Toggles{
-		EnablePrometheusMetrics:    true,
-		EnablePprof:                true,
-		EnableHTTPDebugAPI:         true,
-		EnableV2Controllers:        false,
-		ShouldPlaceAzureChainFirst: false,
+		EnablePrometheusMetrics: true,
+		EnablePprof:             true,
+		EnableHTTPDebugAPI:      true,
+		EnableV2Controllers:     false,
+		PlaceAzureChainFirst:    false,
 	},
 }
 
@@ -30,9 +30,9 @@ type Config struct {
 }
 
 type Toggles struct {
-	EnablePrometheusMetrics    bool
-	EnablePprof                bool
-	EnableHTTPDebugAPI         bool
-	EnableV2Controllers        bool
-	ShouldPlaceAzureChainFirst bool
+	EnablePrometheusMetrics bool
+	EnablePprof             bool
+	EnableHTTPDebugAPI      bool
+	EnableV2Controllers     bool
+	PlaceAzureChainFirst    bool
 }
--- a/npm/config/config.go
+++ b/npm/config/config.go
@@ -18,7 +18,7 @@ var DefaultConfig = Config{
 		EnablePprof:             true,
 		EnableHTTPDebugAPI:      true,
 		EnableV2Controllers:     false,
-		PlaceAzureChainFirst:    false,
+		PlaceAzureChainFirst:    true,
 	},
 }
 
--- a/npm/config/config.go
+++ b/npm/config/config.go
@@ -14,11 +14,11 @@ var DefaultConfig = Config{
 	ListeningPort:         defaultListeningPort,
 	ListeningAddress:      "0.0.0.0",
 	Toggles: Toggles{
-		EnablePrometheusMetrics:          true,
-		EnablePprof:                      true,
-		EnableHTTPDebugAPI:               true,
-		EnableV2Controllers:              false,
-		PlaceAzureChainBeforeKubeForward: false,
+		EnablePrometheusMetrics:    true,
+		EnablePprof:                true,
+		EnableHTTPDebugAPI:         true,
+		EnableV2Controllers:        false,
+		ShouldPlaceAzureChainFirst: false,
 	},
 }
 
@@ -30,9 +30,9 @@ type Config struct {
 }
 
 type Toggles struct {
-	EnablePrometheusMetrics          bool
-	EnablePprof                      bool
-	EnableHTTPDebugAPI               bool
-	EnableV2Controllers              bool
-	PlaceAzureChainBeforeKubeForward bool
+	EnablePrometheusMetrics    bool
+	EnablePprof                bool
+	EnableHTTPDebugAPI         bool
+	EnableV2Controllers        bool
+	ShouldPlaceAzureChainFirst bool
 }
--- a/npm/config/config.go
+++ b/npm/config/config.go
@@ -14,10 +14,11 @@ var DefaultConfig = Config{
 	ListeningPort:         defaultListeningPort,
 	ListeningAddress:      "0.0.0.0",
 	Toggles: Toggles{
-		EnablePrometheusMetrics: true,
-		EnablePprof:             true,
-		EnableHTTPDebugAPI:      true,
-		EnableV2Controllers:     false,
+		EnablePrometheusMetrics:          true,
+		EnablePprof:                      true,
+		EnableHTTPDebugAPI:               true,
+		EnableV2Controllers:              false,
+		PlaceAzureChainBeforeKubeForward: false,
 	},
 }
 
@@ -29,8 +30,9 @@ type Config struct {
 }
 
 type Toggles struct {
-	EnablePrometheusMetrics bool
-	EnablePprof             bool
-	EnableHTTPDebugAPI      bool
-	EnableV2Controllers     bool
+	EnablePrometheusMetrics          bool
+	EnablePprof                      bool
+	EnableHTTPDebugAPI               bool
+	EnableV2Controllers              bool
+	PlaceAzureChainBeforeKubeForward bool
 }
--- a/npm/config/config.go
+++ b/npm/config/config.go
@@ -18,7 +18,7 @@ var DefaultConfig = Config{
 		EnablePprof:             true,
 		EnableHTTPDebugAPI:      true,
 		EnableV2Controllers:     false,
-		PlaceAzureChainFirst:    true,
+		PlaceAzureChainFirst:    false,
 	},
 }
 
--- a/npm/iptm/iptm.go
+++ b/npm/iptm/iptm.go
@@ -204,15 +204,6 @@ func (iptMgr *IptablesManager) checkAndAddForwardChain() error {
 		},
 	}
 
-	// retrieve KUBE-SERVICES index
-	kubeServicesLine, err := iptMgr.getChainLineNumber(util.IptablesKubeServicesChain, util.IptablesForwardChain)
-	if err != nil {
-		metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to get index of KUBE-SERVICES in FORWARD chain with error: %s", err.Error())
-		return err
-	}
-
-	index := kubeServicesLine + 1
-
 	exists, err := iptMgr.exists(entry)
 	if err != nil {
 		return err
@@ -221,6 +212,7 @@ func (iptMgr *IptablesManager) checkAndAddForwardChain() error {
 	if !exists {
 		// position Azure-NPM chain after Kube-Forward and Kube-Service chains if it exists
 		iptMgr.OperationFlag = util.IptablesInsertionFlag
+		index := 1
 		entry.Specs = append([]string{strconv.Itoa(index)}, entry.Specs...)
 		if _, err = iptMgr.run(entry); err != nil {
 			metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to add AZURE-NPM chain to FORWARD chain.")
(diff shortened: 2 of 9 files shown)


npm/iptm/iptm.go
```diff
@@ -221,6 +212,7 @@ func (iptMgr *IptablesManager) checkAndAddForwardChain() error {
 	if !exists {
 		// position Azure-NPM chain after Kube-Forward and Kube-Service chains if it exists
 		iptMgr.OperationFlag = util.IptablesInsertionFlag
+		index := 1
 		entry.Specs = append([]string{strconv.Itoa(index)}, entry.Specs...)
```
rbtr, line 216: is `index` never not `1`? seems like we could save the strconv on the next line. maybe also the inline array initialization

huntergregory: the initial commits here were just to validate something. now we use a toggle so index is variable

npm/config/config.go
```diff
@@ -14,10 +14,11 @@ var DefaultConfig = Config{
...
 	Toggles: Toggles{
-		EnablePrometheusMetrics: true,
-		EnablePprof:             true,
-		EnableHTTPDebugAPI:      true,
-		EnableV2Controllers:     false,
+		EnablePrometheusMetrics:    true,
+		EnablePprof:                true,
+		EnableHTTPDebugAPI:         true,
+		EnableV2Controllers:        false,
+		ShouldPlaceAzureChainFirst: true, // FIXME! default to false after validating
```
vakalapa, line 21: Can we remove "Should" and let the flag be "PlaceAzureChainFirst: true"?

npm/iptm/iptm.go
```diff
@@ -201,18 +221,26 @@ func (iptMgr *IptablesManager) checkAndAddForwardChain() error {
 		Specs: []string{
 			util.IptablesJumpFlag,
 			util.IptablesAzureChain,
+			util.IptablesModuleFlag,
```
vakalapa, line 224: Are we doing ct new condition for when the flag is false too? We will need to test this scenario too. Through conformance and cyclonus

npm/iptm/iptm.go
```diff
@@ -253,8 +285,8 @@ func (iptMgr *IptablesManager) checkAndAddForwardChain() error {
 		return err
 	}
 	iptMgr.OperationFlag = util.IptablesInsertionFlag
-	// Reduce index for deleted AZURE-NPM chain
-	if index > 1 {
+	if !iptMgr.shouldPlaceAzureChainFirst {
+		// Reduce index for deleted AZURE-NPM chain
 		index--
 	}
 	entry.Specs = append([]string{strconv.Itoa(index)}, entry.Specs...)
```
vakalapa, line 292: in entry.specs, we do not have to explicitly say index 1 while inserting, we can just give -I flag and rule will be added to top. I think that is better than having to pass in index, what if the index gets corrupted tomorrow due to a bug ? 
```go
if !iptMgr.shouldPlaceAzureChainFirst {
	// Reduce index for deleted AZURE-NPM chain
	index--
}
entry.Specs = append([]string{strconv.Itoa(index)}, entry.Specs...)
if errCode, err = iptMgr.run(entry); err != nil {
```
Change To 

```go


if !iptMgr.shouldPlaceAzureChainFirst {
	// Reduce index for deleted AZURE-NPM chain
	index--
        entry.Specs = append([]string{strconv.Itoa(index)}, entry.Specs...)
}
if errCode, err = iptMgr.run(entry); err != nil {
```

npm/pkg/controlplane/controllers/v1/controllers_v1_test.go
```diff
@@ -8,13 +8,14 @@ import (
...
 	"github.com/Azure/azure-container-networking/npm/metrics"
+	"github.com/Azure/azure-container-networking/npm/util"
 	"k8s.io/utils/exec"
 )
 
 func TestMain(m *testing.M) {
 	metrics.InitializeAll()
 	realexec := exec.New()
-	iptMgr := iptm.NewIptablesManager(realexec, iptm.NewFakeIptOperationShim())
+	iptMgr := iptm.NewIptablesManager(realexec, iptm.NewFakeIptOperationShim(), util.PlaceAzureChainAfterKubeServices)
```
vakalapa, line 18: instead of `util.PlaceAzureChainAfterKubeServices` use  `npmconfig.DefaultConfig.Toggles.ShouldPlaceAzureChainFirst`

vakalapa (changes requested):

npm/iptm/iptm.go
huntergregory: we are. We'll make this change in v2 as well

npm/iptm/iptm.go
huntergregory: Good point. Will change

vakalapa:
Lgtm, with PlaceAzureFirst flag turned on below are the test results and they are all green:

Cyclonus https://github.com/Azure/azure-container-networking/runs/4184026281?check_suite_focus=true
Conformance: https://msazure.visualstudio.com/One/_build/results?buildId=48875491&view=results

Now let's switch the flag and see the normal is fine too.

vakalapa:
placeAzureChainFirst == true test results:

Cyclonus: https://github.com/Azure/azure-container-networking/runs/4193283318?check_suite_focus=true
Conformance: https://msazure.visualstudio.com/One/_build/results?buildId=48905776&view=results 

lets switch the flag and test with false to make sure there is no regressions

vakalapa (approved):
Lgtm

PR #1087, opened 2021-11-05
Use avast to retry init cns and register node. 

paulgmiller:
**Reason for Change**:
apiserver nnc might not be registered or api server might be down and crashloop backoff puts us outside of 5-10 minutes we have for aks addons to come up so retry a bit more aggresively here.

We could also exponential backoff here but wouldn't want ot go above 1 minute or we'd be in same place.

Use avast retry library in cns main so we don't hand roll our own. 

This is the iron man bandaid version of https://github.com/Azure/azure-container-networking/pull/1084

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr (approved):
❤️ a real retrier

PR #1095, opened 2021-11-10
feat: [NPM] Support pod grace period for v2 

### LINKED ISSUE #1066: NPM blocks graceful shutdown connections for pods in Terminating state

JungukCho:
**Reason for Change**:

PodController in NPM removes ipset information as soon as Pod's status becomes “Terminating”.
So, even though terminated Pod has some activities (e.g., sending traffic to other pods to notify something) during "Terminating" status, NPM does not allow this connection from the Pod.
Thus, NPM needs to delay cleanup process until Pod completes graceful shutdown.

**Issue Fixed**:

Fixes #1066

**Requirements**:

- [x] includes documentation - in `support-pod-grace-period` directory in acndoc
- [x] adds unit tests

**Notes**:
Previous PR for v1 - feat: [NPM] Support graceful shutdown in pod #1083

### DIFF
--- a/npm/pkg/controlplane/controllers/v1/podController.go
+++ b/npm/pkg/controlplane/controllers/v1/podController.go
@@ -21,11 +21,10 @@ import (
 	coreinformer "k8s.io/client-go/informers/core/v1"
 	corelisters "k8s.io/client-go/listers/core/v1"
 
+	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
-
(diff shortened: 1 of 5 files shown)


huntergregory (commented):
Looks good if k8slabels.Equals() has community support

huntergregory (approved):

PR #1104, opened 2021-11-12
test: [NPM] Profiles

huntergregory:
These ConfigMaps will be used to test multiple configurations of NPM in our pipelines.

v2 will default to placing Azure chain first. v1 will default placing it after KUBE-SERVICES.

Includes toggle added in PR #1086


npm/profiles/default-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+  namespace: kube-system
+data:
+  azure-npm.json: |
+    {
+      "ResyncPeriodInMinutes": 15,
+      "ListeningPort": 10091,
+      "ListeningAddress": "0.0.0.0",
+      "Toggles": {
+        "EnablePrometheusMetrics": true,
+        "EnablePprof": true,
```
vakalapa, line 14: Have pprof to be false always in all profiles.

.github/workflows/cyclonus-netpol-test.yaml
```diff
@@ -17,6 +17,10 @@ on:
 jobs:
   cyclonus-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # run cyclonus tests in parallel for NPM with the given ConfigMaps
+        profile: [v1-default.yaml, v2-default.yaml]
```
vakalapa, line 23: Can you remove v2-default for now and have the placefirst profile here, then we can merge this, when required and when we have all green we can enable v2-profile too. Wdyt ? 

I am okay to merge with v2-default and get some results but the master branch will show failures.

vakalapa (approved):

PR #1108, opened 2021-11-16
feat: [NPM] Reset ipsets & update a Prometheus function ResetIPSetEntries (previously just used for UTs)

huntergregory:
Logic for resetting IPSets in Linux:
- get current npm IPSets: 
  - run `ipset list --name` (which returns just the set names per line) and grep for "azure-npm-"
- create a restore file which flushes all these sets and then deletes all of them
- see the code for detailed description of the error handling logic and Prometheus metrics logic

Update for Prometheus `ResetIPSetEntries`: 
- removes all sets from the Prometheus metric for IPSet entries

### DIFF
--- a/npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
+++ b/npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
@@ -92,19 +92,19 @@ var (
 func (iMgr *IPSetManager) resetIPSets() error {
 	listCommand := iMgr.ioShim.Exec.Command(ipsetCommand, ipsetListFlag, ipsetNameFlag)
 	grepCommand := iMgr.ioShim.Exec.Command(ioutil.Grep, azureNPMPrefix)
-	searchResults, gotMatches, grepError := ioutil.PipeCommandToGrep(listCommand, grepCommand)
-	if grepError != nil {
-		return npmerrors.SimpleErrorWrapper("failed to run ipset list for resetting IPSets", grepError)
+	azureIPSets, haveAzureIPSets, commandError := ioutil.PipeCommandToGrep(listCommand, grepCommand)
+	if commandError != nil {
+		return npmerrors.SimpleErrorWrapper("failed to run ipset list for resetting IPSets", commandError)
 	}
-	if !gotMatches {
+	if !haveAzureIPSets {
 		metrics.ResetNumIPSets()
 		metrics.ResetIPSetEntries()
 		return nil
 	}
-	creator, originalNumSets, destroyFailureCount := iMgr.fileCreatorForReset(searchResults)
+	creator, originalNumAzureSets, destroyFailureCount := iMgr.fileCreatorForReset(azureIPSets)
 	restoreError := creator.RunCommandWithFile(ipsetCommand, ipsetRestoreFlag)
 	if restoreError != nil {
-		metrics.SetNumIPSets(originalNumSets)
+		metrics.SetNumIPSets(originalNumAzureSets)
 		// NOTE: the num entries for sets may be incorrect if the restore fails
 		return npmerrors.SimpleErrorWrapper("failed to run ipset restore for resetting IPSets", restoreError)
 	}
@@ -118,7 +118,7 @@ func (iMgr *IPSetManager) resetIPSets() error {
 }
 
 // this needs to be a separate function because we need to check creator contents in UTs
-func (iMgr *IPSetManager) fileCreatorForReset(ipsetListOutput []byte) (creator *ioutil.FileCreator, numSets int, destroyFailureCount *int) {
+func (iMgr *IPSetManager) fileCreatorForReset(ipsetListOutput []byte) (creator *ioutil.FileCreator, originalNumAzureSets int, destroyFailureCount *int) {
 	zero := 0
 	destroyFailureCount = &zero
 	creator = ioutil.NewFileCreator(iMgr.ioShim, maxTryCount, ipsetRestoreLineFailurePattern)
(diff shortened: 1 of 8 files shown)


npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -57,14 +60,135 @@ var (
...
+	names := make([]string, 0)
+	readIndex := 0
+	var line []byte
+	// flush all the sets and create a list of the sets so we can destroy them
+	for readIndex < len(ipsetListOutput) {
+		line, readIndex = parse.Line(readIndex, ipsetListOutput)
+		hashedSetName := string(line)
+		names = append(names, hashedSetName)
+		// error handlers specific to resetting ipsets
+		errorHandlers := []*ioutil.LineErrorHandler{
```
JungukCho, line 134: Great work to classify all errors and corresponding actions.
I understood the high-level ideas about `LineErrorHandler` which holds necessary error handling information to the corresponding executed line when there is error.

While I am fine as is, but just curious.
Do we consolidate all error handlings code without this `errorHandlers` here - https://github.com/Azure/azure-container-networking/blob/master/npm/pkg/dataplane/ioutil/restore_linux.go#L199?
For maintenance, it would be nice to consolidate error handling in one places if possible.
I thought `Callback` has only error message, but it changed `destroyFailureCount` in some places.

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -57,14 +60,135 @@ var (
...
+		If a flush fails, we could update the num entries for that set, but that would be a lot of overhead.
+*/
 func (iMgr *IPSetManager) resetIPSets() error {
-	// called on failure or when NPM is created
-	// so no ipset cache. need to use ipset list like in ipsm.go
+	listCommand := iMgr.ioShim.Exec.Command(ipsetCommand, ipsetListFlag, ipsetNameFlag)
+	grepCommand := iMgr.ioShim.Exec.Command(ioutil.Grep, azureNPMPrefix)
+	searchResults, gotMatches, grepError := ioutil.PipeCommandToGrep(listCommand, grepCommand)
+	if grepError != nil {
+		return npmerrors.SimpleErrorWrapper("failed to run ipset list for resetting IPSets", grepError)
```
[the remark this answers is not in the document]
huntergregory: good point. I will rename

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -57,14 +60,135 @@ var (
...
+	}
+	if !gotMatches {
+		metrics.ResetNumIPSets()
+		metrics.ResetIPSetEntries()
+		return nil
+	}
+	creator, originalNumSets, destroyFailureCount := iMgr.fileCreatorForReset(searchResults)
```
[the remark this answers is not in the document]
huntergregory: perhaps we can rename all the functions and variables with (file)creator in them, but this should be a separate cleanup PR

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
JungukCho: Yep. I understood the intention of codes. 

My fundamental comments are whether we can consolidate all error definitions and handlers in one place per `ipset` and `iptables` instead of sparsely declaring them in each function and using `Callback`. 
It will be easier to understand and maintain codes, but I may miss somethings and it will need big changes.
wdyt?

BTW, it is fine as is.

huntergregory: Still might not fully understand your question. To try to help consolidate, for `creatorForApply`, I group all of the error handling code at the bottom of the file. I can't move the error handling code out of `creatorForReset` because of the dependency on `destroyFailureCount`.

We could consolidate the definition and method (e.g. `ContinueAndAbortSection`) but the callback will always be unique and will eventually do more than logging for most (e.g. mark a set as needing reconcile).

huntergregory: for instance though, we could make this error handler a function of the set name and destroyFailureCount pointer though

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -57,14 +60,135 @@ var (
...
+		metrics.ResetNumIPSets()
+	}
+	metrics.ResetIPSetEntries() // NOTE: the num entries for sets that fail to flush may be incorrect after this
+	return nil
+}
 
-	// create restore file that flushes all sets, then deletes all sets
-	// technically don't need to flush a hashset
+// this needs to be a separate function because we need to check creator contents in UTs
+func (iMgr *IPSetManager) fileCreatorForReset(ipsetListOutput []byte) (creator *ioutil.FileCreator, originalNumAzureSets int, destroyFailureCount *int) {
```
matmerr, line 121: as discussed offline avoid usage of named returns, and potentially just return creator, originalNumAzureSets and destroyFailureCounts are purely used for for metrics tracking and don't particularly related to the usage of creator in clients

huntergregory: should we fail the lint check in favor of unnamed returns? Also, I'm not sure how we could decouple the numbers for metrics from the function output because we set metrics based on the success/failure of `creator.RunCommand`

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -118,10 +118,9 @@ func (iMgr *IPSetManager) resetIPSets() error {
 }
 
 // this needs to be a separate function because we need to check creator contents in UTs
-func (iMgr *IPSetManager) fileCreatorForReset(ipsetListOutput []byte) (creator *ioutil.FileCreator, originalNumAzureSets int, destroyFailureCount *int) {
-	zero := 0
-	destroyFailureCount = &zero
-	creator = ioutil.NewFileCreator(iMgr.ioShim, maxTryCount, ipsetRestoreLineFailurePattern)
+func (iMgr *IPSetManager) fileCreatorForReset(ipsetListOutput []byte) (*ioutil.FileCreator, int, *int) {
```
JungukCho, line 121: It seems we do not return pointer.
```go
(*ioutil.FileCreator, int, *int) -> (*ioutil.FileCreator, int, int)
```
and can use non-pointer operation in caller as well.

JungukCho (approved):
lgtm! You can resolve my last comment in next PR if it makes sense.
However, please check @matmerr comments.

PR #1110, opened 2021-11-17
fix ipampool scaling - only release IPs we have

rbtr:

**Reason for Change**:

When scaling up, we should account for the IPs we have requested when we calculate our free. But when scaling down, we should only account for the IPs that we _have_ when calculating free.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/ipampool/monitor.go
+++ b/cns/ipampool/monitor.go
@@ -103,29 +103,29 @@ func (pm *Monitor) reconcile(ctx context.Context) error {
 	allocatedPodIPCount := len(pm.httpService.GetAllocatedIPConfigs())
 	pendingReleaseIPCount := len(pm.httpService.GetPendingReleaseIPConfigs())
 	availableIPConfigCount := len(pm.httpService.GetAvailableIPConfigs()) // TODO: add pending allocation count to real cns
-	requestedIPConfigCount := pm.spec.RequestedIPCount
-	unallocatedIPConfigCount := cnsPodIPConfigCount - allocatedPodIPCount
-	freeIPConfigCount := requestedIPConfigCount - int64(allocatedPodIPCount)
+	requestedIPConfigCount := int(pm.spec.RequestedIPCount)
+	unallocated := cnsPodIPConfigCount - allocatedPodIPCount
+	requestedUnallocated := requestedIPConfigCount - allocatedPodIPCount
 	batchSize := pm.scaler.BatchSize
 	maxIPCount := pm.scaler.MaxIPCount
 
-	msg := fmt.Sprintf("[ipam-pool-monitor] Pool Size: %v, Goal Size: %v, BatchSize: %v, MaxIPCount: %v, Allocated: %v, Available: %v, Pending Release: %v, Free: %v, Pending Program: %v",
-		cnsPodIPConfigCount, pm.spec.RequestedIPCount, batchSize, maxIPCount, allocatedPodIPCount, availableIPConfigCount, pendingReleaseIPCount, freeIPConfigCount, pendingProgramCount)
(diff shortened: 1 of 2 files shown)


cns/ipampool/monitor.go
```diff
@@ -103,19 +103,20 @@ func (pm *Monitor) reconcile(ctx context.Context) error {
 	allocatedPodIPCount := len(pm.httpService.GetAllocatedIPConfigs())
 	pendingReleaseIPCount := len(pm.httpService.GetPendingReleaseIPConfigs())
 	availableIPConfigCount := len(pm.httpService.GetAvailableIPConfigs()) // TODO: add pending allocation count to real cns
-	requestedIPConfigCount := pm.spec.RequestedIPCount
+	requestedIPConfigCount := int(pm.spec.RequestedIPCount)
 	unallocatedIPConfigCount := cnsPodIPConfigCount - allocatedPodIPCount
-	freeIPConfigCount := requestedIPConfigCount - int64(allocatedPodIPCount)
+	currentFree := cnsPodIPConfigCount - allocatedPodIPCount
+	futureFree := requestedIPConfigCount - allocatedPodIPCount
```
neaggarwMS, line 109: Can we rename it something else, futureFree is little confusing specially when it is being used for ScaleUP. You will scaleup if you dont have ips left. Cant think of a better name too :)

May be "ipsToBeAssignedToCNS"...I'm sure you will come with a better name

neaggarwMS (changes requested):

cns/ipampool/monitor.go
rbtr: I actually can't think of a better name...can we move forward with this one and I will follow up with a change to clean up the way this state is gathered and passed around?

neaggarwMS: How About "totalIpsToBeAssignedToCNS"

neaggarwMS (changes requested):

cns/ipampool/monitor.go
rbtr: I think we're using assigned and allocated backwards. It should be: 
"IPs are allocated to CNS"
"CNS assigns IPs to Pods" 

Then, the total number of IPs that CNS _has_ are the its "allocated" IPs, and the total number if IPs _in use_ are how many that CNS has "assigned" to specific Pods.
"currentFree" becomes `unassigned := allocated - assigned`
"futureFree" becomes something like `requestedUnassigned := requested - assigned`

cns/ipampool/monitor.go
```diff
@@ -103,19 +103,20 @@ func (pm *Monitor) reconcile(ctx context.Context) error {
 	allocatedPodIPCount := len(pm.httpService.GetAllocatedIPConfigs())
 	pendingReleaseIPCount := len(pm.httpService.GetPendingReleaseIPConfigs())
 	availableIPConfigCount := len(pm.httpService.GetAvailableIPConfigs()) // TODO: add pending allocation count to real cns
-	requestedIPConfigCount := pm.spec.RequestedIPCount
+	requestedIPConfigCount := int(pm.spec.RequestedIPCount)
 	unallocatedIPConfigCount := cnsPodIPConfigCount - allocatedPodIPCount
```
neaggarwMS, line 107: We can avoid this, it is same as CurrentFree

cns/ipampool/monitor.go
neaggarwMS: I like the above analogy.
However, since we have defined allocated => CNS assigned ips to pod

lets rename currentFree = unAllocated
FutureFree = requestedUnAllocated

IMO, FutureFree is really hard to explain to a new pair of eyes.

neaggarwMS (changes requested):

neaggarwMS (approved):

PR #1113, opened 2021-11-17
refactor: [NPM] Minor fixes (does not change functionality, but for better maintenance)

JungukCho:
**Reason for Change**:

Few changes (use better functions, update comments and remove todo for better maintenance.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/pkg/dataplane/policies/policy.go
+++ b/npm/pkg/dataplane/policies/policy.go
@@ -46,8 +46,9 @@ type ACLPolicy struct {
 	Target Verdict
 	// Direction defines the flow of traffic
 	Direction Direction
-	// DstPorts holds the destination port information
-	// TODO(jungukcho): It may be better to use pointer to differentiate default value.
+	// DstPorts always holds the destination port information.
+	// The valid value for port must be between 1 and 65535, inclusive
+	// and the endPort must be equal or greater than port.
 	DstPorts Ports
 	// Protocol is the value of traffic protocol
 	Protocol Protocol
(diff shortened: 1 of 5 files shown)


npm/pkg/dataplane/policies/policy.go
```diff
@@ -110,7 +111,6 @@ func (aclPolicy *ACLPolicy) hasKnownProtocol() bool {
 	return aclPolicy.Protocol != "" && (aclPolicy.Protocol == TCP ||
 		aclPolicy.Protocol == UDP ||
 		aclPolicy.Protocol == SCTP ||
-		aclPolicy.Protocol == ICMP ||
```
vakalapa, line 113: Is there a reason for us to not support ICMP ? Network policy rules can include ICMP as protocol right ?

JungukCho: Thank you for bring it up. 
Confirmed it with [networkpolicy spec](https://github.com/kubernetes/api/blob/cf43cc39e22390e13043e176fdb787cfccdb1ef8/networking/v1/types.go#L140) and AKS cluster.

```
// The protocol (TCP, UDP, or SCTP) which traffic must match. If not specified, this
// field defaults to TCP.
```
```yaml
kind: NetworkPolicy
metadata:
  name: icmp-test
  namespace: default
spec:
  podSelector:
    matchLabels:
      role: db
  policyTypes:
  - Ingress
  ingress:
  - from:
    - ipBlock:
        cidr: 10.0.0.0/24
    ports:
    - protocol: ICMP
```
```shell
kubectl apply -f icmp.yaml
The NetworkPolicy "icmp-test" is invalid: spec.ingress[0].ports[0].protocol: Unsupported value: "ICMP": supported values: "TCP", "UDP", "SCTP"
```

vakalapa (approved):

PR #1120, opened 2021-11-22
fix: [NPM] V2 NPM (with only linux data plane) bug fix after first run

huntergregory:
**Reason for Change**:
After running conformance test, fix finding issues.

1. Had a bug where we were jumping to AZURE-NPM-EGRESS chain instead. We need to jump to this other chain since the ingress allow mark will prevent us from returning to ingress chain again after a no-op pass through egress chain.

2. Observed multiple restarting NPM. Checked log and there is nil pointer errors in `NPMSimpleError`.
- For the second error, we need to UT to catch this `nil` error.  
- Need to resolve lint warning.

3. Fix case-sensitivity on protocol value
- need to remove dependency later 
- need better management with `unspecified` and `Any`

4. Fix problem of trying to destroy sets in use by a kernel component
- previously when running `ipset restore`, we would try to destroy sets before deleting any references to them in lists
- fix by destroying sets last in the `ipset restore` file

5. Fix duplicated `namedPort`  prefix. 

6. Fix translating `egress (previously included only the default drop rule)`
- Add UTs for `egress`

7. Fix string parsing issue in except information of ipblock
- Need more UTs (in dataplane_test.go)
- Better data structures to indicate `nomatch` information in ipblock
- Need to resolve lint warning.

8. Fix using correct, unique and consistent key in policy cache. (now deleting network policy works)
- Need more UTs
- Do not walk through `windows` part. Please revise `windows` if needed.

9. Fix Linux policy manager to use PodSelectorList instead of PodSelectorIPSets (this caused an impossible ipset match condition for nested pod labels e.g. pod-a-b AND pod:a AND pod:b).
- Will update the NPMNetworkPolicy struct to just have one of these fields eventually instead of the two.

- [ ] includes documentation
- [ ] adds unit tests

**Notes:**

### DIFF
--- a/.github/workflows/cyclonus-netpol-test.yaml
+++ b/.github/workflows/cyclonus-netpol-test.yaml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         # run cyclonus tests in parallel for NPM with the given ConfigMaps
-        profile: [v1-default.yaml, v1-place-azure-chain-first.yaml]
+        profile: [v1-default.yaml, v1-place-azure-chain-first.yaml, v2-default.yaml]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/npm/azure-npm.yaml
+++ b/npm/azure-npm.yaml
@@ -153,7 +153,7 @@ data:
             "EnablePrometheusMetrics": true,
             "EnablePprof":             true,
             "EnableHTTPDebugAPI":      true,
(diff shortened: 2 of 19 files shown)


vakalapa (dismissed):

JungukCho (approved):

vakalapa (approved):
LGTM, great effort finding all the issues!!!!

PR #1123, opened 2021-11-22
fix: [NPM] V2 NPM with linux DP bug fix after first run

JungukCho:
**Reason for Change**:

After running conformance test, fix finding issues.

1. Observed multiple restarting NPM. Checked log and there is `nil` pointer errors in NPMSimpleError.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
TODO list.
1. For the first error, we need to UT to catch this `nil` error.
2. I will resolve `lint warning` in next PR.

### DIFF
--- a/npm/util/errors/errors.go
+++ b/npm/util/errors/errors.go
@@ -192,7 +192,7 @@ type NPMSimpleError struct {
 }
 
 func SimpleError(errstring string) *NPMSimpleError {
-	return nil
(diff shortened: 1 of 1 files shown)


JungukCho:
close this PR since we made decision to deal with this in https://github.com/Azure/azure-container-networking/pull/1120

PR #1124, opened 2021-11-23
perf: [NPM] Remove random factor for resync and set resync time 0 as default value

JungukCho:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

Random factor for resync is unnecessary since it re-lists cached objects from a local cache in client-go.  In addition, `resync` (15 mins) is not helpful for NPM. 
More detail refers to doc in acn

- [x] includes documentation in **acndocs**
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/azure-npm.yaml
+++ b/npm/azure-npm.yaml
@@ -146,7 +146,7 @@ metadata:
 data:
   azure-npm.json: |
     {
-        "ResyncPeriodInMinutes": 15,
+        "ResyncPeriodInMinutes": 0,
         "ListeningPort":         10091,
         "ListeningAddress":      "0.0.0.0",
         "Toggles": {
--- a/npm/cmd/start.go
+++ b/npm/cmd/start.go
@@ -6,7 +6,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"math/rand"
 	"time"
 
 	"github.com/Azure/azure-container-networking/common"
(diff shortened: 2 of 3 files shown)


npm/azure-npm.yaml
```diff
@@ -146,7 +146,7 @@ metadata:
 data:
   azure-npm.json: |
     {
-        "ResyncPeriodInMinutes": 15,
```
vakalapa, line 149: It is recommended to have the resync, unless client go and our controller is completely bug free. Resync is kind of a insurance for us, it does good for us than any harm (except for some CPU cycles?)

npm/cmd/start.go
```diff
@@ -99,10 +98,7 @@ func start(config npmconfig.Config) error {
 
 	// Setting reSyncPeriod
 	minResyncPeriod := time.Duration(config.ResyncPeriodInMinutes) * time.Minute
-
-	// Adding some randomness so all NPM pods will not request for info at once.
-	factor := rand.Float64() + 1
-	resyncPeriod := time.Duration(float64(minResyncPeriod.Nanoseconds()) * factor)
+	resyncPeriod := time.Duration(float64(minResyncPeriod.Nanoseconds()))
```
vakalapa, line 101: Agreed, random factor can be removed.
https://github.com/kubernetes/kubernetes/pull/71893#discussion_r240956177 

Resync will not relist.

PR #1133, opened 2021-11-29
feat: nnc client sets ownerref

thatmattlong:
**Reason for Change**:
Adds SetOwnerRef to nnc client which sets the NNC ownership to the given object

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/crd/nodenetworkconfig/client.go
+++ b/crd/nodenetworkconfig/client.go
@@ -16,6 +16,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	ctrlcli "sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // Scheme is a runtime scheme containing the client-go scheme and the NodeNetworkConfig scheme.
@@ -24,6 +25,12 @@ var Scheme = runtime.NewScheme()
 func init() {
 	_ = clientgoscheme.AddToScheme(Scheme)
 	_ = v1alpha.AddToScheme(Scheme)
+}
+
+type OwnerExistsError struct{}
+
+func (*OwnerExistsError) Error() string {
+	return "OwnerRef already exists and does not match"
 }
 
 // Client is provided to interface with the NodeNetworkConfig CRDs.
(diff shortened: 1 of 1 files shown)


crd/nodenetworkconfig/client.go
```diff
@@ -26,6 +28,12 @@ func init() {
 	_ = v1alpha.AddToScheme(Scheme)
 }
 
+type OwnerExistsError struct{}
```
rbtr, line 31: can we flip this around to ErrOwnerExists? seems more canonical (https://go.dev/src/os/error.go) even though k/k likes to do "ThingError"

crd/nodenetworkconfig/client.go
```diff
@@ -136,3 +144,20 @@ func (c *Client) UpdateSpec(ctx context.Context, key types.NamespacedName, spec
 	}
 	return nnc, nil
 }
+
+// SetOwnerRef sets the owner of the NodeNetworkConfig to the given object, using HTTP Patch
+func (c *Client) SetOwnerRef(ctx context.Context, nnc *v1alpha.NodeNetworkConfig, owner metav1.Object) error {
+	newNNC := nnc.DeepCopy()
```
rbtr, line 150: I think we should either return the `newNNC` or mutate the passed `nnc`, since the patch call will write the resultant object to the ref it gets passed.

crd/nodenetworkconfig/client.go
```diff
@@ -136,3 +144,20 @@ func (c *Client) UpdateSpec(ctx context.Context, key types.NamespacedName, spec
...
+	if err := ctrlutil.SetControllerReference(owner, newNNC, Scheme); err != nil {
+		// SetControllerReference returns an error if the object already has a different owner or
+		// if there are issues with the scheme. Since we control the scheme in this package, we can safely
+		// assume than any error is caused by an owner already existing
+		return &OwnerExistsError{}
```
rbtr, line 155: why not lean on ctrlutil's [`AlreadyOwnedError`](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L38)?

rbtr (changes requested):

crd/nodenetworkconfig/client.go
rbtr: _if_ we still need it instead of using ctrlutil's similar defined error

crd/nodenetworkconfig/client.go
thatmattlong: because I didn't read the docs and didn't know it existed 😆

crd/nodenetworkconfig/client.go
thatmattlong: good point, I'll change that

thatmattlong: decided to return instead of mutate since the `nnccli.Patch()` could fail and we don't want to mutate the caller if we dn't actually make a successful change

crd/nodenetworkconfig/client.go
thatmattlong: removed, caller can rely on ctrlutil's error with `errors.Is/As` since we just wrap it

rbtr (approved):
lgtm

PR #1136, opened 2021-11-30
proposal: separate mods for cni, cns, and npm

rbtr:
**Reason for Change**:

Makes the primary component projects separate go modules so that they can be versioned and released independently.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


ramiro-gamarra:
Trying to understand the proposal here. This will allow each module to manage its dependencies independently, but it's not clear to me how they can be versioned independently. In regular cases, the module version is still based on a git tag, so even if we have multiple modules, we still only have the one repo tag (unless using different sets of tags per module is suggested?). Let me know what I'm missing :).

rbtr:
Modules in a monorepo can be versioned independently by prefixing the tag with their subpath: `cns/v1.4.16`, `npm/v2.0.0`, etc.
Also, once an initial version for each module is tagged, the `replace` directives (like for the CNS CNI dependency) can be replaced by `require`ing the real tag.

cni/go.mod
```diff
@@ -0,0 +1,83 @@
...
+	github.com/Azure/azure-container-networking => ../
+	github.com/Azure/azure-container-networking/cns => ../cns
+	github.com/Microsoft/go-winio => github.com/microsoft/go-winio v0.4.17
+	github.com/Microsoft/hcsshim => github.com/microsoft/hcsshim v0.8.22
+	github.com/onsi/ginkgo => github.com/onsi/ginkgo v1.12.0
+	github.com/onsi/gomega => github.com/onsi/gomega v1.10.0
+)
+
+require (
+	github.com/Azure/azure-container-networking v0.0.0-00010101000000-000000000000
```
tamilmani1989, line 15: what this v0.0.0-00010101000000-000000000000 signifies?

cns/go.mod
```diff
@@ -0,0 +1,89 @@
+module github.com/Azure/azure-container-networking/cns
+
+go 1.17
+
+replace (
+	github.com/Azure/azure-container-networking => ../
+	github.com/Azure/azure-container-networking/cni => ../cni
```
tamilmani1989, line 7: same as above. go.mod under cni directory?

cni/go.mod
```diff
@@ -0,0 +1,83 @@
+module github.com/Azure/azure-container-networking/cni
+
+go 1.17
+
+replace (
+	github.com/Azure/azure-container-networking => ../
+	github.com/Azure/azure-container-networking/cns => ../cns
```
tamilmani1989, line 7: go standard says (https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive)
`Note: if the right-hand side of a replace directive is a filesystem path, then the target must have a go.mod file at that location`

do we need to create go.mod under cns directory?

cni/go.mod
rbtr: go mod uses the 0.0.0 version when we do `replace` to a filesystem path, since it is not fetching a published module version

cni/go.mod
rbtr: this `replace` is temporary - once versions of the cni and cni module are tagged, we can remove the `replace` and instead use `require` for the specific verson of the module that we have tagged.
for example, if we tag `cns/v1.4.16` and `cni/v1.4.16`, then cni and cns can depend on those published module versions, respectively.

thatmattlong:
I'm wondering if it might be easier to keep CNS and CNI versioned together since they share quite a bit of dependent interaction, and it might get difficult to know which CNI versions and which CNS versions are safe to use together. I suppose semantic versioning would help with this, but there are even cases where CNS isn't used (static Azure CNI) and I'm a little bit afraid of potential complexity that we'd introduce by versioning these separately. NPM of course can (and should) live on its own. But CNS and CNI are tightly coupled and might be best to keep them together.

rbtr:
CNS and CNI still have feature changes and bugfixes that can (and are, at least in AKS) shipped independently. I would adhere to semver and decouple the component releases.

rsagasthya (approved):

PR #1140, opened 2021-12-01
fix: [CNS] fixing a regex to extract hostip in getncversion req

vakalapa:
**Reason for Change**:

Due to wrong regex the hostip is extracted out wrong resulting in 400 badrequest from NMAgent

https://regex101.com/r/pkrHQv/1 

https://github.com/Azure/azure-container-networking/blob/4a4370b0be7c4c7b0293275cd30a470b54a125bf/cns/restserver/api.go#L1085

below is a log snippet, the host ip should be just the IP and not "168.63.129.16/machine/plugins/?comp=nmagent&type=NetworkManagement" regex is not parsing.

[Azure-CNS] Storing ncGetVersionURL for NC ID 53347415-65c7-47af-b13d-31ed8acd1272:  HOST IP 168.63.129.16/machine/plugins/?comp=nmagent&type=NetworkManagement , join net url http://168.63.129.16/machine/plugins/?comp=nmagent&type=NetworkManagement/joinedVirtualNetworks/4941a21f-1a8d-4d0f-8256-cc6e73a8cd22/api-version/1, urlfmt, http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/%s/networkContainers/%s/version/authenticationToken/%s/api-version/1 , ncGetVersionURL: http://168.63.129.16/machine/plugins/?comp=nmagent&type=NetworkManagement/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/10.92.0.29/networkContainers/53347415-65c7-47af-b13d-31ed8acd1272/version/authenticationToken/7598a099-1292-480b-a1cd-3d6df5038578/api-version/1

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -9,7 +9,7 @@ import (
 	"io"
 	"net"
 	"net/http"
-	"regexp"
+	"net/url"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


rbtr (changes requested):
```go
import "net/url"
...
u, err := url.Parse(urlStr)
if err != nil {
    return err
}
host := u.Host
```
the net package has solved parsing URLs to extract hosts, paths, queries, etc (and recomposing safe URLs from those pieces). let's lean on them 🙂

rbtr (approved):

ramiro-gamarra (approved):
"The plural of regex is regrets"

ashvindeodhar (approved):

PR #1150, opened 2021-12-02
fix: [NPM] DP bootup and Linux DP refinements

huntergregory:
Main changes in this PR:
- allow configuration of the dataplane
- changes the dataplane to preform reset/initialization all at once
- in Linux, deactivates NPM when there are no network policies. In deactivated mode, there are no iptables rules in AZURE-NPM chain
- preserve counters in iptables chains (flush chains in iptables-restore files instead of recreating them)

Side Effects of this PR:
- added feature: prevent the ability to add a core NPM chain (e.g. AZURE-NPM-INGRESS) to the stale chains for later deletion
- fixed integration tests to use PolicyKey (after the change in the bug fix PR)
- prevent a false error from logging every time bootup is called in policy manager (see `allCurrentAzureChains` in chain-management_linux.go)


npm/cmd/start.go
```diff
@@ -110,7 +111,7 @@ func start(config npmconfig.Config) error {
 
 	var dp dataplane.GenericDataplane
 	if config.Toggles.EnableV2NPM {
-		dp, err = dataplane.NewDataPlane(npm.GetNodeName(), common.NewIOShim())
+		dp, err = dataplane.NewDataPlane(npm.GetNodeName(), common.NewIOShim(), policies.IPSetAndNoRebootConfig)
```
vakalapa, line 114: We might have a lot more configs coming our way our dp, policy and ipsets mgrs, so lets create a new Dataplane config ? which inherits both policyMgrCfg and ipsetMgrCfg, that way we will have to send in only one cfg option ?

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -56,8 +67,13 @@ func newStaleChains() *staleChains {
 	return &staleChains{make(map[string]struct{})}
 }
 
+// add adds the chain if it isn't one of the iptablesAzureChains.
```
vakalapa, line 70: you can remove the redundant "adds"

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -358,12 +403,25 @@ func (pMgr *PolicyManager) allCurrentAzureChains() ([]string, error) {
...
 	lines := strings.Split(string(searchResults), "\n")
+	if len(lines) == 1 && lines[0] == "" {
+		// this should never happen: gotMatches is true, but there is no content in the searchResults
+		return nil, errInvalidGrepResult
+	}
+	lastIndex := len(lines) - 1
+	lastLine := lines[lastIndex]
+	if lastLine == "" {
+		// remove the last empty line (since each line ends with a newline)
+		lines = lines[:lastIndex]
```
vakalapa, line 414: instead of resizing the while array, could we just check if line != "" in the for loop below ?

huntergregory: looked into this while implementing, and golang slices reference arrays under the hood. This operation will just update the slice length (just an integer <= array length), and not impact memory allocations for the array.

Let me add a short comment about this

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -30,13 +30,22 @@ const (
...
 var (
+	// NOTE any update to this variable should be reflected in the map below, and vice versa
 	iptablesAzureChains = []string{
 		util.IptablesAzureChain,
 		util.IptablesAzureIngressChain,
 		util.IptablesAzureIngressAllowMarkChain,
 		util.IptablesAzureEgressChain,
 		util.IptablesAzureAcceptChain,
 	}
+	iptablesAzureChainsMap = map[string]struct{}{
```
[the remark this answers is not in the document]
huntergregory: I'll update to the former idea. The latter is difficult because of the ordering of fexec commands for UTs.

npm/cmd/start.go
```diff
@@ -29,6 +31,12 @@ import (
 	"k8s.io/utils/exec"
 )
 
+var npmV2DataplaneCfg = &dataplane.Config{
+	PolicyMode:       "",
```
vakalapa, line 35: PolicyMode is what we use in PolicyManagerCfg i think ? we can safly remove this PolicyMode from DpConfig and use it from PolicyManagerCfg ?

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -37,6 +37,18 @@ type IPSetManagerCfg struct {
 	NetworkName string
 }
 
+var (
+	ApplyOnNeedCfg = &IPSetManagerCfg{
```
vakalapa, line 41: Do we need to standardize this ? We know that IPSetMode is a enum, we do not have to define enum of the Cfg obj.

npm/pkg/dataplane/policies/policymanager_linux.go
```diff
@@ -152,17 +167,25 @@ func egressJumpSpecs(networkPolicy *NPMNetworkPolicy) []string {
...
 
-// noflush add to chains impacted
-// TODO use array instead of ...
-func (pMgr *PolicyManager) creatorForNewNetworkPolicies(allChainNames []string, networkPolicies ...*NPMNetworkPolicy) *ioutil.FileCreator {
-	creator := pMgr.newCreatorWithChains(allChainNames)
+func (pMgr *PolicyManager) creatorForNewNetworkPolicies(policyChains []string, networkPolicies []*NPMNetworkPolicy) *ioutil.FileCreator {
+	creator := pMgr.newCreatorWithChains(policyChains)
+
+	// 1. Activate NPM if necessary
+	if pMgr.shouldActivate() {
```
vakalapa, line 174: This might be little confusing, can we change the name to isFirstPolicy, that way we will know that for first policy alone we need these rules?

Also how will we manage duplicates of these rules on very first bootup ?

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -253,13 +229,26 @@ func (pMgr *PolicyManager) runIPTablesCommand(operationFlag string, args ...stri
...
+			chainsToCreate = append(chainsToCreate, chain)
+		}
+	}
 
-	// add AZURE-NPM chain rules
-	creator.AddLine("", nil, util.IptablesAppendFlag, util.IptablesAzureChain, util.IptablesJumpFlag, util.IptablesAzureIngressChain)
-	creator.AddLine("", nil, util.IptablesAppendFlag, util.IptablesAzureChain, util.IptablesJumpFlag, util.IptablesAzureEgressChain)
-	creator.AddLine("", nil, util.IptablesAppendFlag, util.IptablesAzureChain, util.IptablesJumpFlag, util.IptablesAzureAcceptChain)
+	// Step 2.1 in bootup() comment: cleanup old NPM chains, and configure base chains and their rules
+	// To leave NPM deactivated, don't specify any rules for AZURE-NPM chain.
```
vakalapa, line 245: I see we do not duplicate rules here, only activate it when the first policy is applied.

npm/cmd/start.go
vakalapa: Also we can just name it npmDataplaneCfg..

huntergregory: Eventually, I think we should rename Mode to PolicyMode since Dataplane.Config inherits from PolicyManagerCfg, but we'll leave that for another PR

npm/pkg/dataplane/ipsets/ipsetmanager.go
huntergregory: I made variables for IPSetManagerCfg and PolicyManagerCfg unexported and only kept them for UTs

npm/pkg/dataplane/policies/policymanager_linux.go
huntergregory: updated to isFirstPolicy() and isLastPolicy()

vakalapa (approved):

nitishm (approved):
LGTM

PR #1154, opened 2021-12-03
fix: [NPM] v2 backwards compatibility

huntergregory:
Delete the old jump to azure chain (without ctstate new) so that we can destroy azure chain on reset. 

Prior to this fix, deploying v2 after v1 caused a CrashLoopBackOff.

### DIFF
--- a/npm/pkg/dataplane/policies/chain-management_linux.go
+++ b/npm/pkg/dataplane/policies/chain-management_linux.go
@@ -46,6 +46,11 @@ var (
 		util.IptablesCtstateFlag,
 		util.IptablesNewState,
 	}
+	deprecatedJumpFromForwardToAzureChainArgs = []string{
+		util.IptablesForwardChain,
+		util.IptablesJumpFlag,
+		util.IptablesAzureChain,
+	}
 )
 
 type staleChains struct {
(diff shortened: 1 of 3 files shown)


vakalapa (dismissed):

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -130,16 +135,31 @@ func (pMgr *PolicyManager) initializeNPMChains() error {
 // removeNPMChains removes the jump rule from FORWARD chain to AZURE-NPM chain
 // and flushes and deletes all NPM Chains.
 func (pMgr *PolicyManager) removeNPMChains() error {
+	// 1.1 delete the deprecated jump rule from FORWARD chain to AZURE-NPM chain, if it exists
+	deprecatedDeleteErrCode, deprecatedDeleteErr := pMgr.runIPTablesCommand(util.IptablesDeletionFlag, deprecatedJumpFromForwardToAzureChainArgs...)
```
vakalapa, line 139: Does this mean we will suppress any other kind of error without log ? we should at least log it!

huntergregory: Oh, I meant to do the negated version of that boolean. Updating to do a switch statement for all possible error codes

vakalapa (approved):

PR #1155, opened 2021-12-03
fix: [NPM] wait for Azure network to be created in windows DP for 20 …

vakalapa:
…mins

**Reason for Change**:

Currently when an AKS-e cluster is created, "azure" HNS network is not created right away. Only after pods are scheduled will "azure" network be created. Making change in NPM to not error out right away and wait 20 mins and retry for azure network info.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/pkg/dataplane/dataplane_windows.go
+++ b/npm/pkg/dataplane/dataplane_windows.go
@@ -2,6 +2,8 @@ package dataplane
 
 import (
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/Azure/azure-container-networking/npm/pkg/dataplane/policies"
 	npmerrors "github.com/Azure/azure-container-networking/npm/util/errors"
@@ -10,8 +12,10 @@ import (
 )
 
 const (
-	policyWithSets policyMode = "policyWithSets"
-	policyWithIPs  policyMode = "policyWithIPs"
+	policyWithSets     policyMode = "policyWithSets"
+	policyWithIPs      policyMode = "policyWithIPs"
+	maxNoNetRetryCount int        = 240 // max wait time 240*5 == 20 mins
+	maxNoNetSleepTime  int        = 5   // in seconds
 )
 
 func (dp *DataPlane) setPolicyMode() {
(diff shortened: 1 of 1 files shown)


npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -43,6 +47,34 @@ func (dp *DataPlane) initializeDataPlane() error {
 	return nil
 }
 
+func (dp *DataPlane) getNetworkInfo() error {
+	err := dp.setNetworkIDByName(AzureNetworkName)
```
huntergregory, line 51: We can remove the if-statement and all other code outside of it because the ticker starts right away (if I'm not mistaken)

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -43,6 +47,34 @@ func (dp *DataPlane) initializeDataPlane() error {
...
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				err := dp.setNetworkIDByName(AzureNetworkName)
+				if err == nil || !isNetworkNotFoundErr(err) {
+					return nil
+				}
+				retryNumber++
```
huntergregory, line 68: perhaps log when we retry?

huntergregory: or log every 50th try or so since the count is high

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -43,6 +47,34 @@ func (dp *DataPlane) initializeDataPlane() error {
...
+	if isNetworkNotFoundErr(err) {
+		retryNumber := 0
+		ticker := time.NewTicker(time.Second * time.Duration(maxNoNetSleepTime))
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				err := dp.setNetworkIDByName(AzureNetworkName)
+				if err == nil || !isNetworkNotFoundErr(err) {
```
huntergregory, line 65: Discrepancy between code in this if-statment and outside. For an error that isn't networkNotFound, we return nil here, but err outside

npm/pkg/dataplane/dataplane_windows.go
vakalapa: Ticker by default does not start right away, had to change it up a bit.

huntergregory (approved):

PR #1158, opened 2021-12-07
appinsights zapcore

rbtr:

**Reason for Change**:

Adds `zapai` package containing an MVP working AppInsights implementation of a zap.Core.
Because we need to improve our logging framework (and zap would be good for that) but things like CNS need to send logs via AI.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


zapai/core.go
```diff
@@ -0,0 +1,125 @@
...
+		fieldMappers: make(map[string]fieldTagMapper),
+		out:          out,
+	}
+}
+
+func (c *Core) WithFieldMappers(fieldMappers ...map[string]string) *Core {
+	clone := c.clone()
+	for _, fieldMapper := range fieldMappers {
+		for field, tag := range fieldMapper {
+			field, tag := field, tag // anchor vars in the closure
```
ramiro-gamarra, line 51: `field` is not closed over, does it need to be new var in this scope?

zapai/sink.go
```diff
@@ -0,0 +1,159 @@
...
+		dec:        newTraceDecoder(),
+	}
+}
+
+// Write accepts a gob []byte that must be Decodable to an appinsights.TraceTelemetry{}, which is then sent
+// to appinsights via the telemetryTracker.
+func (s *Sink) Write(b []byte) (int, error) {
+	t, err := s.dec.decode(b)
+	if err != nil {
+		return -1, errors.Wrap(err, "sink failed to decode trace")
```
ramiro-gamarra, line 136: `-1` looks like an in-band error, is it needed here?

zapai/example/main.go
```diff
@@ -0,0 +1,46 @@
+package main
+
+import (
+	"net/http"
+	"os"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
```
ramiro-gamarra, line 8: is this an example of just `zap`? don't see references to `zapai`

ramiro-gamarra (commented):
Just some very minor things. Would be great to see a demo. It's a pleasure reviewing a PR like this.

zapai/example/main.go
thatmattlong: If I'm reading this right, the AI sink is registered in an init function of the zapai package (`zapai/sink.go`) which I (having no experience with zap) assume means we just need to `import _ "github.com/Azure/azure-container-networking/zapai"` to get the side-effect?

rbtr: i accidentally dropped the ai zapcore stash, i have updated this file with example usages for the standard json logger, logfmt, and zapai

rbtr: > the AI sink is registered in an init function of the zapai package

this registers our[ ai sink factory method](https://github.com/Azure/azure-container-networking/pull/1158/files#diff-4482faabbd028f6614b30a0970ef8c3695da11e46b5f70ff0c214a9dd24df620R37) with the zap sink machinery, so that we can `zap.Open("appinsights://xxx")` (the URI should be generated [by the method attached to the config struct](https://github.com/Azure/azure-container-networking/pull/1158/files#diff-4482faabbd028f6614b30a0970ef8c3695da11e46b5f70ff0c214a9dd24df620R82)) and we get a zapai.Sink.

zapai/core.go
rbtr: updated

zapai/sink.go
rbtr: 0 works just as well, updated.

ramiro-gamarra (approved):

PR #1163, opened 2021-12-07
feat: [NPM] define transport shim layer using gRPC

nitishm:
User story AB#12707964

**Reason for Change**:
Define protobuf message and services for transport shim

**Issue Fixed**:
AB#12707965

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

Tasks:

- [X] Define protobuf file for all Datapath interface methods

**Notes**:


rbtr:
how are you generating the pb.go? would be nice to define gen targets in the makefile for consistency 🙂 
even better if you also don't assume protoc is installed and take care of that in a similar fashion as the existing go tools dependency targets.

nitishm:
> how are you generating the pb.go? would be nice to define gen targets in the makefile for consistency 🙂 even better if you also don't assume protoc is installed and take care of that in a similar fashion as the existing go tools dependency targets.

Sure I can do that. Are we using it any place else today? I did see a few proto files across the code base.

vakalapa:
> > how are you generating the pb.go? would be nice to define gen targets in the makefile for consistency 🙂 even better if you also don't assume protoc is installed and take care of that in a similar fashion as the existing go tools dependency targets.
> 
> Sure I can do that. Are we using it any place else today? I did see a few proto files across the code base.

@nitishm  https://github.com/Azure/azure-container-networking/blob/master/npm/pkg/dataplane/Makefile Check this, Mat has created this MakeFile for auto-generating go mock. You can do something similar. My opinion is this does not have to be in global MakeFile, since we will only generate this once in a while.

scripts/install-protoc.sh
```diff
@@ -0,0 +1,29 @@
...
+		unzip -o ${PROTOC_ZIP_NAME} -d ${PROTOC_INSTALL_PATH}
+		chmod +x ${PROTOC_INSTALL_PATH}/bin/protoc 
+		export PATH=${PROTOC_INSTALL_PATH}/bin:$PATH
+		echo "Removing protoc zip"
+		rm ${PROTOC_ZIP_NAME}
+else
+	echo "Protobuf compiler already installed at ${PROTOC_INSTALL_PATH}/bin/protoc"
+	echo "Add protoc to your PATH if you want to use it from anywhere else using the following command:"
+	echo "export PATH=${PROTOC_INSTALL_PATH}/bin:\$PATH"
+fi
```
rbtr, line 29: can you set up your editor to add a trailing newline on save?

Makefile
```diff
@@ -425,6 +425,9 @@ $(CONTROLLER_GEN): $(TOOLS_DIR)/go.mod
 
 controller-gen: $(CONTROLLER_GEN) ## Build controller-gen
 
+protoc: 
+	./scripts/install-protoc.sh
```
rbtr, line 429: we don't have a well-defined project layout, but should this scripts dir be under `/build` maybe?

nitishm: From what I have observed most scripts go into $ROOT/scripts (or at least in the kubebuilder project and other go projects). But if build seems like the right place for all scripts then we should definitely move it there. 
@vakalapa chime in?

scripts/install-protoc.sh
nitishm: Oh - how do I do that? I am using vscode

nitishm:
To test this out use the following instructions:

Download the protoc binary
```terminal
./scripts/install-protoc.sh 
```

Install the grpc go gen tool
```terminal
make tools
```

Run generate target
```terminal
cd npm
make generate
```

scripts/install-protoc.sh
rbtr: Settings > Files: Insert Final Newline

npm/pkg/transport/pb/transport.proto
```diff
@@ -0,0 +1,47 @@
...
+
+// DataplaneEvents represents the Service RPC exposed by the gRPC server.
+service DataplaneEvents{
+	rpc Connect(DatapathPodMetadata) returns (stream Events);
+}
+
+// DatapathPodMetadata is the metadata for a datapath pod
+message DatapathPodMetadata {
+  string id = 1; // Daemonset Pod ID
+  string node = 2; // Node name
```
vakalapa, line 15: Do we need a separate ID ? because we can use Node name as the ID.

vakalapa (approved):
Approving with minor comments.

npm/pkg/transport/pb/transport.proto
nitishm: I think pod name would be good to track (for future use probably)?

nitishm:
@vakalapa can I get another approval. Also any idea why the CI doesn't like it?

npm/pkg/transport/pb/transport.proto
```diff
@@ -0,0 +1,47 @@
+syntax = "proto3";
+package pb;
+option go_package = "github.com/Azure/azure-container-networking/npm/pkg/transport/pb;pb";
```
ramiro-gamarra, line 3: For package names, it might be better to communicate domain. In this case `pb` seems a bit generic and it might clash with other packages. Thoughts `dataplane` or `networkpolicy`? Seems like that is the domain of the services and types below.

npm/pkg/transport/pb/transport.proto
ramiro-gamarra: nit: maybe better to be very explicit about the identifiers? `pod_id` and `node_name` seem less ambiguous to me.

npm/pkg/transport/pb/transport.proto
```diff
@@ -0,0 +1,47 @@
...
+  // Payload can contain one or more Event objects.
+  repeated Event event = 3;
+}
+
+// Event is a generic object that can be Created, 
+// Updated, Deleted by the controlplane.
+message Event {
+  // Data can container one or more instances of IPSet or NetworkPolicy
+  // objects.
+	repeated google.protobuf.Struct data = 3;
```
ramiro-gamarra, line 46: I'm guessing this is still in prototyping stages, but I've usually found working with `Struct` difficult. Is the structure of IPSet and NetworkPolicy well defined? Could use `oneof`

ramiro-gamarra (commented):
Got curious about this, so I'm leaving a few comments/questions :)

npm/pkg/transport/pb/transport.proto
nitishm: I am afraid that we might get out of sync if things change where someone might forget to update the proto file

npm/pkg/transport/pb/transport.proto
nitishm: sure I can do that

npm/pkg/transport/pb/transport.proto
nitishm: Since it is local to the transport package I don't think it should matter

npm/pkg/transport/pb/transport.proto
rbtr: make updating the proto part of the build and fail if it diffs dirty

nitishm: How do you diff it with a standard go struct? The proto generated structure will most likely vary from the existing objects due to lack of type support.

ramiro-gamarra: you would diff what the build generates vs what's checked in. this ensures that go proto code is up to date with the source of truth, the proto file.

ramiro-gamarra: I still don't think I understand how `Struct` helps in this case. For prototyping purposes it should be fine, but it's a lot easier to for client and server to get out of sync when using it since it is basically like json and foregoes the benefits of strong typing.

rbtr: > diff what the build generates vs what's checked in

to expand, you `make gen` and then something like `test -z "$(git status --porcelain)"` which returns 0 if the pwd is clean and 1 if there are changes vs source

nitishm: > you would diff what the build generates vs what's checked in. this ensures that go proto code is up to date with the source of truth, the proto file.

I don't mean the proto generated code. I mean the actual IPSet, and other structs defined in `types.go`.

nitishm: > I still don't think I understand how `Struct` helps in this case. For prototyping purposes it should be fine, but it's a lot easier to for client and server to get out of sync when using it since it is basically like json and foregoes the benefits of strong typing.

I agree. This is not set in stone. Good for testing. I will play around with it and figure out what works best.

npm/pkg/transport/pb/transport.proto
vakalapa: Also, do we want to introduce an "api_version" field to help us maintain the structure in long run for upgrade cases ? Both for PodMetadata and also events

vakalapa: <Nit> we normally use "daemonPod" instead of DataplanePod, helps us with consistent naming.

npm/pkg/transport/pb/transport.proto
```diff
@@ -0,0 +1,47 @@
...
+  // The following events are supported:
+  enum EventType {
+    // Send a CREATE/UPDATE event to a datapath Pod
+    APPLY = 0;
+    // Send a DELETE event to a datapath Pod
+    REMOVE = 1;
+  }
+
+  ObjectType object = 2;
+  // Supported objects are IPSet rules and NetworkPolicy objects.
```
vakalapa, line 32: <nit> remove "rules" like below 

>// Supported objects are IPSet and NetworkPolicy objects.

npm/pkg/transport/pb/transport.proto
```diff
@@ -0,0 +1,47 @@
...
+    POLICY = 1;
+  }
+  // Payload can contain one or more Event objects.
+  repeated Event event = 3;
+}
+
+// Event is a generic object that can be Created, 
+// Updated, Deleted by the controlplane.
+message Event {
+  // Data can container one or more instances of IPSet or NetworkPolicy
```
vakalapa, line 44: <nit> container -> contain ?

npm/pkg/transport/pb/transport.proto
vakalapa: @nitishm  once we build a goatStateProcessor with the type generated from this proto file then if proto generation gets dirty, GSProcessor build should fail (ideally). Lets move forward at this moment with current "struct" design, this is a 2-way door, as we build bits around this, i am sure we will tweak this struct further, so it should be okay with current type. Wdygt ?

npm/pkg/transport/transport.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (m *Manager) start(ctx context.Context) {
+	go m.handle()
+
+	for {
+		select {
+		case client := <-m.regCh:
+			klog.Infof("Registering remote client %s", client)
+			m.Registrations[client.String()] = client
+		case client := <-m.deregCh:
```
vakalapa, line 72: there can some potential race conditions here ?  what if a connection got lost and created for the same pod in immediate succession, and lets say deregCh is overloaded and regCh is free, then there is a high chance that we would delete a legitimate connection right ? how do we manage those kinds of race conditions ?

We might need to introduce an aspect of timestamp ?

npm/pkg/transport/transport.go
```diff
@@ -0,0 +1,112 @@
...
+	"net"
+
+	"github.com/Azure/azure-container-networking/npm/pkg/transport/pb"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+	"google.golang.org/grpc/stats"
+	"k8s.io/klog/v2"
+)
+
+type Manager struct {
```
vakalapa, line 15: We might have to think about certs management in Manager too ?

vakalapa (commented):
Minor comments, please resolve all lints.

npm/pkg/transport/transport.go
nitishm: I think a ping-ping health check similar to Felix and typha is a must to avoid such issues. We need some keepalives and checks in place. Thoughts?

nitishm: Yea I think timestamps makes sense.

npm/pkg/transport/pb/transport.proto
nitishm: Sticking to Pod and Node to keep it generic. I don't understand why the file is showing the older version 
We have `pod` and `node` as the two fields

npm/pkg/transport/pb/transport.proto
```diff
@@ -0,0 +1,51 @@
...
+// DataplaneEvents represents the Service RPC exposed by the gRPC server.
+service DataplaneEvents{
+	rpc Connect(DatapathPodMetadata) returns (stream Events);
+}
+
+// DatapathPodMetadata is the metadata for a datapath pod
+message DatapathPodMetadata {
+  string pod = 1; // Daemonset Pod ID
+  string node = 2; // Node name
+  enum APIVersion {
```
nitishm, line 16: @vakalapa making this an enum so that we default to version 1 if not specified. Easier to track versions with enum. WDYT?

npm/pkg/transport/pb/transport.proto
nitishm: > nit: maybe better to be very explicit about the identifiers? `pod_id` and `node_name` seem less ambiguous to me.

Changed it to `pod` and `node` to keep it simple. The id suffix is redundant IMO. WDYT?

npm/pkg/transport/pb/transport.proto
nitishm: @ramiro-gamarra Thoughts?

npm/pkg/transport/pb/transport.proto
ramiro-gamarra: Don't let my feedback on this be a blocker for prototyping purposes btw :), just sharing my 2 cents. Thing about contracts is to keep ambiguity to a minimum IMO. `pod` could be understood as the name of the pod, or the uid of the pod, 2 different things. Maybe documentation could help in the proto file itself, but when used in the generated code, it would also have a bit of ambiguity.

npm/pkg/transport/pb/transport.proto
ramiro-gamarra: Not sure I understand the part about local. The imported package would still be `pb` and referenced as such across the code. If the client had a second grpc service to interact with for example, and its package was under `pkg/foo/pb`, it would also be imported as `pb` but would conflict with the other one, even if they are unrelated paths. Maybe I'm missing a detail, let me know

npm/pkg/transport/transport.go
```diff
@@ -0,0 +1,117 @@
...
+			return
+		}
+	}
+}
+
+func (m *Manager) handle() {
+	klog.Info("Starting transport manager listener")
+	lis, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", m.port))
+	if err != nil {
+		m.errCh <- fmt.Errorf("failed to handle server connections: %w", err)
```
ramiro-gamarra, line 94: should probably return here, otherwise the rest of the goroutine executes with a nil listener after the error is sent. maybe it would be better to have the grpc server wired synchronously and then spinning `server.Server` in a goroutine?

npm/pkg/transport/pb/transport.proto
nitishm: I see your point now. @vakalapa thoughts? Since this proto is common across both cp and dp how would we name it
Maybe we restructure as we get it it? @vakalapa what are your thoughts on this?

npm/pkg/transport/transport.go
nitishm: Yup that works. Making the change now.

nitishm: Done - PTAL @ramiro-gamarra

npm/pkg/transport/pb/transport.proto
vakalapa: Ack, i can see that confusion arising, and we will be expanding later on, so having a specific pkg name instead of pb would help for sure, 

Other option is to consolidate all protos into a central location under like below (calico does this) so all other pks can access these types without any cyclic issues 

>npm/pkg/protos

npm/pkg/transport/pb/transport.proto
vakalapa: that makes sense!

npm/pkg/transport/transport.go
```diff
@@ -110,8 +116,13 @@ func (m *Manager) handle() {
...
+
 	// Start gRPC Server in background
-	if err := server.Serve(lis); err != nil {
-		m.errCh <- fmt.Errorf("failed to start gRPC server: %w", err)
-	}
+	go func(errCh chan<- error) {
+		if err := server.Serve(lis); err != nil {
+			errCh <- fmt.Errorf("failed to start gRPC server: %w", err)
+		}
+	}(m.errCh)
```
nitishm, line 125: I cant recall if closures would prevent the member channel from being passed as-is into the go routine's scope or not. So I explicitly pass it in as a safety measure 🤷

nitishm, line 125: I cant recall if closures would prevent the member channel from being passed as-is into the go routine's scope or not. So I explicitly pass it in as a safety measure 🤷

rbtr: as long as it's not reassigned in a loop, the variable gets captured in the closure

rbtr: [this](https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables) is probably what you're thinking of?

nitishm: Ye I knew about that one but wasn't sure if channels might have a similar fate. Seems unlikely

npm/pkg/transport/consts.go
```diff
@@ -0,0 +1,7 @@
+package transport
+
+const (
+	// concurrentInputRegistrations = 10
+	grpcMaxConcurrentStreams = 100
+	remoteAddrContextKey     = "remote-addr"
```
ramiro-gamarra, line 6: seems like this is used as a key to fetch context values. usually these are their own unexported types in order to avoid key collision in the global scope.

nitishm: Done.

npm/pkg/transport/pb/transport.proto
nitishm: Sure I can update that to `pod_name` & `node_name`

npm/pkg/transport/pb/transport.proto
nitishm: Let me see what I can do

nitishm: Done

npm/pkg/transport/pb/transport.proto
nitishm: Done

vakalapa (approved):

PR #1167, opened 2021-12-08
feat: [NPM] iptables comments for v2

huntergregory:
iptables comments will allow us to debug/understand which policies and what ACL rules are which within iptables. For an overview of what the comments will look like, see the description in *iptables-comments_linux.go*.

This PR also updates the validation check for port/protocol constraints in *policy.go* so that ACLs with a namedport are required to have `UnspecifiedProtocol`


npm/pkg/dataplane/policies/iptables-comments_linux.go
```diff
@@ -0,0 +1,202 @@
...
+				- otherwise:
+					ns-[!]label1-AND-ns-[!]label2...
+			- other w/out ns:
+				[!]label1-AND-[!]label2...
+			- other w/ ns:
+				[!]label1-AND-[!]label2...-AND-ns-[!]labelM-IN-ns-name
+*/
+
+func (networkPolicy *NPMNetworkPolicy) commentForJumpToIngress() string {
+	return networkPolicy.commentForJump(true)
```
vakalapa, line 95: Can you change true/false to constant types for ingressJump or egressJump, that was it will be easier to read ?

npm/pkg/dataplane/policies/iptables-comments_linux.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (proto Protocol) comment() string {
+	if proto == UnspecifiedProtocol {
+		return ""
+	}
+	return fmt.Sprintf("-ON-%s", string(proto))
+}
+
+func (portRange *Ports) comment() string {
```
vakalapa, line 194: instead of having these  <type>.comment() functions for all in this file, can we place these near their type declarations? that will be cleaner in my opinion ? No strong opinion though

huntergregory: I think we should leave as is since comments are only used in Linux, and we don't have a policies_linux.go file

vakalapa (approved):

npm/pkg/dataplane/policies/iptables-comments_linux.go
rbtr: @vakalapa is correct, methods should be defined next to their receivers
the farther a declaration is from the usage, the harder it is to grok

huntergregory: edit: we do have a policies_linux.go file. Might make sense to compile iptables code into one file?

huntergregory: as discussed offline, moving these Linux-specific codes to policy_linux.go

vakalapa (approved):

PR #1188, opened 2022-01-08
feat: [NPM] call policy reconcile in dataplane

huntergregory:
Adds a go-routine for policy manager reconciling, which:
1. adds the jump from AZURE-NPM to FORWARD chain
2. deletes stale chains

Lock stale chains when:
- modifying the stale chains cache
- deleting chains that were in `staleChains` (in `reconcile()`)
- adding policy chains/rules via iptables-restore in case a chain is in `staleChains`

As an optimization, policy manager can "force lock" `staleChains`, which pauses stale chain deletion until next reconcile period.


npm/pkg/dataplane/policies/policymanager_linux.go
```diff
@@ -43,14 +48,17 @@ func (pMgr *PolicyManager) removePolicy(networkPolicy *NPMNetworkPolicy, _ map[s
...
+	chainsToDelete := chainNames([]*NPMNetworkPolicy{networkPolicy})
 	creator := pMgr.creatorForRemovingPolicies(chainsToDelete)
 	restoreErr := restore(creator)
 	if restoreErr != nil {
 		return npmerrors.SimpleErrorWrapper("failed to flush policies", restoreErr)
 	}
 
 	// 3. Delete policy chains in the background.
+	// lock here since stale chains are only affected if we successfully remove policies
+	pMgr.staleChains.forceLock()
```
vakalapa, line 60: 1. calculate creator:
2. acquire reconcileLock (stalechain lock )
3. Delete jump rules
4. restore

npm/pkg/dataplane/dataplane.go
```diff
@@ -24,9 +24,12 @@ type PolicyMode string
 type Config struct {
 	*ipsets.IPSetManagerCfg
 	*policies.PolicyManagerCfg
+	// helpful for UTs (defaults to false for external packages)
+	disableGoRoutines bool
```
nitishm, line 28: can we rename this to something like `concurrencyEnabled` or `concurrencyDisabled`

npm/pkg/dataplane/dataplane.go
```diff
@@ -48,23 +51,28 @@ type NPMEndpoint struct {
...
 
 	err := dp.ResetDataPlane()
 	if err != nil {
 		klog.Errorf("Failed to reset dataplane: %v", err)
 		return nil, err
 	}
+	// necessary for UTs because of ioshim
+	if !dp.disableGoRoutines {
+		dp.policyMgr.Reconcile(dp.stopChannel)
+	}
```
nitishm, line 75: so if this is disabled we're not running reconcile at all? 😕  or am I misunderstanding this

npm/pkg/dataplane/dataplane.go
nitishm: Looking at what it does below - maybe something like `disableReconcile`?

npm/pkg/dataplane/dataplane.go
huntergregory: yes, this is just for unit testing because we test exec calls, and don't want to intermingle exec calls from reconciling with other exec calls

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -64,7 +66,25 @@ type staleChains struct {
...
+	rm.Lock()
+}
+
+// forceUnlock makes sure that the releaseLockSignal channel is empty (in case reconciling
+// wasn't running when forceLock was called), and then unlocks the reconcileManager.
+func (rm *reconcileManager) forceUnlock() {
+	select {
+	case <-rm.releaseLockSignal:
+	default:
+	}
```
nitishm, line 86: You dont need a select block when using just one channel/case.
```suggestion
	<-rm.releaseLockSignal:
```

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -128,6 +148,11 @@ func isBaseChain(chain string) bool {
 func (pMgr *PolicyManager) bootup(_ []string) error {
 	klog.Infof("booting up iptables Azure chains")
 
+	// Stop reconciling so we don't centend for iptables, and so we don't update the staleChains at the same time as reconcile()
+	// Reconciling would only be happening if this function were called to reset iptables well into the azure-npm pod lifecycle.
+	pMgr.reconcileManager.forceLock()
+	defer pMgr.reconcileManager.forceUnlock()
+
 	// 1. delete the deprecated jump to AZURE-NPM
 	deprecatedErrCode, deprecatedErr := pMgr.runIPTablesCommand(util.IptablesDeletionFlag, deprecatedJumpFromForwardToAzureChainArgs...)
```
nitishm, line 157: Do we want it locked for both `pMgr.runIPTablesCommand` and `pMgr.allCurrentAzureChains()`? If not we can just wrap the portion with lock/unlock without using defer

npm/pkg/dataplane/policies/chain-management_linux_test.go
```diff
@@ -47,6 +47,50 @@ Chain AZURE-NPM-ACCEPT (1 references)
...
+	// don't verify calls because there shouldn't be as many commands as we create if forceLock works properly
+	pMgr := NewPolicyManager(ioshim, ipsetConfig)
+
+	start := make(chan struct{}, 1)
+	done := make(chan struct{}, 1)
+	go func() {
+		pMgr.reconcileManager.Lock()
+		defer pMgr.reconcileManager.Unlock()
+		start <- struct{}{}
+		require.NoError(t, pMgr.cleanupChains(testChains))
```
nitishm, line 69: for cleanupChains function can we move https://github.com/Azure/azure-container-networking/blob/fea82838d9c7b8c23f20eaca9e8b16789e40a1e6/npm/pkg/dataplane/policies/chain-management_linux.go#L227-L237 to the `default` case in the select loop above it?

npm/pkg/dataplane/policies/chain-management_linux.go
huntergregory: The idea of this code is to empty the signal, which might be empty already. This is where the default case comes in because then we might come into deadlock.

npm/pkg/dataplane/policies/chain-management_linux.go
huntergregory: Ya we'd like to for runIPTables, allCurrentChains, and restore since they all use iptables. We want to lock for this whole period so we don't start using iptables in the other go routine

test/integration/npm/main.go
```diff
@@ -73,7 +74,7 @@ var (
 )
 
 func main() {
-	dp, err := dataplane.NewDataPlane(nodeName, common.NewIOShim(), dpCfg)
+	dp, err := dataplane.NewDataPlane(nodeName, common.NewIOShim(), dpCfg, nil)
```
nitishm, line 77: Any reason to keep the stop channel nil? Can we pass a legit channel?

test/integration/npm/main.go
```diff
@@ -162,6 +163,9 @@ func main() {
 	printAndWait(true)
 	panicOnError(dp.AddPolicy(policies.TestNetworkPolicies[0]))
 	fmt.Println("AZURE-NPM should have rules now")
+
+	fmt.Println("waiting for reconcile to finish (will be a while if you don't update the reconcile time in policymanager.go")
+	time.Sleep(finalWaitTimeInMinutes * time.Minute)
```
nitishm, line 168: Why do we need this? Is there a channel we could introduce into the code that would signal that things are done instead? This slows down the integration tests

nitishm (changes requested):
functionality wise I think it LGTM. But I have some nits and comments/questions

test/integration/npm/main.go
huntergregory: we don't run these in our pipeline (I just run it locally), so speed shouldn't be a problem

test/integration/npm/main.go
huntergregory: no reason. I can do that

npm/pkg/dataplane/dataplane.go
```diff
@@ -48,23 +51,28 @@ type NPMEndpoint struct {
...
+		stopChannel:    stopChannel,
 	}
 
 	err := dp.ResetDataPlane()
 	if err != nil {
 		klog.Errorf("Failed to reset dataplane: %v", err)
 		return nil, err
 	}
+	// necessary for UTs because of ioshim
+	if !dp.disableReconcileForUTs {
```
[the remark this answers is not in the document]
huntergregory: Removed this flag and created a `dp.RunPeriodicTasks()`

vakalapa (approved):
Great work ! 💯

nitishm (approved):
LGTM

PR #1189, opened 2022-01-10
feat: [NPM] Create controlplane manager for decomposed model

nitishm:
Implement the controlplane component of the decomposed model of
operation.
- Start the transport server in addition to the different watch
  controllers

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


npm/config/config.go
```diff
@@ -3,16 +3,23 @@ package npmconfig
 const (
 	defaultResyncPeriod  = 15
 	defaultListeningPort = 10091
-
+	defaultGrpcPort      = 10092
```
vakalapa, line 6: This makes me think, we should add a default network policy within kube-system NS to always allow these two ports to NPM pods.

npm/controlplane.go
```diff
@@ -0,0 +1,145 @@
...
+	informerFactory informers.SharedInformerFactory,
+	dp dataplane.GenericDataplane,
+	npmVersion string,
+	k8sServerVersion *version.Info,
+) *NetworkPolicyControlplane {
+	klog.Infof("API server version: %+v AI metadata %+v", k8sServerVersion, aiMetadata)
+
+	mgr := transport.NewManager(ctx, config.Transport.Port)
+
+	// TODO (nitishm): This needs to be passed to the Dataplane instance
```
nitishm, line 48: @vakalapa Here the input channel needs to be passed to the Generic(?)Dataplane instance since `ApplyDatapath` would write to this channel.

npm/controlplane.go
```diff
@@ -0,0 +1,145 @@
...
+	if !cache.WaitForCacheSync(stopCh, n.nsInformer.Informer().HasSynced) {
+		return fmt.Errorf("Namespace informer failed to sync")
+	}
+
+	if !cache.WaitForCacheSync(stopCh, n.npInformer.Informer().HasSynced) {
+		return fmt.Errorf("Network policy informer failed to sync")
+	}
+
+	// TODO (nitishm): The input channel for the transport layer needs to be
+	// passed to the Dataplane instance (i.e. ApplyDataplane should use this channel to send messages)
```
nitishm, line 132: Should probably happen in the constructor and not here.

npm/controlplane.go
```diff
@@ -0,0 +1,145 @@
...
+	go n.podControllerV2.Run(stopCh)
+	go n.namespaceControllerV2.Run(stopCh)
+	go n.netPolControllerV2.Run(stopCh)
+
+	// start the transport layer (gRPC) server
+	// We block the main thread here until the server is stopped.
+	// This is unlike the other start methods in this package, which returns nil
+	// and blocks in the main thread during command invocation through the select {}
+	// statement.
+	return n.tm.Start(stopCh)
```
nitishm, line 144: We want to handle any errors returned by this, so instead of looping forever in the invoker function we block here and return any error when things fail for any reason.

npm/controlplane.go
vakalapa: I am working on Control Plane DP Shim and can push a skeleton along with GS Processor.

npm/cmd/start_controlplane.go
```diff
@@ -0,0 +1,109 @@
...
+	klog.Infof("loaded config: %+v", config)
+	klog.Infof("Start NPM version: %s", version)
+
+	var err error
+
+	err = initLogging()
+	if err != nil {
+		return err
+	}
+
```
vakalapa, line 58: We will need to initialize metrics here ?

npm/cmd/start_controlplane.go
```diff
@@ -0,0 +1,109 @@
...
+	// Adding some randomness so all NPM pods will not request for info at once.
+	factor := rand.Float64() + 1 //nolint
+	resyncPeriod := time.Duration(float64(minResyncPeriod.Nanoseconds()) * factor)
+	klog.Infof("Resync period for NPM pod is set to %d.", int(resyncPeriod/time.Minute))
+	factory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
+
+	k8sServerVersion := k8sServerVersion(clientset)
+
+	var dp dataplane.GenericDataplane
+	dp, err = dataplane.NewDataPlane(npm.GetNodeName(), common.NewIOShim(), npmV2DataplaneCfg)
```
vakalapa, line 94: This will be DPShim, can you add a todo saying this needs to be replaced? In control plane we will not be using the actual DP instead the DP's gRPC implementation (DPShim)

npm/cmd/start_dataplane.go
```diff
@@ -0,0 +1,50 @@
...
+	klog.Infof("loaded config: %+v", config)
+	klog.Infof("Start NPM version: %s", version)
+
+	err := initLogging()
+	if err != nil {
+		klog.Errorf("failed to init logging : %v", err)
+		return err
+	}
+
+	n, err := npm.NewNetworkPolicyDataplane(context.Background(), config)
```
vakalapa, line 43: i think we might need to change the naming convention here, ideally this would be newNetworkPolicyDaemon, and the CP would be newNetworkPolicyController.

Because we have verticals and horizontal distinctions, 

Horizontal divisions are the different pods: Controller Pod and Daemon Pod.
Verticals divisions are: ControlPlane and Dataplane.Because in each of the above pods we have a CP and DP component.

Wdyt ?

npm/controlplane.go
```diff
@@ -0,0 +1,144 @@
...
+	dp dataplane.GenericDataplane,
+	npmVersion string,
+	k8sServerVersion *version.Info,
+) *NetworkPolicyControlplane {
+	klog.Infof("API server version: %+v AI metadata %+v", k8sServerVersion, aiMetadata)
+
+	mgr := transport.NewManager(context.Background(), config.Transport.Port)
+
+	// TODO (nitishm): This needs to be passed to the Dataplane instance
+	_ = mgr.InputChannel()
```
vakalapa, line 48: i think we will need to push TM one step above, because, DP expects the input channel to be provided when the object is created. So start.go (for controller) can be our main orch point, no strong opinion though

npm/cmd/start_dataplane.go
nitishm: Sounds fair to me. I will rename it.

npm/cmd/start_controlplane.go
nitishm: What metrics are we referring to?

npm/cmd/start_dataplane.go
```diff
@@ -0,0 +1,50 @@
...
+	npmconfig "github.com/Azure/azure-container-networking/npm/config"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+)
+
+func newStartNPMDataplaneCmd() *cobra.Command {
+	startNPMDataplaneCmd := &cobra.Command{
+		Use:   "dataplane",
```
matmerr, line 17: after dockerfile update, what's the path for reverting out of dataplane/control plane architecture? will `dataplane` or `controlplane` retain classic behavior?

npm/pkg/controlplane/goalstateprocessor/goalstateprocessor.go
```diff
@@ -52,11 +60,20 @@ func (gsp *GoalStateProcessor) Stop() {
...
 }
 
-func (gsp *GoalStateProcessor) run() {
+func (gsp *GoalStateProcessor) run(stopCh <-chan struct{}) {
 	klog.Infof("Starting dataplane for node %s", gsp.nodeID)
 
 	for {
-		gsp.processNext()
+		select {
+		case <-gsp.ctx.Done():
```
vakalapa, line 68: Can you remove redundant, gsp.ctx.Done() in processNext ?

npm/daemon.go
```diff
@@ -0,0 +1,52 @@
...
+		ctx:     ctx,
+		config:  config,
+		gsp:     gsp,
+		client:  client,
+		version: npmVersion,
+	}, nil
+}
+
+func (n *NetworkPolicyDaemon) Start(config npmconfig.Config, stopCh <-chan struct{}) {
+	go n.gsp.Start(stopCh)
```
vakalapa, line 45: gsp.Start spawns a go routine already, this is redundant ?

npm/manager.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2018 Microsoft. All rights reserved.
+// MIT License
+package npm
```
vakalapa, line 3: Is manager.go the right name for this file ?

npm/pkg/transport/controlplane.go
```diff
@@ -119,11 +119,14 @@ func (m *Manager) start() error {
 				}
 			}
 		case <-m.ctx.Done():
-			klog.Info("Stopping transport manager")
+			klog.Info("Context Done. Stopping transport manager")
 			return nil
 		case err := <-m.errCh:
 			klog.Errorf("Error in transport manager: %v", err)
 			return err
+		case <-stopCh:
```
vakalapa, line 127: Not related to this line but, NewManager and Manager is confusing, can we have the right names for this struct ? 

may be ControllerEventsManager ?

vakalapa (changes requested):
Initial component naming comments.

npm/cmd/start_daemon.go
```diff
@@ -0,0 +1,88 @@
...
+	ctx := context.Background()
+	err := initLogging()
+	if err != nil {
+		klog.Errorf("failed to init logging : %v", err)
+		return err
+	}
+
+	var dp dataplane.GenericDataplane
+
+	dp, err = dataplane.NewDataPlane(npm.GetNodeName(), common.NewIOShim(), npmV2DataplaneCfg)
```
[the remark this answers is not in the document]
nitishm: in `start.go`

npm/pkg/transport/controlplane.go
nitishm: If you read it using the fully-qualified path i.e. `transport.Manager` it makes sense to keep it this way. It states that this is our manager that encapsulates the server and watchdog as well. My suggestion is to keep it this way.

npm/cmd/start_controlplane.go
vakalapa: https://github.com/Azure/azure-container-networking/blob/3d1e7399f6abd72888b13c6464479f88c91cbaba/npm/cmd/start.go#L103 

this metrics initialize.

huntergregory: Metrics are initialized in start.go, but initializing again couldn't hurt. I'll have to schedule a meeting to discuss plans for metrics in the fan-out model

npm/pkg/transport/controlplane.go
vakalapa: right but searching for it will be a nightmare for new joinees, probably transport.TransportManager will be better,

npm/cmd/start.go
```diff
@@ -85,6 +85,9 @@ func newStartNPMCmd() *cobra.Command {
 
 	startNPMCmd.Flags().String(flagKubeConfigPath, flagDefaults[flagKubeConfigPath], "path to kubeconfig")
 
+	startNPMCmd.AddCommand(newStartNPMControlplaneCmd())
```
huntergregory, line 88: can we add comments here and/or for newStartNPMDaemonCmd()? Saying that this daemon refers to just the fan-out mode

huntergregory (commented):
Do we want to delegate fan-out vs regular mode to the Dockerfile (start commands), or should it be configurable via the NPM config?

npm/cmd/start.go
nitishm: done

npm/npm_v1_test.go
```diff
@@ -24,14 +23,14 @@ const (
...
 
-func getKey(obj interface{}, t *testing.T) string {
-	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
-	if err != nil {
-		t.Errorf("Unexpected error getting key for obj %v: %v", obj, err)
-		return ""
-	}
-	return key
-}
+// func getKey(obj interface{}, t *testing.T) string {
```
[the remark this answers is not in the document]
nitishm: done

npm/pkg/transport/controlplane.go
nitishm: renamed to GrpcManager

npm/cmd/start_dataplane.go
nitishm: As discussed in the call,
- `start` command would start in the current model
- `start controlplane` and `start deamon` would start the controller and daemon in the decomposed mode.

npm/consts.go
```diff
@@ -0,0 +1,17 @@
+package npm
+
+const (
```
vakalapa, line 3: Can we combine these two consts ?

nitishm: done

vakalapa (approved):
Lgtm 🚢

PR #1192, opened 2022-01-11
guard against unstructured url forwarding in CNS API's

aegal:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/nmagent/client.go
+++ b/cns/nmagent/client.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/Azure/azure-container-networking/cns/logger"
@@ -20,14 +21,16 @@ const (
 	GetNmAgentSupportedApiURLFmt       = "http://%s/machine/plugins/?comp=nmagent&type=GetSupportedApis"
 	GetNetworkContainerVersionURLFmt   = "http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/%s/networkContainers/%s/version/authenticationToken/%s/api-version/1"
 	GetNcVersionListWithOutTokenURLFmt = "http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/api-version/%s"
-	JoinNetworkURLFmt                  = "http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/joinedVirtualNetworks/%s/api-version/1"
-	PutNetworkContainerUrlFmt          = "http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/%s/networkContainers/%s/authenticationToken/%s/api-version/1"
-	DeleteNetworkContainerURLFmt       = "http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/%s/networkContainers/%s/authenticationToken/%s/api-version/1/method/DELETE"
+	JoinNetworkURLFmt                  = "NetworkManagement/joinedVirtualNetworks/%s/api-version/1"
+	PutNetworkValueFmt                 = "NetworkManagement/interfaces/%s/networkContainers/%s/authenticationToken/%s/api-version/1"
+	DeleteNetworkContainerURLFmt       = "NetworkManagement/interfaces/%s/networkContainers/%s/authenticationToken/%s/api-version/1/method/DELETE"
 )
 
 // WireServerIP - wire server ip
 var (
 	WireserverIP                           = "168.63.129.16"
+	WireServerPath 						   =  "machine/plugins"
+	WireServerScheme					   =  "http"
 	getNcVersionListWithOutTokenURLVersion = "2"
 )
 
@@ -75,12 +78,23 @@ func JoinNetwork(networkID string) (*http.Response, error) {
 	var body bytes.Buffer
 	json.NewEncoder(&body).Encode("")
 
-	joinNetworkURL := fmt.Sprintf(
+	joinNetworkTypeValue := fmt.Sprintf(
 		JoinNetworkURLFmt,
-		WireserverIP,
 		networkID)
 
-	response, err := common.GetHttpClient().Post(joinNetworkURL, "application/json", &body)
+	joinNetworkUrl := url.URL{
+		Host:   WireserverIP,
+		Path:   WireServerPath,
+		Scheme: WireServerScheme,
+	}
+
+	queryString := joinNetworkUrl.Query()
+	queryString.Set("type", joinNetworkTypeValue)
+	queryString.Set("comp", "nmagent")
+
+	joinNetworkUrl.RawQuery = queryString.Encode()
+
+	response, err := common.GetHttpClient().Post(joinNetworkUrl.String(), "application/json", &body)
 
 	if err == nil && response.StatusCode == http.StatusOK {
 		defer response.Body.Close()
@@ -96,15 +110,29 @@ func JoinNetwork(networkID string) (*http.Response, error) {
 func PublishNetworkContainer(networkContainerID, associatedInterfaceID, accessToken string, requestBodyData []byte) (*http.Response, error) {
 	logger.Printf("[NMAgentClient] PublishNetworkContainer NC: %s", networkContainerID)
 
-	createNCURL := fmt.Sprintf(
-		PutNetworkContainerUrlFmt,
-		WireserverIP,
+	createNcTypeValue := fmt.Sprintf(
+		PutNetworkValueFmt,
 		associatedInterfaceID,
 		networkContainerID,
 		accessToken)
 
+	createUrl := url.URL{
+		Host:   WireserverIP,
+		Path:   WireServerPath,
+		Scheme: WireServerScheme,
+	}
+
+	queryString := createUrl.Query()
+	queryString.Set("type", createNcTypeValue)
+	queryString.Set("comp", "nmagent")
+
+	createUrl.RawQuery = queryString.Encode()
+
 	requestBody := bytes.NewBuffer(requestBodyData)
-	response, err := common.GetHttpClient().Post(createNCURL, "application/json", requestBody)
+	response, err := common.GetHttpClient().Post(createUrl.String(), "application/json", requestBody)
+
+	logger.Printf("[NMAgentClient][Response] Create NC URL: %s",
+		createUrl.String())
 
 	logger.Printf("[NMAgentClient][Response] Publish NC: %s. Response: %+v. Error: %v",
 		networkContainerID, response, err)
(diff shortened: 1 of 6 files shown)


cns/nmagent/client.go
```diff
@@ -84,11 +93,18 @@ func JoinNetwork(networkID, joinNetworkURL string) (*http.Response, error) {
 }
 
 // PublishNetworkContainer publishes given network container
-func PublishNetworkContainer(networkContainerID, createNetworkContainerURL string, requestBodyData []byte) (*http.Response, error) {
+func PublishNetworkContainer(networkContainerID, associatedInterfaceID, accessToken string, requestBodyData []byte) (*http.Response, error) {
 	logger.Printf("[NMAgentClient] PublishNetworkContainer NC: %s", networkContainerID)
 
+	createNCURL := fmt.Sprintf(
+		PutNetworkContainerUrlFmt,
+		WireserverIP,
```
aegal, line 101: With this we are going with the wireserver IP that is in CNS, let me know if you think that isn't best.

cns/restserver/api.go
```diff
@@ -1148,19 +1180,16 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
 			defer publishResponse.Body.Close()
 		}
 
-		// Store ncGetVersionURL needed for calling NMAgent to check if vfp programming is completed for the NC
-		primaryInterfaceIdentifier := getInterfaceIdFromCreateNetworkContainerURL(req.CreateNetworkContainerURL)
-		authToken := getAuthTokenFromCreateNetworkContainerURL(req.CreateNetworkContainerURL)
-
 		// we attempt to extract the wireserver IP to use from the request, otherwise default to the well-known IP.
 		hostIP := extractHostFromJoinNetworkURL(req.JoinNetworkURL)
```
rbtr, line 1184: should remove this from all of these paths

cns/nmagent/client.go
```diff
@@ -84,11 +93,18 @@ func JoinNetwork(networkID, joinNetworkURL string) (*http.Response, error) {
 }
 
 // PublishNetworkContainer publishes given network container
-func PublishNetworkContainer(networkContainerID, createNetworkContainerURL string, requestBodyData []byte) (*http.Response, error) {
+func PublishNetworkContainer(networkContainerID, associatedInterfaceID, accessToken string, requestBodyData []byte) (*http.Response, error) {
 	logger.Printf("[NMAgentClient] PublishNetworkContainer NC: %s", networkContainerID)
 
+	createNCURL := fmt.Sprintf(
```
rbtr, line 99: I think the URLs should be built from a url.URL, not by string templating

rbtr (changes requested):

cns/nmagent/client.go
rbtr: a common pattern for a client pkg is to have a baseURL that has the scheme and host set (and port if required), make copies of it in the client methods, then set the paths and queries on the copies and use url.String() to make the final input string for the http call.

cns/restserver/api.go
```diff
@@ -1072,14 +1073,32 @@ func (service *HTTPRestService) getNumberOfCPUCores(w http.ResponseWriter, r *ht
 	logger.Response(service.Name, numOfCPUCoresResp, resp.ReturnCode, err)
 }
 
-func getInterfaceIdFromCreateNetworkContainerURL(
-	createNetworkContainerURL string) string {
-	return strings.Split(strings.Split(createNetworkContainerURL, "interfaces/")[1], "/")[0]
-}
+func getAuthTokenAndInterfaceIdFromNcURL(networkContainerURL string) (string, string, error) {
```
ramiro-gamarra, line 1076: This is technically returning interface id and auth token in that order. Since the return types of the func are repeated strings, it's not clear from the call site what each actually represents. I'd suggest changing the function name to reflect the order of the returns and name the returns in the function signature, or have the function return a type that makes it very explicit, along the lines of:
```go
type ncURLParams struct {
    ncID        string
    interfaceID string
    authToken   string
}
```

ramiro-gamarra (commented):
few initial comments

cns/restserver/api.go
```diff
@@ -1119,14 +1140,24 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
...
+	// TODO - refactor this method for better error handling
 	if err != nil {
 		return
 	}
 
+	var associatedInterfaceID, authToken string
+	associatedInterfaceID, authToken, err = getAuthTokenAndInterfaceIdFromNcURL(creteNcUrlCopy)
+	if err != nil {
+		logger.Errorf("[Azure-CNS] %+v", err)
+		w.WriteHeader(http.StatusBadRequest)
```
ramiro-gamarra, line 1153: - more context on log line would helpful, like the action being taken at the moment of error (publish network container) and things that may not be in the error string (like nc id)
- is there no payload returned to client? i think this change merits a response body with error details, since otherwise all we will see in DNC logs is 400

cns/restserver/api_test.go
```diff
@@ -515,25 +515,60 @@ func createNC(
...
 
 func TestPublishNCViaCNS(t *testing.T) {
 	fmt.Println("Test: publishNetworkContainer")
-	publishNCViaCNS(t, "vnet1", "ethWebApp")
+
+	createNetworkContainerURL := "http://" + nmagentEndpoint +
+		"/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/dummyIntf/networkContainers/dummyNCURL/authenticationToken/dummyT/api-version/1"
+	err := publishNCViaCNS(t, "vnet1", "ethWebApp", createNetworkContainerURL, false)
+	if err != nil {
+		t.Fatal(err)
```
ramiro-gamarra, line 534: more context on logged error. same for all below

cns/restserver/api.go
aegal: Added more context and returned a cns response that DNC will parse

cns/restserver/api_test.go
aegal: Added

cns/restserver/api.go
aegal: Changed result to a struct

cns/restserver/api.go
```diff
@@ -1072,14 +1073,32 @@ func (service *HTTPRestService) getNumberOfCPUCores(w http.ResponseWriter, r *ht
...
+	// current format of create network url has a path after a query parameter "type"
+	// doing this parsing due to this structure
+	typeQueryParamVal := queryParams.Get("type")
+	if typeQueryParamVal == "" {
+		return "", "", fmt.Errorf("Invalid create container url format, no type query param")
+	}
+
+	// .{0,128} gets from zero to 128 characters of any kind
+	// ()? is optional
+	ncRegex := regexp.MustCompile(`NetworkManagement/interfaces/(.{0,128})/networkContainers/(.{0,128})/authenticationToken/(.{0,128})/api-version/1(/method/DELETE)?`)
```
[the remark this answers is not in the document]
aegal: Updated it to be 36 characters and updated the regex scope

cns/nmagent/client.go
aegal: Changed it to incorporate some of this feedback

cns/restserver/api.go
aegal: removed

cns/nmagent/client.go
```diff
@@ -65,13 +71,30 @@ func NewClient(url string) (*Client, error) {
...
+		Scheme: WireServerScheme,
+	}
+
+	queryString := joinNetworkURL.Query()
+	queryString.Set("type", joinNetworkTypeValue)
+	queryString.Set("comp", "nmagent")
+
+	joinNetworkURL.RawQuery = queryString.Encode()
+
+	response, err := common.GetHttpClient().Post(joinNetworkURL.String(), "application/json", &body)
```
ramiro-gamarra, line 97: since there are a few instances of this linting error, I'd suggest a small utility function:
```go
func PostCtx(ctx context.Context, cl *http.Client, url string, contentType string, body io.Reader) (*http.Response, error) {
	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, body)
	if err != nil {
		return nil, fmt.Errorf("could not create POST request: %w", err)
	}

	req.Header.Set("Content-Type", contentType)
	return cl.Do(req)
}
```
such that this line
```go
response, err := common.GetHttpClient().Post(joinNetworkURL.String(), "application/json", &body)
```
can be replaced with
```go
response, err := common.PostCtx(context.TODO(), common.GetHttpClient(), joinNetworkURL.String(), "application/json", &body)
```
which I think should address this.

cns/restserver/api.go
```diff
@@ -1119,14 +1145,35 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
...
+		err = service.Listener.Encode(w, &badRequestResponse)
+		logger.Response(service.Name, badRequestResponse, badRequestResponse.Response.ReturnCode, err)
+		return
+	}
+
 	switch r.Method {
 	case "POST":
 		// Join the network
-		publishResponse, publishError, err = service.joinNetwork(req.NetworkID, req.JoinNetworkURL)
+		publishResponse, publishError, err = service.joinNetwork(req.NetworkID)
```
ramiro-gamarra, line 1176: The existing code is __so__ bad that I am inclined to suggest just suppressing the linting error with
```
//nolint:bodyclose // please refactor this mess
```
since blindly appeasing the linter doesn't address real issues. Why is this code bad? Because
- `joinNetwork` has __already__ closed the request body, if there was one to close, even though it consumed __none__ of the body bytes. This causes inefficient TCP connection reuse, not sure what the effect is since it's agains the Wireserver host.
- The `publishResponse` var is reassigned below, with the result of a different http request. `defer`ing a body close earlier in the function __may__ have the equivalent effect while addressing the lint error, but it just really obfuscates that we are really handling 2 different `io.Readers`
- Control flow and scope of variables is puzzling.

cns/restserver/api.go
```diff
@@ -1233,12 +1274,31 @@ func (service *HTTPRestService) unpublishNetworkContainer(w http.ResponseWriter,
...
+		return
+	}
+
 	switch r.Method {
 	case "POST":
 		// Join Network if not joined already
 		isNetworkJoined = service.isNetworkJoined(req.NetworkID)
 		if !isNetworkJoined {
-			unpublishResponse, unpublishError, err = service.joinNetwork(req.NetworkID, req.JoinNetworkURL)
+			unpublishResponse, unpublishError, err = service.joinNetwork(req.NetworkID)
```
ramiro-gamarra, line 1301: same as above, just add a comment to ignore the linter

cns/restserver/api_test.go
```diff
@@ -509,31 +509,64 @@ func TestGetNetworkContainerVersionStatus(t *testing.T) {
 
 func createNC(
 	t *testing.T,
-	params createOrUpdateNetworkContainerParams) {
+	params createOrUpdateNetworkContainerParams,
```
ramiro-gamarra, line 512: I would also ignore this one with 
```go
//nolint:gocritic // param is just for testing
```

cns/restserver/api_test.go
```diff
@@ -546,59 +579,69 @@ func publishNCViaCNS(t *testing.T,
 	json.NewEncoder(&body).Encode(publishNCRequest)
 	req, err := http.NewRequest(http.MethodPost, cns.PublishNetworkContainer, &body)
 	if err != nil {
-		t.Fatal(err)
+		return err
```
ramiro-gamarra, line 582: this one should be simple enough to wrap

cns/restserver/api_test.go
```diff
@@ -610,19 +653,23 @@ func TestUnpublishNCViaCNS(t *testing.T) {
 	json.NewEncoder(&body).Encode(unpublishNCRequest)
 	req, err := http.NewRequest(http.MethodPost, cns.UnpublishNetworkContainer, &body)
 	if err != nil {
-		t.Fatal(err)
+		return err
```
ramiro-gamarra, line 656: same as above, wrap err

ramiro-gamarra (commented):
Some linting still needs to be addressed

cns/restserver/api.go
aegal: Cool, yeah I put a comment on there

rbtr (dismissed):
lgtm once the CI is fully green

rbtr (approved):

PR #1193, opened 2022-01-11
chore: Log CNI events to kusto for debugging

tamilmani1989:
**Reason for Change**:

This PR logs critical cni events to AI resource which would help for debugging as logs getting rotated in some cases.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -319,13 +319,16 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		cnsNetworkConfig *cns.GetNetworkContainerResponse
 		enableInfraVnet  bool
 		enableSnatForDns bool
+		k8sPodName       string
 		cniMetric        telemetry.AIMetric
 	)
 
 	startTime := time.Now()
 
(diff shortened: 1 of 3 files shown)


cni/network/network.go
```diff
@@ -517,16 +521,23 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		}()
 	}
 
+	msg = fmt.Sprintf("Allocated IPAddress from ipam:%+v v6:%+v", result, resultV6)
```
vakalapa, line 524: Instead of individual calls to metrics and logs, can you probably combine them into a single call ?

vakalapa: https://github.com/Azure/azure-container-networking/blob/da719d7a7b5fdf89cd841ecbb966c73bd3fc0ee1/npm/metrics/ai-utils.go#L51 

We did something similar in NPM,

vakalapa (approved):

PR #1194, opened 2022-01-12
fix initialize from kube pods flow

rbtr:
Signed-off-by: Evan Baker <rbtr@users.noreply.github>.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -46,7 +46,6 @@ import (
 	"github.com/Azure/azure-container-networking/store"
 	"github.com/avast/retry-go/v3"
 	"github.com/pkg/errors"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


cns/service/main.go
```diff
@@ -918,11 +910,42 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+
+	var podInfoByIPProvider cns.PodInfoByIPProvider
+	if cnsconfig.InitializeFromCNI {
+		logger.Printf("Initializing from CNI")
+		podInfoByIPProvider, err = cnireconciler.NewCNIPodInfoProvider()
+		if err != nil {
+			return errors.Wrap(err, "failed to create CNI PodInfoProvider")
+		}
+	} else {
+		logger.Printf("Initializing from Kubernetes")
```
neaggarwMS, line 926: Shouldnt we move this under retry loop to avoid crashloop backof not hitting and thus delaying cns startup in case apiserver is not ready

cns/service/main.go
```diff
@@ -918,11 +910,42 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 		}
 	}()
 
+	clientset, err := kubernetes.NewForConfig(kubeConfig)
```
neaggarwMS, line 913: I still suggest to move this logic inside initCNS. It is more contained there as part of CNS initialization workflow.

neaggarwMS: podInfoByProvider is kind of an internal state which we are passing to InitCNS func.

neaggarwMS (changes requested):

cns/service/main.go
rbtr: I have moved the list pods inside the closure so that, if the apiserver is not ready, we also retry that during the podinfobyip call

cns/service/main.go
rbtr: moving the clientset/the rest of this initialization into the `initCNS` breaks IOC - we should provide that func all of the objects it needs as inputs. passing it ready-to-use implementations of the interfaces it needs is the better pattern

cns/service/main.go
```diff
@@ -920,11 +911,40 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+	if cnsconfig.InitializeFromCNI {
+		logger.Printf("Initializing from CNI")
+		podInfoByIPProvider, err = cnireconciler.NewCNIPodInfoProvider()
+		if err != nil {
+			return errors.Wrap(err, "failed to create CNI PodInfoProvider")
+		}
+	} else {
+		logger.Printf("Initializing from Kubernetes")
+		podInfoByIPProvider = cns.PodInfoByIPProviderFunc(func() (map[string]cns.PodInfo, error) {
+			pods, err := clientset.CoreV1().Pods("").List(ctx, metav1.ListOptions{ //nolint:govet // ignore err shadow
```
matmerr, line 929: I think I'm okay with the contents of this pr, but I do want to mention that hostnetworking pods on windows do require a hacky kubeconfig to be constructed externally and loaded via file as opposed to using an inclusterconfig, if running in hostnw on windows becomes a thing

rbtr: good to know, we would need to update how we source the kubeconfigs in a couple of places for that

neaggarwMS (approved):

PR #1197, opened 2022-01-18
fix: [NPM] Windows Dataplane logic fix for src and dst ports

vakalapa:
**Reason for Change**:

According to the newly found info in VFP, the src and dst port mapping we used earlier are wrong. So correcting the logic as per below mapping:

// HNS has confusing Local and Remote address defintions
// For Traffic Direction INGRESS
// 		LocalAddresses  = Source Sets
// 		RemoteAddresses = Destination Sets
//      LocalPorts      = Destination Ports
//      RemotePorts     = Source Ports

// For Traffic Direction EGRESS
// 		LocalAddresses  = Source Sets
// 		RemoteAddresses = Destination Sets
//      LocalPorts      = Source Ports
//      RemotePorts     = Destination Ports

// If we use IPs in ACLs, then INGRESS mapping is same, but EGRESS mapping will change to below
// For Traffic Direction INGRESS
// 		LocalAddresses  = Source IPs
// 		RemoteAddresses = Destination IPs
// For Traffic Direction EGRESS
// 		LocalAddresses  = Destination IPs
// 		RemoteAddresses = Source IPs

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/pkg/dataplane/policies/policy_windows.go
+++ b/npm/pkg/dataplane/policies/policy_windows.go
@@ -94,14 +94,14 @@ func (acl *ACLPolicy) convertToAclSettings() (*NPMACLPolSettings, error) {
 
 	// HNS has confusing Local and Remote address defintions
 	// For Traffic Direction INGRESS
-	// 		LocalAddresses  = Source Sets
-	// 		RemoteAddresses = Destination Sets
+	// 	    LocalAddresses  = Source Sets
+	// 	    RemoteAddresses = Destination Sets
 	//      LocalPorts      = Destination Ports
 	//      RemotePorts     = Source Ports
 
 	// For Traffic Direction EGRESS
-	// 		LocalAddresses  = Source Sets
-	// 		RemoteAddresses = Destination Sets
+	// 	    LocalAddresses  = Source Sets
+	// 	    RemoteAddresses = Destination Sets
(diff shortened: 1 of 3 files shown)


nitishm (approved):
LGTM

npm/pkg/dataplane/policies/policy_windows.go
```diff
@@ -94,20 +94,34 @@ func (acl *ACLPolicy) convertToAclSettings() (*NPMACLPolSettings, error) {
...
+
+	// Switch ports based on direction
+	policySettings.RemotePorts = ""
+	policySettings.LocalPorts = dstPortStr
 	if policySettings.Direction == hcn.DirectionTypeOut {
-		policySettings.LocalAddresses = dstListStr
-		policySettings.LocalPorts = dstPortStr
-		policySettings.RemotePorts = ""
-		policySettings.RemoteAddresses = srcListStr
+		policySettings.LocalPorts = ""
```
huntergregory, line 123: should we do something for IP mode too?

vakalapa: No, we have de-prioritized IpMode. No need to do it.

huntergregory (approved):

PR #1198, opened 2022-01-19
Server-side filtering for NodeNetworkConfig, Node, and Pod objects

rbtr:

**Reason for Change**:

Instead of watching all NNCs (and Nodes and Pods) and filtering on the client-side, we should modify the Manager's cache to only Watch objects with the appropriate fields set which match our Node. This filtering happens server-side as part of the WatchList initialization and should effectively scope CNS's apiserver traffic to its own Node's objects instead of letting it scale with Node and Pod count.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -46,7 +46,6 @@ import (
 	"github.com/Azure/azure-container-networking/store"
 	"github.com/avast/retry-go/v3"
 	"github.com/pkg/errors"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


cns/service/main.go
```diff
@@ -932,10 +935,28 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+	// the nodeScopedCache sets Selector options on the Manager cache which are used
+	// to perform *server-side* filtering of the cached objects. This is very important
+	// for high node/pod count clusters, as it keeps us from watching objects at the
+	// whole cluster scope when we are only interested in the Node's scope.
+	nodeScopedCache := cache.BuilderWithOptions(cache.Options{
+		SelectorsByObject: cache.SelectorsByObject{
+			&v1alpha.NodeNetworkConfig{}: {
+				Field: fields.SelectorFromSet(fields.Set{"metadata.name": nodeName}),
+			},
+			&corev1.Pod{}: {
```
thatmattlong, line 947: Can we remove pods for now since we aren't using it? We have some tight rbac rules in place so adding this would require rbac change as well, which I don't think we're quite ready to make.

thatmattlong (changes requested):

cns/service/main.go
rbtr: dropped pods and nodes, maybe they will reappear at a later date :)

PR #1200, opened 2022-01-20
feat: [NPM] Forward Compatibility: delete v2 chains in v1, and Restructure Code

huntergregory:
Instead of deleting a fixed list of v1 and deprecated chains, delete all current azure iptables chains like we do in v2. 

Added benefit: removes unneeded iptables flush/delete calls both at bootup and when the last network policy is deleted.

### Formalities
- move "ioutil" package to a util package called "osutil_linux"
- add Linux build tags for each file in the package


npm/util/osutil_linux/current-azure-chains.go
```diff
@@ -0,0 +1,62 @@
...
+const (
+	azureChainGrepPattern   string = "Chain AZURE-NPM"
+	minAzureChainNameLength int    = len("AZURE-NPM")
+	// the minimum number of sections when "Chain NAME (1 references)" is split on spaces (" ")
+	minSpacedSectionsForChainLine int = 2
+)
+
+var errInvalidGrepResult = errors.New("unexpectedly got no lines while grepping for current Azure chains")
+
+func AllCurrentAzureChains(exec utilexec.Interface, defaultlockWaitTimeInSeconds string) (map[string]struct{}, error) {
```
vakalapa, line 25: Is there a particular reason for us to introduce this file ? I think this function being in PMGR makes sense, since all the iptable related ones are together, ifanything this can be part of chain-management

npm/iptm/iptm.go
```diff
@@ -127,13 +127,23 @@ func (iptMgr *IptablesManager) UninitNpmChains() error {
...
+		for _, chain := range allAzureChains {
+			delete(currentAzureChains, chain)
+		}
+		for chain := range currentAzureChains {
+			allAzureChains = append(allAzureChains, chain)
+		}
+	}
 
 	iptMgr.OperationFlag = util.IptablesFlushFlag
 	for _, chain := range allAzureChains {
```
vakalapa, line 150: Can we do a flush through restore ?

npm/iptm/iptm.go
```diff
@@ -127,13 +128,23 @@ func (iptMgr *IptablesManager) UninitNpmChains() error {
...
-	// TODO(jungukcho): need to check K8s or NPM version and do it selectively
-	// to avoid unnecessary call.
+	// Clean old NPM chains. This is forward compatible with NPM v2.
 	allAzureChains := append(IptablesAzureChainList,
 		util.IptablesAzureTargetSetsChain,
 		util.IptablesAzureIngressWrongDropsChain,
 	)
+	currentAzureChains, err := linuxutil.AllCurrentAzureChains(iptMgr.exec, defaultlockWaitTimeInSeconds)
+	if err != nil {
+		metrics.SendErrorLogAndMetric(util.IptmID, "Error: failed to get all current AZURE-NPM chains")
```
nitishm, line 138: Is this just a benign error? If so can we add to the error message that states this isn't bad?

npm/iptm/iptm.go
```diff
@@ -127,13 +128,23 @@ func (iptMgr *IptablesManager) UninitNpmChains() error {
...
+	} else {
+		// add any extra current azure chains to the list of all azure chains.
+		for _, chain := range allAzureChains {
+			delete(currentAzureChains, chain)
+		}
+		for chain := range currentAzureChains {
+			allAzureChains = append(allAzureChains, chain)
```
nitishm, line 145: We are adding indices here and not the element at that index. Is that what you intend to do?

nitishm, line 145: `for chain := range currentAzureChains` gives us the key and we are adding this key to `allAzureChains`. Is that what was intended?

npm/iptm/iptm_test.go
```diff
@@ -146,14 +117,90 @@ func TestInitNpmChains(t *testing.T) {
...
-
-	if err := iptMgr.UninitNpmChains(); err != nil {
-		t.Errorf("TestUninitNpmChains @ iptMgr.UninitNpmChains")
+	tests := []struct {
+		name  string
+		calls []testutils.TestCmd
+	}{
+		{
+			name: "no v2 npm chains exist",
+			calls: []testutils.TestCmd{
```
nitishm, line 126: Can we convert this to a helper function somewhere. It would be useful in other tests as well IMO.

npm/iptm/iptm.go
huntergregory: Technically benign for v1, so I can say "Warning: ..."

npm/iptm/iptm.go
huntergregory: yes, the map here is essentially a set as type `map[string]struct{}`

npm/util/osutil_linux/current-azure-chains.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/Azure/azure-container-networking/npm/util"
+	npmerrors "github.com/Azure/azure-container-networking/npm/util/errors"
+	"k8s.io/klog"
+	utilexec "k8s.io/utils/exec"
+)
+
+const (
+	azureChainGrepPattern   string = "Chain AZURE-NPM"
+	minAzureChainNameLength int    = len("AZURE-NPM")
```
nitishm, line 18: move `AZURE-NAME` into a const of it's own. You can then use it in `azureChainGrepPattern`. If we ever have to change the name in the future (there are discussions on renaming things IIRC) we can change it one place.

npm/util/osutil_linux/current-azure-chains_test.go
```diff
@@ -0,0 +1,142 @@
...
+const grepOutputAzureChainsWithPolicies = `Chain AZURE-NPM (1 references)
+Chain AZURE-NPM-ACCEPT (1 references)
+Chain AZURE-NPM-EGRESS (1 references)
+Chain AZURE-NPM-EGRESS-123456 (1 references)
+Chain AZURE-NPM-INGRESS (1 references)
+Chain AZURE-NPM-INGRESS-123456 (1 references)
+Chain AZURE-NPM-INGRESS-ALLOW-MARK (1 references)
+`
+
+var listAllCommandStrings = []string{"iptables", "-w", "60", "-t", "filter", "-n", "-L"}
```
nitishm, line 23: Seems like a useful command. You could export it. Maybe in some follow up PRs we move test helpers into their own pkg?

nitishm (commented):
Nit picking for the most part. But overall looks good to me

npm/util/osutil_linux/current-azure-chains.go
huntergregory: using a constant now

npm/util/osutil_linux/current-azure-chains_test.go
huntergregory: perhaps in another PR to clean up UTs? There are a couple areas of improvement

npm/iptm/iptm_test.go
huntergregory: perhaps in another PR to clean up UTs? There are a couple areas of improvement

vakalapa (approved):

nitishm (approved):
LGTM 🚀

PR #1202, opened 2022-01-21
fix: [NPM] fix incorrect ipset create, fix 1-off prometheus logic, and add cache checking for UTs

huntergregory:
Fixes a bug where a set would be created if it doesn't exist and Delete member API call is made on the set. 

Adds exec time metrics for applying ipsets and adding ACL rules.

Did the following UTs:
- apply sets
- create set
- delete set
- add to set

Need to eventually do:
- delete from set
- add to list
- delete from list
- add/delete selector


npm/metrics/prometheus-metrics.go
```diff
@@ -62,21 +62,27 @@ var (
...
-		numIPSetEntries = createGauge(numIPSetEntriesName, numIPSetEntriesHelp, false)
-		ipsetInventory = createGaugeVec(ipsetInventoryName, ipsetInventoryHelp, false, setNameLabel, setHashLabel)
-		log.Logf("Finished initializing all Prometheus metrics")
-		haveInitialized = true
+		ReinitializeAll()
 	}
 }
 
+// ReinitializeAll creates new Prometheus metrics.
+func ReinitializeAll() {
```
huntergregory, line 73: use a func to set haveInitialized instead

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -174,12 +174,23 @@ func (iMgr *IPSetManager) AddToSets(addToSets []*IPSetMetadata, ip, podKey strin
...
+	for _, metadata := range addToSets {
+		prefixedName := metadata.GetPrefixName()
+		set, exists := iMgr.setMap[prefixedName]
+		if !exists {
+			// NOTE: any newly created IPSet will still be in the cache if an error is returned later
+			iMgr.createIPSet(metadata)
+			set = iMgr.setMap[prefixedName]
+		}
+		if set.Kind != HashSet {
+			return npmerrors.Errorf(npmerrors.AppendIPSet, false, fmt.Sprintf("ipset %s is not a hash set", prefixedName))
```
vakalapa, line 187: Why the double loop here? we can put this in one giant loop ? At least some of the sets will have the right IPs if an error occurs.

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -199,13 +210,25 @@ func (iMgr *IPSetManager) RemoveFromSets(removeFromSets []*IPSetMetadata, ip, po
...
+			return npmerrors.Errorf(npmerrors.DeleteIPSet, false, fmt.Sprintf("ipset %s is not a hash set", prefixedName))
+		}
 	}
 
-	for _, setMetadata := range removeFromSets {
-		prefixedName := setMetadata.GetPrefixName()
-		set := iMgr.setMap[prefixedName]
+	// 2. remove ip from all existing sets
+	for _, metadata := range removeFromSets {
+		prefixedName := metadata.GetPrefixName()
```
vakalapa, line 227: Single loop will also eliminate the duplicate prefixed name and set checks

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -170,22 +170,36 @@ func (iMgr *IPSetManager) DeleteReference(setName, referenceName string, referen
...
-		prefixedName := setMetadata.GetPrefixName()
-		set := iMgr.setMap[prefixedName]
+		// 2. add ip to the set
 		cachedPodKey, ok := set.IPPodKey[ip]
 		set.IPPodKey[ip] = podKey
-		if ok && cachedPodKey != podKey {
-			klog.Infof("AddToSet: PodOwner has changed for Ip: %s, setName:%s, Old podKey: %s, new podKey: %s. Replace context with new PodOwner.",
-				ip, set.Name, cachedPodKey, podKey)
+		if ok {
+			if cachedPodKey != podKey {
```
vakalapa, line 197: Do we have to check the cachedPoKey here ? i think we can safely ignore this and not print any message ?

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -227,40 +250,100 @@ func (iMgr *IPSetManager) RemoveFromSets(removeFromSets []*IPSetMetadata, ip, po
...
+			// NOTE: any newly created IPSet will still be in the cache if an error is returned later
+			iMgr.createIPSet(listMetadata)
+			list = iMgr.setMap[listName]
+		}
+
+		if list.Kind != ListSet {
+			return npmerrors.Errorf(npmerrors.AppendIPSet, false, fmt.Sprintf("ipset %s is not a list set", listName))
+		}
+
+		if _, exists := memberNames[listName]; exists {
```
vakalapa, line 294: I do not think we need this check if listname == membername since, in the above loop we are checking if the set is HashSet and in this loop we are checking if it is ListSet, And our PrefixName depends on the "kind" right ? so we guarantee the name to be unique ? i am okay with keeping this check for safety reasons, Wdyt ?

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -227,40 +250,100 @@ func (iMgr *IPSetManager) RemoveFromSets(removeFromSets []*IPSetMetadata, ip, po
...
+			return npmerrors.Errorf(npmerrors.AppendIPSet, false, fmt.Sprintf("ipset %s cannot be added to itself", listName))
+		}
+
+		// 3. add all members to the list
+		for _, memberMetadata := range setMetadatas {
+			memberName := memberMetadata.GetPrefixName()
+			iMgr.addMemberIPSet(listName, memberName)
```
vakalapa, line 301: What is this setMetadatas are already part of the listmetadas ? Can we have the modified flag here too ad let addmemberIpset return a bool ?

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -227,40 +250,100 @@ func (iMgr *IPSetManager) RemoveFromSets(removeFromSets []*IPSetMetadata, ip, po
...
+		}
+
+		// Nested IPSets are only supported for windows
+		// Check if we want to actually use that support
+		if set.Kind != HashSet {
+			return npmerrors.Errorf(npmerrors.DeleteIPSet, false, fmt.Sprintf("ipset %s is not a hash set and nested list sets are not supported", memberName))
+		}
+
+		// 2. remove member from the list
+		iMgr.removeMemberIPSet(listName, memberName)
```
vakalapa, line 341: let removeMemberIpset return a bool and assign that to modified,

npm/pkg/dataplane/ipsets/ipsetmanager.go
huntergregory: makes sense. I'll replace with a comment

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -170,108 +171,209 @@ func (iMgr *IPSetManager) DeleteReference(setName, referenceName string, referen
...
-		if ok && cachedPodKey != podKey {
-			klog.Infof("AddToSet: PodOwner has changed for Ip: %s, setName:%s, Old podKey: %s, new podKey: %s. Replace context with new PodOwner.",
-				ip, set.Name, cachedPodKey, podKey)
+		if ok {
 			continue
 		}
 
-		iMgr.modifyCacheForKernelMemberUpdate(prefixedName)
+		iMgr.modifyCacheForKernelMemberUpdate(set)
 		metrics.AddEntryToIPSet(prefixedName)
```
vakalapa, line 202: We do not have to check this and print a info message, we can ignore mismatched podKey in add. Wdyt ?

vakalapa (approved):

vakalapa:
@huntergregory  V2 conformance is failing consistently.

huntergregory:
> @huntergregory V2 conformance is failing consistently.

investigating

vakalapa (approved):

PR #1203, opened 2022-01-21
feat: [NPM] Restructure code and add deploy manifests

nitishm:
- Moved files from top level npm/ folder to their own respective folders
  for controller and daemon components
- Generated new deployment manifests for each components and moved into
  the respective folders under npm/deploy

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


nitishm:
Checking this in for tracking. Will port to kustomize in follow up commits.

npm/deploy/npm/azure-npm.yaml
```diff
@@ -99,18 +99,12 @@ spec:
           volumeMounts:
           - name: xtables-lock
             mountPath: /run/xtables.lock
-          - name: log
```
vakalapa, line 102: We will need to update this new file location for all tests

nitishm: @matmerr mentioned this is obsolete? Do the tests still use this?

huntergregory: impacts these: 
- https://github.com/Azure/azure-container-networking/blob/9d8e61e7ae02ae572e16df34e3da67f2870bd041/.github/workflows/cyclonus-netpol-test.yaml#L53 
- https://github.com/Azure/azure-container-networking/blob/9d8e61e7ae02ae572e16df34e3da67f2870bd041/.pipelines/npm/npm-conformance-tests.yaml#L167

npm/deploy/npm/azure-npm.yaml
```diff
@@ -99,18 +99,12 @@ spec:
...
-            mountPath: /var/log
           - name: protocols
             mountPath: /etc/protocols
           - name: azure-npm-config
             mountPath: /etc/azure-npm
       hostNetwork: true
       volumes:
-      - name: log
```
huntergregory, line 110: wondering why we are removing /var/log?

npm/deploy/manifests/common/npm-configmap.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: v1
+kind: ConfigMap
```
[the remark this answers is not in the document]
huntergregory: similarly for npm-serviceaccount.yaml and rbac.yaml in the common/ folder

npm/deploy/kustomize/overlays/daemon/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: npm-deamon-metrics-cluster-service
```
huntergregory, line 4: per our Prometheus discussion yesterday, we'll need to serve up the same "cluster-metrics" service that we have currently 
https://github.com/Azure/azure-container-networking/blob/6a63ac05a757cdd70c996c149c14ac09a7841458/npm/azure-npm.yaml#L128

huntergregory (changes requested):
This is awesome. We should consider better names for services. We are stuck with "npm-metrics-cluster-service" for vanilla npm and npm daemon for Prometheus backwards compatibility. Perhaps we should reserve the term "cluster-metrics" for this service only

npm/deploy/kustomize/overlays/daemon/service.yaml
huntergregory: If we add daemon-specific metrics to npm-metrics-cluster-service, then we'll need to update a scrape config owned by Container Insights team

npm/deploy/kustomize/base/configmap.yaml
```diff
@@ -0,0 +1,24 @@
+---
+apiVersion: v1
+kind: ConfigMap
```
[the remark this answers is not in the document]
huntergregory: Let's move profiles/ to deploy/profiles/ to keep everything together. Can we move all configmap yamls to this folder? 

After moving the profiles/ folder, we'll need to update these (similar to the ones for moving azure-npm.yaml):
- https://github.com/Azure/azure-container-networking/blob/9d8e61e7ae02ae572e16df34e3da67f2870bd041/.github/workflows/cyclonus-netpol-test.yaml#L55
- https://github.com/Azure/azure-container-networking/blob/9d8e61e7ae02ae572e16df34e3da67f2870bd041/.pipelines/npm/npm-conformance-tests.yaml#L173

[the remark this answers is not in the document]
nitishm: Upstream the common practice is to add deployment manifests to the `deploy` folder. You keep your helm charts, kustomize, vanilla k8s yamls and any related manifests that get picked up by these resources. 
The `Configmap`s are common to both overlays hence they are in the `base` folder. If you still think `profiles` should go there I am happy to do that :)

npm/deploy/manifests/common/npm-configmap.yaml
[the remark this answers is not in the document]
nitishm: Common across all deployments (aka `overlays` in kustomize)

npm/deploy/npm/azure-npm.yaml
nitishm: @matmerr said we are using this in production anymore. I made an assumption it wasn't being used at all. I will add it back in :)

npm/deploy/kustomize/overlays/daemon/service.yaml
nitishm: This is a good point. Can we discuss this on chat?

npm/deploy/kustomize/base/configmap.yaml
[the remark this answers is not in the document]
huntergregory: makes sense. base/ seems good (as long as the current profiles are in the same folder, for testing sake)

npm/deploy/npm/azure-npm.yaml
huntergregory: Mat said we are or aren't?

huntergregory:
Discussed offline: we can use one service for the daemon (whether for vanilla NPM or the fan-out mode), which will have 3+ endpoints. The current ones will be for customers:
- /cluster-metrics
- /node-metrics

and new endpoints can be for internal use

npm/deploy/kustomize/base/configmap.yaml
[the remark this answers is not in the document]
vakalapa: Lets not increase the scope of this PR, for now, keep it simple with the filename changes and we can do a separate effort/PR to standardize this process , wdygt ?

npm/deploy/npm/azure-npm.yaml
vakalapa: We are NOT logging to a file anymore so we can safely remove this.

Tiltfile
```diff
@@ -0,0 +1,10 @@
+allow_k8s_contexts('acn-dev-azure-cni')
+default_registry('ttl.sh/nitishm-12390')
```
matmerr, line 2: if this is an ephemeral image can we put it on acnpublic instead?

npm/controller/server.go
```diff
@@ -17,6 +18,8 @@ import (
 	"k8s.io/klog"
 )
 
+var aiMetadata string //nolint // aiMetadata is set in Makefile
```
matmerr, line 21: will require further verification of how this is used, but this is set via the Makefile on build in the dockerfile. Not sure of implications with introducing it into subpackages, and/or if aiMetadata is even relevant in this scope if used in the parent npm/

npm/config/config.go
```diff
@@ -16,8 +17,9 @@ var DefaultConfig = Config{
 	ListeningAddress: "0.0.0.0",
 
 	Transport: GrpcServerConfig{
-		Address: "0.0.0.0",
-		Port:    defaultGrpcPort,
+		Address:    "0.0.0.0",
+		Port:       defaultGrpcPort,
```
matmerr, line 21: maybe `ServerPort` to be extra explicit alongside the `RemotePort`?

npm/deploy/manifests/common/npm-configmap.yaml
```diff
@@ -0,0 +1,24 @@
...
+        "ListeningPort":         10091,
+        "ListeningAddress":      "0.0.0.0",
+        "Toggles": {
+            "EnablePrometheusMetrics": true,
+            "EnablePprof":             true,
+            "EnableHTTPDebugAPI":      true,
+            "EnableV2NPM":             false,
+            "PlaceAzureChainFirst":    false
+        },
+        "Transport": {
```
matmerr, line 20: can we add this to examples/windows/* manifests too

Tiltfile
nitishm: I couldnt find how to provide the credentials (plus we don't want the credentials checked in). Need to look into this

npm/config/config.go
nitishm: ServicePort maybe?

vakalapa (approved):

huntergregory (approved):
Will need to revisit service names when we finalize the fan-out model

PR #1206, opened 2022-01-24
feat: [NPM] Adding DPShim layer in controller pods

vakalapa:
**Reason for Change**:

Adding DPShim component, 

DPShim works as a translator, cache and transporter of Resource events into sets/policies and relays them to gRPC server. The gRPC server will distribute these goal states to all the registered daemon clients. 

DPShim:

1) Cache locally all the sets and policies
2) On Apply Dataplane, convert dirty cache to relevant GoalState events and send to gRPC server
3) periodically delete unused sets
4) Provide a function for clients on first time registration to get the state of whole cache

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


npm/pkg/controlplane/types.go
```diff
@@ -21,16 +23,61 @@ type ControllerIPSets struct {
...
+		NetPolReference: make(map[string]struct{}),
 	}
 }
 
+// GetMetadata returns the metadata of the ipset
 func (c *ControllerIPSets) GetMetadata() *ipsets.IPSetMetadata {
 	return c.IPSetMetadata
 }
+
+// DecIPSetReferCount decrements the ipset refer count
```
nitishm, line 50: comment doesnt match the method name

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+	return &protos.Events{
+		Payload: goalStates,
+	}, nil
+}
+
+func (dp *DPShim) RunPeriodicTasks() {
+	// Here Run periodic task to check if any sets with empty references are present and delete them
+}
 
 func (dp *DPShim) GetIPSet(setName string) *ipsets.IPSet {
```
nitishm, line 90: is this no-op? If so can we add a comment explaining why?

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
 	return nil
 }
 
+func (dp *DPShim) getIPSet(setName string) *controlplane.ControllerIPSets {
+	return dp.setCache[setName]
```
nitishm, line 95: Should this be `getCache`?

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+}
+
+func (dp *DPShim) setExists(setName string) bool {
+	_, ok := dp.setCache[setName]
```
nitishm, line 99: same here? getCache??

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+		return
+	}
+
+	dp.setCache[setName] = controlplane.NewControllerIPSets(set)
+	dp.dirtyCache.modifyAddorUpdateSets(setName)
 }
 
 func (dp *DPShim) DeleteIPSet(setMetadata *ipsets.IPSetMetadata) {
+	dp.Lock()
+	defer dp.Unlock()
```
nitishm, line 124: might make more sense to move the lock/unlock to the helper functions `deleteIPSet`

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+	}
+
+	delete(dp.setCache, setName)
+	dp.dirtyCache.modifyDeleteSets(setName)
 }
 
-func (dp *DPShim) RemoveFromSets(setNames []*ipsets.IPSetMetadata, podMetadata *dataplane.PodMetadata) error {
+func (dp *DPShim) AddToSets(setMetadatas []*ipsets.IPSetMetadata, podMetadata *dataplane.PodMetadata) error {
+	dp.Lock()
+	defer dp.Unlock()
```
nitishm, line 146: If we have this in `createIPSet` we wouldn't have to lock on a large code block

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+	toApplyPolicies := make([]*policies.NPMNetworkPolicy, len(dp.dirtyCache.toAddorUpdatePolicies))
+	idx := 0
+
+	for policyKey := range dp.dirtyCache.toAddorUpdatePolicies {
+		if !dp.policyExists(policyKey) {
+			return nil, npmerrors.Errorf(npmerrors.AddPolicy, false, fmt.Sprintf("policy %s not found", policyKey))
+		}
+
+		policy := dp.policyCache[policyKey]
+		toApplyPolicies[idx] = policy
```
nitishm, line 477: dont need to use ID - we can just do this,
```go
toApplyPolicies = append(toApplyPolicies, policy)
```

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+
+	toApplyPolicies := make([]*policies.NPMNetworkPolicy, len(dp.dirtyCache.toAddorUpdatePolicies))
+	idx := 0
+
+	for policyKey := range dp.dirtyCache.toAddorUpdatePolicies {
+		if !dp.policyExists(policyKey) {
+			return nil, npmerrors.Errorf(npmerrors.AddPolicy, false, fmt.Sprintf("policy %s not found", policyKey))
+		}
+
+		policy := dp.policyCache[policyKey]
```
nitishm, line 476: What if the policy doesnt exist?

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+func (dp *DPShim) processPoliciesRemove() (*protos.GoalState, error) {
+	if len(dp.dirtyCache.toDeletePolicies) == 0 {
+		return nil, nil
+	}
+
+	toDeletePolicies := make([]string, len(dp.dirtyCache.toDeletePolicies))
+	idx := 0
+
+	for policyKey := range dp.dirtyCache.toDeletePolicies {
+		toDeletePolicies[idx] = policyKey
```
nitishm, line 499: Same - make use of `append`

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+		for {
+			select {
+			case <-stopChannel:
+				return
+			case <-ticker.C:
+				klog.Info("deleteUnusedSets: cleaning up unused sets")
+				dp.checkSetReferences()
+				err := dp.ApplyDataPlane()
+				if err != nil {
+					klog.Errorf("deleteUnusedSets: failed to apply dataplane %v", err)
```
nitishm, line 569: I've been curious about error propagation in go routines. Should we create an error channel to feedback errors and handle them in the caller code?

nitishm: This is across our codebase I mean

nitishm (changes requested):
Functionality looks good - just some nit picking

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+func (dp *DPShim) getIPSet(setName string) *controlplane.ControllerIPSets {
+	return dp.setCache[setName]
+}
+
+func (dp *DPShim) setExists(setName string) bool {
+	_, ok := dp.setCache[setName]
+	return ok
+}
+
 func (dp *DPShim) CreateIPSets(setNames []*ipsets.IPSetMetadata) {
```
huntergregory, line 103: rename params to setMetadatas in line with types.go

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+func (dp *DPShim) deleteIPSet(setMetadata *ipsets.IPSetMetadata) {
+	setName := setMetadata.GetPrefixName()
+	klog.Infof("deleteIPSet: cleaning up %s", setName)
+	set, ok := dp.setCache[setName]
+	if !ok {
+		return
+	}
+
+	if set.HasReferences() {
+		return
```
huntergregory, line 137: should we log errors here? Is this expected behavior in any situation?

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
 	return nil
 }
 
-func (dp *DPShim) AddToLists(listName, setNames []*ipsets.IPSetMetadata) error {
+func (dp *DPShim) RemoveFromSets(setMetadatas []*ipsets.IPSetMetadata, podMetadata *dataplane.PodMetadata) error {
+	dp.Lock()
+	defer dp.Unlock()
+	for _, set := range setMetadatas {
+		prefixedSetName := set.GetPrefixName()
+		if !dp.setExists(prefixedSetName) {
```
huntergregory, line 177: remove this (bug from ipsetmanager)

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+
+func (dp *DPShim) policyExists(policyKey string) bool {
+	_, ok := dp.policyCache[policyKey]
+	return ok
+}
+
+func (dp *DPShim) checkForListMemberUpdateErrors(listMetadata, memberMetadatas []*ipsets.IPSetMetadata, npmErrorString string) error {
+	for _, listMetadata := range listMetadata {
+		prefixedListName := listMetadata.GetPrefixName()
+		if !dp.setExists(prefixedListName) {
```
huntergregory, line 387: bug from ipsetmanager

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+	defer func() {
+		dperr := dp.ApplyDataPlane()
+		if dperr != nil {
+			err = fmt.Errorf("failed with error %w, apply failed with %v", err, dperr)
+		}
+	}()
+	dp.Lock()
+	defer dp.Unlock()
+
+	if dp.policyExists(networkpolicies.PolicyKey) {
```
huntergregory, line 270: For another PR: what do we think of normalizing/validating policies before they're sent to dp? See these functions: https://github.com/Azure/azure-container-networking/blob/35dcb32b18dff9683e79007729713892acb21781/npm/pkg/dataplane/policies/policymanager.go#L146

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+	dp.Lock()
+	defer dp.Unlock()
+	for _, set := range setMetadatas {
+		prefixedSetName := set.GetPrefixName()
+		if !dp.setExists(prefixedSetName) {
+			dp.createIPSet(set)
+		}
+
+		set := dp.setCache[prefixedSetName]
+		if set.IPSetMetadata.GetSetKind() != ipsets.HashSet {
```
huntergregory, line 182: we should generalize addToSet, deleteFromSet, addToList, ... validations into a dp func for both dpshim and Dataplane

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+	toDeletePolicies, err := dp.processPoliciesRemove()
+	if err != nil {
+		return err
+	}
+	if toDeletePolicies != nil {
+		goalStates[controlplane.PolicyRemove] = toDeletePolicies
+	}
+
+	if len(goalStates) == 0 {
+		klog.Info("ApplyDataPlane: No changes to apply")
```
huntergregory, line 365: could this ever happen?

npm/pkg/dataplane/dpshim/dpshim.go
huntergregory: if you use append, you should set the initial capacity of the slice to avoid resizing: `make([]thing, 0, len(cache))`

huntergregory (changes requested):
dpshim looks pretty good to me. Made several suggestions. This PR seems dependent on #1202 for validation logic

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -1,77 +1,590 @@
...
+		}
+
+		set := dp.setCache[prefixedSetName]
+		if set.IPSetMetadata.GetSetKind() != ipsets.HashSet {
+			return npmerrors.Errorf(npmerrors.AppendIPSet, false, fmt.Sprintf("ipset %s is not a hash set", prefixedSetName))
+		}
+
+		cachedPod, ok := set.IPPodMetadata[podMetadata.PodIP]
+		set.IPPodMetadata[podMetadata.PodIP] = podMetadata
+		if ok && cachedPod.PodKey != podMetadata.PodKey {
```
vakalapa, line 160: Need to cover if ok condition.

npm/pkg/dataplane/dpshim/dpshim.go
vakalapa: I am following the practice of Exposed function deal with locks and internal functions to not worry about locks. Internal functions by default we can assume that they are to be handled with care. Also, createIPSet is called from some external functions and we cannot deal with locking and unlocking always. Wdyt ?

npm/pkg/dataplane/dpshim/dpshim.go
vakalapa: Can log error.

npm/pkg/dataplane/dpshim/dpshim.go
vakalapa: except for the exists check everything needs to be locked anyway, so i believe trying to lock for each loop might add more overhead. wdyt ?

npm/pkg/dataplane/dpshim/dpshim.go
vakalapa: Thats the check we are doing one line above ?

npm/pkg/dataplane/dpshim/dpshim.go
vakalapa: Avoiding resizing since we know the size of data.

npm/pkg/dataplane/ipsets/ipsetmanager_test.go
```diff
@@ -375,3 +375,49 @@ func TestMain(m *testing.M) {
 
 	os.Exit(exitCode)
 }
+
+func TestValidateIPBlock(t *testing.T) {
```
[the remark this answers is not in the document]
huntergregory: disregard. We will move this validation to the controller

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -258,8 +317,8 @@ func (dp *DPShim) AddPolicy(networkpolicies *policies.NPMNetworkPolicy) error {
 			err = fmt.Errorf("failed with error %w, apply failed with %v", err, dperr)
 		}
 	}()
-	dp.Lock()
-	defer dp.Unlock()
+	dp.lock()
```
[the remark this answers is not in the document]
huntergregory: I see that ApplyDataplane locks too. Maybe we make a little applyDataplane that doesn't lock?

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -143,8 +144,13 @@ func (dp *DPShim) deleteIPSet(setMetadata *ipsets.IPSetMetadata) {
...
 func (dp *DPShim) AddToSets(setMetadatas []*ipsets.IPSetMetadata, podMetadata *dataplane.PodMetadata) error {
-	dp.Lock()
-	defer dp.Unlock()
+	if len(setMetadatas) == 0 {
+		return nil
+	}
+
+	dp.lock()
+	defer dp.unlock()
+
```
huntergregory, line 153: needs to be updated to be like ipsetmanager.go
```
		_, ok := set.IPPodKey[ip]
		set.IPPodKey[ip] = podKey
		if ok {
			continue
		}
```

huntergregory: or perhaps this, because it needs to share when the podkey has changed
```
		oldPodKey, ok := set.IPPodKey[ip]
		if ok && oldPodKey == podKey {
			continue
		}
		set.IPPodKey[ip] = podKey
```

npm/pkg/dataplane/dpshim/dpshim.go
huntergregory: `continue` instead of creating the set

npm/http/server/server.go
```diff
@@ -32,7 +32,7 @@ func NPMRestServerListenAndServe(config npmconfig.Config, npmEncoder json.Marsha
 		rs.router.Handle(api.ClusterMetricsPath, metrics.GetHandler(metrics.ClusterMetrics))
 	}
 
-	if config.Toggles.EnableHTTPDebugAPI {
+	if config.Toggles.EnableHTTPDebugAPI && npmEncoder != nil {
```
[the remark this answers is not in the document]
vakalapa: This is not relevant in V1 case, this is specifically for daemon where we will need the PProf endpoint but epmencoder is not yet supported. That is a todo item for later.

npm/pkg/dataplane/ipsets/ipsetmanager_test.go
```diff
@@ -68,6 +68,153 @@ var (
...
+
+	assertExpectedInfo(t, iMgr, &expectedInfo{
+		mainCache:        nil,
+		toAddUpdateCache: nil,
+		toDeleteCache:    nil,
+		setsForKernel:    nil,
+	})
+}
+
+func TestApplyIPSetsMain(t *testing.T) {
```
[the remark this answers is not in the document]
vakalapa: Residual from master merge, will fix it.

npm/pkg/dataplane/dpshim/dpshim.go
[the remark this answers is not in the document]
vakalapa: This is intentional and keeping the same design from our existing DP, we do not anticipate another process coming in between and making changes.

npm/pkg/dataplane/dpshim/dpshim.go
```diff
@@ -285,8 +344,8 @@ func (dp *DPShim) RemovePolicy(policyKey string) error {
 		}
 	}()
 
-	dp.Lock()
-	defer dp.Unlock()
+	dp.lock()
```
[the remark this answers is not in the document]
vakalapa: See above reply.

npm/pkg/dataplane/dpshim/dpshim.go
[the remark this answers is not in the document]
vakalapa: Even if it does i do not think there is a major issue, except for one, i.e. if a Set is added/created and immediately deleted then kernel might complain about it does not exists. But i think you had suggested a fix for it to use --exists, is that in master ?

npm/pkg/dataplane/dpshim/dpshim.go
vakalapa: getIPSet is in line with our interface, so prefer to keep it this. Unless you have strong reservations ?

npm/pkg/controlplane/goalstateprocessor/goalstateprocessor.go
```diff
@@ -324,9 +336,7 @@ func (gsp *GoalStateProcessor) processPolicyRemoveEvent(goalState *protos.GoalSt
 
 func validatePayload(payload map[string]*protos.GoalState) bool {
 	for _, v := range payload {
-		if len(v.GetData()) != 0 {
-			return true
-		}
+		return len(v.GetData()) != 0
```
huntergregory, line 339: this will only reach the first item in the slice

npm/pkg/dataplane/dpshim/dpshim.go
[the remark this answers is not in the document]
huntergregory: will it add system overhead to lock/unlock twice for each Policy call though? We haven't solved the created then deleted edge case yet (in the non-fan-out model, we thought that situation shouldn't happen)

npm/pkg/controlplane/controllers/v1/networkPolicyController.go
```diff
@@ -66,7 +66,7 @@ func NewNetworkPolicyController(npInformer networkinginformers.NetworkPolicyInfo
 
 // initializeDataPlane do all initialization tasks for data plane
 // TODO(jungukcho) Need to refactor UninitNpmChains since it assumes it has already AZURE-NPM chains
-func (c *NetworkPolicyController) ResetDataPlane() error {
+func (c *NetworkPolicyController) BootupDataplane() error {
```
nitishm, line 69: Can you add the name to the comment for the docs

npm/pkg/controlplane/goalstateprocessor/goalstateprocessor.go
```diff
@@ -203,9 +207,16 @@ func (gsp *GoalStateProcessor) applySets(ipSet *cp.ControllerIPSets, cachedIPSet
 	}
 
 	if cachedIPSet != nil {
-		for podIP, podKey := range cachedIPSet.IPPodKey {
-			if _, ok := ipSet.IPPodMetadata[podIP]; !ok {
-				err := gsp.dp.RemoveFromSets([]*ipsets.IPSetMetadata{setMetadata}, dataplane.NewPodMetadata(podIP, podKey, ""))
+		for podIP, cachedPodKey := range cachedIPSet.IPPodKey {
+			if podMetadata, ok := ipSet.IPPodMetadata[podIP]; !ok {
+				if cachedPodKey != podMetadata.PodKey {
```
nitishm, line 212: a `!ok` means the entry is not found in which case the `podMetadata` is an empty object. Something doesn't seem right here. 🤔

nitishm (changes requested):
Logic seems wrong in https://github.com/Azure/azure-container-networking/pull/1206/commits/b7ffd3545e9f63e4a290ef3073bf729c8252c26b#r806133602

npm/pkg/controlplane/goalstateprocessor/goalstateprocessor.go
vakalapa: That is intentional, payload is invalid if all items are at 0, else it is valid.

npm/pkg/controlplane/goalstateprocessor/goalstateprocessor.go
vakalapa: fixing that

nitishm (approved):
LGTM

huntergregory (approved):
🚀

PR #1216, opened 2022-01-31
feat: don't use CNS for CNI DEL command in windows multitenancy

thatmattlong:
**Reason for Change**:

Today in Azure CNI in windows multitenancy, we depend on the NC state to exist in CNS in order for host endpoint cleanup to work. But, if the NC state got deleted from CNS before the CNI DEL command was invoked, the CNI DEL will fail since the CNS state doesn't exist. This PR allows the CNI DEL command to use only the CNI state instead of relying on CNS.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -969,7 +969,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 	if err != nil {
 		log.Printf("[cni-net] Failed to extract network name from network config. error: %v", err)
 
-		if !cnscli.IsNotFound(err) {
+		if !cnscli.IsNotFound(err) || network.IsNetworkNotFoundError(err) {
 			err = plugin.Errorf("Failed to extract network name from network config. error: %v", err)
 			return err
 		}
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -453,7 +453,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 	}
 
 	// Initialize values from network config.
-	networkID, err := plugin.getNetworkName(k8sPodName, k8sNamespace, args.IfName, nwCfg)
+	networkID, err := plugin.getNetworkName(k8sPodName, k8sNamespace, args.IfName, args.Netns, nwCfg)
 	if err != nil {
 		log.Printf("[cni-net] Failed to extract network name from network config. error: %v", err)
 		return err
@@ -843,7 +843,7 @@ func (plugin *NetPlugin) Get(args *cniSkel.CmdArgs) error {
 	}
 
 	// Initialize values from network config.
-	if networkId, err = plugin.getNetworkName(k8sPodName, k8sNamespace, args.IfName, nwCfg); err != nil {
+	if networkId, err = plugin.getNetworkName(k8sPodName, k8sNamespace, args.IfName, args.Netns, nwCfg); err != nil {
 		// TODO: Ideally we should return from here only.
 		log.Printf("[cni-net] Failed to extract network name from network config. error: %v", err)
 	}
@@ -963,7 +963,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 		}
 	}
 	// Initialize values from network config.
-	networkID, err = plugin.getNetworkName(k8sPodName, k8sNamespace, args.IfName, nwCfg)
+	networkID, err = plugin.getNetworkName(k8sPodName, k8sNamespace, args.IfName, args.Netns, nwCfg)
 
 	// If error is not found error, then we ignore it, to comply with CNI SPEC.
 	if err != nil {
@@ -1024,8 +1024,12 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 	telemetry.LogAndSendEvent(plugin.tb, fmt.Sprintf("Deleting endpoint:%v", endpointID))
 	// Delete the endpoint.
 	if err = plugin.nm.DeleteEndpoint(cnsclient, networkID, endpointID); err != nil {
-		err = plugin.Errorf("Failed to delete endpoint: %v", err)
-		return err
+		pluginErr := plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
+		// If it is not a networkNotFound error, then return a retriable error so the container runtime will keep retrying.
+		// If it is a networkNotFoundError, then simply skip it and move on to cleanup of InfraVnet.
+		if !network.IsNetworkNotFoundError(err) {
+			return pluginErr
+		}
 	}
 
 	if !nwCfg.MultiTenancy {
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -1013,17 +1013,11 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 		return err
 	}
 
-	cnsclient, err := cnscli.New(nwCfg.CNSUrl, defaultRequestTimeout)
-	if err != nil {
-		log.Printf("failed to initialized cns client with URL %s: %v", nwCfg.CNSUrl, err.Error())
-		return plugin.Errorf(err.Error())
-	}
-
 	// schedule send metric before attempting delete
 	defer sendMetricFunc()
 	telemetry.LogAndSendEvent(plugin.tb, fmt.Sprintf("Deleting endpoint:%v", endpointID))
 	// Delete the endpoint.
-	if err = plugin.nm.DeleteEndpoint(cnsclient, networkID, endpointID); err != nil {
+	if err = plugin.nm.DeleteEndpoint(networkID, endpointID); err != nil {
 		pluginErr := plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
 		// If it is not a networkNotFound error, then return a retriable error so the container runtime will keep retrying.
 		// If it is a networkNotFoundError, then simply skip it and move on to cleanup of InfraVnet.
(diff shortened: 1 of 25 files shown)


cni/network/network.go
```diff
@@ -1024,8 +1024,12 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 	telemetry.LogAndSendEvent(plugin.tb, fmt.Sprintf("Deleting endpoint:%v", endpointID))
 	// Delete the endpoint.
 	if err = plugin.nm.DeleteEndpoint(cnsclient, networkID, endpointID); err != nil {
-		err = plugin.Errorf("Failed to delete endpoint: %v", err)
-		return err
+		pluginErr := plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
+		// If it is not a networkNotFound error, then return a retriable error so the container runtime will keep retrying.
+		// If it is a networkNotFoundError, then simply skip it and move on to cleanup of InfraVnet.
+		if !network.IsNetworkNotFoundError(err) {
```
tamilmani1989, line 1030: isn't it endpoint not found?

cni/network/network_windows.go
```diff
@@ -177,6 +177,15 @@ func (plugin *NetPlugin) getNetworkName(podName, podNs, ifName string, nwCfg *cn
...
+		// First try to get the network name from the state file
+		if networkName, err := plugin.nm.FindNetworkNameFromNetNs(netNs); err != nil {
+			log.Printf("Error getting network name from state: %v. Fall back to CNS.", err)
+		} else {
+			return networkName, nil
+		}
+
+		// If it is not found, then fallback to CNS, in case it wasn't in the statefile for some reason
+		// TODO: investigate if we can remove this fallback
 		_, cnsNetworkConfig, _, err = plugin.multitenancyClient.GetContainerNetworkConfiguration(context.TODO(), nwCfg, podName, podNs, ifName)
```
tamilmani1989, line 189: i would say remove it. even if it fails, we know whether to fix this or not

network/endpoint_windows.go
```diff
@@ -281,14 +282,28 @@ func (nw *network) configureHcnEndpoint(epInfo *EndpointInfo) (*hcn.HostComputeE
...
-	log.Printf("[net] Deleting HostNCApipaEndpoint for network container: %s", networkContainerID)
-	err := cli.DeleteHostNCApipaEndpoint(context.TODO(), networkContainerID)
-	log.Printf("[net] Completed HostNCApipaEndpoint deletion for network container: %s"+
-		" with error: %v", networkContainerID, err)
+	// First try to delete the host NC apipa endpoint directly from CNI
+	log.Printf("[net] Deleting HosNCApipaEndpoint for network container [%s] directly")
+	if err := nw.deleteHostNCApipaEndpointDirect(networkContainerID); err != nil {
+		log.Printf("[net] Error deleting HostNCApipaEndpoint for network container [%s] directly, fallback to CNS")
+		// Fallback to deleting the host NC apipa endpoint with CNS
+		log.Printf("[net] Deleting HostNCApipaEndpoint for network container [%s] with CNS", networkContainerID)
```
tamilmani1989, line 290: why fallback to cns? if it fails, then cns also does the same right?

network/endpoint_windows.go
```diff
@@ -281,14 +282,28 @@ func (nw *network) configureHcnEndpoint(epInfo *EndpointInfo) (*hcn.HostComputeE
...
+			log.Printf("[net] Completed HostNCApipaEndpoint deletion for network container: %s with error: %v", networkContainerID, err)
+			return err
+		}
+	}
 
+	log.Printf("[net] Completed HostNCApipaEndpoint deletion for network container [%s] successfully", networkContainerID)
 	return nil
 }
 
+func (nw *network) deleteHostNCApipaEndpointDirect(networkContainerID string) error {
```
tamilmani1989, line 301: shouldn't this be endpoint function? func (ep *endpoint) deleteHostNCApipaEndpointDirect(networkContainerID string) error {

network/endpoint_windows.go
```diff
@@ -281,14 +282,28 @@ func (nw *network) configureHcnEndpoint(epInfo *EndpointInfo) (*hcn.HostComputeE
 }
 
 func (nw *network) deleteHostNCApipaEndpoint(cli apipaClient, networkContainerID string) error {
```
tamilmani1989, line 284: should be endpoint function

network/network.go
```diff
@@ -244,3 +244,25 @@ func (nm *networkManager) getNetwork(networkId string) (*network, error) {
 
 	return nil, errNetworkNotFound
 }
+
+// getNetworkNameForNetNs finds the network that contains the endpoint that was created for this netNs. Returns
+// and errNetworkNotFound if the netNs is not found in any network
+func (nm *networkManager) FindNetworkNameFromNetNs(netNs string) (string, error) {
```
tamilmani1989, line 250: FindNetworkIDFromNetNs?

tamilmani1989 (commented):
add unit test for finding network from netns

network/network.go
aegal: Same, also thinking we should put a not on what the param is

cni/network/network.go
thatmattlong: we already skip endpoint not found by returning a nil error if the endpoint is not found, but before that we call `getNetwork` which could return `errNetworkNotFound`. If that happens, we also want to skip it since that means there is no endpoint if there is no network. Thus we can move on to the infra vnet cleanup. If we get any other error, we want to retry later.

network/endpoint_windows.go
thatmattlong: the create and delete HostNCApipaEndpoint functions were already as `(nw *network)` so I don't want to refactor too much in this PR, I would prefer to leave this change for later

network/endpoint_windows.go
thatmattlong: removed

network/endpoint_windows.go
thatmattlong: removed this func

thatmattlong:
> add unit test for finding network from netns

Done

cni/network/network_windows.go
thatmattlong: actually I think we need this, during CNI ADD it will call this function too and network won't be in state file yet since it hasn't been created, so in that case we need to query CNS

cni/network/network.go
tamilmani1989: if network not found, we return nil too above (line no: 979)..probably we dont need to check this at all IsNetworkNotFoundError

cni/network/network_windows.go
tamilmani1989: hmm right.. we are calling CNS twice in ADD case then.. we should remove this code and rely on GetMultiTenancyCNIResult which returns cnsNetworkConfig too and we have to construct from that

network/endpoint_windows.go
tamilmani1989: hmm right..should ve been endpoint function..that' fine..let it be

tamilmani1989 (commented):
left comments

cni/network/network.go
tamilmani1989: one more thing to verify is that are we handling if hns delete endpoint call returns endpoint not found error.. in that case we shouldn't return error

thatmattlong: we first check if endpoint is found before hns call so we shouldn't end up in that scenario, and even so if we do, then the next time the cni is invoked we would hit that check and it would succeed

cni/network/network_windows.go
thatmattlong: done

cni/network/network_windows.go
```diff
@@ -160,42 +159,43 @@ func updateSubnetPrefix(cnsNwConfig *cns.GetNetworkContainerResponse, subnetPref
...
+
+		// networkName will look like ~ azure-vlan1-172-28-1-0_24
+		networkSuffix := strings.Replace(subnet.String(), ".", "-", -1)
+		networkSuffix = strings.Replace(networkSuffix, "/", "_", -1)
+		networkName := fmt.Sprintf("%s-vlan%v-%v", nwCfg.Name, cnsResponse.MultiTenancyInfo.ID, networkSuffix)
+
+		return networkName, nil
+	}
+
+	// If no cnsResponse was present, try to get the network name from the state file
```
tamilmani1989, line 190: add a comment probably saying this code will hit only on delete call

cni/network/network_windows.go
```diff
@@ -160,42 +159,43 @@ func updateSubnetPrefix(cnsNwConfig *cns.GetNetworkContainerResponse, subnetPref
...
-				networkName = strings.Replace(networkName, "/", "_", -1)
-				networkName = fmt.Sprintf("%s-vlan%v-%v", nwCfg.Name, cnsNetworkConfig.MultiTenancyInfo.ID, networkName)
-			}
+	// in multitenancy case, the network name will be in the state file or can be built from cnsResponse
+	determineWinVer()
+	if len(strings.TrimSpace(netNs)) == 0 || len(strings.TrimSpace(podName)) == 0 || len(strings.TrimSpace(podNs)) == 0 {
+		return "", fmt.Errorf("POD info cannot be empty. PodName: %s, PodNamespace: %s, NetNs: %s", podName, podNs, netNs)
+	}
+
+	// First try to build the network name from the cnsResponse if present
```
tamilmani1989, line 174: add a comment - it will hit only for add not delete

network/endpoint_windows.go
```diff
@@ -280,11 +284,47 @@ func (nw *network) configureHcnEndpoint(epInfo *EndpointInfo) (*hcn.HostComputeE
...
+
+	log.Printf("[net] Completed HostNCApipaEndpoint deletion for network container [%s] successfully", networkContainerID)
+	return nil
+}
+
+func getHostNCApipaEndpointName(networkContainerID string) string {
+	return fmt.Sprintf("%s-%s", hostNCApipaEndpointNamePrefix, networkContainerID)
+}
+
+func deleteEndpointByNameHnsV2(endpointName string) error {
```
tamilmani1989, line 309: rename and remove repetitive logs

tamilmani1989 (approved):
lgtm 🚢

PR #1219, opened 2022-02-01
feat: [NPM] add dataplane health metrics

huntergregory:
Introduce dataplane health metrics, and pave the way to add other metrics endpoints easily.

TODO: Consider how to initialize only the necessary metrics/endpoints.


huntergregory:
moving to #1220

PR #1220, opened 2022-02-02
feat: [NPM] perf metrics for pod/ns/policy CRUD

huntergregory:
This PR adds controller exec time metrics for pods/namespaces/policies. Metric labels are "operation" (create/update/delete) and "error" ("true" or "false").

Completed UTs:
- apply time metrics
- v1 controllers

TODO:
- UTs for v2 controllers

### DIFF
--- a/npm/http/server/server.go
+++ b/npm/http/server/server.go
@@ -28,9 +28,8 @@ func NPMRestServerListenAndServe(config npmconfig.Config, npmEncoder json.Marsha
 
 	// prometheus handlers
 	if config.Toggles.EnablePrometheusMetrics {
-		rs.router.Handle(api.NodeMetricsPath, metrics.GetHandler(metrics.CustomerNodeMetrics))
-		rs.router.Handle(api.ClusterMetricsPath, metrics.GetHandler(metrics.CustomerClusterMetrics))
-		rs.router.Handle(api.DataplaneHealthMetricsPath, metrics.GetHandler(metrics.DataplaneHealthMetrics))
+		rs.router.Handle(api.NodeMetricsPath, metrics.GetHandler(metrics.NodeMetrics))
+		rs.router.Handle(api.ClusterMetricsPath, metrics.GetHandler(metrics.ClusterMetrics))
 	}
 
 	if config.Toggles.EnableHTTPDebugAPI {
--- a/npm/metrics/namespaces.go
+++ b/npm/metrics/namespaces.go
@@ -1,0 +1,16 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// RecordNamespaceApplyTime adds an observation of namespace apply time for the specified apply mode.
+// The execution time is from the timer's start until now.
+func RecordNamespaceApplyTime(timer *Timer, mode ApplyMode) {
+	timer.stopAndRecordApplyTime(namespaceApplyTime, mode)
+}
+
+// GetPodApplyCount returns the number of observations for namespace apply time for the specified apply mode.
+// This function is slow.
+func GetNamespaceApplyCount(mode ApplyMode) (int, error) {
+	labels := prometheus.Labels{applyModeLabel: string(mode)}
+	return getCountValue(namespaceApplyTime.With(labels))
+}
--- a/npm/metrics/namespaces.go
+++ b/npm/metrics/namespaces.go
@@ -1,18 +1,13 @@
 package metrics
 
-import (
-	"github.com/prometheus/client_golang/prometheus"
-)
-
-// RecordNamespaceExecTime adds an observation of namespace exec time for the specified operation.
+// RecordControllerNamespaceExecTime adds an observation of namespace exec time (unless the operation is NoOp).
 // The execution time is from the timer's start until now.
-func RecordNamespaceExecTime(timer *Timer, op OperationKind) {
-	timer.stopAndRecordCRUDExecTime(controllerNamespaceExecTime, op)
+func RecordControllerNamespaceExecTime(timer *Timer, op OperationKind, hadError bool) {
+	timer.stopAndRecordCRUDExecTime(controllerNamespaceExecTime, op, hadError)
 }
 
-// GetNamespaceExecCount returns the number of observations for namespace exec time for the specified operation.
+// GetControllerNamespaceExecCount returns the number of observations for namespace exec time for the specified operation.
 // This function is slow.
-func GetNamespaceExecCount(op OperationKind) (int, error) {
-	labels := prometheus.Labels{operationLabel: string(op)}
-	return getCountVecValue(controllerNamespaceExecTime, labels)
+func GetControllerNamespaceExecCount(op OperationKind, hadError bool) (int, error) {
+	return getCountVecValue(controllerNamespaceExecTime, getCRUDExecTimeLabels(op, hadError))
 }
--- a/npm/metrics/namespaces.go
+++ b/npm/metrics/namespaces.go
@@ -1,16 +1,18 @@
 package metrics
 
-import "github.com/prometheus/client_golang/prometheus"
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
 
-// RecordNamespaceApplyTime adds an observation of namespace apply time for the specified apply mode.
+// RecordNamespaceExecTime adds an observation of namespace exec time for the specified operation.
 // The execution time is from the timer's start until now.
-func RecordNamespaceApplyTime(timer *Timer, mode ApplyMode) {
-	timer.stopAndRecordApplyTime(namespaceApplyTime, mode)
+func RecordNamespaceExecTime(timer *Timer, op OperationKind) {
+	timer.stopAndRecordCRUDExecTime(controllerNamespaceExecTime, op)
 }
 
-// GetPodApplyCount returns the number of observations for namespace apply time for the specified apply mode.
+// GetNamespaceExecCount returns the number of observations for namespace exec time for the specified operation.
 // This function is slow.
-func GetNamespaceApplyCount(mode ApplyMode) (int, error) {
-	labels := prometheus.Labels{applyModeLabel: string(mode)}
-	return getCountValue(namespaceApplyTime.With(labels))
+func GetNamespaceExecCount(op OperationKind) (int, error) {
+	labels := prometheus.Labels{operationLabel: string(op)}
+	return getCountVecValue(controllerNamespaceExecTime, labels)
 }
--- a/npm/metrics/pods.go
+++ b/npm/metrics/pods.go
@@ -1,16 +1,13 @@
 package metrics
 
-import "github.com/prometheus/client_golang/prometheus"
-
-// RecordPodExecTime adds an observation of pod exec time for the specified operation.
+// RecordControllerPodExecTime adds an observation of pod exec time for the specified operation (unless the operation is NoOp).
 // The execution time is from the timer's start until now.
-func RecordPodExecTime(timer *Timer, op OperationKind) {
-	timer.stopAndRecordCRUDExecTime(controllerPodExecTime, op)
+func RecordControllerPodExecTime(timer *Timer, op OperationKind, hadError bool) {
+	timer.stopAndRecordCRUDExecTime(controllerPodExecTime, op, hadError)
 }
 
-// GetPodExecCount returns the number of observations for pod exec time for the specified operation.
+// GetControllerPodExecCount returns the number of observations for pod exec time for the specified operation.
 // This function is slow.
-func GetPodExecCount(op OperationKind) (int, error) {
-	labels := prometheus.Labels{operationLabel: string(op)}
-	return getCountVecValue(controllerPodExecTime, labels)
+func GetControllerPodExecCount(op OperationKind, hadError bool) (int, error) {
+	return getCountVecValue(controllerPodExecTime, getCRUDExecTimeLabels(op, hadError))
 }
--- a/npm/metrics/pods.go
+++ b/npm/metrics/pods.go
@@ -1,0 +1,16 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// RecordPolicyApplyTime adds an observation of pod apply time for the specified apply mode.
+// The execution time is from the timer's start until now.
+func RecordPodApplyTime(timer *Timer, mode ApplyMode) {
+	timer.stopAndRecordApplyTime(podApplyTime, mode)
+}
+
+// GetPodApplyCount returns the number of observations for pod apply time for the specified apply mode.
+// This function is slow.
+func GetPodApplyCount(mode ApplyMode) (int, error) {
+	labels := prometheus.Labels{applyModeLabel: string(mode)}
+	return getCountValue(podApplyTime.With(labels))
+}
--- a/npm/metrics/pods.go
+++ b/npm/metrics/pods.go
@@ -2,15 +2,15 @@ package metrics
 
 import "github.com/prometheus/client_golang/prometheus"
 
-// RecordPolicyApplyTime adds an observation of pod apply time for the specified apply mode.
+// RecordPodExecTime adds an observation of pod exec time for the specified operation.
 // The execution time is from the timer's start until now.
-func RecordPodApplyTime(timer *Timer, mode ApplyMode) {
-	timer.stopAndRecordApplyTime(podApplyTime, mode)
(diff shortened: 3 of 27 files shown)


npm/metrics/namespaces.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// RecordNamespaceApplyTime adds an observation of namespace apply time for the specified apply mode.
+// The execution time is from the timer's start until now.
+func RecordNamespaceApplyTime(timer *Timer, mode ApplyMode) {
+	timer.stopAndRecordApplyTime(namespaceApplyTime, mode)
+}
+
+// GetPodApplyCount returns the number of observations for namespace apply time for the specified apply mode.
```
nitishm, line 11: nit: comment doesnt match the function name

npm/http/server/server.go
```diff
@@ -28,8 +28,9 @@ func NPMRestServerListenAndServe(config npmconfig.Config, npmEncoder json.Marsha
 
 	// prometheus handlers
 	if config.Toggles.EnablePrometheusMetrics {
-		rs.router.Handle(api.NodeMetricsPath, metrics.GetHandler(true))
-		rs.router.Handle(api.ClusterMetricsPath, metrics.GetHandler(false))
+		rs.router.Handle(api.NodeMetricsPath, metrics.GetHandler(metrics.CustomerNodeMetrics))
+		rs.router.Handle(api.ClusterMetricsPath, metrics.GetHandler(metrics.CustomerClusterMetrics))
+		rs.router.Handle(api.DataplaneHealthMetricsPath, metrics.GetHandler(metrics.DataplaneHealthMetrics))
```
nitishm, line 33: What is the reasoning behind 3 different endpoint URLs for scraping prometheus metrics? Traditionally it's just one endpoint serving metrics for all components of the workload.

npm/metrics/prometheus-metrics.go
```diff
@@ -55,29 +78,102 @@ const (
...
+	UpdateMode ApplyMode = "update"
+	DeleteMode ApplyMode = "delete"
+)
+
+var knownApplyModes = map[ApplyMode]struct{}{
+	CreateMode: {},
+	UpdateMode: {},
+	DeleteMode: {},
+}
+
```
nitishm, line 141: Another option is to associate an `isValid()` method to `ApplyMode` where you can check if a declared enum value is part of the `ApplyMode` collection. 
Nothing major - it just got me thinking about what would be a more idiomatic way of checking for validity in Go

npm/metrics/prometheus-metrics.go
```diff
@@ -55,29 +78,102 @@ const (
...
+	DataplaneHealthMetrics RegistryType = "dataplane-health"
+)
+
 var (
-	nodeLevelRegistry    = prometheus.NewRegistry()
-	clusterLevelRegistry = prometheus.NewRegistry()
-	haveInitialized      = false
+	allRegistries   = []RegistryType{CustomerNodeMetrics, CustomerClusterMetrics, DataplaneHealthMetrics}
+	registries      = make(map[RegistryType]*prometheus.Registry)
+	haveInitialized = false
```
nitishm, line 125: Not the biggest fan of globals for storing state. If you get a chance maybe create a `Metric` struct to store state and associate all the functions that you have below to that struct?

npm/metrics/prometheus-values.go
```diff
@@ -25,8 +25,12 @@ func getVecValue(gaugeVecMetric *prometheus.GaugeVec, labels prometheus.Labels)
 
 // getCountValue returns the number of times a Summary metric has recorded an observation.
 // This function is slow.
-func getCountValue(summaryMetric prometheus.Summary) (int, error) {
-	dtoMetric, err := getDTOMetric(summaryMetric)
+func getCountValue(summaryMetric interface{}) (int, error) {
```
nitishm, line 28: why `interface{}`? Will this ever be of any type other than `prometheus.Collector`?

npm/pkg/controlplane/controllers/v1/networkPolicyController.go
```diff
@@ -259,6 +266,11 @@ func (c *NetworkPolicyController) syncNetPol(key string) error {
 	// If DeletionTimestamp of the netPolObj is set, start cleaning up lastly applied states.
 	// This is early cleaning up process from updateNetPol event
 	if netPolObj.ObjectMeta.DeletionTimestamp != nil || netPolObj.ObjectMeta.DeletionGracePeriodSeconds != nil {
+		if _, ok := c.rawNpMap[key]; ok {
+			// record time to delete policy if it exists (can't call within cleanUpNetworkPolicy because this can be called by a pod update)
+			timer := metrics.StartNewTimer()
+			defer metrics.RecordPolicyApplyTime(timer, metrics.DeleteMode)
```
nitishm, line 272: Out of curiosity what happens if we error out while performing deletion? The time elapsed will skew the metrics, will it not, when compared to successful deletes, unless we can correlate the two events? 
My concern is when we aggregate metrics over a period of time to get a sense of deletion times (for performance statistic), if we were erroring out we would have short deletion times, which might inadvertently convey that we are performing well in deletions since we do not have the error context associated with the captured time metric.
Something to think about. WDYT? /cc @vakalapa @matmerr

nitishm: Maybe an extra field that passes the success state when `RecordXYZApplyTime(...)` is called?

npm/pkg/controlplane/controllers/v1/parsePolicy.go
```diff
@@ -33,6 +33,7 @@ func isSamePolicy(old, new *networkingv1.NetworkPolicy) bool {
 		return false
 	}
 
+	fmt.Println("same!!!")
```
nitishm, line 36: stray print?

npm/pkg/controlplane/controllers/v2/networkPolicyController.go
```diff
@@ -269,6 +280,15 @@ func (c *NetworkPolicyController) syncAddAndUpdateNetPol(netPolObj *networkingv1
 		return errNetPolTranslationFailure
 	}
 
+	_, policyExisted := c.rawNpSpecMap[netpolKey]
+	var mode metrics.ApplyMode
+	if policyExisted {
+		mode = metrics.UpdateMode
+	} else {
+		mode = metrics.CreateMode
+	}
```
nitishm, line 289:
```suggestion
var mode metrics.ApplyMode
if _, ok := c.rawNpSpecMap[netpolKey]; ok {
    mode = metrics.UpdateMode
} else {
    mode = metrics.CreateMode
}
```

nitishm: ah I see - you need to use it in the follow up code. Nevermind - you can ignore this suggestion

nitishm (commented):
Nits and questions. Otherwise looks good to me functionally

npm/metrics/prometheus-values.go
huntergregory: it can be a prometheus.Observer too. This function is only used for UTs

npm/pkg/controlplane/controllers/v1/networkPolicyController.go
huntergregory: In the past we've recorded regardless of success. NPM rarely fails unless there's an IcM-worthy problem if I'm not wrong @matmerr, @vakalapa ?

An error or two shouldn't really affect an overall average of say 30+ adds I think. Those quick runs also wouldn't impact the quantiles (0.5, 0.9, 0.99).

npm/metrics/prometheus-metrics.go
huntergregory: I like this suggestion, although it would require changes throughout the codebase to use a struct and methods for external calls to the package (if that's what you're suggesting?). Or are you suggesting a Registries struct for internal use?

npm/http/server/server.go
huntergregory: We already have two endpoints and have to keep this for back compat, so to keep in that same vein, I was thinking we leave customer-facing metrics on separate endpoints (nodemetrics and clustermetrics) and internal on others

huntergregory: we could rename dataplane health to "daemon-metrics", "internal-metrics", or just "metrics". Wdyt?

huntergregory: option 2: we can dump all dataplane health metrics onto node-metrics

We would probably want to update Container Insights to drop those num_failures metrics then though (they're set up to scrape all metrics from "/node-metrics"

vakalapa: @huntergregory  lets think the future, this distinction we add now may live forever for the same backward compatibility reason. so lets think about next 2-3 year plan, what is the best course forward. I would not want to have multiple different endpoints for different kinds of metrics. All the "internal" metrics are relevant for customers too and they would want to see them in their workbooks. What if we keep existing metrics as is and just distribute the new metrics onto node metrics for daemon specifc and cluster metrics for generic metrics ?

npm/metrics/namespaces.go
```diff
@@ -0,0 +1,16 @@
...
+// This function is slow.
+func GetNamespaceApplyCount(mode ApplyMode) (int, error) {
```
vakalapa, line 13: Can you remind me again why we are doing recording delete events ?

npm/pkg/controlplane/controllers/v1/nameSpaceController.go
```diff
@@ -348,6 +360,8 @@ func (nsc *NamespaceController) syncAddNameSpace(nsObj *corev1.Namespace) error
 
 // syncUpdateNameSpace handles updating namespace in ipset.
 func (nsc *NamespaceController) syncUpdateNameSpace(newNsObj *corev1.Namespace) error {
+	timer := metrics.StartNewTimer()
```
vakalapa, line 363: Instead of dealing these metrics here, we can just put then at the start of syncAddNS and cleanupDeleteNS, that will keep the code fairly simple ?

npm/pkg/controlplane/controllers/v1/networkPolicyController.go
vakalapa: Agreed, the error percent will be low, so the results wont be skewed way too much. Good thing in V2 we are logging faulre counts too. Lets keep V1 simple and enhance V2 with such error check logic as separate effort.

npm/pkg/controlplane/controllers/v1/networkPolicyController.go
```diff
@@ -245,6 +245,13 @@ func (c *NetworkPolicyController) syncNetPol(key string) error {
 	if err != nil {
 		if errors.IsNotFound(err) {
 			klog.Infof("Network Policy %s is not found, may be it is deleted", key)
+
+			if _, ok := c.rawNpMap[key]; ok {
+				// record time to delete policy if it exists (can't call within cleanUpNetworkPolicy because this can be called by a pod update)
+				timer := metrics.StartNewTimer()
```
vakalapa, line 251: Again, we can put this at the top of the clean up and syncAdd functions, that will reduce the complexity of adding it in multiple places.

npm/metrics/namespaces.go
```diff
@@ -0,0 +1,16 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// RecordNamespaceApplyTime adds an observation of namespace apply time for the specified apply mode.
+// The execution time is from the timer's start until now.
+func RecordNamespaceApplyTime(timer *Timer, mode ApplyMode) {
```
vakalapa, line 7: Can we rename this func to RecordNamespaceApplyTime -> RecordNamespaceExecTime, same with other resource functions too, since this is not just apply instead it lso records delete and update

npm/metrics/pods.go
```diff
@@ -0,0 +1,16 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// RecordPolicyApplyTime adds an observation of pod apply time for the specified apply mode.
+// The execution time is from the timer's start until now.
+func RecordPodApplyTime(timer *Timer, mode ApplyMode) {
```
vakalapa, line 7: RecordPodApplyTime ->  RecordPodExecTime

npm/metrics/prometheus-metrics.go
```diff
@@ -24,16 +30,33 @@ var (
...
 
+	// TODO do update/delete
+	policyApplyTimeName = "policy_apply_time"
+	policyApplyTimeHelp = "Execution time in milliseconds for updating/deleting a network policy. NOTE: for apply time for adding, see add_policy_exec_time"
+	applyModeLabel      = "apply_mode"
+
+	podApplyTimeName = "pod_apply_time"
+	podApplyTimeHelp = "Execution time in milliseconds for adding/updating/deleting a pod"
+
+	namespaceApplyTimeName = "namespace_apply_time"
```
vakalapa, line 57: Is there a reason for us to combine all resource events into the same metric ? I think there is value in keeping one metric for one type of resource event. Wdyt ?

npm/http/server/server.go
huntergregory: Good points. Let's go with this strategy for the daemon. 

Not necessary to decide now, but I'm thinking we should just have a "/metrics" endpoint for the fan-out controller, and those metrics should have their own registry and initialization function.

npm/metrics/prometheus-metrics.go
huntergregory: reverting these changes since we will keep the original two registries

npm/metrics/prometheus-metrics.go
huntergregory: discussed offline that using vector metrics with labels is a good strategy

npm/pkg/controlplane/controllers/v1/networkPolicyController.go
huntergregory: synced offline about keeping as is

npm/http/server/server.go
huntergregory: Final update on this:
Now we have controller metrics and daemon metrics. Seems easiest to preserve the /node-metrics and /cluster-metrics endpoints for both the fan-out and vanilla npm, since both the vanilla controller and daemon have metrics on each endpoint.

npm/metrics/prometheus-values.go
huntergregory: changed back to prometheus.Collector and introduced a new function for getting counts for vectors

npm/pkg/controlplane/controllers/v1/networkPolicyController.go
huntergregory: Incorporated this suggestion. Now metrics have an "error" label

npm/metrics/prometheus-metrics.go
```diff
@@ -53,71 +39,183 @@ const (
...
+	controllerExecTimeLabels    = []string{operationLabel, hadErrorLabel}
+
+	// TODO add health metrics
 )
 
-// InitializeAll creates all the Prometheus Metrics. The metrics will be nil before this method is called.
+type RegistryType bool
+
+const (
+	NodeMetrics    RegistryType = false
```
vakalapa, line 104: Why are we switching off node metrics by default ?

npm/metrics/prometheus-metrics.go
```diff
@@ -53,71 +39,183 @@ const (
...
+}
+
+// initializeControllerMetrics creates metrics modified by the controller
+func initializeControllerMetrics() {
+	// CLUSTER METRICS
+	numPolicies = createClusterGauge(numPoliciesName, numPoliciesHelp)
+	// TODO include health metrics: num failures for validating policies & ipsets
+
+	// NODE METRICS
+	addPolicyExecTime = createNodeSummaryVec(addPolicyExecTimeName, "", addPolicyExecTimeHelp, addPolicyExecTimeLabels)
```
vakalapa, line 183: addPolicyExecTime should be a daemon metric, like ACLRuleExectime or aaddIPsetExectime.

vakalapa: i understand that it is being used in controller space now, but ideally this should be taken care of in dp.AddPOlicy

npm/metrics/prometheus-metrics.go
huntergregory: let me switch the type of RegistryType to a string for easier readability

vakalapa (approved):

PR #1222, opened 2022-02-04
introduce a code coverage diff check

pjohnst5:
**Reason for Change**:
Introduce a code coverage check


rbtr:
what's the `"$(COVERAGE_DROP_THRESHOLD)"`? i don't even think that should be configurable, we should not allow PRs to decrease coverage

pjohnst5:
Closed #1270 in favor or this one

PR #1224, opened 2022-02-07
Mellanox priority vlan tag set script

pjohnst5:
**Reason for Change**:
Provide a script for partner teams to set the PriorityVLANTag on windows multitenant VMs

Connect X-4 case:
image

Connect X-3 case:
image


scripts/Set-NetAdapterPriorityVLanTag.ps1
```diff
@@ -0,0 +1,52 @@
+function Set-NetAdapterPriorityVLanTag
```
vakalapa, line 1: Can you add a comment at the very top on why we need this script.

pjohnst5: Yessir

PR #1228, opened 2022-02-08
fix: panic caused by nil Writer provided to Logger

### LINKED ISSUE #1227: Creating a new logger on Linux without write permissions to /var causes panic
When creating a new `log.(*Logger)` with `NewLogger`, and then subsequently initializing it with `log.(*Logger).Init`, a panic will result when any of the logging methods are invoked (e.g. `log.(*Logger).Debugf`). The resulting panic (abridged for privacy) is:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x4feb76]

goroutine 1 [running]:
log.(*Logger).Output(0xc00051e7d0, 0x13, {0xc000524b70, 0x23})
        /home/tim/goroots/go1.17/src/log/log.go:184 +0x356
log.(*Logger).Printf(0x198e233, {0xc00055e198, 0xc000511ac8}, {0xc00022d9a0, 0x2, 0x73e3a5})
        /home/tim/goroots/go1.17/src/log/log.go:191 +0x4c
github.com/Azure/azure-container-networking/log.(*Logger).logf(0xc000213810, {0x19937f5, 0xb}, {0xc00022d9a0, 0x1, 0x1})
        /home/tim/go/pkg/mod/github.com/!azure/azure-container-networking@v1.4.16/log/logger.go:204 +0xf0
github.com/Azure/azure-container-networking/log.(*Logger).Logf(0xc000213810, {0x19937f5, 0x7fa856e433a0}, {0xc00022d9a0, 0x10, 0x7fa87dc72a68})
        /home/tim/go/pkg/mod/github.com/!azure/azure-container-networking@v1.4.16/log/logger.go:210 +0x8c
github.com/Azure/azure-container-networking/log.(*Logger).Debugf(...)
        /home/tim/go/pkg/mod/github.com/!azure/azure-container-networking@v1.4.16/log/logger.go:229
```

The root cause of this is the `nil` writer being provided to the `log.New` invocation. If write permissions were available, this would be set properly by `log.SetOutput`.

I don't know if this behavior is also present on Windows since I haven't personally tested it. In an out-of-band conversation, @ramiro-gamarra mentioned that this is a Linux-specific issue with this package.

timraymond:
**Reason for Change**:

In the log.(*Logger).Init method, several different strategies are
considered based on the configuration provided by the user. In one
common case, a rotating log file is opened and the log output is tee-ed
into it. If there is any issue with opening that log file, the
`log.(*Log).SetOutput` method is never invoked on the underlying logger.
Since this standard library logger was created with `nil` passed as the
`io.Writer`, this results in a nil pointer deref panic. The error
causing the issue is also not surfaced to the user.

This initializes the standard library logger with the `io.Discard`
writer. This adheres to the original intention here, while also being a
safer alternative to nil (effectively writes are to /dev/null). Also,
the error is written out to stderr, in the hopes that an operator will
be able to diagnose and fix the issue (in the absense of a correct
logger configuration).

Finally, the logger initialization has been moved to a new
function, `log.NewLoggerE` that properly exposes the underlying error.
The original function, `NewLogger`, has been modified to use this
function instead and has been marked Deprecated.

**Issue Fixed**:
Fixes #1227

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -58,7 +58,11 @@ type Logger struct {
 
 var pid = os.Getpid()
 
-// NewLoggerE creates a new Logger and surfaces any errors encountered during the process
+// NewLoggerE creates a new Logger and surfaces any errors encountered during
+// the process. The returned logger is guaranteed to be safe to use when a
+// non-nil error is returned, but may have undesired behavior. Callers should
+// treat the logger as nil under error conditions unless necessary for
+// backwards compatibility reasons.
 func NewLoggerE(name string, level, target int, logDir string) (*Logger, error) {
 	logger := &Logger{
 		l:            log.New(io.Discard, logPrefix, log.LstdFlags),
@@ -72,7 +76,8 @@ func NewLoggerE(name string, level, target int, logDir string) (*Logger, error) 
 
 	err := logger.SetTarget(target)
 	if err != nil {
-		return nil, fmt.Errorf("setting log target: %w", err)
+		// we *do* want to return the logger here for backwards compatibility
+		return logger, fmt.Errorf("setting log target: %w", err)
 	}
 	return logger, nil
 }


log/logger_test.go
```diff
@@ -14,6 +15,28 @@ const (
...
+	targetDir := "/definitelyDoesNotExist"
+
+	// TODO(timraymond): this is some duplicated logic from
+	// (*Logger).getFileName. It should be possible to make it a publicly
+	// callable function to make this a little less brittle
+	fullLogPath := path.Join(targetDir, logName+".log")
+
+	// confirm the assumptions of this test before we run it:
+	if _, err := os.Stat(fullLogPath); err == nil {
+		t.Skipf("The log file at %q exists, so this test cannot sensibly run. Delete it first", fullLogPath)
```
rbtr, line 31: would it be better to use a relative path and always delete it during test setup?

timraymond: @rbtr I like the general idea. I did a bit more research and found [this SO post](https://unix.stackexchange.com/questions/643520/is-there-a-file-that-will-always-not-exist). The Go equivalent is [`os.TempDir`](https://pkg.go.dev/os@go1.17.6#TempDir) which also does the right thing on Linux and Windows, so I'm inclined to use it. Patch forthcoming :)

ramiro-gamarra: +1 on `os.TempDir`

rbtr (changes requested):
deleting the wrong thing in the test, i think

log/logger_test.go
```diff
@@ -14,6 +15,29 @@ const (
...
+	// unwriteable directory
+
+	// this test needs a guaranteed empty directory, so we create a temporary one
+	// and ensure that it gets destroyed afterward.
+	targetDir := os.TempDir()
+	t.Cleanup(func() {
+		// This removal could produce an error, but since it's a temporary
+		// directory anyway, this is a best-effort cleanup
+		os.Remove(targetDir)
+	})
```
rbtr, line 29: my reading of [os.TempDir()](https://pkg.go.dev/os#TempDir) is that it will return `/tmp`, and i don't think deleting _that_ is what you want to do here

rbtr (changes requested):
forgot to include the comment 🙃

log/logger_test.go
timraymond: Ah, yep--good catch. `os.MkdirTemp` is what I wanted. Yay code reviews :)

log/logger.go
```diff
@@ -58,20 +58,39 @@ type Logger struct {
...
+	if err != nil {
+		// ideally this would be returned to the caller, but this API is depended
+		// on by unknown parties. Given at this point we have an unusable (but
+		// safe) logger, we log to stderr with the standard library in hopes that
+		// an operator will see the error and be able to take corrective action
+		log.Println("error initializing logger: err:", err)
+	}
 
-	return &logger
+	return logger
```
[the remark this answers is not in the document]
timraymond: Also a good catch. Habit makes me `return nil, err` :)

rbtr (approved):
lgtm

PR #1230, opened 2022-02-09
feat: [NPM] send more AI logs

huntergregory:
This PR does the following:
- Sends a log when NPM starts up (allows us to see if a version reached a customer since we can look for that log)
- Sends a heartbeat log which runs in iptables/policy manager periodic tasks. It includes the number of policies in the cluster
- Sends error logs in the v2 dataplane, which previously sent nothing

### DIFF
--- a/npm/metrics/ai-utils.go
+++ b/npm/metrics/ai-utils.go
@@ -69,7 +69,7 @@ func SendErrorLogAndMetric(operationID int, format string, args ...interface{}) 
 // SendMetric sends metrics
 func SendMetric(metric aitelemetry.Metric) {
 	if th == nil {
-		log.Logf("AppInsights didn't initialized.")
+		log.Logf("AppInsights didn't initialize")
 		return
 	}
 	th.TrackMetric(metric)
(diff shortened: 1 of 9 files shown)


npm/metrics/ai-utils.go
```diff
@@ -81,10 +76,23 @@ func SendMetric(metric aitelemetry.Metric) {
...
 // SendLog sends log
-func SendLog(report aitelemetry.Report) {
+func SendLog(operationID int, msg string) {
+	report := aitelemetry.Report{
+		Message:          msg,
+		Context:          strconv.Itoa(operationID),
+		CustomDimensions: make(map[string]string),
+	}
 	if th == nil {
 		log.Logf("AppInsights didn't initialized.")
```
nitishm, line 86: grammar

vakalapa (dismissed):

vakalapa (approved):

vakalapa:
@rbtr  Is there an example zap implementation in our repo, we can add a backlog item for this improvement and base the change on example implementation.

PR #1232, opened 2022-02-10
Allow cns to register node if no NCs are present

pjohnst5:
**Reason for Change**:
Allow CNS to register the node if no NCs are present

**Issue Fixed**:
Livesite issue 13242071

**Validation**
#unitTests and,
When NCs are not present, CNS **allows** re-registration of node:
image
image
image
image

When NCs **are** present, CNS **denies** re-registration of node:
image
image
image
image

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -746,7 +746,7 @@ func (service *HTTPRestService) setOrchestratorType(w http.ResponseWriter, r *ht
 	service.dncPartitionKey = req.DncPartitionKey
 	nodeID = service.state.NodeID
 
-	if nodeID == "" || nodeID == req.NodeID {
+	if nodeID == "" || nodeID == req.NodeID || !service.areNCsPresent() {
 		switch req.OrchestratorType {
 		case cns.ServiceFabric:
 			fallthrough
--- a/cns/restserver/util.go
+++ b/cns/restserver/util.go
@@ -600,6 +600,20 @@ func (service *HTTPRestService) getNetworkContainerDetails(networkContainerID st
 	return containerDetails, containerExists
 }
 
+// areNCsPresent returns true if NCs are present in CNS, false if no NCs are present
+func (service *HTTPRestService) areNCsPresent() bool {
+	service.RLock()
+	defer service.RUnlock()
+
+	if len(service.state.ContainerStatus) == 0 {
+		if len(service.state.ContainerIDByOrchestratorContext) == 0 {
+			return false
+		}
+	}
+
+	return true
+}
+
 // Check if the network is joined
 func (service *HTTPRestService) isNetworkJoined(networkID string) bool {
 	namedLock.LockAcquire(stateJoinedNetworks)
--- a/cns/restserver/util.go
+++ b/cns/restserver/util.go
@@ -602,10 +602,8 @@ func (service *HTTPRestService) getNetworkContainerDetails(networkContainerID st
 
 // areNCsPresent returns true if NCs are present in CNS, false if no NCs are present
 func (service *HTTPRestService) areNCsPresent() bool {
-	if len(service.state.ContainerStatus) == 0 {
-		if len(service.state.ContainerIDByOrchestratorContext) == 0 {
-			return false
-		}
+	if len(service.state.ContainerStatus) == 0 && len(service.state.ContainerIDByOrchestratorContext) == 0 {
+		return false
 	}
 	return true
 }
--- a/cns/restserver/util.go
+++ b/cns/restserver/util.go
@@ -602,15 +602,11 @@ func (service *HTTPRestService) getNetworkContainerDetails(networkContainerID st
 
 // areNCsPresent returns true if NCs are present in CNS, false if no NCs are present
 func (service *HTTPRestService) areNCsPresent() bool {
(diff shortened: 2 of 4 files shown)


cns/restserver/util.go
```diff
@@ -600,6 +600,20 @@ func (service *HTTPRestService) getNetworkContainerDetails(networkContainerID st
 	return containerDetails, containerExists
 }
 
+// areNCsPresent returns true if NCs are present in CNS, false if no NCs are present
+func (service *HTTPRestService) areNCsPresent() bool {
+	service.RLock()
+	defer service.RUnlock()
```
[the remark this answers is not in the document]
pjohnst5: If I remove those two lines, is that good then?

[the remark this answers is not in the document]
ramiro-gamarra: yes, removing these would be fine

pjohnst5:
What do you think @rbtr about these go-lint failures?
I think it's because there's a lock object in CNS
` Error: copylocks: range var tt copies lock:`
`Error: copylocks: assignment copies lock value to tt:`

and then this one is because I'm trying to do just in-line structs:
`Error: Error return value of `encoding/json.Marshal` is not checked (errchkjson)`
https://github.com/Azure/azure-container-networking/pull/1232/files#diff-053ea523f14080c985c2992e7a397d824ced2e6eaa58dada5886438ae58d8111R236

Do you think I could just merge it in as is 😅 ?

rbtr:
yeah, it's mad because you're not supposed to copy a mutex and you're making a local copy of that tt struct. should be fine to throw a `//nolint` and an explanation of why at the end of those lines

cns/restserver/util.go
```diff
@@ -600,6 +600,16 @@ func (service *HTTPRestService) getNetworkContainerDetails(networkContainerID st
 	return containerDetails, containerExists
 }
 
+// areNCsPresent returns true if NCs are present in CNS, false if no NCs are present
+func (service *HTTPRestService) areNCsPresent() bool {
+	if len(service.state.ContainerStatus) == 0 {
+		if len(service.state.ContainerIDByOrchestratorContext) == 0 {
```
rbtr, line 606: can you one-line this with an `||`? or maybe it should be a switch?
```suggestion
	if len(service.state.ContainerStatus) == 0 || len(service.state.ContainerIDByOrchestratorContext) == 0 {
```

pjohnst5:
Does the `//nolint` go at the end of the line itself? Or at the top of the file?

rbtr (changes requested):

cns/restserver/util.go
pjohnst5: I would probably `&&` it instead, so that both things need to be empty for it to say it's empty

rbtr:
> Does the `//nolint` go at the end of the line itself? Or at the top of the file?

end of line. ex:
```
tt := tt //nolint:copylocks // this mutex copy is to keep a local reference to this variable in the test func closure, and is ok
```
//nolint:<the linter to ignore> // <the reason for the nolint>

cns/restserver/util.go
pjohnst5: I put an `&&` @rbtr so it's one line now

pjohnst5:
> end of line. ex:

Thanks @rbtr, linter seems happy now

cns/restserver/util.go
```diff
@@ -600,6 +600,14 @@ func (service *HTTPRestService) getNetworkContainerDetails(networkContainerID st
 	return containerDetails, containerExists
 }
 
+// areNCsPresent returns true if NCs are present in CNS, false if no NCs are present
+func (service *HTTPRestService) areNCsPresent() bool {
+	if len(service.state.ContainerStatus) == 0 && len(service.state.ContainerIDByOrchestratorContext) == 0 {
+		return false
+	}
+	return true
```
rbtr, line 608: just a thought, this pattern is fundamentally 
```go
if (bool) { return false }
return true
```
which is really
```go
return !bool
```

cns/restserver/api.go
```diff
@@ -746,7 +746,7 @@ func (service *HTTPRestService) setOrchestratorType(w http.ResponseWriter, r *ht
 	service.dncPartitionKey = req.DncPartitionKey
 	nodeID = service.state.NodeID
 
-	if nodeID == "" || nodeID == req.NodeID {
+	if nodeID == "" || nodeID == req.NodeID || !service.areNCsPresent() {
 		switch req.OrchestratorType {
 		case cns.ServiceFabric:
```
rbtr, line 751: i know you didn't change this, but this `fallthrough` pattern is very unusual in go. if all of these switch tests should actually execute the same case, it's better to do:
```go
case cns.ServiceFabric, cns.Kubernetes, cns.KubernetesCRD, [... other tests ...]:
    // the case
```

rbtr (approved):
lgtm, but left a couple comments that for things you could improve if you feel like it

cns/restserver/api.go
pjohnst5: Good point

cns/restserver/util.go
pjohnst5: Ah I didn't think about that

PR #1233, opened 2022-02-11
feat: [NPM] v2 as default config & PlaceFirst toggle functionality for v2

huntergregory:
Default toggle updates:
- EnableV2NPM = true
- PlaceAzureChainFirst = true

Now PlaceAzureChainFirst = false would impact v2 so that the jump from FORWARD chain to AZURE-NPM chain would be placed after the jump to KUBE-SERVICES.


vakalapa (dismissed):

vakalapa (approved):

huntergregory:
tests passed before the last commit (only affected a UT), so merging now since "Build ACN and Test" passed

PR #1236, opened 2022-02-12
Don't send NNC to Pool Monitor before Reconciler starts

rbtr:

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -954,6 +954,8 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 	}
 	logger.Printf("reconciled initial CNS state after %d attempts", attempt)
 
+	// start the pool Monitor before the Reconciler, since it needs to be ready to receive an
+	// NodeNetworkConfig update by the time the Reconciler tries to send it.
 	go func() {
 		logger.Printf("Starting IPAM Pool Monitor")
 		if e := poolMonitor.Start(ctx); e != nil {
@@ -997,7 +999,9 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 		return errors.Wrapf(err, "failed to setup reconciler with manager")
 	}
 
-	// Start the RequestController which starts the reconcile loop
+	// Start the Manager which starts the reconcile loop.
+	// The Reconciler will send an initial NodeNetworkConfig update to the PoolMonitor, starting the
+	// Monitor's internal loop.
 	go func() {
 		logger.Printf("Starting NodeNetworkConfig reconciler.")
 		for {
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -820,8 +820,8 @@ type ncStateReconciler interface {
 }
 
 // TODO(rbtr) where should this live??
-// InitCNS initializes cns by passing pods and a createnetworkcontainerrequest
-func initCNS(ctx context.Context, cli nodeNetworkConfigGetter, ncReconciler ncStateReconciler, podInfoByIPProvider cns.PodInfoByIPProvider) error {
+// reconcileInitialCNSState initializes cns by passing pods and a CreateNetworkContainerRequest
+func reconcileInitialCNSState(ctx context.Context, cli nodeNetworkConfigGetter, ncReconciler ncStateReconciler, podInfoByIPProvider cns.PodInfoByIPProvider) error {
 	// Get nnc using direct client
 	nnc, err := cli.Get(ctx)
 	if err != nil {
@@ -864,8 +864,6 @@ func initCNS(ctx context.Context, cli nodeNetworkConfigGetter, ncReconciler ncSt
 
 // InitializeCRDState builds and starts the CRD controllers.
 func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cnsconfig *configuration.CNSConfig) error {
-	logger.Printf("[Azure CNS] Starting request controller")
-
 	// convert interface type to implementation type
 	httpRestServiceImplementation, ok := httpRestService.(*restserver.HTTPRestService)
 	if !ok {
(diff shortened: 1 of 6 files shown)


cns/service/main.go
```diff
@@ -939,10 +918,18 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+	}
+	// TODO(rbtr): nodename and namespace should be in the cns config
+	scopedcli := kubecontroller.NewScopedClient(nnccli, types.NamespacedName{Namespace: "kube-system", Name: nodeName})
+
 	// apiserver nnc might not be registered or api server might be down and crashloop backof puts us outside of 5-10 minutes we have for
 	// aks addons to come up so retry a bit more aggresively here.
 	// will retry 10 times maxing out at a minute taking about 8 minutes before it gives up.
 	err = retry.Do(func() error {
+		logger.Printf("attempting to initCNS")
 		err = initCNS(ctx, scopedcli, httpRestServiceImplementation, podInfoByIPProvider)
```
neaggarwMS, line 933: rename it to ReconcileCNSState?

cns/service/main.go
```diff
@@ -864,8 +864,6 @@ func initCNS(ctx context.Context, cli nodeNetworkConfigGetter, ncReconciler ncSt
 
 // InitializeCRDState builds and starts the CRD controllers.
 func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cnsconfig *configuration.CNSConfig) error {
```
neaggarwMS, line 866: InitAndStartController?

cns/service/main.go
```diff
@@ -939,10 +918,18 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+		return errors.Wrap(err, "failed to create NNC client")
+	}
+	// TODO(rbtr): nodename and namespace should be in the cns config
+	scopedcli := kubecontroller.NewScopedClient(nnccli, types.NamespacedName{Namespace: "kube-system", Name: nodeName})
+
 	// apiserver nnc might not be registered or api server might be down and crashloop backof puts us outside of 5-10 minutes we have for
 	// aks addons to come up so retry a bit more aggresively here.
 	// will retry 10 times maxing out at a minute taking about 8 minutes before it gives up.
 	err = retry.Do(func() error {
+		logger.Printf("attempting to initCNS")
```
neaggarwMS, line 932: Can you also add a log when CNS state is reconciled successfully

cns/service/main.go
```diff
@@ -939,19 +920,46 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
 	if err != nil {
 		return err
 	}
+	logger.Printf("reconciled initial CNS state after %d attempts", attempt)
+
+	// initialize the ipam pool monitor
+	poolOpts := ipampool.Options{
+		RefreshDelay: poolIPAMRefreshRateInMilliseconds * time.Millisecond,
+	}
+	poolMonitor := ipampool.NewMonitor(httpRestServiceImplementation, scopedcli, &poolOpts)
```
neaggarwMS, line 954: Can you initialize the poolMonitor before reconcileInitialCNSState retrier loop.

cns/service/main.go
```diff
@@ -939,19 +920,46 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+	httpRestServiceImplementation.IPAMPoolMonitor = poolMonitor
+	go func() {
```
neaggarwMS, line 956: Can you move this after Reconciler is started.

cns/service/main.go
```diff
@@ -1005,8 +1014,8 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 		}
 	}()
 
-	logger.Printf("Starting SyncHostNCVersion")
 	go func() {
```
neaggarwMS, line 1017: This also must be started only after reconciler is started. Can you add a Started flag in reconciler and control the order?

neaggarwMS (changes requested):

cns/service/main.go
rbtr: the PoolMonitor needs to be started (and waiting to receive an NNC) before the Reconciler, otherwise the Reconciler will block when it tries to update the Monitor with a new NNC

cns/singletenantcontroller/reconciler.go
```diff
@@ -37,18 +38,24 @@ type Reconciler struct {
...
 		ipampoolmonitorcli: ipampipampoolmonitorcli,
 		nnccli:             nnccli,
+		started:            make(chan interface{}),
 	}
 }
 
 // Reconcile is called on CRD status changes
 func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	defer func() {
+		r.once.Do(func() { close(r.started) })
```
neaggarwMS, line 57: Does this gets set even if we are returning error?

cns/service/main.go
neaggarwMS: Can you update this in comments? Also in IPAMPoolMonitor we should also ensure reconciler is Started. Today we rely only on Update API and determine that reconciler got started. That is an incorrect assumption as any other component can call Update function

neaggarwMS (changes requested):

cns/singletenantcontroller/reconciler.go
rbtr: moved it to the end

cns/service/main.go
rbtr: I have updated the comments and docs on these methods to be clear what the effects and side-effects of calling them are. I think trying to have the Monitor check if the Reconciler is the component sending it an NNC update tangles them together too much.

neaggarwMS: Exactly, we cant expect to track who is calling update. But IPAM Pool Monitor can ensure to Start only it gets a NNC update + Reconciler is Started.

neaggarwMS (approved):

PR #1239, opened 2022-02-15
Create rbac.yaml in acn repo to use in automation

aegal:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/tools/acncli/deployment/rbac.yaml
+++ b/tools/acncli/deployment/rbac.yaml
@@ -1,0 +1,30 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: network-mgr
+rules:
+- apiGroups: [""]
+  resources: ["pods","nodes"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: ["acn.azure.com"]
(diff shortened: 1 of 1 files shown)


rbtr:
What is the use case here? We already have [this](https://github.com/Azure/azure-container-networking/tree/master/test/integration/manifests/cns) and [this](https://github.com/Azure/azure-container-networking/blob/master/cns/azure-cns.yaml).

aegal:
> What is the use case here? We already have [this](https://github.com/Azure/azure-container-networking/tree/master/test/integration/manifests/cns) and [this](https://github.com/Azure/azure-container-networking/blob/master/cns/azure-cns.yaml).

This is a little different to those and used in the mdnc scenario to validate rc can talk to the customer cluter, was going to use it in our runner automation.  My initial thought was since this is suppose to get applied on the worker nodes where CNS lives, then it would be a good fit for it here since it will make it easier to update. 

But after chatting with Matt Long, I'll leave it where it currently is with the runner code and close this P.R. 

Also, sorry for the late response, didn't see this comment.

aegal:
found another place to have this yaml for our automation

PR #1241, opened 2022-02-15
fix: [NPM] disable v2 CLI & polish linux policy manager

huntergregory:
Changes:
- disable the CLI for v2
- update some comments
- polish linux policy manager
  - increase the max try count to 2 for iptables-restore
  - fix the exit code that we ignore when deleting jumps to policy chains (now we ignore when the rule doesn't exist

### DIFF
--- a/npm/http/server/server.go
+++ b/npm/http/server/server.go
@@ -32,8 +32,10 @@ func NPMRestServerListenAndServe(config npmconfig.Config, npmEncoder json.Marsha
 		rs.router.Handle(api.ClusterMetricsPath, metrics.GetHandler(metrics.ClusterMetrics))
 	}
 
-	if config.Toggles.EnableHTTPDebugAPI && npmEncoder != nil {
-		// ACN CLI debug handlerss
+	// TODO support the debug CLI for v2
+	// the nil check is for fan-out npm
+	if config.Toggles.EnableHTTPDebugAPI && npmEncoder != nil && !config.Toggles.EnableV2NPM {
+		// ACN CLI debug handlers
 		rs.router.Handle(api.NPMMgrPath, rs.npmCacheHandler(npmEncoder)).Methods(http.MethodGet)
 	}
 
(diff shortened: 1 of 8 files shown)


npm/http/server/server.go
```diff
@@ -32,8 +32,10 @@ func NPMRestServerListenAndServe(config npmconfig.Config, npmEncoder json.Marsha
 		rs.router.Handle(api.ClusterMetricsPath, metrics.GetHandler(metrics.ClusterMetrics))
 	}
 
-	if config.Toggles.EnableHTTPDebugAPI && npmEncoder != nil {
-		// ACN CLI debug handlerss
+	// TODO support the debug CLI for v2
+	// the nil check is for fan-out npm
+	if config.Toggles.EnableHTTPDebugAPI && npmEncoder != nil && !config.Toggles.EnableV2NPM {
```
vakalapa, line 37: Why do we need to disable GetCache for V2, this is useful irrespective of debug CLI.

npm/pkg/dataplane/policies/policymanager_linux.go
```diff
@@ -12,14 +12,34 @@ import (
...
+	// the error message describes the error if this pattern is within the message
+	knownLineErrorPattern = "Error occurred at line: (\\d+)"
+
+	chainSectionPrefix = "chain"
 )
 
+/*
+Error handling for iptables-restore:
+Currently we retry on any error and will make two tries max.
+The section IDs and line error patterns are pointless currently. Although we can eventually use them to skip a section with an error and salvage the rest of the file.
```
vakalapa, line 28: nit: Can you reduce this oneline into two, its too long.

vakalapa (approved):

npm/http/server/server.go
huntergregory: currently the cache is buggy for v2. Let's fix this in another PR

npm/pkg/dataplane/policies/policymanager_linux.go
huntergregory: addressing in followup PR

PR #1252, opened 2022-02-18
feat: [NPM] lazy delete ipsets

huntergregory:
Place the burden of cleaning up empty/unreferenced ipsets on the dataplane. The ipset manager will remove all deletable sets from the main cache, and for `ApplyAllIPSets` mode, will add those sets to the `toDeleteCache`, ready to be removed on the next `ApplyIPSets` call. 

This PR does the following:
- removes `DeleteIPSet` calls from the controllers:
  - previously, we would only delete `Namespace` ipsets in namespace controller
- creates an ipset manager reconcile function, and moves the reconcile go thread out of policy manager into DP

Minor updates:
- fixes integration testing structs (some NPMNetPols didn't have RuleIPSets prior)
- adds thorough UTs for Add/Delete Reference in ipset manager
- moves some windows-specific ipset manager code to the windows file


npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -484,6 +440,27 @@ func (iMgr *IPSetManager) exists(name string) bool {
...
+			return
+		}
+	} else if !set.canBeDeleted() {
+		return
+	}
+
+	delete(iMgr.setMap, set.Name)
+	metrics.DecNumIPSets()
+	if iMgr.iMgrCfg.IPSetMode == ApplyAllIPSets {
+		// NOTE: in ApplyAllIPSets mode, if this ipset has never been created in the kernel, it would be added to the deleteCache, and then the OS would fail to delete it
```
vakalapa, line 458: <nit> line too long. Make it into two lines

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -484,6 +440,27 @@ func (iMgr *IPSetManager) exists(name string) bool {
...
+		iMgr.modifyCacheForKernelRemoval(set.Name)
+	}
+	// if mode is ApplyOnNeed, the set will not be in the kernel (or will be in the delete cache already) since there are no references
```
vakalapa, line 461: Even in applyONNeed we might end in situation where a set is needed with members and then we delete the referenced netpol/list. We can end up with empty sets in kernel right ?

test/mystuff/main.go
```diff
@@ -0,0 +1,100 @@
+package main
```
vakalapa, line 1: Can you rename this folder from mystuff to something else if we intend to keep this file ?

huntergregory: whoops shouldn't have added this

npm/pkg/dataplane/ipsets/ipsetmanager.go
huntergregory: anytime the set is modified, we see if it should be in the kernel now, and it to the deleteCache if needed

npm/pkg/dataplane/dataplane.go
```diff
@@ -78,7 +81,26 @@ func (dp *DataPlane) BootupDataplane() error {
...
+				return
+			case <-ticker.C:
+				// locks ipset manager
+				dp.ipsetMgr.Reconcile()
+
+				// in Windows, does nothing
+				// in Linux, locks policy manager but can be interrupted
+				dp.policyMgr.Reconcile()
+
+				metrics.SendHeartbeatLog()
```
vakalapa, line 100: Can we have the heart beat at the very top and in a gofunc, so it wont block the setMgr or pMgr reconcile.

vakalapa (approved):

PR #1258, opened 2022-02-25
Add TLS to transport shim using Secure gRPC

nitishm:
gRPC now uses self signed certificates for secure gRPC comms. 
Note: Self signed certificate is valid only for 365 days at the moment so that we can annually rotate these certs manually

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


nitishm:
TODO: Need to add a make target to generate new certificates. But this is a makeshift arrangement. We need a better way to rotate our certificates on an annual basis. Ideal is to use cert-manager or some automation. Another (unsafe) option is to increase the validity to a really large period.
Thoughts @vakalapa ?

npm/deploy/certs/server-key.pem
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
rbtr, line 1: this doesn't feel like the correct thing to do

nitishm: Yep I agree but we dont have an option here. We do not have a certificate management controller so we have to hard code these.
We could programmatically generate these but for now everything will be embedded in a Kubernetes Secret. 
Programmatically might be a better option TBH but just to move ahead we check these in since this isn't yet going into production.
Thoughts @vakalapa ?

nitishm:
Abandoning this in favor of programmatically generating the certs.

PR #1260, opened 2022-02-28
Added the Pod Net ARM ID as a Label.

rsagasthya:
Added Pod Net Arm Id as labels to the CNS Metrics.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/ipampool/metrics.go
+++ b/cns/ipampool/metrics.go
@@ -5,66 +5,81 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
+const (
+	subnetLabel     = "subnet"
+	subnetCIDRLabel = "subnet_cidr"
+)
+
 var (
-	ipamAllocatedIPCount = prometheus.NewGauge(
+	ipamAllocatedIPCount = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "ipam_pod_allocated_ips",
 			Help: "Count of IPs CNS has allocated to Pods.",
 		},
+		[]string{subnetLabel, subnetCIDRLabel},
 	)
-	ipamAvailableIPCount = prometheus.NewGauge(
+	ipamAvailableIPCount = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "ipam_available_ips",
 			Help: "Available IP count.",
 		},
+		[]string{subnetLabel, subnetCIDRLabel},
 	)
-	ipamBatchSize = prometheus.NewGauge(
+	ipamBatchSize = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "ipam_batch_size",
 			Help: "IPAM IP pool batch size.",
 		},
+		[]string{subnetLabel, subnetCIDRLabel},
 	)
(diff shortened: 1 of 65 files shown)


cns/ipampool/metrics.go
```diff
@@ -6,80 +6,81 @@ import (
...
+	subnetLabel      = "subnet"
+	subnetCIDRLabel  = "subnet_cidr"
+	podnetARMIDLabel = "podnet_arm_id"
 )
 
 var (
 	ipamAllocatedIPCount = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Name: "ipam_pod_allocated_ips",
+			Name: "cx_ipam_pod_allocated_ips",
```
rbtr, line 17: for the cx metrics, can you also add a ConstantLabel in the Opts with key `customer_metric`

cns/ipampool/monitor.go
```diff
@@ -18,6 +19,8 @@ const (
 	DefaultRefreshDelay = 1 * time.Second
 	// DefaultMaxIPs default maximum allocatable IPs
 	DefaultMaxIPs = 250
+	// Subnet ARM ID /subscriptions/$(SUB)/resourceGroups/$(GROUP)/providers/Microsoft.Network/virtualNetworks/$(VNET)/subnets/$(subnet)
```
rbtr, line 22: consistent casing -> `$(SUBNET)`

cns/ipampool/monitor.go
```diff
@@ -49,8 +52,8 @@ type Monitor struct {
 	once        sync.Once
 }
 
-// Global Variables for Subnet and SubnetAddressSpace
-var subnet, subnetCIDR string
+// Global Variables for Subnet, Subnet Address Space and Pod Network ARM ID
+var subnet, subnetCIDR, subnetARMID string
```
rbtr, line 56: previously called `podnetARMID`, change one or the other to be consistent (i think calling it subnetARMID is better)

cns/ipampool/monitor.go
```diff
@@ -343,6 +347,20 @@ func (pm *Monitor) GetStateSnapshot() cns.IpamPoolMonitorStateSnapshot {
 	}
 }
 
+// GenerateARMID uses the Subnet ARM ID format to populate the ARM ID with the metadata.
+// If either of the metadata attributes are empty, then the ARM ID will be an empty string.
+func GenerateARMID(nnc *v1alpha.NodeNetworkConfig) string {
+	subscription := nnc.Status.NetworkContainers[0].SubscriptionID
+	resourceGroup := nnc.Status.NetworkContainers[0].ResourceGroupID
+	vnetID := nnc.Status.NetworkContainers[0].VNETID
+	subnetID := nnc.Status.NetworkContainers[0].SubnetID
```
rbtr, line 356: this should probably accept an NC, not the whole NNC, and the caller should pick which NC is passed in so that we have the option to do the right thing still if the NNC has multiple NCs

rbtr (changes requested):

rbtr (changes requested):

rbtr (approved):
lgtm

PR #1262, opened 2022-03-01
[NPM] Generate TLS certificates during docker build and bake into image for gRPC secure channels

nitishm:
This change adds the generation of the Server cert and CA cert to the docker image. These certs are available to both the client and the server side to establish a secure gRPC connection between client & server.

Changes include,
- Adding tls helper functions to transport package
- Generate script to generate the TLS certificates
- Dockerfile (Linux) updated to run the generate script at build time.

TODOs,
- Create a generate script for Windows.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,10 @@ ipam-*.xml
 controller-gen
 build/tools/bin
 npm/debug/http
+
+# certificates
+*/**/certs/
+*.crt
+*.pem
+*.pem
+*.srl
--- a/.gitignore
+++ b/.gitignore
@@ -29,5 +29,4 @@ npm/debug/http
 */**/certs/
 *.crt
 *.pem
-*.pem
 *.srl
--- a/npm/Dockerfile
+++ b/npm/Dockerfile
@@ -20,6 +20,9 @@ FROM docker.io/ubuntu:focal
 COPY --from=builder /usr/local/bin/azure-npm \
 	/usr/bin/azure-npm
 
+COPY --from=builder /usr/local/src/npm/npm/scripts \
+  /usr/local/npm
+
 # Install dependencies.
 RUN apt-get update
 RUN apt-get install -y iptables
@@ -28,6 +31,9 @@ RUN apt-get install -y ca-certificates
 RUN apt-get upgrade -y
 
 RUN chmod +x /usr/bin/azure-npm
+
+WORKDIR /usr/local/npm
+RUN ./generate_certs.sh
 
 # Run the npm command by default when the container starts.
 ENTRYPOINT ["/usr/bin/azure-npm", "start"]
--- a/npm/cmd/root.go
+++ b/npm/cmd/root.go
@@ -1,7 +1,14 @@
 package main
 
 import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	npmconfig "github.com/Azure/azure-container-networking/npm/config"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"k8s.io/klog"
 )
 
 // NewRootCmd returns a root cobra command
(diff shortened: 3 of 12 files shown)


npm/pkg/transport/tls.go
```diff
@@ -0,0 +1,48 @@
...
+	"os"
+
+	"google.golang.org/grpc/credentials"
+)
+
+const (
+	serverCertPEMFilename = "tls.crt"
+	serverKeyPEMFilename  = "tls.key"
+	caCertPEMFilename     = "ca.crt"
+	path                  = "/usr/local/npm"
```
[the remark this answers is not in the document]
nitishm: The certs are baked in through the Dockerfile. Not sure how to make that dynamic. Hence I made this static :/

.gitignore
```diff
@@ -24,3 +24,10 @@ ipam-*.xml
 controller-gen
 build/tools/bin
 npm/debug/http
+
+# certificates
+*/**/certs/
+*.crt
+*.pem
+*.pem
```
huntergregory, line 32: nit: duplicate

npm/pkg/transport/events_client.go
```diff
@@ -81,11 +90,12 @@ func (c *EventsClient) run(ctx context.Context, stopCh <-chan struct{}) error {
 		default:
 			if connectClient == nil {
 				klog.Info("Reconnecting to gRPC server controller")
-				opts := []grpc.CallOption{grpc.WaitForReady(true)}
+				opts := []grpc.CallOption{grpc.WaitForReady(false)}
```
huntergregory, line 93: curious what this represents and why we had it true before, false now

npm/scripts/generate_certs.sh
```diff
@@ -0,0 +1,38 @@
...
+if [ ! -f "$SAN_CNF_FILE" ]
+then
+		echo "SAN_CNF_FILE does not exist"
+		exit
+fi
+
+if [ ! -d "$CERTS_STAGING_DIR" ]
+then
+    echo "Creating $CERTS_STAGING_DIR"
+		mkdir -p $CERTS_STAGING_DIR
```
huntergregory, line 25: nit: inconsistent tab spacing (in above if statement too)

npm/Dockerfile
```diff
@@ -20,6 +20,9 @@ FROM docker.io/ubuntu:focal
 COPY --from=builder /usr/local/bin/azure-npm \
 	/usr/bin/azure-npm
 
+COPY --from=builder /usr/local/src/npm/npm/scripts \
```
huntergregory, line 23: is `npm/npm/` intentional?

npm/cmd/root.go
```diff
@@ -12,6 +19,27 @@ func NewRootCmd() *cobra.Command {
 		CompletionOptions: cobra.CompletionOptions{
 			DisableDefaultCmd: true,
 		},
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
```
huntergregory, line 22: why move this from the start cmd? Also has some impacts like to npm/Dockerfile, so might be better to scope to another PR?

nitishm: What Dockerfile changes do you anticipate @huntergregory ?

npm/Dockerfile
nitishm: Yep. Since I didnt want to change the WORKDIR in the builder image ([here](https://github.com/Azure/azure-container-networking/pull/1262/files/143d39929ebbd7773bb3a2a1662bc1c8137ec331#diff-9109d573bc356b2f5cfb6ccfa3a9a94643911cb8ef8e6b9de1e3a40debcf26f3R8-R11)) which copies the host's source to `/usr/local/src/npm` it has to be this way.

npm/pkg/transport/events_client.go
nitishm: True blocks the connection until the server is ready to accept the connection. We don't want to block our connection since the server pod could crash which we are connecting or in the load-balancing case reject the connection in which case the agent would need to block until timeout. Instead we try sending, error out and try reconnecting.

npm/cmd/root.go
huntergregory: perhaps?
https://github.com/Azure/azure-container-networking/blob/d6ba4b9e363cbd394cb544753f7bb0b31fdcf9dc/npm/Dockerfile#L33

nitishm: So are you suggesting we update this to "controller"/"daemon"? The default behavior, as for now, should not use fan-out IMO.

huntergregory: Took another look at the code, and this makes sense now.

vakalapa (approved):

npm/pkg/transport/tls.go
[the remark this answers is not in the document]
rbtr: you could make it dynamic using a volume mount at runtime; it would be better for this to be configurable imo

PR #1265, opened 2022-03-02
test: [NPM] Make Cyclonus Profiles Independent

huntergregory:
Also adds non-truncated logs for every NPM pod


vakalapa:
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits 

We will need to keep these limits in mind

huntergregory:
> https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits
> 
> We will need to keep these limits in mind

I think this limit is the only relevant one for the PR:

"Workflow run queue - No more than 500 workflow runs can be queued in a 10 second interval per repository. If a workflow run reaches this limit, the workflow run is terminated and fails to complete."

Worst case, 5 cyclonus workflows will be triggered now instead of 1.

huntergregory:
Closing because GitHub now has a feature to rerun individual failed jobs

PR #1267, opened 2022-03-05
[CNI] Return retriable error for transient errors

tamilmani1989:
2. cni to return retriable error for transient errors

**Reason for Change**:

This PR addresses CNI to return retriable error for transient errors so that cri would retry on error. This will prevent ipleaks in case if cni cannot connect to cns or cni cannot remove endpoint due to transient error

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/plugin/main.go
+++ b/cni/network/plugin/main.go
@@ -6,7 +6,6 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	cniTypes "github.com/containernetworking/cni/pkg/types"
 	"io"
 	"os"
 	"reflect"
@@ -23,6 +22,7 @@ import (
 	"github.com/Azure/azure-container-networking/store"
 	"github.com/Azure/azure-container-networking/telemetry"
 	"github.com/containernetworking/cni/pkg/skel"
+	cniTypes "github.com/containernetworking/cni/pkg/types"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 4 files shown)


cni/network/plugin/main.go
```diff
@@ -266,7 +280,7 @@ func main() {
 		}
 	}
 
-	handled, err := handleIfCniUpdate(netPlugin.Update)
+	handled, _ := handleIfCniUpdate(netPlugin.Update)
```
vakalapa, line 283: Should we log this error atleast ? if it is not nil

tamilmani1989: its getting logged inside handleIfCniUpdate already

cni/network/plugin/main.go
```diff
@@ -174,6 +175,13 @@ func main() {
 		&network.Multitenancy{},
 		&acnnetwork.AzureHNSEndpoint{},
 	)
+
+	defer func() {
+		if err != nil {
+			os.Exit(1)
+		}
+	}()
```
rbtr, line 183: i don't like this defer pattern. it would be better to make a function which can return err, have main only call it and exit with 0 or 1 based on the err, and move all of this in to that function.
short of that...defers go in to a call stack (LIFO), so this would exit before the log.Close above, so this should be the first defer

rbtr (changes requested):

tamilmani1989:
linux dualstack pipeline succeeded after i retrigger manually. so should be transient failure https://msazure.visualstudio.com/One/_build/results?buildId=52794160&view=results

rbtr (approved):

PR #1270, opened 2022-03-08
Code coverage check and publish coverage to kusto

pjohnst5:
**Reason for Change**:
Adds a code coverage check to our builds, and publishes the code coverage percentage to kusto

### DIFF
--- a/.pipelines/pipeline.yaml
+++ b/.pipelines/pipeline.yaml
@@ -1,21 +1,21 @@
 pr:
   branches:
     include:
-    - master
-    - honeycomb-dev
+      - master
+      - honeycomb-dev
 
 trigger:
   branches:
     include:
-    - master
-    - honeycomb-dev
+      - master
+      - honeycomb-dev
 
 stages:
   - stage: setup
     displayName: ACN
     jobs:
       - job: env
-        displayName: Setup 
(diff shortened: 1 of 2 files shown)


.pipelines/pipeline.yaml
```diff
@@ -128,90 +128,138 @@ stages:
...
-            os: linux
-            name: npm
-      steps:
-      - template: containers/container-template.yaml
-        parameters:
-          name: $(name)
-          os: $(os)
-          arch: $(arch)
-          tag: $(TAG)
+      - job: containerize
```
rbtr, line 189: is this whitespace change from some autoformatter? i think lists should be dedented to the same level as their parent item for max readability

rbtr:
@pjohnst5 does this have a different intent from https://github.com/Azure/azure-container-networking/pull/1222? if not, we should keep using the same PR for the same conceptual change

.pipelines/pipeline.yaml
pjohnst5: Yeah it was, I can change it

pjohnst5:
@rbtr  yes it is actually, I'll switch to #1222

pjohnst5:
Moving to #1222

PR #1271, opened 2022-03-08
test: [NPM] cylonus extended fail-fast false

vakalapa:
**Reason for Change**:

If one of the job fails, then the other jobs are getting cancelled. 

`fail-fast: false` will let other jobs in an action run until completion.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


.github/workflows/cyclonus-netpol-test.yaml
```diff
@@ -7,27 +7,35 @@ on:
...
   cyclonus-test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         # run cyclonus tests in parallel for NPM with the given ConfigMaps
-        profile: [v1-default.yaml, v1-place-azure-chain-first.yaml, v2-default.yaml, v2-apply-on-need.yaml, v2-place-azure-after-kube-services.yaml]
+        profile:
+          [
+            v1-default.yaml,
```
huntergregory, line 26: also working on this file for #1265, why don't we introduce fail fast for this file in that PR?

huntergregory (changes requested):
Can we remove updates to cyclonus-netpol-test.yaml? Also working on this file for #1265 , and we won't need fail-fast after it since there won't be any matrix strategy

test/cyclonus/install-cyclonus-exclude-sctp.yaml
```diff
@@ -20,7 +20,7 @@ spec:
             - --pod-creation-timeout-seconds=30
             - --job-timeout-seconds=2
             - --server-protocol=TCP,UDP
-            - --exclude=sctp
+            - --exclude=sctp,end-port # TODO update kind cluster to use latest k8s version and remove endport exclusion
```
matmerr, line 23: this can be done by bumping this version here
https://github.com/Azure/azure-container-networking/blob/master/test/kind/kind.yaml#L7

PR #1278, opened 2022-03-14
fix: [NPM] add a toggle to diables AITelemetry in NPM

vakalapa:
**Reason for Change**:

NPM is crashing in Gov cloud as there is no AI telemetry to be enabled. Adding this toggle to disable it in GOV regions.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/examples/windows/azure-npm.yaml
+++ b/npm/examples/windows/azure-npm.yaml
@@ -15,7 +15,7 @@ metadata:
     addonmanager.kubernetes.io/mode: EnsureExists
 rules:
   - apiGroups:
-    - ""
+      - ""
     resources:
       - pods
       - nodes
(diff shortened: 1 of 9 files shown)


npm/examples/windows/azure-npm.yaml
```diff
@@ -25,7 +25,7 @@ rules:
       - list
       - watch
   - apiGroups:
-    - networking.k8s.io
+      - networking.k8s.io
```
huntergregory, line 28: double indent?

huntergregory (commented):
So this change prevents us from failing if CreateTelemetryHandle fails. 

Do we need to create a toggle? I see that we have two options:
1. have a toggle for EnableAITelemetry
2. try to create telemetry handle no matter what

PR #1280, opened 2022-03-15
add hostname and version in the correct field in ai logs

rbtr:

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/aitelemetry/telemetrywrapper.go
+++ b/aitelemetry/telemetrywrapper.go
@@ -2,6 +2,7 @@ package aitelemetry
 
 import (
 	"fmt"
+	"os"
 	"runtime"
 	"time"
 
@@ -24,6 +25,7 @@ const (
 	vmIDStr                          = "VMID"
 	versionStr                       = "AppVersion"
 	azurePublicCloudStr              = "AzurePublicCloud"
+	hostNameKey                      = "hostname"
 	defaultTimeout                   = 10
 	defaultBatchIntervalInSecs       = 15
 	defaultBatchSizeInBytes          = 32768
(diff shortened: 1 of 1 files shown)


rsagasthya (approved):

aitelemetry/telemetrywrapper.go
```diff
@@ -207,6 +209,8 @@ func (th *telemetryHandle) TrackLog(report Report) {
 	trace.Tags.User().SetAuthUserId(runtime.GOOS)
 	trace.Tags.Operation().SetId(report.Context)
 	trace.Tags.Operation().SetParentId(th.appVersion)
+	trace.Tags.Application().SetVer(th.appVersion)
+	trace.Properties[hostNameKey], _ = os.Hostname()
```
tamilmani1989, line 213: is this different from metadata.VMName?

rbtr: yes - AI queries IMDS for some metadata like the "VMName", so you will actually get the VM Name. But the hostname that the OS reports may vary from the vm name and has no connection. Crucially for k8s, the hostname is the Pod name, and we will be able to tell what _Pod_ the logs are coming from, instead of just which VM.

aitelemetry/telemetrywrapper.go
```diff
@@ -207,6 +209,8 @@ func (th *telemetryHandle) TrackLog(report Report) {
 	trace.Tags.User().SetAuthUserId(runtime.GOOS)
 	trace.Tags.Operation().SetId(report.Context)
 	trace.Tags.Operation().SetParentId(th.appVersion)
+	trace.Tags.Application().SetVer(th.appVersion)
```
tamilmani1989, line 212: some fields are not exported to kusto. did you check if this visible in kusto? in that case we can remove `trace.Tags.Operation().SetParentId(th.appVersion)`

PR #1293, opened 2022-03-17
CNI Delete to ignore network not found error

tamilmani1989:
**Reason for Change**:

CNI delete should not throw error if network not found, This can happen if vm rebooted with pod in running state. After reboot, CRI will call CNI DEL and CNI ADD to create it again. During CNI DEL, network would not be present and cni delete throws errors and prevents pod from getting created. In this PR, when cni gets network not found, there is no point in calling ipam and it should swallow error and return nil.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -976,21 +976,12 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 
 	// Query the network.
 	if nwInfo, err = plugin.nm.GetNetworkInfo(networkID); err != nil {
-
 		if !nwCfg.MultiTenancy {
-			// attempt to release address associated with this Endpoint id
-			// This is to ensure clean up is done even in failure cases
-			log.Errorf("[cni-net] Failed to query network: %v", err)
-			telemetry.LogAndSendEvent(plugin.tb, fmt.Sprintf("Release ip by ContainerID (network not found):%v", args.ContainerID))
-			err = plugin.ipamInvoker.Delete(nil, nwCfg, args, nwInfo.Options)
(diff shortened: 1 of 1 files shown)


cni/network/network.go
```diff
@@ -976,21 +976,14 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 
 	// Query the network.
 	if nwInfo, err = plugin.nm.GetNetworkInfo(networkID); err != nil {
-
 		if !nwCfg.MultiTenancy {
```
ashvindeodhar, line 979: instead of ignoring all the non-nil errors, we should only ignore if the error is of type NetworkNotFoundError
https://github.com/microsoft/hcsshim/blob/7fa8bda4e6ba503caf0d53d0a4ee99b9a64ceed8/internal/hns/hns.go#L21

If the error is not of type NetworkNotFound, we should return error.

we do have similar code for hns v2 path (hcn). I believe this should work for v1 as well.
https://github.com/Azure/azure-container-networking/blob/96bec09d413e53e06e3776bbab3a2d6bd0e6ad31/cns/hnsclient/hnsclient_windows.go#L287

ashvindeodhar (changes requested):

cni/network/network.go
tamilmani1989: GetNetworkInfo checks cni state and return network. It doesn't call hns. If there is state it returns that else it returns error. The only error is not found

ashvindeodhar (approved):

PR #1301, opened 2022-03-23
Multitenant configs for cni manager

pjohnst5:
**Reason for Change**:
In cni manager, allow setting cns url and enable exact match for pod name


rbtr:
what's the use case?

pjohnst5:
@rbtr for multitenancy, linux
https://msazure.visualstudio.com/One/_git/Networking-Aquarius/pullrequest/5834231

rbtr:
i worry about allowing arbitrary URL injection at runtime, that could provide a path to privilege escalation

pjohnst5:
What do you say @rbtr ? Given the scenario, is it okay to merge?

jaer-tsun (approved):
lgtm

rbtr:
> What do you say @rbtr ? Given the scenario, is it okay to merge?

I'm not blocking 🙂

PR #1305, opened 2022-03-30
Add NMAgent Client

timraymond:
This adds a new nmagent package containing a Client for interacting with NMAgent. There's also some additional supporting logic in the nmagent/internal package for handling Wireserver, automatic retries, and validating requests.

Each endpoint that we use has a method on the client that accepts a context and some kind of request struct (see requests.go). The client deals with any asynchronous behavior and presents a synchronous abstraction to callers (consequently it's important to set timeouts on the inbound context if that's what's desired).

`nmagent/internal` contains an `http.RoundTripper` used to deal with the idiosyncrasies of proxying requests through Wireserver. It also contains a few other utilities for struct validation and retries also used here.

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Notes**:


nmagent/error.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Temporary reports whether the error encountered from NMAgent should be
+// considered temporary, and thus retriable
+func (e Error) Temporary() bool {
+	// We consider Unauthorized responses from NMAgent to be temporary for a
+	// certain period of time. This is to allow for situations where an
+	// authorization token may not yet be available
+	if e.Code == http.StatusUnauthorized {
+		return e.Runtime < e.Limit
+	}
```
ramiro-gamarra, line 46: This is a check that to me seems clearer higher up the stack, still trying to elaborate better on why. Let me get back to this comment at a later point.

nmagent/internal/wireserver.go
```diff
@@ -0,0 +1,123 @@
...
+	resp, err := w.Transport.RoundTrip(req)
+	if err != nil {
+		return resp, err
+	}
+	// we want to close this because we're going to replace it
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return resp, nil
+	}
```
ramiro-gamarra, line 95: It doesn't seem like this path consumes body to EOF? Sorry if I keep bringing it up, but I find it in general to be better practice to consume all bytes, especially since these should be fairly small payloads (no great advantage to streaming at the moment IMO). When these are not consumed, its possible that tcp connections won't be reused properly.

ramiro-gamarra (commented):
Few initial comments

nmagent/internal/retry.go
```diff
@@ -0,0 +1,119 @@
...
+					return fmt.Errorf("sleeping during retry: %w", err)
+				}
+				time.Sleep(delay)
+				continue
+			}
+
+			// since it's not temporary, it can't be retried, so...
+			return err
+		}
+		return nil
```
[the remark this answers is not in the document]
timraymond: Good point, in practice it probably would be pretty close to immediate since the context is propagated down to the `http.(*Client)`, but that's not a good assumption for this to have (since it's supposed to work with anything). We're definitely converging on your implementation of retries, but with some functional programming and temporary errors :) We may want to have that discussion about retry logic in general... possibly move this into a top-level package, etc.

I'm also probably going to tweak this so that the function being retried passes down the context as well... it could break up the context heirarchy a bit if it gets passed in via the closure when you actually go to use this.

nmagent/client.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// DeleteNetworkContainer removes a Network Container, its associated IP
+// addresses, and network policies from an interface
+func (c *Client) DeleteNetworkContainer(ctx context.Context, dcr DeleteContainerRequest) error {
+	requestStart := time.Now()
+
+	req, err := c.buildRequest(ctx, dcr)
+	if err != nil {
+		return fmt.Errorf("building request: %w", err)
```
rbtr, line 165: I prefer `errors.Wrap`, wdyt?

nmagent/error.go
rbtr: it does seem like this is making the error object too smart and that this is something the consumer should be making decisions about

nmagent/internal/validate.go
```diff
@@ -0,0 +1,55 @@
+package internal
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// ValidationError {{{1
```
rbtr, line 9: drop these

nmagent/internal/validate.go
```diff
@@ -0,0 +1,55 @@
...
+	val := reflect.ValueOf(obj)
+	typ := reflect.TypeOf(obj)
+
+	for i := 0; i < val.NumField(); i++ {
+		fieldVal := val.Field(i)
+		fieldTyp := typ.Field(i)
+
+		op := fieldTyp.Tag.Get("validate")
+		switch op {
+		case "presence":
```
rbtr, line 39: i'm not convinced this is necessary. we are the consumers of this package and know what inputs it needs

nmagent/internal/wireserver.go
```diff
@@ -0,0 +1,123 @@
...
+		return resp, fmt.Errorf("decoding json response from wireserver: %w", err)
+	}
+
+	// set the response status code with the *real* status code
+	realCode, err := wsResp.StatusCode()
+	if err != nil {
+		return resp, fmt.Errorf("retrieving status code from wireserver response: %w", err)
+	}
+
+	resp.StatusCode = realCode
```
rbtr, line 110: is this a change of behavior vs our current nma usage? making this not-a-drop-in change?

rbtr (commented):
globally, docstrings need punctuation.
very nice package but I feel YAGNI - we aren't providing the official nma-go library, just writing a little client for our personal use

nmagent/internal/validate.go
timraymond: So this came about from porting the existing validations over. I had a bunch of logic in the `Validate` methods for each request that looked something like:

```go
if r.SomeField == "" {
       err.MissingFields = append(err.MissingFields, "SomeField")
}

// repeat for a bunch of other stuff
```

`Validate` is another method, so when it came time to answer the question "what fields are actually required for this request?" I had to go digging into that method to figure it out. I started adding comments to the struct saying like "this field is required"... which led me to think "well, a struct tag can serve this purpose too, and it can actually do the validation." I initially thought `required:"true"`, but that brought about the question "what does `required:"false"` mean and how is that distinct from not providing it at all?" (it isn't). So I settled on `validate:"presence"`, since it does allow pulling in other common validation as well. The `Validate` methods still exist as a place for anything particularly weird, but they're pretty vestigial at this point.

It's kinda fancy, true, but it was born out of a practical refactoring. Plus I can snag the name of the field out with reflection rather than having to remember to update the string if it ever gets renamed.

timraymond:
@ramiro-gamarra Take a look at 088b710. It now reads the entire response and handles unexpected content types.

nmagent/client.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+// NewClient returns an initialized Client using the provided configuration
+func NewClient(host string, port uint16, opts ...Option) *Client {
+	client := &Client{
+		httpClient: &http.Client{
+			Transport: &internal.WireserverTransport{
+				Transport: http.DefaultTransport,
+			},
+		},
```
ramiro-gamarra, line 23: I've been thinking more about this, and I think using a Wireserver Transport will make wireserver/nmagent issues more difficult for the team to debug. My main worry is that from this layer, it's not clear if the http status code came from wireserver or nmagent, while with the existing protocol we still have both in scope. I really appreciate the investigation and effort invested in `wireserver.go`, which would still be nice to keep for documentation purposes, but for now I'd suggest to keep `httpStatusCode` in the response structs and skip the transport approach. Thoughts @timraymond @rbtr?

timraymond:
Here's the remaining items that we decided last week on this PR:

- [ ] Remove the `UnauthorizedGracePeriod` and leave that determination to the consumer of the `nmagent.Client`
- [X] Remove the functional options (see a9f5d5d)
- [ ] Remove the validation struct tags
- [ ] Remove the wireserver transport
- [ ] Address all linter feedback

timraymond:
I made a few attempts to remove the `WireserverTransport` this week... none came out well because every method of the client has to do the status code extraction. It also gets a little weird trying to communicate "did my request actually work?" What we want to avoid is confusion over "who produced this HTTP status code?" The WireserverTransport knows this information, so I changed it to communicate it upwards. This is then encapsulated in the error. Rather than deciphering two status codes by hand, the logs can say "Wireserver broke. Here's the code it gave us." Existing logic can also use the `Source` field if they care (where it seems there may be some caring).

I ran all this by @ramiro-gamarra yesterday on a call, and the only addition was the body of the request in an error case to aid in debugging.

rbtr:
@timraymond I think you still need punctuation on docstrings everywhere, otherwise I have no significant comments

nmagent/client.go
```diff
@@ -0,0 +1,208 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return c.error(resp.StatusCode, resp.Header, resp.Body)
+	}
+
+	return nil
+}
+
+func (c *Client) error(code int, headers http.Header, body io.ReadCloser) error {
+	// read the entire body
+	defer body.Close()
```
ramiro-gamarra, line 163: - Does this need to be a method? Doesn't seem like the receiver is used.
- Seems like this close is redundant with all usages of `error` (caller always closes). It could be removed, and this just accept a `Reader`

nmagent/client.go
```diff
@@ -0,0 +1,208 @@
...
+			return c.error(resp.StatusCode, resp.Header, resp.Body)
+		}
+		return nil
+	})
+
+	return err // nolint:wrapcheck // wrapping this just introduces noise
+}
+
+// GetNetworkConfiguration retrieves the configuration of a customer's virtual
+// network. Only subnets which have been delegated will be returned
```
ramiro-gamarra, line 84: My understanding is that all subnets in vnet are returned, not just delegated ones. We mark the subnets as delegated at later steps in DNC.

ramiro-gamarra (dismissed):
I think this is good to get started with testing. We can make tweaks as needed in later PRs.

timraymond:
@ramiro-gamarra I apparently vacated your approval by pushing those commits. PTAL :) @rbtr Docstrings addressed in 4fcbdd5 . I wrote a linter for that if we want to look at incorporating it.

nmagent/client.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	err = c.retrier.Do(ctx, func() error {
+		resp, err := c.httpClient.Do(req) // nolint:govet // the shadow is intentional
+		if err != nil {
+			return errors.Wrap(err, "executing request")
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK {
```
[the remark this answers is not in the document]
timraymond: The standard library will recycle connections if you close the body. I suspect that requiring package authors to drain the bodies was too obtuse, so they made `Close()` affirmatively mean "I'm done with this connection" here:
https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/transport.go;l=2135-2140

The check to see whether an EOF was seen is still present, but it's short-circuited by `alive` in this conditional:
https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/net/http/transport.go;l=2150-2153

`alive` is basically a latch that makes the connection eligible for recycling.

nmagent/client.go
timraymond: I pulled most of these doc comments straight from the NMA docs just so users wouldn't have to cross reference them all the time... This one comes from section 8.2.6. We should find out if that's true and update the upstream docs if all the subnets are returned.

nmagent/client.go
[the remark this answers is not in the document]
ramiro-gamarra: Connection re-use (or lack of) can be shown with this example:
```go
package main

import (
	"context"
	"io"
	"io/ioutil"
	"log"
	"net/http"
	"net/http/httptrace"
)

func main() {
	clientTrace := &httptrace.ClientTrace{
		GotConn: func(info httptrace.GotConnInfo) {
			log.Printf("conn: %+v", info)
		},
	}
	ctx := httptrace.WithClientTrace(context.Background(), clientTrace)

	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "http://example.com", nil)
	if err != nil {
		log.Fatal(err)
	}
	res, err := http.DefaultClient.Do(req)
	if err != nil {
		log.Fatal(err)
	}

	if _, err := io.Copy(ioutil.Discard, res.Body); err != nil {
		log.Fatal(err)
	}
	res.Body.Close()

	req, err = http.NewRequestWithContext(ctx, http.MethodGet, "http://example.com", nil)
	if err != nil {
		log.Fatal(err)
	}
	_, err = http.DefaultClient.Do(req)
	if err != nil {
		log.Fatal(err)
	}
}
```
You can comment out the code that consumes the body, and the trace should show no connection reuse. This is still the behavior in 1.18. It's possible that there are some scenarios where this won't be the case but this has bit me enough in the past to make me want to be extra careful :)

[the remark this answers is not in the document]
timraymond: Mind if I reuse this example as a repro for a Go issue? (I'll mention you on it) IMO, this violates least surprise, and it appears there's been some effort to make `Close()` do the right thing... it's apparently not complete though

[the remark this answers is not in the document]
ramiro-gamarra: Feel free. Code slightly modified from [this article](https://golang.cafe/blog/how-to-reuse-http-connections-in-go.html) btw. Would be nice to see some clarification around the code you pointed to, though it seems to me that this behavior is still consistent with the docs.

ramiro-gamarra (dismissed):

rbtr (approved):

PR #1307, opened 2022-03-31
perf: [NPM] make apply ipsets faster

huntergregory:
dramatically improves apply time for medium to large scale.

Previously, we would parse through an ipset save file for every batch dataplane operation. We can forego that by maintaining member diff for our ipset dirty cache.

New dirty cache adds robustness to all combinations of create/add/del/destroy calls. Windows doesn't need to maintain member diff by nature of HNS.

### DIFF
--- a/npm/pkg/dataplane/ipsets/dirtycache.go
+++ b/npm/pkg/dataplane/ipsets/dirtycache.go
@@ -1,12 +1,30 @@
 package ipsets
 
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Azure/azure-container-networking/npm/metrics"
+	"github.com/Azure/azure-container-networking/npm/util"
+	"k8s.io/klog"
+)
+
 /*
 	dirtyCacheInterface will maintain the dirty cache.
 	It may maintain membersToAdd and membersToDelete.
 	Members are either IPs, CIDRs, IP-Port pairs, or prefixed set names if the parent is a list.
 
 	Assumptions:
-	FIXME: TODO
+	- if the set becomes dirty via update or destroy, then the set WAS in the kernel before
+		- update won't be called unless the set IS in the kernel OR create was called before
+		- destroy won't be called unless the set IS in the kernel OR create was called before
+	- if the set becomes dirty via create, then the set was NOT in the kernel before
+		- create won't be called twice (unless a destroy call comes in between)
+
+	Examples of Expected Behavior:
+	- if a set is created and then destroyed, that set will not be in the dirty cache anymore
+	- if a set is updated and then destroyed, that set will be in the delete cache
+	- if a the only operations on a set are adding and removing the same member, the set may still be in the dirty cache, but the member will be untracked
 */
 type dirtyCacheInterface interface {
 	// reset empties dirty cache
@@ -21,10 +39,10 @@ type dirtyCacheInterface interface {
 	deleteMember(set *IPSet, member string)
 	// delete will mark the set to be deleted in the cache
 	destroy(set *IPSet)
-	// getSetsToAddOrUpdate returns the set names to be added or updated
-	getSetsToAddOrUpdate() map[string]struct{}
-	// getSetsToDelete returns the set names to be deleted
-	getSetsToDelete() map[string]struct{}
+	// setsToAddOrUpdate returns the set names to be added or updated
+	setsToAddOrUpdate() map[string]struct{}
+	// setsToDelete returns the set names to be deleted
+	setsToDelete() map[string]struct{}
 	// numSetsToAddOrUpdate returns the number of sets to be added or updated
 	numSetsToAddOrUpdate() int
 	// numSetsToDelete returns the number of sets to be deleted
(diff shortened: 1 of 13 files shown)


npm/pkg/dataplane/ipsets/dirtycache_linux.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (dc *memberAwareDirtyCache) update(originalSet *IPSet) {
+	putIntoAndRemoveFromOther(originalSet, dc.toAddOrUpdateCache, dc.toDeleteCache)
+}
+
+func (dc *memberAwareDirtyCache) delete(originalSet *IPSet) {
+	putIntoAndRemoveFromOther(originalSet, dc.toDeleteCache, dc.toAddOrUpdateCache)
+}
+
+func putIntoAndRemoveFromOther(originalSet *IPSet, intoCache, fromCache map[string]map[string]struct{}) {
```
vakalapa, line 45: into -> toCache ?

npm/pkg/dataplane/ipsets/dirtycache_linux.go
```diff
@@ -0,0 +1,119 @@
...
+	setName := originalSet.Name
+	if _, ok := intoCache[setName]; ok {
+		return
+	}
+	members, ok := fromCache[setName]
+	if !ok {
```
vakalapa, line 51: also if set is present in fromCache, then we will need to update members with current cache members

npm/pkg/dataplane/ipsets/dirtycache_windows.go
```diff
@@ -0,0 +1,86 @@
+package ipsets
+
+import "fmt"
+
+type nameOnlyDirtyCache struct {
```
vakalapa, line 5: lets just call it dirtyCache,

npm/pkg/dataplane/ipsets/dirtycache_windows.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (dc *nameOnlyDirtyCache) update(originalSet *IPSet) {
+	putIntoAndRemoveFromOther(originalSet, dc.toAddOrUpdateCache, dc.toDeleteCache)
+}
+
+func (dc *nameOnlyDirtyCache) delete(originalSet *IPSet) {
+	putIntoAndRemoveFromOther(originalSet, dc.toDeleteCache, dc.toAddOrUpdateCache)
+}
+
+func putIntoAndRemoveFromOther(set *IPSet, intoCache, fromCache map[string]struct{}) {
```
vakalapa, line 34: pointers would be better here too

npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -34,14 +34,10 @@ const (
...
-	setMap  map[string]*IPSet
-	// Map with Key as IPSet name to to emulate set
-	// and value as struct{} for minimal memory consumption.
-	toAddOrUpdateCache map[string]struct{}
-	// IPSets referred to in this cache may be in the setMap, but must be deleted from the kernel
-	toDeleteCache map[string]struct{}
-	ioShim        *common.IOShim
+	iMgrCfg    *IPSetManagerCfg
+	setMap     map[string]*IPSet
+	dirtyCache dirtyCacheMaintainer
```
vakalapa, line 39: pointer for dirtycache.

npm/pkg/dataplane/ipsets/dirtycache_linux.go
```diff
@@ -0,0 +1,119 @@
+package ipsets
+
+import "fmt"
+
+type memberAwareDirtyCache struct {
+	// map of prefixed set names to original members
+	toAddOrUpdateCache map[string]map[string]struct{}
```
vakalapa, line 7: instead of maintaining all original members, i think saving just the members to be added and members to be deleted is memory friendly. Because take the example of recent scale testing we have done, if an IPSET has large member count and we just need to a single pod ip from it, we are duplicating the large member set into dirtycache .

vakalapa (changes requested):

npm/pkg/dataplane/ipsets/dirtycache_linux.go
```diff
@@ -0,0 +1,119 @@
...
+	info, ok := dc.toDeleteCache[setName]
+	if !ok {
+		info = make(map[string]struct{})
+	}
+	dc.toAddOrUpdateCache[setName] = info
+	delete(dc.toDeleteCache, setName)
+}
+
+func (dc *memberAwareDirtyCache) update(originalSet *IPSet) {
+	putIntoAndRemoveFromOther(originalSet, dc.toAddOrUpdateCache, dc.toDeleteCache)
```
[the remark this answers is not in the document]
rbtr: maps _are_ pointers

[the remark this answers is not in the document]
rbtr: https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it

npm/pkg/dataplane/ipsets/dirtycache.go
```diff
@@ -0,0 +1,32 @@
+package ipsets
+
+// dirtyCacheMaintainer will maintain the dirty cache
+type dirtyCacheMaintainer interface {
```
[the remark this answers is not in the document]
huntergregory: let's use without lock since ipsetmanager locks itself before every operation

npm/pkg/dataplane/ipsets/ipsetmanager.go
huntergregory: as Evan mentioned, maps are pointers

rbtr: @vakalapa does it feel like [we're bullying](https://c.tenor.com/t6V-MIkkT8kAAAAC/old-man-how-many-times-do-we-have-to-teach-you.gif) you about this

npm/pkg/dataplane/ipsets/dirtycache_linux.go
```diff
@@ -0,0 +1,119 @@
+package ipsets
+
+import "fmt"
+
+type memberAwareDirtyCache struct {
```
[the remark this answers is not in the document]
huntergregory: discussed offline naming this dirtyCache in both OS's

npm/pkg/dataplane/ipsets/dirtycache.go
```diff
@@ -0,0 +1,264 @@
...
+func (dc *dirtyCache) resetAddOrUpdateCache() {
+	dc.toCreateCache = make(map[string]*memberDiff)
+	dc.toUpdateCache = make(map[string]*memberDiff)
+}
+
+func (dc *dirtyCache) create(set *IPSet) {
+	// error checking
+	if _, ok := dc.toCreateCache[set.Name]; ok {
+		msg := fmt.Sprintf("create: set %s should not already be in the toCreateCache", set.Name)
+		klog.Error(msg)
```
vakalapa, line 91: Why do we want to log this error ? we can just log a warning and move on right ?

npm/pkg/dataplane/ipsets/dirtycache.go
```diff
@@ -0,0 +1,264 @@
...
+		dc.toCreateCache[set.Name] = diffOnCreate(set)
+	}
+}
+
+// could optimize Linux to remove from toUpdateCache if there were no member diff afterwards,
+// but leaving as is prevents difference between OS caches
+func (dc *dirtyCache) addMember(set *IPSet, member string) {
+	// error checking
+	if dc.isSetToDelete(set.Name) {
+		msg := fmt.Sprintf("addMember: set %s should not be in the toDestroyCache", set.Name)
```
vakalapa, line 118: If we are trying to add a member, it means that the set should NOT be deleted, shouldnt we remove it from todeletecache and add to toupdate cache ?

npm/pkg/dataplane/ipsets/dirtycache.go
```diff
@@ -0,0 +1,264 @@
...
+	diff.addMember(member)
+}
+
+// could optimize Linux to remove from toUpdateCache if there were no member diff afterwards,
+// but leaving as is prevents difference between OS caches
+func (dc *dirtyCache) deleteMember(set *IPSet, member string) {
+	// error checking
+	if dc.isSetToDelete(set.Name) {
+		msg := fmt.Sprintf("deleteMember: set %s should not be in the toDestroyCache", set.Name)
+		klog.Error(msg)
```
vakalapa, line 141: we can make this warning

npm/pkg/dataplane/ipsets/dirtycache.go
```diff
@@ -0,0 +1,264 @@
...
+		metrics.SendErrorLogAndMetric(util.IpsmID, msg)
+		return
+	}
+
+	diff, ok := dc.toCreateCache[set.Name]
+	if ok {
+		// don't mark a member to be deleted if it never existed in the kernel
+		diff.deleteMemberIfExists(member)
```
vakalapa, line 149: Can we name this deleteMemberInDiffToAdd, more appropriate than if exists

npm/pkg/dataplane/ipsets/dirtycache.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+		diff.deleteMember(member)
+	}
+}
+
+func (dc *dirtyCache) destroy(set *IPSet) {
+	// error checking
+	if dc.isSetToDelete(set.Name) {
+		msg := fmt.Sprintf("destroy: set %s should not already be in the toDestroyCache", set.Name)
+		klog.Error(msg)
```
vakalapa, line 164: same here can be a warning, do not need an error here, and also, are these warnings required ? Do these indicate problems elsewhere in DP/setMgr code?

npm/pkg/dataplane/ipsets/dirtycache.go
huntergregory: can log it as a warning. It breaks the assumptions of the dirty cache though and shouldn't happen if ipsetmanager is using it correclty

npm/pkg/dataplane/ipsets/dirtycache.go
huntergregory: assumptions require that create is called after a destroy

npm/pkg/dataplane/ipsets/dirtycache.go
huntergregory: the error checks make sure that ipsetmanager is following the assumptions of the dirty cache. Assumptions are in the comments of the dirtyCacheInterface

npm/pkg/dataplane/ipsets/dirtycache.go
huntergregory: talked offline. Will ignore consecutive creates, but will warn on create after a set is in the update cache

npm/pkg/dataplane/ipsets/dirtycache.go
huntergregory: talked offline. Can handle this by moving to the update cache instead

npm/pkg/dataplane/ipsets/dirtycache.go
huntergregory: updated deleteMember checks to just ignore calls when isSetToDelete

npm/pkg/dataplane/ipsets/dirtycache.go
huntergregory: talked offline. Safe to make multiple destroy calls in a row

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -307,31 +349,76 @@ func (iMgr *IPSetManager) fileCreatorForApply(maxTryCount int, saveFile []byte)
...
+	}
+	for prefixedName := range setsToDelete {
+		iMgr.destroySetForApply(creator, prefixedName)
+	}
+	return creator
+}
+
+// NOTE: duplicate code in the first step in this function and fileCreatorForApplyWithSaveFile
+func (iMgr *IPSetManager) fileCreatorForApply(maxTryCount int) *ioutil.FileCreator {
+	creator := ioutil.NewFileCreator(iMgr.ioShim, maxTryCount, ipsetRestoreLineFailurePattern) // TODO make the line failure pattern into a definition constant eventually
```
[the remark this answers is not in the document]
huntergregory: talked offline. rough to duplicate a lot between the functions but no clean solution

vakalapa (approved):
Great work, this improves our perf by about 10x !! :shipit:

PR #1311, opened 2022-04-01
refactor the TLS package to reduce multiplat sprawl

rbtr:

**Reason for Change**:

Refactors the TLS package to delint it and consolidate reduce functionality that is currently spread between _linux and _windows build files down to strictly what differs between those OS implementations.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


server/tls/certificateretriever.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, block := range cr.pemBlock {
+		if block.Type == PrivateKeyLabel {
+			pk, err := x509.ParsePKCS8PrivateKey(block.Bytes)
+			if err != nil {
+				return nil, fmt.Errorf("could not parse private key %w", err)
+			}
+			return pk, nil
+		}
+	}
+	return nil, errors.Wrapf(ErrNoPrivateKey, "certificate path: %s", cr.settings.CertificatePath)
```
timraymond, line 80: Should this also use `fmt.Errorf` instead? There's a couple other instances of `pkg/errors` as well... I won't bother marking them all, but the same question applies to them :)

cns/service.go
```diff
@@ -73,14 +73,14 @@ func (service *Service) Initialize(config *common.ServiceConfig) error {
...
 			return err
 		}
-		if config.TlsSettings.TLSPort != "" {
+		if config.TLSSettings.Port != "" {
 			// listener.URL.Host will always be hostname:port, passed in to CNS via CNS command
 			// else it will default to localhost
 			// extract hostname and override tls port.
 			hostParts := strings.Split(listener.URL.Host, ":")
-			config.TlsSettings.TLSEndpoint = hostParts[0] + ":" + config.TlsSettings.TLSPort
+			config.TLSSettings.Endpoint = hostParts[0] + ":" + config.TLSSettings.Port
```
timraymond, line 81: `net.JoinHostPort` is a bit nicer for this

timraymond (approved):
Looks good. I've got a couple questions, but no showstoppers :)

server/tls/certificateretriever.go
rbtr: I would like to move towards `errors.Wrap` and more usage of `pkg/errors` since I think it is clearer. The usages of `fmt.Errorf` are existing usages that I didn't update

cns/service.go
rbtr: trying not to make any functional change in this PR, but that is a good util to keep in mind in the future

rbtr:
saving this for later

PR #1323, opened 2022-04-06
create nnc listener concept and adapt existing poolmonitor and swift …

rbtr:
…service to it

Signed-off-by: Evan Baker <rbtr@users.noreply.github.com>

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

PR #1325, opened 2022-04-08
Support NetworkContainers of Static or Dynamic types

rbtr:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

Depends on #1323 

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/singletenantcontroller/conversion.go
```diff
@@ -59,19 +58,19 @@ func SwiftNodeNetworkConfigListener(cnscli cnsClient) NodeNetworkConfigListenerF
 	}
 }
 
-// CRDStatusToNCRequest translates a crd status to createnetworkcontainer request
-func CRDStatusToNCRequest(status *v1alpha.NodeNetworkConfigStatus) (cns.CreateNetworkContainerRequest, error) {
+// CreateNCRequestFromDynamicNNC translates a crd status to createnetworkcontainer request for dynamic NNC (swift)
+func CreateNCRequestFromDynamicNNC(nnc *v1alpha.NodeNetworkConfig) (cns.CreateNetworkContainerRequest, error) {
```
thatmattlong, line 62: any particular reason we pass a pointer to nnc here? I guess since we don't do a `nil` check then it'll never be `nil`, so we can probably just not use a pointer here

cns/singletenantcontroller/conversion.go
```diff
@@ -109,7 +108,86 @@ func CRDStatusToNCRequest(status *v1alpha.NodeNetworkConfigStatus) (cns.CreateNe
...
+	subnet := cns.IPSubnet{
+		IPAddress:    primaryPrefix.Addr().String(),
+		PrefixLength: uint8(subnetPrefix.Bits()),
+	}
+
+	secondaryIPConfigs := map[string]cns.SecondaryIPConfig{}
+
+	// iterate through all IP addresses in the subnet described by primaryPrefix and
+	// add them to the request as secondary IPConfigs. Skip the specific IP of the
+	// primaryPrefix; that is the gateway.
```
thatmattlong, line 175: don't think this is true, DNC will reserve a gateway if necessary and wouldn't assign to first IP of NC CIDR but instead will be in NC's Gateway property (DefaultGateway of NNC status) so we don't need to reserve this first address here

thatmattlong (changes requested):
Don't need to reserve first IP of CIDR

cns/singletenantcontroller/conversion.go
rbtr: cheaper than doing pbv, but nil checks are a good idea

rbtr: actually, I want this communication to eventually become channels, so changing to pbv

cns/singletenantcontroller/conversion.go
rbtr: ack

thatmattlong (approved):

cns/service/main.go
```diff
@@ -865,7 +865,7 @@ func reconcileInitialCNSState(ctx context.Context, cli nodeNetworkConfigGetter,
 	}
 
 	// Convert to CreateNetworkContainerRequest
-	ncRequest, err := kubecontroller.CRDStatusToNCRequest(&nnc.Status)
+	ncRequest, err := kubecontroller.CreateNCRequestFromDynamicNC(nnc.Status.NetworkContainers[0]) // swift will only ever support 1 NC
```
neaggarwMS, line 868: >// swift will only ever support 1 NC [](http://example.com/codeflow?start=96&length=36)

This may not be true in future.

Can you update the comment " For now AKS Swift supports 1 NC"

rbtr: updated to support multiple NCs

neaggarwMS (changes requested):

cns/service/main.go
```diff
@@ -865,19 +865,23 @@ func reconcileInitialCNSState(ctx context.Context, cli nodeNetworkConfigGetter,
...
-	if err != nil {
-		return errors.Wrap(err, "failed to convert NNC status to network container request")
-	}
-	// rebuild CNS state
-	podInfoByIP, err := podInfoByIPProvider.PodInfoByIP()
-	if err != nil {
-		return errors.Wrap(err, "provider failed to provide PodInfoByIP")
-	}
+	for i := range nnc.Status.NetworkContainers {
+		ncRequest, err := kubecontroller.CreateNCRequestFromDynamicNC(nnc.Status.NetworkContainers[i])
```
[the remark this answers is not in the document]
rbtr: no, we're making the NC Request(s) from the NC(s), not from the Status

thatmattlong (approved):

PR #1326, opened 2022-04-11
Return if execution mode is v4OnlyOverlay

alam-tahmid:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -7,6 +7,7 @@ import (
 	"net"
 
 	"github.com/Azure/azure-container-networking/cni"
+	"github.com/Azure/azure-container-networking/cni/util"
 	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/iptables"
 	"github.com/Azure/azure-container-networking/log"
@@ -25,9 +26,10 @@ var (
 )
 
 type CNSIPAMInvoker struct {
-	podName      string
-	podNamespace string
-	cnsClient    cnsclient
+	podName       string
+	podNamespace  string
+	cnsClient     cnsclient
+	executionMode util.ExecutionMode
 }
 
 type IPv4ResultInfo struct {
@@ -40,11 +42,12 @@ type IPv4ResultInfo struct {
 	hostGateway        string
 }
 
-func NewCNSInvoker(podName, namespace string, cnsClient cnsclient) *CNSIPAMInvoker {
+func NewCNSInvoker(podName, namespace string, cnsClient cnsclient, executionMode util.ExecutionMode) *CNSIPAMInvoker {
 	return &CNSIPAMInvoker{
-		podName:      podName,
-		podNamespace: namespace,
-		cnsClient:    cnsClient,
+		podName:       podName,
+		podNamespace:  namespace,
+		cnsClient:     cnsClient,
+		executionMode: executionMode,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


cni/util/const.go
```diff
@@ -4,7 +4,8 @@ type ExecutionMode string
 
 // CNI execution modes
 const (
-	Default   ExecutionMode = "default"
-	Baremetal ExecutionMode = "baremetal"
-	AKSSwift  ExecutionMode = "aksswift"
+	Default       ExecutionMode = "default"
+	Baremetal     ExecutionMode = "baremetal"
+	AKSSwift      ExecutionMode = "aksswift"
+	V4OnlyOverlay ExecutionMode = "v4OnlyOverlay"
```
rbtr, line 10: `/util` is definitely not the right place for these, but I am also not convinced that an additional execution made is the right choice. 

@thatmattlong currently we have an IPAM block in the config:
```json
{
   "cniVersion":"0.3.0",
   "name":"azure",
   "plugins":[
      {
         "type":"azure-vnet",
         "mode":"transparent",
         "ipsToRouteViaHost":["..."],
         "ipam":{
            "type":"azure-cns"
         }
      },
      {
         "type":"portmap",
         "capabilities":{
            "portMappings":true
         },
         "snat":true
      }
   ]
}
```
which is unambiguous in describing Swift (azure-vnet, mode: transparent, ipam: cns). should we add the mode _here_, as maybe 
```json
         "ipam":{
            "type":"azure-cns"
            "mode":"[dynamic|static]"
         }
```

rbtr: /cc @jaer-tsun who might know more about this "execution mode" concept

rbtr (changes requested):
where we put it and what exactly the swift vs overlay config flag means needs to be considered

cni/util/const.go
thatmattlong: Not sure if ipam is the right place for it, since from CNI perspective the IPAM is just CNS in this case. To me the ipam block describes how CNI goes about getting an IP to use for the Pod, and executionMode describes how it programs the datapath for that IP, which in this case is slightly different than Swift (even though IPAM is exactly the same).

I think either way we go is tricky because a whole new execution mode just to skip adding one chain to iptables feels like overkill, but then putting this flag in ipam also doesn't feel right. Probably eventually CNS should just handle these iptables chains and then there is only one CNI mode.

tamilmani1989: i like the idea of moving swift iptable chains to cns . Even for cilium integration we are considering moving to cns

tamilmani1989, line 10: instead of v4OnlyOverlay how about swiftv4Overlay or just v4Overlay?

thatmattlong: @neaggarwMS what do you think?

neaggarwMS: Lets do v4Overlay

cni/network/invoker_cns.go
```diff
@@ -128,6 +131,11 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 		},
 	}
 
+	// ip-masq-agent handles snat in v4 overlay mode so just return here
+	if invoker.executionMode == util.V4OnlyOverlay {
```
neaggarwMS, line 135: Instead of early return, can you call the SetHostOption if executionMode != util.v4OnlyOverlay

cni/util/const.go
neaggarwMS: Actually do we need v4Overlay? How about just Overlay? @tamilmani1989 ? DualStack option should come if we have v4 and v6 IPAM

rbtr, line 10: > Lets do v4Overlay

rbtr (changes requested):

cni/util/const.go
neaggarwMS: I suggest to keep just "Overlay"

neaggarwMS (dismissed):

neaggarwMS (approved):

PR #1328, opened 2022-04-12
Don't overwrite Spec in IPAM PoolMonitor from NNC again after initialization

rbtr:

**Reason for Change**:

After init, the PoolMonitor writes the Spec, so it has the target state locally and should not let the NNC updates it receives overwrite that local state.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/ipampool/monitor.go
+++ b/cns/ipampool/monitor.go
@@ -94,7 +94,6 @@ func (pm *Monitor) Start(ctx context.Context) error {
 				continue
 			}
 		case nnc := <-pm.nncSource: // received a new NodeNetworkConfig, extract the data from it and re-reconcile.
-			pm.spec = nnc.Spec
 			scaler := nnc.Status.Scaler
 
 			// Set SubnetName, SubnetAddressSpace and Pod Network ARM ID values to the global subnet, subnetCIDR and subnetARM variables.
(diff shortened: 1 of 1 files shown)


neaggarwMS:
> 			pm.spec = nnc.Spec

WE need to put this under ReadLock too? or might as well move pm.mu.lock() here

---
Refers to: cns/ipampool/monitor.go:98 in efa059e. [](commit_id = efa059e78d57d4ccc105347093b336acb73a2d62, deletion_comment = False)

neaggarwMS (changes requested):
:clock1:

neaggarwMS (changes requested):
:clock1:

neaggarwMS:
> 	pm.clampScaler(&nnc.Status.Scaler)

We need pm.mu.Lock here too as you are updating scaler

---
Refers to: cns/ipampool/monitor.go:375 in efa059e. [](commit_id = efa059e78d57d4ccc105347093b336acb73a2d62, deletion_comment = False)

cns/ipampool/monitor.go
```diff
@@ -105,7 +104,10 @@ func (pm *Monitor) Start(ctx context.Context) error {
 			pm.metastate.batch = scaler.BatchSize
 			pm.metastate.max = scaler.MaxIPCount
 			pm.metastate.minFreeCount, pm.metastate.maxFreeCount = CalculateMinFreeIPs(scaler), CalculateMaxFreeIPs(scaler)
-			pm.once.Do(func() { close(pm.started) }) // close the init channel the first time we receive a NodeNetworkConfig.
+			pm.once.Do(func() {
+				pm.spec = nnc.Spec // set the spec from the NNC initially (afterwards we write the Spec so we know target state).
```
neaggarwMS, line 108: Can you add a log line here

neaggarwMS: Also add a log line in func (pm *Monitor) Update(nnc *v1alpha.NodeNetworkConfig) error { function.

neaggarwMS (changes requested):

neaggarwMS (dismissed):

neaggarwMS (approved):

PR #1332, opened 2022-04-15
fix: [NPM] clean up logs

huntergregory:
- remove error messages for non-error (unable to delete deprecated jump to AZURE-NPM)
- fix log that printed a pointer instead of the ipset

### DIFF
--- a/npm/pkg/dataplane/policies/chain-management_linux.go
+++ b/npm/pkg/dataplane/policies/chain-management_linux.go
@@ -21,18 +21,14 @@ const (
 	// TODO replace all util constants with local constants
 	defaultlockWaitTimeInSeconds string = "60"
 
-	doesNotExistErrorCode      int    = 1 // stderr possibility: Bad rule (does a matching rule exist in that chain?)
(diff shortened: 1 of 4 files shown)


npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -267,6 +291,11 @@ func (pMgr *PolicyManager) runIPTablesCommand(operationFlag string, args ...stri
 		errCode := exitError.ExitStatus()
 		allArgsString := strings.Join(allArgs, " ")
 		msgStr := strings.TrimSuffix(string(output), "\n")
+		for _, info := range ignored {
+			if errCode == info.exitCode && strings.Contains(msgStr, info.msg) {
+				return errCode, fmt.Errorf("not able to run iptables command [%s %s] due to benign Stderr: [%s]: %w", util.Iptables, allArgsString, msgStr, errBenignIPTablesFailure)
```
vakalapa, line 296: any advantages of returning error here?, we can safely log a message and not send any error back

vakalapa (approved):

PR #1335, opened 2022-04-18
Add dropgz module for CNI installer

rbtr:
**Reason for Change**:

Adds a CLI tool ("dropgz") for reliably deploying arbitrary files via a single self-contained binary/image.
Initial use case: to install the CNI binaries at runtime via an init container in the CNS daemonset.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
`dropgz` is effectively a CLI wrapper around Go's `embed.FS`, allowing us to pack the contents of a directory in to a binary and deploy the contents back out to the filesystem on-demand. The contents of the binary are determined at build-time, and SHA checksums are computed and stored with the files in the embed FS to allow us to verify integrity of the deployed files (bonus, we can skip re-deploying if the existing file-on-disk has the correct SHA).

Q: Why not just do this with by copying the bins to deploy in to a busybox container (or other base image with a shell that can `cp`)? 
A: Packing the files in to a Go bin lets us build in a scratch container with _no_ shell, reducing the potential attack surface for a privileged container that is writing files out to the host disk (in our use-case, writing the CNI bins which are invoked by CRI on all Pod create/delete events on a K8s Node).


dropgz/build/cni.Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.18 AS builder
+ARG VERSION
+WORKDIR /azure-container-networking
+COPY . .
+RUN CGO_ENABLED=0 go build -a -o bin/azure-vnet -trimpath -ldflags "-X main.version="$VERSION"" -gcflags="-dwarflocationlists=true" cni/network/plugin/main.go
```
wedaly, line 5: could you help me understand the steps required to add CNI binaries from another image? Particularly interested in how this interacts with multi-arch builds (I'm not very familiar with that process)

dropgz/cmd/root.go
```diff
@@ -0,0 +1,68 @@
...
+		leveler,
+	))
+
+	// bind root flags
+	root.PersistentFlags().StringVarP(&levelFlag, "log-level", "v", "info", "log level [trace,debug,info,warn,error]")
+}
+
+func Execute() {
+	if err := root.Execute(); err != nil {
+		os.Exit(1)
```
wedaly, line 57: Log the error before exiting?

dropgz/pkg/embed/payload.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		contents = append(contents, strings.TrimPrefix(path, cwd+string(filepath.Separator)))
+		return nil
+	}); err != nil {
+		return nil, errors.Wrap(err, "error walking gzfs")
+	}
+	return contents, nil
+}
+
+func Extract(path string) (io.ReadCloser, io.Closer, error) {
```
wedaly, line 45: It seems a bit awkward to me to make the caller responsible for closing both the file and the gzip reader. Maybe we could include these both in a single struct that implements `io.ReadCloser`, with a `Close()` method that closes both readers in the correct order?

dropgz/pkg/embed/payload.go
```diff
@@ -0,0 +1,83 @@
...
+		return err
+	}
+	defer c.Close()
+	defer r.Close()
+	target, err := os.Create(dest)
+	if err != nil {
+		return errors.Wrapf(err, "failed to create file %s", dest)
+	}
+	defer target.Close()
+	_, err = io.Copy(bufio.NewWriter(target), r)
```
wedaly, line 66: Might consider using [renamio](https://pkg.go.dev/github.com/google/renameio) here to reduce the risk of another process reading a half-written file (for instance, if dropgz crashes mid-write). It's definitely not perfect, but safer than writing directly to the target path.

dropgz/cmd/root.go
```diff
@@ -0,0 +1,68 @@
...
+// root represent the base invocation.
+var root = &cobra.Command{
+	Use:          "dropgz",
+	SilenceUsage: true,
+}
+
+func init() {
+	// set up signal handlers
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(context.Background())
```
wedaly, line 33: Where does this `ctx` get used? I can't seem to find any references

dropgz/pkg/embed/payload.go
```diff
@@ -0,0 +1,83 @@
...
+func Contents() ([]string, error) {
+	contents := []string{}
+	if err := fs.WalkDir(gzfs, cwd, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			return nil
+		}
+		contents = append(contents, strings.TrimPrefix(path, cwd+string(filepath.Separator)))
```
wedaly, line 37: nit: `cwd+string(filepath.Separator)` could be calculated once and reused for every path?

wedaly (commented):
overall looks good! had a few small comments/questions

One other thing to think about is testing. I'd suggest validating that all the expected files have been embedded, can be written out, and have the correct checksums (better to catch problems before this gets deployed to a cluster)

dropgz/build/cni.Dockerfile
rbtr: Answering your second question first: we don't need to worry about multi-arch here, that is handled automatically through [flags[1]](https://github.com/rbtr/azure-container-networking/blob/f09087817e0d802c34ea1a8a46f3f107a99164be/.pipelines/pipeline.yaml#L161-L168) [passed[2]](https://github.com/rbtr/azure-container-networking/blob/f09087817e0d802c34ea1a8a46f3f107a99164be/.pipelines/containers/container-template.yaml#L23-L24) [to[3]](https://github.com/rbtr/azure-container-networking/blob/f09087817e0d802c34ea1a8a46f3f107a99164be/Makefile#L248-L254) the [container builder[4]](https://github.com/rbtr/azure-container-networking/blob/f09087817e0d802c34ea1a8a46f3f107a99164be/Makefile#L181-L188). Any `FROM <image>` in this Dockerfile will do the right thing as long as `<image>` is also a multiplat image.
Which means, to your first question, we can leverage multi-stage builds and reference another CNI image like `FROM docker.io/other/cni:x.x.x as othercni`.  Then we add a line to copy bins out of that image into our target dir, right before we calculate checksums and compress (which is the next line in this file): 
```Dockerfile
FROM docker.io/other/cni:x.x.x as othercni

[do other prep]
COPY --from=othercni /bin/othercnibin /dropgz/pkg/embed/gz/othercnibin
[checksum and compress]
...
```

dropgz/pkg/embed/payload.go
rbtr: good idea, will update

dropgz/pkg/embed/payload.go
rbtr: I did consider using the temp-write and rename pattern originally but decided against it. I think it's probably not a failure mode of any consequence to us and whoever is calling dropgz should be responsible for retrying and controlling access to the target files. In the intended init-container scenario, k8s will retry before the main container comes up, and we won't be impacted by any transient partial writes.
Maybe also worth noting that renamio doesn't support Windows (and says it's not even possible to do atomic writes on Windows...) 🙁

dropgz/pkg/embed/payload.go
```diff
@@ -34,36 +35,62 @@ func Contents() ([]string, error) {
...
+
+func (rc *gzipCompoundReadCloser) Read(p []byte) (n int, err error) {
+	return rc.gzreader.Read(p)
+}
+
+func (rc *gzipCompoundReadCloser) Close() error {
+	if err := rc.gzreader.Close(); err != nil {
+		return err
+	}
+	if err := rc.file.Close(); err != nil {
```
wedaly, line 64: do we still need to close the file if gzreader.Close() fails?

rbtr: I guess we don't really need to close either because these are handles to the `embed.FS` and the process is about to exit. But maybe. I need to check the conditions that the gzreader might error on close to see if that would preclude that the underlying file be closed.

rbtr: the impl of `Close()` on the `embed.FS` file is noop 😅

dropgz/pkg/embed/payload.go
```diff
@@ -0,0 +1,110 @@
...
+	return &gzipCompoundReadCloser{file: f, gzreader: r}, nil
+}
+
+func deploy(src, dest string) error {
+	rc, err := Extract(src)
+	if err != nil {
+		return err
+	}
+	defer rc.Close()
+	target, err := os.OpenFile(dest, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o766) //nolint:gomnd // executable file bitmask
```
timraymond, line 88: I understand the immediate intention is deploying executables, but it might make sense to add the file's mode to `gzipCompoundReadCloser` so that you can recreate files with the same mode here. It might get... um, interesting with things like symlink, directory, and pipe mode bits. Though, at least if you wanted to do SUID or have a read-only config, that should work too.

dropgz/cmd/root.go
```diff
@@ -0,0 +1,68 @@
...
+
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		<-sig
+		cancel()
+		fmt.Println("exiting")
+	}()
```
timraymond, line 41: Similar to the other question about `ctx`, this seems like it effectively just swallows SIGTERMs.

timraymond (commented):
Looks pretty good--no showstoppers that I can see. I have a couple of suggestions and questions though.

dropgz/pkg/embed/payload.go
rbtr: copying the file mode from the embed is a much better idea 🤦🏼

rbtr: it's not feasible to keep the file perms without introducing additional dependencies (tar) or tabulating and restoring the bitmask ourselves. i think it is reasonable to deploy 766 and say that the caller can chmod if necessary.

dropgz/cmd/root.go
rbtr: fixed to exit, since the context is not threaded to timeouts anywhere yet

wedaly (approved):

tamilmani1989 (commented):
unit tests are missing for this new package.

tamilmani1989 (approved):
lgtm.. please add work item to cover unit tests in separate PR as discussed offline

tamilmani1989 (approved):

PR #1336, opened 2022-04-18
CNI Linux: Add 10 seconds timeout for ExecuteCommand

msvik:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


thatmattlong (dismissed):
LGTM

thatmattlong:
Any reason to do this only for linux and not windows?

msvik:
> Any reason to do this only for linux and not windows?

The livesite was seen/debugged in Linux CNI. Second, Tamilmani pointed out that his observation is that these system commands finish much faster in Linux (order of couple of seconds) and on windows take longer anyway. So, it is just safer to add timeout to Linux.

tamilmani1989 (dismissed):
lgtm

tamilmani1989:
> Any reason to do this only for linux and not windows?

In addition to what vipin said, in windows we use hns calls lot than exec cmds. we have a repair item to add timeout for hns calls

platform/os_linux.go
```diff
@@ -81,7 +82,13 @@ func (p *execClient) ExecuteCommand(command string) (string, error) {
 
 	var stderr bytes.Buffer
 	var out bytes.Buffer
-	cmd := exec.Command("sh", "-c", command)
+
+	// Create a new context and add a timeout to it
+	const timeoutSeconds = 10
```
tamilmani1989, line 87: how about we defined this in execClient object so that we configure it for unit test as well

msvik: I like the idea. Moved timeout to execclient with both default option and an option to override default.

platform/os_linux_test.go
```diff
@@ -0,0 +1,29 @@
+package platform
+
+import (
+	"testing"
+	"time"
+)
+
+// Default timeout in ExecuteCommand is 10 seconds so test should fail with 12 seconds execution time
+func TestExecuteCommandTimeout(t *testing.T) {
+	const timeout = 10 * time.Second
```
tamilmani1989, line 10: can we have less timeout value so that tests won't hang for long?

platform/os_linux_test.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+	t.Logf("%s", err.Error())
+}
+
+// Default timeout in ExecuteCommand is 10 seconds so test should pass with 8 seconds execution time
+func TestExecuteCommandNoTimeout(t *testing.T) {
+	const timeout = 10 * time.Second
+	client := NewExecClientTimeout(timeout)
+
+	_, err := client.ExecuteCommand("sleep 8")
```
tamilmani1989, line 25: same as above

tamilmani1989 (changes requested):

platform/os_linux_test.go
msvik: done

platform/os_linux_test.go
msvik: done

tamilmani1989 (approved):
lgtm

PR #1342, opened 2022-04-20
Total IPs metric

rsagasthya:
Now includes Primary IP count

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/ipampool/monitor.go
```diff
@@ -52,9 +52,13 @@ type Monitor struct {
 	once        sync.Once
 }
 
-// Global Variables for Subnet, Subnet Address Space and Subnet ARM ID
+// Global Variables:
+// For Subnet, Subnet Address Space and Subnet ARM ID
 var subnet, subnetCIDR, subnetARMID string
 
+// For List of Unique Primary IP Addresses
+var ipAddresses map[string]int
```
rbtr, line 60: These definitely shouldn't be globals; they are not properties of the Monitor, but properties of the Pool (really, of the NCs). I think they belong in the state/metastate. 
There's a problem here in that these are per-NC and we can have multiple NCs, but I don't have a suggestion on how to handle that just yet and until we _do_ have multiple NCs we don't need to deal with it

cns/ipampool/monitor.go
```diff
@@ -101,6 +107,11 @@ func (pm *Monitor) Start(ctx context.Context) error {
 			subnetCIDR = nnc.Status.NetworkContainers[0].SubnetAddressSpace
 			subnetARMID = GenerateARMID(&nnc.Status.NetworkContainers[0])
 
+			// Add Primary IP to Map, if not present.
+			for i := 0; i < len(nnc.Status.NetworkContainers); i++ {
+				ipAddresses[nnc.Status.NetworkContainers[i].PrimaryIP] = 1
+			}
```
rbtr, line 113: I don't understand why we're setting a value of `1` here, can you explain?

rbtr (changes requested):

cns/ipampool/monitor.go
rsagasthya: It is a hashmap, we need to set the value to something. The other alternative was to set it to true.

rbtr: Are you trying to use the `map` as a `Set`? The preferred way to do that is to make the Value type `interface{}` and set `ipAddresses[nnc.Status.NetworkContainers[i].PrimaryIP] = nil`, as that has no memory overhead

cns/ipampool/monitor.go
rsagasthya: Localized this map.

cns/ipampool/monitor.go
```diff
@@ -101,6 +103,15 @@ func (pm *Monitor) Start(ctx context.Context) error {
 			subnetCIDR = nnc.Status.NetworkContainers[0].SubnetAddressSpace
 			subnetARMID = GenerateARMID(&nnc.Status.NetworkContainers[0])
 
+			pm.metastate.primaryIPAddresses = make(map[string]struct{})
+			// Add Primary IP to Map, if not present.
+			// This is only for Swift i.e. if NC Type is vnet.
+			for i := 0; i < len(nnc.Status.NetworkContainers); i++ {
+				if nnc.Status.NetworkContainers[i].Type == v1alpha.VNET {
```
thatmattlong, line 110: I'd suggest checking if this is `v1alpha.VNET` or `""`, since it is possible that this CNS could be operating on NNCs created by a DNC-RC that does not yet set the type, in which case they are of type VNET

rbtr (changes requested):
commit history needs to be fixed

rbtr (approved):

cns/ipampool/monitor.go
```diff
@@ -101,6 +103,16 @@ func (pm *Monitor) Start(ctx context.Context) error {
 			subnetCIDR = nnc.Status.NetworkContainers[0].SubnetAddressSpace
 			subnetARMID = GenerateARMID(&nnc.Status.NetworkContainers[0])
 
+			pm.metastate.primaryIPAddresses = make(map[string]struct{})
+			// Add Primary IP to Map, if not present.
+			// This is only for Swift i.e. if NC Type is vnet.
+			for i := 0; i < len(nnc.Status.NetworkContainers); i++ {
```
thatmattlong, line 109: I would suggest using range instead, like: `for _, nc := range nnc.Status.NetworkContainers {...}

rbtr: that makes a copy of the NC, which we would need to update it and write it back in to the status. indexing makes more sense here. (we could `for i := range ...`, because it's a little prettier, but the C style is fine too)

PR #1345, opened 2022-04-21
build: Add Windows CNS & NPM as Part of Multi-Arch Manifest

jaer-tsun:
**Reason for Change**:

Adding windows cns to multi arch image to support windows AKS-Swift

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


.pipelines/pipeline.yaml
```diff
@@ -181,6 +181,7 @@ stages:
     dependsOn:
       - setup
       - containerize
+      - build_windows_images
```
rbtr, line 184: please work Windows into the existing pattern (using things like template variables) instead of adding special-case complexity

rbtr: what I mean is if, for example, `containers/container-template.yaml` is not generic enough to support Windows builds directly, maybe it should be `containers/container-template-$(os).yaml`. At that point we have to make no other structural changes to the main pipeline and only need to create that specific step template.

Makefile
```diff
@@ -304,7 +304,7 @@ cns-image-windows:
 	$(MKDIR) $(IMAGE_DIR); 
 	docker build \
 	--no-cache \
-	-f cns/windows.Dockerfile \
+	-f cns/Dockerfile.windows \
```
rbtr, line 307: I prefer it the other way around, `Dockerfile` is the type of file and therefore goes as the extension

rbtr (changes requested):

Makefile
jaer-tsun: ok I'll renamed the npm one

.pipelines/pipeline.yaml
jaer-tsun: I'm still confused, borrowing the existing windows-image.yaml (build_windows_images) and then waiting for it to complete before creating the multi arch manifests here.

rbtr: the `containerize` stage has been set up generically with parameters passed in to build the whole matrix of arch x image that we ship. instead of having Windows separate, fold it in so that `containerize` builds the matrix of os x arch x image.

jaer-tsun: I thought the windows images had to be done from windows pool though

rbtr: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/pool?view=azure-pipelines
> You can specify a pool at the pipeline, stage, or job level.

.pipelines/pipeline.yaml
```diff
@@ -136,8 +136,6 @@ stages:
           - publish: ./test/apimodels/
             artifact: clusterdefinitions
 
-  - template: windows-image.yaml
-
```
rbtr, line 140: 😍

rbtr, line 140: 😍

rbtr (approved):
🚀

rbtr:
@Azure/acn-npm-reviewers for npm failures here?

rbtr (approved):
lgtm pending npm check failures

PR #1346, opened 2022-04-21
Adding Keyvault Shim

ramiro-gamarra:
**Reason for Change**:

As part of enabling zero touch cert rotation for CNS deployments, we need to be able to fetch TLS certificates from Keyvault using MSI. This PR introduces the `keyvault` package, the main export of which is a `keyvault.Shim` which should make it easy to retrieve Keyvault certificates and transform them into the appropriate types for consumption in a Go http server.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:
The added modules from the Azure SDK make use of generics, which are known to be not fully compatible with the linter/CI


keyvault/shim.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Shim provides convenience methods for working with KeyVault.
+type Shim struct {
+	sf secretFetcher
+}
+
+// NewShim constructs a Shim for a KeyVault instance located at the provided url. The azcore.TokenCredential will
+// only be used during method calls, it is not verified at initialization.
+func NewShim(vaultURL string, cred azcore.TokenCredential) (*Shim, error) {
+	c, err := azsecrets.NewClient(vaultURL, cred, nil)
```
timraymond, line 38: If you make `sf` public, you can eliminate this constructor and the dep on `azcore`. It might also let you test with the `keyvault_test` package as well since you'd no longer need access to private members.

keyvault/shim.go
```diff
@@ -0,0 +1,158 @@
...
+	pemContentType    = "application/x-pem-file"
+	pkcs12ContentType = "application/x-pkcs12"
+)
+
+type secretFetcher interface {
+	GetSecret(ctx context.Context, secretName string, opts *azsecrets.GetSecretOptions) (azsecrets.GetSecretResponse, error)
+}
+
+// Shim provides convenience methods for working with KeyVault.
+type Shim struct {
```
timraymond, line 31: Would `CertificateProvider` be closer to what this does?

keyvault/shim.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// GetLatestTLSCertificate fetches the latest version of a keyvault certificate and transforms it into a usable tls.Certificate.
+func (s *Shim) GetLatestTLSCertificate(ctx context.Context, certName string) (tls.Certificate, error) {
+	resp, err := s.sf.GetSecret(ctx, certName, nil)
+	if err != nil {
+		return tls.Certificate{}, errors.Wrap(err, "could not get secret")
+	}
+
+	pemBlocks, err := getPEMBlocks(*resp.Properties.ContentType, *resp.Value)
```
timraymond, line 53: Does PEM follow MIME to the letter? Like, could you end up with surprise things after the media type (e.g., as absurd as this would be: `application/x-pem-file; charset=utf-8`. There's a `mime.ParseMediaType` that does the right thing, if so.

keyvault/_example/main.go
```diff
@@ -0,0 +1,147 @@
...
+		logger.Fatal("could not get response", zap.String("host", addr), zap.Error(err))
+	}
+
+	printTLSConnState(resp.TLS)
+
+	bs, _ := io.ReadAll(resp.Body)
+	logger.Info("response from tls server", zap.String("body bytes", string(bs)))
+}
+
+func createClientTLSConfig(tlsCert tls.Certificate) (*tls.Config, error) {
```
timraymond, line 99: I think this would be nice to have pushed down into the `keyvault` package... usually when you go to create an `http.Server` with TLS enabled, you hit that `TLSConfig` section and have to look up a bunch of documentation on how to assemble the certs, the root CA, etc., etc. Ideally, it would be "just drop whatever gets returned from this into TLSConfig."

rbtr (dismissed):
lgtm

timraymond (commented):
Looks pretty good. Just have some questions and thoughts.

keyvault/shim.go
ramiro-gamarra: For keyvault-generated certs (our target use case), what I've observed are these specific content types, though [documentation about kv secrets is not totally clear](https://docs.microsoft.com/en-us/azure/key-vault/secrets/about-secrets).

keyvault/shim.go
ramiro-gamarra: Yeah, given the api surface that is probably the most apt description. Maybe it's a bit premature to think our keyvault usage will grow beyond retrieving certs, but my original thinking was along those lines: this type is a wrapper over the az keyvault sdk that we can extend to simplify keyvault operations as needed.

keyvault/shim.go
rbtr: I think a `New` is appropriate here; my benchmark is: if I incompletely initialize this struct will things break, that seems true here.

PR #1369, opened 2022-05-09
Alegal/add timeout to hns calls

aegal:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


network/hnswrapper/hnsv2wrapperwithtimeout.go
```diff
@@ -0,0 +1,344 @@
...
+			endpoint: endpoint,
+			Err:      err,
+		}
+	}()
+
+	// Listen on our channel AND a timeout channel - which ever happens first.
+	select {
+	case res := <-r:
+		return res.endpoint, res.Err
+	case <-time.After(3 * time.Second):
```
matmerr, line 66: alternatively use context.Context and cancel that context withTimeout

https://www.sohamkamani.com/golang/context-cancellation-and-values/#emitting-a-cancellation-event

network/hnswrapper/hnsv2wrapperwithtimeout.go
```diff
@@ -1,12 +1,28 @@
+// Copyright 2022 Microsoft. All rights reserved.
+// MIT License
```
rbtr, line 2: we have a global LICENSE in the root of the repo, we do not need license headers unless this file should fall under a different license.

network/hnswrapper/hnsv2wrapperwithtimeout.go
```diff
@@ -63,13 +80,15 @@ func (h Hnsv2wrapperwithtimeout) CreateEndpoint(endpoint *hcn.HostComputeEndpoin
 	select {
 	case res := <-r:
 		return res.endpoint, res.Err
-	case <-time.After(3 * time.Second):
-		return nil, fmt.Errorf("out of time :(")
+	case <-ctx.Done():
+		return nil, fmt.Errorf("CreateEndpoint %w , timeout value is %s seconds", ErrHNSCallTimeout, hnsCallTimeout.String())
```
rbtr, line 84: errors.Wrap/errors.Wrapf (everywhere, instead of fmt.Errorf)

rbtr (changes requested):

network/hnswrapper/hnsv2wrapperwithtimeout.go
```diff
@@ -0,0 +1,388 @@
...
+}
+
+type GetNetworkByIDFuncResult struct {
+	network *hcn.HostComputeNetwork
+	Err     error
+}
+
+func (h Hnsv2wrapperwithtimeout) CreateEndpoint(endpoint *hcn.HostComputeEndpoint) (*hcn.HostComputeEndpoint, error) {
+	r := make(chan CreateEndpointFuncResult)
+	ctx, cancel := context.WithTimeout(context.TODO(), hnsCallTimeout)
```
tamilmani1989, line 67: can hnsCallTimeout be part of Hnsv2wrapperwithtimeout structure? Also do you want to follow camelcase notation for this structure name?

network/hnswrapper/hnsv2wrapperwithtimeout.go
aegal: Doesn't %w introduced in go 1.13 replace those packages for wrapping the error? I am using %w here to wrap the error with the standard library, I know errors.Wrapf includes the stack trace, is that the extra benefit we are looking to achieve by leveraging this external package github.com/pkg/errors? 

Otherwise these two do the same thing but one uses the standard library.

network/hnswrapper/hnsv2wrapperwithtimeout.go
```diff
@@ -0,0 +1,388 @@
...
+	defer cancel()
+
+	go func() {
+		network, err := h.Hnsv2.CreateNetwork(network)
+
+		r <- CreateNetworkFuncResult{
+			network: network,
+			Err:     err,
+		}
+
```
tamilmani1989, line 118: extra space

tamilmani1989 (changes requested):
let also add a config  - one for disable/enable timeout and other for configuring timeout seconds

network/hnswrapper/hnsv2wrapperwithtimeout.go
aegal: If its just a matter of formatting, wouldn't it be more prudent to create some sort of common method to return errors in the format we want? This way we control the format and we use the standard library and folks don't have to think to much about this. 

I don't have a strong opinion on this and will go with wrapf, but slightly curious.

rbtr: > common method to return errors in the format we want

sure, we could call it something like `pkg/errors.Wrap`? 😄 
`pkg/errors` is commonly used, feature complete, stable, and not going anywhere, I don't see a reason to reimplement it ourselves

network/hnswrapper/hnsv2wrapperwithtimeout.go
aegal: updated

network/hnswrapper/hnsv2wrapperwithtimeout.go
aegal: removed

network/hnswrapper/hnsv2wrapperwithtimeout.go
aegal: For me, when I read the go official documentation mentioning %w as the way to wrap errors and in this repo, there is examples of doing that in that way - this leads to confusion for folks that are not aware we started using pkg/errors everywhere. Maybe having this as part of the linter would help? 

In any case, I updated it to use errors.wrapf.

network/hnswrapper/hnsv2wrapperwithtimeout.go
aegal: I updated it to make it part of Hnsv2wrapperwithtimeout struct

network/network_windows.go
```diff
@@ -55,6 +58,17 @@ func UseHnsV2(netNs string) (bool, error) {
 	return useHnsV2, err
 }
 
+// this hnsv2 variable is package level variable in network
+// we do this to avoid passing around os specific objects in platform agnostic code
+var hnsv2 hnswrapper.HnsV2WrapperInterface = hnswrapper.Hnsv2wrapper{}
+
+func enableHnsTimeout(timeoutValue int) {
+	if _, ok := hnsv2.(hnswrapper.Hnsv2wrapperwithtimeout); !ok {
+		var timeoutDuration = time.Duration(timeoutValue) * time.Second
```
aegal, line 67: should I return a error if timeoutvalue is zero or default to to a constant we define? @tamilmani1989 what do you think?

network/manager.go
```diff
@@ -143,6 +149,18 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
 		}
 	}
 
+	// Parsing the Stdin the CRI will pass to CNI
+	// this is to leverage it in Delete/Create network during the initial CNI rehydration
+	stdinData, err := io.ReadAll(os.Stdin)
```
aegal, line 154: @tamilmani1989 reading in stdin to recreate the CNI config here to ensure when we restore/rehydrate we can leverage the timeout , let me know if see any issues with that

cni/netconfig.go
```diff
@@ -80,7 +80,9 @@ type NetworkConfig struct {
 }
 
 type WindowsSettings struct {
-	EnableLoopbackDSR bool `json:"enableLoopbackDSR,omitempty"`
+	EnableLoopbackDSR         bool `json:"enableLoopbackDSR,omitempty"`
+	EnableHNSTimeout          bool `json:"enableHNSTimeout,omitempty"`
+	HnsTimeoutDurationSeconds int  `json:"hnsTimeoutDurationSeconds,omitempty"`
```
rbtr, line 85: make this a `time.Duration` so we can't use it incorrectly

cni/netconfig.go
```diff
@@ -80,7 +80,9 @@ type NetworkConfig struct {
 }
 
 type WindowsSettings struct {
-	EnableLoopbackDSR bool `json:"enableLoopbackDSR,omitempty"`
+	EnableLoopbackDSR         bool `json:"enableLoopbackDSR,omitempty"`
+	EnableHNSTimeout          bool `json:"enableHNSTimeout,omitempty"`
```
rbtr, line 84: we can treat the zero value of the timeout duration as "disable timeouts" - then we don't need two parameters for it

network/hnswrapper/hnsv2wrapperwithtimeout.go
```diff
@@ -0,0 +1,384 @@
...
+	"github.com/Microsoft/hcsshim/hcn"
+)
+
+// ErrHNSCallTimeout - hns call timeout
+var ErrHNSCallTimeout = errors.New("timed out calling hns")
+
+type Hnsv2wrapperwithtimeout struct {
+	Hnsv2 HnsV2WrapperInterface
+	// hnsCallTimeout indicates the time in seconds to wait for hns calls before timing out
```
rbtr, line 21: not necessarily "seconds"

rbtr (changes requested):

cni/netconfig.go
rbtr: can't be unmarshalled directly but check [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration)

network/manager.go
tamilmani1989: lets remove this code..its not required..rehydration will be always false for cni..also you cannot read stdin twice..it can be read only once

network/endpoint_windows.go
```diff
@@ -441,12 +443,15 @@ func (nw *network) newEndpointImplHnsV2(cli apipaClient, epInfo *EndpointInfo) (
...
 
 // deleteEndpointImpl deletes an existing endpoint from the network.
-func (nw *network) deleteEndpointImpl(_ netlink.NetlinkInterface, _ platform.ExecClient, ep *endpoint) error {
+func (nw *network) deleteEndpointImpl(_ netlink.NetlinkInterface, _ platform.ExecClient, ep *endpoint, cniConfig *cni.NetworkConfig) error {
 	if useHnsV2, err := UseHnsV2(ep.NetNs); useHnsV2 {
 		if err != nil {
 			return err
 		}
 
+		if cniConfig.WindowsSettings.EnableHNSTimeout == true {
```
tamilmani1989, line 452: let detect based on timeout seconds itself

tamilmani1989 (changes requested):

cni/netconfig.go
aegal: that's better, updated it to use that

network/manager.go
aegal: removed the code

network/network_windows.go
```diff
@@ -55,6 +58,26 @@ func UseHnsV2(netNs string) (bool, error) {
 	return useHnsV2, err
 }
 
+// this hnsv2 variable is package level variable in network
+// we do this to avoid passing around os specific objects in platform agnostic code
+var hnsv2 hnswrapper.HnsV2WrapperInterface = hnswrapper.Hnsv2wrapper{}
+
+var Hnsv1 hnswrapper.HnsV1WrapperInterface = hnswrapper.Hnsv1wrapper{}
```
aegal, line 65: Unfortunate that hnsv1 would need to exported here, its used by cni/network in the consecutive add

tamilmani1989: you can use hnsendpoitclient interface intialized in netplugin and use it as like before?

tamilmani1989 (changes requested):
also add timeout setting in windows conflist

network/network_windows.go
aegal: Closing this comment out per our discussion offline, to use hnsendpoitclient would require some non trivial refactor between cni/network and root network package.

cni/netconfig.go
aegal: Yeah, can't be unmarshalled directly. I'll keep it as an integer, easier to glean the format from the conflist and also easier to do the conditional check.

network/hnswrapper/hnsv2wrapperwithtimeout.go
aegal: updated the comment

network/endpoint_windows.go
aegal: updated

network/hnswrapper/hnsv1wrapperwithtimeout.go
```diff
@@ -0,0 +1,230 @@
...
+			HnsGlobals: hnsGlobals,
+			Err:        err,
+		}
+	}()
+
+	select {
+	case res := <-r:
+		return res.HnsGlobals, res.Err
+	case <-ctx.Done():
+		return nil, errors.Wrapf(ErrHNSCallTimeout, "IsHnsEndpointAttached timeout value is %v ", h.HnsCallTimeout.String())
```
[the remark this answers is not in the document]
aegal: done

network/manager.go
```diff
@@ -163,7 +162,7 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
 					for _, extIf := range nm.ExternalInterfaces {
 						for _, nw := range extIf.Networks {
 							log.Printf("[net] Deleting the network %s on reboot\n", nw.Id)
-							nm.deleteNetwork(nw.Id, &cni.NetworkConfig{})
+							nm.deleteNetwork(nw.Id, &cni.NetworkConfig{})  //nolint:errcheck // error is not needed to be checked in rehydration
```
rbtr, line 165: you should discard the error explicitly like `_ = rm.deleteNetwork` so that it's clear to future readers that this was intentional and not a mistake

cni/network/network_windows_test.go
```diff
@@ -19,6 +20,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func init() {
+	network.Hnsv2 = hnswrapper.NewHnsv2wrapperFake()
+	network.Hnsv1 = hnswrapper.Hnsv1wrapperfake{}
```
tamilmani1989, line 25: lets keep it consistent..either have New function like hnsv2 or remove function and intialize hnsv2 directly

cnm/network/network.go
```diff
@@ -173,7 +174,7 @@ func (plugin *netPlugin) createNetwork(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	err = plugin.nm.CreateNetwork(&nwInfo)
+	err = plugin.nm.CreateNetwork(&nwInfo, &cni.NetworkConfig{})
```
tamilmani1989, line 177: with platformInit change, i dont think its required to pass networkconfig...is that right? if so, can you please remove passing cni config from all places

tamilmani1989 (changes requested):

network/endpoint_windows.go
```diff
@@ -97,15 +65,24 @@ func ConstructEndpointID(containerID string, netNsPath string, ifName string) (s
...
+
 	if useHnsV2, err := UseHnsV2(epInfo.NetNsPath); useHnsV2 {
 		if err != nil {
 			return nil, err
 		}
 
+		if cniConfig.WindowsSettings.HnsTimeoutDurationInSeconds > 0 {
+			enableHnsV2Timeout(cniConfig.WindowsSettings.HnsTimeoutDurationInSeconds)
+		}
+
```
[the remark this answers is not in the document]
aegal: Thanks for this suggestion!

cni/network/network_windows.go
```diff
@@ -388,3 +388,11 @@ func getNATInfo(executionMode string, ncPrimaryIPIface interface{}, multitenancy
 
 	return natInfo
 }
+
+func platformInit(cniConfig *cni.NetworkConfig) {
+	if cniConfig.WindowsSettings.HnsTimeoutDurationInSeconds > 0 {
+		log.Printf(fmt.Sprintf("Enabling timeout for Hns calls with a timeout value of : %v", cniConfig.WindowsSettings.HnsTimeoutDurationInSeconds))
```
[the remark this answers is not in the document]
aegal: removed it

tamilmani1989 (approved):
lgtm :shipit:

PR #1372, opened 2022-05-10
Fix CNS logging

camrynl:
**Reason for Change**:

 OrchestratorContext cannot always be unmarshalled due to different orchestrator types, so updating CNS logging to exclude it

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -92,7 +92,7 @@ type CreateNetworkContainerRequest struct {
 
 // CreateNetworkContainerRequest implements fmt.Stringer for logging
 func (req *CreateNetworkContainerRequest) String() string {
-	return fmt.Sprintf("[Azure CNS] cns.CreateNetworkContainerRequest "+
+	return fmt.Sprintf("CreateNetworkContainerRequest"+
 		"{Version: %s, NetworkContainerType: %s, NetworkContainerid: %s, PrimaryInterfaceIdentifier: %s, "+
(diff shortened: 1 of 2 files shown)


rbtr (changes requested):
if we want to dump (only some) struct fields, maybe `cns.CreateNetworkContainerRequest` should implement [`fmt.Stringer`](https://pkg.go.dev/fmt#Stringer)?

tamilmani1989 (dismissed):
lgtm

cns/NetworkContainerContract.go
```diff
@@ -90,6 +90,16 @@ type CreateNetworkContainerRequest struct {
 	EndpointPolicies           []NetworkContainerRequestPolicies
 }
 
+// CreateNetworkContainerRequest implements fmt.Stringer for logging
+func (req *CreateNetworkContainerRequest) String() string {
+	return fmt.Sprintf("[Azure CNS] cns.CreateNetworkContainerRequest "+
```
rbtr, line 95: this log metadata shouldn't be in the String method, it should only return a string representation of the struct
```suggestion
	return fmt.Sprintf("CreateNetworkContainerRequest"+
```

rbtr (changes requested):

rbtr (approved):
thanks for cleaning this up 🚀

rbtr:
hey @camrynl you are welcome to merge. you should also usually have an "enable auto-merge" button so that you don't have to come back later after the CI and approvals are in 🙂

camrynl:
@rbtr noted, thanks!

PR #1374, opened 2022-05-11
fix/refactor: [NPM] eliminate struct fields to prevent errors & reduce memory

huntergregory:
- 1st commit: remove Name field from NPMNetworkPolicy because we should be using PolicyKey field instead
  - Windows Dataplane was using the wrong field in some scenarios
- 2nd commit: move PolicyID field in ACLPolicy to NPMNetworkPolicy since it's the same for all ACLs in a netpol

### DIFF
--- a/npm/pkg/dataplane/dataplane_windows.go
+++ b/npm/pkg/dataplane/dataplane_windows.go
@@ -213,6 +213,8 @@ func (dp *DataPlane) getSelectorIPsByPolicy(policy *policies.NPMNetworkPolicy) (
 		selectorIpSets[ipset.Metadata.GetPrefixName()] = struct{}{}
 	}
 
+	klog.Infof("policy %s has policy selector: %+v", policy.PolicyKey, selectorIpSets) // FIXME remove after debugging
+
 	return dp.ipsetMgr.GetIPsFromSelectorIPSets(selectorIpSets)
 }
 
@@ -238,9 +240,9 @@ func (dp *DataPlane) getEndpointsToApplyPolicy(policy *policies.NPMNetworkPolicy
 			continue
 		}
 		endpointList[ip] = endpoint.ID
-		// TODO make sure this is netpol key and not name
 		endpoint.NetPolReference[policy.PolicyKey] = struct{}{}
 	}
+	klog.Infof("[DataPlane] Endpoints to apply policy %s: %+v", policy.PolicyKey, endpointList) // FIXME remove after debugging
 	return endpointList, nil
 }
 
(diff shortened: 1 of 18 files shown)


npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -192,16 +192,16 @@ func (dp *DataPlane) updatePod(pod *updateNPMPod) error {
...
-		endpoint.NetPolReference[policyName] = struct{}{}
+		endpoint.NetPolReference[policyKey] = struct{}{}
 	}
 
 	return nil
 }
 
-func (dp *DataPlane) getSelectorIPsByPolicyName(policyName string) (map[string]struct{}, error) {
-	policy, ok := dp.policyMgr.GetPolicy(policyName)
+func (dp *DataPlane) getSelectorIPsByPolicyName(policyKey string) (map[string]struct{}, error) {
```
vakalapa, line 201: <nit> probably update this func name to getSelectorIPsByPolicyKey

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -278,7 +280,9 @@ func (dp *DataPlane) refreshAllPodEndpoints() error {
 		}
 
 		dp.endpointCache[ep.IP] = ep
+		klog.Infof("updating endpoint cache to include %s: %+v", ep.IP, ep) // FIXME remove after debugging
 	}
+	klog.Infof("endpoint cache after refreshing all pod endpoints: %+v", dp.endpointCache) // FIXME remove after debugging
```
vakalapa, line 285: remove these prints ?

vakalapa (commented):
minor comments

npm/pkg/dataplane/dataplane_windows.go
huntergregory: gonna fix the bug related to these print statements in a followup PR

npm/pkg/dataplane/dataplane_windows.go
huntergregory: let me fix the followup PR?

npm/pkg/dataplane/policies/policy_windows.go
```diff
@@ -27,6 +28,13 @@ var (
 	ErrProtocolNotSupported       = errors.New("Protocol mentioned is not supported")
 )
 
+// aclPolicyID returns azure-acl-<network policy namespace>-<network policy name> format
+// to differentiate ACLs among different network policies,
+// but aclPolicy in the same network policy has the same aclPolicyID.
+func aclPolicyID(policyNS, policyName string) string {
```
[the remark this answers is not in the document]
huntergregory: not adamant either, but prefer to let Linux return an empty string to keep in par with separate dirty cache etc. between OS's

vakalapa (approved):

PR #1375, opened 2022-05-13
Remove duplicate logs

tamilmani1989:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/network/network.go
```diff
@@ -754,8 +751,7 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
 	}
 
 	// Create the endpoint.
-	telemetry.SendCNIEvent(plugin.tb, fmt.Sprintf("[cni-net] Creating endpoint %+v.", epInfo))
-	log.Printf("[cni-net] Creating endpoint %v.", epInfo.Id)
+	telemetry.LogAndSendEvent(plugin.tb, fmt.Sprintf("[cni-net] Creating endpoint %+v.", epInfo.String()))
```
aegal, line 754: minor, shouldn't this be %s since we are getting epInfo.String()

aegal, line 754: minor, shouldn't this be %s since we are getting epInfo.String()

aegal (approved):
lgtm

cni/network/network.go
rbtr: yeah, %v is "make a string representation" so it works, but it should probably just be %s

aegal (dismissed):

network/network.go
```diff
@@ -89,6 +90,11 @@ type DNSInfo struct {
 	Options []string
 }
 
+func (nwInfo *NetworkInfo) String() string {
```
huntergregory, line 93: In NPM, we elected to use a method name different than `String()` to make investigations easier. If you print the NetworkInfo pointer or print it with "%v", then it will implicitly use the `String()` method, which might not have all the info in the struct.

https://github.com/Azure/azure-container-networking/pull/1235

Stringer interface:
https://towardsdev.com/type-stringer-golang-88d89777b959

PR #1383, opened 2022-05-16
fix: Start OVS Switchd if not running on Ubuntu 18

jaer-tsun:
**Reason for Change**:

Start OVS Switchd if not running on Ubuntu 18

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


ovsctl/ovsctl.go
```diff
@@ -55,7 +57,20 @@ type Ovsctl struct {
 }
 
 func NewOvsctl() Ovsctl {
-	return Ovsctl{execcli: platform.NewExecClient()}
+	execcli := platform.NewExecClient()
+	// try to ensure ovs-vswitchd is active (for Ubuntu 18), otherwise log failure to start the daemon
+	if output, err := execcli.ExecuteCommand("lsb_release -is"); err == nil && strings.Contains(output, "Ubuntu") {
```
rbtr, line 62: `lsb_release` may not be installed. is there a better way to do this than exec and `string.Contains`?

rbtr (changes requested):

rbtr:
is this really our responsibility to start a system service? it feels to me like a prerequisite, but not something that we should do ourselves.

ovsctl/ovsctl.go
jaer-tsun: parse /etc/os-release?

ovsctl/ovsctl.go
```diff
@@ -55,7 +57,20 @@ type Ovsctl struct {
...
+		if output, err = execcli.ExecuteCommand("lsb_release -rs"); err == nil {
+			if version, _ := strconv.ParseFloat(strings.TrimSpace(output), strconv.IntSize); int32(version) >= defaultUbuntuMajorVerion {
+				if _, err = execcli.ExecuteCommand("systemctl is-active ovs-vswitchd"); err != nil {
```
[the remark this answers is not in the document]
jaer-tsun: fixed

ovsctl/ovsctl.go
tamilmani1989: /etc/os-release is better

tamilmani1989:
> is this really our responsibility to start a system service? it feels to me like a prerequisite, but not something that we should do ourselves.

ya this should be started on bootup. in one of incidents, service was stopped for some reason and cni failing to run ovs commands. this fix makes sure cni recovers from it by starting service. Do you think any better way to address this?

ovsctl/ovsctl.go
```diff
@@ -55,7 +63,26 @@ type Ovsctl struct {
...
+
+						if _, err := conn.StartUnitContext(ctx, "ovs-vswitchd.service", "fail", nil); err != nil {
+							log.Printf("[ovs] Failed to start ovs-vswitchd err:%v", err)
+						}
+					}
+				}
+			}
+		}
+	}
+	return Ovsctl{execcli: execcli}
```
tamilmani1989, line 85: can you add field with started as true so that cni would not try get os version and start ovs service for each invocation of this api

ovsctl/ovsctl.go
```diff
@@ -55,7 +63,26 @@ type Ovsctl struct {
 }
 
 func NewOvsctl() Ovsctl {
-	return Ovsctl{execcli: platform.NewExecClient()}
+	execcli := platform.NewExecClient()
+	// try to ensure ovs-vswitchd is active (for Ubuntu 18), otherwise log failure to start the daemon
+	if output, err := execcli.ExecuteCommand("cat /etc/os-release"); err == nil {
+		if capture := osRE.FindStringSubmatch(output); len(capture) == 2 {
+			if fields := strings.Fields(capture[1]); fields[0] == "Ubuntu" {
```
tamilmani1989, line 70: instead of doing nested ifs, how about you check for error and log it and return?

tamilmani1989 (changes requested):

rbtr:
> > is this really our responsibility to start a system service? it feels to me like a prerequisite, but not something that we should do ourselves.
> 
> ya this should be started on bootup. in one of incidents, service was stopped for some reason and cni failing to run ovs commands. this fix makes sure cni recovers from it by starting service. Do you think any better way to address this?

I'm hesitant for us to start manipulating the state of the host for things that are not, strictly speaking, our responsibility.

PR #1385, opened 2022-05-17
fix: [NPM] fix windows intersection of IPs

huntergregory:
Used to never apply policies to endpoints if the policy selector involved a list set. Two changes to enable this:
1. fix calculation getting IPs in the intersection of pod selector IPSets
2. separate out member IPSets of list IPSets in the pod selector

Other change: fail the dataplane if a non "azure" network is specified or if IPPolicyMode is used, which is unimplemented.

### DIFF
--- a/npm/pkg/controlplane/translation/translatePolicy.go
+++ b/npm/pkg/controlplane/translation/translatePolicy.go
@@ -214,42 +214,46 @@ func ipBlockRule(policyName, ns string, direction policies.Direction, matchType 
 	return ipBlockIPSet, setInfo
 }
 
-// PodSelector translates podSelector of NetworkPolicyPeer field in networkpolicy object to translatedIPSet and SetInfo.
+// PodSelector translates podSelector of NetworkPolicyPeer field in networkpolicy object to translatedIPSets, children of translated IPSets, and SetInfo.
+// Children are members of a list-type IPSet.
 // This function is called only when the NetworkPolicyPeer has namespaceSelector field.
-func podSelector(policyKey string, matchType policies.MatchType, selector *metav1.LabelSelector) ([]*ipsets.TranslatedIPSet, []policies.SetInfo, error) {
+func podSelector(policyKey string, matchType policies.MatchType, selector *metav1.LabelSelector) (psSets, childPSSets []*ipsets.TranslatedIPSet, psList []policies.SetInfo, err error) {
 	podSelectors, err := parsePodSelector(policyKey, selector)
 	if err != nil {
-		return nil, nil, err
+		return
 	}
 	lenOfPodSelectors := len(podSelectors)
-	podSelectorIPSets := []*ipsets.TranslatedIPSet{}
-	podSelectorList := make([]policies.SetInfo, lenOfPodSelectors)
+	psSets = make([]*ipsets.TranslatedIPSet, 0)
+	childPSSets = make([]*ipsets.TranslatedIPSet, 0)
+	psList = make([]policies.SetInfo, lenOfPodSelectors)
 
 	for i := 0; i < lenOfPodSelectors; i++ {
 		ps := podSelectors[i]
-		podSelectorIPSets = append(podSelectorIPSets, ipsets.NewTranslatedIPSet(ps.setName, ps.setType, ps.members...))
+		psSets = append(psSets, ipsets.NewTranslatedIPSet(ps.setName, ps.setType, ps.members...))
+
 		// if value is nested value, create translatedIPSet with the nested value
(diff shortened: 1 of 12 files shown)


npm/pkg/dataplane/ipsets/ipsetmanager_windows.go
```diff
@@ -33,40 +36,42 @@ func (iMgr *IPSetManager) GetIPsFromSelectorIPSets(setList map[string]struct{})
...
-			}
+			ips = set.affiliatedIPs()
 			firstLoop = false
-		}
-		klog.Infof("set [%s] has ippodkey: %+v", set.Name, set.IPPodKey) // FIXME remove after debugging
-		setintersections, err = set.getSetIntersection(setintersections)
-		if err != nil {
-			return nil, err
+		} else {
+			set.intersectAffiliatedIPs()
```
vakalapa, line 61: you are passing in the existing ops list.

vakalapa (changes requested):

npm/pkg/dataplane/ipsets/ipsetmanager_windows.go
```diff
@@ -327,6 +329,11 @@ func getPolicyNetworkRequestMarshal(setPolicySettings map[string]*hcn.SetPolicyS
 		)
 	}
 
+	if len(policyNetworkRequest.Policies) == 0 {
```
vakalapa, line 332: why move this check down ?

huntergregory: Added the other check back, but this is a separate check. We can have 1+ setpolicysettings but have 0 policies if there are no setpolicies with the given policyType

npm/pkg/dataplane/ipsets/ipset.go
```diff
@@ -372,21 +372,49 @@ func (set *IPSet) hasMember(memberName string) bool {
...
-			fmt.Sprintf("[IPSet] Selector IPSet cannot be of type %s", set.Type.String()))
+// affiliatedIPs returns the IPs for a hash set or the IPs of the member sets for a list set
+// This method, intersectAffiliatedIPs, and GetSetContents are good examples of how the
+// ipset struct may have been better designed as an interface with hash and list implementations.
+// Not worth it to redesign though.
+func (set *IPSet) affiliatedIPs() map[string]struct{} {
+	if set.Kind == HashSet {
+		result := make(map[string]struct{}, len(set.IPPodKey))
+		for ip := range set.IPPodKey {
+			result[ip] = struct{}{}
```
matmerr, line 383: can we use the set.IPPodKey map directly ignore values instead of creating a map with the same keys?

npm/pkg/controlplane/translation/translatePolicy.go
```diff
@@ -214,42 +214,46 @@ func ipBlockRule(policyName, ns string, direction policies.Direction, matchType
 	return ipBlockIPSet, setInfo
 }
 
-// PodSelector translates podSelector of NetworkPolicyPeer field in networkpolicy object to translatedIPSet and SetInfo.
+// PodSelector translates podSelector of NetworkPolicyPeer field in networkpolicy object to translatedIPSets, children of translated IPSets, and SetInfo.
+// Children are members of a list-type IPSet.
 // This function is called only when the NetworkPolicyPeer has namespaceSelector field.
-func podSelector(policyKey string, matchType policies.MatchType, selector *metav1.LabelSelector) ([]*ipsets.TranslatedIPSet, []policies.SetInfo, error) {
+func podSelector(policyKey string, matchType policies.MatchType, selector *metav1.LabelSelector) (psSets, childPSSets []*ipsets.TranslatedIPSet, psList []policies.SetInfo, err error) {
```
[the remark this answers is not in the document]
rbtr: yeah please don't.
it suggests that when you're returning several of the same type, which starts to get to where you should be returning a tuple or more complex type imo, and i would prefer it suggest that

npm/pkg/dataplane/ipsets/ipset.go
huntergregory: this has to be a separate map because we'll be removing members from it based on IPPodKeys of other sets later

huntergregory: thinking about this more, there are 2 scenarios:
1. when determining which IPs to apply a new policy on, we use the full map
2. when updating the IPSets that a pod is a part of, we only check to see if the Pod IP is in the map

The second scenario doesn't require the full map and could use O(1) space, but I'm not sure if this a worthwhile optimization if we already do the O(N) space option for another scenario

matmerr (approved):

PR #1386, opened 2022-05-17
perf: [NPM] remove unnecessary computation in windows DP

huntergregory:
- Remove unnecessary code chunk where we convert member sets of lists to policies (see old line 202, which would never be reached)
  - not worth adding complexity if unnecessary
  - also slightly computation heavy: concatenate all members in each hash set
- Replace verbose log lines with lines that rarely occur (only occur when sets are first added to an endpoint)

### DIFF
--- a/npm/pkg/dataplane/ipsets/ipsetmanager_windows.go
+++ b/npm/pkg/dataplane/ipsets/ipsetmanager_windows.go
@@ -152,27 +152,27 @@ func (iMgr *IPSetManager) applyIPSets() error {
 // toDeleteSets:
 //      this function will loop through the dirty delete cache and adds existing set obj in HNS to toDeleteSets
 func (iMgr *IPSetManager) calculateNewSetPolicies(networkPolicies []hcn.NetworkPolicy) (*networkPolicyBuilder, error) {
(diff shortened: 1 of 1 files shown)


npm/pkg/dataplane/ipsets/ipsetmanager_windows.go
```diff
@@ -152,59 +152,45 @@ func (iMgr *IPSetManager) applyIPSets() error {
...
-		// TODO we should add members first and then the Lists
 		_, ok := existingSetNames[setName]
 		if ok {
 			setPolicyBuilder.toUpdateSets[setName] = setPol
 		} else {
+			setsToAdd = append(setsToAdd, setName)
 			setPolicyBuilder.toAddSets[setName] = setPol
 		}
-		if set.Kind == ListSet {
-			for _, memberSet := range set.MemberIPSets {
```
vakalapa, line 191: Just in case if we have a bug one level up in adding member sets to add updatecache, do you think it is worth it for us to compute a set of all the member ipsets and then check against the to update or to add sets and see if they are present ?

vakalapa (approved):
Approved with a single comment.

PR #1389, opened 2022-05-18
Bugfix/14337706/remove_ipam_state

ecigar13:
Fix: refactored hard coded ipam state file to a const
Fix: check for IPAM state presence when azure-vnet.json not present, if true then delete IPAM state.
Fix: changed error to uppercase to allow export.

**Reason for Change**:
Prevent CNI from falling into unrecoverable state: if kubelet restarts when container is running, azure-vnet.json not present, IPAM will assume pool exist and all ADD commands will fail.

- [ ] includes documentation
- [ ] adds unit tests


cni/ipam/ipam.go
```diff
@@ -168,6 +170,21 @@ func (plugin *ipamPlugin) Add(args *cniSkel.CmdArgs) error {
 		// Allocate an address pool.
 		poolID, subnet, err = plugin.am.RequestPool(nwCfg.Ipam.AddrSpace, "", "", options, isIpv6)
 		if err != nil {
+
+			if errors.Is(err, ipam.ErrNoAvailableAddressPools) || errors.Is(err, ipam.ErrAddressPoolNotFound) {
```
tamilmani1989, line 174: lets do only for ErrNoAvailableAddressPools

tamilmani1989 (changes requested):
i see you changed permission of scripts..can you please revert it?

PR #1392, opened 2022-05-19
Bugfix/14337706/remove state file

ecigar13:
**Reason for Change**:
fix: make user enter a file path to specify state file
fix: auto delete IPAM state to recover when CNI state file is missing and RefCount is not 0.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


cni/ipam/ipam.go
```diff
@@ -167,7 +169,24 @@ func (plugin *ipamPlugin) Add(args *cniSkel.CmdArgs) error {
 
 		// Allocate an address pool.
 		poolID, subnet, err = plugin.am.RequestPool(nwCfg.Ipam.AddrSpace, "", "", options, isIpv6)
+
 		if err != nil {
+			if errors.Is(err, ipam.ErrNoAvailableAddressPools) {
```
tamilmani1989, line 174: can you move this code to invoker_azure.go?

tamilmani1989 (changes requested):

cni/network/invoker_azure.go
```diff
@@ -25,7 +27,26 @@ type delegatePlugin interface {
 	Errorf(format string, args ...interface{}) *cniTypes.Error
 }
 
+// Create an instance of CNI, then call a delegate function for CNI action (ADD/DEL)
+// Avoid having IPAM state but no CNI state.
 func NewAzureIpamInvoker(plugin *NetPlugin, nwInfo *network.NetworkInfo) *AzureIPAMInvoker {
+	cniStateExists, err := platform.CheckIfFileExists(platform.CNIStateFilePath)
```
tamilmani1989, line 33: should do only if it gets no availabe address pool error from ipam

tamilmani1989 (changes requested):

cni/network/invoker_azure.go
```diff
@@ -57,20 +57,20 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 	if err != nil {
 
 		if errors.Is(err, ipam.ErrNoAvailableAddressPools) {
-			cniStateExists, err := platform.CheckIfFileExists(platform.CNIStateFilePath)
+			cniStateExists, er := platform.CheckIfFileExists(platform.CNIStateFilePath)
```
rbtr, line 60: why?

tamilmani1989 (changes requested):

cni/network/invoker_azure.go
ecigar13: I fixed it.

cni/network/invoker_azure.go
```diff
@@ -97,6 +122,24 @@ func (invoker *AzureIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkCo
 
 	if address == nil {
 		if err := invoker.plugin.DelegateDel(nwCfg.Ipam.Type, nwCfg); err != nil {
+			if errors.Is(err, ipam.ErrNoAvailableAddressPools) {
```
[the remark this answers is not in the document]
tamilmani1989: can you remove this from delete call?

cni/network/invoker_azure.go
```diff
@@ -48,7 +53,28 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
...
+		}
+
+		if !cniStateExists && ipamStateExists {
+			er = os.Remove(platform.CNIIpamStatePath)
+			if er != nil {
+				log.Printf("[cni] error deleting state file %v", er)
+			}
+		}
+
+		log.Printf("[cni] Deleted IPAM state file")
```
tamilmani1989, line 74: can you move this code to separate function?

cni/network/invoker_azure_test.go
```diff
@@ -132,12 +133,13 @@ func TestAzureIPAMInvoker_Add(t *testing.T) {
...
-		args    args
-		want    *cniTypesCurr.Result
-		want1   *cniTypesCurr.Result
-		wantErr bool
+		name        string
+		fields      fields
+		args        args
+		want        *cniTypesCurr.Result
+		want1       *cniTypesCurr.Result
+		wantErrType error
```
tamilmani1989, line 141: why did you add wantErrType..?

tamilmani1989 (changes requested):

cni/network/invoker_azure.go
```diff
@@ -112,6 +95,26 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 	return addResult, err
 }
 
+func (invoker *AzureIPAMInvoker) deleteIpamState() {
+	log.Printf("[cni] Deleting IPAM state file")
```
tamilmani1989, line 99: move this log to line 111 before os.remove call

tamilmani1989 (changes requested):

cni/network/invoker_azure_test.go
tamilmani1989: revert this change if not used

tamilmani1989 (approved):
lgtm 🚢

cni/network/invoker_azure.go
```diff
@@ -48,6 +53,10 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 
 	// Call into IPAM plugin to allocate an address pool for the network.
 	addResult.ipv4Result, err = invoker.plugin.DelegateAdd(addConfig.nwCfg.Ipam.Type, addConfig.nwCfg)
+
+	if err != nil && strings.Contains(err.Error(), ipam.ErrNoAvailableAddressPools.Error()) {
```
rbtr, line 57: is there an issue with doing this via `errors.Is/As`?

ipam/api.go
```diff
@@ -18,7 +18,7 @@ var (
 	errAddressPoolNotFound     = fmt.Errorf("Address pool not found")
 	errAddressPoolInUse        = fmt.Errorf("Address pool already in use")
 	errAddressPoolNotInUse     = fmt.Errorf("Address pool not in use")
-	errNoAvailableAddressPools = fmt.Errorf("No available address pools")
+	ErrNoAvailableAddressPools = fmt.Errorf("No available address pools")
```
rbtr, line 21: it's easier to find them when you put exported vars in a separate block

rbtr: because they might be wrapped or appear in logs, error strings should not be sentence case (and should not have closing punctuation)

store/json.go
```diff
@@ -42,9 +39,8 @@ type jsonFileStore struct {
 // NewJsonFileStore creates a new jsonFileStore object, accessed as a KeyValueStore.
 func NewJsonFileStore(fileName string, lockclient processlock.Interface) (KeyValueStore, error) {
 	if fileName == "" {
-		fileName = defaultFileName
+		return &jsonFileStore{}, errors.New("Need to pass in a json file path")
```
rbtr, line 42: no sentence case

cni/network/invoker_azure.go
```diff
@@ -86,6 +95,29 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
...
+		return
+	}
+
+	ipamStateExists, err := platform.CheckIfFileExists(platform.CNIIpamStatePath)
+	if err != nil {
+		log.Printf("[cni] Error checking IPAM state exist: %v", err)
+		return
+	}
+
+	if !cniStateExists && ipamStateExists {
```
rbtr, line 111: could drop this and return immediately if `cniStateExists` above?

rbtr: similarly could you return immediately if `!ipamStateExists`

tamilmani1989: agree on 1st case..2nd case would not happen

cni/network/invoker_azure.go
tamilmani1989: yes this error is from ipam and the ipam code unwrapping to string before writing to stdout and in cni side Error.Is couldn't find it

ecigar13: From CNI specification: https://www.cni.dev/docs/spec/#Error
CNI delegate actions will type erase the error to a JSON string, so `errors.Is/As` won't work. Only way to check error type is doing it inside IPAM. Outside delegated functions, I can use string contains, or parse into a JSON and also use string contains for the Msg field.

ipam/api.go
ecigar13: Do I change all other error strings to lower case, too? They're not part of this work item: 14337706.

rbtr: 😕
You did the right thing and matched the convention here. But it could be something to file as a future repair, and keep in mind when creating new error strings later.

PR #1399, opened 2022-05-23
fix: [NPM] manage policies on pod endpoints

huntergregory:
- maintain the netpol references of endpoints so we can delete netpols off the endpoint if needed
- don't need to update an endpoint if a pod update is called for a stale (deleted) pod
- garbage collect the endpoint cache

### DIFF
--- a/npm/pkg/dataplane/dataplane_windows.go
+++ b/npm/pkg/dataplane/dataplane_windows.go
@@ -162,7 +162,9 @@ func (dp *DataPlane) updatePod(pod *updateNPMPod) error {
 		}
 
 		for policyKey := range selectorReference {
-			toAddPolicies[policyKey] = struct{}{}
+			if _, ok := dp.pendingPolicies[policyKey]; !ok {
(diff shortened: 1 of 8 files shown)


npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -250,12 +263,15 @@ func (dp *DataPlane) getAllPodEndpoints() ([]hcn.HostComputeEndpoint, error) {
...
 
-// refreshAllPodEndpoints will refresh all the pod endpoints
+// refreshAllPodEndpoints will refresh all the pod endpoints and create empty netpol references for new endpoints
 func (dp *DataPlane) refreshAllPodEndpoints() error {
 	endpoints, err := dp.getAllPodEndpoints()
 	if err != nil {
 		return err
 	}
+
+	currentTime := time.Now().Unix()
```
matmerr, line 273: probably move this closer to usage to avoid time slip

matmerr (approved):

PR #1401, opened 2022-05-25
fix: NPM ipv6/CIDR validation

ck319:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/pkg/dataplane/ipsets/ipsetmanager.go
+++ b/npm/pkg/dataplane/ipsets/ipsetmanager.go
@@ -554,9 +554,15 @@ func validateIPSetMemberIP(ip string) bool {
 	// 192.168.0.0/24 nomatch
 	// always guaranteed to have ip, not guaranteed to have port + protocol
 	ipDetails := strings.Split(ip, ",")
+	isIPBlock := strings.Contains(ipDetails[0], "/")
+
 	if util.IsIPV4(ipDetails[0]) {
-		err := ValidateIPBlock(ipDetails[0])
-		return err == nil
+
+		if isIPBlock {
+			err := ValidateIPBlock(ipDetails[0])
+			return err == nil
+		}
+		return true
 	}
 	return false
 }
--- a/npm/pkg/dataplane/ipsets/ipsetmanager.go
+++ b/npm/pkg/dataplane/ipsets/ipsetmanager.go
@@ -555,9 +555,8 @@ func validateIPSetMemberIP(ip string) bool {
 	// always guaranteed to have ip, not guaranteed to have port + protocol
 	ipDetails := strings.Split(ip, ",")
 	if util.IsIPV4(ipDetails[0]) {
-		return true
+		err := ValidateIPBlock(ipDetails[0])
+		return err == nil
 	}
-
-	err := ValidateIPBlock(ip)
-	return err == nil
+	return false
 }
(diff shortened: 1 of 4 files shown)


npm/pkg/dataplane/ipsets/ipsetmanager.go
```diff
@@ -555,9 +555,8 @@ func validateIPSetMemberIP(ip string) bool {
 	// always guaranteed to have ip, not guaranteed to have port + protocol
 	ipDetails := strings.Split(ip, ",")
 	if util.IsIPV4(ipDetails[0]) {
-		return true
+		err := ValidateIPBlock(ipDetails[0])
```
huntergregory, line 558: should we validate the whole `ip` like we used to?

npm/pkg/dataplane/ipsets/ipsetmanager_test.go
```diff
@@ -1509,3 +1509,55 @@ func assertExpectedInfo(t *testing.T, iMgr *IPSetManager, info *expectedInfo) {
 		require.Equal(t, expectedNumEntries, numEntries, "numEntries mismatch for set %s", set.Name)
 	}
 }
+
+func TestValidateIPSetMemberIP(t *testing.T) {
```
huntergregory, line 1513: nit: let's put this code next to the other validate ip block tests

npm/pkg/dataplane/ipsets/ipsetmanager.go
ck319: I had to change this because it would fail in the case of tcp since that method splits on a space and tcp has a comma. That method only validates CIDR so sending it already split seemed to make sense in this case

npm/pkg/dataplane/ipsets/ipsetmanager_test.go
ck319: done

npm/util/util.go
```diff
@@ -355,10 +355,6 @@ func SliceToString(list []string) string {
...
 
 func IsIPV4(ip string) bool {
-	if net.ParseIP(ip).To4() != nil {
-		return true
-	}
-
-	_, _, err := net.ParseCIDR(ip)
-	return err == nil
+	ipOnly := strings.Split(ip, "/")
+	return net.ParseIP(ipOnly[0]).To4() != nil
```
rbtr, line 359: consider using the new [`netip`](https://pkg.go.dev/net/netip#Addr.Is4) package (here but really everywhere)

ck319: updated

npm/util/util.go
```diff
@@ -361,8 +361,8 @@ func IsIPV4(ip string) bool {
 	address, err := netip.ParseAddr(ipOnly[0])
 
 	if address.Is4() && isIPBlock {
-		_, _, err = net.ParseCIDR(ip)
-		return (err == nil && address.Is4())
+		_, _, err2 := net.ParseCIDR(ip)
+		return (err == nil && err2 == nil && address.Is4())
```
rbtr, line 365: check err where it is assigned and you don't have to avoid reassigning it and checking so many things here.

ck319: ok, updated

huntergregory (approved):

PR #1426, opened 2022-06-09
Adding TLS cert refreshing from KeyVault

ramiro-gamarra:
**Reason for Change**:

This PR adds a `CertRefresher` construct, which allows background pull of new certificates from KeyVault. We use this construct to provide a `tls.Certificate` in a dynamic manner to an HTTP server that uses a `tls.Config`.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/cns/service.go
+++ b/cns/service.go
@@ -4,7 +4,10 @@
 package cns
 
 import (
+	"context"
+	"crypto/tls"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -12,8 +15,11 @@ import (
 	"github.com/Azure/azure-container-networking/cns/common"
 	"github.com/Azure/azure-container-networking/cns/logger"
 	acn "github.com/Azure/azure-container-networking/common"
+	"github.com/Azure/azure-container-networking/keyvault"
 	"github.com/Azure/azure-container-networking/log"
+	localtls "github.com/Azure/azure-container-networking/server/tls"
 	"github.com/Azure/azure-container-networking/store"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/pkg/errors"
 )
 
@@ -68,19 +74,27 @@ func (service *Service) Initialize(config *common.ServiceConfig) error {
 		if err != nil {
 			return err
 		}
-		// Create the listener.
+
 		listener, err := acn.NewListener(u)
 		if err != nil {
 			return err
 		}
+
 		if config.TlsSettings.TLSPort != "" {
 			// listener.URL.Host will always be hostname:port, passed in to CNS via CNS command
 			// else it will default to localhost
 			// extract hostname and override tls port.
 			hostParts := strings.Split(listener.URL.Host, ":")
-			config.TlsSettings.TLSEndpoint = hostParts[0] + ":" + config.TlsSettings.TLSPort
+			tlsAddress := net.JoinHostPort(hostParts[0], config.TlsSettings.TLSPort)
+
 			// Start the listener and HTTP and HTTPS server.
-			if err = listener.StartTLS(config.ErrChan, config.TlsSettings); err != nil {
+			tlsConfig, err := getTLSConfig(config.TlsSettings, config.ErrChan)
+			if err != nil {
+				log.Printf("Failed to compose Tls Configuration with error: %+v", err)
+				return errors.Wrap(err, "could not get tls config")
+			}
+
+			if err := listener.StartTLS(config.ErrChan, tlsConfig, tlsAddress); err != nil {
 				return err
 			}
 		}
@@ -93,6 +107,89 @@ func (service *Service) Initialize(config *common.ServiceConfig) error {
 
 	log.Debugf("[Azure CNS] Successfully initialized a service with config: %+v", config)
 	return nil
+}
+
+func getTLSConfig(tlsSettings localtls.TlsSettings, errChan chan<- error) (*tls.Config, error) {
+	if tlsSettings.TLSCertificatePath != "" {
+		return getTLSConfigFromFile(tlsSettings)
+	}
+
+	if tlsSettings.KeyVaultURL != "" {
+		return getTLSConfigFromKeyVault(tlsSettings, errChan)
+	}
+
+	return nil, errors.Errorf("invalid tls settings: %+v", tlsSettings)
+}
+
+func getTLSConfigFromFile(tlsSettings localtls.TlsSettings) (*tls.Config, error) {
+	tlsCertRetriever, err := localtls.GetTlsCertificateRetriever(tlsSettings)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get certificate retriever")
+	}
+
+	leafCertificate, err := tlsCertRetriever.GetCertificate()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get certificate")
+	}
+
+	if leafCertificate == nil {
+		return nil, errors.New("certificate retrieval returned empty")
+	}
+
+	privateKey, err := tlsCertRetriever.GetPrivateKey()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get certificate private key")
+	}
+
+	tlsCert := tls.Certificate{
+		Certificate: [][]byte{leafCertificate.Raw},
+		PrivateKey:  privateKey,
+		Leaf:        leafCertificate,
+	}
+
+	tlsConfig := &tls.Config{
+		MaxVersion: tls.VersionTLS13,
+		MinVersion: tls.VersionTLS12,
+		Certificates: []tls.Certificate{
+			tlsCert,
+		},
+	}
+
+	return tlsConfig, nil
+}
+
+func getTLSConfigFromKeyVault(tlsSettings localtls.TlsSettings, errChan chan<- error) (*tls.Config, error) {
+	credOpts := azidentity.ManagedIdentityCredentialOptions{ID: azidentity.ResourceID(tlsSettings.MSIResourceID)}
+	cred, err := azidentity.NewManagedIdentityCredential(&credOpts)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not create managed identity credential")
+	}
+
+	kvs, err := keyvault.NewShim(tlsSettings.KeyVaultURL, cred)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not create new keyvault shim")
+	}
+
+	ctx := context.TODO()
+
+	cr, err := keyvault.NewCertRefresher(ctx, kvs, logger.Log, tlsSettings.KeyVaultCertificateName)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not create new cert refresher")
+	}
+
+	go func() {
+		errChan <- cr.Refresh(ctx, tlsSettings.KeyVaultCertificateRefreshInterval)
+	}()
+
+	tlsConfig := tls.Config{
+		MinVersion: tls.VersionTLS12,
+		MaxVersion: tls.VersionTLS13,
+		GetCertificate: func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			return cr.GetCertificate(), nil
+		},
+	}
+
+	return &tlsConfig, nil
 }
 
 func (service *Service) StartListener(config *common.ServiceConfig) error {
--- a/common/listener.go
+++ b/common/listener.go
@@ -6,26 +6,25 @@ package common
 import (
 	"crypto/tls"
 	"encoding/json"
-	"fmt"
 	"net"
 	"net/http"
 	"net/url"
 	"os"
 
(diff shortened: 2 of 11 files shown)


cns/service.go
```diff
@@ -68,19 +68,27 @@ func (service *Service) Initialize(config *common.ServiceConfig) error {
...
 			return err
 		}
+
 		if config.TlsSettings.TLSPort != "" {
 			// listener.URL.Host will always be hostname:port, passed in to CNS via CNS command
 			// else it will default to localhost
 			// extract hostname and override tls port.
 			hostParts := strings.Split(listener.URL.Host, ":")
-			config.TlsSettings.TLSEndpoint = hostParts[0] + ":" + config.TlsSettings.TLSPort
+			tlsAddress := hostParts[0] + ":" + config.TlsSettings.TLSPort
```
rbtr, line 82: it would be safer to use [JoinHostPort](https://pkg.go.dev/net#JoinHostPort), but I wish we were using the related net pkg tooling everywhere we deal with this stuff already.

keyvault/certrefresher.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, errors.Wrap(err, "could not fetch initial cert")
+	}
+
+	cf.cert = &cert
+	l.Info("initial certificate fetched", zap.Object("certificate", &cf))
+	return &cf, nil
+}
+
+// MarshalLogObject makes *CertRefresher a zapcore.ObjectMarshaler. Not concurrent-safe.
+func (c *CertRefresher) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
```
rbtr, line 58: 😍

rbtr (approved):
some suggestions but looks good already 🚀

keyvault/certrefresher.go
```diff
@@ -0,0 +1,125 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Wrap(ctx.Err(), "refresh canceled")
+		case <-ticker.C:
+			nextTick := time.Now().Add(interval) // approx
+
+			// todo: add retries for transient errors
```
timraymond, line 85: Should we combine the retry package in DNC with the one that I wrote for the NMA client? I could hoist that up the package tree and you'd be able to use it here... you'd just have to define what a transient error is.

Non-blocking comment though... we can do this in a separate PR if you think it's a good idea.

keyvault/certrefresher.go
```diff
@@ -0,0 +1,125 @@
...
+			if err := c.refresh(ctx); err != nil {
+				// best effort to determine if we should stop the refresh loop
+				certExpires := c.cert.Leaf.NotAfter
+				if certExpires.Before(nextTick) {
```
timraymond, line 89: I like how readable that is :)

keyvault/certrefresher.go
```diff
@@ -0,0 +1,125 @@
...
+					return &EventualExpirationErr{certExpires}
```
timraymond, line 90: Instead of failing, would it make sense to use `ticker.Reset()` to bump the refresh rate so you'll catch it in time?

common/listener.go
```diff
@@ -41,142 +45,188 @@ func NewListener(u *url.URL) (*Listener, error) {
...
+	l, _ := zap.NewProduction()
+	ctx := context.TODO()
+
+	cr, err := keyvault.NewCertRefresher(ctx, kvs, l, tlsSettings.KeyVaultCertificateName)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not create new cert refresher")
+	}
+
+	go func() {
+		errChan <- cr.Refresh(ctx, tlsSettings.KeyVaultCertificateRefreshInterval)
```
timraymond, line 119: I don't think this will ever return cleanly unless the certificate's expiration is before the next refresh interval. The context has no cancellation. That might be fine if it's intended to share the lifetime of the application, but I figured I'd mention it.

keyvault/certrefresher.go
```diff
@@ -0,0 +1,125 @@
...
+	GetLatestTLSCertificate(ctx context.Context, certName string) (tls.Certificate, error)
+}
+
+// CertRefresher offers a mechanism to present the latest version of a tls.Certificate from KeyVault, refreshed at an interval.
+type CertRefresher struct {
+	certName string
+	kvc      tlsCertFetcher
+	logger   *zap.Logger
+	m        sync.RWMutex
+	cert     *tls.Certificate
```
timraymond, line 35: It's a small thing, but I like to separate out the mutex and the thing (or things) it's protecting... I usually put the mutex below it.
```suggestion
	logger   *zap.Logger
	
	cert     *tls.Certificate
	m        sync.RWMutex
```

keyvault/certrefresher_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+func TestCertRefresher(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
timraymond, line 19: I have a little helper function that might be useful here:

```go
func testContext(t *testing.T) (context.Context, context.CancelFunc) {
	ctx := context.Background()
	if deadline, ok := t.Deadline(); ok {
		// a timeout was set, so use that as the context's deadline
		return context.WithDeadline(ctx, deadline)
	}
	return context.WithCancel(ctx)
}
```

So the usage here would look like:

```go
ctx, cancel := testContext(t)
defer cancel()
```

It's similar, but it'll use the deadline set in the `*testing.T` if a `-timeout` flag was provided.

keyvault/certrefresher.go
aegal: +1, I was thinking the same thing.

aegal (dismissed):
this looks good

common/listener.go
ramiro-gamarra: Yeah, ideally, context would be propagated all the way from `main` to handle signals cleanly but there are a few layers of indirection unfortunately, so I opted for a `context.TODO()` in the meantime. This goroutine is indeed tied to the lifetime of the application so a temporary leak _might_ be ok for now. Since I have some changes to push, I may try to propagate context, but will depend on size of changes.

keyvault/certrefresher.go
ramiro-gamarra: I see we already use `avast/retry-go` in this repo so I think I'll use that for now. Maybe we should see if we can accomplish what we require in `nmagent` and DNC using that package? Would be nice to standardize retries across the board.

cns/logger/log.go
```diff
@@ -22,113 +24,90 @@ var (
...
-}
 
-// Initialize CNS Logger
-func InitLogger(fileName string, logLevel, logTarget int, logDir string) {
-	Log = &CNSLogger{
-		logger: log.NewLogger(fileName, logLevel, logTarget, logDir),
-	}
+	m            sync.RWMutex
+	Orchestrator string
+	NodeID       string
```
ramiro-gamarra, line 33: motivation for changes in this file are making this cns logger easier to stub out in tests, as well as addressing what I'm pretty sure is a race condition when setting these fields, since the logging methods here read them all the time but they are set, possibly multiple times, from an http handler.

keyvault/certrefresher.go
```diff
@@ -0,0 +1,130 @@
...
+	retryFn := func() (err error) {
+		latestCert, err = c.kvc.GetLatestTLSCertificate(ctx, c.certName)
+		if err != nil {
+			c.logger.Errorf("could not fetch latest tls certificate: %v. retrying...", err)
+			return errors.Wrap(err, "could not fetch latest tls certificate")
+		}
+		return nil
+	}
+
+	if err := retry.Do(retryFn, retry.Context(ctx), retry.Delay(time.Second), retry.DelayType(retry.FixedDelay)); err != nil {
```
ramiro-gamarra, line 108: changed behavior a bit from latest revision. the intent is now to retry to fetch from keyvault until a deadline set to the current cert expiration date is met. basically, we try to keep cns alive for as long as the cert is valid.

ramiro-gamarra (commented):
Added some context to a few of the changes

keyvault/certrefresher_test.go
ramiro-gamarra: added this, it's nice convenience function for any test, so added a todo to find a better home for it in the future

keyvault/certrefresher.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (e *EventualExpirationErr) Error() string {
+	return fmt.Sprintf("could not refresh before expiration on %s", e.Time.String())
+}
+
+type tlsCertFetcher interface {
+	GetLatestTLSCertificate(ctx context.Context, certName string) (tls.Certificate, error)
+}
+
+type logger interface {
```
ramiro-gamarra, line 28: I switched from `zap` to existing logging mechanism to keep any possible side effects from this PR to a minimum. I enjoyed adding logs with `zap` so when given a chance I plan to revisit this package (or CNS in general) to contribute to its adoption.

cns/logger/log.go
```diff
@@ -22,113 +24,90 @@ var (
...
-	Orchestrator         string
-	NodeID               string
 	DisableTraceLogging  bool
 	DisableMetricLogging bool
 	DisableEventLogging  bool
-}
 
-// Initialize CNS Logger
-func InitLogger(fileName string, logLevel, logTarget int, logDir string) {
-	Log = &CNSLogger{
```
timraymond, line 34: ❤️

rbtr: ditto, but maybe fixing the logger should be a separate set of changes (that go first, if necessary)?

cns/logger/log.go
```diff
@@ -178,16 +156,97 @@ func ResponseEx(tag string, request interface{}, response interface{}, returnCod
...
+func (c *CNSLogger) sendTraceInternal(msg string) {
+	orch, nodeID := c.getOrchestratorAndNodeID()
+
+	report := aitelemetry.Report{
+		Message: msg,
+		Context: nodeID,
+		CustomDimensions: map[string]string{
+			OrchestratorTypeStr: orch,
+			NodeIDStr:           nodeID,
+		},
```
timraymond, line 178: Is it worth storing a map as just a general bag of keys and values like many structured logging tools do? We can set "orchestrator" and "nodeID" by default in there, but have the flexibility to add more as necessary. Unsure what the limitations of AI are though, so it may not be feasible :)

timraymond (commented):
I think this looks good, but I'll defer to someone with more familiarity in CNS for approval. I've got just a couple of questions / suggestions.

keyvault/certrefresher_test.go
timraymond: @ramiro-gamarra  Yeah, we'll have to chat about a good home for it in DNC as well, since I made a similar suggestion on a PR over there :)

cns/logger/log.go
```diff
@@ -178,16 +156,97 @@ func ResponseEx(tag string, request interface{}, response interface{}, returnCod
...
 
-// Send AI telemetry metric
-func SendMetric(metric aitelemetry.Metric) {
-	if Log.th == nil || Log.DisableMetricLogging {
+func (c *CNSLogger) getOrchestratorAndNodeID() (orch, nodeID string) {
+	c.m.RLock()
+	orch, nodeID = c.Orchestrator, c.NodeID
+	c.m.RUnlock()
+	return
+}
```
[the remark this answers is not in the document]
ramiro-gamarra: the linter actually suggested this. I kind of agree, only when there are multiple returns of the same type.

cns/logger/log.go
ramiro-gamarra: If I understand correctly, is the suggestion to add a map of default custom dimensions to CNSLogger? I would agree, and we do something similar in DNC. In this case, these properties are lazily added; there will be a time during CNS runtime when they haven't been set by DNC yet, so they are not quite "default". Ideally, once set they should never change, but I don't want to change behavior too much this time around.

cns/logger/log.go
```diff
@@ -178,16 +156,97 @@ func ResponseEx(tag string, request interface{}, response interface{}, returnCod
...
+	Log.Response(tag, response, returnCode, err)
+}
+
+func ResponseEx(tag string, request, response interface{}, returnCode types.ResponseCode, err error) {
+	Log.ResponseEx(tag, request, response, returnCode, err)
+}
+
+func SendMetric(metric aitelemetry.Metric) {
+	Log.SendMetric(metric)
 }
```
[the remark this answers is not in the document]
ramiro-gamarra: Good idea. Will address in the PR I'm extracting

cns/logger/log.go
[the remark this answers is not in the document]
timraymond: Yeah, I can see the names would be useful in the signature. I think in this case I would name them, but still explicitly return. Edited the suggestion to show what I mean.

rbtr (approved):
lgtm

PR #1429, opened 2022-06-15
Refactoring CNSLogger

ramiro-gamarra:
**Reason for Change**:

This PR refactors CNSLogger to make it easier for unit tests to stub it out. It is currently used as a singleton, manipulated with package level functions, which is not ideal.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/logger/cnslogger.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (c *CNSLogger) SetContextDetails(orchestrator, nodeID string) {
+	c.logger.Logf("SetContext details called with: %v orchestrator nodeID %v", orchestrator, nodeID)
+	c.m.Lock()
+	c.Orchestrator = orchestrator
+	c.NodeID = nodeID
+	c.m.Unlock()
+}
+
+func (c *CNSLogger) Printf(format string, args ...any) {
```
rbtr, line 66: `any` 😍

cns/logger/cnslogger.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (c *CNSLogger) getOrchestratorAndNodeID() (orch, nodeID string) {
+	c.m.RLock()
+	orch, nodeID = c.Orchestrator, c.NodeID
+	c.m.RUnlock()
+	return
+}
+
+func (c *CNSLogger) sendTraceInternal(msg string) {
+	orch, nodeID := c.getOrchestratorAndNodeID()
```
rbtr, line 164: is this still racy, even with the mutex? I imagine we could lock, update, unlock, in two threads and when the first actually logs it will use the later value.
since that is current state, it might be out of scope to fix here...something that will be fixed as a side-effect of structured logging.

rbtr (approved):

cns/logger/cnslogger.go
ramiro-gamarra: yeah, I suppose this fixes the data race, not the possible race condition. I mentioned in a comment on the keyvault pr, I think the behavior really should be to lazily initialize these 2 variables (orchestrator and node id shouldn't become different for any CNS instance once set, is my understanding), but leaving that for a later PR.

PR #1430, opened 2022-06-15
Remove azure-vnet-telemetry for windows multitenancy

tamilmani1989:
…vice for windows multitenancy will be started from cns.

**Reason for Change**:

This PR removes azure-vnet-telemetry binary from windows multitenancy package. Instead, added option in cns to start telemetry service which will receive events from CNI. For windows multitenancy scenario, customer needs to start cns with this option "-ts". eg: ./azure-cns.exe -ts

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -376,6 +376,14 @@ func sendRegisterNodeRequest(httpc *http.Client, httpRestService cns.HTTPService
 }
 
 func startTelemetryService(ctx context.Context) {
+	var config aitelemetry.AIConfig
+
+	err := telemetry.CreateAITelemetryHandle(config, false, false, false)
+	if err != nil {
+		log.Errorf("AI telemetry handle creation failed..:%w", err)
+		return
+	}
+
 	tbtemp := telemetry.NewTelemetryBuffer()
 	tbtemp.Cleanup(telemetry.FdName)
 	tb := telemetry.NewTelemetryBuffer()
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -385,9 +385,15 @@ func startTelemetryService(ctx context.Context) {
 	}
 
 	tbtemp := telemetry.NewTelemetryBuffer()
+	//nolint:errcheck // best effort to cleanup leaked pipe/socket before start
 	tbtemp.Cleanup(telemetry.FdName)
+
 	tb := telemetry.NewTelemetryBuffer()
-	tb.StartServer()
+	err = tb.StartServer()
+	if err != nil {
+		log.Errorf("Telemetry service failed to start: %w", err)
+		return
+	}
 	tb.PushData(rootCtx)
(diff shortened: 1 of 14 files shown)


telemetry/telemetrybuffer.go
```diff
@@ -181,6 +182,9 @@ func (tb *TelemetryBuffer) PushData() {
 		case <-tb.cancel:
 			log.Logf("[Telemetry] server cancel event")
 			goto EXIT
+		case <-ctx.Done():
+			log.Logf("[Telemetry] received context done event")
+			goto EXIT
```
rbtr, line 187: using `goto` is unusual and makes this harder to understand for me - can we just `break`? it seems like that would be equivalent.

cns/service/main.go
```diff
@@ -475,6 +506,10 @@ func main() {
 		logger.InitAI(aiConfig, ts.DisableTrace, ts.DisableMetric, ts.DisableEvent)
 	}
 
+	if telemetryDaemonEnabled {
+		go startTelemetryService(rootCtx)
```
ashvindeodhar, line 510: since we are not returning any error, failure to start telemetry service will go undetected. Should we exit the CNS if we can't start telemetry service? It will provide early failure. Otherwise we may not have telemetry data to debug an issue w/o telemetry

telemetry/telemetrybuffer.go
```diff
@@ -171,20 +170,23 @@ func (tb *TelemetryBuffer) Connect() error {
 }
 
 // PushData - PushData running an instance if it isn't already being run elsewhere
-func (tb *TelemetryBuffer) PushData() {
-	for {
+func (tb *TelemetryBuffer) PushData(ctx context.Context) {
+	loop := true
+	for loop {
```
ashvindeodhar, line 175: just curious to understand if goto is not the correct way.

rbtr: `goto` in general is confusing, it isn't commonly seen in `go` because there tends to be a clearer way to accomplish the same effect. 
here, instead of `goto EXIT`, or using this `loop` variable, we could:
`defer tb.Close()` - this is the only code in this func outside of the for/select, and we want it to execute after we break the `for`, 
so then we `return` out of those select branches which should terminate the loop.

cns/service/main.go
tamilmani1989: hmm thats more aggresive imo. also i dont want to change existing behaviour. this telemetry service is intended only for cni and cns telemetry not dependent on this.. if we couldn't get cni logs in worst case, we can ask cx to fetch logs.. even if telemetry service is not started, cns is functional, so i dont think exit would be good option here

ashvindeodhar (approved):

PR #1432, opened 2022-06-16
Delegated Azure IPAM

nddq:
This is the draft PR for the delegated Azure IPAM


rbtr:
@wedaly fyi

azure-ipam/ipam.go
```diff
@@ -0,0 +1,163 @@
...
+	versionedCniResult, err := cniResult.GetAsVersion(nwCfg.CNIVersion)
+	if err != nil {
+		p.logger.Error("Failed to interpret CNI result with netconf CNI version",
+			zap.Error(err),
+			zap.Any("cniVersion", nwCfg.CNIVersion),
+		)
+		return errors.Wrapf(err, "failed to interpret CNI result as version %s", nwCfg.CNIVersion)
+	}
+
+	versionedCniResult.Print()
```
tamilmani1989, line 127: log add call completed with pod details?

azure-ipam/cnsutils.go
```diff
@@ -0,0 +1,108 @@
...
+	podCfg := K8SPodEnvArgs{}
+	err := cniTypes.LoadArgs(args, &podCfg)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse pod config from stdin")
+	}
+	return &podCfg, nil
+}
+
+func getEndpointID(containerID, ifName string) string {
+	const minContainerLength = 8
```
wedaly, line 82: little confused that this is called "min" -- seems like we're truncating any ID *longer* than this, which means it's a max?

azure-ipam/ipam.go
```diff
@@ -0,0 +1,163 @@
...
+// CmdDel handles CNI delete commands.
+func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
+	p.logger.Info("DEL called")
+	// Create CNS request from args
+	req, err := createCNSRequest(args)
+	if err != nil {
+		p.logger.Error("Failed to create CNS IP config request",
+			zap.Error(err),
+		)
+		return errors.Wrapf(err, "failed to create CNS IP config request")
```
tamilmani1989, line 141: should we consider throwing retriable error here as well?

azure-ipam/cnsutils.go
```diff
@@ -0,0 +1,108 @@
...
+		return ""
+	}
+
+	infraEpName := containerID + "-" + ifName
+
+	return infraEpName
+}
+
+// Parse network config from given byte array
+func parseNetConf(b []byte) (*cniTypes.NetConf, error) {
```
wedaly, line 96: this isn't specific to CNS, so maybe it should be in `ipam.go` rather than `cnsutils.go`?

azure-ipam/ipam.go
```diff
@@ -0,0 +1,163 @@
...
+	ctx := context.TODO()
+	p.logger.Info("Making request to CNS")
+	if err := p.cnsClient.ReleaseIPAddress(ctx, req); err != nil {
+		p.logger.Error("Failed to release IP address from CNS",
+			zap.Error(err),
+			zap.Any("request", req),
+		)
+		return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "")
+	}
+
```
tamilmani1989, line 155: log for del completed?

azure-ipam/main.go
```diff
@@ -0,0 +1,60 @@
...
+		log.Printf("error executing azure-ipam plugin: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func executePlugin() error {
+	zapConfig := []byte(`{
+		"level": "debug",
+		"encoding": "json",
+		"outputPaths": ["/tmp/logs"],
```
tamilmani1989, line 27: isn't /tmp/logs not persistent? do we need to log in /var/logs/ instead?

azure-ipam/cnsutils.go
```diff
@@ -0,0 +1,108 @@
...
+func getEndpointID(containerID, ifName string) string {
+	const minContainerLength = 8
+	if len(containerID) > minContainerLength {
+		containerID = containerID[:8]
+	} else {
+		log.Printf("Container ID length is not greater than 8: %v", containerID)
+		return ""
+	}
+
+	infraEpName := containerID + "-" + ifName
```
wedaly, line 90: should check with @tamilmani1989 whether we still want ifName here

azure-ipam/ipam_test.go
```diff
@@ -0,0 +1,303 @@
...
+				req.Error(err)
+			} else {
+				req.NoError(err)
+			}
+		})
+	}
+}
+
+func TestCmdCheck(t *testing.T) {
+}
```
wedaly, line 303: should call the plugin here and expect no error

azure-ipam/ipam_test.go
```diff
@@ -0,0 +1,303 @@
...
+			fmt.Printf("test : %v\n", tt.name)
+			mockCNSClient := &MockCNSClient{}
+			logger, err := zap.NewProduction()
+			defer logger.Sync() // nolint
+			if err != nil {
+				return
+			}
+			ipamPlugin, _ := NewPlugin(logger, mockCNSClient)
+			err = ipamPlugin.CmdDel(tt.args)
+			fmt.Println()
```
wedaly, line 292: leftover from debugging?

azure-ipam/ipam_test.go
```diff
@@ -0,0 +1,303 @@
...
+)
+
+type scenario struct {
+	name    string
+	args    *cniSkel.CmdArgs
+	wantErr bool
+}
+
+func TestCmdAdd(t *testing.T) {
+	req := require.New(t)
```
wedaly, line 103: this is a little confusing (initially thought "req" was an HTTP request, not "require"). Any reason not to just call `require.NoError()` and `require.Error()` directly? (that's usually how I see this library used)

azure-ipam/ipam_test.go
```diff
@@ -0,0 +1,303 @@
...
+			wantErr: true,
+		},
+		{
+			name:    "Fail process CNS response during CmdAdd",
+			args:    failProcessCNSResp,
+			wantErr: true,
+		},
+		{
+			name:    "Fail parse netconf during CmdAdd",
+			args:    failParseNetConf,
```
wedaly, line 196: nit: I'm finding this a bit hard to read, because the construction of these args happens in different parts of the file. Consider changing this to a function that does all the construction in one place?

Imagining something like:
```
buildArgs: func() *cniSkel.CmdArgs {
   // do all the setup here...
}
```

azure-ipam/ipam_test.go
```diff
@@ -0,0 +1,303 @@
...
+				return
+			}
+			ipamPlugin, _ := NewPlugin(logger, mockCNSClient)
+			err = ipamPlugin.CmdAdd(tt.args)
+			fmt.Println()
+			if tt.wantErr {
+				req.Error(err)
+			} else {
+				req.NoError(err)
+			}
```
wedaly, line 222: could we also validate the output? That's probably the most important thing -- need to make sure the IPs that CNS returns make it to the upper plugin.

I'm not totally sure how we can capture the output, but we might have examples of how we tested the other plugins?

azure-ipam/ipam_test.go
```diff
@@ -0,0 +1,303 @@
...
+		StdinData:   invalidNetConf,
+	}
+
+	failGetVersionedResult := &cniSkel.CmdArgs{
+		ContainerID: "failGetVersionedResult",
+		Netns:       "testnetns",
+		IfName:      "testifname",
+		Args:        happyPodArgs,
+		StdinData:   invalidVersionNetConfByteArr,
+	}
```
wedaly, line 171: do we need to add any scenarios for network plugin = overlay?

azure-ipam/ipam.go
```diff
@@ -0,0 +1,163 @@
...
+// IPAMPlugin is the struct for the delegated azure-ipam plugin
+type IPAMPlugin struct {
+	Name      string
+	Version   string
+	Options   map[string]interface{}
+	logger    *zap.Logger
+	cnsClient cnsClient
+}
+
+type cnsClient interface {
```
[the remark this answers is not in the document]
rbtr: I don't like the idea of interfaces being used across packages (and here, modules). I prefer to scope the interface to the usage (and declare it there). That makes it easy to mock or swap impls, since then my code is only using the interface that I have defined around just the behavior that I use.

azure-ipam/const.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+import "time"
+
+const (
+	pluginName    = "azure-ipam"
```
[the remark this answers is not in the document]
rbtr: Is our goal that this be the single ipam plugin, or will they coexist forever? If it's going to be the only one, I think `azure-ipam` is better

azure-ipam/ipam.go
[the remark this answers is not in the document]
wedaly: fair points 👍

azure-ipam/main.go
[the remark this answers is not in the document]
rbtr: This might be an okay default, which we usually expect to get overridden by an explicit config option. I think this yaml []byte is unnecessary and we should be putting the logger/config together directly, though.

azure-ipam/ipam_test.go
nddq: can you elaborate on this?

azure-ipam/cnsutils.go
```diff
@@ -0,0 +1,108 @@
+package main
```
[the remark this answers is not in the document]
nddq: would swift-ipam works? I think the name got thrown around in meetings :)

azure-ipam/cnsutils.go
```diff
@@ -0,0 +1,108 @@
...
+	return req, nil
+}
+
+func processCNSResponse(resp *cns.IPConfigResponse) (*net.IPNet, net.IP, error) {
+	podCIDR := fmt.Sprintf(
+		"%s/%d",
+		resp.PodIpInfo.PodIPConfig.IPAddress,
+		resp.PodIpInfo.NetworkContainerPrimaryIPConfig.IPSubnet.PrefixLength,
+	)
+	podIP, podIPNet, err := net.ParseCIDR(podCIDR)
```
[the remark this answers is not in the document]
nddq: what's the difference between the two?

azure-ipam/cnsutils.go
nddq: you're right :)

azure-ipam/cnsutils.go
```diff
@@ -0,0 +1,108 @@
...
+// Parse network config from given byte array
+func parseNetConf(b []byte) (*cniTypes.NetConf, error) {
+	netConf := &cniTypes.NetConf{}
+	err := json.Unmarshal(b, netConf)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal net conf")
+	}
+
+	if netConf.CNIVersion == "" {
+		netConf.CNIVersion = "0.2.0"
```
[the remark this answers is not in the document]
nddq: This is the default version that the current azure-vnet fall back to when there is no version specified in the network config file. I don't know if this is the case for other CNI like Cilium though.

azure-ipam/cnsutils.go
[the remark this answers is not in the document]
tamilmani1989: lets keep it delegated-ipam.. if we want to merge v1ipam tomorrow, then it wouldn't make sense to have swift-ipam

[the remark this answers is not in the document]
rbtr: i'll suggest again that if we're going to converge to a single ipam, it should just be `azure-ipam`

azure-ipam/cnsutils.go
tamilmani1989: @nddq  should just return containerid here without truncating it

azure-ipam/ipam.go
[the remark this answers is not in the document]
nddq: ^

azure-ipam/ipam.go
```diff
@@ -0,0 +1,163 @@
...
+		)
+		return errors.Wrapf(err, "failed to create CNS IP config request")
+	}
+	p.logger.Info("Created CNS IP config request",
+		zap.Any("request", req),
+	)
+
+	// cnsClient sets a request timeout.
+	ctx := context.TODO()
+	p.logger.Info("Making request to CNS")
```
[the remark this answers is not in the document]
nddq: can you elaborate more on the what do you mean by a debug() function? Is that related to the log level set in the config?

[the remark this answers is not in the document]
tamilmani1989: does p.logger have Debug function like Info..if thats the case, we can some of the verbose logs to p.logger.Debug

azure-ipam/main.go
```diff
@@ -0,0 +1,60 @@
...
+	if err := json.Unmarshal(zapConfig, &cfg); err != nil {
+		return errors.Wrapf(err, "failed to unmarshal zap config")
+	}
+
+	logger, err := cfg.Build()
+	defer logger.Sync() // nolint
+	if err != nil {
+		return errors.Wrapf(err, "failed to setup IPAM logging")
+	}
+	logger.Info("logger construction succeeded")
```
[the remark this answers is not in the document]
nddq: It does make sense to move the logger construction to a helper function. Can you clarify what do you mean by the helper function to return a cleanup function?

azure-ipam/cnsutils.go
```diff
@@ -0,0 +1,108 @@
...
+		resp.PodIpInfo.NetworkContainerPrimaryIPConfig.IPSubnet.PrefixLength,
+	)
+	podIP, podIPNet, err := net.ParseCIDR(podCIDR)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "cns returned invalid pod CIDR %q", podCIDR)
+	}
+
+	resultIPNet := &net.IPNet{
+		IP:   podIP,
+		Mask: podIPNet.Mask,
```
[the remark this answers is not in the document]
nddq: would switching to package `netip` be beneficial here if we have to parse them back to `net.IP` anyway when we are populating the result afterward?

azure-ipam/ipam_test.go
wedaly: Here's an example of how to capture and verify the output (from the CNI reference plugins): https://github.com/containernetworking/plugins/blob/9c59728d39b729558bb712c79e2c80acc0864191/pkg/testutils/cmd.go#L42-L47

I think you can probably use the testutils package directly (it's part of the same library we're already using to invoke the delegated plugin)

Really think we should test this, because if something gets output incorrectly then the plugin won't work.

azure-ipam/logger.go
```diff
@@ -0,0 +1,56 @@
...
+		MessageKey:  "msg",
+		LevelKey:    "level",
+		EncodeLevel: zapcore.LowercaseLevelEncoder,
+	}
+
+	logger, err := loggerCfg.Build()
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to build zap logger")
+	}
+	return logger, CleanUpLog, nil
```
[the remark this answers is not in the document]
nddq: That's look so much cleaner! I will make the change

azure-ipam/main.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+func executePlugin() error {
+	logger, cleanup, err := NewLogger()
+	if err != nil {
+		return errors.Wrapf(err, "failed to setup IPAM logging")
+	}
+	logger.Debug("logger construction succeeded")
+	defer cleanup(logger) // nolint
```
rbtr: if you don't intend to do anything with the err, probably shouldn't return it

azure-ipam/cnsutils.go
```diff
@@ -0,0 +1,73 @@
...
+	ncGatewayIPAddress := resp.PodIpInfo.NetworkContainerPrimaryIPConfig.GatewayIPAddress
+	gwIP, err := netip.ParseAddr(ncGatewayIPAddress)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "cns returned an invalid gateway address %q", ncGatewayIPAddress)
+	}
+
+	return &podIPNet, &gwIP, nil
+}
+
+type K8SPodEnvArgs struct {
```
rbtr, line 59: does this need to be exported?

azure-ipam/logger.go
```diff
@@ -0,0 +1,54 @@
...
+		EncodeLevel: zapcore.LowercaseLevelEncoder,
+	}
+
+	logger, err := loggerCfg.Build()
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to build zap logger")
+	}
+
+	cleanup := func() {
+		logger.Sync() // nolint
```
rbtr, line 37: instead of squashing the linter, indicate that you are intentionally ignoring the error by blank assigning it
```suggestion
		_ = logger.Sync() // nolint
```

rbtr (changes requested):

azure-ipam/ipam_test.go
nddq: looping in @thatmattlong

azure-ipam/ipam.go
```diff
@@ -0,0 +1,213 @@
...
+	nwCfg, err := parseNetConf(args.StdinData)
+	if err != nil {
+		p.logger.Error("Failed to parse CNI network config from stdin",
+			zap.Error(err),
+			zap.Any("argStdinData", args.StdinData),
+		)
+		return errors.Wrapf(err, "failed to parse CNI network config from stdin")
+	}
+	p.logger.Debug("Parsed network config",
+		zap.Any("netconf", nwCfg),
```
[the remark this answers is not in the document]
nddq: I think that trading off vertical real-estate for code readability is worth it, no? Not line-wrap for log entries with small numbers of field but for log entries with large no of fields help us save vertical space but by doing so, we would lose the uniformity of our code, just my 2 cents.

azure-ipam/internal/buildinfo/buildinfo.go
```diff
@@ -0,0 +1,12 @@
...
+
+// this will be populate by the Go compiler via
+// the -ldflags, which insert dynamic information
+// into the binary at build time
+var Version string
+
+var (
+	LogLevel         string
+	OutputPaths      string // comma separated list of paths
+	ErrorOutputPaths string // comma separated list of paths
```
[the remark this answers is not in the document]
nddq: we would still want them to be populate at build time right?

azure-ipam/ipam_test.go
```diff
@@ -0,0 +1,291 @@
...
+	case "failRequestCNSReleaseIPArgs":
+		return errFoo
+	default:
+		return nil
+	}
+}
+
+const (
+	happyPodArgs    = "K8S_POD_NAMESPACE=testns;K8S_POD_NAME=testname;K8S_POD_INFRA_CONTAINER_ID=testid"
+	nothappyPodArgs = "K8S_POD_NAMESPACE=testns;K8S_POD_NAME=testname;K8S_POD_INFRA_CONTAINER_ID=testid;break=break" // this will break the pod config parsing
```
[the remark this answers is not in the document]
nddq: can you elaborate?

azure-ipam/ipam.go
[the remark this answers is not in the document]
tamilmani1989: whichever you feel not critical can be debug.. for eg: "Making request to CNS" can be debug

azure-ipam/internal/buildinfo/buildinfo.go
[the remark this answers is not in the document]
rbtr: no, they should be configurable and set at runtime (with default value)

azure-ipam/ipam.go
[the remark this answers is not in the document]
rbtr: this does not improve readability.

azure-ipam/go.mod
```diff
@@ -0,0 +1,82 @@
+module github.com/Azure/azure-container-networking/azure-ipam
```
wedaly, line 1: Could you try running a `go mod tidy` from within the `azure-ipam` directory? When I tried it, I got this error:
```
go: finding module for package github.com/containernetworking/cni/pkg/types/current
github.com/Azure/azure-container-networking/azure-ipam imports
        github.com/Azure/azure-container-networking/cns/client imports
        github.com/Azure/azure-container-networking/cns/restserver imports
        github.com/Azure/azure-container-networking/cns/ipamclient imports
        github.com/Azure/azure-container-networking/ipam tested by
        github.com/Azure/azure-container-networking/ipam.test imports
        github.com/containernetworking/cni/pkg/types/current: module github.com/containernetworking/cni@latest found (v1.1.1), but does not contain package github.com/containernetworking/cni/pkg/types/current
```

Actually, do we even need `azure-ipam` to be a go module? It seems like there's already a go.mod at the repository root.

azure-ipam/ipam_test.go
```diff
@@ -0,0 +1,283 @@
...
+		{
+			name:    "Fail request CNS release IP during CmdDel",
+			args:    buildArgs("failRequestCNSReleaseIPArgs", happyPodArgs, happyNetConfByteArr),
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			fmt.Printf("test : %v\n", tt.name)
```
wedaly, line 254: nit: the Go test framework should print the (sub)test name for you, I think? (Same for the other tests in this file)

wedaly (approved):
all my comments have been addressed, thanks! please check whether you need / should remove the go.mod in azure-ipam

azure-ipam/go.mod
rbtr: I want it to be a separate module; I've been trying to move the functionally distinct binaries in this monorepo to modules [for a while](https://github.com/Azure/azure-container-networking/pull/1136).
As for your build error - multimodule workspaces are new in go1.18 and with nested modules require that you have a [`go.work`](https://go.dev/doc/tutorial/workspaces) in the root module. (there is a `make workspace` target that _should_ set this up but I wouldn't rely on it yet.

azure-ipam/ipam_test.go
rbtr: @nddq tests shouldn't print directly to stdout, they should just pass or fail with message - `testify`'s assert and require functions have an optional message parameter at the end. Parallel tests all printing to stdout just make a mess 🥲

nddq: should I try using Ginko 🙂

wedaly (approved):

PR #1449, opened 2022-06-29
adding port field in endpoint struct & use that

kmurudi:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

 fix: CNI read port number from statefile than fetching from client

### DIFF
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -476,14 +476,6 @@ func (nm *networkManager) connectExternalInterface(extIf *externalInterface, nwI
 		return err
 	}
 
-	//fetch ovs port number and add it to nwInfo struct
-	extIf.OVSSwitchPort, err = networkClient.GetOVSPort(extIf)
-	if err != nil {
-		log.Printf("[ovs] Get ofport failed with error %v", err)
-		return err
-	}
-	nwInfo.OVSSwitchPort = extIf.OVSSwitchPort
-
 	// Add the bridge rules.
 	err = networkClient.AddL2Rules(extIf)
 	if err != nil {
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -476,6 +476,14 @@ func (nm *networkManager) connectExternalInterface(extIf *externalInterface, nwI
 		return err
 	}
 
+	//fetch ovs port number and add it to nwInfo struct
+	extIf.OVSSwitchPort, err = networkClient.GetOVSPort(extIf)
+	if err != nil {
+		log.Printf("[ovs] Get ofport failed with error %v", err)
+		return err
+	}
+	nwInfo.OVSSwitchPort = extIf.OVSSwitchPort
+
 	// Add the bridge rules.
 	err = networkClient.AddL2Rules(extIf)
 	if err != nil {
--- a/network/ovs_endpointclient_linux.go
+++ b/network/ovs_endpointclient_linux.go
@@ -99,6 +99,12 @@ func (client *OVSEndpointClient) AddEndpoints(epInfo *EndpointInfo) error {
 		return err
 	}
 
+	log.Printf("[ovs] Get ovs port for interface %v.", client.hostVethName)
+	epInfo.VethOVSSwitchPort, err = client.ovsctlClient.GetOVSPortNumber(client.hostVethName)
+	if err != nil {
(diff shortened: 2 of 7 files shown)


network/network_linux.go
```diff
@@ -476,6 +476,14 @@ func (nm *networkManager) connectExternalInterface(extIf *externalInterface, nwI
 		return err
 	}
 
+	//fetch ovs port number and add it to nwInfo struct
+	extIf.OVSSwitchPort, err = networkClient.GetOVSPort(extIf)
+	if err != nil {
+		log.Printf("[ovs] Get ofport failed with error %v", err)
+		return err
+	}
+	nwInfo.OVSSwitchPort = extIf.OVSSwitchPort
```
tamilmani1989, line 485: since we directly set to extIf which is part if nw structure...no need to copy this in nwInfo

network/ovs_endpointclient_linux.go
```diff
@@ -109,18 +115,10 @@ func (client *OVSEndpointClient) AddEndpointRules(epInfo *EndpointInfo) error {
...
-	}
+	containerOVSPort := epInfo.VethOVSSwitchPort
 
 	log.Printf("[ovs] Get ovs port for interface %v.", client.hostPrimaryIfName)
-	hostPort, err := client.ovsctlClient.GetOVSPortNumber(client.hostPrimaryIfName)
-	if err != nil {
-		log.Printf("[ovs] Get ofport failed with error %v", err)
-		return err
-	}
+	hostPort := epInfo.VethOVSSwitchPort
```
tamilmani1989, line 121: this should be extIf.OvsSwitchPort

network/ovs_endpointclient_linux.go
```diff
@@ -154,16 +152,10 @@ func (client *OVSEndpointClient) AddEndpointRules(epInfo *EndpointInfo) error {
...
-		log.Printf("[ovs] Get portnum failed with error %v", err)
-	}
+	containerPort := ep.VethOVSSwitchPort
 
 	log.Printf("[ovs] Get ovs port for interface %v.", client.hostPrimaryIfName)
-	hostPort, err := client.ovsctlClient.GetOVSPortNumber(client.hostPrimaryIfName)
-	if err != nil {
-		log.Printf("[ovs] Get portnum failed with error %v", err)
-	}
+	hostPort := ep.VethOVSSwitchPort
```
tamilmani1989, line 158: this should be extIf.OvsSwitchPort

network/network_linux.go
```diff
@@ -476,6 +476,14 @@ func (nm *networkManager) connectExternalInterface(extIf *externalInterface, nwI
 		return err
 	}
 
+	//fetch ovs port number and add it to nwInfo struct
```
tamilmani1989, line 479: this code shouldn't be here.. it should be in ovs specific file

tamilmani1989 (changes requested):

network/ovs_endpointclient_linux.go
kmurudi: this function does not have access to extIf? epInfo is the struct passed in here so I thought to fetch from epInfo

network/ovs_endpointclient_linux.go
kmurudi: same as above - this func is not getting extIf passed anywhere, it is passing ep endpoint struct, do we need to pass * network struct to this function caller here?

network/network_linux.go
kmurudi: will move this to AddL2Rules() in ovs_networkclient_linux.go

network/ovs_endpointclient_linux.go
tamilmani1989: yes..newovsendpointclient pass nw field, we should save in ovsendpointclient structure

network/ovs_endpointclient_linux.go
tamilmani1989: replied below

tamilmani1989 (changes requested):

network/ovs_endpointclient_linux.go
kmurudi: added

PR #1454, opened 2022-07-01
Port DeleteNetworkContainer and SetOrchestratorType from DNC

timraymond:
We currently use `/network/deletenetworkcontainer` and `/network/setorchestratortype` in DNC by directly making HTTP requests.  There was a small CNS client present to provide access to these two endpoints. This adds these two endpoints to the CNS client we already maintain so that we can just use the "official" one instead in DNC.

- [X] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/cns/client/client.go
+++ b/cns/client/client.go
@@ -609,3 +609,49 @@ func (c *Client) PublishNetworkContainer(ctx context.Context, pncr cns.PublishNe
 	// ...otherwise the request was successful so
 	return nil
 }
+
+// UnpublishNC unpublishes the network container via the NMAgent running
+// alongside the CNS service. This is useful to avoid throttling issues imposed
+// by Wireserver.
+func (c *Client) UnpublishNC(ctx context.Context, uncr cns.UnpublishNetworkContainerRequest) error {
+	// In order to unpublish a Network Container, we need its ID. If the ID is
+	// missing, we can assume that the request is invalid and immediately return
+	// an error
+	if uncr.NetworkContainerID == "" {
+		return errors.New("request missing network container id")
+	}
+
+	// Now that the request is valid it can be packaged as an HTTP request:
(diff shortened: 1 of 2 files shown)


cns/client/client.go
```diff
@@ -411,3 +411,105 @@ func (c *Client) GetHTTPServiceData(ctx context.Context) (*restserver.GetHTTPSer
...
+	var out cns.DeleteNetworkContainerResponse
+	err = json.NewDecoder(resp.Body).Decode(&out)
+	if err != nil {
+		return errors.Wrap(err, "decoding response as JSON")
+	}
+
+	// if a non-zero response code was received from CNS, it means something went
+	// wrong and it should be surfaced to the caller as an error
+	if out.Response.ReturnCode != 0 {
+		return errors.New(out.Response.Message)
```
ramiro-gamarra, line 455: Would a client want to handle errors differently for different error codes? I see other methods doing something similar but if we are going to start consuming this from DNC, I would like to leave this as open for future compatibility and return the actual API contract struct. Probably create new methods for all the methods that we will use from DNC.

ramiro-gamarra (commented):
One comment for discussion

cns/client/client.go
timraymond: I think it is likely that a client would want to handle errors differently (notably retries for temporary errors). I still think the better way to do it is by having a `TemporaryError` like I did in the NMAgent client (this is the same idea as Dave Cheney's ["don't just check errors, handle them gracefully"](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully)).

I like this because the knowledge about what constitutes a temporary error (i.e. what do the error codes mean) is in exactly one place: the `IsTemporary()` method (or equivalent for other classes of error).

rbtr (approved):
lgtm

PR #1471, opened 2022-07-18
feat: Add native linux endpoint client to prep removing OVS

QxBytes:
**Reason for Change**:

Creates a linux native endpoint client which is meant to replace the existing endpoint client that uses OVS. Uses the native linux kernel to replicate the functionality of OVS. Using the native client should improve the reliability of the multitenancy datapath because there will no longer be a dependency on OVS.

**Issue Fixed**:

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Notes**:
Creates a new netns package and uses vishvananda's netns and netlink library.


netns/netns.go
```diff
@@ -0,0 +1,28 @@
+package netns
+
+import "github.com/vishvananda/netns"
+
+type Netns struct{}
+
+func NewNetns() *Netns {
```
rbtr, line 7:
```suggestion
func New() *Netns {
```
you will always call this through the `netns` package, so it is unambiguous to say `netns.New()`

netns/netns.go
```diff
@@ -0,0 +1,28 @@
...
+	return &Netns{}
+}
+
+func (f *Netns) Get() (uintptr, error) {
+	nsHandle, err := netns.Get()
+	return uintptr(nsHandle), err
```
rbtr, line 13: why are you wrapping this `int` type in `uintptr`?

netns/netnsinterface.go
```diff
@@ -0,0 +1,9 @@
+package netns
+
+type NetnsInterface interface {
```
rbtr, line 3: This is a Java OOP pattern 👎🏼. In Go, _we_ do not define the interfaces when we define the types (because we don't need to!). The _consumer_ of our types is free to define the interface representing the (subset of the) behavior of our types that they need to use abstractly, and our types will implicitly implement their interface.

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,438 @@
...
+const (
+	azureMac         = "12:34:56:78:9a:bc" // Packets leaving the VM should have this MAC
+	loopbackIf       = "lo"                // The name of the loopback interface
+	numDefaultRoutes = 2                   // VNET NS, when no containers use it, has this many routes
+)
+
+var errorNativeEndpointClient = errors.New("NativeEndpointClient Error")
+
+func newErrorNativeEndpointClient(msg string, errStr string) error {
+	return fmt.Errorf("%w : %s : %s", errorNativeEndpointClient, msg, errStr)
```
rbtr, line 28: use `pkg/errors.Wrap` for consistent wrapping

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,438 @@
...
+func NewNativeEndpointClient(
+	eth0VethName string,
+	ethXVethName string,
+	vnetVethName string,
+	containerVethName string,
+	vnetNSName string,
+	mode string,
+	vlanid int,
+	nl netlink.NetlinkInterface,
+	plc platform.ExecClient,
```
rbtr, line 62: That you need to wrap the method signature tells me there are way too many args. Consider creating a config struct which wraps the args that you need to pass in to build this object.
OR, if there is no "magic" happening in this `New`, remove it entirely? Directly instantiating the struct should always be your default new object creation method, up until creating your object is too complicated to trust the consumer of your type to perform correctly.

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,438 @@
...
+	} else {
+		log.Printf("[native] VM Namespace: %s", returnedTo.file.Name())
+	}
+
+	log.Printf("[native] Checking if NS exists...")
+	var existingErr error
+	vnetNS, existingErr := client.netnsClient.GetFromName(client.vnetNSName)
+	// If the ns does not exist, the below code will trigger to create it
+	if existingErr != nil {
+		if !strings.Contains(strings.ToLower(existingErr.Error()), "no such file or directory") {
```
rbtr, line 120: could use `errors.Is` instead of stringly matching?

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,438 @@
...
+			return newErrorNativeEndpointClient("Error other than vnet NS doesn't exist", existingErr.Error())
```
rbtr, line 121: - errors should not be `Sentence case`, they tend to be embedded as logs and should be readable as fragments
- should use error wrapping

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,438 @@
...
+	var routeInfoList []RouteInfo
+	// Add route specifying which device the pod ip(s) are on
+	for _, ipAddr := range ipAddresses {
+		var (
+			routeInfo RouteInfo
+			ipNet     net.IPNet
+		)
+
+		if ipAddr.IP.To4() != nil {
+			ipNet = net.IPNet{IP: ipAddr.IP, Mask: net.CIDRMask(ipv4FullMask, ipv4Bits)}
```
rbtr, line 312: I think new code should use the `netip` package instead of `net`. The types and fns are a significant improvement.

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,438 @@
...
+		if delErr != nil {
+			return newErrorNativeEndpointClient("Failed to delete namespace", delErr.Error())
+		}
+	}
+	return nil
+}
+
+// Helper function that allows executing a function with one parameter in a VM namespace
+// Does not work for process namespaces
+func ExecuteInNS[T any](nsName string, param *T, f func(param *T) error) error {
```
rbtr, line 402: generics are cool, but I don't think they're necessary here. This function should accept a parameter-less function which is closed over the inputs that it needs.
For example, you could close any arbitrary context and execute it in the namespace like:
```go
func ExecuteInNS(ns string, fn func() error) error {
	//<...enter ns...>
	return fn()
}

// close the input func over arg "param"
param := "some input to my anonymous func"
ExecuteInNS("abcnamespace", func() error {
	fmt.Printf("closed over param %s", param)
	//<...do things...>
	return nil
})
```

rbtr (changes requested):
some initial comments - several things which apply globally, and I expect fixes to be applied to the whole PR even though I have only commented on a single instance.
pls also address the feedback from the linter

tamilmani1989 (commented):
Thanks for this PR.. Left some comments

network/native_endpointclient_linux.go
QxBytes: I am unable to use the `netip` package because I need to call `addRoutes`, which takes in a `[]RouteInfo` parameter. This type requires types from `net.IP` (which is what is being constructed in the code above). There does not appear to be an equivalent of `addRoutes` in the `netip` package so I need to use the `addRoutes` that uses `net`.

rbtr: ack

netns/netns.go
```diff
@@ -1,7 +1,9 @@
+//nolint
 package netns
 
 import (
 	"github.com/pkg/errors"
+	//nolint
```
rbtr, line 6: what was the lint here? shadow?

QxBytes: I've been trying to fix a linter error and was testing if any location for `//nolint` would cause the windows-latest linter to pass. However, the linter still appears to fail so I'm not sure what is causing it.

QxBytes: Fixed by adding +build linux

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,403 @@
...
+		virtualGwNet.String(), destMac)
+	hardwareAddr, err := net.ParseMAC(destMac)
+	if err != nil {
+		return errors.Wrap(err, "unable to parse mac")
+	}
+	if err := client.netlink.AddOrRemoveStaticArp(netlink.ADD,
+		interfaceName,
+		virtualGwNet.IP,
+		hardwareAddr,
+		false); err != nil {
```
rbtr, line 328: unwrap
splitting method signatures or invocations across lines is worse for readability

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,403 @@
...
+			if deleteNSIfNotNilErr != nil {
+				if delErr := client.netlink.DeleteLink(client.vlanVethName); delErr != nil {
+					log.Errorf("deleting vlan veth failed on addendpoint failure")
+				}
+				return errors.Wrap(deleteNSIfNotNilErr, "deleting vlan veth in vm ns due to addendpoint failure")
+			}
+		} else {
+			// Any other error
+			return errors.Wrap(deleteNSIfNotNilErr, "failed to create vlan vnet link after making new ns")
+		}
```
rbtr, line 128: i think you could flip this conditional to eliminate the `else` branch entirely

rbtr (dismissed):
some nitpicks but generally lgtm

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,397 @@
...
+
+	// Open the network namespace
+	log.Printf("[ExecuteInNS] Opening ns %v.", fmt.Sprintf("/var/run/netns/%s", nsName))
+	ns, err := OpenNamespace(fmt.Sprintf("/var/run/netns/%s", nsName))
+	if err != nil {
+		return err
+	}
+	defer ns.Close()
+	// Enter the network namespace
+	log.Printf("[ExecuteInNS] Entering vnetns %s.", ns.file.Name())
```
tamilmani1989, line 378: log looks confusing. why vnetns? i guess this is generic function

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,397 @@
...
+	if err := ns.Enter(); err != nil {
+		return err
+	}
+
+	// Exit network namespace
+	defer func() {
+		log.Printf("[ExecuteInNS] Exiting vnetns %s.", ns.file.Name())
```
tamilmani1989, line 385: same here

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,397 @@
...
+		if err != nil {
+			return errors.Wrap(err, ("failed to get route list"))
+		}
+		log.Printf("[native] Routes remaining: %v", routes)
+
+		return client.DeleteEndpointsImpl(ep, len(routes))
+	})
+}
+
+func (client *NativeEndpointClient) DeleteEndpointsImpl(ep *endpoint, routesLeft int) error {
```
tamilmani1989, line 340: wanted to understand on routesLeft logic.. lets talk offline about this

tamilmani1989 (changes requested):

network/native_endpointclient_linux.go
```diff
@@ -0,0 +1,407 @@
...
+
+type netnsClient interface {
+	Get() (fileDescriptor int, err error)
+	GetFromName(name string) (fileDescriptor int, err error)
+	Set(fileDescriptor int) (err error)
+	NewNamed(name string) (fileDescriptor int, err error)
+	DeleteNamed(name string) (err error)
+}
+type NativeEndpointClient struct {
+	eth0VethName      string // So like eth0
```
tamilmani1989, line 30: this is not veth..rename as "primaryIfName"

tamilmani1989 (changes requested):

tamilmani1989 (approved):
lgtm

rbtr (approved):

rbtr:
rebased for CI fixeds; lgtm

PR #1475, opened 2022-07-18
Update Swift pipeline w byo cluster

camrynl:
**Reason for Change**:

Current e2e tests are utilizing an existing cluster. Each PR waits for the cluster to be available before running tests.
Adding byo cluster feature to pipeline so each PR can run tests in parallel using its own cluster.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


camrynl:
PR still in progress. Can't convert to draft since I'm testing the pipeline. I'll notify when to review, ty :)

.pipelines/singletenancy/aks-swift/e2e-step-template.yaml
```diff
@@ -26,18 +26,17 @@ steps:
 
   - task: AzureCLI@1
     inputs:
-      azureSubscription: $(BUILD_VALIDATIONS_SERVICE_CONNECTION)
+      azureSubscription: "Azure Container Networking - Test"
```
rbtr, line 29: there is a pipeline var for this Sub, we should use it

cns/azure-cns.yaml
```diff
@@ -96,7 +96,7 @@ spec:
           effect: NoSchedule
       containers:
         - name: cns-container
-          image: mcr.microsoft.com/containernetworking/azure-cns:v1.4.7
+          image: mcr.microsoft.com/containernetworking/azure-cns:v1.4.29
```
rbtr, line 99: Can you revert this? I think this manifest is used by partners, and is unrelated to this change anyway

rbtr (changes requested):

rbtr:
You can also remove this line https://github.com/Azure/azure-container-networking/blob/637d5affe966f2aadf0240e64afc0cc34e481696/.pipelines/singletenancy/aks-swift/e2e-job-template.yaml#L9 which currently prevents this test job from running in parallel 🙂

rbtr (approved):
🚀🚀

PR #1477, opened 2022-07-19
Cilium Integration fixes 

nddq:
As described in title.

### DIFF
--- a/azure-ipam/main.go
+++ b/azure-ipam/main.go
@@ -24,7 +24,7 @@ func executePlugin() error {
 	loggerCfg := &logger.Config{
 		Level:            "debug",
 		OutputPaths:      "var/log/azure-ipam.log",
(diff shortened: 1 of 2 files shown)


azure-ipam/main.go
```diff
@@ -21,7 +21,7 @@ func main() {
 
 func executePlugin() error {
 	// logger config
-	var loggerCfg *logger.Config
+	loggerCfg := &logger.Config{}
```
rbtr, line 24: maybe inline setting the opts from right below?
```suggestion
	loggerCfg := &logger.Config{
		Level: 		  "debug",
		OutputPaths: 	  "stdout",
		ErrorOutputPaths: "stderr"
	}
```

rbtr (changes requested):

rbtr (approved):

tamilmani1989 (approved):

PR #1483, opened 2022-07-20
add cilium and azure-ipam to dropgz image

rbtr:

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/azure-ipam/azilium.conflist
+++ b/azure-ipam/azilium.conflist
@@ -1,0 +1,14 @@
+{
+    "cniVersion": "0.3.1",
+    "name": "cilium",
+    "plugins": [
+        {
+            "type": "cilium-cni",
+            "ipam": {
+                "type": "azure-ipam"
+            },
+            "enable-debug": true,
+            "log-file": "/var/log/cilium-cni.log"
+        }
+    ]
+}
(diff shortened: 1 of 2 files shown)


azure-ipam/azilium.conflist
```diff
@@ -0,0 +1,14 @@
+{
+    "cniVersion": "0.3.1",
+    "name": "cilium",
```
nddq, line 3: we need to change the name here, otherwise Cilium won't invoke cmdDel from azure-ipam. Pinging @tamilmani1989 and @wedaly

nddq: apparently, the name conflicts with another existing chaining plugin's name, which makes the cmdDel flow returns early

nddq: https://github.com/cilium/cilium/blob/8bcc5fe2c594b5fb4b9d2966906c490dea215d83/plugins/cilium-cni/cilium-cni.go#L664

wedaly: I think "cilium" will work, actually -- it's "azure" that was causing problems. The `chainingapi.DefaultConfigName` in the CNI code is "cilium", so setting name="cilium" tells the CNI to skip chaining (which os what we want).

nddq (approved):
lgtm 👌

PR #1484, opened 2022-07-20
Program IPTables SNAT rules in CNS

nddq:
Draft PR for programming iptables snat rules in CNS, which is a part of delegated IPAM.

### DIFF
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -35,6 +35,7 @@ type CNSConfig struct {
 	WireserverIP                string
 	KeyVaultSettings            KeyVaultSettings
 	MSISettings                 MSISettings
+	DelegatedIPAMUsed           bool
 }
 
 type TelemetrySettings struct {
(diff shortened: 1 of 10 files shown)


rbtr (changes requested):
looks good, needs one addition - programming the iptables rules should be gated on a config flag eg: `EnableProgramIPTables` that we default false. this way we can keep rolling out CNS without expecting this functionality to just start running unexpectedly.

cns/configuration/configuration.go
```diff
@@ -35,6 +35,7 @@ type CNSConfig struct {
 	WireserverIP                string
 	KeyVaultSettings            KeyVaultSettings
 	MSISettings                 MSISettings
+	DelegatedIPAMUsed           bool
```
rbtr, line 38: i think updating this var name was missed

rbtr (changes requested):

rbtr (approved):

PR #1500, opened 2022-07-31
feat: CNS manages endpoints state for delegate IPAM use case

nddq:
This is the draft PR for the endpoint state management functionality for CNS for when delegated IPAM is used
TODO: Handling state for release IP config request ✅


cns/cnireconciler/podinfoprovider.go
```diff
@@ -44,3 +67,32 @@ func cniStateToPodInfoByIP(state *api.AzureCNIState) (map[string]cns.PodInfo, er
 	}
 	return podInfoByIP, nil
 }
+
+func endpointStateToPodInfoByIP(state map[string]*restserver.EndpointInfo) (map[string]cns.PodInfo, error) {
+	podInfoByIP := map[string]cns.PodInfo{}
+	for containerID, endpointInfo := range state {
+		for ifname, ipinfo := range endpointInfo.IfnameToIPMap {
+			if _, ok := podInfoByIP[ipinfo.IPv4.IP.String()]; ok {
```
tamilmani1989, line 75: its good to check ipinfo.IPv4!= nil as like ipv6. if in future, if azure supports single stack ipv6 , then this would be an issue

cns/restserver/ipam.go
```diff
@@ -76,6 +97,54 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
 	logger.ResponseEx(service.Name+operationName, ipconfigRequest, reserveResp, reserveResp.Response.ReturnCode, err)
 }
 
+var errNilStateStore = errors.New("nil endpoint state store")
+
+func (service *HTTPRestService) updateEndpointState(ipconfigRequest cns.IPConfigRequest, podInfo cns.PodInfo, podIPInfo cns.PodIpInfo) error {
+	if service.EndpointStateStore == nil {
+		return errNilStateStore
```
tamilmani1989, line 104: errStoreEmpty?

cns/restserver/restserver.go
```diff
@@ -58,7 +59,20 @@ type HTTPRestService struct {
...
+}
+
+type EndpointInfo struct {
+	PodName       string
+	PodNamespace  string
+	IfnameToIPMap map[string]*IPInfo // key : interface name, value : IPInfo
+}
+
+type IPInfo struct {
+	IPv4 *net.IPNet
```
tamilmani1989, line 74: can this be array of addresses? e.g []net.IPNet

tamilmani1989: also why this is a pointer instead of net.ipnet?

cns/restserver/ipam.go
```diff
@@ -76,6 +97,54 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
...
+	}
+	service.Lock()
+	defer service.Unlock()
+	logger.Printf("[updateEndpointState] Updating endpoint state for infra container %s", ipconfigRequest.InfraContainerID)
+	if endpointInfo, ok := service.EndpointState[ipconfigRequest.InfraContainerID]; ok {
+		logger.Printf("[updateEndpointState] Found existing endpoint state for infra container %s", ipconfigRequest.InfraContainerID)
```
tamilmani1989, line 110: when do we get this case? if code hits here, isnt something wrong?

tamilmani1989 (changes requested):

cns/restserver/restserver.go
nddq: Why would it be an array? Would an ifname have multiple IP addresses?

tamilmani1989: not today..if in future if we support multiple v4 address in interface, then it can be extensible

cns/restserver/ipam.go
nddq: I don't have a case in mind, but it is generally better to check for duplicates in a map?

cns/restserver/ipam.go
```diff
@@ -76,6 +97,66 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
...
+		ip, ipnet, err := net.ParseCIDR(podIPInfo.PodIPConfig.IPAddress + "/" + fmt.Sprint(podIPInfo.PodIPConfig.PrefixLength))
+		if err != nil {
+			return fmt.Errorf("failed to parse pod ip address: %w", err)
+		}
+		ipconfig := net.IPNet{IP: ip, Mask: ipnet.Mask}
```
tamilmani1989, line 115: instead of this, there s better way to do it.
```
ip := net.ParseIP(podIPInfo.PodIPConfig.IPAddress)
if ip.To4() == nil {
   ipconfig := net.IPNet{IP: ip, Mask: net.CIDRMask(podIPInfo.PodIPConfig.PrefixLength, 32)}
} else {
  ipconfig := net.IPNet{IP: ip, Mask: net.CIDRMask(podIPInfo.PodIPConfig.PrefixLength, 128)}
}
```
}

cns/restserver/ipam.go
```diff
@@ -76,6 +97,66 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
...
+
+		service.EndpointState[ipconfigRequest.InfraContainerID] = endpointInfo
+
+	} else {
+		endpointInfo := &EndpointInfo{PodName: podInfo.Name(), PodNamespace: podInfo.Namespace(), IfnameToIPMap: make(map[string]*IPInfo)}
+		ip, ipnet, err := net.ParseCIDR(podIPInfo.PodIPConfig.IPAddress + "/" + fmt.Sprint(podIPInfo.PodIPConfig.PrefixLength))
+		if err != nil {
+			return fmt.Errorf("failed to parse pod ip address: %w", err)
+		}
+		ipconfig := net.IPNet{IP: ip, Mask: ipnet.Mask}
```
tamilmani1989, line 142: same as above..

tamilmani1989 (changes requested):

cns/restserver/restserver.go
```diff
@@ -58,7 +59,20 @@ type HTTPRestService struct {
...
-	dncPartitionKey string
+	dncPartitionKey    string
+	EndpointState      map[string]*EndpointInfo // key : container id
+	EndpointStateStore store.KeyValueStore
+}
+
+type EndpointInfo struct {
+	PodName       string
+	PodNamespace  string
+	IfnameToIPMap map[string]*IPInfo // key : interface name, value : IPInfo
```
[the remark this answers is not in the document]
nddq: yeah, I only used pointer because I don't know how big the struct would grow to be. If the struct consistently stay relatively small then I guess we don't have to use pointers.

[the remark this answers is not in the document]
nddq: actually, we need pointers for IPInfo during IPconfig request and release, as we don't want to create a new copy of the arrays within IPInfo, update with new entries, and then reassign it every time.

cns/restserver/restserver.go
```diff
@@ -58,7 +59,20 @@ type HTTPRestService struct {
 	state                    *httpRestServiceState
 	podsPendingIPAssignment  *bounded.TimedSet
 	sync.RWMutex
-	dncPartitionKey string
+	dncPartitionKey    string
+	EndpointState      map[string]*EndpointInfo // key : container id
```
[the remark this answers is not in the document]
nddq: maybe we don't need a pointer for this case?

cns/restserver/ipam.go
nddq: this looks cleaner! Will make the change.

tamilmani1989 (approved):

PR #1505, opened 2022-08-02
Update to Go 1.19

rbtr:

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
- **All necessary Go 1.19 base images have not been published yet; this will stay as draft until the dependencies are available.**


timraymond (approved):
Looks good to me... I _think_ this is everywhere? It'd almost be nice to have a GH action to bump Go versions just so we don't miss any place in the future.

rbtr:
I think so. An action with paths to sed or something might be nice, but hard to maintain... And even if we automatically update, someone needs to manually update the required PR checks because Github does string matching and the names have the version like `Lint(1.18.x, ubuntu-latest)`

timraymond:
mmm... fair point. We'd likely only run it every six months, where we find that other things have encoded knowledge about the Go version that weren't also updated in the script. Probably more work than it's worth.

timraymond (approved):

PR #1512, opened 2022-08-05
[NPM] include os in log telemetry

huntergregory:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/metrics/ai-utils.go
+++ b/npm/metrics/ai-utils.go
@@ -83,7 +83,11 @@ func SendMetric(metric aitelemetry.Metric) {
 
 // SendLog sends log
 func SendLog(operationID int, msg string, printLog bool) {
-	msg = fmt.Sprintf("%s - (NPM v%d)", msg, npmVersion)
+	osName := "linux"
+	if util.IsWindowsDP() {
+		osName = "windows"
+	}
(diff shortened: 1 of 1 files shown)


npm/metrics/ai-utils.go
```diff
@@ -83,7 +83,11 @@ func SendMetric(metric aitelemetry.Metric) {
 
 // SendLog sends log
 func SendLog(operationID int, msg string, printLog bool) {
-	msg = fmt.Sprintf("%s - (NPM v%d)", msg, npmVersion)
+	osName := "linux"
+	if util.IsWindowsDP() {
+		osName = "windows"
+	}
+	msg = fmt.Sprintf("%s - (NPM v%d) - (%s)", msg, npmVersion, osName)
```
vakalapa, line 90: have a different identifier, so that it will be easy to add a regex to query this out in kusto

vakalapa: may be create a json and log it, so it will be easier to use parse_json in kusto

vakalapa: We have multiple values in messages db which point to windows. Like Auth id column gives the OS name, custom dimensions have the os version etc. So adding this might be redundant

huntergregory:
os info is already in metadata

PR #1518, opened 2022-08-09
Add missing client paths to CNS Client

timraymond:
These were forgotten during a previous effort to add endpoints to the
CNS client. Consequently, those endpoints don't actually work unless
these paths are present here.

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr (approved):

rbtr:
bypassing broken akse tests

timraymond:
Heh, I was confused why those were broken and wailed on "re-build" a couple of times. Thanks for merging it :)

PR #1523, opened 2022-08-10
feat: adding log rotation to Azure-IPAM Zap logger via lumberjack

AzureAhai:
Since Zap logger does not have a native rotation support. I added lumberjack filewriter that can rotate based on a given file size.


azure-ipam/logger/logger.go
```diff
@@ -6,56 +6,48 @@ import (
...
-
+	// set the log level - Debug by default
+	var logLevel zapcore.Level
+	if strings.ToLower(cfg.Level) == "info" {
+		logLevel = zap.InfoLevel
+	} else if strings.ToLower(cfg.Level) == "error" {
+		logLevel = zap.ErrorLevel
+	} else {
+		logLevel = zap.DebugLevel
+	}
```
rbtr, line 35: imo [this was previously done correctly](https://github.com/Azure/azure-container-networking/pull/1523/files#diff-00339433bfed6e3f6830a7e8f1ecd201d726875a0313512452c04a20a8bc8c6fL21-L25) as 
```go
level, err := zapcore.ParseLevel(cfg.Level)
```
what is the reason to change this to parse the strings ourselves?

azure-ipam/logger/logger.go
```diff
@@ -6,56 +6,48 @@ import (
...
-		TimeKey:     "time",
-		MessageKey:  "msg",
-		LevelKey:    "level",
-		EncodeLevel: zapcore.LowercaseLevelEncoder,
-		EncodeTime:  zapcore.ISO8601TimeEncoder,
-	}
-
-	logger, err := loggerCfg.Build()
-	if err != nil {
+	//check the filepath is not empty
```
rbtr, line 22: the linter is being lenient for you... comments must have a space after the `//`. (and they should explain _why_, not what, you are doing).

azure-ipam/logger/logger.go
```diff
@@ -6,56 +6,48 @@ import (
...
+	if cfg.Filename == "" {
+		err := errors.New("no Filename is provided")
```
rbtr, line 24: no filename should not be an error, it just means don't log to a file

azure-ipam/logger/logger.go
```diff
@@ -6,56 +6,48 @@ import (
...
-}
 
-func getErrOutputPath(paths string) []string {
-	if paths == "" {
-		return nil
-	}
-	return strings.Split(paths, ",")
-}
+	return logger, cleanup, nil
+}
```
rbtr, line 53: configure your editor to add trailing newlines on save

azure-ipam/main.go
```diff
@@ -22,9 +22,9 @@ func main() {
 func executePlugin() error {
 	// logger config
 	loggerCfg := &logger.Config{
-		Level:            "debug",
-		OutputPaths:      "var/log/azure-ipam.log",
-		ErrorOutputPaths: "var/log/azure-ipam.log",
+		Level:                "debug",
+		Filename:             "var/log/azure-ipam.log",
```
rbtr, line 26: should be an absolute path?

azure-ipam/logger/logger.go
```diff
@@ -6,56 +6,48 @@ import (
...
 	"go.uber.org/zap/zapcore"
+	"gopkg.in/natefinch/lumberjack.v2"
 )
 
 type Config struct {
-	Level            string
-	OutputPaths      string // comma separated list of paths
-	ErrorOutputPaths string // comma separated list of paths
+	Level       string // Debug by default
+	Filename   	string
```
rbtr, line 14: `FilePath` is probably more connect naming

rbtr (changes requested):

azure-ipam/logger/logger.go
AzureAhai: I followed lumberjack naming convention. But have no objection to change it to FilePath

azure-ipam/main.go
AzureAhai: If you don't start it with "/" then it will consider it relative to "/" 
It was previously written like this and I didn't change it.

azure-ipam/logger/logger.go
AzureAhai: Will address in the new commit

azure-ipam/logger/logger.go
AzureAhai: The issue that I had with the parseLevel was that if you pass "" as level or any of the non-defined string, it will through an Error and apparently prevent Azure-IPAM execution (That's what I observed). @tamilmani1989 suggested to have Debug mode as default for Azure-IPAM. 
I can still use ParseLevel() but if an error occurs, just set the level as debug instead of returning an error.

azure-ipam/logger/logger.go
AzureAhai: will address in the next commit

azure-ipam/logger/logger.go
```diff
@@ -6,56 +6,48 @@ import (
...
+	MaxSize    	int    // megabytes
+	MaxBackups 	int    // # of backups, no limitation by default
+	MaxAge     	int    // days, no limitation by default
```
tamilmani1989, line 17: can remove MaxAge. Lets add if needed

azure-ipam/main.go
tamilmani1989: lets fix this and have absolute path

azure-ipam/logger/logger.go
tamilmani1989: @AzureAhai  lets not change this and use ParseLevel itself and debug is defined as default in main.go already

azure-ipam/main.go
```diff
@@ -22,9 +22,9 @@ func main() {
...
+		MaxSize:              10, //MB
```
tamilmani1989, line 27: lets have 40 MB and 5 MB per file and 8 files in total

tamilmani1989 (changes requested):

azure-ipam/logger/logger.go
AzureAhai: Lumberjack log in <processname>-lumberjack.log within os.TempDir() if filename is empty. 

I decided to avoid such scenario. If we make this package universal, then we can address this better.

azure-ipam/main.go
AzureAhai: Will address in the new commit

azure-ipam/logger/logger.go
AzureAhai: Will address in the new commit

azure-ipam/main.go
rbtr: > If you don't start it with "/" then it will consider it relative to "/"

It will consider it relative to the current working directory, which _happens_ to be `/`. That is not guaranteed.

azure-ipam/logger/logger.go
```diff
@@ -1,61 +1,42 @@
...
 	"go.uber.org/zap/zapcore"
+	"gopkg.in/natefinch/lumberjack.v2"
 )
 
 type Config struct {
-	Level            string
-	OutputPaths      string // comma separated list of paths
-	ErrorOutputPaths string // comma separated list of paths
+	Level       string // Debug by default
+	Filepath   	string // if Empty log into <processname>-lumberjack.log within os.TempDir()
```
rbtr, line 12: something is off with the spacing here

azure-ipam/logger/logger.go
```diff
@@ -1,61 +1,42 @@
...
+	MaxSize    	int    // megabytes
```
rbtr, line 13: It will not be clear to me reading or writing the config that this is megabytes. We should either use the base unit, or rename the config appropriately.

azure-ipam/logger/logger.go
```diff
@@ -1,61 +1,42 @@
...
+	MaxBackups 	int    // # of backups, no limitation by default
 }
 
 // NewLogger creates and returns a zap logger and a clean up function
 func New(cfg *Config) (*zap.Logger, func(), error) {
-	loggerCfg := &zap.Config{}
-
-	level, err := zapcore.ParseLevel(cfg.Level)
+	// check the filepath is not empty
```
rbtr, line 19: comment seems out of place

rbtr (changes requested):

azure-ipam/logger/logger.go
rbtr: I think you misunderstood what I meant - if I don't specify a log file, you should not log to a file at all

rbtr:
@nddq do you mind giving this a look?

azure-ipam/logger/logger.go
nddq: I suggest changing it to MaxSizeInMB.

tamilmani1989 (dismissed):
lgtm

AzureAhai:
@rbtr Can you plz review the new commit?

rbtr:
have you addressed my request that not setting a file name not be an error?

AzureAhai:
> The only error that the function returns is when the log level is set incorrectly. 
In case of empty filepath then it will log into `<processname>-lumberjack.log` within os.TempDir()` 
which is the default expectation of the lumberjack pkg.

rbtr:
> @[rbtr](https://github.com/rbtr) [3 days ago](https://github.com/Azure/azure-container-networking/pull/1523#discussion_r946099397)
I think you misunderstood what I meant - if I don't specify a log file, you should not log to a file at all

AzureAhai:
> > @[rbtr](https://github.com/rbtr) [3 days ago](https://github.com/Azure/azure-container-networking/pull/1523#discussion_r946099397)
> > I think you misunderstood what I meant - if I don't specify a log file, you should not log to a file at all

I tried os.

rbtr:
So don't load lumberjack. If I don't specify a log file, I don't need log file rotation.
That's the nice thing about the plugin/composable architecture of zap - we can construct exactly what we need at runtime out of the building blocks.

rbtr:
I was not suggesting that we start logging on stderr; I'm not even sure if that is okay. I know that we can't log on stdout because that's how the CNI plugins communicate. Does the invoker expect request responses on stderr as well as stdout, or does it ignore it? 
/cc @wedaly

wedaly:
> Does the invoker expect request responses on stderr as well as stdout, or does it ignore it?

I believe the invoker copies stderr to `os.Stderr`

https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/pkg/invoke/raw_exec.go#L67

https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/pkg/invoke/exec.go#L185-L187

This is what the CNI spec has to say:

1. Plugin errors should go to stderr: https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/SPEC.md?plain=1#L195
2. Delegated plugins stderr should go to calling plugin's stderr: https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/SPEC.md?plain=1#L460

AzureAhai:
> > Does the invoker expect request responses on stderr as well as stdout, or does it ignore it?
> 
> I believe the invoker copies stderr to `os.Stderr`
> 
> https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/pkg/invoke/raw_exec.go#L67
> 
> https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/pkg/invoke/exec.go#L185-L187
> 
> This is what the CNI spec has to say:
> 
> 1. Plugin errors should go to stderr: https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/SPEC.md?plain=1#L195
> 2. Delegated plugins stderr should go to calling plugin's stderr: https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/SPEC.md?plain=1#L460

Do we have any specification for info and debug logs? Do we need to send them to stdout? @rbtr We have the option to set the outputpath to stdout and errorpath to stderr.

AzureAhai:
> > > Does the invoker expect request responses on stderr as well as stdout, or does it ignore it?
> > 
> > 
> > I believe the invoker copies stderr to `os.Stderr`
> > https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/pkg/invoke/raw_exec.go#L67
> > https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/pkg/invoke/exec.go#L185-L187
> > This is what the CNI spec has to say:
> > 
> > 1. Plugin errors should go to stderr: https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/SPEC.md?plain=1#L195
> > 2. Delegated plugins stderr should go to calling plugin's stderr: https://github.com/containernetworking/cni/blob/40fa4795fbee95922aa6f04b3cf0639563419055/SPEC.md?plain=1#L460
> 
> Do we have any specification for info and debug logs? Do we need to send them to stdout? @rbtr We have the option to set the outputpath to stdout and errorpath to stderr.

@wedaly Another thing that concerns me is that according to CNI spec Do we need to always have the delegated plugin Errors passed to the stderr even if we are writing all the logs to a file?

rbtr:
Thanks @wedaly
I read that as _only errors_ should go on stderr, and that logging to it might be interpreted as errors.
@AzureAhai while this might be Zap's default behavior, the logger config was [previously hardcoded to have file paths](https://github.com/Azure/azure-container-networking/blob/2417b7effc3826ea7ca32514f9039ef602c40e73/azure-ipam/main.go#L24-L28) making this a behavior change to `azure-ipam` that I don't think the CNI will handle.
CNI plugins communicate over stdout and stderr, where they expect input/output arguments and real errors. We can't log to those, because the CNI calling us will (fail to) interpret the logs as data.

AzureAhai:
> Thanks @wedaly I read that as _only errors_ should go on stderr, and that logging to it might be interpreted as errors. @AzureAhai while this might be Zap's default behavior, the logger config was [previously hardcoded to have file paths](https://github.com/Azure/azure-container-networking/blob/2417b7effc3826ea7ca32514f9039ef602c40e73/azure-ipam/main.go#L24-L28) making this a behavior change to `azure-ipam` that I don't think the CNI will handle. CNI plugins communicate over stdout and stderr, where they expect input/output arguments and real errors. We can't log to those, because the CNI calling us will (fail to) interpret the logs as data.

AzureAhai:
> Thanks @wedaly I read that as _only errors_ should go on stderr, and that logging to it might be interpreted as errors. @AzureAhai while this might be Zap's default behavior, the logger config was [previously hardcoded to have file paths](https://github.com/Azure/azure-container-networking/blob/2417b7effc3826ea7ca32514f9039ef602c40e73/azure-ipam/main.go#L24-L28) making this a behavior change to `azure-ipam` that I don't think the CNI will handle. CNI plugins communicate over stdout and stderr, where they expect input/output arguments and real errors. We can't log to those, because the CNI calling us will (fail to) interpret the logs as data.

We may need to always write the errors to stderr. For the info and debug log, we may need to decide to have a default path as before.

AzureAhai:
> Thanks @wedaly I read that as _only errors_ should go on stderr, and that logging to it might be interpreted as errors. @AzureAhai while this might be Zap's default behavior, the logger config was [previously hardcoded to have file paths](https://github.com/Azure/azure-container-networking/blob/2417b7effc3826ea7ca32514f9039ef602c40e73/azure-ipam/main.go#L24-L28) making this a behavior change to `azure-ipam` that I don't think the CNI will handle. CNI plugins communicate over stdout and stderr, where they expect input/output arguments and real errors. We can't log to those, because the CNI calling us will (fail to) interpret the logs as data.

Just trying to make sure that I get your point here:
So currently we are hardcoding the filepath in the caller in main.go which is the same as this PR from the beginning. I was assuming your concern was that if we don't specify a filepath in the caller we should not log to a file. Due to issues with writing to Stderr or Stdout, I'm going to just specify a default filepath in the logger in case that the caller is not specifying a filepath. @rbtr is this alright?

rbtr:
I do not like lumberjack's random filename default. I could be convinced that a hardcoded default filepath is okay, but you are correct that my initial suggestion was that if the caller does not specify a path, we do not log to a file (and this does not imply anything about any other behavior, only that no path => no log file).

AzureAhai:
> I do not like lumberjack's random filename default. I could be convinced that a hardcoded default filepath is okay, but you are correct that my initial suggestion was that if the caller does not specify a path, we do not log to a file (and this does not imply anything about any other behavior, only that no path => no log file).

Based on today's meeting I will put `/var/log/azure-ipam.log` as the default in the logger, even if we pass "" as the filepath because if we don't do that it will got to Stderr (If we pass "" the current code will also log in Stderr)
We can have this issue in mind to define a default path later on as @tamilmani1989 suggested.

azure-ipam/logger/logger.go
```diff
@@ -1,61 +1,53 @@
...
-		EncodeTime:  zapcore.ISO8601TimeEncoder,
+	if cfg.Filepath == "" {
+		cfg.Filepath = Filepath
 	}
-
-	logger, err := loggerCfg.Build()
-	if err != nil {
-		return nil, nil, errors.Wrapf(err, "failed to build zap logger")
-	}
-
```
tamilmani1989, line 42: why did we remove this custom encoder config?

AzureAhai: zap.conf.build() will return an error if the encoding or filepath have issues. But Zapcore.newcore() and zap.New(core) do not return error that I can capture. Do you have any specific scenario in mind that I will not capture an error?

tamilmani1989: not sure i followed your comment. returning error is good right if there any issue instead of ignoring it?

AzureAhai: That customize encoder config was similar to  NewProductionEncoderConfig() which is already available in zap. you can edit the fields of the encoder returned by this function.  
like what I did here `encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder` 

And the error catch that has been removed is the result of build() which is no longer being used since we are using lumberjack and it seems we should use `Zapcore.newcore()` and `zap.New(core)` . These functions do not return error like build()

tamilmani1989: got it.. we didn't set EncodeLevel: zapcore.LowercaseLevelEncoder in this new code...is that ok.?

AzureAhai: The constructor that I used is already setting that:
`func NewProductionEncoderConfig() zapcore.EncoderConfig {
	return zapcore.EncoderConfig{
		TimeKey:        "ts",
		LevelKey:       "level",
		NameKey:        "logger",
		CallerKey:      "caller",
		FunctionKey:    zapcore.OmitKey,
		MessageKey:     "msg",
		StacktraceKey:  "stacktrace",
		LineEnding:     zapcore.DefaultLineEnding,
		EncodeLevel:    zapcore.LowercaseLevelEncoder,
		EncodeTime:     zapcore.EpochTimeEncoder,
		EncodeDuration: zapcore.SecondsDurationEncoder,
		EncodeCaller:   zapcore.ShortCallerEncoder,
	}
}`

I also checked the log format and it seems fine. This is a sample:
`{"level":"debug","ts":"2022-08-24T03:33:38.371Z","msg":"Making request to CNS"}`

tamilmani1989 (dismissed):

AzureAhai:
@rbtr Can you review this please? The merge is blocked because of the change request.

azure-ipam/logger/logger.go
```diff
@@ -1,61 +1,53 @@
...
+		MaxBackups:  cfg.MaxBackups,
+	})
+	// define the log encoding
+	encoderConfig := zap.NewProductionEncoderConfig()
+	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	jsonEncoder := zapcore.NewJSONEncoder(encoderConfig)
+	// create a new zap logger
+	core := zapcore.NewCore(jsonEncoder, logFileWriter, logLevel)
+	logger := zap.New(core)
+	return logger
```
rbtr, line 52: nit: i think it's clear enough to `return zap.New(core)` and save the line. vertical space is valuable

azure-ipam/main.go
```diff
@@ -22,9 +22,10 @@ func main() {
 func executePlugin() error {
 	// logger config
 	loggerCfg := &logger.Config{
-		Level:            "debug",
-		OutputPaths:      "var/log/azure-ipam.log",
-		ErrorOutputPaths: "var/log/azure-ipam.log",
+		Level:                "debug",
+		Filepath:             "/var/log/azure-ipam.log",
+		MaxSizeInMB:          5, // MegaBytes
+		MaxBackups:           8,
```
rbtr, line 28: wondering if you did any testing to arrive at these values? I want to make sure we are not rotating the logs too aggressively - it would hard to troubleshoot if that was accidentally like, 10 minutes of logs in a decently sized cluster

rbtr (approved):
nothing blocking, lgtm

azure-ipam/main.go
AzureAhai: I did test that. We don't log much into this file for now. It took alot of time for me with an script to reach to 5 MB. It was @tamilmani1989 suggestion to go with these numbers for now.

tamilmani1989: keeping same as cni log files

PR #1526, opened 2022-08-11
Accept NNC with no NC as valid and consider pool monitor and reconciler started

rbtr:

**Reason for Change**: Due to superficial usage of the NC in the IPAM Pool Monitor, it was considered un-started until it received an NNC with an NC. This change fixes that and considers the reconciler and monitor started as soon as an NNC is ready, whether or not the NC is ready also.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/ipampool/monitor.go
+++ b/cns/ipampool/monitor.go
@@ -121,14 +121,14 @@ func (pm *Monitor) Start(ctx context.Context) error {
 				// if we have initialized and enter this case, we proceed out of the select and continue to reconcile.
 			}
 		case nnc := <-pm.nncSource: // received a new NodeNetworkConfig, extract the data from it and re-reconcile.
-			scaler := nnc.Status.Scaler
-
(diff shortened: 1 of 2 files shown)


cns/ipampool/monitor.go
```diff
@@ -140,6 +140,7 @@ func (pm *Monitor) Start(ctx context.Context) error {
 				}
 			}
 
+			scaler := nnc.Status.Scaler
```
thatmattlong, line 143: If we don't have NCs we probably don't have a `Status` at all yet, does this matter? Will the pool monitor start respecting the real scaler when it comes in later?

thatmattlong (approved):
Approving assuming the answer to my question is that it doesn't matter if we don't have a scaler yet and it will get updated later.

cns/ipampool/monitor.go
rbtr: Yes, next time we get an NNC we will copy the scaler out again, and I think this zero-value scaler is safe to use

rbtr:
bypassing broken aks-e check

PR #1527, opened 2022-08-11
feat: compare node IP with NNCs node IP

thatmattlong:
**Reason for Change**:

CNS shouldn't `CreateOrUpdateNCInternal` a NC that was created for a different Node IP. If the Node IPs don't match, DNC-RC will eventually clean it up, so it is not valid for use.


rbtr (approved):
lgtm ✅

rbtr (approved):

PR #1528, opened 2022-08-11
implement ObjectEncoder

ZetaoZhuang:
feat: support zap.Object in zapai

**Reason for Change**:

currently the zapai evaluates a zapcore.Field into a best-effort string, per the[ fieldStringer ](https://github.com/Azure/azure-container-networking/blob/master/zapai/encoder.go#L88), this bring some troubles to log the zap object. Ex, while logging zap.Object("ex1", &ex1), the result looks like this `"ex1":"&{ex1 2 0 0xc0000ef880}"`  instead of logging the actual object fields ,so I would like to make zapai support zap.Object by implementing the [ObjectEncoder](https://pkg.go.dev/go.uber.org/zap@v1.22.0/zapcore#ObjectEncoder) interface, and leverage [ObjectMarshaler](https://pkg.go.dev/go.uber.org/zap@v1.22.0/zapcore#ObjectMarshaler) to help adding object fields into logging context. However the appinsights.TraceTelemetry.[BaseTelemetry](https://github.com/Microsoft/ApplicationInsights-Go/blob/master/appinsights/telemetry.go#L50) only take (string: string) pair as custom properties,, that is why I only implement the **AddObject** and **AddString** in the interface for now. But I can still implement the rest of methods and convert those non-string and non-object type into a best-effort string. 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr:
@ZetaoZhuang you need to provide much more context in the description on the why and how of this change. It looks like you're polluting this interface with a bunch of unimplemented methods - I don't understand what the point is, and you're going to need to convince me this is the best way to do it.

ZetaoZhuang:
> *

@rbtr I updated the description with more contexts. Let me know if you have any feedbacks. Thanks

rbtr (approved):

PR #1533, opened 2022-08-17
Pipeline support for Go submodules versioned independently of root repo

rbtr:
**Reason for Change**:

Certain components in this (mono)repo are structured as separate Go modules. Go supports independently versioning nested modules in Git repos by tagging them with their path prefix, eg: a Git tag of `zapai/v0.0.1` on this monorepo is understood by Go mod and other tooling to represent the dependency `github.com/Azure/azure-container-networking/zapai@v0.0.1`. Bare tags can coexist with module prefixed tags, and will continue to represent the module located at the root of the repo.

This change updates the Makefile and pipelines to support building and publishing nested modules using their prefixed tags, instead of the root repo tag.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


camrynl (approved):
looks good

tamilmani1989 (dismissed):

camrynl (approved):
good to go 👍

PR #1541, opened 2022-08-23
Add two more CNS endpoints to the CNS client

timraymond:
These are the final two endpoints that DNC uses without going through the CNS client. Adding these makes it possible to use the CNS client exclusively in DNC.

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -30,6 +30,7 @@ const (
 	PathDebugPodContext                      = "/debug/podcontext"
 	PathDebugRestData                        = "/debug/restdata"
 	NumberOfCPUCores                         = "/hostcpucores"
+	NMAgentSupportedAPIs                     = "/network/nmagentsupportedapis"
 )
 
 // NetworkContainer Prefixes
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -29,6 +29,7 @@ const (
 	PathDebugIPAddresses                     = "/debug/ipaddresses"
 	PathDebugPodContext                      = "/debug/podcontext"
 	PathDebugRestData                        = "/debug/restdata"
+	NumberOfCPUCores                         = "/hostcpucores"
 )
 
 // NetworkContainer Prefixes
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -29,8 +29,8 @@ const (
 	PathDebugIPAddresses                     = "/debug/ipaddresses"
 	PathDebugPodContext                      = "/debug/podcontext"
 	PathDebugRestData                        = "/debug/restdata"
-	NumberOfCPUCores                         = "/hostcpucores"
-	NMAgentSupportedAPIs                     = "/network/nmagentsupportedapis"
+	NumberOfCPUCores                         = NumberOfCPUCoresPath
+	NMAgentSupportedAPIs                     = NmAgentSupportedApisPath
 )
 
 // NetworkContainer Prefixes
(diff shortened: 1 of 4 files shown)


cns/client/client.go
```diff
@@ -659,3 +697,56 @@ func (c *Client) UnpublishNC(ctx context.Context, uncr cns.UnpublishNetworkConta
...
 }
+
+// NMAgentSupportedAPIs returns the supported API names from NMAgent. This can
+// be used, for example, to detect whether the node is capable for GRE
+// allocations.
+func (c *Client) NMAgentSupportedAPIs(ctx context.Context) (*cns.NmAgentSupportedApisResponse, error) {
+	// build the request
+	reqBody := &cns.NmAgentSupportedApisRequest{
+		// the IP used below is that of the Wireserver
+		GetNmAgentSupportedApisURL: "http://168.63.129.16/machine/plugins/?comp=nmagent&type=GetSupportedApis",
```
rbtr, line 708: CNS doesn't use this client-supplied URL anymore, right? That was a potential CSRF

timraymond: Right, I couldn't find any (correct) instance where the URL doesn't resolve to this. The Wireserver IP is pretty much adamant, since it's in documentation far and wide. Also the path & query string is... well, the documented path & query string.

So yeah... I saw no reason to continue allowing it to be adjustable by anyone.

cns/NetworkContainerContract.go
```diff
@@ -29,6 +29,8 @@ const (
 	PathDebugIPAddresses                     = "/debug/ipaddresses"
 	PathDebugPodContext                      = "/debug/podcontext"
 	PathDebugRestData                        = "/debug/restdata"
+	NumberOfCPUCores                         = "/hostcpucores"
+	NMAgentSupportedAPIs                     = "/network/nmagentsupportedapis"
```
rbtr, line 33: these consts already exist already in [`api.go`](https://github.com/Azure/azure-container-networking/blob/master/cns/api.go#L34)

cns/client/client.go
rbtr: I think we're misunderstanding each other - CNS should not accept a URL for it to call in any of its APIs. That's begging for a CSRF exploit. I thought we had found and fixed all of them ([ref](https://github.com/Azure/azure-container-networking/pull/1192/files)).
Hardcoding the URL in the client is not the move - we have this path in the server already, we can ignore the input from the client especially if it is always the same.

rbtr (changes requested):

cns/NetworkContainerContract.go
timraymond: Yeah, I was of two minds on this one. These seemed like `consts` for the client's consumption. I didn't like the duplication, so I'm more than happy to use the ones in `api.go` :)

cns/client/client.go
timraymond: No, I follow... I just did a poor job acknowledging the server-side aspect in my previous comment. We do still need these client changes as-is to have a compatible client with existing CNS deployments. I'd like to do the server-side changes in a separate PR, if that's alright.

rbtr: Okay, gotcha. That sounds good to me.

cns/NetworkContainerContract.go
rbtr: I think we should only have one instance of these consts that there is no chance that the server and client can drift apart. WDYT about just redefining these (in whichever direction prevents import cycles)?
eg `NumberOfCPUCores = api.NumberOfCPUCoresPath`

timraymond: @rbtr I changed it over to that, and I like it better since they at least can't diverge which is 💯 . Turns out they're both in the same package too, so yay no import cycles... but it kind of makes this second block's existence a little more dubious. I'm okay with letting it live for now if ultimately there's a future where we just generate this client entirely from swagger docs.

rbtr (dismissed):

rbtr (dismissed):

rbtr (approved):

PR #1543, opened 2022-08-24
fix: [NPM-WIN] lock the policyMap & updatePodCache

huntergregory:
When applying lots of pod events and policy events at once, there's a chance of a fatal concurrent write to a map from the both the Pod Controller and NetPol controller. _**This is strictly a Windows issue**_ because in Linux, only the NetPol controller's code path will touch the PolicyManager, but in Windows, we also add/remove any necessary policies to a pod's endpoint when the pod's IPSets are updated.
Now we lock the policyMap to prevent this from happening.

This PR also locks the updatePodCache (_**only used by Windows**_) because sometimes we would update a pod twice. We haven't seen crashes for this before, but there could be a chance of concurrent map access here too, and either way, we shouldn't make a repeated OS call.
ApplyDataplane() can be called by the Pod controller and NetPol controller code paths, so in the old code, a pod could be updated while the other thread is waiting for HNS to respond.


npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -349,8 +357,8 @@ func (dp *DataPlane) setNetworkIDByName(networkName string) error {
 }
 
 func (dp *DataPlane) getAllEndpointIDs() []string {
-	endpointIDs := make([]string, 0, len(dp.endpointCache))
-	for _, endpoint := range dp.endpointCache {
+	endpointIDs := make([]string, 0, len(dp.endpointCache.cache))
```
vakalapa, line 360: epCache lock unlock here.

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -118,7 +118,7 @@ func (dp *DataPlane) updatePod(pod *updateNPMPod) error {
 	}
 
 	// Check if pod is already present in cache
-	endpoint, ok := dp.endpointCache[pod.PodIP]
+	endpoint, ok := dp.endpointCache.cache[pod.PodIP]
```
vakalapa, line 121: Need to lock and unlock here

npm/pkg/dataplane/dataplane.go
```diff
@@ -23,6 +24,24 @@ type Config struct {
 	*policies.PolicyManagerCfg
 }
 
+type updatePodCache struct {
```
vakalapa, line 27: Move both updatepodcache and endpointcache to types.go and have Get, Set functions which take care of lock unlock inside.

npm/pkg/dataplane/policies/policymanager.go
```diff
@@ -36,14 +36,45 @@ type PolicyManagerCfg struct {
 }
 
 type PolicyMap struct {
+	mutex sync.RWMutex
 	cache map[string]*NPMNetworkPolicy
 }
 
+func (m *PolicyMap) RLock() {
+	if util.IsWindowsDP() {
```
vakalapa, line 44: Dont need this check,

huntergregory: removed

npm/pkg/dataplane/dataplane.go
huntergregory: discussed offline that this is irrelevant now

npm/pkg/dataplane/dataplane_windows.go
huntergregory: fixed locking

npm/pkg/dataplane/dataplane_windows.go
huntergregory: this is used only at bootup, so no need to use locks

vakalapa (approved):

PR #1544, opened 2022-08-24
NC refresh optimization

bohuini:
Added API on CNS:
- If received  "GET": Return all NCs in CNS's state file to DNC in order to check if NC refresh is needed
- If received "POST": Store all the NCs (from the request body that client sent) into CNS's state file

Added test for the two handler above.

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -17,11 +17,11 @@ const (
 	SetOrchestratorType                      = "/network/setorchestratortype"
 	CreateOrUpdateNetworkContainer           = "/network/createorupdatenetworkcontainer"
 	DeleteNetworkContainer                   = "/network/deletenetworkcontainer"
-	GetNetworkContainerStatus                = "/network/getnetworkcontainerstatus"
 	PublishNetworkContainer                  = "/network/publishnetworkcontainer"
 	UnpublishNetworkContainer                = "/network/unpublishnetworkcontainer"
 	GetInterfaceForContainer                 = "/network/getinterfaceforcontainer"
 	GetNetworkContainerByOrchestratorContext = "/network/getnetworkcontainerbyorchestratorcontext"
+	NetworkContainersURLPath                 = "/network/networkcontainers"
 	AttachContainerToNetwork                 = "/network/attachcontainertonetwork"
 	DetachContainerFromNetwork               = "/network/detachcontainerfromnetwork"
 	RequestIPConfig                          = "/network/requestipconfig"
@@ -340,12 +340,12 @@ type CreateNetworkContainerResponse struct {
 	Response Response
 }
 
-// GetNetworkContainerStatusRequest specifies the details about the request to retrieve status of a specifc network container.
+// GetNetworkContainerStatusRequest specifies the details about the request to retrieve status of a specific network container.
 type GetNetworkContainerStatusRequest struct {
 	NetworkContainerid string
 }
 
-// GetNetworkContainerStatusResponse specifies response of retriving a network container status.
+// GetNetworkContainerStatusResponse specifies response of retrieving a network container status.
 type GetNetworkContainerStatusResponse struct {
 	NetworkContainerid string
 	Version            string
@@ -353,13 +353,26 @@ type GetNetworkContainerStatusResponse struct {
 	Response           Response
 }
 
-// GetNetworkContainerRequest specifies the details about the request to retrieve a specifc network container.
+type GetAllNetworkContainersResponse struct {
+	NetworkContainers []GetNetworkContainerResponse
+	Response          Response
+}
+
+type PostNetworkContainersRequest struct {
+	CreateNetworkContainerRequests []CreateNetworkContainerRequest
+}
+
+type PostNetworkContainersResponse struct {
+	Response Response
+}
+
+// GetNetworkContainerRequest specifies the details about the request to retrieve a specific network container.
 type GetNetworkContainerRequest struct {
 	NetworkContainerid  string
 	OrchestratorContext json.RawMessage
 }
 
-// GetNetworkContainerResponse describes the response to retrieve a specifc network container.
+// GetNetworkContainerResponse describes the response to retrieve a specific network container.
 type GetNetworkContainerResponse struct {
 	NetworkContainerID         string
 	IPConfiguration            IPConfiguration
@@ -436,12 +449,12 @@ type IPAddressState struct {
 	State     string
 }
 
-// DeleteNetworkContainerRequest specifies the details about the request to delete a specifc network container.
+// DeleteNetworkContainerRequest specifies the details about the request to delete a specific network container.
 type DeleteNetworkContainerRequest struct {
 	NetworkContainerid string
 }
 
-// DeleteNetworkContainerResponse describes the response to delete a specifc network container.
+// DeleteNetworkContainerResponse describes the response to delete a specific network container.
 type DeleteNetworkContainerResponse struct {
 	Response Response
 }
@@ -470,7 +483,7 @@ type DetachContainerFromNetworkResponse struct {
 	Response Response
 }
 
-// NetworkInterface specifies the information that can be used to unquely identify an interface.
+// NetworkInterface specifies the information that can be used to uniquely identify an interface.
 type NetworkInterface struct {
 	Name      string
 	IPAddress string
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -305,7 +305,7 @@ type IPConfiguration struct {
 // SecondaryIPConfig contains IP info of SecondaryIP
 type SecondaryIPConfig struct {
 	IPAddress string
-	// NCVesion will help in determining whether IP is in pending programming or available when reconciling.
+	// NCVersion will help in determining whether IP is in pending programming or available when reconciling.
 	NCVersion int
 }
 
@@ -353,15 +353,18 @@ type GetNetworkContainerStatusResponse struct {
 	Response           Response
 }
 
+// GetAllNetworkContainersResponse specifies response of retrieving all NCs from CNS during the process of NC refresh association.
 type GetAllNetworkContainersResponse struct {
 	NetworkContainers []GetNetworkContainerResponse
 	Response          Response
 }
 
+// PostNetworkContainersRequest specifies the request of creating all NCs that are sent from DNC.
 type PostNetworkContainersRequest struct {
 	CreateNetworkContainerRequests []CreateNetworkContainerRequest
 }
 
+// PostNetworkContainersResponse specifies response of creating all NCs that are sent from DNC.
 type PostNetworkContainersResponse struct {
 	Response Response
 }
@@ -386,14 +389,12 @@ type GetNetworkContainerResponse struct {
 	AllowNCToHostCommunication bool
 }
 
-// DeleteNetworkContainerRequest specifies the details about the request to delete a specifc network container.
 type PodIpInfo struct {
 	PodIPConfig                     IPSubnet
 	NetworkContainerPrimaryIPConfig IPConfiguration
 	HostPrimaryIPInfo               HostIPInfo
 }
 
-// DeleteNetworkContainerRequest specifies the details about the request to delete a specifc network container.
 type HostIPInfo struct {
 	Gateway   string
 	PrimaryIP string
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -5,7 +5,6 @@ package restserver
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -23,6 +22,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/wireserver"
 	"github.com/Azure/azure-container-networking/common"
 	"github.com/Azure/azure-container-networking/platform"
+	"github.com/pkg/errors"
 )
 
 var (
@@ -54,7 +54,7 @@ func (service *HTTPRestService) setEnvironment(w http.ResponseWriter, r *http.Re
 	}
 
 	switch r.Method {
-	case "POST":
+	case http.MethodPost:
 		logger.Printf("[Azure CNS]  POST received for SetEnvironment.")
 		service.state.Location = req.Location
 		service.state.NetworkType = req.NetworkType
@@ -88,7 +88,7 @@ func (service *HTTPRestService) createNetwork(w http.ResponseWriter, r *http.Req
 			returnCode = types.InvalidParameter
 		} else {
 			switch r.Method {
-			case "POST":
+			case http.MethodPost:
 				dc := service.dockerClient
 				rt := service.routingTable
 				err = dc.NetworkExists(req.NetworkName)
@@ -190,7 +190,7 @@ func (service *HTTPRestService) deleteNetwork(w http.ResponseWriter, r *http.Req
 	}
 
 	switch r.Method {
-	case "POST":
+	case http.MethodPost:
 		dc := service.dockerClient
 		err := dc.NetworkExists(req.NetworkName)
 
@@ -249,7 +249,7 @@ func (service *HTTPRestService) createHnsNetwork(w http.ResponseWriter, r *http.
 		returnCode = types.InvalidParameter
 	} else {
 		switch r.Method {
-		case "POST":
+		case http.MethodPost:
 			if err := hnsclient.CreateHnsNetwork(req); err == nil {
 				// Save the newly created HnsNetwork name. CNS deleteHnsNetwork API
 				// will only allow deleting these networks.
@@ -300,7 +300,7 @@ func (service *HTTPRestService) deleteHnsNetwork(w http.ResponseWriter, r *http.
 		returnCode = types.InvalidParameter
 	} else {
 		switch r.Method {
-		case "POST":
+		case http.MethodPost:
 			networkInfo, found := service.getNetworkInfo(req.NetworkName)
 			if found && networkInfo.NetworkName == req.NetworkName {
 				if err = hnsclient.DeleteHnsNetwork(req.NetworkName); err == nil {
@@ -357,7 +357,7 @@ func (service *HTTPRestService) reserveIPAddress(w http.ResponseWriter, r *http.
 	}
 
 	switch r.Method {
-	case "POST":
+	case http.MethodPost:
 		ic := service.ipamClient
 
 		var ifInfo *wireserver.InterfaceInfo
@@ -434,7 +434,7 @@ func (service *HTTPRestService) releaseIPAddress(w http.ResponseWriter, r *http.
 	}
 
 	switch r.Method {
-	case "POST":
+	case http.MethodPost:
 		ic := service.ipamClient
 
 		var ifInfo *wireserver.InterfaceInfo
@@ -490,7 +490,7 @@ func (service *HTTPRestService) getHostLocalIP(w http.ResponseWriter, r *http.Re
 
 	if service.state.Initialized {
 		switch r.Method {
-		case "GET":
+		case http.MethodGet:
 			switch service.state.NetworkType {
 			case "Underlay":
 				if service.wscli != nil {
@@ -543,7 +543,7 @@ func (service *HTTPRestService) getIPAddressUtilization(w http.ResponseWriter, r
 	var unhealthyAddrs []string
 
 	switch r.Method {
-	case "GET":
+	case http.MethodGet:
 		ic := service.ipamClient
 
 		ifInfo, err := service.getPrimaryHostInterface(context.TODO())
@@ -601,11 +601,6 @@ func (service *HTTPRestService) getAvailableIPAddresses(w http.ResponseWriter, r
 	logger.Printf("[Azure CNS] getAvailableIPAddresses")
 	logger.Request(service.Name, "getAvailableIPAddresses", nil)
 
-	switch r.Method {
-	case "GET":
-	default:
-	}
-
 	resp := cns.Response{ReturnCode: 0}
 	ipResp := &cns.GetIPAddressesResponse{Response: resp}
 	err := service.Listener.Encode(w, &ipResp)
@@ -617,11 +612,6 @@ func (service *HTTPRestService) getAvailableIPAddresses(w http.ResponseWriter, r
 func (service *HTTPRestService) getReservedIPAddresses(w http.ResponseWriter, r *http.Request) {
 	logger.Printf("[Azure CNS] getReservedIPAddresses")
 	logger.Request(service.Name, "getReservedIPAddresses", nil)
-
-	switch r.Method {
-	case "GET":
-	default:
-	}
 
 	resp := cns.Response{ReturnCode: 0}
 	ipResp := &cns.GetIPAddressesResponse{Response: resp}
@@ -642,7 +632,7 @@ func (service *HTTPRestService) getUnhealthyIPAddresses(w http.ResponseWriter, r
 	var unhealthyAddrs []string
 
 	switch r.Method {
-	case "GET":
+	case http.MethodGet:
 		ic := service.ipamClient
 
 		ifInfo, err := service.getPrimaryHostInterface(context.TODO())
@@ -698,11 +688,6 @@ func (service *HTTPRestService) getAllIPAddresses(w http.ResponseWriter, r *http
 	logger.Printf("[Azure CNS] getAllIPAddresses")
 	logger.Request(service.Name, "getAllIPAddresses", nil)
 
-	switch r.Method {
-	case "GET":
-	default:
-	}
-
 	resp := cns.Response{ReturnCode: 0}
 	ipResp := &cns.GetIPAddressesResponse{Response: resp}
 	err := service.Listener.Encode(w, &ipResp)
@@ -714,11 +699,6 @@ func (service *HTTPRestService) getAllIPAddresses(w http.ResponseWriter, r *http
 func (service *HTTPRestService) getHealthReport(w http.ResponseWriter, r *http.Request) {
 	logger.Printf("[Azure CNS] getHealthReport")
 	logger.Request(service.Name, "getHealthReport", nil)
-
-	switch r.Method {
-	case "GET":
-	default:
-	}
 
 	resp := &cns.Response{ReturnCode: 0}
 	err := service.Listener.Encode(w, &resp)
@@ -798,7 +778,7 @@ func (service *HTTPRestService) createOrUpdateNetworkContainer(w http.ResponseWr
 	var returnCode types.ResponseCode
 	var returnMessage string
 	switch r.Method {
-	case "POST":
+	case http.MethodPost:
 		if req.NetworkContainerType == cns.WebApps {
 			// try to get the saved nc state if it exists
 			existing, ok := service.getNetworkContainerDetails(req.NetworkContainerid)
@@ -899,6 +879,27 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
 	logger.Response(service.Name, getNetworkContainerResponse, returnCode, err)
 }
 
+// getOrRefreshNetworkContainers is to check whether refresh association is needed.
+// If received  "GET": Return all NCs in CNS's state file to DNC in order to check if NC refresh is needed
+// If received "POST": Store all the NCs (from the request body that client sent) into CNS's state file
+func (service *HTTPRestService) getOrRefreshNetworkContainers(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] getOrRefreshNetworkContainers")
+
+	switch r.Method {
+	case http.MethodGet:
+		service.handleGetNetworkContainers(w)
+		return
+	case http.MethodPost:
+		service.handlePostNetworkContainers(w, r)
+		return
+	default:
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		err := errors.New("[Azure CNS] getOrRefreshNetworkContainers did not receive a GET or POST")
+		logger.Response(service.Name, nil, types.InvalidParameter, err)
+		return
+	}
+}
+
 func (service *HTTPRestService) deleteNetworkContainer(w http.ResponseWriter, r *http.Request) {
 	logger.Printf("[Azure CNS] deleteNetworkContainer")
 
@@ -918,7 +919,7 @@ func (service *HTTPRestService) deleteNetworkContainer(w http.ResponseWriter, r 
 	}
 
 	switch r.Method {
-	case "POST":
+	case http.MethodPost:
 		var containerStatus containerstatus
 		var ok bool
 
@@ -1069,7 +1070,7 @@ func (service *HTTPRestService) getNumberOfCPUCores(w http.ResponseWriter, r *ht
 	)
 
 	switch r.Method {
-	case "GET":
+	case http.MethodGet:
 		num = runtime.NumCPU()
 	default:
 		errMsg = "[Azure-CNS] getNumberOfCPUCores API expects a GET."
@@ -1170,7 +1171,7 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
 	}
 
 	switch r.Method {
-	case "POST":
+	case http.MethodPost:
 		// Join the network
 		// Please refactor this
 		// do not reuse the below variable between network join and publish
@@ -1298,7 +1299,7 @@ func (service *HTTPRestService) unpublishNetworkContainer(w http.ResponseWriter,
 	}
 
 	switch r.Method {
-	case "POST":
+	case http.MethodPost:
 		// Join Network if not joined already
 		isNetworkJoined = service.isNetworkJoined(req.NetworkID)
 		if !isNetworkJoined {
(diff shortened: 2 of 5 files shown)


ramiro-gamarra (changes requested):

cns/restserver/api.go
```diff
@@ -899,6 +902,147 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
 
+func (service *HTTPRestService) getAllNetworkContainers(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] getAllNetworkContainers")
+
+	var (
+		req           cns.GetAllNetworkContainersRequest
+		returnCode    types.ResponseCode
+		returnMessage string
+		nodeID        string
+	)
```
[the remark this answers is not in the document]
rbtr: it's a terrible pattern 😢

cns/NetworkContainerContract.go
```diff
@@ -22,6 +22,8 @@ const (
 	UnpublishNetworkContainer                = "/network/unpublishnetworkcontainer"
 	GetInterfaceForContainer                 = "/network/getinterfaceforcontainer"
 	GetNetworkContainerByOrchestratorContext = "/network/getnetworkcontainerbyorchestratorcontext"
+	GetAllNetworkContainers                  = "/network/getallnetworkcontainers"
+	ReplenishNetworkContainer                = "/network/replenishnetworkcontainer"
```
rbtr, line 26: I get trying match the style of the existing endpoints, but we really should be making any new endpoints as good REST URLs. ([This might be a good intro doc](https://apiguide.readthedocs.io/en/latest/build_and_publish/use_RESTful_urls.html), but there is a LOT of info out there and opinions on what that means, exactly.)

As a concrete example, these GET endpoint should simply be:
```suggestion
	GetAllNetworkContainers                  = "/network/networkcontainers"
	ReplenishNetworkContainer                = "/network/networkcontainers/<something?>"
```
I'm trying to understand what `replenish` means in this context exactly, but I'm thinking that it should actually be a PUT on the same URI (`/network/networkcontainers`) with a payload containing the NCs that need to be updated - this request is coming from DNC, which has that data already, right? No need to call CNS to tell it to call DNC, from DNC.

rbtr (changes requested):

cns/restserver/api.go
```diff
@@ -899,6 +899,109 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+	allNcExist := false
+
+	switch r.Method {
+	case "GET":
+		ncDetails, allNcExist = service.getAllNetworkContainersDetails(req.NetworkContainersID)
+
+	default:
+		returnMessage = "[Azure CNS] Error. getAllNetworkContainers did not receive a GET."
+		returnCode = types.InvalidParameter
+	}
```
rbtr, line 927: I have concerns about the dataflow here, it's pretty convoluted for how simple this request response is.
What if, early in the handler, we do a simple: 
```go
if r.Method != "GET" { return http.MethodNotAllowed } // but write the response 
``` 
and otherwise proceed with serving the request. Then there is no need to preallocate vars, maybe assign them in a switch, and return something later when it's not actually clear to the reader what has happened.
Plus, in the case where we do not handle the request, we should not return a half-populated response struct. We should *only* return an error.

cns/restserver/api.go
```diff
@@ -899,6 +899,109 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+				}
+			}
+		}
+
+		returnCode, returnMessage = service.saveNetworkContainerGoalState(req)
+
+	default:
+		returnMessage = "[Azure CNS] Error. replenishNetworkContainer did not receive a POST."
+		returnCode = types.InvalidParameter
+	}
```
rbtr, line 992: Same comment about the dataflow. Rethink this and optimize it into a single path.

cns/restserver/util.go
```diff
@@ -616,6 +616,37 @@ func (service *HTTPRestService) getNetworkContainerDetails(networkContainerID st
...
+	allNcExist := true
+	containerDetails := make(map[string]cns.ContainerStatus)
+
+	for _, ncID := range networkContainersID {
+		ncDetails, ncExists := service.state.ContainerStatus[ncID]
+		if !ncExists {
+			allNcExist = false
+			break
+		}
+
```
rbtr, line 632: nit: tighten up, remove extra vertical whitespace. It is okay to separate logical blocks, but use empty lines very very sparingly.

cns/restserver/util.go
```diff
@@ -616,6 +616,37 @@ func (service *HTTPRestService) getNetworkContainerDetails(networkContainerID st
...
+func storeNcStatus(ncDetails containerstatus) cns.ContainerStatus {
+	var containerDetails cns.ContainerStatus
+	containerDetails.ID = ncDetails.ID
+	containerDetails.VMVersion = ncDetails.VMVersion
+	containerDetails.HostVersion = ncDetails.HostVersion
+	containerDetails.CreateNetworkContainerRequest = ncDetails.CreateNetworkContainerRequest
+	containerDetails.VfpUpdateComplete = ncDetails.VfpUpdateComplete
+
+	return containerDetails
```
rbtr, line 647: pls inline and use the short form for struct instantiation:
```suggestion
	return cns.ContainerStatus{
		ID: ncDetails.ID,
		VMVersion: ncDetails.VMVersion,
		HostVersion: ncDetails.HostVersion,
		CreateNetworkContainerRequest: ncDetails.CreateNetworkContainerRequest,
		VfpUpdateComplete: ncDetails.VfpUpdateComplete,
	}
```

rbtr (changes requested):

cns/restserver/api.go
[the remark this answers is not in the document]
bohuini: Addressed.

cns/restserver/api.go
```diff
@@ -899,6 +902,147 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+	if req.NodeID != nodeID {
+		returnMessage = fmt.Sprintf("[Azure CNS] Invalid request for node %s", nodeID)
+		returnCode = types.InvalidRequest
+	}
+
+	// Get all network containers' ID in CNS
+	networkContainerIDs := make([]string, len(service.state.ContainerStatus))
+	i := 0
+
+	for NcID := range service.state.ContainerStatus {
```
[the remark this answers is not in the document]
bohuini: Addressed.

cns/restserver/api.go
bohuini: Addressed.

cns/restserver/api.go
bohuini: Addressed.

cns/restserver/util.go
bohuini: Fixed.

cns/restserver/util.go
bohuini: Fixed.

cns/NetworkContainerContract.go
bohuini: Addressed. Thanks for the link. It's very good to know.

cns/restserver/api.go
```diff
@@ -899,6 +902,147 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+		networkContainerIDs[i] = NcID
+		i++
+	}
+
+	response := cns.GetAllNetworkContainersResponse{
+		NetworkContainerIDs: networkContainerIDs,
+		Response: cns.Response{
+			ReturnCode: returnCode,
+			Message:    returnMessage,
```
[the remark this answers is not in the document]
bohuini: Addressed.

cns/restserver/api.go
```diff
@@ -899,6 +902,147 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+	logger.Request(service.Name, &req, err)
+	if err != nil {
+		return
+	}
+
+	nodeID = service.state.NodeID
+	if req.NodeID != nodeID {
+		returnMessage = fmt.Sprintf("[Azure CNS] Invalid request for node %s", nodeID)
+		returnCode = types.InvalidRequest
+	}
```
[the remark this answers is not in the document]
bohuini: Addressed.

cns/NetworkContainerContract.go
```diff
@@ -338,26 +340,47 @@ type CreateNetworkContainerResponse struct {
...
+type GetAllNetworkContainersResponse struct {
+	NetworkContainerIDs []string
+	Response            Response
+}
+
+type ReplenishNetworkContainerRequest struct {
+	DncEP     string
+	InfraVnet string
+	NodeID    string
+}
```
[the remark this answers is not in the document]
bohuini: No, CNS does **not** reach out to DNC. The reason I use this pattern was to try to achieve the operation of `GET` the NCs details from DNC. Then I found out, as a matter of fact, this can be a `POST` on the same URL in the first API whose request body already has a payload containing the NCs that need to be updated. We can store all NCs without calling DNC.

cns/NetworkContainerContract.go
```diff
@@ -338,26 +340,47 @@ type CreateNetworkContainerResponse struct {
...
 
-// GetNetworkContainerRequest specifies the details about the request to retrieve a specifc network container.
+// GetAllNetworkContainersRequest specifies the details about the request to retrieve all network containers on CNS.
+type GetAllNetworkContainersRequest struct {
+	NodeID string
+}
+
+// GetAllNetworkContainersResponse returns all network containers' ID on the node.
+type GetAllNetworkContainersResponse struct {
+	NetworkContainerIDs []string
```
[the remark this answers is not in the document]
bohuini: I was planning to check if there is any NC mis-matched or lost, by only comparing the NC IDs. Or even just simply compare the size/numbers of how many NCs that DNC & CNS each has.
However, if there are any occasions when the NC ID matches, but the other GS's properties do not match, then this will be terrible solution.
Hence it is better to return all details of NC. Approach updated.

cns/NetworkContainerContract.go
```diff
@@ -338,26 +340,47 @@ type CreateNetworkContainerResponse struct {
...
 	Version            string
 	AzureHostVersion   string
 	Response           Response
 }
 
-// GetNetworkContainerRequest specifies the details about the request to retrieve a specifc network container.
+// GetAllNetworkContainersRequest specifies the details about the request to retrieve all network containers on CNS.
+type GetAllNetworkContainersRequest struct {
+	NodeID string
+}
```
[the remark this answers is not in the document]
bohuini: If I'm understanding the flow correctly, DNC will send a `GET` request to CNS to obtain all the NCs' details in the response. Since we are looping over the nodes on DNC (one call per node), initially I was thinking to retrieve all the NCs on a particular node by node ID.
What I am not understand here is that what does the `check for the http method in the handler` refer to? And what does the request body from DNC look like?
For now, I changed it to `ncIDs[]` in the request. The details of each NC in the CNS will be returned by the NC ID.

cns/NetworkContainerContract.go
```diff
@@ -17,11 +17,12 @@ const (
 	SetOrchestratorType                      = "/network/setorchestratortype"
 	CreateOrUpdateNetworkContainer           = "/network/createorupdatenetworkcontainer"
 	DeleteNetworkContainer                   = "/network/deletenetworkcontainer"
-	GetNetworkContainerStatus                = "/network/getnetworkcontainerstatus"
 	PublishNetworkContainer                  = "/network/publishnetworkcontainer"
 	UnpublishNetworkContainer                = "/network/unpublishnetworkcontainer"
 	GetInterfaceForContainer                 = "/network/getinterfaceforcontainer"
 	GetNetworkContainerByOrchestratorContext = "/network/getnetworkcontainerbyorchestratorcontext"
+	GetAllNetworkContainers                  = "/network/networkcontainers"
+	NetworkContainerRefreshAssociation       = "/network/networkcontainerrefreshassociation"
```
ramiro-gamarra, line 25: Saw another comment to this effect which I agree with: a more RESTful approach is to think of these as operations on a `networkcontainers` collection:
- GET `/network/networkcontainers` returns all the NCs in CNS's state file
- PUT `/network/networkcontainers` allows client to send a request body with the NCs that need to be stored in CNS's state file. (Could also be a POST IMO since there's no standard for bulk request as far as I'm aware of).

This would be a bit more declarative if CNS used an http routing library, but in this case it's a bit barebones, so to be clear, I'm suggesting:
- A single route, no `/network/networkcontainerrefreshassociation`
- A single handler for the route, which dispatches to appropriate methods:
```go
func (s *HTTPRestService) networkContainers(w http.ResponseWriter, r *http.Request){
    switch r.Method {
    case http.MethodGet:
        getNetworkContainers(w,r)
    case http.MethodGet:
        putNetworkContainers(w,r)
    default:
        w.WriteHeader(http.StatusMethodNotAllowed)
    }
}
```

cns/restserver/api.go
```diff
@@ -899,6 +899,95 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+// getAllNetworkContainers is used to check if NC refresh association is needed
+// If any NC from DNC request is not existing in CNS, refresh association will be triggered
+func (service *HTTPRestService) getAllNetworkContainers(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] getAllNetworkContainers")
+	if r.Method != "GET" {
+		logger.Printf("[Azure CNS] Error. getAllNetworkContainers did not receive a GET.")
+		return
+	}
+
+	var req cns.GetAllNetworkContainersRequest
```
ramiro-gamarra, line 911: - GET requests usually don't involve bodies. All the necessary information the server needs to handle the request should come from the URI.
- The purpose of this functionality is for DNC to learn about **any** NCs CNS is aware of. No need for this endpoint to take individual NC ids. If in the future that needs to be supported, they should be provided via query parameters on the URL.

cns/restserver/api.go
```diff
@@ -899,6 +899,95 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+
+// refreshNetworkContainers receives and stores the GS for all the NCs of node from DNC
+func (service *HTTPRestService) refreshNetworkContainers(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] refreshNetworkContainers")
+	if r.Method != "POST" {
+		logger.Printf("[Azure CNS] Error. refreshNetworkContainers did not receive a POST.")
+		return
+	}
+
+	var req cns.CreateNetworkContainerRequest
```
ramiro-gamarra, line 941: This is information for a single NC. We need a new type that allows the client to send multiple NCs.

cns/NetworkContainerContract.go
bohuini: Merged into one method, and use a single route.
If received "GET":
1. Check NCs in CNS's state file, return "false" if any NC is missing. It means NC refresh association is needed;
2. If all NCs exist, return all NCs in CNS's state file for follow-up check;

If received "POST": Store all the NCs (from the request body that client sent) into CNS's state file

cns/restserver/api.go
bohuini: Addressed.

cns/restserver/api.go
```diff
@@ -899,6 +902,94 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+
+	default:
+		response := cns.GetOrRefreshNetworkContainersResponse{
+			Response: cns.Response{
+				ReturnCode: types.InvalidParameter,
+				Message:    "[Azure CNS] Error. getOrRefreshNetworkContainers did not receive a GET or POST.",
+			},
+		}
+		err := service.Listener.Encode(w, &response)
+		logger.Response(service.Name, response, response.Response.ReturnCode, err)
```
rbtr, line 989: let's just return an http.MethodNotAllowed instead of this struct

cns/restserver/api.go
```diff
@@ -899,6 +902,94 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+
+	case "POST":
+		var req cns.GetOrRefreshNetworkContainersRequest
+		err := service.Listener.Decode(w, r, &req)
+		logger.Request(service.Name, &req, err)
+		if err != nil {
+			logger.Printf("[Azure CNS] getOrRefreshNetworkContainers failed with error: %s", err.Error())
+			return
```
rbtr, line 934: am I missing something, or do these error cases just not return anything to the client?

rbtr (dismissed):

cns/restserver/api.go
bohuini: Addressed.

cns/NetworkContainerContract.go
```diff
@@ -338,26 +338,44 @@ type CreateNetworkContainerResponse struct {
...
+}
+
+type GetOrRefreshNetworkContainersRequest struct {
+	CreateNetworkContainerRequests []CreateNetworkContainerRequest
+}
+
+type GetOrRefreshNetworkContainersResponse struct {
+	NetworkContainersDetails map[string]ContainerStatus // NetworkContainerID is the key.
+	Response                 Response
+}
```
ramiro-gamarra, line 370: Some suggestions about the response struct. Usually, I'm of the opinion that IDs shouldn't be json keys, since they make the response less descriptive. I'd like to hear some thoughts from others, but I would recommend this struct be
```go
type GetNetworkContainersResponse struct {
    NetworkContainers []GetNetworkContainerResponse
    Response          Response
}
```

cns/restserver/api.go
```diff
@@ -899,6 +902,99 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
 }
 
+// getOrRefreshNetworkContainers is to check whether refresh association is needed.
+// If received  "GET": return all NCs in CNS's state file to DNC for follow-up check
+// If received "POST": Store all the NCs (from the request body that client sent) into CNS's state file
+func (service *HTTPRestService) getOrRefreshNetworkContainers(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] getOrRefreshNetworkContainers")
+
+	switch r.Method {
+	case "GET":
```
ramiro-gamarra, line 912: - use constants from `http` package: `http.MethodGet`, `http.MethodPost`
- extract functions specific to the method:
```go
switch r.Method {
case http.MethodGet:
    handleGetNetworkContainers(w, r)
    return
case http.MethodPost:
    handlePostNetworkContainers(w, r)
    return
default:
    // ...
}
```

cns/restserver/api.go
```diff
@@ -899,6 +902,99 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+		// return all NCs in CNS
+		networkContainersDetails := make(map[string]cns.ContainerStatus)
+		for ncID := range service.state.ContainerStatus {
```
ramiro-gamarra, line 915: does this not need a read lock?

cns/restserver/api.go
```diff
@@ -899,6 +902,99 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+			}
+			createNcRequest, requestErr := http.NewRequestWithContext(context.Background(), http.MethodPost, "", bytes.NewBuffer(body))
+			if requestErr != nil {
+				returnCode = types.InvalidRequest
+				returnMessage = fmt.Sprintf("[Azure CNS] Create Network Container failed with error: %s", err.Error())
+				break
+			}
+			createNcRequest.Header.Set(common.ContentType, common.JsonContent)
+			writer := httptest.NewRecorder()
+			service.createOrUpdateNetworkContainer(writer, createNcRequest)
```
ramiro-gamarra, line 961: this does not look right. we shouldn't need to be creating `http.Requests` for this functionality.

cns/restserver/util.go
```diff
@@ -616,6 +616,16 @@ func (service *HTTPRestService) getNetworkContainerDetails(networkContainerID st
 	return containerDetails, containerExists
 }
 
+func storeNcStatus(ncDetails containerstatus) cns.ContainerStatus {
```
ramiro-gamarra, line 619: this reads like the function is storing a nc status. maybe `toCNSContainerStatus` is a more appropriate name?

ramiro-gamarra (dismissed):
more comments

cns/NetworkContainerContract.go
```diff
@@ -17,11 +17,11 @@ const (
 	SetOrchestratorType                      = "/network/setorchestratortype"
 	CreateOrUpdateNetworkContainer           = "/network/createorupdatenetworkcontainer"
 	DeleteNetworkContainer                   = "/network/deletenetworkcontainer"
-	GetNetworkContainerStatus                = "/network/getnetworkcontainerstatus"
 	PublishNetworkContainer                  = "/network/publishnetworkcontainer"
 	UnpublishNetworkContainer                = "/network/unpublishnetworkcontainer"
 	GetInterfaceForContainer                 = "/network/getinterfaceforcontainer"
 	GetNetworkContainerByOrchestratorContext = "/network/getnetworkcontainerbyorchestratorcontext"
+	GetOrRefreshNetworkContainers            = "/network/networkcontainers"
```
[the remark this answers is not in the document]
bohuini: Fixed.

cns/NetworkContainerContract.go
```diff
@@ -338,26 +338,44 @@ type CreateNetworkContainerResponse struct {
...
+// ContainerStatus is to save status of an existing container
+type ContainerStatus struct {
+	ID                            string
+	VMVersion                     string
+	HostVersion                   string
+	CreateNetworkContainerRequest CreateNetworkContainerRequest
+	VfpUpdateComplete             bool // True when VFP programming is completed for the NC
+}
+
+type GetOrRefreshNetworkContainersRequest struct {
```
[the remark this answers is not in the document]
bohuini: Fixed.

cns/restserver/util.go
bohuini: This function is deleted since the struct changed.

cns/restserver/api.go
bohuini: Fixed.

cns/restserver/api.go
bohuini: Yes, add a read lock would be more accurate.

cns/restserver/api.go
bohuini: Implementation updated. Extract functions to specific method.

cns/restserver/api.go
bohuini: Fixed.

cns/restserver/util.go
```diff
@@ -840,3 +840,107 @@ func (service *HTTPRestService) isNCWaitingForUpdate(
...
+			},
+		}
+		err = service.Listener.Encode(w, &response)
+		logger.Response(service.Name, response, response.Response.ReturnCode, err)
+		return
+	}
+
+	returnCode := types.Success
+	returnMessage := ""
+	for i := 0; i <= len(req.CreateNetworkContainerRequests); i++ {
```
ramiro-gamarra, line 896:
```suggestion
	for i := 0; i < len(req.CreateNetworkContainerRequests); i++ {
```
Otherwise you will get an index out of bounds on the last iteration.

cns/restserver/util.go
```diff
@@ -840,3 +840,107 @@ func (service *HTTPRestService) isNCWaitingForUpdate(
...
+		createNcReq := req.CreateNetworkContainerRequests[i]
+		if createNcReq.NetworkContainerType == cns.WebApps {
+			// Try to get the saved NC state if it exists
+			ncDetails, found := service.getNetworkContainerDetails(createNcReq.NetworkContainerid)
+			// Create/update nc only if it doesn't exist, or it exists and the requested version is different from the saved version
```
ramiro-gamarra, line 901: Are these comments pulled from a different section? This code branch only creates, doesn't update.

cns/restserver/util.go
```diff
@@ -840,3 +840,107 @@ func (service *HTTPRestService) isNCWaitingForUpdate(
...
+		logger.Response(service.Name, response, response.Response.ReturnCode, err)
+		return
+	}
+
+	returnCode := types.Success
+	returnMessage := ""
+	for i := 0; i <= len(req.CreateNetworkContainerRequests); i++ {
+		createNcReq := req.CreateNetworkContainerRequests[i]
+		if createNcReq.NetworkContainerType == cns.WebApps {
+			// Try to get the saved NC state if it exists
```
ramiro-gamarra, line 899: IMO code is clear enough, this comment is redundant.

cns/restserver/util.go
```diff
@@ -840,3 +840,107 @@ func (service *HTTPRestService) isNCWaitingForUpdate(
...
+			if !found || (found && ncDetails.VMVersion != createNcReq.Version) {
+				nc := service.networkContainer
+				if err = nc.Create(createNcReq); err != nil {
+					returnCode = types.UnexpectedError
+					returnMessage = fmt.Sprintf("[Azure CNS] Create Network Container failed with error: %s", err.Error())
+					break
+				}
+			}
+		} else if createNcReq.NetworkContainerType == cns.AzureContainerInstance {
+			// Try to get the saved NC state if it exists
```
ramiro-gamarra, line 911: Same here re: comment being redundant.

cns/restserver/util.go
```diff
@@ -840,3 +840,107 @@ func (service *HTTPRestService) isNCWaitingForUpdate(
...
+			ncDetails, found := service.getNetworkContainerDetails(createNcReq.NetworkContainerid)
+			// Create/update nc only if it doesn't exist, or it exists and the requested version is different from the saved version
```
ramiro-gamarra, line 913: Comment is not accurate. This code branch updates if the nc exists AND the version is different.

ramiro-gamarra (commented):
Few more things. Also, we need some tests.

cns/restserver/util.go
bohuini: Fixed.

cns/restserver/util.go
bohuini: Fixed.

cns/restserver/util.go
bohuini: Addressed.

cns/NetworkContainerContract.go
bohuini: Applied.

cns/restserver/api_test.go
```diff
@@ -828,6 +848,149 @@ func TestCreateHostNCApipaEndpoint(t *testing.T) {
...
+			return errMismatchedNCs
+		}
+	}
+
+	fmt.Printf("GetNetworkContainers succeeded with response %+v, raw:%+v\n", resp, w.Body)
+	return nil
+}
+
+func TestPostNetworkContainers(t *testing.T) {
+	t.Run("Test Post Network Containers", func(t *testing.T) {
```
ramiro-gamarra, line 911: same suggestion as above

cns/restserver/api_test.go
```diff
@@ -828,6 +848,149 @@ func TestCreateHostNCApipaEndpoint(t *testing.T) {
...
+	req, err := http.NewRequestWithContext(context.TODO(), http.MethodPost, cns.NetworkContainersURLPath, &body)
+	if err != nil {
+		return fmt.Errorf("postAllNetworkContainers failed with error: %w", err)
+	}
+
+	w := httptest.NewRecorder()
+	mux.ServeHTTP(w, req)
+	var resp cns.PostNetworkContainersResponse
+	err = decodeResponse(w, &resp)
+	fmt.Printf("Raw response: %+v", w.Body)
```
ramiro-gamarra, line 984: please confirm what `w.Body` prints

cns/restserver/util.go
```diff
@@ -840,3 +840,90 @@ func (service *HTTPRestService) isNCWaitingForUpdate(
...
+
+		// If NC was created successfully, log NC snapshot.
+		if saveNcReturnCode == types.Success {
+			logNCSnapshot(createNcReq)
+		} else {
+			returnCode = saveNcReturnCode
+			returnMessage = saveNcReturnMessage
+			break
+		}
+	}
```
ramiro-gamarra, line 919: setting variables and then breaking is difficult to follow. it's almost akin to a `goto` which should be discouraged. please extract this to a function that returns the response type. Something like:
```go
func (s *HTTPRestService) createOrUpdateNetworkContainers(...) cns.Response
```

such that it can be used in this scope like
```go
resp := createOrUpdateNetworkContainers(req.CreateNetworkContainerRequests)
err = service.Listener.Encode(w, &response)
logger.Response(service.Name, response, response.Response.ReturnCode, err)

```

cns/restserver/api.go
```diff
@@ -899,6 +879,27 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
 	logger.Response(service.Name, getNetworkContainerResponse, returnCode, err)
 }
 
+// getOrRefreshNetworkContainers is to check whether refresh association is needed.
+// If received  "GET": Return all NCs in CNS's state file to DNC in order to check if NC refresh is needed
+// If received "POST": Store all the NCs (from the request body that client sent) into CNS's state file
+func (service *HTTPRestService) getOrRefreshNetworkContainers(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] getOrRefreshNetworkContainers")
```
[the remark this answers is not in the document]
bohuini: Added log to specific handler.

cns/NetworkContainerContract.go
```diff
@@ -340,26 +340,39 @@ type CreateNetworkContainerResponse struct {
...
+	Response          Response
+}
+
+type PostNetworkContainersRequest struct {
+	CreateNetworkContainerRequests []CreateNetworkContainerRequest
+}
+
+type PostNetworkContainersResponse struct {
+	Response Response
+}
```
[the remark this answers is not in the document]
bohuini: Added documentation.

cns/restserver/api_test.go
```diff
@@ -828,6 +848,149 @@ func TestCreateHostNCApipaEndpoint(t *testing.T) {
 	fmt.Printf("createHostNCApipaEndpoint Responded with %+v\n", createHostNCApipaEndpointResponse)
 }
 
+func TestGetNetworkContainers(t *testing.T) {
+	t.Run("Test Get Network Containers", func(t *testing.T) {
```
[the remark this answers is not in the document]
bohuini: Removed.

cns/restserver/api_test.go
```diff
@@ -828,6 +848,149 @@ func TestCreateHostNCApipaEndpoint(t *testing.T) {
...
+		setEnv(t)
+		err := setOrchestratorType(t, cns.Kubernetes)
+		if err != nil {
+			t.Errorf("TestGetNetworkContainers failed with error:%+v", err)
+			t.Fatal(err)
```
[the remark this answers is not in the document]
bohuini: Addressed.

cns/restserver/api_test.go
```diff
@@ -828,6 +848,149 @@ func TestCreateHostNCApipaEndpoint(t *testing.T) {
...
+	}
+
+	// If any NC in response is not found in ncParams, it means get all NCs failed
+	for i := 0; i < len(ncParams); i++ {
+		if !contains(resp.NetworkContainers, cns.SwiftPrefix+ncParams[i].ncID) {
+			return errMismatchedNCs
+		}
+	}
+
+	fmt.Printf("GetNetworkContainers succeeded with response %+v, raw:%+v\n", resp, w.Body)
```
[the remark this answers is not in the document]
bohuini: - fixed
- W.body is unformatted JSON data of all NCs. Removed it to reduce redundancy.

cns/restserver/api_test.go
bohuini: Fixed

cns/restserver/api_test.go
bohuini: `W.body` is unformatted JSON data of all NCs. Removed it to reduce redundancy.

cns/restserver/util.go
bohuini: Added a helper function `createNetworkContainers()` that returns the response, making the logic be more clearer.

cns/restserver/util.go
```diff
@@ -840,3 +840,95 @@ func (service *HTTPRestService) isNCWaitingForUpdate(
...
+			ReturnCode: types.Success,
+		},
+	}
+	err := service.Listener.Encode(w, &response)
+	logger.Response(service.Name, response, response.Response.ReturnCode, err)
+}
+
+// handlePostNetworkContainers stores all the NCs (from the request that client sent) into CNS's state file
+func (service *HTTPRestService) handlePostNetworkContainers(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] getOrRefreshNetworkContainers do POST operation and store all the NCs from the request into CNS")
```
ramiro-gamarra, line 880:
```suggestion
	logger.Printf("[Azure CNS] handlePostNetworkContainers")
```

ramiro-gamarra (commented):
Only 2 comments regarding logs.

cns/restserver/util.go
```diff
@@ -840,3 +840,95 @@ func (service *HTTPRestService) isNCWaitingForUpdate(
 	}
 	return
 }
+
+// handleGetNetworkContainers returns all NCs in CNS
+func (service *HTTPRestService) handleGetNetworkContainers(w http.ResponseWriter) {
+	logger.Printf("[Azure CNS] getOrRefreshNetworkContainers do GET operation and return all NCs in CNS")
```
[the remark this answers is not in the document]
bohuini: Fixed.

cns/restserver/util.go
bohuini: Fixed.

ramiro-gamarra (approved):

PR #1564, opened 2022-08-31
fix: [NPM] controllers return dataplane error and skip OS calls if doing nothing

huntergregory:
Current pod/namespace controller code doesn't handle dataplane errors. 

Also, a controller doesn't need to apply dataplane changes if it makes none. This could cause an item to requeue if encountering errors when applying other controllers' changes.


npm/pkg/controlplane/controllers/v2/namespaceController.go
```diff
@@ -221,7 +221,7 @@ func (nsc *NamespaceController) processNextWorkItem() bool {
 }
 
 // syncNamespace compares the actual state with the desired, and attempts to converge the two.
-func (nsc *NamespaceController) syncNamespace(nsKey string) error {
+func (nsc *NamespaceController) syncNamespace(nsKey string) (err error) {
```
vakalapa, line 224: Can we fix this golint.

vakalapa (dismissed):

npm/pkg/controlplane/controllers/v2/namespaceController.go
huntergregory: discussed offline that we shouldn't return sync errors when the dataplane fails. The dataplane should reconcile on its own

npm/pkg/dataplane/dataplane.go
```diff
@@ -107,6 +107,11 @@ func (dp *DataPlane) RunPeriodicTasks() {
 				// in Windows, does nothing
 				// in Linux, locks policy manager but can be interrupted
 				dp.policyMgr.Reconcile()
+
+				// the pod and namespace controllers won't retry failed calls to ApplyDataplane
+				if err := dp.ApplyDataPlane(); err != nil {
```
vakalapa, line 112: Is there a possibility for us to applyDP when a diff controller is in between its CRUD event ?

ck319 (approved):

npm/pkg/dataplane/dataplane.go
huntergregory: yes, but I'm not worried about this because we already have 3 controller threads that could call ApplyDataplane at any time

huntergregory: now, we will fix this current behavior. We're now going to prevent another thread from applying changes in between a Controller's CRUD event

huntergregory:
Superseded by #1875

PR #1568, opened 2022-09-01
fix: [WIN-NPM] don't remove policy from cache when updating pod

huntergregory:
When removing a policy during updatePod(), the policy should not be removed from the cache. This would cause the policy to not be found when deleting the policy.

This PR also fixes the prometheus ACL metric for windows.

### DIFF
--- a/npm/pkg/dataplane/policies/policymanager.go
+++ b/npm/pkg/dataplane/policies/policymanager.go
@@ -130,13 +130,18 @@ func (pMgr *PolicyManager) AddPolicy(policy *NPMNetworkPolicy, endpointList map[
 	metrics.RecordACLRuleExecTime(timer) // record execution time regardless of failure
 	if err != nil {
 		// NOTE: in Linux, Prometheus metrics may be off at this point since some ACL rules may have been applied successfully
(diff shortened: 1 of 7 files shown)


npm/pkg/dataplane/policies/policymanager.go
```diff
@@ -152,22 +157,57 @@ func (pMgr *PolicyManager) RemovePolicy(policyKey string, endpointList map[strin
...
+	metrics.DecNumACLRulesBy(policy.numACLRulesProducedInKernel() * numEndpoints)
 
+	// remove policy from cache
 	delete(pMgr.policyMap.cache, policyKey)
 	return nil
 }
 
+// RemovePolicyForEndpoints is identical to RemovePolicy except it will not remove the policy from the cache.
+// This function is intended for Windows only.
+func (pMgr *PolicyManager) RemovePolicyForEndpoints(policyKey string, endpointList map[string]string) error {
```
vakalapa, line 184: Where is this function being used ?

huntergregory: forgot to commit. updated

vakalapa (approved):

PR #1569, opened 2022-09-01
fix: [WIN-NPM] don't apply policy to new pod with same IP as an old pod

huntergregory:
When getting endpoints to apply a policy, ignore IPs that have a different pod key in the ipsetmanager versus the endpoint cache (in case the pod controller hasn't updated the dp yet that the IP's pod owner has changed).

### DIFF
--- a/npm/pkg/dataplane/dataplane-test-cases_windows_test.go
+++ b/npm/pkg/dataplane/dataplane-test-cases_windows_test.go
@@ -370,6 +370,40 @@ func getAllSerialTests() []*SerialTestCase {
 				},
 			},
 		},
+		{
+			Description: "Pod B replaces Pod A with same IP",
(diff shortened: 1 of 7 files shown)


ck319 (dismissed):

matmerr (approved):

huntergregory:
UT would be:
1. Endpoint X created
2. Pod A created w/ Endpoint X's IP and label k1:v1
3. Endpoint X deleted (Pod A's endpoint)
4. Endpoint Y created with the same IP
5. Policy created w/ Pod Selector k1:v1 (before Pod A delete event comes in)

ck319 (approved):

matmerr (approved):

vakalapa (approved):

PR #1575, opened 2022-09-02
Update Makefile to build all binaries

camrynl:
**Reason for Change**:

The target make all-binaries does not currently include building azure-ipam.
Including azure-ipam in the build so that it's available to use in the release pipeline when releasing dropgz.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -589,9 +589,9 @@ endif
 
 # Create a azure-ipam archive for the target platform.
 .PHONY: azure-ipam-archive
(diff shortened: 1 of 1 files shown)


rbtr:
`zapai` is strictly a library, we don't need to build binaries for it (it has no `func main`).

rbtr (approved):

PR #1576, opened 2022-09-06
fix: Remove retry from Windows pipelines

vakalapa:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/containers/container-template-windows.yaml
+++ b/.pipelines/containers/container-template-windows.yaml
@@ -13,7 +13,7 @@ steps:
 
 - powershell: |
     echo "using tag ${{ parameters.tag }}"
-    powershell.exe -command "& { . .\build\scripts\windows.ps1; {${{ parameters.name }}-image windows-${{ parameters.arch }}-${{ parameters.tag }}} }"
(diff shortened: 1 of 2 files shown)


.pipelines/containers/container-template-windows.yaml
```diff
@@ -13,9 +13,10 @@ steps:
 
 - powershell: |
     echo "using tag ${{ parameters.tag }}"
-    powershell.exe -command "& { . .\build\scripts\windows.ps1; Retry({${{ parameters.name }}-image windows-${{ parameters.arch }}-${{ parameters.tag }}}) }"
+    powershell.exe -command "& { . .\build\scripts\windows.ps1; ${{ parameters.name }}-image windows-${{ parameters.arch }}-${{ parameters.tag }} }"
```
rbtr, line 16: don't think we need the subshell invocation here:
```suggestion
    . .\build\scripts\windows.ps1; ( ${{ parameters.name }}-image windows-${{ parameters.arch }}-${{ parameters.tag }} )
```

PR #1588, opened 2022-09-08
Update dropgz to use azure-ipam release

camrynl:
**Reason for Change**:

cni.Dockerfile is currently building local versions of azure-ipam, updating to use the new release.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


dropgz/build/cni.Dockerfile
```diff
@@ -2,9 +2,11 @@ FROM mcr.microsoft.com/oss/cilium/cilium:1.12.1.1 as cilium
...
 FROM mcr.microsoft.com/oss/go/microsoft/golang:1.19 AS azure-ipam
 ARG VERSION
+ARG OS
+ARG ARCH
 WORKDIR /azure-ipam
 COPY ./azure-ipam .
-RUN CGO_ENABLED=0 go build -a -o bin/azure-ipam -trimpath -ldflags "-X main.version="$VERSION"" -gcflags="-dwarflocationlists=true" .
+RUN curl -LO https://github.com/Azure/azure-container-networking/releases/download/azure-ipam%2Fv0.0.2/azure-ipam-$OS-$ARCH-v0.0.2.tgz && tar -xvf azure-ipam-$OS-$ARCH-v0.0.2.tgz
 
 FROM mcr.microsoft.com/oss/go/microsoft/golang:1.19 AS azure-vnet
```
rbtr, line 11: Since we are not doing Go compiling, these probably don't need to be `FROM` a Go base image anymore. 
We could create an intermediate mariner image with tar:
```dockerfile
FROM mcr.microsoft.com/cbl-mariner/base/core:2.0 as tar
tdnf install -y tar
```
then change the azure-ipam/azure-vnet images to
```dockerfile
FROM tar AS azure-ipam
...
```

rbtr (changes requested):

dropgz/build/cni.Dockerfile
rbtr: while I was checking the docs for that, I noticed that it's possible to use external images directly without making them in to intermediate images...so we can remove line 1 entirely, and change the cilium copy line to:
```dockerfile
COPY --from=mcr.microsoft.com/oss/cilium/cilium:1.12.1.1 /opt/cni/bin/cilium-cni pkg/embed/fs
```

camrynl: running into issues after creating the mariner image -- curl for azure-ipam/azure-vnet is now failing with:

> curl: (60) SSL certificate problem: unable to get local issuer certificate

rbtr: probably need to install `ca-certificates`

tamilmani1989: @camrynl lets also cilium image from dropgz since cilium cni going to be installed by cilium agent

tamilmani1989: are we using same dockerfile for e2e pipeline as well.? If so, this wouldn't test PR branch azure-ipam version right?

camrynl: @tamilmani1989 yes this is used in the e2e pipeline to make the image, and the tests are using the tagged version of dropgz and cns

tamilmani1989: in that case we should use separate dockerfile for e2e pipeline so that it uses PR branch version instead of officially release version of azure-ipam or cns

rbtr (approved):

PR #1589, opened 2022-09-08
fix: fixing the CIDR format of getNetworkName() on Windows

AzureAhai:
The issue is that right now with an IP "172.168.1.1/24" networkName "azure" and multitenancy id "1", we expect the function to return "azure-vlan1-172.168.1-0_24" but instead it returns "azure-vlan1-172.168.1-1_24"

This PR is fixing the issue by parsing the CIDR correctly.

### DIFF
--- a/cni/network/network_windows.go
+++ b/cni/network/network_windows.go
@@ -154,8 +154,13 @@ func (plugin *NetPlugin) getNetworkName(netNs string, ipamAddResult *IPAMAddResu
 	// This will happen during ADD call
 	if ipamAddResult != nil && ipamAddResult.ncResponse != nil {
 		// networkName will look like ~ azure-vlan1-172-28-1-0_24
-		subnet := ipamAddResult.ipv4Result.IPs[0].Address
-		networkName := strings.Replace(subnet.String(), ".", "-", -1)
+		ipAddrNet := ipamAddResult.ipv4Result.IPs[0].Address
+		_, ipNet, err := net.ParseCIDR(ipAddrNet.String())
+		if err != nil {
+			log.Printf("Error parsing network CIDR: %v.", err)
+			return "", err
+		}
+		networkName := strings.Replace(ipNet.String(), ".", "-", -1)
 		networkName = strings.Replace(networkName, "/", "_", -1)
 		networkName = fmt.Sprintf("%s-vlan%v-%v", nwCfg.Name, ipamAddResult.ncResponse.MultiTenancyInfo.ID, networkName)
 		return networkName, nil
--- a/cni/network/network_windows.go
+++ b/cni/network/network_windows.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"net/netip"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


cni/network/network_windows.go
```diff
@@ -154,8 +154,13 @@ func (plugin *NetPlugin) getNetworkName(netNs string, ipamAddResult *IPAMAddResu
 	// This will happen during ADD call
 	if ipamAddResult != nil && ipamAddResult.ncResponse != nil {
 		// networkName will look like ~ azure-vlan1-172-28-1-0_24
-		subnet := ipamAddResult.ipv4Result.IPs[0].Address
-		networkName := strings.Replace(subnet.String(), ".", "-", -1)
+		ipAddrNet := ipamAddResult.ipv4Result.IPs[0].Address
+		_, ipNet, err := net.ParseCIDR(ipAddrNet.String())
+		if err != nil {
+			log.Printf("Error parsing network CIDR: %v.", err)
```
tamilmani1989, line 160: log ipAddrNet.String() as well in error case

cni/network/network_windows.go
```diff
@@ -154,8 +154,13 @@ func (plugin *NetPlugin) getNetworkName(netNs string, ipamAddResult *IPAMAddResu
...
+			return "", err
+		}
+		networkName := strings.Replace(ipNet.String(), ".", "-", -1)
```
rbtr, line 163: I think new code should use the `netip` package.
```suggestion
		prefix, err := netip.ParsePrefix(ipAddrNet.String())
		if err != nil {
			log.Printf("Error parsing network CIDR: %v.", err)
			return "", err
		}
		networkName := strings.Replace(prefix.Masked(), ".", "-", -1)
```

cni/network/network_windows_test.go
```diff
@@ -301,3 +301,59 @@ func TestDSRPolciy(t *testing.T) {
...
+
+func TestGetNetworkNameFromCNS(t *testing.T) {
+	plugin, _ := cni.NewPlugin("name", "0.3.0")
+	tests := []struct {
+		name  			string
+		plugin      	*NetPlugin
+		netNs 			string
+		nwCfg 			*cni.NetworkConfig
+		ipamAddResult   *IPAMAddResult
+		want 			string
```
rbtr, line 313: I don't know why the linter is not yelling at you for this, but it needs to be gofmt

AzureAhai: I used tab here and has configured my tab to be 3 spaces! for some reason it didn't follow through. Will use space instead.

cni/network/network_windows.go
AzureAhai: @tamilmani1989 you agree?

tamilmani1989: yep lets use netip package

tamilmani1989 (approved):
lgtm

AzureAhai:
> lgtm

I'll merge it after the multi-tenancy swift test is done

tamilmani1989:
> > lgtm
> 
> I'll merge it after the multi-tenancy swift test is done

Let also e2e and lint succeed

tamilmani1989 (approved):

cni/network/network_windows_test.go
```diff
@@ -301,3 +301,59 @@ func TestDSRPolciy(t *testing.T) {
...
+				},
+			},
+			want: "azure-vlan1-10-240-0-0_24",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			networkName, err := tt.plugin.getNetworkName(tt.netNs, tt.ipamAddResult, tt.nwCfg)
```
vipul-21, line 354: Do we want to cover the error case scenario as well ? Since this test case is checking the happy path right now.

vipul-21 (dismissed):

cni/network/network_windows_test.go
AzureAhai: At this point we only wanted to make sure that we're producing correct CIDR when the CNS response is valid. Seems that the failure path is not that critical for unit test.

rbtr: Checking the failure cases is absolutely critical in unit tests, come on. You should at least be testing that the input validation that you added correctly returns the error that you added when the input is invalid. How else do you know it rejects bad inputs?

rbtr (changes requested):
@vipul-21 was right to call out the incomplete UT, please finish those.

cni/network/network_windows_test.go
AzureAhai: I added the unit test that was asked for this particular case. Yes, in general unit test is critical for non-happy paths. I didn't consider it critical here, since I was informed malformed CIDR is not reaching to this function. Nevertheless, I will add a test case for malformed CIDR since it is the only error check that I have added.

rbtr (approved):

PR #1590, opened 2022-09-08
Remove parsing GatewayIP and passing down to CNI

tamilmani1989:
…ture

**Reason for Change**:

**Issue Fixed**:

This PR fixes azure-ipam support for aks swift overlay cluster. In AKS swift overlay mode, empty gatewayIP is getting passed from cns to azure-ipam and which is expected. Cilium/Azure CNI really doesn't use this gatewayip. So its better to skip parsing this gatewayIP and passing to Azure CNI 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -80,12 +80,12 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 	p.logger.Debug("Received CNS IP config response", zap.Any("response", resp))
 
 	// Get Pod IP and gateway IP from ip config response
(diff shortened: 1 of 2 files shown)


camrynl (approved):

rbtr (approved):
for my curiosity, would we use this in an AzCNIv1 scenario where we NAT traffic through the node?

AzureAhai (approved):

tamilmani1989:
Cilium E2e pipeline

> for my curiosity, would we use this in an AzCNIv1 scenario where we NAT traffic through the node?

nope not really. used only in bridge mode which we deprecated long ago

PR #1593, opened 2022-09-09
align pool to batch during scale up

rbtr:

**Reason for Change**:

Currently the scale-up logic blindly adds +1 batch to whatever the current request is, meaning that if the current request is not aligned to the batch, the post scale-up pool will still not be aligned to the batch. During scale down, we calculate the skew from the batch and use that to realign the pool to a multiple of the batch.
This change copies the mod arithmetic from the scale-down so that we realign to the next greater multiple of the batch size during a scale-up. This may be within the threshold that triggers _another_ scale-up, ie: 
a pool of batch 16 with 15 IPs and 14 assignments will scale up to -> 16, which is less than the minimum free IPs, so it will scale up again to 32 on the next tick. This behavior is seen when coming out of a constrained subnet state, when the batch would have been temporarily adjusted to 1 and thus the requested IP count could be anything.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rsagasthya (dismissed):
Approving the changes. But, looks like the branch is out of date.

rsagasthya (approved):

PR #1597, opened 2022-09-09
pipeline: use platform tag make targets in e2e test calls

camrynl:
**Reason for Change**:

In swifte2e and ciliume2e tests are called with CNS and CNI dropgz versions. Including new make targets for all platform tags to be used in e2e test calls. 

Changing from current:
image

to also include OS/ARCH for clarity, and to keep it consistent with the pattern of image build. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/singletenancy/aks-swift/e2e-step-template.yaml
+++ b/.pipelines/singletenancy/aks-swift/e2e-step-template.yaml
@@ -57,7 +57,7 @@ steps:
       sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
       kubectl cluster-info
       kubectl get po -owide -A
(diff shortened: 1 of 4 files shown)


rbtr:
The existing behavior of using the multiplat tag is correct.
The platform tags are a private implementation detail that no one should know or care about. Once the multiplat manifest is built and that omnibus tag is published, _everything_ downstream of that (customers, or our tests) must use the multiplat tag.
This is a step backwards if we want to test heterogeneous node pools (which we should), and is not a realistic view of the cx deployment because it is a different tag than they would use.
cc @tamilmani1989 I consider this a functional regression.

PR #1598, opened 2022-09-09
Update NetworkContainer struct to include ipv6 for Dualstack

rjdenney:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -95,6 +95,52 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 			NCVersion: int(nc.Version),
 		}
 	}
+
+	// Check to see if there is a v6 Ip in the struct denoting a dualstack container
+	if nc.PrimaryIPv6 != nil {
```
rbtr, line 100: Nothing in this block cares that the IP is an IPv6 address, so I don't understand why we need to duplicate the parsing code which already handles IPv4/IPv6 interchangeable.

rbtr (commented):
It seems like the IP tooling in `netip` (and probably `net` too?) supports IPv6/IPv4 interchangeably. Why do we need separate fields and codepaths for IPv6 in our code? I would rather rely on those packages that already do that right thing, and have our code be v4/v6 agnostic. Duplicating the existing machinery of IPv4 is the wrong approach to IPv6 support, imo, we should instead make the tooling more generic as appropriate so that it deals with _any IPs_, instead of IPv4 IPs or IPv6 IPs but not both.

cns/kubecontroller/nodenetworkconfig/conversion.go
rjdenney: Netip should be able to work on either IPv6 or IPv4 and the function would work on either if you were going for either a purely IPv4 or IPv6 approach. The reason I added this was to be able to differentiate between dualstack overlay and IPv4 overlay since in dualstack the NetworkContainer struct as written in the document includes separate fileds for the v4 and v6 IPs.

This is the doc: [Azure IPv6 Overlay Proposal.docx (sharepoint.com)](https://microsoft.sharepoint.com/:w:/t/Aznet/Ec02v6L2_hVPhBOpl9Idl9sB7_9GDSrI6ToeHaNjAbqquQ?e=kUL8Ck&CID=893DD4D3-6289-4E2F-A85F-EA92B5EA805B&isSPOFile=1&clickparams=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiIyNy8yMjA4MTYwMTQwMiIsIkhhc0ZlZGVyYXRlZFVzZXIiOmZhbHNlfQ%3D%3D)

rbtr (changes requested):
blocked pending design review

PR #1610, opened 2022-09-13
NC Refresh Optimization

bohuini:
Added API on CNS:
If received  "GET": Return all NCs in CNS's state file to DNC in order to check if NC refresh is needed
If received "POST": Store all the NCs (from the request body that client sent) into CNS's state file

Added unit test.


rbtr:
@bohuini is this a conceptually different change than https://github.com/Azure/azure-container-networking/pull/1544? If not, please don't open new PRs for the same feature, continue on the same PR - it provides historical context and documentation.

rbtr:
duplicate

PR #1617, opened 2022-09-14
Increase windows container build timeout to 120m

rbtr:
**Reason for Change**:

Windows container build fails due to timeout frequently at current limit of 60m. Doubles to attempt to mitigate and increase the pipeline pass percentage.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


nairashu:
Let's link the incident or work item which invoked this change in the description.

nairashu (approved):
Let's link the incident or work item which invoked this change in the description.

PR #1619, opened 2022-09-16
Add path to CNS client for DeleteNC

timraymond:
This path was missed in a prior iteration, but is necessary for the client to function properly. Without it, the route cannot be found and thus the request cannot be created.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr (dismissed):
oop

rbtr (approved):

PR #1620, opened 2022-09-16
Fix: Added a metric to monitor the subnet exhaustion state within the Ipam Monitor Pool

### LINKED ISSUE #1605: Reclaim IPs in PodSubnet when subnet is full or close to full
**Component (Azure NPM or Azure CNI):**

Azure CNI + pod subnet (CNS)

**Describe in detail the feature/behavior/change you'd like to see:**  

When a pod subnet has completely or nearly exhausted the IPs available, there should be a way to reclaim some of the unused IPs on different nodes since nodes are allocated IPs in batches of 16 but it does not mean IPs are necessarily in use.

**Orchestrator(e.g. Kubernetes, Docker):**  

Kubernetes

Both

**Anything else you would like to add:**  

nairashu:
… 

**Reason for Change**:
The PR is for adding a  metric to be able to identify if the IPAM Pool Monitor is able to read the subnet exhaustion state change and compare it with the Batch size metric to verify the batch changes match with this state change.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/ipampool/metrics.go
```diff
@@ -132,4 +140,10 @@ func observeIPPoolState(state ipPoolState, meta metaState, labels []string) {
 	ipamPrimaryIPCount.WithLabelValues(labels...).Set(float64(len(meta.primaryIPAddresses)))
 	ipamRequestedIPConfigCount.WithLabelValues(labels...).Set(float64(state.requestedIPs))
 	ipamTotalIPCount.WithLabelValues(labels...).Set(float64(state.totalIPs))
+	if meta.exhausted {
+		ipamSubnetExhaustionState.WithLabelValues(labels...).Set(float64(1))
+	} else {
+		ipamSubnetExhaustionState.WithLabelValues(labels...).Set(float64(0))
+	}
+
```
rbtr, line 148: it's more common in go to start with the base case, so that there is no need of an `else` branch at all. 
```suggestion
	var exhaustionState float64 // = 0 if you want to be really clear 
	if meta.exhausted {
		exhaustionState = 1
	}
	ipamSubnetExhaustionState.WithLabelValues(labels...).Set(exhaustionState)
```
here it's minor, but in cases where the branching logic is very complex, this is much more readable.

rbtr, line 148: it's more common in go to start with the base case, so that there is no need of an `else` branch at all. 
```suggestion
	var exhaustionState float64 // = 0 if you want to be really clear 
	if meta.exhausted {
		exhaustionState = 1
	}
	ipamSubnetExhaustionState.WithLabelValues(labels...).Set(exhaustionState)
```
here it's minor, but in cases where the branching logic is very complex, this is much more readable.

cns/ipampool/metrics.go
```diff
@@ -117,7 +125,7 @@ func init() {
 		ipamPrimaryIPCount,
 		ipamRequestedIPConfigCount,
 		ipamTotalIPCount,
-	)
+		ipamSubnetExhaustionState)
```
rbtr, line 128: the parenthesis needs to be on a newline

cns/ipampool/metrics.go
rbtr, line 148: remove this empty newline

rbtr (commented):
some superficial comments

cns/ipampool/metrics.go
nairashu: Fixed

cns/ipampool/metrics.go
nairashu: Thanks for the suggestion. Changed

nairashu: Removed

rbtr:
@nairashu FYI, if you put `Fixes ...` in the description, it wil close the issue automatically when this is merged. I have removed that from your comment because we want the issue to stay open for now 🙂

cns/ipampool/metrics.go
```diff
@@ -133,4 +142,9 @@ func observeIPPoolState(state ipPoolState, meta metaState) {
 	ipamPrimaryIPCount.WithLabelValues(labels...).Set(float64(len(meta.primaryIPAddresses)))
 	ipamRequestedIPConfigCount.WithLabelValues(labels...).Set(float64(state.requestedIPs))
 	ipamTotalIPCount.WithLabelValues(labels...).Set(float64(state.totalIPs))
+	var exhaustionState float64 = 0
```
thatmattlong, line 145: I would suggest `consts` for these states, as a reader I'm not sure what state `0` or `1` means.

cns/kubecontroller/clustersubnetstate/reconciler.go
```diff
@@ -22,8 +22,10 @@ type Reconciler struct {
 func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	css, err := r.Cli.Get(ctx, req.NamespacedName)
 	if err != nil {
+		cssReconcilerErrorState.Set(float64(1))
```
thatmattlong, line 25: same here, if you have exported `const`s you can use them here and it will be clear what you are setting this to

cns/ipampool/metrics.go
```diff
@@ -133,4 +144,9 @@ func observeIPPoolState(state ipPoolState, meta metaState) {
 	ipamPrimaryIPCount.WithLabelValues(labels...).Set(float64(len(meta.primaryIPAddresses)))
 	ipamRequestedIPConfigCount.WithLabelValues(labels...).Set(float64(state.requestedIPs))
 	ipamTotalIPCount.WithLabelValues(labels...).Set(float64(state.totalIPs))
+	if meta.exhausted {
+		ipamSubnetExhaustionState.WithLabelValues(labels...).Set(float64(subnetIPExhausted))
```
neaggarwMS, line 148: Do we need two metrics? This is a gauge then counter metric, so cant we determine subnet exhaustion based on if the value is 1 or 0? @rbtr  Thoughts?

cns/kubecontroller/clustersubnetstate/metrics.go
```diff
@@ -0,0 +1,26 @@
+package clustersubnetstate
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+const (
+	cssReconcilerError   = 1
```
neaggarwMS, line 9: Can you rename it to be more specific? say SubnetExhaustionCRDWatcherFailed?

cns/ipampool/monitor.go
```diff
@@ -266,7 +267,7 @@ func (pm *Monitor) increasePoolSize(ctx context.Context, meta metaState, state i
 
 	if _, err := pm.nnccli.UpdateSpec(ctx, &tempNNCSpec); err != nil {
 		// caller will retry to update the CRD again
-		return err
+		return fmt.Errorf("[ipam-pool-monitor] Increasing pool size: UpdateCRDSpec failed %w", err)
```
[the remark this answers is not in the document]
timraymond: The other thing with the error strings is that I find it's best to just say what you were trying to do when the error was produced... since that's all the information you have at that exact point in the codebase. For example, something like:

```go
return errors.Wrap(err, "executing UpdateSpec")
```

Something _else_ can add the "increasing pool size:" bit, if that's indeed what it was trying to do upstream. The result is basically an inline stack trace of things you actually care about and can grep for: `[some-service] error reticulating splines: err: increasing pool size: executing UpdateSpec: invalid spec format`

rbtr (approved):

PR #1622, opened 2022-09-16
Update test cns daemonset to reconcile its own state

camrynl:
**Reason for Change**:

enabling ManageEndpointState in configmap and updating cns daemonset in e2e tests so that cns will reconcile its own state

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/integration/manifests/cns/daemonset.yaml
```diff
@@ -41,7 +41,7 @@ spec:
           effect: NoSchedule
       containers:
         - name: cns-container
-          image: acnpublic.azurecr.io/azure-cns:v1.4.32
+          image: acnpublic.azurecr.io/azure-cns:v1.4.33
```
tamilmani1989, line 44: lets use latest v1.4.35

tamilmani1989 (changes requested):

test/integration/manifests/cns/daemonset.yaml
rbtr: this is just a placeholder, the tests update the image to the latest working version on the branch

tamilmani1989 (approved):

rbtr (changes requested):
we still need coverage of the reconcile-from-CNI scenario, I don't think we can update this for all swift E2Es

camrynl:
> we still need coverage of the reconcile-from-CNI scenario, I don't think we can update this for all swift E2Es

If I use different configmaps for cilium and swift, would that keep coverage for reconciling from cni?

tamilmani1989 (approved):
lgtm

rbtr (approved):

PR #1624, opened 2022-09-20
feat: adding mechanism to collect VM metadata in AI core

ZetaoZhuang:
**Reason for Change**:

[task link](https://msazure.visualstudio.com/One/_workitems/edit/15404650)
VM metadata are pretty helpful while devs are troubleshooting the prod issues. So adding a mechanism in AI core to collect VM metadata. These VM infos would be part of the custom dimension fields of TraceTelemetry entity, and sent to AI

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


zapai/core.go
```diff
@@ -1,11 +1,34 @@
 package zapai
 
 import (
+	"github.com/Azure/azure-container-networking/aitelemetry"
+	"github.com/Azure/azure-container-networking/common"
+	"github.com/Azure/azure-container-networking/log"
+	"github.com/Azure/azure-container-networking/processlock"
+	"github.com/Azure/azure-container-networking/store"
```
rbtr, line 8: zapai (as an independent go module) **cannot** depend on `github.com/Azure/azure-container-networking/*` if we want to use it in the components of that module (like CNS, CNI, NPM).

zapai/core.go
```diff
@@ -88,11 +143,34 @@ func (c *Core) Write(entry zapcore.Entry, fields []zapcore.Field) error {
...
+		t.Tags.User().SetId(metadata.VMName)
+		t.Properties[locationStr] = metadata.Location
+		t.Properties[resourceGroupStr] = metadata.ResourceGroupName
+		t.Properties[vmIDStr] = metadata.VMID
+		t.Properties[vmNameStr] = metadata.VMName
+		t.Properties[vmSizeStr] = metadata.VMSize
+		t.Properties[osVersionStr] = metadata.OSVersion
+		t.Properties[subscriptionIDStr] = subscriptionIDStr
+		t.Tags.Session().SetId(metadata.VMID)
+	}
```
rbtr, line 172: why should this be the responsibility of the core, instead of the responsibility of the _user_ of the core?

rbtr (changes requested):
this is unmergeable because it introduces a cyclical dependency. 
even if it did not, I don't think this functionality belongs in the core.

zapai/core.go
ZetaoZhuang: Ya. Was trying to find a central place for it so each user don't need to have it own implementation for collecting VM meta. 
but if we want to use it in in the components of that module (like CNS, CNI, NPM)., the cyclical dependency is a thing

timraymond: Is it possible to do this as a user of the library only? If not, maybe offer a callback mechanism to allow the user of the core to do this? Just an idea... not fully immersed in this to figure out if it's a good one or not.

zapai/go.mod
```diff
@@ -11,8 +12,23 @@ require (
...
+	go.opencensus.io v0.23.0 // indirect
 	go.uber.org/atomic v1.9.0 // indirect
 	go.uber.org/goleak v1.1.12 // indirect
 	go.uber.org/multierr v1.8.0 // indirect
+	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
 )
+
+replace github.com/Microsoft/hcsshim => github.com/vakalapa/hcsshim v0.9.1-0.20211203205307-837d4d06df77
```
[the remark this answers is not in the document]
rbtr: that seems reasonable to me. 
however, this need for this replace is invalid, because we should not be taking a dependency on `github.com/Azure/azure-container-networking` and bringing all that baggage in here

zapai/core.go
rbtr: Something like https://github.com/Azure/azure-container-networking/blob/ce5cc41650653287dca7cf96823b3ea66e9bff12/zapai/mappers.go#L50? 🙂 
Setting all of these fields is already possible (though I will concede that xrefing the mappers to figure out what the magic remapped key is is not the best UX), **even the special ones like `.User().SetId()`**. Those tag groups are just convenience setters to the `ai.user.authUserId` key in the same underlying `Properties` map.

rbtr:
Setting all of these tags [is already possible](https://github.com/Azure/azure-container-networking/blob/ce5cc41650653287dca7cf96823b3ea66e9bff12/zapai/mappers.go) by using the correct key (though I will concede that xrefing the mappers to figure out what the magic remapped key is, is not the best UX), **even the special looking ones like `.User().SetId()`**. Those tag groups are [just convenience setters to the `ai.user.` prefixed](https://github.com/microsoft/ApplicationInsights-Go/blob/master/appinsights/contracts/contexttags.go#L454-L463) keys in the same underlying `ContextTags` map.
I am closing this as wontfix because I don't think this belongs in the zapai library, but I would consider a different change which adds the ability to collect the default AI metadata in a _**composable**_ way via the mappers or some other mixin sort of thing, and **does not take any additional dependencies**.

PR #1625, opened 2022-09-20
Allow DNC to create NCs on a node with duplicate orchestrator context -- CNI & CNS 

paulyufan:
First version of 15415401, changes include:
* CNI
* CNS


ashvindeodhar (changes requested):
Not yet finished. Posting comments I have so far

cni/network/cnsclient.go
```diff
@@ -9,5 +9,6 @@ import (
 type cnsclient interface {
 	RequestIPAddress(ctx context.Context, ipconfig cns.IPConfigRequest) (*cns.IPConfigResponse, error)
 	ReleaseIPAddress(ctx context.Context, ipconfig cns.IPConfigRequest) error
-	GetNetworkConfiguration(ctx context.Context, orchestratorContext []byte) (*cns.GetNetworkContainerResponse, error)
+	GetNetworkContainerWithOrchestratorContext(ctx context.Context, orchestratorContext []byte) (*cns.GetNetworkContainerResponse, error)
+	GetNetworkContainersWithOrchestratorContext(ctx context.Context, orchestratorContext []byte) (*[]cns.GetNetworkContainerResponse, error)
```
ashvindeodhar, line 13: is pointer to slice is needed or just slice would do?

cni/network/multitenancy.go
```diff
@@ -203,7 +208,77 @@ func (m *Multitenancy) GetContainerNetworkConfiguration(
 	return ncResponse, hostSubnetPrefix, err
 }
 
-func (m *Multitenancy) getContainerNetworkConfigurationInternal(
+// get all network container configurations by given orchestratorContext
```
ashvindeodhar, line 211: nit: for given orchestrator context

cni/network/multitenancy.go
```diff
@@ -203,7 +208,77 @@ func (m *Multitenancy) GetContainerNetworkConfiguration(
...
+
+	if !nwCfg.EnableExactMatchForPodName {
+		podNameWithoutSuffix = network.GetPodNameWithoutSuffix(podName)
+	} else {
+		podNameWithoutSuffix = podName
+	}
+
+	log.Printf("Podname without suffix %v", podNameWithoutSuffix)
+
+	ncResponses, hostSubnetPrefix, err := m.getNetworkContainersWithOrchestratorContextInternal(ctx, podNamespace, podNameWithoutSuffix)
```
ashvindeodhar, line 225: should it be hostSubnetPrefixes since you are returning a slice?

cni/network/multitenancy.go
```diff
@@ -203,7 +208,77 @@ func (m *Multitenancy) GetContainerNetworkConfiguration(
...
+		for i := 0; i < len(responses); i++ {
+			if nwCfg.EnableSnatOnHost {
+				if responses[i].LocalIPConfiguration.IPSubnet.IPAddress == "" {
+					log.Printf("Snat IP is not populated. Got empty string")
+					return nil, []net.IPNet{}, errSnatIP
+				}
+			}
+		}
+	}
+	return ncResponses, hostSubnetPrefix, err
```
ashvindeodhar, line 238: As mentioned above I think we should just return slice instead of pointer to slice

cni/network/multitenancy.go
```diff
@@ -203,7 +208,77 @@ func (m *Multitenancy) GetContainerNetworkConfiguration(
...
+		PodNamespace: namespace,
+	}
+
+	orchestratorContext, err := json.Marshal(podInfo)
+	if err != nil {
+		log.Printf("Marshalling KubernetesPodInfo failed with %v", err)
+		return nil, []net.IPNet{}, fmt.Errorf("%w", err)
+	}
+
+	ncConfigs, err := m.cnsclient.GetNetworkContainersWithOrchestratorContext(ctx, orchestratorContext)
```
ashvindeodhar, line 256: let's use slice instead of pointer to slice

cni/network/multitenancy_mock.go
```diff
@@ -78,3 +78,74 @@ func (m *MockMultitenancy) GetContainerNetworkConfiguration(
...
+				IPAddress:    "169.254.0.4",
+				PrefixLength: localIPPrefixLen,
+			},
+			GatewayIPAddress: "169.254.0.1",
+		},
+
+		PrimaryInterfaceIdentifier: "10.240.0.4/24",
+		MultiTenancyInfo: cns.MultiTenancyInfo{
+			EncapType: cns.Vlan,
+			ID:        1,
```
ashvindeodhar, line 134: let's try to use the same examples for dual nic - i.e. let's have these 2 NCs belong to different vnets. Which means different ip space / IP / different Vlan ids. You can use the examples from the testbed VM which you created

cni/network/multitenancy_test.go
```diff
@@ -215,6 +227,144 @@ func TestCleanupMultitenancyResources(t *testing.T) {
...
+			GatewayIPAddress: "10.1.0.1",
+		},
+		Routes: []cns.Route{
+			{
+				IPAddress:        "10.1.0.0/16",
+				GatewayIPAddress: "10.1.0.1",
+			},
+		},
+	}
+	nc2 := &cns.GetNetworkContainerResponse{
```
ashvindeodhar, line 264: why are nc1 and nc2 responses are identical?

cni/network/multitenancy.go
timraymond: +1 to this. Slices are basically fat pointers anyway, so adding indirection to them is usually no savings (it just makes things harder to read).

cni/network/multitenancy.go
paulyufan: Yes, it should be named "hostSubnetPrefixes" since getNetworkContainersWithOrchestratorContextInternal returns a slice of hostSubPrefix

cns/restserver/util.go
```diff
@@ -350,11 +359,145 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
 
+func (service *HTTPRestService) getNetworkContainersResponse(
+	req cns.GetNetworkContainerRequest,
+) (getNetworkContainersResponse []cns.GetNetworkContainerResponse) {
+	var (
+		networkContainerID          string
+		getNetworkContainerResponse cns.GetNetworkContainerResponse
+		exists                      bool
+		waitingForUpdate            bool
+	)
```
timraymond, line 370: Pre-declaring variables like this is generally frowned-upon in Go (even though there's a lot of old code around that does this). This causes situations where you use one of these variables, but forget to assign to it (by later deleting the assignment). The compiler would catch that immediately ordinarily, but not if you did this C-style pre-declaration. The zero value would sneak out.

cns/restserver/util.go
```diff
@@ -350,11 +359,145 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+
+	service.Lock()
+	defer service.Unlock()
```
timraymond, line 373: It's never a good idea to have these methods exposed on a type unless the type is a synchronization primitive (like `sync.Cond`, a condition variable). Even if these weren't exposed here, you should use the underlying mutex directly.

cns/restserver/util.go
```diff
@@ -350,11 +359,145 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+				}
+
+				vfpUpdateComplete := !waitingForUpdate
+				ncstatus := service.state.ContainerStatus[ncid]
+				// Update the container status if-
+				// 1. VfpUpdateCompleted successfully
+				// 2. VfpUpdateComplete changed to false
+				if (getNetworkContainerResponse.Response.ReturnCode == types.NetworkContainerVfpProgramComplete &&
+					vfpUpdateComplete && ncstatus.VfpUpdateComplete != vfpUpdateComplete) ||
+					(!vfpUpdateComplete && ncstatus.VfpUpdateComplete != vfpUpdateComplete) {
```
timraymond, line 424: That these comments are needed hints at the fact that there's too much going on in this conditional. Consider moving this to maybe a function or some other type where you can break it up and document the individual parts.

cns/restserver/util.go
```diff
@@ -350,11 +359,145 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+			// If the NC goal state doesn't exist in CNS running in managed mode, call DNC to retrieve the goal state
+			var (
+				dncEP     = service.GetOption(acn.OptPrivateEndpoint).(string)
+				infraVnet = service.GetOption(acn.OptInfrastructureNetworkID).(string)
+				nodeID    = service.GetOption(acn.OptNodeID).(string)
+			)
+
+			service.Unlock()
+			getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.SyncNodeStatus(dncEP, infraVnet, nodeID, req.OrchestratorContext)
+			service.Lock()
```
timraymond, line 443: If you need to release a mutex like this and re-acquire it, there's usually other methods that should be added to better encapsulate the critical sections.

timraymond (changes requested):

cni/network/network.go
```diff
@@ -446,123 +448,157 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
-		return err
+			log.Printf("PrimaryInterfaceIdentifier: %v", ipamAddResult.hostSubnetPrefix.IP.String())
+		} else {
+			if er != nil {
+				er = errors.Wrapf(er, "GetNetworkContainersWithOrchestratorContext failed for podname %v namespace %v", k8sPodName, k8sNamespace)
+				log.Printf("%+v", er)
+				return er
+			}
+
+			var ncResponsesCopy [2]cns.GetNetworkContainerResponse
```
tamilmani1989, line 476: i dont recommend hardcoding number of nc responses in cni code..cni should not restrict here..dnc should handle that

cni/network/network.go
```diff
@@ -446,123 +448,157 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
 
-		ipamAddResult.ipv4Result = convertToCniResult(ipamAddResult.ncResponse, args.IfName)
+		log.Printf("hostSubnetPrefixes are %+v", hostSubnetPrefixes)
 
-		log.Printf("PrimaryInterfaceIdentifier: %v", ipamAddResult.hostSubnetPrefix.IP.String())
-	}
+		if ncResponses == nil {
+			log.Printf("CNS is old version, invoke old CNI API")
+			ipamAddResult.ncResponse, ipamAddResult.hostSubnetPrefix, er = plugin.multitenancyClient.GetNetworkContainerWithOrchestratorContext(
+				context.TODO(), nwCfg, k8sPodName, k8sNamespace)
```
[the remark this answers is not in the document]
tamilmani1989: nope..cni is short running process..not long running like cns or dnc

cni/network/network.go
```diff
@@ -446,123 +448,157 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+		ncResponses, hostSubnetPrefixes, er := plugin.multitenancyClient.GetNetworkContainersWithOrchestratorContext(
 			context.TODO(), nwCfg, k8sPodName, k8sNamespace)
-		if er != nil {
-			er = errors.Wrapf(er, "GetContainerNetworkConfiguration failed for podname %v namespace %v", k8sPodName, k8sNamespace)
-			log.Printf("%+v", er)
-			return er
-		}
 
-		ipamAddResult.ipv4Result = convertToCniResult(ipamAddResult.ncResponse, args.IfName)
+		log.Printf("hostSubnetPrefixes are %+v", hostSubnetPrefixes)
```
tamilmani1989, line 454: what happens if `GetNetworkContainersWithOrchestratorContext` returns err and `hostSubnetPrefixes` is nil? why are we not checking for error at first place here?

cni/network/network.go
```diff
@@ -446,123 +448,157 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
 
-		log.Printf("PrimaryInterfaceIdentifier: %v", ipamAddResult.hostSubnetPrefix.IP.String())
-	}
+		if ncResponses == nil {
```
tamilmani1989, line 456: how do we decide its old cns based on ncResponse? can you please add comments for changes you are making? it would be better for person to understand why its been done this way

cni/network/network.go
```diff
@@ -446,123 +448,157 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+			log.Printf("PrimaryInterfaceIdentifier: %v", ipamAddResult.hostSubnetPrefix.IP.String())
+		} else {
+			if er != nil {
+				er = errors.Wrapf(er, "GetNetworkContainersWithOrchestratorContext failed for podname %v namespace %v", k8sPodName, k8sNamespace)
+				log.Printf("%+v", er)
+				return er
+			}
+
+			var ncResponsesCopy [2]cns.GetNetworkContainerResponse
+			for i := 0; i < len(ncResponses); i++ {
```
tamilmani1989, line 477: should use slice and use for range

cni/network/network.go
```diff
@@ -446,123 +448,157 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+				ncResponsesCopy[i] = ncResponses[i]
+				ipamAddResult.ncResponse = &ncResponsesCopy[i]
+				ipamAddResult.ipv4Result = convertToCniResult(ipamAddResult.ncResponse, args.IfName)
+				ipamAddResult.hostSubnetPrefix = hostSubnetPrefixes[i]
+				ipamAddResults = append(ipamAddResults, ipamAddResult)
```
tamilmani1989, line 482: not sure why are you copying here

tamilmani1989 (changes requested):
from the code, couldn't understand why you implemented in certain way, so im not able to review fully.. it would be better if you setup time for code walkthrough ..

cni/network/network.go
paulyufan: using range will cause github issue and for all for loop, we need to use this type...

cni/network/network.go
paulyufan: because we need to deep copy ncResponse object here, otherwise the object will be missing

cni/network/multitenancy_test.go
```diff
@@ -215,6 +227,143 @@ func TestCleanupMultitenancyResources(t *testing.T) {
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			got1, _, err := tt.args.plugin.multitenancyClient.GetNetworkContainersWithOrchestratorContext(
+				tt.args.ctx,
+				tt.args.nwCfg,
+				tt.args.k8sPodName,
+				tt.args.k8sNamespace)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("GetContainerNetworkConfiguration() error = %v, wantErr %v", err, tt.wantErr)
+				return
```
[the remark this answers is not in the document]
rbtr: prefer `require.NoError()`?

cni/network/network.go
paulyufan: @ashvindeodhar can you please add DNC NC number restrictions here then I can remove hardcoded number 2

cni/network/network.go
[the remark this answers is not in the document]
rbtr: it could still hang, and it should spawn a root context that can be cancelled if the process gets a sigint/sigterm to (as much as possible) gracefully shut down

cni/network/network.go
```diff
@@ -446,123 +448,157 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
 			return err
 		}
 
-		if resultSecondAdd != nil {
-			ipamAddResult.ipv4Result = resultSecondAdd
-			return nil
+		endpointID := GetEndpointID(args)
+		policies := cni.GetPoliciesFromNwCfg(nwCfg.AdditionalArgs)
+
+		options := make(map[string]interface{})
```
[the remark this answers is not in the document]
paulyufan: good to know "any" is a new way to replace interface{}

cns/restserver/restserver.go
```diff
@@ -106,14 +108,72 @@ type containerstatus struct {
 	VfpUpdateComplete             bool // True when VFP programming is completed for the NC
 }
 
+type ncSet map[string]struct{}
```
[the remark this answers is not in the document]
paulyufan: if this becomes a general interface, yes let's name it as "Set"

cni/network/network.go
rbtr: > using range will cause github issue

I am concerned that you don't understand why you have made the change that you made because this statement does not make sense

cns/restserver/restserver.go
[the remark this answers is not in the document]
timraymond: @paulyufan It already is a general type. There's no need to falsely tie it to a domain concept.

cni/network/cnsclient.go
```diff
@@ -9,5 +9,6 @@ import (
 type cnsclient interface {
 	RequestIPAddress(ctx context.Context, ipconfig cns.IPConfigRequest) (*cns.IPConfigResponse, error)
 	ReleaseIPAddress(ctx context.Context, ipconfig cns.IPConfigRequest) error
-	GetNetworkConfiguration(ctx context.Context, orchestratorContext []byte) (*cns.GetNetworkContainerResponse, error)
+	GetNetworkContainerWithOrchestratorContext(ctx context.Context, orchestratorContext []byte) (*cns.GetNetworkContainerResponse, error)
+	GetNetworkContainersWithOrchestratorContext(ctx context.Context, orchestratorContext []byte) ([]cns.GetNetworkContainerResponse, error)
```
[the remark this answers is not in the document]
paulyufan: @ashvindeodhar I agree with Tim that we can change the function name to GetNetworkContainers() and pass the orchestratorContext parameter. What do you think?

cns/restserver/util.go
paulyufan: Need to redesign this part

cns/restserver/util.go
paulyufan: Can we discuss more about this

cns/restserver/util.go
paulyufan: Can we discuss more about this

cns/restserver/util.go
timraymond: Yeah, so the whole reason to use a mutex is because you want some set of operations to be atomic from the perspective of other goroutines. You can usually give that set of things a name, and move all of that into its own method. Generally that method looks like:

```go
func (s *SomeType) doAtomicThings() {
       s.mu.Lock()
       defer s.mu.Unlock()

       // do stuff
}
```

That way, instead of dropping the lock mid-method here, you can have this method call into the atomic methods:

```go
s.doAtomicThings()
syncNodeStatus()
s.doOtherAtomicThings()
```

This turns "s" into something approximating a [Monitor](https://en.wikipedia.org/wiki/Monitor_(synchronization)) in the "thread-safe class" definition.

cni/network/network.go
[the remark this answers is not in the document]
paulyufan: we discussed and it seems it's right thing to do. it's not a block for this PR and we can create a sub-task to optimize this later

cns/restserver/util.go
paulyufan: We can revisit this to optimize. But we can move this to a common function

cns/restserver/util.go
paulyufan: As discussed, if we change this, we will change everywhere and it might be not high priority now,

cns/restserver/util.go
paulyufan: As discussed, we can create a new task to optimize this later on since this does not block the PR

timraymond (approved):

cni/network/multitenancy.go
```diff
@@ -37,11 +37,16 @@ type MultitenancyClient interface {
 	DetermineSnatFeatureOnHost(
 		snatFile string,
 		nmAgentSupportedApisURL string) (bool, bool, error)
-	GetContainerNetworkConfiguration(
+	GetNetworkContainer(
```
paulyufan, line 40: wrap these 2 functions to one function

cni/network/multitenancy.go
```diff
@@ -203,7 +208,74 @@ func (m *Multitenancy) GetContainerNetworkConfiguration(
...
+
+	if !nwCfg.EnableExactMatchForPodName {
+		podNameWithoutSuffix = network.GetPodNameWithoutSuffix(podName)
+	} else {
+		podNameWithoutSuffix = podName
+	}
+
+	log.Printf("Podname without suffix %v", podNameWithoutSuffix)
+
+	ncResponses, hostSubnetPrefixes, err := m.getNetworkContainersWithOrchestratorContextInternal(ctx, podNamespace, podNameWithoutSuffix)
```
paulyufan, line 225: TODO: combine ncResponses and hostSubnetPrefixes together

cni/network/network.go
```diff
@@ -446,122 +445,155 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+			}
 
-	// Initialize values from network config.
-	networkID, err := plugin.getNetworkName(args.Netns, &ipamAddResult, nwCfg)
-	if err != nil {
-		log.Printf("[cni-net] Failed to extract network name from network config. error: %v", err)
-		return err
+			// currently DNC only allows to create dual NICs
+			//nolint:gocritic // copy is ok
+			for i, nc := range ncResponses {
```
paulyufan, line 478: return ipamAddResults directly from GetNetworkContainersConfiguration()

cns/restserver/util.go
```diff
@@ -350,11 +359,149 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+			Routes:                     savedReq.Routes,
+			CnetAddressSpace:           savedReq.CnetAddressSpace,
+			MultiTenancyInfo:           savedReq.MultiTenancyInfo,
+			PrimaryInterfaceIdentifier: savedReq.PrimaryInterfaceIdentifier,
+			LocalIPConfiguration:       savedReq.LocalIPConfiguration,
+			AllowHostToNCCommunication: savedReq.AllowHostToNCCommunication,
+			AllowNCToHostCommunication: savedReq.AllowNCToHostCommunication,
+		}
+
+		getNetworkContainersResponse = append(getNetworkContainersResponse, getNetworkContainerResponse)
```
tamilmani1989, line 493: name of this variable can be different... like ncList or something else

cns/restserver/util.go
```diff
@@ -382,32 +529,41 @@ func (service *HTTPRestService) getNetworkContainerResponse(
...
+
+		networkContainerID = string(ncidBytes)
+		if networkContainerID != "" {
+			exists = true
+		}
 
 		if exists {
 			// If the goal state is available with CNS, check if the NC is pending VFP programming
-			waitingForUpdate, getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.isNCWaitingForUpdate(service.state.ContainerStatus[containerID].CreateNetworkContainerRequest.Version, containerID) //nolint:lll // bad code
+			waitingForUpdate, getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.isNCWaitingForUpdate(service.state.ContainerStatus[networkContainerID].CreateNetworkContainerRequest.Version, networkContainerID) //nolint:lll // bad code
```
tamilmani1989, line 544: same as above.. you can use local variable to get return value and then at end copy it to getneworkcontainerresponse.

tamilmani1989 (changes requested):

cns/restserver/util.go
```diff
@@ -350,11 +359,149 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+				waitingForUpdate, getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.isNCWaitingForUpdate(service.state.ContainerStatus[ncid].CreateNetworkContainerRequest.Version, networkContainerID) //nolint:lll // bad code
+				// If the return code is not success, return the error to the caller
+				if getNetworkContainerResponse.Response.ReturnCode == types.NetworkContainerVfpProgramPending {
+					logger.Errorf("[Azure-CNS] isNCWaitingForUpdate failed for NCID: %s with error: %s",
+						ncid, getNetworkContainerResponse.Response.Message)
+					getNetworkContainersResponse = append(getNetworkContainersResponse, getNetworkContainerResponse)
+					return getNetworkContainersResponse
+				}
+
+				vfpUpdateComplete := !waitingForUpdate
```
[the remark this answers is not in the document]
ashvindeodhar: this is existing code. I'll work with Paul to address this.

cns/restserver/util.go
```diff
@@ -350,11 +359,149 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+		// marshal nc set i.e.,&map[nc1:{} nc2:{}] to bytes
+		ncidBytes, err := json.Marshal(service.state.ContainerIDByOrchestratorContext[podInfo.Name()+podInfo.Namespace()])
+		if err != nil {
+			logger.Errorf("Failed to marlshal NC set")
+			return nil
+		}
+
+		// get networkContainerID, the format is "nc1,nc2"
+		// networkContainerID is used for isNCWaitingForUpdate() and stored for usage out of this case
+		networkContainerID = string(ncidBytes)
```
ashvindeodhar, line 397: the name of var is misleading. It;s not a single NCID, it's a list of ncids

cns/restserver/util.go
```diff
@@ -350,11 +359,149 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+	default:
+		getNetworkContainerResponse.Response.ReturnCode = types.UnsupportedOrchestratorType
+		getNetworkContainerResponse.Response.Message = fmt.Sprintf("Invalid orchestrator type %v", service.state.OrchestratorType)
+		getNetworkContainersResponse = append(getNetworkContainersResponse, getNetworkContainerResponse)
+		return getNetworkContainersResponse
+	}
+
+	// get ncSet from networkContainerID and unmarshal to list, i.e.,["nc1","nc2"]
+	ncSet := &Set{}
+	if err := json.Unmarshal([]byte(networkContainerID), &ncSet); err != nil {
```
ashvindeodhar, line 466: I feel number of unmarshaling in this func can reduced. Let's discuss

ashvindeodhar: There is repeated unmarshaling

cns/restserver/util.go
```diff
@@ -350,11 +359,149 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+	}
+
+	for ncid := range *ncSet {
+		containerStatus := service.state.ContainerStatus
+		containerDetails, ok := containerStatus[ncid]
+		if !ok {
+			getNetworkContainerResponse.Response.ReturnCode = types.UnknownContainerID
+			getNetworkContainerResponse.Response.Message = "NetworkContainer doesn't exist."
+			getNetworkContainersResponse = append(getNetworkContainersResponse, getNetworkContainerResponse)
+			return getNetworkContainersResponse
```
ashvindeodhar, line 477: instead of returning immediately, we should add all the responses and then return.

cns/restserver/util.go
```diff
@@ -382,32 +529,41 @@ func (service *HTTPRestService) getNetworkContainerResponse(
 
 		logger.Printf("pod info %+v", podInfo)
 
-		containerID, exists = service.state.ContainerIDByOrchestratorContext[podInfo.Name()+podInfo.Namespace()]
+		ncidBytes, err := json.Marshal(service.state.ContainerIDByOrchestratorContext[podInfo.Name()+podInfo.Namespace()])
```
ashvindeodhar, line 532: why do you need to marshal? can you not check if the Set is non empty? You can expose a func to the Set that you added to return if isEmpty()

ashvindeodhar (changes requested):

cni/network/multitenancy.go
```diff
@@ -180,7 +180,7 @@ func (m *Multitenancy) SetupRoutingForMultitenancy(
 	setupInfraVnetRoutingForMultitenancy(nwCfg, azIpamResult, epInfo, result)
 }
 
-func (m *Multitenancy) GetContainerNetworkConfiguration(
+func (m *Multitenancy) GetNetworkContainer(
```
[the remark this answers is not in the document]
paulyufan: Yes, we do not use this anymore, need to delete this API

cns/restserver/util.go
```diff
@@ -350,11 +359,149 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+			logger.Errorf("Failed to marlshal NC set")
+			return nil
+		}
+
+		// get networkContainerID, the format is "nc1,nc2"
+		// networkContainerID is used for isNCWaitingForUpdate() and stored for usage out of this case
+		networkContainerID = string(ncidBytes)
+		if networkContainerID != "" {
+			exists = true
+		}
```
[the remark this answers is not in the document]
paulyufan: We need use networkContainerID as string for "if existing" part

ashvindeodhar (changes requested):

cns/restserver/util.go
```diff
@@ -380,34 +507,43 @@ func (service *HTTPRestService) getNetworkContainerResponse(
...
 
-		containerID, exists = service.state.ContainerIDByOrchestratorContext[podInfo.Name()+podInfo.Namespace()]
+		if networkContainerIDs != nil {
+			exists = true
+		}
 
 		if exists {
 			// If the goal state is available with CNS, check if the NC is pending VFP programming
-			waitingForUpdate, getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.isNCWaitingForUpdate(service.state.ContainerStatus[containerID].CreateNetworkContainerRequest.Version, containerID) //nolint:lll // bad code
+			waitingForUpdate, getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.isNCWaitingForUpdate(service.state.ContainerStatus[networkContainerIDs[0]].CreateNetworkContainerRequest.Version, networkContainerIDs[0]) //nolint:lll // bad code
```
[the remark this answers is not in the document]
paulyufan: change the name of getNetworkContainersResponse to getAllNetworkContainerResponse

cns/restserver/restserver.go
```diff
@@ -132,7 +132,7 @@ type httpRestServiceState struct {
 	OrchestratorType                 string
 	NodeID                           string
 	Initialized                      bool
-	ContainerIDByOrchestratorContext map[string]string          // OrchestratorContext is key and value is NetworkContainerID.
+	ContainerIDByOrchestratorContext map[string]Set             // OrchestratorContext is the key and value is NetworkContainerIDs separated by comma.
```
ashvindeodhar, line 135: Set by itself won't store the values as comma separated. Remove the comma separated comment.
You can mention that in the cns state file, it will be stored as comma separated

cns/restserver/api_test.go
```diff
@@ -246,8 +246,90 @@ func TestSetOrchestratorType_NCsPresent(t *testing.T) {
...
+				assert.NotNil(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, tt.response, resp)
+		})
+	}
+}
+
+func TestSetOrchestratorType_DualNCsPresent(t *testing.T) {
```
ashvindeodhar, line 314: we agreed last time that we don't need set orchestrator type test for dual NC. Remove this newly added test.
where are the tests for 
TestCreateNetworkContainerDualNic and
TestGetAllNetworkContainerByOrchestratorContext?

cns/restserver/api_test.go
```diff
@@ -246,8 +246,90 @@ func TestSetOrchestratorType_NCsPresent(t *testing.T) {
 					ContainerStatus: map[string]containerstatus{
 						"nc1": {},
 					},
-					ContainerIDByOrchestratorContext: map[string]string{
-						"nc1": "present",
+					ContainerIDByOrchestratorContext: map[string]Set{
+						"nc1": {"present": {}},
```
ashvindeodhar, line 250: old comment is still not addressed - these should be guids

cns/restserver/api.go
```diff
@@ -874,6 +862,40 @@ func (service *HTTPRestService) getNetworkContainerByID(w http.ResponseWriter, r
...
+func (service *HTTPRestService) getAllNetworkContainers(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] getAllNetworkContainers")
+
+	var req cns.GetNetworkContainerRequest
+
+	logger.Printf("req is %+v", req)
+	err := service.Listener.Decode(w, r, &req)
+	logger.Request(service.Name, &req, err)
+	if err != nil {
+		logger.Errorf("[Azure CNS] failed to decode cns request due to %+v", err)
```
ashvindeodhar, line 875: please log req in the error log. w/o that, we won't be able to debug

cns/restserver/api.go
```diff
@@ -874,6 +862,40 @@ func (service *HTTPRestService) getNetworkContainerByID(w http.ResponseWriter, r
...
+	if err = platform.SetSdnRemoteArpMacAddress(); err != nil {
+		logger.Printf("[Azure CNS] SetSdnRemoteArpMacAddress failed with error: %s", err.Error())
+		return
+	}
+
+	getNetworkContainersResponse := service.getNetworkContainersResponse(req)
+	logger.Printf("getNetworkContainersResponse are %+v", getNetworkContainersResponse)
+	var resp cns.GetAllNetworkContainersResponse
+
+	resp.Response.ReturnCode = getNetworkContainersResponse[0].Response.ReturnCode
```
ashvindeodhar, line 891: we should not just copy the returnCode and Message from 0th NC. We should set those to indicate if all the NC responses succeeded or if it failed. Let's discuss this.

cns/restserver/api.go
```diff
@@ -874,6 +862,40 @@ func (service *HTTPRestService) getNetworkContainerByID(w http.ResponseWriter, r
...
+
+	// getAllNetworkContainers gets called for multitenancy and
+	// setting the SDNRemoteArpMacAddress regKey is essential for the multitenancy
+	// to work correctly in case of windows platform. Return if there is an error
+	if err = platform.SetSdnRemoteArpMacAddress(); err != nil {
+		logger.Printf("[Azure CNS] SetSdnRemoteArpMacAddress failed with error: %s", err.Error())
+		return
+	}
+
+	getNetworkContainersResponse := service.getNetworkContainersResponse(req)
```
ashvindeodhar, line 887: As discussed, change the name to service.getAllNetworkContainerResponses

cns/client/client_test.go
```diff
@@ -639,7 +639,7 @@ func TestGetNetworkConfiguration(t *testing.T) {
 			orchestratorContext, err := json.Marshal(tt.podInfo)
 			assert.NoError(t, err, "marshaling orchestrator context failed")
 
-			got, err := client.GetNetworkConfiguration(tt.ctx, orchestratorContext)
+			got, err := client.GetNetworkContainer(tt.ctx, orchestratorContext)
 			if tt.wantErr {
```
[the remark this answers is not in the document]
ashvindeodhar: is this done?

cns/client/client.go
```diff
@@ -93,8 +94,60 @@ func buildRoutes(baseURL string, paths []string) (map[string]url.URL, error) {
...
+		return nil, errors.Wrap(err, "failed to build request")
+	}
+	req.Header.Set(headerContentType, contentTypeJSON)
+	res, err := c.client.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "http request failed")
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
```
ashvindeodhar, line 123: we should handle the case where the res.StatusCode is 404 indicating that the API is not supported.
Let's add a todo for CNS to respond with 405 if the API is not supported.
Can we add UnsupportedAPI error code in the types and return it similar to line 125?

cns/restserver/util.go
```diff
@@ -184,13 +185,20 @@ func (service *HTTPRestService) saveNetworkContainerGoalState(
...
+				service.state.ContainerIDByOrchestratorContext = make(map[string]Set)
 			}
 
-			service.state.ContainerIDByOrchestratorContext[podInfo.Name()+podInfo.Namespace()] = req.NetworkContainerid
+			if service.state.ContainerIDByOrchestratorContext[orchestratorContext] == nil {
+				service.state.ContainerIDByOrchestratorContext[orchestratorContext] = Set{}
+			}
+
+			ncSet := service.state.ContainerIDByOrchestratorContext[orchestratorContext]
+			ncSet.Add(req.NetworkContainerid)
```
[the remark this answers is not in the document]
paulyufan: This updates the Set at ContainerIDByOrchestratorContext

cns/restserver/util.go
```diff
@@ -350,11 +358,130 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+		getNetworkContainersResponse := []cns.GetNetworkContainerResponse{}
+		getNetworkContainerResponse.Response.ReturnCode = types.UnsupportedOrchestratorType
+		getNetworkContainerResponse.Response.Message = fmt.Sprintf("Invalid orchestrator type %v", service.state.OrchestratorType)
+		getNetworkContainersResponse = append(getNetworkContainersResponse, getNetworkContainerResponse)
+		return getNetworkContainersResponse
+	}
+
+	// get ncList i.e.,["nc1","nc2"]
+	getNetworkContainersResponse := []cns.GetNetworkContainerResponse{}
+	ncList := strings.Split(networkContainerIDs, ",")
```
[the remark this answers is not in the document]
paulyufan: We get this ncList from strings.Split(networkContainerIDs, ",")

cns/restserver/api_test.go
paulyufan: I think TestGetAllNetworkContainers == TestGetAllNetworkContainerByOrchestratorContext() test case

cni/network/network.go
```diff
@@ -446,123 +445,122 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+			}
 		}
-	}
 
-	// Initialize azureipam/cns ipam
-	if plugin.ipamInvoker == nil {
-		switch nwCfg.IPAM.Type {
-		case network.AzureCNS:
-			plugin.ipamInvoker = NewCNSInvoker(k8sPodName, k8sNamespace, cnsClient, util.ExecutionMode(nwCfg.ExecutionMode), util.IpamMode(nwCfg.IPAM.Mode))
+		// Initialize azureipam/cns ipam
```
ashvindeodhar, line 502: remove extra white space

ashvindeodhar (changes requested):

cns/restserver/util.go
```diff
@@ -350,13 +358,131 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+	default:
+		getNetworkContainersResponse := []cns.GetNetworkContainerResponse{}
+		getNetworkContainerResponse.Response.ReturnCode = types.UnsupportedOrchestratorType
+		getNetworkContainerResponse.Response.Message = fmt.Sprintf("Invalid orchestrator type %v", service.state.OrchestratorType)
+		getNetworkContainersResponse = append(getNetworkContainersResponse, getNetworkContainerResponse)
+		return getNetworkContainersResponse
+	}
+
+	getNetworkContainersResponse := []cns.GetNetworkContainerResponse{}
+	ncList := strings.Split(networkContainerIDs, ",")
```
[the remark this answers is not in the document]
paulyufan: networkContainerIDs is the global variable. If we look at getNetworkContainerResponse(), we are using the same way to populate networkContainerID

cns/restserver/util.go
```diff
@@ -350,13 +358,131 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
 	return 0, ""
 }
 
+func (service *HTTPRestService) getAllNetworkContainerResponses(
```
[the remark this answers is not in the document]
paulyufan: We can combine them together, but if we want to get single getNetworkContainerResponse, we need to do one loop operation which will add time complexity for this.

cns/restserver/util.go
```diff
@@ -350,13 +358,131 @@ func (service *HTTPRestService) removeToBeDeletedIPStateUntransacted(
...
+		podInfo, err := cns.UnmarshalPodInfo(req.OrchestratorContext)
+		getNetworkContainersResponse := []cns.GetNetworkContainerResponse{}
+		if err != nil {
+			getNetworkContainerResponse.Response.ReturnCode = types.UnexpectedError
+			getNetworkContainerResponse.Response.Message = fmt.Sprintf("Unmarshalling orchestrator context failed with error %v", err)
+			getNetworkContainersResponse = append(getNetworkContainersResponse, getNetworkContainerResponse)
+		}
+
+		exists := false
+		waitingForUpdate := false
```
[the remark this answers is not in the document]
timraymond: @ashvindeodhar Usually the short declaration (`:=`) is preferred in all circumstances, except where it's necessary to hoist a definition to some higher lexical scope... usually where conditionals or loops are involved. Even when a specific type is required, it's usually better to do something like `foo := uint64(42)`. There's a good post about it here: https://npf.io/2022/05/why-infer/ . The points raised there don't entirely apply to this specific instance, but it at least shows why it's a good habit to choose `:=` first.

cni/network/network.go
```diff
@@ -446,122 +448,120 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+		sendEvent(plugin, fmt.Sprintf("Allocated IPAddress from ipam:%+v v6:%+v", ipamAddResult.ipv4Result, ipamAddResult.ipv6Result))
 	}
 
-	sendEvent(plugin, fmt.Sprintf("Allocated IPAddress from ipam:%+v v6:%+v", ipamAddResult.ipv4Result, ipamAddResult.ipv6Result))
+	// iterate ipamAddResults and program the endpoint
+	for i := 0; i < len(ipamAddResults); i++ {
+		ipamAddResult = ipamAddResults[i]
 
-	defer func() {
+		networkID, err := plugin.getNetworkName(args.Netns, &ipamAddResult, nwCfg)
```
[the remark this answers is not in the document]
tamilmani1989: can you check if above doesn't work?

cni/network/network.go
```diff
@@ -446,123 +445,120 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
 		}
-
-		ipamAddResult.ipv4Result = convertToCniResult(ipamAddResult.ncResponse, args.IfName)
-
-		log.Printf("PrimaryInterfaceIdentifier: %v", ipamAddResult.hostSubnetPrefix.IP.String())
+	} else {
+		// TODO: refactor this code for simplification
+		// Add empty ipamAddResult object for single tenancy mode
+		// this will be used for: ipamAddResult, err = plugin.ipamInvoker.Add(ipamAddConfig)
+		ipamAddResults = append(ipamAddResults, ipamAddResult)
```
tamilmani1989, line 458: instead of this, we need to have better way.. how about having numNetworks variable with default value as 1 and then set to len(ipamAddResult) in case of mutltitenancy case and then use numNetworks to iterate

tamilmani1989 (changes requested):

cni/network/network.go
paulyufan: let's name it as numEnpoints

ashvindeodhar (approved):

ashvindeodhar:
let's get @tamilmani1989 approval before merging this.

cns/restserver/util.go
```diff
@@ -951,3 +999,40 @@ func (service *HTTPRestService) setResponse(w http.ResponseWriter, returnCode ty
 	serviceErr := service.Listener.Encode(w, &response)
 	logger.Response(service.Name, response, returnCode, serviceErr)
 }
+
+// ncList contains comma-separated list of unique NCs
+type ncList string
+
+// only add unique NC to ncList
+func (n *ncList) Add(nc string) {
```
tamilmani1989, line 1007: why did we change this implementation from set?

tamilmani1989 (changes requested):

cns/restserver/util.go
paulyufan: This is because we wanted to keep nc to be added in order as discussed before but seems like they did change from their side. @ashvindeodhar do we keep this now?

cns/restserver/internalapi.go
```diff
@@ -312,32 +312,35 @@ func (service *HTTPRestService) ReconcileNCState(ncRequest *cns.CreateNetworkCon
...
-		delete(service.state.ContainerStatus, req.NetworkContainerid)
+		delete(service.state.ContainerStatus, ncid)
 	}
 
 	if service.state.ContainerIDByOrchestratorContext != nil {
-		for orchestratorContext, networkContainerID := range service.state.ContainerIDByOrchestratorContext {
-			if networkContainerID == req.NetworkContainerid {
-				delete(service.state.ContainerIDByOrchestratorContext, orchestratorContext)
-				break
+		for orchestratorContext, networkContainerIDs := range service.state.ContainerIDByOrchestratorContext { //nolint:gocritic // copy is ok
```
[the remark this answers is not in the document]
paulyufan: There are two apis there before, DeleteNetworkContainer() is part of CNS route deleting container that we created. I am not 100% sure about DeleteNetworkContainerInternal(), it's used for Reconcile () and Reconcile() is called on multi-tenant CRD status changes but this function is only used in a UT test case.

cni/network/network.go
```diff
@@ -941,80 +945,82 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+		// schedule send metric before attempting delete
+		defer sendMetricFunc() //nolint:gocritic
+		logAndSendEvent(plugin, fmt.Sprintf("Deleting endpoint:%v", endpointID))
+		// Delete the endpoint.
+		if err = plugin.nm.DeleteEndpoint(networkID, endpointID); err != nil {
+			// return a retriable error so the container runtime will retry this DEL later
+			// the implementation of this function returns nil if the endpoint doens't exist, so
+			// we don't have to check that here
+			return plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
+		}
```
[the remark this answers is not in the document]
paulyufan: In the old code, deletion order is same:
https://github.com/Azure/azure-container-networking/blob/0863a815f4d7dc37965638fcffaa6252e48d0898/cni/network/network.go#L990

cns/restserver/api.go
```diff
@@ -952,14 +978,17 @@ func (service *HTTPRestService) deleteNetworkContainer(w http.ResponseWriter, r
...
-				if networkContainerID == req.NetworkContainerid {
-					delete(service.state.ContainerIDByOrchestratorContext, orchestratorContext)
-					break
+			for orchestratorContext, networkContainerIDs := range service.state.ContainerIDByOrchestratorContext { //nolint:gocritic // copy is ok
+				if networkContainerIDs.Contains(ncid) {
+					networkContainerIDs.Delete(ncid)
+					if *networkContainerIDs == "" {
+						delete(service.state.ContainerIDByOrchestratorContext, orchestratorContext)
+						break
+					}
```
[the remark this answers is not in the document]
paulyufan: we do not have any UTs for this api.go, we should add them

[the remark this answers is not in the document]
paulyufan: Yes, in api_test.go, deleteNetworkContainerWithParams()

req, err := http.NewRequest(http.MethodPost, ***cns.DeleteNetworkContainer**, &body)

cni/network/multitenancy_test.go
```diff
@@ -381,8 +401,8 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
 				require.Error(err)
 			}
 			require.NoError(err)
-			require.Exactly(tt.want1, got1)
-			require.Exactly(tt.want2, got2)
+			require.Exactly(tt.want1, got[0].ncResponse)
```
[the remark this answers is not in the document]
paulyufan: added

cni/network/network.go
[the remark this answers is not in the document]
tamilmani1989: ok then its good. i got confused with this: https://github.com/Azure/azure-container-networking/blob/0863a815f4d7dc37965638fcffaa6252e48d0898/cni/network/network.go#L976

cns/restserver/internalapi.go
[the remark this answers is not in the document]
tamilmani1989: @rbtr @thatmattlong to review this change and make sure it doesn't affect multitenantcrd

cns/client/client.go
```diff
@@ -93,8 +94,69 @@ func buildRoutes(baseURL string, paths []string) (map[string]url.URL, error) {
...
+
+	//  investigate 404 orchestratorContext which is invalid and make sure this is addressed
+	if res.StatusCode == http.StatusNotFound {
+		return nil, &CNSClientError{
+			Code: types.UnsupportedAPI,
+			Err:  errors.Errorf("Unsupported API"),
+		}
+	}
+
+	// TODO: CNS should return 405 if client is trying to call API that CNS doesn't support
```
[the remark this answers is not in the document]
tamilmani1989: then can we add a case for 405 as well and handle as like 404?

cns/restserver/api.go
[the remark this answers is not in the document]
paulyufan: two UT tests:
(1) Remove one NC
(2) Remove two NCs => state should be not represent

cns/restserver/util.go
paulyufan: add more test cases to cover this.

Add 3 NCs, remove first, second and third one, and check if they are deleted in order

paulyufan: test cases added

cni/network/network.go
```diff
@@ -941,80 +945,82 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-	networkID, err = plugin.getNetworkName(args.Netns, nil, nwCfg)
-	if err != nil {
-		log.Printf("[cni-net] Failed to extract network name from network config. error: %v", err)
-		// If error is not found error, then we ignore it, to comply with CNI SPEC.
-		if !network.IsNetworkNotFoundError(err) {
+	// Loop through all the networks that are created for the given Netns. In case of multi-nic scenario ( currently supported
+	// scenario is dual-nic ), single container may have endpoints created in multiple networks. As all the endpoints are
+	// deleted, getNetworkName will return error of the type NetworkNotFoundError which will result in nil error as compliance
+	// with CNI SPEC as mentioned below.
+	for {
```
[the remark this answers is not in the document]
paulyufan: It's solved by this PR: https://github.com/Azure/azure-container-networking/pull/1828

cns/client/client.go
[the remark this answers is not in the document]
paulyufan: "TODO" is Removed

tamilmani1989 (approved):
lgtm

PR #1628, opened 2022-09-21
validate nnc in swift e2e scale/connectivity test

rbtr:

**Reason for Change**:

Adds a validation of the NNC (requested IPs) during the swift e2e.
This includes functionality to support testing in subnet scarcity mode, but it is hardcoded to false until the clustersubnetstates are being written by RC.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/integration/k8s_test.go
+++ b/test/integration/k8s_test.go
@@ -4,23 +4,23 @@ package k8s
 
 import (
 	"context"
-	"log"
-
-	//"dnc/test/integration/goldpinger"
 	"errors"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"testing"
 	"time"
 
+	"github.com/Azure/azure-container-networking/crd/nodenetworkconfig"
 	"github.com/Azure/azure-container-networking/test/integration/goldpinger"
+	"github.com/Azure/azure-container-networking/test/integration/kube"
 	"github.com/Azure/azure-container-networking/test/integration/retry"
-	v1 "k8s.io/client-go/kubernetes/typed/apps/v1"
-
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	v1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 	"k8s.io/client-go/util/homedir"
 )
 
@@ -86,12 +86,12 @@ todo:
 */
 
 func TestPodScaling(t *testing.T) {
-	clientset, err := mustGetClientset()
+	clientset, err := mustGetClientset(*kubeconfig)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	restConfig := mustGetRestConfig(t)
+	restConfig := mustGetRestConfig(t, *kubeconfig)
 	deployment, err := mustParseDeployment(gpDeployment)
 	if err != nil {
 		t.Fatal(err)
@@ -184,9 +184,33 @@ func TestPodScaling(t *testing.T) {
 				}
 				t.Log("all pods have been allocated IPs")
 			}) {
-				errors.New("Pods don't have IP's")
-				return
+				t.Error("some pods don't have IP's")
 			}
+
+			t.Run("nodenetworkconfig has converged", func(t *testing.T) {
+				clusterCheckCtx, cancel := context.WithTimeout(ctx, 20*time.Minute)
+				defer cancel()
+				nnccli, err := nodenetworkconfig.NewClient(restConfig)
+				if err != nil {
+					t.Fatal(err)
+				}
+				nodes, err := clientset.CoreV1().Nodes().List(clusterCheckCtx, metav1.ListOptions{})
+				if err != nil {
+					t.Fatal(err)
+				}
+				for _, node := range nodes.Items {
+					nnc, err := nnccli.Get(clusterCheckCtx, types.NamespacedName{Namespace: "kube-system", Name: node.Name})
+					if err != nil {
+						t.Fatal(err)
+					}
+					podList, err := kube.GetPodsByNode(clusterCheckCtx, clientset.CoreV1().Pods(""), node.Name)
+					if err != nil {
+						t.Fatal(err)
+					}
+					pods := kube.FilterHostnet(podList.Items)
+					kube.ValidateNNC(t, *nnc, int64(len(pods)), false)
+				}
+			})
 
 			t.Run("all pods can ping each other", func(t *testing.T) {
 				clusterCheckCtx, cancel := context.WithTimeout(ctx, 20*time.Minute)
--- a/test/integration/k8s_test.go
+++ b/test/integration/k8s_test.go
@@ -184,7 +184,7 @@ func TestPodScaling(t *testing.T) {
 				}
 				t.Log("all pods have been allocated IPs")
 			}) {
-				t.Error("some pods don't have IP's")
+				t.Error("some pods don't have IPs")
 			}
 
 			t.Run("nodenetworkconfig has converged", func(t *testing.T) {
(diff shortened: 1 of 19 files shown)


test/integration/manifests/cns/ciliumconfigmap.yaml
```diff
@@ -22,6 +22,7 @@ data:
           "NodeSyncIntervalInSeconds": 30
       },
       "ChannelMode": "CRD",
+      "EnableSubnetScarcity": true,
```
camrynl, line 25: nothing to change here, but I haven't seen this mode in use before, could I get some context?

camrynl (changes requested):

test/integration/k8s_test.go
```diff
@@ -43,7 +41,7 @@ var (
 	kubeconfig          = flag.String("test-kubeconfig", filepath.Join(homedir.HomeDir(), ".kube", "config"), "(optional) absolute path to the kubeconfig file")
 	delegatedSubnetID   = flag.String("delegated-subnet-id", "", "delegated subnet id for node labeling")
 	delegatedSubnetName = flag.String("subnet-name", "", "subnet name for node labeling")
-	gpPodScaleCounts    = []int{3, 15, 150, 3}
+	gpPodScaleCounts    = []int{3, 15, 3}
```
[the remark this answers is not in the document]
rbtr: oops, I took that out to make my runs faster and forgot to put it back

test/integration/manifests/cns/ciliumconfigmap.yaml
rbtr: since we scale in batches of IPs in swift, it is possible to have a lot of allocated but free IPs on the node - this mode will allow CNS to give back its extra IPs (effectively setting the batch size to 1) when the subnet is running out of IPs. DNC/RC will monitor the subnet utilization and push a CRD which CNS to signal CNS. 
If the subnet is not constrained, this flag is a no-op, but I'm enabling it here now to make sure that it doesn't break anything if it's accidentally rolled out before the DNC changes.

test/integration/k8s_test.go
```diff
@@ -184,10 +182,34 @@ func TestPodScaling(t *testing.T) {
 				}
 				t.Log("all pods have been allocated IPs")
 			}) {
-				errors.New("Pods don't have IP's")
-				return
+				t.Error("some pods don't have IP's")
```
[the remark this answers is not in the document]
rbtr: I think the previous code was broken - it didn't do anything with the error it created, and it returned nothing. We enter this block iff the `t.Run` returns an false/an error, so I think we should halt the test and print an error - `t.Error` will fail-fast

test/integration/k8s_test.go
```diff
@@ -184,10 +182,34 @@ func TestPodScaling(t *testing.T) {
...
+					if err != nil {
+						t.Fatal(err)
+					}
+					podList, err := kube.GetPodsByNode(clusterCheckCtx, clientset.CoreV1().Pods(""), node.Name)
+					if err != nil {
+						t.Fatal(err)
+					}
+					pods := kube.FilterHostnet(podList.Items)
+					kube.ValidateNNC(t, *nnc, int64(len(pods)), false)
+				}
```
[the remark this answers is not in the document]
rbtr: tests do not print logs unless they fail 🙁

camrynl (dismissed):
thanks for the explanations 👍

test/integration/k8s_test.go
```diff
@@ -184,10 +184,34 @@ func TestPodScaling(t *testing.T) {
...
+				defer cancel()
+				nnccli, err := nodenetworkconfig.NewClient(restConfig)
+				if err != nil {
+					t.Fatal(err)
+				}
+				nodes, err := clientset.CoreV1().Nodes().List(clusterCheckCtx, metav1.ListOptions{})
+				if err != nil {
+					t.Fatal(err)
+				}
+				for _, node := range nodes.Items {
```
nairashu, line 201: This should run into a rangeVal lint exception that will get thrown:
https://go-critic.com/overview.html#rangeValCopy-ref

test/integration/kube/nnc.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"github.com/Azure/azure-container-networking/crd/nodenetworkconfig/api/v1alpha"
+	"github.com/stretchr/testify/assert"
+)
+
+func ValidateNNC(t *testing.T, nnc v1alpha.NodeNetworkConfig, pods int64, subnetScarcity bool) {
```
nairashu, line 11: The only place this is being validated is where subnet scarcity is set to false in the k8s_test. Should we not be validating the NNC after scaling up the pods such that subnet scarcity is true and then again after scaling down when subnet scarcity is false again?

test/integration/utils_test.go
```diff
@@ -31,8 +27,8 @@ const (
 	SubnetNameLabel        = "kubernetes.azure.com/podnetwork-subnet"
 )
 
-func mustGetClientset() (*kubernetes.Clientset, error) {
-	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+func mustGetClientset(kubeconfig string) (*kubernetes.Clientset, error) { //nolint:unused,deadcode // used by test
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
 	if err != nil {
 		return nil, err
```
nairashu, line 33: wrapcheck lint might throw an error here.

nairashu (commented):
Made some minor review comments. Not sure how this is an E2E integration test where the subnet goes from No scarcity (existing validations should work) --> Scarce flag being set (Existing validations + batchSize change validated + cns ip release code validated) --> No Scarcity (Existing validations + batchSize change reverted)

rbtr:
@nairashu 
> This includes functionality to support testing in subnet scarcity mode, but it is hardcoded to false until the clustersubnetstates are being written by RC.

Once the exhausted state is being written by CRD, that will be passed in to the NNC validation instead of `false`. Then, we update the test to build this cluster with a smaller subnet/scale the pods up past the exhaustion limit, and it becomes an e2e for CNS scarcity behavior.
FWIW I manually validated that the NNC validation is correct with the exhausted bool set to true, so this is ready for those changes as soon as they land.

test/integration/k8s_test.go
timraymond: Fair, but this is a test, so I'd prefer optimizing for readability so long as the tests are reasonably fast. The body of this loop is pretty IO-bound on the operations with the k8s API, so I think that would overwhelm the time spent copying and generating garbage.

test/integration/k8s_test.go
```diff
@@ -184,10 +184,34 @@ func TestPodScaling(t *testing.T) {
...
-				return
+				t.Error("some pods don't have IPs")
 			}
 
+			t.Run("nodenetworkconfig has converged", func(t *testing.T) {
+				clusterCheckCtx, cancel := context.WithTimeout(ctx, 20*time.Minute)
+				defer cancel()
+				nnccli, err := nodenetworkconfig.NewClient(restConfig)
+				if err != nil {
+					t.Fatal(err)
```
timraymond, line 195: Not sure how descriptive the errors are here, would it be worth editorializing them a bit?

test/integration/k8s_test.go
```diff
@@ -184,10 +184,34 @@ func TestPodScaling(t *testing.T) {
 				}
 				t.Log("all pods have been allocated IPs")
 			}) {
-				errors.New("Pods don't have IP's")
-				return
+				t.Error("some pods don't have IPs")
 			}
 
+			t.Run("nodenetworkconfig has converged", func(t *testing.T) {
+				clusterCheckCtx, cancel := context.WithTimeout(ctx, 20*time.Minute)
```
timraymond, line 191: Would it be worthwhile to bind this to the test suite's timeout as well? That way it's the lesser of that timeout and 20 minutes?

rbtr:
@rsagasthya pls review

PR #1636, opened 2022-09-28
pipeline testing for azure-ipam/cni-dropgz submodules

camrynl:
**Reason for Change**:

Current state of e2e tests changes to dropgz/azure-ipam on every run regardless of whether or not change is present. 
Adding submodule pipeline that will only trigger runs if there are changes in the azure-ipam or dropgz directories. This pipeline will test the **cni-dropgz-test image**.  
When changes are not present, the e2e tests will run only on the regular **cni-dropgz image**, which is built from the cni installer that utilizes released versions of components.

New Submodule Pipeline: https://dev.azure.com/msazure/One/_build?definitionId=292133&_a=summary

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


camrynl:
validated pipeline tests via manual run: https://dev.azure.com/msazure/One/_build/results?buildId=61311663&view=results

unable to run the full submodule pipeline via PR trigger since submodules-pipeline.yaml is not merged to master and does not have access to pipeline secrets (for Docker login)

Only testing ipam/dropgz against swifte2e and ciliume2e scenarios. Is there anything else we want to include/exclude in the pipeline? @rbtr

.pipelines/submodules-pipeline.yaml
```diff
@@ -0,0 +1,300 @@
...
+    - "zapai/*"
+    - "azure-ipam/*"
+    - "dropgz/*"
+    exclude:
+    - docs
+  tags:
+    include:
+    - "zapai/*"
+    - "azure-ipam/*"
+    - "dropgz/*"
```
rbtr, line 22: for includes to work as intended here, do we need to exclude everything else?
```suggestion
pr:
  branches:
    include:
      - master
  paths:
    exclude:
    - "*"
    include:
    - "zapai/*"
    - "azure-ipam/*"
    - "dropgz/*"

trigger:
  paths:
    exclude:
    - "*"
    include:
    - "zapai/*"
    - "azure-ipam/*"
    - "dropgz/*"
  tags:
    exclude:
    - "*"
    include:
    - "zapai/*"
    - "azure-ipam/*"
    - "dropgz/*"
```

.pipelines/submodules-pipeline.yaml
```diff
@@ -0,0 +1,300 @@
...
+        variables:
+          STORAGE_ID: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.StorageID'] ]
+        pool:
+          name: "$(BUILD_POOL_NAME_DEFAULT)"
+        steps:
+          - script: |
+              make tools
+              # run test, echo exit status code to fd 3, pipe output from test to tee, which splits output to stdout and go-junit-report (which converts test output to report.xml), stdout from tee is redirected to fd 4. Take output written to fd 3 (which is the exit code of test), redirect to stdout, pipe to read from stdout then exit with that status code. Read all output from fd 4 (output from tee) and write to top stdout
+              { { { {
+                    sudo -E env "PATH=$PATH" make test-all;
```
rbtr, line 66: seems too broad

.pipelines/singletenancy/ipam-dropgz/ipam-dropgz-cilium-e2e-job-template.yaml
```diff
@@ -0,0 +1,29 @@
+parameters:
```
rbtr, line 1: Copy/pasting like this is going to create additional maintenance overhead, and the designs will immediately begin to drift apart. We should evaluate what is actually different between the two pipelines and parameterize the templates accordingly so that they are generic enough to run in both scenarios.

rbtr (changes requested):

.pipelines/submodules-pipeline.yaml
camrynl: yeah, this would make more sense

.pipelines/submodules-pipeline.yaml
```diff
@@ -0,0 +1,319 @@
...
+            npm:
+              name: npm
+              platforms: linux/amd64 linux/arm64 windows/amd64
+        steps:
+          - template: containers/manifest-template.yaml
+            parameters:
+              name: $(name)
+              platforms: $(platforms)
+  
+  # - template: singletenancy/ipam-dropgz/ipam-dropgz-cilium-e2e-job-template.yaml
```
estebancams, line 271: I see this template commented out, should it be enabled? I couldn't find any other reference to ipam-dropgz-cilium-e2e-job or the swift one

camrynl: I commented while I was testing a new parameter with the original swifte2e and ciliume2e template. I am planning to remove the ipam-dropgz templates

camrynl: I'm restructuring things based off of the feedback I've gotten so far

camrynl:
- enabled access of pipeline secrets to forks in repo and validated pipeline with PR trigger: https://dev.azure.com/msazure/One/_build/results?buildId=61950645&view=results

- test is now using swifte2e/ciliume2e templates with new parameter testDropgz set to true.

- unit test stage has been reduced to only testing azure-ipam

estebancams (dismissed):

rbtr (dismissed):

estebancams (approved):

PR #1642, opened 2022-10-03
feat: expose getHomeAzInfo api in cns to retrieve node home az infos from NMAgent

ZetaoZhuang:
**Reason for Change**:

To enable AZR , dnc need to call CNS to get home az infos from NMAgent. This PR is for exposing the getHomeAzInfo api in cns to achieve that.

caching homeaz infos will be covered in separated PR

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:
[task link](https://msazure.visualstudio.com/One/_workitems/edit/15662388)


timraymond (changes requested):

cns/nmagent/client.go
```diff
@@ -236,6 +237,30 @@ func GetNmAgentSupportedApis(httpc *http.Client, getNmAgentSupportedApisURL stri
 	return xmlDoc.SupportedApis, nil
 }
 
+func GetHomeAzInfo() (*http.Response, error) {
```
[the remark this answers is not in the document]
ZetaoZhuang: updated. thanks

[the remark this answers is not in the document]
timraymond: Hey Zetao, the package is still incorrect, I'm afraid. The fully-qualified package is `github.com/Azure/azure-container-networking/nmagent`, not `github.com/Azure/azure-container-networking/cns/nmagent`. Sorry for the confusion... I have a concurrent PR to remove the `./cns/nmagent` package.

timraymond (approved):

timraymond (approved):

cns/restserver/internalapi.go
```diff
@@ -187,7 +187,7 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
 	if len(outdatedNCs) == 0 {
 		return nil
 	}
-	ncVersionListResp, err := service.nmagentClient.GetNCVersionList(ctx)
+	ncVersionListResp, err := service.nmagentMultiClient.OldClient.GetNCVersionList(ctx)
```
ashvindeodhar, line 190: why is this using OldClient?

cns/restserver/restserver.go
```diff
@@ -38,9 +40,16 @@ var (
 type interfaceGetter interface {
 	GetInterfaces(ctx context.Context) (*wireserver.GetInterfacesResult, error)
 }
-
+type NodeInquirer interface {
+	GetHomeAzInfo(ctx context.Context) (nma.HomeAzInfo, error)
+}
+type NmagentMultiClient struct {
+	OldClient nmagentClient
```
ashvindeodhar, line 47: why do we need multiple clients?

cns/service/main.go
```diff
@@ -546,6 +559,11 @@ func main() {
 		return
 	}
 
+	nmagentClients := restserver.NmagentMultiClient{
```
ashvindeodhar, line 562: maybe we can have a call to understand why we need this. Let's discuss these comments in tomorrow's sync

cns/types/codes.go
```diff
@@ -40,6 +40,8 @@ const (
 	UnsupportedNCVersion                   ResponseCode = 38
 	FailedToRunIPTableCmd                  ResponseCode = 39
 	NilEndpointStateStore                  ResponseCode = 40
+	NmAgentNotSupportedAPIError            ResponseCode = 41
```
ashvindeodhar, line 43: how about NmAgentUnsupportedAPIError

nmagent/client.go
```diff
@@ -154,6 +154,33 @@ func (c *Client) DeleteNetworkContainer(ctx context.Context, dcr DeleteContainer
 	return nil
 }
 
+// GetHomeAzInfo gets node's home az info from nmagent
+func (c *Client) GetHomeAzInfo(ctx context.Context) (HomeAzInfo, error) {
+	getHomeAzInfoRequest := &GetHomeAzInfoRequest{}
+	var out HomeAzInfo
```
ashvindeodhar, line 160: can out be renamed to something that better reflects what it is? Is this the style used in other places in this file as well?

nmagent/client_test.go
```diff
@@ -432,3 +432,73 @@ func TestNMAgentDeleteNC(t *testing.T) {
...
+		shouldErr bool
+	}{
+		{
+			"happy path",
+			nmagent.HomeAzInfo{HomeAz: "az01"},
+			"/machine/plugins/?comp=nmagent&type=GetHomeAzInfo",
+			map[string]interface{}{
+				"httpStatusCode": "200",
+				"HomeAz":         "az01",
+				"DcmtRegion":     "uswest2",
```
ashvindeodhar, line 451: we won't be getting dcmtregion anymore. HomeAz returned would be 1 or 01. Let;s ask for the doc from Alok to finalize these values.

nmagent/client_test.go
```diff
@@ -432,3 +432,73 @@ func TestNMAgentDeleteNC(t *testing.T) {
...
+			},
+			false,
+		},
+		{
+			"empty response",
+			nmagent.HomeAzInfo{},
+			"/machine/plugins/?comp=nmagent&type=GetHomeAzInfo",
+			map[string]interface{}{
+				"httpStatusCode": "404",
```
ashvindeodhar, line 460: they should never return 404 if the api is supported. It can be 500. It can also be 401 - in which case we need to treat it as fatal / non-retryable error. We should create alerts on those

nmagent/client.go
ZetaoZhuang: Ya. in some places.

cns/restserver/api.go
```diff
@@ -765,6 +766,76 @@ func (service *HTTPRestService) setOrchestratorType(w http.ResponseWriter, r *ht
 	logger.Response(service.Name, resp, resp.ReturnCode, err)
 }
 
+func (service *HTTPRestService) getHomeAzInfo(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
ZetaoZhuang: Ya. the cache part will be covered in separated PR. Its another sub task.

cns/restserver/api_test.go
```diff
@@ -820,6 +820,50 @@ func TestNmAgentSupportedApisHandler(t *testing.T) {
 	fmt.Printf("nmAgentSupportedApisHandler Responded with %+v\n", nmAgentSupportedApisResponse)
 }
 
+func TestGetHomeAzInfo_UnsupportedHttpMethod(t *testing.T) {
+	fmt.Println("Test: GetHomeAzInfoHandler_UnsupportedHttpMethod")
```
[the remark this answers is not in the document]
timraymond: It should be removed. The test runner outputs the name of the test (in a de-facto standard way) if you ask for verbose output (`-v`). This adds redundant noise.

cns/types/codes.go
```diff
@@ -116,6 +119,12 @@ func (c ResponseCode) String() string {
 		return "UnsupportedOrchestratorType"
 	case UnsupportedVerb:
 		return "UnsupportedVerb"
+	case NmAgentUnSupportedAPIError:
+		return "NmAgentUnSupportedAPIError"
+	case NmAgentServerInternalError:
```
sakmittal, line 124: Same as above

smittal22 (dismissed):
Looks good to me.

cns/restserver/internalapi.go
```diff
@@ -389,3 +389,61 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
+// GetHomeAz gets home AZ from cache or from nmagent
+func (service *HTTPRestService) GetHomeAz(ctx context.Context) cns.GetHomeAzResponse {
+	homeAz := service.readHomeAzCache()
+	if homeAz != "" {
+		return cns.GetHomeAzResponse{HomeAzResponse: nmagent.HomeAzResponse{HomeAz: homeAz}}
+	}
+	suppApisCache := service.readSupportedApisCache()
+	if len(suppApisCache) == 0 || !isAPISupportedByNMAgent(suppApisCache, getHomeAzAPIName) { // to account for if nma getting updated underneath
+		// Get NMAgent supported apis
+		supportedAPIs, err := service.nma.SupportedAPIs(ctx)
```
ashvindeodhar, line 402: can code between lines 401 - 410 be extracted into a separate func like updateNMASupportedAPICache? The goal of that func is to query NMA and update the cache.

cns/restserver/internalapi.go
```diff
@@ -389,3 +389,61 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
+		if err != nil {
+			returnMessage := fmt.Sprintf("[Azure CNS] Error. getHomeAz failed to get NMAgent supported Apis %v", err)
```
ashvindeodhar, line 404: IMO the return msg may be a bit misleading. It should indicate that the "xx" API call to NMA failed. xx - correct API name exposed by NMA to get supported APIs

cns/restserver/internalapi.go
```diff
@@ -389,3 +389,61 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
 
 	return returnCode
 }
+
+// GetHomeAz gets home AZ from cache or from nmagent
+func (service *HTTPRestService) GetHomeAz(ctx context.Context) cns.GetHomeAzResponse {
+	homeAz := service.readHomeAzCache()
+	if homeAz != "" {
+		return cns.GetHomeAzResponse{HomeAzResponse: nmagent.HomeAzResponse{HomeAz: homeAz}}
+	}
```
ashvindeodhar, line 398: As discussed today, this func will only check the cache and return the value and cached NMA response

cns/restserver/internalapi_test.go
```diff
@@ -381,6 +381,32 @@ func TestReconcileNCWithSystemPods(t *testing.T) {
 	validateNCStateAfterReconcile(t, req, expectedNcCount, expectedAssignedPods)
 }
 
+func TestGetHomeAz(t *testing.T) {
+	expHomeAz := "01"
```
ashvindeodhar, line 385: Now that we know homeAZ will be UINT, let's make that change here

cns/restserver/internalapi_test.go
```diff
@@ -381,6 +381,32 @@ func TestReconcileNCWithSystemPods(t *testing.T) {
...
+				"GetHomeAz",
+			}, nil
+		},
+		GetHomeAzF: func(_ context.Context) (nma.HomeAzResponse, error) {
+			return nma.HomeAzResponse{
+				HomeAz: expHomeAz,
+			}, nil
+		},
+	}
+	cleanup := setMockNMAgent(svc, mnma)
```
ashvindeodhar, line 398: why is the mockNMA named as cleanup?

cns/restserver/restserver.go
```diff
@@ -67,6 +68,8 @@ type HTTPRestService struct {
 	dncPartitionKey    string
 	EndpointState      map[string]*EndpointInfo // key : container id
 	EndpointStateStore store.KeyValueStore
+	supportedApisCache []string
```
ashvindeodhar, line 71: the name doesn;t indicate that this is supported APIs from NMA. Need the name to reflect that

cns/restserver/util.go
```diff
@@ -923,3 +923,32 @@ func (service *HTTPRestService) createNetworkContainers(createNetworkContainerRe
...
 	}
 }
+
+func (service *HTTPRestService) setResponse(w http.ResponseWriter, returnCode types.ResponseCode, response interface{}) {
+	serviceErr := service.Listener.Encode(w, &response)
+	logger.Response(service.Name, response, returnCode, serviceErr)
+}
+
+func (service *HTTPRestService) readSupportedApisCache() []string {
+	service.RLock()
```
ashvindeodhar, line 933: it's not advisable to use defers when it's not needed. It comes with some perf penalty. 
https://lk4d4.darth.io/posts/defer/

nmagent/client.go
```diff
@@ -235,6 +235,33 @@ func (c *Client) GetNCVersionList(ctx context.Context) (NCVersionList, error) {
 	return out, nil
 }
 
+// GetHomeAz gets node's home az info from nmagent
```
ashvindeodhar, line 238: nit: drop 'info'. From other place in this func as well

nmagent/client_test.go
```diff
@@ -622,3 +622,71 @@ func TestGetNCVersionList(t *testing.T) {
...
+		resp      map[string]interface{}
+		shouldErr bool
+	}{
+		{
+			"happy path",
+			nmagent.HomeAzResponse{HomeAz: "01"},
+			"/machine/plugins/?comp=nmagent&type=GetHomeAz",
+			map[string]interface{}{
+				"httpStatusCode": "200",
+				"HomeAz":         "01",
```
ashvindeodhar, line 640: change to UINT

nmagent/responses.go
```diff
@@ -36,3 +36,7 @@ type NCVersion struct {
 type NCVersionList struct {
 	Containers []NCVersion `json:"networkContainers"`
 }
+
+type HomeAzResponse struct {
+	HomeAz string `json:"HomeAz"`
```
ashvindeodhar, line 41: UINT

ashvindeodhar (changes requested):

cns/restserver/api.go
```diff
@@ -1461,7 +1494,11 @@ func (service *HTTPRestService) nmAgentSupportedApisHandler(w http.ResponseWrite
 		if err != nil {
 			returnCode = types.NmAgentSupportedApisError
 			returnMessage = fmt.Sprintf("[Azure-CNS] %s", retErr.Error())
+		} else {
+			// caching supportedApis value
+			service.updateSupportedApisCache(apis)
```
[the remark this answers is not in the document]
ZetaoZhuang: I think we should cache the support APIs. and we will not miss out the underneath version update on nmagent because when we find out the api list cache does not contain the target api, we will make call to get the updated api list from nmagent and take different action based on the response

cns/restserver/internalapi_test.go
ZetaoZhuang: because the setMockNMAgent  return a cleanup function that will restore NMAgent back to what it was

cns/restserver/util.go
rbtr: That _was_ true. Note that this linked blog is from 2014 which means it was written when the latest version of Go was <1.5.
If we fast-forward a little, even as far back as [Go 1.14 the performance gap was **significantly** smaller](https://rakyll.org/inlined-defers/).
We're all the way up to go1.19 now and defer efficiency has been continuously improved. IMO, this is not a good reason to _not_ use `defer` anymore, except in the most performance-critical paths. I doubt we have anything running with such tight timing requirements that we care, and certainly not in this codepath.

cns/NetworkContainerContract.go
```diff
@@ -16,6 +16,7 @@ import (
 // Container Network Service DNC Contract
 const (
 	SetOrchestratorType                      = "/network/setorchestratortype"
+	GetHomeAz                                = "/gethomeaz"
```
rbtr, line 19: please make this a good RESTful endpoint, even if the existing ones aren't

cns/api.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/common"
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/crd/nodenetworkconfig/api/v1alpha"
+	nma "github.com/Azure/azure-container-networking/nmagent"
```
rbtr, line 15: @timraymond is this import rename leftover cruft?

cns/api.go
```diff
@@ -335,3 +338,8 @@ type NmAgentSupportedApisResponse struct {
 	Response      Response
 	SupportedApis []string
 }
+
+type GetHomeAzResponse struct {
+	Response       Response           `json:"Response"`
+	HomeAzResponse nma.HomeAzResponse `json:"HomeAzResponse"`
```
rbtr, line 344: JSON keys should not be capitalized

rbtr: further, unless you don't want your JSON keys to match your exported var names for some reason, you don't need explicit `json` struct annotations at all.

cns/restserver/util.go
```diff
@@ -923,3 +923,60 @@ func (service *HTTPRestService) createNetworkContainers(createNetworkContainerRe
...
+	}
+	service.updateGetHomeAzResponseCache(getHomeAzResponse)
+	return getHomeAzResponse
+}
+
+// readNMASupportedApisCache gets the nmsSupportedApisCache value
+func (service *HTTPRestService) readNMASupportedApisCache() []string {
+	service.RLock()
+	supportedApis := service.nmaSupportedApisCache
+	service.RUnlock()
```
rbtr, line 947: I don't think you need to lock the whole service for this, do you? We should have smaller scoped mutexes for things which can be safely accessed in parallel with each other.
/cc @timraymond again

cns/restserver/util.go
```diff
@@ -923,3 +923,60 @@ func (service *HTTPRestService) createNetworkContainers(createNetworkContainerRe
...
+	return supportedApis
+}
```
rbtr, line 949: I know that doing this with a map would **not** be thread-safe; I am not sure if returning a slice like this is thread-safe, since the slice is a sort of reference type and the underlying array can be mutated through any copy of the slice?

nmagent/responses.go
```diff
@@ -36,3 +36,7 @@ type NCVersion struct {
 type NCVersionList struct {
 	Containers []NCVersion `json:"networkContainers"`
 }
+
+type HomeAzResponse struct {
+	HomeAz uint `json:"HomeAz"`
```
rbtr, line 41: all previous JSON comments apply

rbtr (changes requested):
do I understand correctly that nothing is currently populating these "cache" values? why not make them pull-through, so that the first attempt to read the cache does the look-up if the value has not been set? I prefer that to having an out-of-band cache put process.

nmagent/responses.go
rbtr: @ashvindeodhar why uint?

cns/api.go
ZetaoZhuang: per the previous comment

> Please add a json struct tag here. Even if it's identical to the field name, it serves as documentation and it also breaks the binding between the field name and the JSON representation.

cc @timraymond

ZetaoZhuang:
> do I understand correctly that nothing is currently populating these "cache" values? why not make them pull-through, so that the first attempt to read the cache does the look-up if the value has not been set? I prefer that to having an out-of-band cache put process.

Ya, currently nothing is populated to these caches. Based on the design we agreed on, we will have a thread to query home az from nmagent and cache the value when cns starts. @ashvindeodhar

cns/api.go
rbtr: okay, if we want to be explicit and always have the JSON annotations that's fine. They still should not start with a capital letter 🙂

nmagent/responses.go
ashvindeodhar: NMA returns this as UINT. We will be passing this back to NMA when we create NC, so keeping it UINT with us

cns/restserver/util.go
ZetaoZhuang: Ya, the slice is a sort of reference type.  I don't see any scenario we would mutate the copy slice for now. But I can do a deep copy here so it has 0 chance to mutate the underlying cache value.

cns/restserver/util.go
```diff
@@ -923,3 +923,62 @@ func (service *HTTPRestService) createNetworkContainers(createNetworkContainerRe
...
+	service.getHomeAzResponseCacheMux.RUnlock()
+	return response
+}
+
+// updateGetHomeAzResponseCache updates the GetHomeAzResponseCache value
+func (service *HTTPRestService) updateGetHomeAzResponseCache(response cns.GetHomeAzResponse) {
+	service.getHomeAzResponseCacheMux.Lock()
+	service.getHomeAzResponseCache = response
+	service.getHomeAzResponseCacheMux.Unlock()
+}
```
timraymond, line 973: `*HTTPRestService` shouldn't really care that responses for these endpoints are cached. The only thing it cares about is that there is a thing that can get supported APIs and that can also get this HomeAz stuff. This can be better done by composing the client with a new `CachedClient`, and depending on the behavior expected from an NMAgent instead of a concrete implementation. I sketched this out here (apologies for the screencap, it was the only way I could get a decent example without making something contrived):

image

timraymond (changes requested):
I have more thoughts here, but I just want to checkpoint this review so that some feedback can be provided before I'm finished.

nmagent/cachedclient.go
```diff
@@ -0,0 +1,106 @@
...
+// updateHomeAzAndErrorCache update the homeAzResponseCache and homeAzResponseErrorCache value
+func (c *CachedClient) updateHomeAzAndErrorCache(homeAzResponse HomeAzResponse, err error) {
+	c.homeAzMu.Lock()
+	c.homeAzResponseCache = homeAzResponse
+	c.homeAzResponseErrorCache = err
+	c.homeAzMu.Unlock()
+}
+
+// PopulateHomeAzCache makes call to nmagent to retrieve home az if getHomeAz api is supported by nmagent
+func (c *CachedClient) PopulateHomeAzCache(ctx context.Context) error {
```
timraymond, line 87: Instead of this method, why can't the following pseudocode be applied for each shadowed method:

```
// retrieve the cached value
// if none exists, or the TTL indicates that it is stale
       // fetch the value from NMAgent
       // write the value to the cache with an updated TTL
// return the value
```

nmagent/cachedclient.go
```diff
@@ -0,0 +1,106 @@
...
+
+const GetHomeAzAPIName = "GetHomeAz"
+
+type CachedClient struct {
+	*Client
+	// The purpose of nmaSupportedApisCache here is only for reducing the call to nmagent when the getHomeAz api is supported,
+	// but having other unexpected errors while querying homeAz During the retry
+	nmaSupportedApisCache    []string
+	homeAzResponseCache      HomeAzResponse
+	homeAzResponseErrorCache error
```
timraymond, line 20: I don't think it's a good idea to cache errors, because it's more likely that we'll be telling lies to other parts of the system about the state of NMAgent. The same could be said about successes, of course, but at least we know the degree to which those lies extend.

Continuing to say that some dependency is broken, even after it's been fixed just perpetuates the brokenness and allows it to haunt those on-call.

nmagent/cachedclient_test.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+		{
+			"empty response",
+			nmagent.HomeAzResponse{},
+			"/machine/plugins/?comp=nmagent&type=GetHomeAz",
+			map[string]interface{}{
+				"httpStatusCode": "500",
+			},
+			true,
+		},
```
timraymond, line 41: These tests make sense for the real client, but for a cache, we want to test something entirely different: does the cache invoke the real client when we expect it to, and does it serve from the cache when we expect it to.

cns/restserver/api.go
```diff
@@ -763,6 +765,71 @@ func (service *HTTPRestService) setOrchestratorType(w http.ResponseWriter, r *ht
...
+					return
+				}
+			}
+			returnCode := types.UnexpectedError
+			returnMessage := fmt.Sprintf("[Azure CNS] Error. getHomeAz failed %v", err)
+			service.setResponse(w, returnCode, cns.GetHomeAzResponse{
+				Response: cns.Response{ReturnCode: returnCode, Message: returnMessage},
+			})
+			return
+		}
```
timraymond, line 819: I really don't want us to be in the business of re-litigating every error we get from something and figuring out what it means. That just spreads intrinsic knowledge about how subsystems work out to components that shouldn't have it (creating N places to change when that knowledge changes instead of 1). My intention with the nmagent client was that the error tells you what went wrong, so ideally the error handling looks as boring as:

```go
homeAzResponse, err := service.nma.GetHomeAz(ctx)
if err != nil {
       service.setResponse(w, returnCode, cns.GetHomeAzResponse{
              Response: cns.Response{
                     ReturnCode: NMAgentFailed, // constantize this somewhere
                     Message: err.Error()
              }
       })
       return
}
```

This isolates things to what we know in this handler: we tried to do something with NMAgent, but it failed. We don't know what, but the error has all the detail you would need.

The current approach also has too much expectation that the error returned is an `nmagent.Error`. Other types can (and will) be returned.

Why do we need error codes for each possible outcome?

timraymond (changes requested):
Some more questions on the CachedClient (it is looking better!). Also have some other questions about error handling.

nmagent/cachedclient_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+			client := nmagent.NewCachedClient(test.client)
+			client.Start(1)
+
+			// give some time for the thread to complete retrieving home az and update the cache
+			time.Sleep(2 * time.Second)
```
timraymond, line 70: That this is necessary highlights a temporal dependency that shouldn't be here. In reality, a caller could invoke any method immediately after creating the client. Nothing is stopping them.

Also, `time.Sleep` in general is a smell in tests (no matter how much we do it elsewhere). It slows the tests down by default. Also, by failing to address the underlying ordering-of-events problem, it usually means that eventually something else will be added and require even more `time.Sleep` (repeat ad infinitum).

nmagent/cachedclient_test.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+
+			// check the error Cache
+			if errCache != nil && !test.shouldErr {
+				t.Fatal("unexpected error: err:", errCache)
+			}
+			if errCache == nil && test.shouldErr {
+				t.Fatal("expected error but received none")
+			}
+			client.Stop()
```
timraymond, line 84: This should be in a `t.Cleanup()`

nmagent/cachedclient.go
```diff
@@ -0,0 +1,164 @@
...
+// updateHomeAzAndErrorCache update the homeAzResponseCache and homeAzResponseErrorCache value
+func (c *CachedClient) updateHomeAzAndErrorCache(homeAzResponse HomeAzResponse, err error) {
+	c.homeAzMu.Lock()
+	defer c.homeAzMu.Unlock()
+	c.homeAzResponseCache = homeAzResponse
+	c.homeAzResponseErrorCache = err
+}
+
+// Start starts a new thread to refresh home az cache
+func (c *CachedClient) Start(retryIntervalInSecs int) {
```
timraymond, line 104: Rather than take an `int` here, you should take a `time.Duration` to leave units up to the caller. This makes usages self-documenting:

```go
c.Start(2 * time.Second)
```

instead of the cryptic

```go
c.Start(2)
```

nmagent/cachedclient.go
```diff
@@ -0,0 +1,164 @@
...
+		select {
+		case <-c.closing:
+			return
+		default:
+			ctx, cancel := context.WithTimeout(context.Background(), ContextTimeOut)
+			homeAzResponse, populateErr := c.populateHomeAzCache(ctx)
+			// keep retrying when there is an error or getHomeAz api is not supported by nmagent
+			if populateErr != nil || !homeAzResponse.IsSupported {
+				log.Debugf("Failed to retrieve home az from nmagent, will retry. %v", populateErr)
+				time.Sleep(time.Duration(retryIntervalInSecs) * time.Second)
```
timraymond, line 125: I'd prefer this to use a `time.Ticker`, so that `c.closing` has a chance to run outside of the sleep interval.

timraymond (changes requested):

nmagent/cachedclient.go
```diff
@@ -0,0 +1,178 @@
+package nmagent
+
+import (
```
ashvindeodhar, line 3: can you use "github.com/patrickmn/go-cache" ? This has been used by DNC RC. Check the code in nodenetworkconfigreconciler.go

nmagent/client_test.go
```diff
@@ -622,3 +622,71 @@ func TestGetNCVersionList(t *testing.T) {
...
+	tests := []struct {
+		name      string
+		exp       nmagent.HomeAzResponse
+		expPath   string
+		resp      map[string]interface{}
+		shouldErr bool
+	}{
+		{
+			"happy path",
+			nmagent.HomeAzResponse{HomeAz: uint(1)},
```
ashvindeodhar, line 636: are you missing IsSupported ?

nmagent/client_test.go
```diff
@@ -622,3 +622,71 @@ func TestGetNCVersionList(t *testing.T) {
...
+					if err != nil {
+						t.Fatal("unexpected error encoding response: err:", err)
+					}
+					rr.WriteHeader(http.StatusOK)
+					return rr.Result(), nil
+				},
+			})
+
+			got, err := client.GetHomeAz(context.TODO())
+			if err != nil && !test.shouldErr {
```
ashvindeodhar, line 675: should you be testing if the correct values are returned for IsSupported and AZID?

ashvindeodhar (changes requested):

nmagent/cachedclient.go
rbtr: [A little copying is better than a little dependency](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s).
That dependency is 10x bigger than this whole cached-client implementation, and significantly obfuscates the behavior of the client, which IMO should be a very simple pull-through cached client.

cns/types/codes.go
```diff
@@ -40,6 +40,8 @@ const (
 	UnsupportedNCVersion                   ResponseCode = 38
 	FailedToRunIPTableCmd                  ResponseCode = 39
 	NilEndpointStateStore                  ResponseCode = 40
+	NmAgentInternalServerError             ResponseCode = 41
```
ashvindeodhar, line 43: my proposal is to use nma http responses so we won't need to add these

ashvindeodhar (changes requested):

cns/restserver/homeazcache.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// readCacheValue reads home az cache value
+func (h *HomeAzCache) readCacheValue() cns.GetHomeAzResponse {
+	cachedResp, _ := h.values.Get(homeAzCacheKey)
+	return cachedResp.(cns.GetHomeAzResponse)
+}
+
+// Start starts a new thread to refresh home az cache
+func (h *HomeAzCache) Start(retryIntervalInSecs time.Duration) {
```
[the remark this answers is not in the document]
timraymond: I've always found it better to be clear about where goroutines are started and where they die for daemon-like things. Otherwise you lose track of them and they leak. Functions that serve as constructors / factory functions roughly transform configuration into structs that are ready to run. One advantage is that you can have a whole bunch of things that conform to:

```go
type Runner struct {
       Start(context.Context) error
}
```

Collect all of them into a `[]Runner`, and opaquely start all of the goroutines associated with each daemon-like thing. That is, of course, spoiled here because we're accepting additional configuration in `Start()`. I didn't flag it though, since it's an easy change and not a hill I wanted to die on.

cns/restserver/restserver.go
```diff
@@ -128,7 +130,7 @@ type networkInfo struct {
 }
 
 // NewHTTPRestService creates a new HTTP Service object.
-func NewHTTPRestService(config *common.ServiceConfig, wscli interfaceGetter, nmagentClient nmagentClient, endpointStateStore store.KeyValueStore) (cns.HTTPService, error) {
+func NewHTTPRestService(config *common.ServiceConfig, wscli interfaceGetter, nmagentClient nmagentClient, endpointStateStore store.KeyValueStore, homeAzCache *HomeAzCache) (cns.HTTPService, error) {
```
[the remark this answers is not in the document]
ZetaoZhuang: we need it to read the cache value and response to the incoming gethomeaz api call

cns/restserver/homeazmonitor.go
```diff
@@ -0,0 +1,168 @@
...
+	homeAzCacheKey   = "HomeAz"
+)
+
+type HomeAzMonitor struct {
+	nmagentClient
+	values *cache.Cache
+	// channel used as signal to end of the goroutine for populating home az cache
+	closing chan struct{}
+	// channel used as signal to block the Start() of HomeAzMonitor until the first request to retrieve home az completes
+	block                    chan struct{}
```
ashvindeodhar, line 29: nit: space after block

cns/restserver/homeazmonitor.go
```diff
@@ -0,0 +1,168 @@
...
+func (h *HomeAzMonitor) readCacheValue() cns.GetHomeAzResponse {
+	cachedResp, _ := h.values.Get(homeAzCacheKey)
+	return cachedResp.(cns.GetHomeAzResponse)
+}
+
+// Start starts a new thread to refresh home az cache
+func (h *HomeAzMonitor) Start() {
+	go h.refresh()
+	// block until the first request to nmagent for retrieving home az completes
+	<-h.block
```
ashvindeodhar, line 64: why do need to block here?

cns/restserver/homeazmonitor.go
```diff
@@ -0,0 +1,168 @@
...
+				return
+			}
+		}
+		returnMessage := fmt.Sprintf("[HomeAzMonitor] failed with Error. %v", err)
+		returnCode := types.UnexpectedError
+		h.update(returnCode, returnMessage, cns.HomeAzResponse{IsSupported: true})
+		return
+	}
+
+	h.update(types.Success, "Get Home Az successfully", cns.HomeAzResponse{IsSupported: true, HomeAz: azResponse.HomeAz})
```
ashvindeodhar, line 144: nit: succeeded

cns/restserver/homeazmonitor.go
```diff
@@ -0,0 +1,168 @@
...
+	values *cache.Cache
+	// channel used as signal to end of the goroutine for populating home az cache
+	closing chan struct{}
+	// channel used as signal to block the Start() of HomeAzMonitor until the first request to retrieve home az completes
+	block                    chan struct{}
+	cacheRefreshIntervalSecs time.Duration
+}
+
+// New creates a new HomeAzMonitor object
+func New(client nmagentClient, cacheRefreshIntervalSecs time.Duration) *HomeAzMonitor {
```
ashvindeodhar, line 34: should this be renamed to NewHomeAzMonitor? ( which will be inline with NewHTTPRestService )
with just new, the invocation looks like restserver.New() return HomeAzMonitor.

cns/service/main.go
```diff
@@ -830,6 +833,9 @@ func main() {
 		}
 	}
 
+	logger.Printf("end the goroutine for refreshing homeAz")
```
ashvindeodhar, line 836: do you have a corresponding log indicating the monitor started?

cns/restserver/homeazmonitor.go
ZetaoZhuang: we want to make sure the home az cache is not empty (at least wait until the first request to retrieve home az from nmagent completed)

ashvindeodhar: what's the issue with having the empty cache when CNS gets a request from DNC? Can CNS not respond accordingly if the cache is empty indicating internal error?

ashvindeodhar: with this block we are adding latency to CNS start and I am looking to see if we can avoid that

ZetaoZhuang: understand that. we can response internal error if the cache is empty

ashvindeodhar (approved):

rbtr:
@ashvindeodhar please leave a note explaining why this was merged bypassing checks

ashvindeodhar:
@rbtr  sorry, I added a note but seems didn't hit the 'comment' button.
Merging this bypassing the windows check which has been failing for last few days. We need this change for making progress on aquarius changes and testing the critical AZR feature in timely fashion. This change is not updating any existing code / functionality so risk of any windows regressions are very low. This code adds new functionality which is currently not used. After checking with pipeline v-team ( Esteban ), I came to know that this might need few more days to fix this windows pipeline issue.

PR #1643, opened 2022-10-03
Replace the NMAgent client in CNS with the one from the nmagent package

timraymond:
The NMAgent client used previously in CNS is older, and a duplication of the one added in `nmagent/`. In order to maintain one nmagent client and prevent duplication / wasted effort, this changes CNS to use the "official" client instead.

- [X] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/Azure/azure-container-networking/cns/types"
+	"github.com/Azure/azure-container-networking/nmagent"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 )
@@ -496,7 +497,7 @@ type PublishNetworkContainerRequest struct {
 	NetworkContainerID                string
 	JoinNetworkURL                    string
 	CreateNetworkContainerURL         string
-	CreateNetworkContainerRequestBody []byte
+	CreateNetworkContainerRequestBody nmagent.PutNetworkContainerRequest
 }
 
 // NetworkContainerParameters parameters available in network container operations
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -1117,7 +1117,6 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
 	ctx := r.Context()
 
 	var (
-		err                 error
 		req                 cns.PublishNetworkContainerRequest
 		returnCode          types.ResponseCode
 		returnMessage       string
@@ -1127,7 +1126,7 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
 		isNetworkJoined     bool
 	)
 
-	err = service.Listener.Decode(w, r, &req)
+	err := service.Listener.Decode(w, r, &req)
 
 	creteNcURLCopy := req.CreateNetworkContainerURL
 
@@ -1172,7 +1171,7 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
 		// Please refactor this
 		// do not reuse the below variable between network join and publish
 		// nolint:bodyclose // existing code needs refactoring
-		err := service.joinNetwork(req.NetworkID)
+		err = service.joinNetwork(req.NetworkID)
 		if err != nil {
 			returnMessage = err.Error()
 			returnCode = types.NetworkJoinFailed
@@ -1308,7 +1307,7 @@ func (service *HTTPRestService) unpublishNetworkContainer(w http.ResponseWriter,
 				AuthenticationToken: ncParameters.AuthToken,
 			}
 
-			err := service.nma.DeleteNetworkContainer(ctx, dcr)
+			err = service.nma.DeleteNetworkContainer(ctx, dcr)
 			if err != nil {
 				returnMessage = fmt.Sprintf("Failed to unpublish Network Container: %s", req.NetworkContainerID)
 				returnCode = types.NetworkContainerUnpublishFailed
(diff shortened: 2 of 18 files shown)


cns/service/main.go
```diff
@@ -312,8 +313,12 @@ func printVersion() {
 	fmt.Printf("Version %v\n", version)
 }
 
+type NodeInquirer interface {
```
rbtr, line 316: I'm imagining this with a top hat and monocle. I would have gone with "Interrogator" 😅

rbtr (approved):

cns/restserver/util.go
```diff
@@ -643,21 +642,21 @@ func (service *HTTPRestService) setNetworkStateJoined(networkID string) {
 }
 
 // Join Network by calling nmagent
-func (service *HTTPRestService) joinNetwork(
-	networkID string,
-) (*http.Response, error, error) {
-	var err error
-	joinResponse, joinErr := nmagent.JoinNetwork(networkID)
+func (service *HTTPRestService) joinNetwork(ctx context.Context, networkID string) error {
+	jnr := nma.JoinNetworkRequest{
```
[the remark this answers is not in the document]
timraymond: yeah... that's a habit I picked up from a C++ codebase lol. I like `req` too.

cns/service/main.go
timraymond: I like `Interrogator`. I wanted to separate out the read ops from the write ops. `NodeReader` sounded... weird and misleading. `Interrogator` does say what it does what it says on the tin... it interrogates nodes for information. I guess that means the writing side should be `NodeInceptor` 😆

cns/restserver/util.go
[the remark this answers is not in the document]
timraymond: Okay, this was done in 7519cbd

cns/service/main.go
timraymond: Updated in 9381406

cns/restserver/api.go
```diff
@@ -17,10 +16,9 @@ import (
 	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/cns/hnsclient"
 	"github.com/Azure/azure-container-networking/cns/logger"
-	"github.com/Azure/azure-container-networking/cns/nmagent"
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/cns/wireserver"
-	"github.com/Azure/azure-container-networking/common"
+	nma "github.com/Azure/azure-container-networking/nmagent"
```
ramiro-gamarra, line 21: package is used as `nmagent` in other places, any reason to rename it in some places and not on others? personally, I think renaming packages should be done as little as possible. if it's widely understood that `nma` refers to nmagent, then the package should just be `nma`.

cns/restserver/internalapi.go
```diff
@@ -96,22 +96,22 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+		}
+
+		ncVersionURLs.Store(nc.NetworkContainerid, nmaReq)
 		waitingForUpdate, _, _ := service.isNCWaitingForUpdate(nc.Version, nc.NetworkContainerid)
 
 		body, _ = json.Marshal(nc)
 		req, _ = http.NewRequest(http.MethodPost, "", bytes.NewBuffer(body))
 		req.Header.Set(common.ContentType, common.JsonContent)
+
+		w := httptest.NewRecorder()
```
ramiro-gamarra, line 112: I know this is just moving existing lines, but this is... interesting, to say the least. i suppose its a symptom of all routing + business logic living in the same function.

rbtr (approved):

nmagent/client.go
```diff
@@ -113,6 +114,37 @@ func (c *Client) GetNetworkConfiguration(ctx context.Context, gncr GetNetworkCon
 	return out, err // nolint:wrapcheck // wrapping just introduces noise here
 }
 
+// GetNetworkContainerVersion gets the current goal state version of a Network
```
ramiro-gamarra, line 117: godoc is off

cns/restserver/api.go
timraymond: Any instance of this is an oversight from when there was the old and new packages both named `nmagent`. I agree, but I won't fix this in this PR because it is blocking other feature work. I'll add this to remediation items for a future PR.

nmagent/client.go
timraymond: Noted. Will fix in a future cleanup PR.

cns/service/main.go
```diff
@@ -312,8 +313,14 @@ func printVersion() {
 	fmt.Printf("Version %v\n", version)
 }
 
+// NodeInterrogator is functionality necessary to read information about nodes.
+// It is intended to be strictly read-only.
```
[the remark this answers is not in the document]
timraymond: Yeah, the intention is to set some opinions about what this set of behavior means. If you have some methods which aren't about digging information out of Nodes, then it might belong in another interface.

cns/restserver/internalapi.go
timraymond: Not a fan of `httptest` in something that's not a test anyway... but that's a battle for another day.

cns/NetworkContainerContract.go
```diff
@@ -496,7 +497,7 @@ type PublishNetworkContainerRequest struct {
 	NetworkContainerID                string
 	JoinNetworkURL                    string
 	CreateNetworkContainerURL         string
-	CreateNetworkContainerRequestBody []byte
+	CreateNetworkContainerRequestBody nmagent.PutNetworkContainerRequest
```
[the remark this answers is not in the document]
rbtr: is that this? https://github.com/Azure/azure-container-networking/pull/1695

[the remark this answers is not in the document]
pjohnst5: Nvm we know about it: https://github.com/Azure/azure-container-networking/commit/31a0906102e272adeaf4be914c46b80a11a398c4

PR #1650, opened 2022-10-10
Emit a metric in CNS for NC Sync count by status

rbtr:

**Reason for Change**:

- emits a metric for total NC sync (attempt) count, labelled by success/failure
- improves the sync method to consider the scenario where not all of the NCs that CNS knows about are present in the NMA response an error, and return an error indicating this after processing the NCs that are present in the response.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -164,75 +164,79 @@ func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMo
 var errNonExistentContainerStatus = errors.New("nonExistantContainerstatus")
 
 func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMode string) error {
-	hostVersionNeedsUpdateContainers := map[string]struct{}{}
+	outdatedNCs := map[string]struct{}{}
 	for idx := range service.state.ContainerStatus {
 		// Will open a separate PR to convert all the NC version related variable to int. Change from string to int is a pain.
-		hostVersion, err := strconv.Atoi(service.state.ContainerStatus[idx].HostVersion)
+		localNCVersion, err := strconv.Atoi(service.state.ContainerStatus[idx].HostVersion)
 		if err != nil {
 			logger.Errorf("Received err when change containerstatus.HostVersion %s to int, err msg %v", service.state.ContainerStatus[idx].HostVersion, err)
 			continue
 		}
-		dncNcVersion, err := strconv.Atoi(service.state.ContainerStatus[idx].CreateNetworkContainerRequest.Version)
+		dncNCVersion, err := strconv.Atoi(service.state.ContainerStatus[idx].CreateNetworkContainerRequest.Version)
 		if err != nil {
 			logger.Errorf("Received err when change nc version %s in containerstatus to int, err msg %v", service.state.ContainerStatus[idx].CreateNetworkContainerRequest.Version, err)
 			continue
 		}
 		// host NC version is the NC version from NMAgent, if it's smaller than NC version from DNC, then append it to indicate it needs update.
-		if hostVersion < dncNcVersion {
-			hostVersionNeedsUpdateContainers[service.state.ContainerStatus[idx].ID] = struct{}{}
-		} else if hostVersion > dncNcVersion {
-			logger.Errorf("NC version from NMAgent is larger than DNC, NC version from NMAgent is %d, NC version from DNC is %d", hostVersion, dncNcVersion)
+		if localNCVersion < dncNCVersion {
+			outdatedNCs[service.state.ContainerStatus[idx].ID] = struct{}{}
+		} else if localNCVersion > dncNCVersion {
+			logger.Errorf("NC version from NMAgent is larger than DNC, NC version from NMAgent is %d, NC version from DNC is %d", localNCVersion, dncNCVersion)
 		}
 	}
-	if len(hostVersionNeedsUpdateContainers) == 0 {
+	if len(outdatedNCs) == 0 {
 		return nil
 	}
-	ncList, err := service.nmagentClient.GetNCVersionList(ctx)
+	ncVersionListResp, err := service.nmagentClient.GetNCVersionList(ctx)
 	if err != nil {
 		return errors.Wrap(err, "failed to get nc version list from nmagent")
 	}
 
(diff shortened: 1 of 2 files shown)


cns/restserver/internalapi.go
```diff
@@ -157,13 +157,14 @@ func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMo
 	if err != nil {
 		logger.Errorf("sync host error %v", err)
```
neaggarwMS, line 158: Will this get logged if given NC is missing in NMA get call?

cns/restserver/internalapi.go
```diff
@@ -157,13 +157,14 @@ func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMo
...
 	}
```
neaggarwMS, line 159: We dont know here if there is some issue in SyncHostNC API or NMA not able to program a given NC. We need to differentiate it, may be add a error reason.

neaggarwMS (changes requested):

cns/restserver/internalapi.go
```diff
@@ -157,13 +157,14 @@ func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMo
...
-	syncHostNCVersion.WithLabelValues(strconv.FormatBool(err == nil)).Observe(time.Since(start).Seconds())
+	syncHostNCVersionCount.WithLabelValues(strconv.FormatBool(err == nil)).Inc()
```
csfmomo, line 160: Not sure if we inc non err label, how do we set alert? How do we know that there is an error with syncHostNCVersion?

rbtr: we have this label "ok", and it will be set to `0` or `1` based on `err == nil`, and we will increment the correct series based on that. 
the alert would be based on the value of `sync_host_nc_version_total{"ok=0"}[15m]` or something like that.

cns/restserver/internalapi.go
```diff
@@ -195,28 +196,43 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
 		ncInfo, exist := service.state.ContainerStatus[ncID]
 		if !exist {
 			return errors.Wrapf(errNonExistentContainerStatus, "can't find NC with ID %s in service state, stop updating this host NC version", ncID)
 		}
+		hostVersion, err := strconv.Atoi(ncInfo.HostVersion)
+		if err != nil {
+			return errors.Wrapf(err, "failed to parse host nc version string %s", ncInfo.HostVersion)
+		}
+		if hostVersion > version {
+			logger.Errorf("NC version from NMA is decreasing: have %d, got %d", hostVersion, version)
```
neaggarwMS, line 219: is hostVersion the last good version provided by host or it is same as nc?

neaggarwMS: If former, then instead of comparing it with hostVersion we need to compare it with NC version to match it is same or higher than NC new version

neaggarwMS (changes requested):

cns/restserver/internalapi.go
csfmomo: I think this logic is already covered from line 180 to 182

csfmomo: > If former, then instead of comparing it with hostVersion we need to compare it with NC version to match it is same or higher than NC new version

I think this logic is already covered from line 180 to 182

csfmomo: Compare the hostVersion with version is to check whether NMA's version gone backwards. For example, last NMA's version is 10 and we already update all the pending programming IPs, then we got a version 11 from DNC, then CNS try to sync NMA, NMA returns version 9 and we want to  log this error.

csfmomo: > If former, then instead of comparing it with hostVersion we need to compare it with NC version to match it is same or higher than NC new version
Compare the hostVersion with version is to check whether NMA's version gone backwards. For example, last NMA's version is 10 and we already update all the pending programming IPs, then we got a version 11 from DNC, then CNS try to sync NMA, NMA returns version 9 and we want to log this error.

cns/restserver/internalapi.go
```diff
@@ -195,28 +196,43 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
 		// Check whether it exist in service state and get the related nc info
 		ncInfo, exist := service.state.ContainerStatus[ncID]
 		if !exist {
 			return errors.Wrapf(errNonExistentContainerStatus, "can't find NC with ID %s in service state, stop updating this host NC version", ncID)
 		}
+		hostVersion, err := strconv.Atoi(ncInfo.HostVersion)
+		if err != nil {
+			return errors.Wrapf(err, "failed to parse host nc version string %s", ncInfo.HostVersion)
+		}
+		if hostVersion > version {
```
neaggarwMS, line 218: Can we call it newHostVersion? version is too generic

cns/restserver/internalapi.go
neaggarwMS: i got the flow, this looks good

cns/restserver/internalapi.go
```diff
@@ -195,28 +196,43 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
-	for _, ncID := range hostVersionNeedsUpdateContainers {
+	for ncID := range hostVersionNeedsUpdateContainers {
 		versionStr, ok := newHostNCVersionList[ncID]
 		if !ok {
+			// NMA doesn't have this NC that we need programmed yet, bail out
 			continue
 		}
 		version, err := strconv.Atoi(versionStr)
 		if err != nil {
 			return errors.Wrapf(err, "failed to parse container version of %s", ncID)
```
neaggarwMS, line 207: should we return or append error and continue to move to next NC?

neaggarwMS (approved):

cns/restserver/internalapi.go
rbtr: refactored all the variable names here to make it very clear which is which.

neaggarwMS (approved):

csfmomo (approved):

PR #1652, opened 2022-10-13
test: [WIN-NPM] dataplane test framework

huntergregory:
Framework to specify/test any arbitrary sequence of DP events.


network/hnswrapper/hnsv2wrapperfake.go
```diff
@@ -328,6 +331,35 @@ func (fCache FakeHNSCache) SetPolicy(setID string) *hcn.SetPolicySetting {
 	return nil
 }
 
+func (fCache FakeHNSCache) PrettyString() string {
```
timraymond, line 334: Any reason not to make this implement `io.Stringer` directly?

network/hnswrapper/hnsv2wrapperfake.go
```diff
@@ -400,11 +467,40 @@ func NewFakeHostComputeEndpoint(endpoint *hcn.HostComputeEndpoint) *FakeHostComp
 	}
 }
 
+func (fEndpoint *FakeHostComputeEndpoint) PrettyString() string {
```
timraymond, line 470: I realize that existing methods do this, but the generally accepted style in the wider Go community is for a single letter matching the first letter of the type the method is being implemented on for "this":

```go
func (f *FakeHostComputeEndpoint) PrettyString() string
```

npm/pkg/dataplane/dataplane_windows_test.go
```diff
@@ -0,0 +1,629 @@
...
+		if !ok {
+			continue
+		}
+
+		success = assert.Equal(t, len(expectedACLs), len(cachedACLs), fmt.Sprintf("unexpected number of ACLs for Endpoint with ID: %s", epID)) && success
+		for _, expectedACL := range expectedACLs {
+			foundACL := false
+			for _, cacheACL := range cachedACLs {
+				if expectedACL.ID == cacheACL.ID {
+					if reflect.DeepEqual(expectedACL, cacheACL) {
```
timraymond, line 603: `cmp.Equal` from `google.com/go-cmp/cmp` is categorically better than `reflect.DeepEqual`. The talk that lays out the argument for it is here: https://www.youtube.com/watch?v=OuT8YYAOOVI. There's also a `cmp.Diff` that will make it easier to see _why_ it's not equal.

network/hnswrapper/hnsv2wrapperfake.go
huntergregory: we decided against implementing `io.Stringer` in #1223

network/hnswrapper/hnsv2wrapperfake.go
huntergregory: would you suggest keeping the same style throughout the file or adding new methods in the generally accepted way?

ramiro-gamarra: linters complain if all method receivers are not named the same, so all should be changed.

network/hnswrapper/hnsv2wrapperfake.go
timraymond: Okay, the reasoning makes sense to me there.

npm/pkg/dataplane/dataplane_windows_test.go
```diff
@@ -0,0 +1,757 @@
...
+
+			dp, err := NewDataPlane(thisNode, io, tt.cfg, nil)
+			require.NoError(t, err, "failed to initialize dp")
+
+			for i, session := range tt.sessions {
+				// concurrent case with multiple go routines
+				wg := new(sync.WaitGroup)
+				wg.Add(len(session))
+				threadErrors := make(chan error, len(session))
+				for j, th := range session {
```
tamilmani1989, line 616: since it was named as concurrent sessions, shouldn't each session start in separate process or thread?

huntergregory: poor wording. thanks for pointing out

npm/pkg/dataplane/e2es/e2e_windows_test.go
```diff
@@ -0,0 +1,78 @@
...
+						err = s.HNSAction.Do(hns)
+					} else if s.DPAction != nil {
+						err = s.DPAction.Do(dp)
+					}
+
+					assert.Nil(t, err, "failed to run step in foreground: %s", s.ID)
+				}
+			}
+
+			tt.WaitForAllStepsToComplete()
```
ramiro-gamarra, line 69: _Seems_ like this waits for goroutines added with `MarkStepRunningInBackground` (and completed with `MarkStepComplete`)? Are more steps added in other scopes? Otherwise, would a regular waitgroup accomplish the same?

npm/pkg/dataplane/e2es/e2e_windows_test.go
```diff
@@ -0,0 +1,78 @@
...
+			close(backgroundErrors)
+			for err := range backgroundErrors {
+				assert.Nil(t, err, "failed during concurrency")
```
ramiro-gamarra, line 72: Seems easier to just collect any errors and assert that there were none as opposed to handling nils from an error channel?

npm/pkg/dataplane/e2es/e2e_windows_test.go
```diff
@@ -0,0 +1,74 @@
...
+				} else {
+					if !assert.Nil(t, s.Do(), "failed to run step in foreground: %s", s.ID) {
+						// stop processing steps on a foreground failure
+						break
+					}
+				}
+			}
+
+			tt.WaitForAll()
+			close(backgroundErrors)
```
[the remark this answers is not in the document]
ramiro-gamarra: errgroup can be useful, yes, but it seems like the goal here is to wait for multiple things to complete and verify no errors at the end. errgroup will return the first error, which may not paint the entire picture at the end of processing.

npm/pkg/dataplane/e2es/e2e_windows_test.go
```diff
@@ -0,0 +1,78 @@
...
+					tt.MarkStepRunningInBackground(s.ID)
+					go func(s *TestStep) {
+						defer tt.MarkStepComplete(s.ID)
+
+						var err error
+						if s.HNSAction != nil {
+							err = s.HNSAction.Do(hns)
+						} else if s.DPAction != nil {
+							err = s.DPAction.Do(dp)
+						}
```
ramiro-gamarra, line 51: Is a test step valid if both are nil or both are set?

npm/pkg/dataplane/e2es/e2e_windows_test.go
```diff
@@ -0,0 +1,78 @@
...
+					} else if s.DPAction != nil {
+						err = s.DPAction.Do(dp)
+					}
+
+					assert.Nil(t, err, "failed to run step in foreground: %s", s.ID)
+				}
+			}
+
+			tt.WaitForAllStepsToComplete()
+			close(backgroundErrors)
```
ramiro-gamarra, line 70: I believe there's a deadlock here in case of background errors: since its an unbuffered channel, any goroutine that encounters an error will be blocked on the send to the channel until something can consume it, but the only thing that consumes it starts AFTER all goroutines have completed.

ramiro-gamarra (commented):
Some comments

npm/pkg/dataplane/e2es/e2e_windows_test.go
huntergregory: an Action is valid only if exactly one of HNSAction or DPAction is non-nil

npm/pkg/dataplane/e2es/e2e_windows_test.go
huntergregory: simplified things with a new structure. all go routines and wait groups would now be handled in the actual UT in test.Run()

npm/pkg/dataplane/e2es/e2e_windows_test.go
huntergregory: this design emulates cyclonus (upstream in netpol-api now). The Action here is similar to the Action in Cyclonus, although we use interfaces instead of having a separate "interpreter" for each case
https://github.com/mattfenwick/cyclonus/blob/ec73330a5c88654ed5545ccdbcfad33843722ebd/pkg/generator/action.go#L5

npm/pkg/dataplane/e2es/e2e_windows_test.go
huntergregory: thanks for catching this

npm/pkg/dataplane/e2es/e2e_windows_test.go
huntergregory: updated to require.Empty() for better semantics

npm/pkg/dataplane/dataplane_windows_test.go
```diff
@@ -0,0 +1,112 @@
...
+				_, err := hns.CreateEndpoint(ep)
+				require.Nil(t, err, "failed to create initial endpoint %+v", ep)
+			}
+
+			// the dp is necessary for NPM tests
+			dp, err := NewDataPlane(thisNode, io, tt.DpCfg, nil)
+			require.NoError(t, err, "failed to initialize dp")
+
+			wg := new(sync.WaitGroup)
+			wg.Add(len(tt.Threads))
```
ramiro-gamarra, line 80: Don't mean to be pedantic, but is the "thread" terminology necessary? A thread usually carries the connotation of being an OS level construct, which Go abstracts away in the form of goroutines.

npm/pkg/dataplane/dataplane_windows_test.go
```diff
@@ -0,0 +1,112 @@
...
+				}()
+			}
+
+			wg.Wait()
+			close(backgroundErrors)
+			errStrings := make([]string, len(backgroundErrors))
+			for err := range backgroundErrors {
+				errStrings = append(errStrings, fmt.Sprintf("[%s]", err.Error()))
+			}
+			assert.Empty(t, backgroundErrors, "encountered errors in threaded test: %s", strings.Join(errStrings, ","))
```
ramiro-gamarra, line 109: - the `errStrings` initialization is odd: if there are any errors, the len of `backgroundErrors` will be > 0, but then this appends AFTER the last index.
- asserting on channel length is also unorthodox. im not sure buffered channels are what you need here. my recommendation would be another type for `multierr` that's concurrent safe: you can just append to it from any goroutine. after tests, you just assert that the type contains no errors.

npm/pkg/dataplane/dataplane-test-cases_windows_test.go
```diff
@@ -0,0 +1,152 @@
...
+			PolicyTypes: []networkingv1.PolicyType{
+				networkingv1.PolicyTypeIngress,
+				networkingv1.PolicyTypeEgress,
+			},
+		},
+	}
+}
+
+func getAllSerialTests() []*SerialTestCase {
+	return []*SerialTestCase{
```
ramiro-gamarra, line 80: With this pattern, how do you run a single test?

npm/pkg/dataplane/dataplane_windows_test.go
huntergregory: Perhaps `GoRoutines` then? or `Workstreams`?

npm/pkg/dataplane/dataplane_windows_test.go
huntergregory: can use "go.uber.org/multierr"

network/hnswrapper/hnsv2wrapperfake.go
huntergregory: for the reviewer's sake, can address this in a followup PR since it will create a diff for most of the file

npm/pkg/dataplane/dataplane_windows_test.go
timraymond: I'd agree an `error` type for multiple errors is a good idea, but do we really need another dependency for it?

npm/pkg/dataplane/dataplane_windows_test.go
timraymond: Maybe `Jobs`? `GoRoutines` isn't bad... `Workstreams` is corp-speak though that I would avoid.

npm/pkg/dataplane/dataplane_windows_test.go
huntergregory: to avoid the new dependency, went back to a channel approach, except errString is initialized after checking length of backgroundErrors now

npm/pkg/dataplane/dataplane_windows_test.go
huntergregory: now using `Jobs`

npm/pkg/dataplane/dataplane-test-cases_windows_test.go
```diff
@@ -0,0 +1,448 @@
...
+			},
+			PolicyTypes: []networkingv1.PolicyType{
+				networkingv1.PolicyTypeIngress,
+				networkingv1.PolicyTypeEgress,
+			},
+		},
+	}
+}
+
+func getAllSerialTests() []*SerialTestCase {
```
matmerr, line 90: Do we have any way of injecting expected errors to model nonhappy paths? Suppose we wanted to inject dp/hns returning errors and getting coverage in the controllers around those scenarios

matmerr (approved):
sticking to serial tests for the moment, need to add methods of injecting errors into mock hns/dp for testing nonhappy paths

ck319 (dismissed):

PR #1654, opened 2022-10-14
feat: include swift-overlay.conflist in the dropgz build

tyler-lloyd:
adds the azure-swift-overlay.conflist in the dropgz build file so Azure CNI Overlay clusters can use the dropgz to install the cni.

Signed-off-by: Tyler Lloyd <tylloyd@microsoft.com>

**Reason for Change**:

Adds the [azure-swift-overlay](https://github.com/Azure/azure-container-networking/blob/95155ca9b4079cfe9ce4e93c04de419e28a3ad3c/cni/azure-linux-swift-overlay.conflist) conflist to dropgz's build process so that it can be loaded onto a host without the conflist.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/dropgz/build/cni.Dockerfile
+++ b/dropgz/build/cni.Dockerfile
@@ -25,6 +25,7 @@ COPY dropgz .
 COPY --from=azure-ipam /azure-ipam/*.conflist pkg/embed/fs
 COPY --from=azure-ipam /azure-ipam/azure-ipam pkg/embed/fs
 COPY --from=azure-vnet /azure-container-networking/cni/azure-$OS-swift.conflist pkg/embed/fs/azure-swift.conflist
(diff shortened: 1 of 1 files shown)


rbtr (approved):
lgtm

rbtr (approved):

timraymond:
What is this file, and what does it do? I understand what `dropgz` does, but it would be helpful for reviewers (like me) without context to have a "here is the problem, here is why this proposed change solves this problem" answered in the description of the PR.

rbtr:
@timraymond the description already answers that, to my reading. Do you disagree? 
This being the repo for a CNI impl, I think we have to assume familiarity with CNI fundamentals like the function of the conflist in general.

rbtr (approved):

PR #1656, opened 2022-10-17
feat: expose register node api

ZetaoZhuang:
**Reason for Change**:

To enable AZR , dnc need to call CNS to create context selector through NMAgent. This PR is for exposing the register node api api in cns to achieve that.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -1478,3 +1478,57 @@ func (service *HTTPRestService) nmAgentSupportedApisHandler(w http.ResponseWrite
 
 	logger.Response(service.Name, nmAgentSupportedApisResponse, resp.ReturnCode, serviceErr)
 }
+
+func (service *HTTPRestService) registerNode(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] registerNode")
+	logger.Request(service.Name, "registerNode", nil)
+
+	var (
+		returnCode    types.ResponseCode
+		req           cns.RegisterNodeRequest
+		returnMessage string
+		err           error
+		resp          *http.Response
+	)
+
+	err = service.Listener.Decode(w, r, &req)
+	logger.Request(service.Name, &req, err)
+
+	if err != nil {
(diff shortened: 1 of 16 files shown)


cns/restserver/api.go
```diff
@@ -1506,3 +1507,56 @@ func (service *HTTPRestService) nmAgentSupportedApisHandler(w http.ResponseWrite
...
+				case http.StatusInternalServerError:
+					returnMessage = "[Azure CNS] Error. registerNode failed due to Nmagent server internal error."
+					returnCode = types.NmAgentServerInternalError
+				case http.StatusBadRequest:
+					returnMessage = "[Azure CNS] Error. registerNode failed due to bad request error."
+					returnCode = types.NmAgentBadRequestError
+				default:
+					returnCode = types.UnexpectedError
+					returnMessage = fmt.Sprintf("[Azure CNS] Error. registerNode failed with StatusCode: %d", apiError.StatusCode())
+				}
```
timraymond, line 1548: All of this information is encoded by the error itself. If you log the error, you will get something like:

```
nmagent: wireserver: http status 500: something terrible happened: body: a whole bunch of stuff
```

But it also might include issues with mime handling:

```
unexpected content type: application/xml: body: <some><weird><xml></xml></weird></some>
```

@rbtr Since these error codes are a public API, does adding these two make sense? Is there an existing one that's more appropriate? If not, would it make sense to add a code for the failure of a service dependency?

timraymond:
@ZetaoZhuang my work has been merged, so there will be some conflicts to resolve (and also you can remove some of the workarounds we had discussed). If you want to work through those conflicts, please let me know and we can set up some time to do so.

cns/restserver/api.go
ZetaoZhuang: @timraymond we briefly discussed this last week. any new idea on this? Thanks

PR #1667, opened 2022-10-20
fix: unblock channel for cni

camrynl:
**Reason for Change**:

CNI blocks when sending failure report because CNS exits and is not receiving.
Unblocking channel with select/default clause.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/telemetry/telemetrybuffer.go
+++ b/telemetry/telemetrybuffer.go
@@ -216,12 +216,7 @@ func (tb *TelemetryBuffer) Write(b []byte) (c int, err error) {
 
 // Cancel - signal to tear down telemetry buffer
 func (tb *TelemetryBuffer) Cancel() {
-	select {
-	case tb.cancel <- true:
-		log.Logf("server cancel")
-	default:
-		log.Logf("Cancel fn: default case: no message sent")
-	}
+	tb.cancel <- true
 }
 
 // Close - close all connections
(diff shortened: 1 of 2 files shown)


rbtr (approved):

telemetry/telemetrybuffer.go
```diff
@@ -216,7 +216,12 @@ func (tb *TelemetryBuffer) Write(b []byte) (c int, err error) {
 
 // Cancel - signal to tear down telemetry buffer
 func (tb *TelemetryBuffer) Cancel() {
-	tb.cancel <- true
```
rbtr, line 219: I have not read through how the value sent on this channel is being used, so I don't know how messy it is. But the better pattern for this kind of binary "write once, read any" signal via a channel is to `close` the channel instead of sending a single item through it. Closing is non-blocking, and any reader waiting on the channel will immediately unblock and read a zero-value from the channel, and, if using two-value reads (`v, ok := <-tb.cancel`), a [`false` indicating that the channel is closed](https://go.dev/ref/spec#Receive_operator).
It's important that this be the only writer for the channel, because any subsequent attempt to send on the closed channel will panic.

tamilmani1989:
on a second thought, i think we should remove `tb.Cancel()` from `SendReport, SendCNIMetric, SendCNIEvent` from `telemetry/telemetry.go`. I see no value with that call and its invoked only on write failure and it doesn't close server socket since telemetrybuffer has its own copy

tamilmani1989 (approved):

rbtr (approved):

PR #1668, opened 2022-10-20
fix: cns to use controller runtime (cached) clients in reconcilers

rbtr:
**Reason for Change**:

The scoped CRD clients did not previously accept a controller-runtime client as an input, and instead constructed one from the kubeconfig. This creats direct (uncached) clients, which is okay for one-offs like during initialization, but not for ongoing high frequency usage like in the Reconcilers.

This change modifies the CRD client constructors to accept a client to wrap, which allows embedding the (cached) client provided by the ctrlruntime Manager. This takes full advantage of the sharedinformer pattern, improves efficiency of the reconcilers, and reduces apiserver load.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**: I migrated these components away from the Manager's client as part of switching out the scoped clients 😢 that was a mistake, and this is how they should have been from the beginning.

### DIFF
--- a/cns/kubecontroller/clustersubnetstate/reconciler.go
+++ b/cns/kubecontroller/clustersubnetstate/reconciler.go
@@ -3,6 +3,7 @@ package clustersubnetstate
 import (
 	"context"
 
+	"github.com/Azure/azure-container-networking/crd/clustersubnetstate"
 	"github.com/Azure/azure-container-networking/crd/clustersubnetstate/api/v1alpha1"
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


thatmattlong (dismissed):

camrynl (approved):

timraymond (dismissed):
I don't know that I completely follow everything here, but I've got enough of an understanding to approve it (especially given prior approvals). The Go looks good, and I learned something about SharedInformer in the process.

thatmattlong (approved):

PR #1670, opened 2022-10-21
fix: add phony target to crd gen makefiles to ignore path on disk

rbtr:

**Reason for Change**: Adds .PHONY targets (back?) to the CRD gen makefiles. 
`make` by default interprets a target name as being the desired file output to the disk. If the target (`manifests`) exists as a path on disk (`./manifests/`), the target will not execute.
Adding `.PHONY` for a target indicates that the output/target name is not a file, and the target will execute regardless of the file status.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


timraymond (approved):
Seems reasonable to me!

rbtr:
closed in #1776

PR #1673, opened 2022-10-24
Dualstack cilium proposal

rjdenney:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,23 @@
+# Cilium Azure-ipam Dualstack Overlay Proposal 
+
+## Purpose 
+
+This document proposes changes to the azure-ipam contract so that we can work with dualstack clusters between Cilium and Azure CNS. 
+
+## Overview 
+
+With the current CNS solution for dualstack we are setting the nnc to pass multiple NCs with CIDR to be unrolled and added to a pool that contains both v6 and v4 ips. Since we are planning to use Cilium CNI for the linux dualstack solution we need to make changes to azure-ipam to allow for a dualstack cluster to get multiple IPs. Currently azure-Ipam handles the [`CmdAdd`](https://github.com/Azure/azure-container-networking/blob/master/azure-ipam/ipam.go) command from Cilium CNI and sends an IP request to the CNS. The CNS then in overlay then looks through the pool and picks the first IP found that isn’t used. For dualstack we need to change two things: 1) Azure-ipam needs to be able to assign multiple IPs and 2) CNS needs to be able to determine what type of IP it is returning so that we can have one IPv4 and one IPv6.
```
rbtr, line 9: format this list:

>... things:
>1) Azure-ipam needs to be able to assign multiple IPs and 
>2) CNS needs to be able to determine what type of IP it is returning so that we can have one IPv4 and one IPv6.

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Proposal 
+
+For this issue two solutions are being proposed.  
+
+The first is having azure-ipam send two separate requests for both an IPv4 and an IPv6. The add command would need to be able to recognize that the pod is dualstack, create two [`IPConfigRequest`](https://github.com/Azure/azure-container-networking/blob/master/cns/NetworkContainerContract.go)s with an added type field to specify either v4 or v6 and call [`RequestIPAddress`](https://github.com/Azure/azure-container-networking/blob/master/cns/client/client.go) twice (once for each ip type). These two IPs would then both be returned in separate `IPConfigResponse` objects to the azure-ipam and added to the array of IPs in `cniResult` which then get returned to the Cilium CNI.
```
rbtr, line 15: include the proposed new IPConfigRequest in a codeblock

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,23 @@
...
+
+The second solution is to send one request from azure-ipam and have it specify in its `IPConfigRequest` that it wants a dualstack pod. With solution instead of calling `RequestIPAddress` twice to request two IPs it would only call the CNS once and would return both IPs requested in an array of `PodIpInfo`s in the `IPConfigResponse`. This would most likely require a much more extensive code change as requesting an IP would now return an array of IPs instead of just the one IP. The change would have to make the default situation expect that only one IP should be returned in the array unless we know that the pod is dualstack (or possible future reasons for returning multiple IPs).
```
rbtr, line 17: include the proposed new IPConfigRequest and IPConfigResponse specs here in codeblocks

rbtr: >The change would have to make the default situation expect that only one IP

Why should the client care? It makes an IPs requests and gets IPs. It's irrelevant that it may only get 1; 1 is a valid number of IPs.

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,23 @@
...
+
+The second solution is to send one request from azure-ipam and have it specify in its `IPConfigRequest` that it wants a dualstack pod. With solution instead of calling `RequestIPAddress` twice to request two IPs it would only call the CNS once and would return both IPs requested in an array of `PodIpInfo`s in the `IPConfigResponse`. This would most likely require a much more extensive code change as requesting an IP would now return an array of IPs instead of just the one IP. The change would have to make the default situation expect that only one IP should be returned in the array unless we know that the pod is dualstack (or possible future reasons for returning multiple IPs). 
+
+For both of these proposals there would also need to be some changes to how IPs are assigned in the CNS. When we go through the pool to find an available IP in [`AssignAnyAvailableIPConfig`](https://github.com/Azure/azure-container-networking/blob/master/cns/restserver/ipam.go) a type check would need to be added to see if it is either v4 or v6 and the type that we want to check for can be passed into the function.This type check
```
rbtr, line 19: >...function.This type check

?

rbtr, line 19: > a type check would need to be added to see if it is either v4 or v6

This needs to be generic. The general form of this problem would be that we have a heterogeneous pool of IP and want to assign IP(s) based on some filterable properties of the IPs in the pool. v4, v6, subnetA, subnetB, <any other properties>...

docs/feature/dualstack/Cilium-dualstack.md
rbtr: what if we get an error on one IP request?

rbtr (commented):
Needs a twice-over to improve formatting and language, needs to introduce the problem and involved components, etc. This is documentation of not just the _what_ (and needs more of that), but the _why_, so that when we revisit this in the future, we will still be able to understand why we made the decisions we made.

docs/feature/dualstack/Cilium-dualstack.md
rjdenney: added formatting

docs/feature/dualstack/Cilium-dualstack.md
rjdenney: > what if we get an error on one IP request?

This is something I was thinking about too and why I added it to further questions. Right now when just requesting one IP at a time and it fails we print an error (Failed to request IP address from CNS) and return. I would imagine if we retrieved one but fail the other than we send out a similar error and release the IP that we were able to get.

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,55 @@
...
+The second solution is to send one request from azure-ipam and have it specify in its `IPConfigRequest` that it wants a dualstack pod. With solution instead of calling CNS twice to request two IPs it would only call the CNS once and would return both IPs requested in an array. This would most likely require a much more extensive code change as requesting an IP would now return an array of IPs instead of just the one IP.  
+
+```diff
+type IPConfigRequest struct {
+    DesiredIPAddress    string
+    PodInterfaceID      string
+    InfraContainerID    string
+    OrchestratorContext json.RawMessage
+    Ifname              string // Used by delegated IPAM
++   PodType             string // would designate that a pod is 'dualstack' and search for two IPs instead of just one
```
ashvindeodhar, line 39: IMO we shouldn't design a contract that takes the podType as dualStack. If we end up going this route, it should be generic - which allows asking for N IPs with each IP of a specifc type. So you should be able to ask for 5 IPs, 5 ipv4s and 2 ipv6s.

ashvindeodhar: Going with Evan's suggestion below, this should be named something generic like 'IPConstraint' which can specify if a specific type is needed, if a specific subnet is needed, etc.

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Purpose 
+
+This document proposes changes to the azure-ipam contract so that we can work with dualstack clusters between Cilium and Azure CNS. 
+
+## Overview 
+
+With the current CNS solution for dualstack we are setting the nnc to pass multiple NCs with CIDR to be unrolled and added to a pool that contains both v6 and v4 ips. Since we are planning to use Cilium CNI for the linux dualstack solution we need to make changes to azure-ipam to allow for a dualstack cluster to get multiple IPs. Currently azure-Ipam handles the [`CmdAdd`](https://github.com/Azure/azure-container-networking/blob/master/azure-ipam/ipam.go) command from Cilium CNI and sends an IP request to the CNS. The CNS hen in overlay then looks through the pool and picks the first IP found that isn’t used. For dualstack we need to change two things:
+1. Azure-ipam needs to be able to assign multiple IPs
+2. CNS needs to be able to determine what type of IP it is returning so that we can have one IPv4 and one IPv6.
```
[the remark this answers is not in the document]
rjdenney: This is true we need to be able to pass into the AssignAnyAvailableIPConfig function which type of IP that we want

docs/feature/dualstack/Cilium-dualstack.md
rjdenney: I agree this would work better. Do you think it should it be its own type or just a string?

rbtr (changes requested):
this PR should stay scoped to proposal docs. impl should be separate (and should probably wait on this to solidify/get merged at least)

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,67 @@
+# Cilium Azure-ipam Dual Stack Overlay Proposal 
+
+## Purpose 
+
+This document proposes changes to the azure-ipam contract so that we can work with dual stack clusters between Cilium and Azure CNS.
```
rbtr, line 5:
```suggestion
This document proposes changes to the azure-cns contract so that we can work with dual stack clusters. 
```
the proposed changes are to the CNS API and only involve azure-ipam as a side-effect as a consumer of that API

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,67 @@
+# Cilium Azure-ipam Dual Stack Overlay Proposal 
+
+## Purpose 
+
+This document proposes changes to the azure-ipam contract so that we can work with dual stack clusters between Cilium and Azure CNS. 
+
+## Overview 
+
+With the current CNS solution for dual stack we are setting the nnc to pass multiple NCs with CIDR to be unrolled and added to a pool that contains both v6 and v4 ips. Since we are planning to use Cilium CNI for the linux dual stack solution we need to make changes to azure-ipam to allow for a dual stack cluster to get multiple IPs. Currently azure-Ipam handles the [`CmdAdd`](https://github.com/Azure/azure-container-networking/blob/master/azure-ipam/ipam.go#:~:text=func%20(p%20*IPAMPlugin)%20CmdAdd(args%20*cniSkel.CmdArgs)%20error%20%7B) command from Cilium CNI and sends an IP request to the CNS. The CNS then in overlay then looks through the pool and picks the first IP found that isn’t used. For dual stack we need to change two things:
```
rbtr, line 9:
```suggestion
With the current CNS solution for dualstack we are setting the NNC to pass multiple NCs with CIDR to be unrolled and added to a pool that contains both v6 and v4 IPs. Since we are planning to use Cilium CNI for the Linux dualstack solution we need to make changes to the CNS API to allow for a dualstack cluster to get multiple IPs. Currently azure-ipam handles the [`CmdAdd`](https://github.com/Azure/azure-container-networking/blob/master/azure-ipam/ipam.go#:~:text=func%20(p%20*IPAMPlugin)%20CmdAdd(args%20*cniSkel.CmdArgs)%20error%20%7B) command from Cilium CNI and sends an IP request to CNS. In Overlay, CNS then looks through the pool and picks the first IP found that isn’t used. For dualstack we need to change two things:
```

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,67 @@
...
+1. Azure-ipam needs to be able to assign multiple IPs
```
rbtr, line 10:
```suggestion
1. azure-ipam needs to be able to assign multiple IPs
```

rbtr (changes requested):
I have reviewed up to the last comment and stopped - please go over this critically and make sure that it is polished. It is public-facing documentation and should read as such.

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,67 @@
...
+}
+```
+
+```diff
++type IPMode string
++
++const (
++	SingleStack Status = "SingleStack"
++	DualStack   Status = "DualStack"
++)
```
rbtr, line 40: I don't think that this is the concept that we previously discussed and agreed on - I want the NC assignment mode to be more generic and composable. You are trying to address the case of "one v4 NC, one v6 NC, and assign an IP from both", but we could more fundamentally describe that as "two NCs, assign an IP from both". Then, what if we have more than two NCs? This is not different from "multiple NCs, assign an IP from each". What about the scenario of all NCs are v4, or all are v6? What if I have several v4 and one v6?
In other words, you are addressing the _specific case_ of v4+v6, but you should be thinking about how to solve the general problem of CNS having multiple NCs and needing to assign IPs out of one or more of them in certain combinations (and hardcoding the behavior of each option is the unmaintainable way to do it). The solution you propose can still be to the specific case you are interested in, but should be open-ended enough that it could be enhanced to the more general cases without having to throw it away and re-solve the whole problem again.

To get this started, we should initially change the option name to `IPAssignmentMode` or `NCAssignmentMode`, and think about how we tell CNS _dynamically, at runtime_ which NC(s) it should be assigning an IP out of to satisfy any individual IP request.

ashvindeodhar: Hey @rbtr, the last conversation us three had, we talked about having a single field similar to what Ryan proposed here to let CNS know that it needs to operate in the dual stack mode. Maybe there was a misunderstanding on takeaways from that discussion. Having said that, I see your point that about adding a field that couples closely to a scenario. I don't believe we want to design something for the scenarios which are not existent and don't know what the usage will be. e.g. having multiple NCs in NNCs - can there be mixed mode supported where few NCs will be dynamic, few static? In this case, CNS would not want to give IPs from both the NCs as suggested by NNC's IPAssignmentMode NC list. Will there be a valid scenario where there would be say 5 NCs under NNC while CNS is supposed to use only 3 of those to generate IP pool? etc. There are unknowns and IMO we should design for those when we actually have valid scenarios.
In short, I am proposing to stick to the current behavior of CNS to use all the NCs under NNC to populate the IP pool and not add any flag to the NNC.
Proposal -
1. CNS learns IPAssignmentMode based on IPFamily and mode (static/dynamic) from the NCs under NNC. If CNS sees ipv4 and ipv6 static mode NCs under NNC, it will be in dualstack IPAssignmentMode. In this mode, it will validate that there are exactly 2 NCs under NNC.
2. When in dualstack mode, CNS will give out one ipv4 and one ipv6 IP.
3. There won't be any change to ipv4 overlay for now but this can be extended to it as well.

Let me know your thoughts.

rbtr: If we want to go that direction, I suggest we commit to it fully: we can remove any intelligence related to IPv4/v6 and make it so that CNS assigns one IP out of each NC for each IP Request. 
This is the simplest solution to the current problem and requires the least rework in the future if other NC scenarios come up, because we won't have to _remove_ a bunch of logic around how we assign IPs based on the IP families of the NCs.

ashvindeodhar: We don't know what future NC scenarios will come up. If there is a scenario of mix mode, then we still may not want to assign one IP out of each NC for each IP request. So if we go with this route we may still need to remove / update logic. We just don't know at this point :) 
I am ok with giving one IP out of each NC and removing any intelligence about ip familiy. Without digging into the CNS code, do you know if that will affect the current logic of v4 overlay? I'll try to take a look at it

rbtr: No existing singletenancy scenario has multiple NCs so they should be fine

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,67 @@
...
+    Ifname              string // Used by delegated IPAM
+}
+```
+
+```diff
+type IPConfigResponse struct {
+-   PodIpInfo PodIpInfo 
++   PodIpInfo []PodIpInfo 
+    Response  Response
+}
```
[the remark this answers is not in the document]
thatmattlong: actually new type is probably not required, the new api can just return the slice as top-level lists are still valid json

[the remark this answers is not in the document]
rjdenney: When you say that I wouldn't need a new type are you suggesting that I just return a slice of the IPConfigResponse instead of the PodIPInfo within it?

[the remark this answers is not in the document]
ashvindeodhar: @thatmattlong IMO it would be better to create a new type so that there will be a single 'Response' indicating if the client should consume the response. Can't do that if we return slice of IPConfigResponse

rjdenney:
@thatmattlong and @rbtr, I have updated the doc to reflect more of an API and would appreciate if you guys could give me an comments or a thumbs up if you think it is good. I am also going to split up my changes into multiple PRs so that I can push the CNS changes with the new contract first and then push up the azure-ipam changes after.

rbtr (dismissed):
"approved with suggestions"

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,74 @@
...
++// IPConfigsResponse is used in CNS IPAM mode to return a slice of IP configs as a response to CNI ADD
++type IPConfigsResponse struct {
++	PodIpInfo []PodIpInfo
++	Response  Response
++}
+```
+
+### New API
+
+A new API will called `RequestAllIPAddresses` be created for requesting IPs when running in Overlay. This will work similar to the current [`RequestIPAddress`](https://github.com/Azure/azure-container-networking/blob/master/cns/client/client.go) and will accept the same [`IPConfigRequest`](https://github.com/Azure/azure-container-networking/blob/master/cns/NetworkContainerContract.go), the main difference will be that it will return the new contract type `IPConfigsResponse` with the slice of `PodIpInfo`.
```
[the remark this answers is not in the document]
rjdenney: I agree that `RequestIPs` would be a better naming convention and will shift some of the names to be more consistent with `IPConfigsResponse`

rbtr:
@ashvindeodhar I'm sad that this design proposal PR is still open, but there are already a half dozen implementation PRs in-flight.
cc @thatmattlong @rjdenney

ashvindeodhar:
> @ashvindeodhar I'm sad that this design proposal PR is still open, but there are already a half dozen implementation PRs in-flight. cc @thatmattlong @rjdenney

Hey Evan, I didn't realize this is not yet merged. I remember you approving it some time back so I thought this was merged with an ack from Matt too.
@thatmattlong Evan had approved this design and the PR that Ryan is creating adheres to the design. AFAIK you had looked thru the design and didn't have anything major. Do you mind taking a look at the design doc again? Ryan can merge the PR if things look ok to you.

rjdenney:
> @ashvindeodhar I'm sad that this design proposal PR is still open, but there are already a half dozen implementation PRs in-flight. cc @thatmattlong @rjdenney

@thatmattlong @rbtr @ashvindeodhar Just made the updates with the suggestions and bit more that I ended up changing in the implementation. Mostly just variable names and mentioned that I also updated GetExistingIPConfig and AssignDesiredIPConfig so that they can also work with multiple IPs. Should be ready to merge and the CNS side of changes I have up at https://github.com/Azure/azure-container-networking/pull/1773 . I also have a separate PR for the azure-ipam with the CNS changes here https://github.com/Azure/azure-container-networking/pull/1715 but we will need to push in CNS before azure-ipam goes in

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,69 @@
...
+// IPConfigResponse is used in CNS IPAM mode as a response to CNI ADD
+type IPConfigResponse struct {
+	PodIpInfo PodIpInfo
+	Response  Response
+}
+
++// IPConfigsResponse is used in CNS IPAM mode to return a slice of IP configs as a response to CNI ADD
++type IPConfigsResponse struct {
++	PodIpInfo []PodIpInfo
++	Response  Response
```
thatmattlong, line 34: I don't know what we use this `Response` type for, but if it exists in each `PodIpInfo`, do we also need one for the whole list?

thatmattlong (approved):

docs/feature/dualstack/Cilium-dualstack.md
rjdenney: From my understanding I believe the response just is a return code and a message for the `IPConfigsResponse` and isn't specific to each `PodIpInfo`. If for some reason we get an error trying to get IPs it is adds a `ReturnCode` and the error message to `Response`

docs/feature/dualstack/Cilium-dualstack.md
```diff
@@ -0,0 +1,69 @@
...
++}
+```
+
+### New API
+
+A new API will called `RequestIPs` be created for requesting IPs when running in Overlay. This will work similar to the current [`RequestIPAddress`](https://github.com/Azure/azure-container-networking/blob/master/cns/client/client.go) and will accept the same [`IPConfigRequest`](https://github.com/Azure/azure-container-networking/blob/master/cns/NetworkContainerContract.go), the main difference will be that it will return the new contract type `IPConfigsResponse` with the slice of `PodIpInfo`. 
+
+```diff
++// RequestIPs calls the RequestIPConfigs in CNS
++func (c *Client) RequestIPs(ctx context.Context, ipconfig cns.IPConfigRequest) (*cns.IPConfigsResponse, error)
```
tamilmani1989, line 44: will this return just ipv4 in an array when running in overlay singlestack ipv4?

docs/feature/dualstack/Cilium-dualstack.md
rbtr: @rjdenny you need to consider what happens if we get one IPConfig and then fail the next one. Is the first one marked as "Assigned" already? We would need to roll that back, and make the whole op transactional - entirely successful or entirely failed.

ashvindeodhar: @rbtr yes we had thought about it and have added a todo in the code as a reminder to address it before merging the change. We cannot have the partial success because in that case, we may mark IPs as 'Assigned' w/o the containers using those IPs. We cannot rely on the client to do the right thing of releasing the partially succeeded requests. We may run out of the IPs if we don't handle this ourselves.

docs/feature/dualstack/Cilium-dualstack.md
ashvindeodhar: Ryan is making changes in azure-ipam and CNS. 
Paul Yu is making changes in azure-cni. So ipv4 overlay will still keep using the old API until new changes made in azure-cni get deployed which is when it will also start using slice of IPs. It will have single result in the slice.

ashvindeodhar:
@rjdenney can this be merged?

rjdenney:
> @rjdenney can this be merged?

Yeah this should be able to be merged it. It reflects how the CNS currently is 

> @rjdenney can this be merged?

It should be able to be merged. It we make more updates down the line I can add more. I may add some more based off of the changes I have been making related to CNI but what it covers now is accurate.

Also I have made the request IPs backwards compatible as mentioned in the doc but I am starting to think I may need to reintroduce the old ReleaseIPAddress function as well. I ended up having to update that function to allow us to have multiple IPs in the `DesiredIPAddress` field in the contract. This is needed so that when CNI makes a DELETE command on a pod with two IPs IPAM knows to put two back in the pool.

rjdenney:
> @rjdenney can this be merged?

yes

rbtr:
go on then

PR #1680, opened 2022-10-26
feat: allow the user to pass an aikey in config

thatmattlong:
**Reason for Change**:
Allows the user of CNS to specify an appinsights instrumentation key in config so we can split out the telemetry backends to address some of our scale issues.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
no unit tests since there are already no unit tests for the cnslogger and this is a trivial change to grab a string from config and pass it to the Init function

### DIFF
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -63,6 +63,8 @@ type TelemetrySettings struct {
 	DebugMode bool
 	// Interval for sending snapshot events.
 	SnapshotIntervalInMins int
+	// AppInsightsInstrumentationKey allows the user to override the default appinsights ikey
+	AppInsightsInstrumentationKey string
 }
 
 type ManagedSettings struct {
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -544,7 +544,11 @@ func main() {
 			DebugMode:                    ts.DebugMode,
 		}
 
-		logger.InitAI(aiConfig, ts.DisableTrace, ts.DisableMetric, ts.DisableEvent)
+		if aiKey := cnsconfig.TelemetrySettings.AppInsightsInstrumentationKey; aiKey != "" {
+			logger.InitAIWithIKey(aiConfig, aiKey, ts.DisableTrace, ts.DisableMetric, ts.DisableEvent)
+		} else {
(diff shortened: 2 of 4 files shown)


cns/configuration/configuration.go
```diff
@@ -63,6 +63,8 @@ type TelemetrySettings struct {
 	DebugMode bool
 	// Interval for sending snapshot events.
 	SnapshotIntervalInMins int
+	// AppInsightsInstrumentationKey allows the user to override the default appinsights ikey
```
timraymond, line 66: Typo?
```suggestion
	// AppInsightsInstrumentationKey allows the user to override the default appinsights key
```

timraymond (approved):

cns/configuration/configuration.go
thatmattlong: It was intentional though I can see why it seems that way. 😄  I've seen appinsights instrumentation keys abbreviated as "ikey" which is what I was going for here.

cns/service/main.go
```diff
@@ -544,7 +544,11 @@ func main() {
 			DebugMode:                    ts.DebugMode,
 		}
 
-		logger.InitAI(aiConfig, ts.DisableTrace, ts.DisableMetric, ts.DisableEvent)
+		if aiKey := cnsconfig.TelemetrySettings.AppInsightsInstrumentationKey; aiKey != "" {
+			logger.InitAIWithIKey(aiConfig, aiKey, ts.DisableTrace, ts.DisableMetric, ts.DisableEvent)
+		} else {
```
aegal, line 549: Just a NIT: might be better to set the default in setTelemetrySettingDefaults and not have to do this if/else scenario

thatmattlong: I tried your suggestion, but the problem I found was that the default value is defined as an unexported variable in cns/logger, so there was no good way for me to set the default in the `setTelemetrySettingDefaults` function becuase there is no way for me to read that value without at least exporting the variable. And the problem with _that_, is that we inject the value of that string at build-time using `go build`, and I'm not confident that I can enumerate all of the places where that value gets set, since we build CNS in a few different places in a few different pipelines. That's all to say - I think we'll do a big telemetry refactor soon once we have a path forward without AI, so I'm kind of ok with merging this as-is (especially if you feel the feedback is a nit).

PR #1682, opened 2022-10-27
fix: correct supportedApis path

ZetaoZhuang:
**Reason for Change**:

the supportedApis path in cns is wrong. PR to fix it

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


timraymond (approved):
Good catch, thanks :)

PR #1691, opened 2022-11-07
errors.Wrap automatically add colon and append error

ecigar13:
**Reason for Change**:
errors.Wrap() automatically appends colon and error. We don't need to add `: %w`. Here's the log:
`IPAM Invoker Add failed with error: Failed to get IP address from CNS with error: %w: http request failed: Post "`

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -81,7 +81,7 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 	response, err := invoker.cnsClient.RequestIPAddress(context.TODO(), ipconfig)
 	if err != nil {
 		log.Printf("Failed to get IP address from CNS with error %v, response: %v", err, response)
-		return IPAMAddResult{}, errors.Wrap(err, "Failed to get IP address from CNS with error")
+		return IPAMAddResult{}, errors.Wrap(err, "failed to get IP address from CNS")
 	}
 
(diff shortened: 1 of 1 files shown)


cni/network/invoker_cns.go
```diff
@@ -81,7 +81,7 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 	response, err := invoker.cnsClient.RequestIPAddress(context.TODO(), ipconfig)
 	if err != nil {
 		log.Printf("Failed to get IP address from CNS with error %v, response: %v", err, response)
-		return IPAMAddResult{}, errors.Wrap(err, "Failed to get IP address from CNS with error: %w")
+		return IPAMAddResult{}, errors.Wrap(err, "Failed to get IP address from CNS with error")
```
rbtr, line 84:
```suggestion
		return IPAMAddResult{}, errors.Wrap(err, "failed to get IP address from CNS")
```
since you're fixing this, error strings should also not be sentence case (and we can drop the "with error" because of course it is)

nairashu (dismissed):

rbtr (approved):

timraymond (approved):
Thanks for lower-casing "failed" too.

rbtr (dismissed):

rbtr (approved):

PR #1695, opened 2022-11-09
Fix inability to unmarshal nmagent request

### LINKED ISSUE #1694: Requests to /network/publishnetworkcontainer fail
Requests made to the `/network/publishnetworkcontainer` endpoint fail with the following error message:

```
Failed to decode *cns.PublishNetworkContainerRequest &{NetworkID:3ea285c0-2c61-47d0-bcc0-1d7529f978f7 NetworkContainerID:617a3122-1ef6-481c-b6dc-9851fef130b2 JoinNetworkURL:http://127.0.0.1:9001/machine/plugins/?comp=nmagent&type=NetworkManagement/joinedVirtualNetworks/3ea285c0-2c61-47d0-bcc0-1d7529f978f7/api-version/1 CreateNetworkContainerURL:http://127.0.0.1:9001/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/172.17.0.1/networkContainers/617a3122-1ef6-481c-b6dc-9851fef130b2/ CreateNetworkContainerRequestBody:{ID: VNetID: Version:0 SubnetName: IPv4Addrs:[] Policies:[] VlanID:0 GREKey:0 AuthenticationToken: PrimaryAddress:}} json: cannot unmarshal string into Go struct field PublishNetworkContainerRequest.CreateNetworkContainerRequestBody of type nmagent.PutNetworkContainerRequest.
```

This is a regression caused by a change to the requests expected by CNS from clients. This violates the de-facto API contract established by CNS.

timraymond:
During the previous switchover to using the client from the `nmagent` client (as opposed to the `cns/nmagent` client), an assumption was made that "proxied" requests to NMAgent were provided by clients as nested JSON. This assumption was wrong--they are Base64-encoded strings of JSON. Even though they're ultimately similar, they're very different from the perspective of the JSON unmarshaler. Consequently, this restores the nested request body back to a []byte and performs the second-stage decoding manually (similarly to how it was previously done).

Fixes #1694

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -5,6 +5,7 @@ package restserver
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"net/http"
@@ -18,6 +19,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/logger"
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/cns/wireserver"
+	"github.com/Azure/azure-container-networking/nmagent"
 	nma "github.com/Azure/azure-container-networking/nmagent"
 	"github.com/Azure/azure-container-networking/platform"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


rbtr:
> they are Base64-encoded strings of JSON

do you mean base64? these look like plaintext in the error message and in your test changes. UTF-8 or similar, maybe?

timraymond:
@rbtr `[]byte` ends up getting encoded that way by `json.Marshal`. It's certainly possible to write the tests using "e30=" for that value... but it doesn't help readability much. You can see this happening here: https://go.dev/play/p/9_mvDSnT0MT

cns/restserver/api.go
```diff
@@ -1188,11 +1190,20 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
...
+			innerReqBytes := req.CreateNetworkContainerRequestBody
 
-			err = service.nma.PutNetworkContainer(ctx, &pncr)
+			var innerReq nmagent.PutNetworkContainerRequest
+			err = json.Unmarshal(innerReqBytes, &innerReq)
+			if err != nil {
+				returnMessage = fmt.Sprintf("Failed to publish Network Container: %s", req.NetworkContainerID)
+				returnCode = types.NetworkContainerPublishFailed
+				logger.Errorf("[Azure-CNS] %s", returnMessage)
+			}
```
ramiro-gamarra, line 1201: Is this the previous version of the code verbatim? It seems odd that there is no return after this error (the below might blow up). Since it seems like there's another change to be made still to pass linting, would a `return` be appropriate here? If there could be unintended consequences and this is the code as it always had been, then disregard.

timraymond: Fixed in [b866e39](https://github.com/Azure/azure-container-networking/pull/1695/commits/b866e39136e21be893df5c57bb8127a69ae7200a)

rbtr (approved):

ramiro-gamarra (approved):

rbtr:
bypassing as relevant tests have passed

pjohnst5:
Started a fix for this here: #1740 but need Tim's help to understand why `version` can't be `0`

PR #1702, opened 2022-11-18
feat: cns writes cni conflist

thatmattlong:
**Reason for Change**:

This PR will allow CNS to write the CNI conflist (right now for Azure CNI Overlay only), so it can keep the Node in the NotReady state until the network is ready.

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cni/netconfig.go
+++ b/cni/netconfig.go
@@ -40,43 +40,45 @@ type RuntimeDNSConfig struct {
 	Options  []string `json:"options,omitempty"`
 }
 
+type IPAM struct {
+	Mode          string `json:"mode,omitempty"`
+	Type          string `json:"type"`
+	Environment   string `json:"environment,omitempty"`
+	AddrSpace     string `json:"addressSpace,omitempty"`
+	Subnet        string `json:"subnet,omitempty"`
+	Address       string `json:"ipAddress,omitempty"`
+	QueryInterval string `json:"queryInterval,omitempty"`
+}
+
 // NetworkConfig represents Azure CNI plugin network configuration.
 type NetworkConfig struct {
-	CNIVersion                    string   `json:"cniVersion,omitempty"`
-	Name                          string   `json:"name,omitempty"`
-	Type                          string   `json:"type,omitempty"`
-	Mode                          string   `json:"mode,omitempty"`
-	Master                        string   `json:"master,omitempty"`
-	AdapterName                   string   `json:"adapterName,omitempty"`
-	Bridge                        string   `json:"bridge,omitempty"`
-	LogLevel                      string   `json:"logLevel,omitempty"`
-	LogTarget                     string   `json:"logTarget,omitempty"`
-	InfraVnetAddressSpace         string   `json:"infraVnetAddressSpace,omitempty"`
-	IPV6Mode                      string   `json:"ipv6Mode,omitempty"`
-	ServiceCidrs                  string   `json:"serviceCidrs,omitempty"`
-	VnetCidrs                     string   `json:"vnetCidrs,omitempty"`
-	PodNamespaceForDualNetwork    []string `json:"podNamespaceForDualNetwork,omitempty"`
-	IPsToRouteViaHost             []string `json:"ipsToRouteViaHost,omitempty"`
-	MultiTenancy                  bool     `json:"multiTenancy,omitempty"`
-	EnableSnatOnHost              bool     `json:"enableSnatOnHost,omitempty"`
-	EnableExactMatchForPodName    bool     `json:"enableExactMatchForPodName,omitempty"`
-	DisableHairpinOnHostInterface bool     `json:"disableHairpinOnHostInterface,omitempty"`
-	DisableIPTableLock            bool     `json:"disableIPTableLock,omitempty"`
-	CNSUrl                        string   `json:"cnsurl,omitempty"`
-	ExecutionMode                 string   `json:"executionMode,omitempty"`
-	Ipam                          struct {
-		Mode          string `json:"mode,omitempty"`
-		Type          string `json:"type"`
-		Environment   string `json:"environment,omitempty"`
-		AddrSpace     string `json:"addressSpace,omitempty"`
-		Subnet        string `json:"subnet,omitempty"`
-		Address       string `json:"ipAddress,omitempty"`
-		QueryInterval string `json:"queryInterval,omitempty"`
-	} `json:"ipam,omitempty"`
-	DNS             cniTypes.DNS    `json:"dns,omitempty"`
-	RuntimeConfig   RuntimeConfig   `json:"runtimeConfig,omitempty"`
-	WindowsSettings WindowsSettings `json:"windowsSettings,omitempty"`
-	AdditionalArgs  []KVPair        `json:"AdditionalArgs,omitempty"`
+	CNIVersion                    string          `json:"cniVersion,omitempty"`
+	Name                          string          `json:"name,omitempty"`
+	Type                          string          `json:"type,omitempty"`
+	Mode                          string          `json:"mode,omitempty"`
+	Master                        string          `json:"master,omitempty"`
+	AdapterName                   string          `json:"adapterName,omitempty"`
+	Bridge                        string          `json:"bridge,omitempty"`
+	LogLevel                      string          `json:"logLevel,omitempty"`
+	LogTarget                     string          `json:"logTarget,omitempty"`
+	InfraVnetAddressSpace         string          `json:"infraVnetAddressSpace,omitempty"`
+	IPV6Mode                      string          `json:"ipv6Mode,omitempty"`
+	ServiceCidrs                  string          `json:"serviceCidrs,omitempty"`
+	VnetCidrs                     string          `json:"vnetCidrs,omitempty"`
+	PodNamespaceForDualNetwork    []string        `json:"podNamespaceForDualNetwork,omitempty"`
+	IPsToRouteViaHost             []string        `json:"ipsToRouteViaHost,omitempty"`
+	MultiTenancy                  bool            `json:"multiTenancy,omitempty"`
+	EnableSnatOnHost              bool            `json:"enableSnatOnHost,omitempty"`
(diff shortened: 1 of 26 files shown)


cns/cniconflist/generator.go
```diff
@@ -0,0 +1,40 @@
...
+	DisableCheck bool          `json:"disableCheck,omitempty"`
+	Plugins      []interface{} `json:"plugins,omitempty"`
+}
+
+const (
+	cniVersion     = "0.3.0"
+	cniName        = "azure"
+	cniType        = "azure-vnet"
+	nodeLocalDNSIP = "169.254.20.10"
+)
```
timraymond, line 40: Nit: const blocks at the top are usually the tradition.

cns/cniconflist/generator_windows.go
```diff
@@ -0,0 +1,13 @@
+package cniconflist
+
+import "errors"
+
+var errNotImplemented = errors.New("cni conflist generator not implemented on Windows")
```
timraymond, line 5: Out of curiosity, why? Seems like it just generates JSON... is it that we don't want that JSON to be consumed by anything on Windows?

cns/restserver/restserver.go
```diff
@@ -281,3 +306,19 @@ func (service *HTTPRestService) Stop() {
...
 	logger.Printf("[Azure CNS]  Service stopped.")
 }
+
+// MustGenerateCNIConflistOnce will generate the CNI conflist once if the service was initialized with
+// a conflist generator. If not, this is a no-op.
+func (service *HTTPRestService) MustGenerateCNIConflistOnce() {
+	service.generateCNIConflistOnce.Do(func() {
+		if err := service.cniConflistGenerator.Generate(); err != nil {
+			logger.Errorf("unable to generate cni conflist with error: %v", err)
+			os.Exit(1)
```
timraymond, line 316: Not crazy about `os.Exit` appearing anywhere other than `main()` because it prevents `defers` from running elsewhere. Can we return here instead? Alternatively, `runtime.Goexit()` is slightly better, since it's like a goroutine-specific `os.Exit` (it's how `testing.(*T).Fatal` works under the hood).

It's a bit tricky to see the forest through the trees on the diff though, so if it's a pervasive thing, don't let it block the PR.

cni/netconfig.go
```diff
@@ -40,43 +40,45 @@ type RuntimeDNSConfig struct {
 	Options  []string `json:"options,omitempty"`
 }
 
+type Ipam struct {
```
timraymond, line 43: That looks much better pulled out of `NetworkConfig`

cns/cniconflist/generator_linux_test.go
```diff
@@ -0,0 +1,77 @@
...
+	for _, test := range generatorTests {
+		tt := test
+		t.Run(fmt.Sprintf("generate-%s", tt.s), func(t *testing.T) {
+			buffer := new(bytes.Buffer)
+			g, err := cniconflist.NewGenerator(&bufferWriteCloser{buffer}, tt.s)
+			if err != nil {
+				if tt.expectedNewGeneratorErr {
+					return
+				}
+				t.Fail()
```
timraymond, line 49: I'd combine these, just for readability, and editorialize the error a little:

```go
if err != nil && !tt.expectedNewGeneratorErr {
       t.Fatal("unexpected error creating generator: err:", err)
}

if err == nil && tt.expectedNewGeneratorErr {
       t.Fatal("exepcted an error creating generator but received none")
}
```

This reads, "if the error was present and we didn't expect one, fail" and "if there was no error, but we expected one, fail". `t.Fatal`, I assume, (equivalent of `t.Log` and `t.FailNow`) because it's nonsensical to make any assertions after that point in either circumstance.

cns/cniconflist/generator_linux_test.go
```diff
@@ -0,0 +1,77 @@
...
+			err = g.Generate()
+			assert.NoError(t, err)
+
+			// by unmarshalling the json and comparing the structs, we're only checking
+			// for data equality and not really diffing the files, for example
+			// the fixture could have spaces and the generated conflist could have tabs,
+			// but this is fine since all we really care about is the data equality
+			fixture, err := os.ReadFile(tt.fixture)
+			if err != nil {
+				t.Fail()
```
timraymond, line 61: I'd recomment using `t.Error` / `t.Fatal`, or their respective `require` / `assert` equivalents to make sure the failure messages are useful.

cns/cniconflist/generator.go
```diff
@@ -0,0 +1,40 @@
...
+
+type Generator struct {
+	writer   io.WriteCloser
+	scenario configuration.CNIConflistScenario
+}
+
+func (g *Generator) Close() error {
+	if err := g.writer.Close(); err != nil {
+		return fmt.Errorf("error closing generator: %w", err)
```
timraymond, line 17: `errors.Wrap` is used in CNS, typically.

cns/cniconflist/generator.go
```diff
@@ -0,0 +1,40 @@
...
+// cniConflist represents the containernetworking/cni/pkg/types.NetConfList
+//
+// TODO: should use the containernetworking/cni/pkg/types.NetConfList, but our CNI config
+// doesn't embed the NetConf struct, so we need to refactor that before we can use the
+// upstream containernetworking types
+type CNIConflist struct {
+	CNIVersion   string        `json:"cniVersion,omitempty"`
+	Name         string        `json:"name,omitempty"`
+	DisableCheck bool          `json:"disableCheck,omitempty"`
+	Plugins      []interface{} `json:"plugins,omitempty"`
```
timraymond, line 32: Not necessary for this PR, but I've been trying to force myself to use `any` over `interface{}` now that we can. It's a hard habit to break.

cns/cniconflist/generator_linux.go
```diff
@@ -0,0 +1,75 @@
...
+		Type         string          `json:"type"`
+		Capabilities map[string]bool `json:"capabilities"`
+		SNAT         bool            `json:"snat"`
+	}{
+		Type: portmapType,
+		Capabilities: map[string]bool{
+			portMappingsCapability: true,
+		},
+		SNAT: true,
+	}
```
timraymond, line 60: Is it worth reifying portmap into its own type where it's pretty common in CNI configs?

timraymond (approved):
Looks pretty good to me. I have some suggestions, but no showstoppers that I saw.

cns/cniconflist/generator_windows.go
```diff
@@ -0,0 +1,13 @@
...
+
+func NewGenerator(io.Writer, configuration.CNIConflistScenario) (*Generator, error) {
```
rbtr, line 7: missing these imports

rbtr (commented):
it feels like there may be a slightly better way to define the abstractions/implementations, but already looks good to me pending windows import/build fix

cns/cniconflist/generator.go
```diff
@@ -0,0 +1,40 @@
+package cniconflist
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/Azure/azure-container-networking/cns/configuration"
```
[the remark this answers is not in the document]
timraymond: I like this too, it makes `cniconflist` more modular and reduces coupling. You can probably even directly assign the CNS config to the `conflist` config using the Go v1.8 trick: https://go.dev/play/p/eXB4jRzbuE0

cns/cniconflist/generator_linux.go
thatmattlong: the problem I ran into was that their conflist includes this `SNAT` bool that is not in the containernetworking/cni types, and IIRC they didn't export their config type. But I'll check on that part. If I can import their config type, that would be good.

cns/cniconflist/generator.go
thatmattlong: fixed

cns/restserver/restserver.go
thatmattlong: changed to `panic`

cni/netconfig.go
thatmattlong: ah missed that when I was moving things around, I'm usually a stickler for capitalizing acronyms 😆

cns/cniconflist/generator_linux_test.go
thatmattlong: agreed, changed to the suggestion

cns/cniconflist/generator_linux_test.go
thatmattlong: fixed to use assert

cns/cniconflist/generator.go
thatmattlong: changed to `errors.Wrap`

cns/cniconflist/generator.go
thatmattlong: done

cns/cniconflist/generator.go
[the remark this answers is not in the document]
thatmattlong: yep good suggestion, fixed

cns/cniconflist/generator_linux.go
```diff
@@ -0,0 +1,75 @@
...
+	default:
+		return nil, fmt.Errorf("can't build generator: %w", errUnsupportedScenario)
+	}
+
+	return &g, nil
+}
+
+func (g *Generator) Generate() error {
+	var azureVnetConfig cni.NetworkConfig
+	if g.scenario == configuration.CNIConflistScenarioV4Overlay {
```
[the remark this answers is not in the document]
thatmattlong: I thought about that but kind of expected the scenarios to share a lot of code, and basically just change a couple of lines like `executionMode` based on the scenario, so I thought sharing the code in a single generator made more sense. But it is a little bit harder to discern what actually gets generated since you kind of have to trace the code mentally to figure it out. I kind of accepted that tradeoff because we have the testdata/fixtures to see exactly what we expect to get generated, but I may refactor a bit to try our your suggestion and see which one I like better.

rbtr (approved):

PR #1707, opened 2022-11-29
Fix failure to start CNS when setting WireserverIP

timraymond:
There was insufficient coverage over cases involving different permutations of the "WireserverIP" configuration option. Consequently, there were instances where reasonable values for this option caused CNS to fail to start.

This moves the logic for transforming the CNS configuration into configuration suitable for the NMAgent client into a method off the CNSConfig. It also permits adding coverage over different scenarios that are likely to emerge.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -3,8 +3,12 @@ package configuration
 
 import (
 	"encoding/json"
+	"net"
+	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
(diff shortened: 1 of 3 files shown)


timraymond:
@paulyufan was the original reporter of this issue and tried out the changes made here. He reports:
paul

rbtr (approved):

PR #1712, opened 2022-12-05
fix: repair windows cni lock issue

AzureAhai:
Moving the lock from InitializeKeyValueStore() function to restore/save functions to improve cni performance on windows.

**Reason for Change**:
In Windows CNI We are currently lock access to statefile in the init phase which may cause some CNI performance issue. This PR will move the lock to the exact point when we read/write from/to statefile.

### DIFF
--- a/cni/network/plugin/main.go
+++ b/cni/network/plugin/main.go
@@ -225,9 +225,10 @@ func rootExecute() error {
 		// Start telemetry process if not already started. This should be done inside lock, otherwise multiple process
 		// end up creating/killing telemetry process results in undesired state.
 		tb = telemetry.NewTelemetryBuffer()
-		tb.ConnectToTelemetryService(telemetryNumRetries, telemetryWaitTimeInMilliseconds)
+		if err = tb.ConnectCNIToTelemetryService(telemetryNumRetries, telemetryWaitTimeInMilliseconds, netPlugin.Plugin); err != nil {
+			log.Errorf("connection to telemetry service failed.")
+		}
 		defer tb.Close()
-
 		netPlugin.SetCNIReport(cniReport, tb)
 
 		t := time.Now()
--- a/cni/plugin.go
+++ b/cni/plugin.go
@@ -174,12 +174,6 @@ func (plugin *Plugin) InitializeKeyValueStore(config *common.PluginConfig) error
 		}
 	}
 
-	// Acquire store lock.
-	if err := plugin.Store.Lock(store.DefaultLockTimeout); err != nil {
-		log.Printf("[cni] Failed to lock store: %v.", err)
-		return err
-	}
-
 	config.Store = plugin.Store
 
 	return nil
--- a/cni/plugin.go
+++ b/cni/plugin.go
@@ -174,6 +174,14 @@ func (plugin *Plugin) InitializeKeyValueStore(config *common.PluginConfig) error
 		}
 	}
 
+	// Acquiring store lock at this stage for optimization purpuses on Windows
+	if runtime.GOOS != "windows" {
+		// Acquire store lock.
+		if err := plugin.Store.Lock(store.DefaultLockTimeout); err != nil {
+			log.Printf("[cni] Failed to lock store: %v.", err)
+			return err
+		}
+	}
 	config.Store = plugin.Store
 
 	return nil
--- a/cni/plugin.go
+++ b/cni/plugin.go
@@ -201,3 +201,25 @@ func (plugin *Plugin) UninitializeKeyValueStore() error {
 
 	return nil
 }
+
+// Lock key-value store. This function is being used for locking access to TelemetryService start for Windows Runtime.
+func (plugin *Plugin) LockKeyValueStore() error {
+	// Acquire store lock.
+	if err := plugin.Store.Lock(store.DefaultLockTimeout); err != nil {
+		log.Printf("[cni] Failed to lock store: %v.", err)
+		return errors.Wrap(err, "error Acquiring store lock")
+	}
+	return nil
+}
+
+// Unlock key-value store
+func (plugin *Plugin) UnLockKeyValueStore() error {
+	if plugin.Store != nil {
+		err := plugin.Store.Unlock()
+		if err != nil {
+			log.Printf("[cni] Failed to unlock store: %v.", err)
+			return errors.Wrap(err, "error unlock store lock")
+		}
+	}
+	return nil
+}
--- a/cni/plugin.go
+++ b/cni/plugin.go
@@ -174,12 +174,13 @@ func (plugin *Plugin) InitializeKeyValueStore(config *common.PluginConfig) error
 		}
 	}
 
-	// Acquiring store lock at this stage for optimization purpuses on Windows
+	// Acquiring store lock at this stage only for Linux. For optimization purpuses on Windows, the store lock will take place at later stage.
+	// TODO: The Linux store locking should be removed from here as well after some more validation.
 	if runtime.GOOS != "windows" {
 		// Acquire store lock.
 		if err := plugin.Store.Lock(store.DefaultLockTimeout); err != nil {
 			log.Printf("[cni] Failed to lock store: %v.", err)
-			return err
+			return errors.Wrap(err, "error Acquiring store lock")
 		}
 	}
 	config.Store = plugin.Store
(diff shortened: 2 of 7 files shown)


network/manager.go
```diff
@@ -115,8 +115,22 @@ func (nm *networkManager) Initialize(config *common.PluginConfig, isRehydrationR
 func (nm *networkManager) Uninitialize() {
 }
 
+//Remove the lock on the key-value store
+func (nm *networkManager) unlockStore() error {
```
tamilmani1989, line 119: why do we need separate function for unlock alone?

network/manager.go
```diff
@@ -115,8 +115,22 @@ func (nm *networkManager) Initialize(config *common.PluginConfig, isRehydrationR
...
+			return err
+		}
+	}
+	nm.store = nil
+	return nil
+}
+
 // Restore reads network manager state from persistent store.
 func (nm *networkManager) restore(isRehydrationRequired bool) error {
+	log.Printf("entering restore func...")
```
tamilmani1989, line 133: can it be `[net] Entering restore function`

network/manager.go
```diff
@@ -127,18 +141,26 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
 	// After a reboot, all address resources are implicitly released.
 	// Ignore the persisted state if it is older than the last reboot time.
 
+	// Acquire store lock.
+	if err := nm.store.Lock(store.DefaultLockTimeout); err != nil {
+		log.Printf("[cni] Failed to lock store: %v.", err)
+		return err
+	}
```
tamilmani1989, line 148: add defer function to unlock

network/manager.go
```diff
@@ -127,18 +141,26 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
...
 	// Read any persisted state.
 	err := nm.store.Read(storeKey, nm)
 	if err != nil {
 		if err == store.ErrKeyNotFound {
 			log.Printf("[net] network store key not found")
 			// Considered successful.
+			nm.unlockStore()
```
tamilmani1989, line 155: no need to have unlock in each if condition..defer should take care of

network/manager.go
```diff
@@ -225,12 +252,18 @@ func (nm *networkManager) save() error {
 	// Update time stamp.
 	nm.TimeStamp = time.Now()
 
+	// Acquire store lock.
+	if err := nm.store.Lock(store.DefaultLockTimeout); err != nil {
+		log.Printf("[cni] Failed to lock store: %v.", err)
+		return err
+	}
```
tamilmani1989, line 259: add defer for unlock and remove from other places

tamilmani1989 (changes requested):

network/manager.go
AzureAhai: I put this for my own debugging. Forgot to remove it. Will delete it in next commit. I need to fix the error checking for nm.unlockStore() too

network/manager.go
AzureAhai: I had to release lock in the restore function at multiple blocks, pretty much before every return().

network/manager.go
```diff
@@ -217,6 +233,7 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
 
 // Save writes network manager state to persistent store.
 func (nm *networkManager) save() error {
+	log.Printf("entering save func...")
```
tamilmani1989, line 236: remove this as well

network/manager.go
```diff
@@ -117,6 +117,7 @@ func (nm *networkManager) Uninitialize() {
 
 // Restore reads network manager state from persistent store.
 func (nm *networkManager) restore(isRehydrationRequired bool) error {
+	log.Printf("entering restore func...")
```
tamilmani1989, line 120: remove this.. you can resolve once you removed it..

tamilmani1989 (changes requested):
I'm also thinking..we should add these changes under a config field somethinng like "optimizeLock" so that we can release only to customers we want and also revert back easily for any mitigation

cni/plugin.go
```diff
@@ -174,12 +174,14 @@ func (plugin *Plugin) InitializeKeyValueStore(config *common.PluginConfig) error
 		}
 	}
 
-	// Acquire store lock.
-	if err := plugin.Store.Lock(store.DefaultLockTimeout); err != nil {
-		log.Printf("[cni] Failed to lock store: %v.", err)
-		return err
+	// Acquiring store lock at this stage for optimization purpuses on Windows
```
tamilmani1989, line 177: nit pic: spell check purposes and also add TODO, should remove this condition in future

ipam/manager.go
```diff
@@ -100,6 +101,22 @@ func (am *addressManager) restore(rehydrateIpamInfoOnReboot bool) error {
...
+	// Acquiring store lock at this stage for optimization purpuses on Windows
+	if runtime.GOOS == "windows" {
+		// Acquire store lock.
+		if err := am.store.Lock(store.DefaultLockTimeout); err != nil {
+			log.Printf("[ipam] Failed to lock store: %v.", err)
+			return err
+		}
+		// Remove the lock on the key-value store
+		defer func() {
+			err := am.store.Unlock()
```
tamilmani1989, line 113: same condition for unlock?  unlock only if not windows?

ipam/manager.go
```diff
@@ -169,6 +186,22 @@ func (am *addressManager) save() error {
...
```
tamilmani1989, line 198: add same condition to unlock

network/manager.go
```diff
@@ -127,6 +128,22 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
...
+	// Acquiring store lock at this stage for optimization purpuses on Windows
+	if runtime.GOOS == "windows" {
+		// Acquire store lock.
+		if err := nm.store.Lock(store.DefaultLockTimeout); err != nil {
+			log.Printf("[cni] Failed to lock store: %v.", err)
+			return err
+		}
+		// Remove the lock on the key-value store
+		defer func() {
+			err := nm.store.Unlock()
```
tamilmani1989, line 140: same as above

network/manager.go
```diff
@@ -225,6 +241,22 @@ func (nm *networkManager) save() error {
...
```
tamilmani1989, line 253: same as above

tamilmani1989 (changes requested):

cni/network/plugin/main.go
```diff
@@ -224,10 +224,15 @@ func rootExecute() error {
 
 		// Start telemetry process if not already started. This should be done inside lock, otherwise multiple process
 		// end up creating/killing telemetry process results in undesired state.
+		if err = netPlugin.Plugin.LockKeyValueStore(); err != nil {
```
tamilmani1989, line 227: lets not do it here..lets move it to a place where we start telemetry service otherwise this lock will be acquired for every cni invocation

tamilmani1989 (changes requested):

cni/plugin.go
```diff
@@ -198,3 +201,27 @@ func (plugin *Plugin) UninitializeKeyValueStore() error {
 
 	return nil
 }
+
+// Lock key-value store for Windows Runtime. This function is being used for locking access to TelemetryService start.
+func (plugin *Plugin) LockKeyValueStore() error {
+	if runtime.GOOS == "windows" {
```
tamilmani1989, line 207: can we move this condition to a place where we call LockKeyValueStore()

telemetry/telemetrybuffer.go
```diff
@@ -319,6 +321,34 @@ func (tb *TelemetryBuffer) ConnectToTelemetryService(telemetryNumRetries, teleme
...
+		if err := tb.Connect(); err != nil {
+			log.Logf("Connection to telemetry socket failed: %v", err)
+			if err = netPlugin.LockKeyValueStore(); err != nil {
+				log.Logf("lock acquire error: %v", err)
+				return errors.Wrap(err, "lock acquire error")
+			}
+			tb.Cleanup(FdName)
+			StartTelemetryService(path, args)
+			WaitForTelemetrySocket(telemetryNumRetries, time.Duration(telemetryWaitTimeInMilliseconds))
+			if err = netPlugin.UnLockKeyValueStore(); err != nil {
```
tamilmani1989, line 339: unlock only if windows

tamilmani1989 (changes requested):

cni/plugin.go
AzureAhai: addressed in the new commit

telemetry/telemetrybuffer.go
AzureAhai: addressed in the new commit

tamilmani1989 (approved):
lgtm

PR #1715, opened 2022-12-10
Dualstack CNS and azure-ipam changes

rjdenney:
**Reason for Change**:

This PR is to address changes in CNS and the azure-ipam to allow for a dualstack implementation. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Currently still testing the azure-ipam/cilium side of changes

### DIFF
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -9,6 +9,7 @@ import (
 	"github.com/Azure/azure-container-networking/azure-ipam/internal/buildinfo"
 	"github.com/Azure/azure-container-networking/azure-ipam/ipconfig"
 	"github.com/Azure/azure-container-networking/cns"
+	cnscli "github.com/Azure/azure-container-networking/cns/client"
 	cniSkel "github.com/containernetworking/cni/pkg/skel"
 	cniTypes "github.com/containernetworking/cni/pkg/types"
 	types100 "github.com/containernetworking/cni/pkg/types/100"
@@ -75,8 +76,27 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 	// https://www.cni.dev/docs/spec/#delegated-plugin-execution-procedure
 	resp, err := p.cnsClient.RequestIPs(context.TODO(), req)
 	if err != nil {
-		p.logger.Error("Failed to request IP address from CNS", zap.Error(err), zap.Any("request", req))
-		return cniTypes.NewError(ErrRequestIPConfigFromCNS, err.Error(), "failed to request IP address from CNS")
+		// if we fail a request with a 404 error try using the old API
+		p.logger.Error("Failed to request IPs using RequestIPs from CNS, going to try RequestIPAddress", zap.Error(err), zap.Any("request", req))
+		if errors.Is(err, cnscli.ErrAPINotFound) {
+			res, err := p.cnsClient.RequestIPAddress(context.TODO(), req)
+			
+			// if the old API fails as well then we just return the error
+			if err != nil {
+				p.logger.Error("Failed to request IP address from CNS using RequestIPAddress", zap.Error(err), zap.Any("request", req))
+				return cniTypes.NewError(ErrRequestIPConfigFromCNS, err.Error(), "failed to request IP address from CNS")
+			}
+			// takes values from the IPConfigResponse struct and puts them in a IPConfigsResponse struct
+			resp = &cns.IPConfigsResponse{
+				Response: res.Response,
+				PodIPInfo: []cns.PodIpInfo{
+					res.PodIpInfo,
+				},
+			}
+		} else {
+			p.logger.Error("Failed to request IP address from CNS", zap.Error(err), zap.Any("request", req))
+			return cniTypes.NewError(ErrRequestIPConfigFromCNS, err.Error(), "failed to request IP address from CNS")
+		}
 	}
 	p.logger.Debug("Received CNS IP config response", zap.Any("response", resp))
 
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -89,7 +89,11 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 		p.logger.Debug("Parsed pod IP", zap.String("podIPNet", IPNet.String()))
 	}
 
-	// ryand hardcoding until we can use labels
+	if len(*podIPNet) > 2 {
+		p.logger.Error("Returned an invalid number of IPs in IPConfigResponse", zap.Error(err), zap.Any("pod IPs", podIPNet))
+		return cniTypes.NewError(ErrProcessIPConfigResponse, err.Error(), "Returned an invalid number of IPs in IPConfigResponse")
+	}
+
 	cniResult := &types100.Result{
 		IPs: []*types100.IPConfig{
 			{
@@ -98,13 +102,26 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 					Mask: net.CIDRMask((*podIPNet)[0].Bits(), 32), // nolint
 				},
 			},
-			{
-				Address: net.IPNet{
-					IP:   net.ParseIP((*podIPNet)[1].Addr().String()),
-					Mask: net.CIDRMask((*podIPNet)[1].Bits(), 128), // nolint
+		},
+	}
+
+	if len(*podIPNet) == 2 { 
+		cniResult = &types100.Result{
+			IPs: []*types100.IPConfig{
+				{
+					Address: net.IPNet{
+						IP:   net.ParseIP((*podIPNet)[0].Addr().String()),
+						Mask: net.CIDRMask((*podIPNet)[0].Bits(), 32), // nolint
+					},
+				},
+				{
+					Address: net.IPNet{
+						IP:   net.ParseIP((*podIPNet)[1].Addr().String()),
+						Mask: net.CIDRMask((*podIPNet)[1].Bits(), 128), // nolint
+					},
 				},
 			},
-		},
+		}
 	}
 
 	// Get versioned result
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -85,14 +85,21 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 		p.logger.Error("Failed to interpret CNS IPConfigResponse", zap.Error(err), zap.Any("response", resp))
 		return cniTypes.NewError(ErrProcessIPConfigResponse, err.Error(), "failed to interpret CNS IPConfigResponse")
 	}
-	p.logger.Debug("Parsed pod IP", zap.String("podIPNet", podIPNet.String()))
+	p.logger.Debug("Parsed pod IP", zap.String("podIPNet", (*podIPNet)[0].String()))
 
+	// ryand hardcoding until we can use labels
 	cniResult := &types100.Result{
 		IPs: []*types100.IPConfig{
 			{
 				Address: net.IPNet{
-					IP:   net.ParseIP(podIPNet.Addr().String()),
-					Mask: net.CIDRMask(podIPNet.Bits(), 32), // nolint
+					IP:   net.ParseIP((*podIPNet)[0].Addr().String()),
+					Mask: net.CIDRMask((*podIPNet)[0].Bits(), 32), // nolint
+				},
+			},
+			{
+				Address: net.IPNet{
+					IP:   net.ParseIP((*podIPNet)[1].Addr().String()),
+					Mask: net.CIDRMask((*podIPNet)[1].Bits(), 128), // nolint
 				},
 			},
 		},
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -85,7 +85,9 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 		p.logger.Error("Failed to interpret CNS IPConfigResponse", zap.Error(err), zap.Any("response", resp))
 		return cniTypes.NewError(ErrProcessIPConfigResponse, err.Error(), "failed to interpret CNS IPConfigResponse")
 	}
-	p.logger.Debug("Parsed pod IP", zap.String("podIPNet", (*podIPNet)[0].String()))
+	for _, IPNet := range *podIPNet {
+		p.logger.Debug("Parsed pod IP", zap.String("podIPNet", IPNet.String()))
+	}
 
 	// ryand hardcoding until we can use labels
 	cniResult := &types100.Result{
@@ -104,6 +106,8 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 			},
 		},
 	}
+
+	p.logger.Debug("unversioned cni result", zap.Any("CNIresult", cniResult))
 
 	// Get versioned result
 	versionedCniResult, err := cniResult.GetAsVersion(nwCfg.CNIVersion)
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -107,8 +107,6 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 		},
 	}
 
-	p.logger.Debug("unversioned cni result", zap.Any("CNIresult", cniResult))
-
 	// Get versioned result
 	versionedCniResult, err := cniResult.GetAsVersion(nwCfg.CNIVersion)
 	if err != nil {
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -29,7 +29,8 @@ type IPAMPlugin struct {
 
 type cnsClient interface {
 	RequestIPAddress(context.Context, cns.IPConfigRequest) (*cns.IPConfigResponse, error)
-	ReleaseIPAddress(context.Context, cns.IPConfigRequest) error
+	RequestAllIPAddresses(context.Context, cns.IPConfigRequest) (*cns.IPConfigsResponse, error)
+	ReleaseIPs(context.Context, cns.IPConfigRequest) error
 }
 
 // NewPlugin constructs a new IPAM plugin instance with given logger and CNS client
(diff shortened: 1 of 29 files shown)


azure-ipam/ipconfig/ipconfig.go
```diff
@@ -39,18 +39,24 @@ func CreateIPConfigReq(args *cniSkel.CmdArgs) (cns.IPConfigRequest, error) {
...
+	podIPNets = make([]netip.Prefix, len(resp.PodIpInfo))
+
+	for i := range resp.PodIpInfo {
+		podCIDR := fmt.Sprintf(
+			"%s/%d",
+			resp.PodIpInfo[i].PodIPConfig.IPAddress,
+			resp.PodIpInfo[i].NetworkContainerPrimaryIPConfig.IPSubnet.PrefixLength,
+		)
+		podIPNet, err := netip.ParsePrefix(podCIDR)
+		podIPNets[i] = podIPNet
```
ashvindeodhar, line 53: there is no point in assigning this if err is non-nil. This should come after line 56

cilium/cilium_helm_values.yaml
```diff
@@ -22,10 +22,12 @@ endpointRoutes:
 
 extraArgs:
   - --local-router-ipv4=169.254.23.0
+  - --local-router-ipv6=fe80::1
```
ashvindeodhar, line 25: I am just curious on where did you get this from?

cni/network/invoker_cns.go
```diff
@@ -53,6 +53,7 @@ func NewCNSInvoker(podName, namespace string, cnsClient cnsclient, executionMode
 	}
 }
 
+// ryand IPv6 not currently supported
```
ashvindeodhar, line 56: not sure I understand. Which scenario is this code path covering? Don't we support ipv6 in CNS now with overlay?

cni/network/invoker_cns.go
```diff
@@ -85,13 +86,13 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
 
 	info := IPv4ResultInfo{
-		podIPAddress:       response.PodIpInfo.PodIPConfig.IPAddress,
-		ncSubnetPrefix:     response.PodIpInfo.NetworkContainerPrimaryIPConfig.IPSubnet.PrefixLength,
-		ncPrimaryIP:        response.PodIpInfo.NetworkContainerPrimaryIPConfig.IPSubnet.IPAddress,
-		ncGatewayIPAddress: response.PodIpInfo.NetworkContainerPrimaryIPConfig.GatewayIPAddress,
-		hostSubnet:         response.PodIpInfo.HostPrimaryIPInfo.Subnet,
-		hostPrimaryIP:      response.PodIpInfo.HostPrimaryIPInfo.PrimaryIP,
-		hostGateway:        response.PodIpInfo.HostPrimaryIPInfo.Gateway,
+		podIPAddress:       response.PodIpInfo[0].PodIPConfig.IPAddress,
```
ashvindeodhar, line 89: it's wrong to assume that 0th index will always hold ipv4. You need to check if the IPAddress is v4 family or not

cni/network/invoker_cns_test.go
```diff
@@ -56,23 +56,25 @@ func TestCNSIPAMInvoker_Add(t *testing.T) {
...
 						result: &cns.IPConfigResponse{
-							PodIpInfo: cns.PodIpInfo{
-								PodIPConfig: cns.IPSubnet{
-									IPAddress:    "10.0.1.10",
-									PrefixLength: 24,
-								},
-								NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
-									IPSubnet: cns.IPSubnet{
-										IPAddress:    "10.0.1.0",
+							PodIpInfo: []cns.PodIpInfo{
```
ashvindeodhar, line 59: need to add v6 test

cns/NetworkContainerContract.go
```diff
@@ -441,7 +441,7 @@ type GetIPAddressStatusResponse struct {
 
 // GetPodContextResponse is used in CNS Client debug mode to get mapping of Orchestrator Context to Pod IP UUID
 type GetPodContextResponse struct {
-	PodContext map[string]string
+	PodContext map[string][]string
```
ashvindeodhar, line 444: please add a comment explaining why there can be one orch context to many pod IPs

cns/client/client_test.go
```diff
@@ -124,8 +124,8 @@ func getIPNetFromResponse(resp *cns.IPConfigResponse) (net.IPNet, error) {
 	)
 
 	// set result ipconfig from CNS Response Body
-	prefix := strconv.Itoa(int(resp.PodIpInfo.PodIPConfig.PrefixLength))
-	ip, ipnet, err := net.ParseCIDR(resp.PodIpInfo.PodIPConfig.IPAddress + "/" + prefix)
+	prefix := strconv.Itoa(int(resp.PodIpInfo[0].PodIPConfig.PrefixLength))
```
ashvindeodhar, line 127: this code needs to be updated to have a loop that runs over PodIpInfo instead of hardcoding 0th index

cns/restserver/ipam.go
```diff
@@ -86,7 +88,7 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
 		}
 	}
 
-	reserveResp := &cns.IPConfigResponse{
+	reserveResp := cns.IPConfigResponse{
```
ashvindeodhar, line 91: I remember you were experimenting with this change to see if that helps with some error you were having. Is this change needed? I wouldn't think so

cns/restserver/ipam.go
```diff
@@ -102,64 +104,76 @@ var (
...
+						return nil
+					}
+				}
+				endpointInfo.IfnameToIPMap[ipconfigRequest.Ifname].IPv4 = append(endpointInfo.IfnameToIPMap[ipconfigRequest.Ifname].IPv4, ipconfig)
 			}
-			endpointInfo.IfnameToIPMap[ipconfigRequest.Ifname].IPv4 = append(endpointInfo.IfnameToIPMap[ipconfigRequest.Ifname].IPv4, ipconfig)
-		}
 
-		service.EndpointState[ipconfigRequest.InfraContainerID] = endpointInfo
+			service.EndpointState[ipconfigRequest.InfraContainerID] = endpointInfo
```
ashvindeodhar, line 147: there seems be extra white space on line 147

cns/restserver/ipam.go
```diff
@@ -478,31 +499,33 @@ func (service *HTTPRestService) MarkExistingIPsAsPendingRelease(pendingIPIDs []s
...
 	defer service.RUnlock()
 
-	ipID := service.PodIPIDByPodInterfaceKey[podInfo.Key()]
-	if ipID != "" {
-		if ipState, isExist := service.PodIPConfigState[ipID]; isExist {
-			err := service.populateIPConfigInfoUntransacted(ipState, &podIpInfo)
-			return podIpInfo, isExist, err
-		}
+	for _, ipID := range service.PodIPIDByPodInterfaceKey[podInfo.Key()] {
+		if ipID != "" {
```
ashvindeodhar, line 512: remove extra white spaces

cns/restserver/ipam.go
```diff
@@ -478,31 +499,33 @@ func (service *HTTPRestService) MarkExistingIPsAsPendingRelease(pendingIPIDs []s
...
 	}
 
 	return podIpInfo, isExist, nil
 }
 
-func (service *HTTPRestService) AssignDesiredIPConfig(podInfo cns.PodInfo, desiredIPAddress string) (cns.PodIpInfo, error) {
-	var podIpInfo cns.PodIpInfo
+func (service *HTTPRestService) AssignDesiredIPConfig(podInfo cns.PodInfo, desiredIPAddress string) ([]cns.PodIpInfo, error) {
+	var podIpInfo []cns.PodIpInfo
+	podIpInfo = make([]cns.PodIpInfo, 2)
```
ashvindeodhar, line 528: need to remove hardcoded value of 2

cns/restserver/ipam.go
```diff
@@ -526,42 +549,73 @@ func (service *HTTPRestService) AssignDesiredIPConfig(podInfo cns.PodInfo, desir
 			default:
 				return podIpInfo, errors.Errorf("[AllocateDesiredIPConfig] Desired IP is not available %+v", ipConfig)
 			}
-			err := service.populateIPConfigInfoUntransacted(ipConfig, &podIpInfo)
+			err := service.populateIPConfigInfoUntransacted(ipConfig, &podIpInfo[0])
```
ashvindeodhar, line 552: why is this assigning to 0th index only?

cns/restserver/ipam.go
```diff
@@ -526,42 +549,73 @@ func (service *HTTPRestService) AssignDesiredIPConfig(podInfo cns.PodInfo, desir
...
 	return podIpInfo, fmt.Errorf("Requested IP not found in pool")
 }
 
-func (service *HTTPRestService) AssignAnyAvailableIPConfig(podInfo cns.PodInfo) (cns.PodIpInfo, error) {
+// ryand setting to default dualstack mode for testing while waiting to get pod labels - will change
+func (service *HTTPRestService) AssignAnyAvailableIPConfig(podInfo cns.PodInfo) ([]cns.PodIpInfo, error) {
 	service.Lock()
 	defer service.Unlock()
+	var podIpInfo []cns.PodIpInfo
+	podIpInfo = make([]cns.PodIpInfo, 2)
```
ashvindeodhar, line 564: same comment as above

cns/restserver/ipam.go
```diff
@@ -526,42 +549,73 @@ func (service *HTTPRestService) AssignDesiredIPConfig(podInfo cns.PodInfo, desir
...
-
-			return podIPInfo, nil
+			break
+		}
+	}
+	JsonData, err := json.Marshal(&podIpInfo[0])
+	if err != nil {
+		return []cns.PodIpInfo{}, err
+	}
+	logger.Warnf(string(JsonData))
```
ashvindeodhar, line 586: why is this printed as warning?

cns/restserver/ipam.go
```diff
@@ -526,42 +549,73 @@ func (service *HTTPRestService) AssignDesiredIPConfig(podInfo cns.PodInfo, desir
...
-			err := service.populateIPConfigInfoUntransacted(ipConfig, &podIpInfo)
+			err := service.populateIPConfigInfoUntransacted(ipConfig, &podIpInfo[0])
 			return podIpInfo, err
 		}
 	}
 	return podIpInfo, fmt.Errorf("Requested IP not found in pool")
 }
 
-func (service *HTTPRestService) AssignAnyAvailableIPConfig(podInfo cns.PodInfo) (cns.PodIpInfo, error) {
+// ryand setting to default dualstack mode for testing while waiting to get pod labels - will change
```
ashvindeodhar, line 559: this logic needs to change so that it gives out one IP from each available NC. There is no need to be concerned about v4 / v6

cns/restserver/restserver.go
```diff
@@ -58,13 +58,13 @@ type HTTPRestService struct {
...
 	homeAzMonitor            *HomeAzMonitor
 	networkContainer         *networkcontainers.NetworkContainers
-	PodIPIDByPodInterfaceKey map[string]string                    // PodInterfaceId is key and value is Pod IP (SecondaryIP) uuid.
-	PodIPConfigState         map[string]cns.IPConfigurationStatus // Secondary IP ID(uuid) is key
-	IPAMPoolMonitor          cns.IPAMPoolMonitor
-	routingTable             *routes.RoutingTable
-	store                    store.KeyValueStore
-	state                    *httpRestServiceState
-	podsPendingIPAssignment  *bounded.TimedSet
+	PodIPIDByPodInterfaceKey map[string][]string // PodInterfaceId is key and value is Pod IP (SecondaryIP) uuid.
```
ashvindeodhar, line 61: update the comment to reflect the change

cns/restserver/restserver.go
```diff
@@ -106,7 +106,7 @@ type GetHTTPServiceDataResponse struct {
 
 // HTTPRestServiceData represents in-memory CNS data in the debug API paths.
 type HTTPRestServiceData struct {
-	PodIPIDByPodInterfaceKey map[string]string                    // PodInterfaceId is key and value is Pod IP uuid.
+	PodIPIDByPodInterfaceKey map[string][]string                  // PodInterfaceId is key and value is Pod IP uuid.
```
ashvindeodhar, line 109: same comment as above

cns/restserver/ipam_test.go
```diff
@@ -67,26 +67,28 @@ func NewPodState(ipaddress string, prefixLength uint8, id, ncid string, state ty
...
+	}
 
 	// retrieve podinfo from orchestrator context
 	podInfo, err := cns.UnmarshalPodInfo(req.OrchestratorContext)
 	if err != nil {
 		return cns.IPConfigurationStatus{}, errors.Wrap(err, "failed to unmarshal pod info")
 	}
 
-	ipID := svc.PodIPIDByPodInterfaceKey[podInfo.Key()]
+	ipID := svc.PodIPIDByPodInterfaceKey[podInfo.Key()][0]
```
ashvindeodhar, line 91: I don;t see a single ipv6 test in this PR.

ashvindeodhar (changes requested):

azure-ipam/ipam.go
```diff
@@ -85,14 +85,23 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 		p.logger.Error("Failed to interpret CNS IPConfigResponse", zap.Error(err), zap.Any("response", resp))
 		return cniTypes.NewError(ErrProcessIPConfigResponse, err.Error(), "failed to interpret CNS IPConfigResponse")
 	}
-	p.logger.Debug("Parsed pod IP", zap.String("podIPNet", podIPNet.String()))
+	for _, IPNet := range *podIPNet {
+		p.logger.Debug("Parsed pod IP", zap.String("podIPNet", IPNet.String()))
+	}
 
+	// ryand hardcoding until we can use labels
```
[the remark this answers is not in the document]
rjdenney: I think this was a comment added back when we're discussing node labels and is now out of date. I'll remove it.

cilium/cilium_helm_values.yaml
rjdenney: So for the local router ipv4 we use the link-local address which is that line above it. When I was talking with Will he mentioned that is something I would have to add and also needs to be added on the AKS side. 

I looked up what is used for an ipv6 link-local address and found here https://blogs.infoblox.com/ipv6-coe/fe80-1-is-a-perfectly-valid-ipv6-default-gateway-address/ that it needs to be an address with a header of fe80.

cni/network/invoker_cns.go
rjdenney: This may be able to be changed back to normal or will have to be something different if we end up using the new API. I originally was in the mindset that the first ip would be ipv4 since that is how I originally had it set up as and if there was only one IP it owuld default to the first index anyways. Will still need to be changed though.

cni/network/invoker_cns.go
rjdenney: I am going to remove this since it is the CNI and relates more to what is being done on Windows. Should have been removed earlier.

ashvindeodhar (changes requested):

azure-ipam/ipam_test.go
```diff
@@ -31,10 +31,12 @@ func (c *MockCNSClient) RequestIPAddress(ctx context.Context, ipconfig cns.IPCon
 	case "failProcessCNSResp":
 		result := &cns.IPConfigResponse{
 			PodIpInfo: cns.PodIpInfo{
-				PodIPConfig: cns.IPSubnet{
-					IPAddress:    "10.0.1.10.2", // invalid ip address
-					PrefixLength: 24,
-				},
+				{
+					PodIPConfig: cns.IPSubnet{
+						IPAddress:    "10.0.1.10.2", // invalid ip address
```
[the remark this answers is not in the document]
rjdenney: There is an additional .2 at the end which would make it 40 bytes. A valid IP would stop at 10.0.1.10

azure-ipam/ipam_test.go
```diff
@@ -31,10 +31,12 @@ func (c *MockCNSClient) RequestIPAddress(ctx context.Context, ipconfig cns.IPCon
 	case "failProcessCNSResp":
 		result := &cns.IPConfigResponse{
 			PodIpInfo: cns.PodIpInfo{
```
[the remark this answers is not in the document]
rjdenney: This test is using the old version of RequestIPAddress which expects the old contract without the slice

cns/restserver/ipam_test.go
```diff
@@ -105,38 +112,60 @@ func NewPodStateWithOrchestratorContext(ipaddress, id, ncid string, state types.
...
-	for _, ipconfig := range ipconfigs {
-		secIpConfig := cns.SecondaryIPConfig{
-			IPAddress: ipconfig.IPAddress,
-			NCVersion: -1,
+func UpdatePodIpConfigState(t *testing.T, svc *HTTPRestService, ipconfigs map[string]cns.IPConfigurationStatus, ncIDs []string) error {
+	// Create each NC 
+	for _, NCID := range ncIDs {
+		secondaryIPConfigs := make(map[string]cns.SecondaryIPConfig)
+		for _, ipconfig := range ipconfigs {
+			if ipconfig.NCID == NCID {
```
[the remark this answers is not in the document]
rjdenney: So this was a part of the updates that I had to make to the ipam_test.go suite. There may be a way to do it but what this part of the function does is fill up the NCs with the ipconfigs that are passed in. 

The outer loop create an NC for each NCID that I pass in. The inner loop then goes through the ipconfigs passed in and searches for all of the ipconfigs that are assigned to that NC. By checking if the ipconfig equals the NCID we are on we can keep IPs assigned to only the NC we want them to be on.

What I eventually did later down the suite was make all of the tests able to have variable inputs for IPs, NCs, and prefixes. Each test should be able to use those to create whatever pods needed to run the test and can work for single or multiple NC cases.

cns/restserver/internalapi_test.go
```diff
@@ -581,28 +581,28 @@ func validateNCStateAfterReconcile(t *testing.T, ncRequest *cns.CreateNetworkCon
...
+			ipConfigstate := svc.PodIPConfigState[ipID]
+			if ipConfigstate.GetState() != types.Assigned {
+				t.Fatalf("IpAddress %s is not marked as assigned to Pod: %+v, ipState: %+v", ipaddress, podInfo, ipConfigstate)
+			}
 
-		// Validate if IPAddress matches
-		if ipConfigstate.IPAddress != ipaddress {
-			t.Fatalf("IpAddress %s is not same, for Pod: %+v, actual ipState: %+v", ipaddress, podInfo, ipConfigstate)
-		}
+			// Validate if IPAddress matches
```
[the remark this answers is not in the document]
rjdenney: This section got indented because I had to add an extra for loop for ranging over svc.PodIPIDByPodInterfaceKey[podInfo.Key()]  since the map's value is now a slice of IPs instead of just an IP.

PR #1716, opened 2022-12-12
fix: deadlock when os.Stat errors within rotate method

ramiro-gamarra:
**Reason for Change**:
This PR addresses a deadlock path in file rotation: when reading file stats to get file size, an error is handled with a recursive call to `logger.Debugf`, which deadlocks since a mutex cannot be acquired recursively.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

Cleaned up a few more things in the `rotate` method. The changes should be equivalent in behavior, with todos added to investigate if the original rotation logic is correct.

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -149,42 +149,54 @@ func (logger *Logger) getLogFileName() string {
 	return logFileName
 }
 
-// todo: handle errors and use atomic file rotation
-// Rotate checks the active log file size and rotates log files if necessary.
-func (logger *Logger) rotate() {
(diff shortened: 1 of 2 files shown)


rbtr:
seems like this could have been caught by a UT, can you add one that exercises this?

rbtr (approved):
@ramiro-gamarra is this one good to go?

timraymond (approved):
Looks good

PR #1728, opened 2022-12-16
Fix incorrect empty body provided to Wireserver

timraymond:
In cases where clients using a WireserverTransport (namely the NMAgent one) do not wish to provide a body, one must be provided to satisfy quirks of Wireserver. Initially, an empty Go string was thought to be sufficient, but this is not. Wireserver expects an empty JSON string: specifically the 2-byte sequence "".

Furthermore, this logic was only being triggered on cases where a nil body was presented to the WireserverTransport's logic. Given that there are now more linters enforcing the use of `http.NoBody`, this is distinct from `nil` and must be separately checked for (but treated in the same way).

Tests and inline documentation have been updated to reflect this new understanding.

- [X] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/nmagent/internal/wireserver_test.go
+++ b/nmagent/internal/wireserver_test.go
@@ -2,6 +2,7 @@ package internal
 
 import (
 	"encoding/json"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"testing"
@@ -220,12 +221,16 @@ func TestWireserverTransportPostBody(t *testing.T) {
 	// all PUT and POST requests must have an empty string body
 	t.Parallel()
 
-	bodyIsNil := false
+	var gotBody []byte
 	client := &http.Client{
 		Transport: &WireserverTransport{
 			Transport: &TestTripper{
 				RoundTripF: func(req *http.Request) (*http.Response, error) {
-					bodyIsNil = req.Body == nil
+					body, err := io.ReadAll(req.Body)
+					if err != nil {
+						t.Fatal("unexpected error reading request body: err:", err)
+					}
+					gotBody = body
 					rr := httptest.NewRecorder()
 					_, _ = rr.WriteString(`{"httpStatusCode": "200"}`)
 					rr.WriteHeader(http.StatusOK)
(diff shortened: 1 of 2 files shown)


rbtr (approved):
horrible

rbtr (approved):

nmagent/internal/wireserver_test.go
```diff
@@ -247,8 +252,9 @@ func TestWireserverTransportPostBody(t *testing.T) {
 	}
 	defer resp.Body.Close()
 
-	if bodyIsNil {
-		t.Error("downstream request body to wireserver was nil, but not expected to be")
+	exp := `""` // an empty JSON string
+	if got := string(gotBody); got != exp {
+		t.Errorf("unexpected body received: got: %q, exp: %q\n", got, exp)
```
ramiro-gamarra, line 257: nit: is the new line needed? same for others.

ramiro-gamarra (approved):

nmagent/internal/wireserver_test.go
timraymond: @ramiro-gamarra do you mean the changed content, or a literal `\n`? There are no extra `\n`s (look at the line number sequence). It might be a GH display bug.

smittal22:
This fix works for us. Have tested by building a CNS executable and joinNetwork calls do not fail anymore.

nmagent/internal/wireserver_test.go
ramiro-gamarra: the `\n` in the error message

timraymond: Okay, so apparently, the presence or absence of it makes no difference in the output--they're stripped internally. I add them out of habit it because it matters for `fmt.Println` vs `fmt.Printf`. Since it won't create janky output, I'll keep it in mind for next time, but I won't fix it here to avoid blocking the merge.

ashvindeodhar:
This is fixing a regression in the master which has caused the releases 1.4.38 and 1.4.39 as non-usable for swift standalone. 
Bypassing the windows test check which has been failing for past few days. This is a contained change w/o which the master will be in a broken state.

PR #1730, opened 2022-12-19
Network name as an optional variable

pjohnst5:
Adding in the capabilty of setting the network name in the azure cni conflist, for the acn-cli image


thatmattlong:
I'll approve to unblock, but we probably shouldn't be adding any new code to the acncli and instead this functionality should exist in dropgz, our new tool for installing CNI + conflists

thatmattlong (approved):

PR #1736, opened 2022-12-22
remove windows 2004 aks-e tests from pipeline

estebancams:
**Reason for Change**:
We still don't support windows 2004 for aks and aks-e tests are failing very often. While we determine the root cause, in order to allow PRs to go in, and since we'll still run windows 1903 tests, we can disable 2004 tests as a mitigation
A new task was reported to add win 22 tests once we migrate over from aks-e (https://msazure.visualstudio.com/One/_workitems/edit/16667635)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/pipeline.yaml
+++ b/.pipelines/pipeline.yaml
@@ -390,7 +390,6 @@ stages:
       - "aks_swift_e2e"
       - "cilium_e2e"
       - "windows_19_03_e2e"
-      - "windows_20_04_e2e"
     jobs:
       - job: delete_remote_artifacts
         displayName: Delete remote artifacts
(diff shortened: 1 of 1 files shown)


pjohnst5 (approved):
LGTM as long as it's not available in AKS anyways (and aks-e is deprecated now)
It does fail quite often so in my opinion it is best to remove it while we then put the win 22 tests in for aks

pjohnst5 (dismissed):
Yeah maybe we should first get win 22 operational before removing this based on discussion offline

rbtr (approved):

PR #1744, opened 2023-01-06
nmagent get nv version list api V2 refactor

csfmomo:
**Reason for Change**:

Refactor the way we get nc version. CNS used to call nmagent nc version api v1 which needs auth token, today, nmagent provides v2 api which doesn't need token and the process can be simplified.

work item 14734259

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/restserver/api_test.go
```diff
@@ -588,13 +601,26 @@ func TestGetNetworkContainerVersionStatus(t *testing.T) {
...
+	// 	return nmagent.NCVersionList{
+	// 		Containers: []nmagent.NCVersion{
+	// 			{
+	// 				NetworkContainerID: params.ncID,
+	// 				Version:            "0", // explicitly 1 less than the version above
+	// 			},
+	// 		},
+	// 	}, nil
+	// }
+
```
timraymond, line 623: Please remove the commented-out code. We can find it and bring it back with `git` if we need to.

cns/restserver/internalapi.go
```diff
@@ -94,6 +94,15 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
 	}
 	service.RUnlock()
 
+	ncVersionListResp, err := service.nma.GetNCVersionList(context.TODO())
```
timraymond, line 97: For all of these NMAgent methods, a `context.Context` is required to control the request. Please plumb in the incoming request to CNS (available via `http.(*Request).Context()` so that requests to NMAgent will be retried only as long as the upstream requestor to CNS is willing to wait (retries will occur indefinitely with a TODO).

There's several instances of `context.TODO()` in this patch... won't flag them to keep the noise down.

cns/restserver/util.go
```diff
@@ -789,27 +808,15 @@ func (service *HTTPRestService) isNCWaitingForUpdate(
 		}
 	}
 
-	getNCVersionURL, ok := ncVersionURLs.Load(ncid)
+	ncTargetVersion, _ := strconv.Atoi(ncVersion)
```
timraymond, line 811: This error needs to be handled.

cns/restserver/util.go
```diff
@@ -384,12 +384,22 @@ func (service *HTTPRestService) getNetworkContainerResponse(
 
 		containerID, exists = service.state.ContainerIDByOrchestratorContext[podInfo.Name()+podInfo.Namespace()]
 
+		ncVersionListResp, err := service.nma.GetNCVersionList(context.TODO())
+		if err != nil {
+			logger.Errorf("failed to get nc version list from nmagent")
```
timraymond, line 389: The function does not terminate here.

timraymond (changes requested):

cns/restserver/internalapi.go
csfmomo: Updated with timeout.

cns/restserver/util.go
csfmomo: Updated. Terminate it when it needs termination. Some of the case should not terminate though nc version return has error, then I just skip version check and put nc as pending programing from nmagent side.

cns/restserver/util.go
```diff
@@ -773,13 +802,13 @@ func (service *HTTPRestService) populateIPConfigInfoUntransacted(ipConfigStatus
...
-// isNCWaitingForUpdate :- Determine whether NC version on NMA matches programmed version
+// isNCWaitingForUpdateV2 :- Determine whether NC version on NMA matches programmed version
 // Return error and waitingForUpdate as true only CNS gets response from NMAgent indicating
 // the VFP programming is pending
 // This returns success / waitingForUpdate as false in all other cases.
-func (service *HTTPRestService) isNCWaitingForUpdate(
-	ncVersion, ncid string,
-) (waitingForUpdate bool, returnCode types.ResponseCode, message string) {
+// V2 is using the nmagent get nc version list api v2 which doesn't need authentication token
+func (service *HTTPRestService) isNCWaitingForUpdateV2(
```
thatmattlong, line 810: If we're not keeping both versions of this function in code, I'd suggest to just call this `isNCWaitingForUpdate` (the existing name) and just update the implementation.

csfmomo: OK

cns/restserver/internalapi.go
```diff
@@ -94,32 +92,41 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
-		}
-
-		ncVersionURLs.Store(nc.NetworkContainerid, nmaReq)
-		waitingForUpdate, _, _ := service.isNCWaitingForUpdate(nc.Version, nc.NetworkContainerid)
-
-		body, _ = json.Marshal(nc)
-		req, _ = http.NewRequest(http.MethodPost, "", bytes.NewBuffer(body))
-		req.Header.Set(common.ContentType, common.JsonContent)
+	skipNCVersionCheck := false
+	ctx, cancel := context.WithTimeout(context.Background(), ContextTimeOut)
```
thatmattlong, line 96: This may already be an existing pattern so I won't block on this, but I would recommend renaming the `ContextTimeout` variable to something meaningful, like `nmaAPICallTimeout`

cns/restserver/internalapi.go
```diff
@@ -94,32 +92,41 @@ func (service *HTTPRestService) SyncNodeStatus(dncEP, infraVnet, nodeID string,
...
+	defer cancel()
+	ncVersionListResp, err := service.nma.GetNCVersionList(ctx)
+	if err != nil {
+		skipNCVersionCheck = true
+		logger.Errorf("failed to get nc version list from nmagent")
+	}
 
-		w := httptest.NewRecorder()
-		service.createOrUpdateNetworkContainer(w, req)
+	if !skipNCVersionCheck {
```
thatmattlong, line 104: Doesn't look like we ever run any code if this condition doesn't match. Instead, I suggest to eliminate this variable `skipNCVersionCheck` completely and just return in the `err` check block. Then you can remove this `if` statement and it would be cleaner.

cns/restserver/internalapi.go
csfmomo: I can update it with a nma specific varaible and rename it.

rbtr:
@csfmomo there are several linting issues on this branch still, can you fix those? You can check the lint locally with `make lint`

cns/restserver/api_test.go
```diff
@@ -623,14 +631,18 @@ func TestGetNetworkContainerVersionStatus(t *testing.T) {
...
-		return nmagent.NCVersion{}, errors.New("boom") //nolint:goerr113 // it's just a test
+	mnma.GetNCVersionListF = func(_ context.Context) (nmagent.NCVersionList, error) {
+		rsp := nmagent.NCVersionList{
+			Containers: []nmagent.NCVersion{},
+		}
+		return rsp, errTest
 	}
+
 	mnma.JoinNetworkF = func(_ context.Context, _ nmagent.JoinNetworkRequest) error {
-		return errors.New("boom") //nolint:goerr113 // it's just a test
```
timraymond, line 630: These should be reverted back. The point of these dynamic errors is that the actual error is completely inconsequential. At the same time, we want to make it obvious that we definitely _intend_ to return an error here. Assigning it to a package variable does a few things opposing this:

1) I'm not really sure what `errTest` actually is: it's typed as an error, but is it nil? Did someone else assign to it? I don't know.
2) Because there is no way to make it immutable, it is inter-test state that can create subtle flakiness as these tests change over time.
3) It gives the error an identity that we want to avoid. `JoinNetwork` can fail in any number of ways, and we want to say here that the specific error is completely inconsequential (only that it implements `error`). This is to encompass timeouts, bad requests, actual failures, etc.

So yeah, even though the linter complains about this, the linter is wrong.

nmagent/requests.go
```diff
@@ -448,7 +448,7 @@ func (NCVersionListRequest) Method() string {
 
 // Path returns the path required to issue the request.
 func (NCVersionListRequest) Path() string {
-	return "/NetworkManagement/interfaces/api-version/1"
```
timraymond, line 451: Would there be any reason for a client to use the version 1 API going forward? Can't remember if we discussed this...

timraymond (changes requested):

nmagent/requests.go
csfmomo: Actually, I didn't find any official support for "/NetworkManagement/interfaces/api-version/1", so it's not correct here.
I only find support for "/NetworkManagement/interfaces/api-version/2", version 2 is an api without token.
or "NetworkManagement/interfaces/NetworkID/networkContainers/NCID/version/authenticationToken/%s/api-version/1" version 1 is an api with token.
But, I tested on machine, "/NetworkManagement/interfaces/api-version/1" returning the same result as "/NetworkManagement/interfaces/api-version/2"
However, I think we still should make it correct.

timraymond: @csfmomo Okay cool, makes sense to me then :)

cns/restserver/api_test.go
csfmomo: I think we can leave it as it is for this PR and I can open another PR to address this issue.

rbtr: @csfmomo, @timraymond is asking that you _not_ make the change here, so please revert it. No need to change it, then open another PR to change it back.

rbtr (changes requested):
see comment

cns/restserver/api_test.go
csfmomo: Reverted.

rbtr (approved):

PR #1756, opened 2023-01-11
add azid and enableazr fields to putNC call

smittal22:
**Reason for Change**:

To enable AZR support, we need to pass two extra params `azID` (uint) and `enableAZR` (boolean) as part of body to NMAgent.

**Issue Fixed**:

https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/16223444

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**: e2e flow has been tested by creating and deleting NCs with both `enableAZR` set to `true` and `false`

### DIFF
--- a/nmagent/requests.go
+++ b/nmagent/requests.go
@@ -68,6 +68,12 @@ type PutNetworkContainerRequest struct {
 	// PrimaryAddress is the primary customer address of the interface in the
 	// management VNet
 	PrimaryAddress string
+
+	// AzID is the home AZ ID of the network container
+	AzID uint `json:"azID"`
+
+	// EnableAZR denotes whether AZR is enabled for network container or not
+	EnableAZR bool `json:"enableAZR"`
 }
 
 type internalNC struct {
@@ -83,6 +89,8 @@ type internalNC struct {
 	Policies   []Policy `json:"policies"`
 	VlanID     int      `json:"vlanId"`
 	GREKey     uint16   `json:"greKey"`
+	AzID       uint     `json:"azID"`
+	EnableAZR  bool     `json:"enableAZR"`
 }
 
 func (p *PutNetworkContainerRequest) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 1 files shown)


nmagent/requests.go
```diff
@@ -68,6 +68,12 @@ type PutNetworkContainerRequest struct {
 	// PrimaryAddress is the primary customer address of the interface in the
 	// management VNet
 	PrimaryAddress string
+
+	// AzID is the home AZ ID of the network container
+	AzID uint
```
rbtr, line 73: does the request get serialized differently if the type is `uint` vs `int`? I don't think JSON represents these types differently. Same question for the use of `uint16`. Do we need to constrain these types or is `int` ok?

rbtr: It's fine if that's what they support but that isn't exactly my question. Does JSON have a different representation of `int`/`uint`/`uint16`? The request is being serialized to JSON when we query NMA, right?

smittal22: @rbtr You are correct. There is no differentiation between `uint` and `int` for JSON, or `uint16` for that matter.
Also, we encode the body as JSON from DNC and send it to CNS. CNS then unmarshals the body to the required struct and passes it onto NMAgent.

rbtr: let's align all of the values in these requests on a single integer type, then

rbtr (changes requested):

nmagent/requests.go
smittal22: Well, this still needs to be in `uint`, right? This is what we are unmarshalling into. I can chat briefly if there's some confusion.

smittal22: @timraymond, I chatted with @rbtr and it seems even though we have these different data types like `uint16`, `uint64`, before passing the request to NMA, we're just converting it to a JSON blob and since JSON does not distinguish between these different bit lengths or even between `uint` and `int` we're losing this info anyway.
Even the `Version` that we make sure to be `uint64`, we're type casting it to `int` and then converting to string.
He suggested I leave `AzID` as `uint` for now but this is a broader discussion that we should have in future.

ashvindeodhar (dismissed):

ashvindeodhar (dismissed):

rbtr (approved):

PR #1757, opened 2023-01-11
Fix incorrect HTTP status from publish NC

timraymond:
CNS was responding with an HTTP status code of "0" from NMAgent. Successes are supposed to be 200. The C-style var block at the beginning of publishNetworkContainer was the reason for this. During refactoring, the location where this status code was set to a successful value of 200 was accidentally removed. Because the var block declared the variable and silently initialized it to 0, the compiler did not flag this bug as it otherwise would have. The status code has been removed from this block and explicitly defined and initialized to a correct value of 200. Subsequent error handling will change this as necessary.

Also, despite consumers depending on this status, there were no tests to verify that the status was set correctly. Tests have been added to reflect this dependency.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -1162,11 +1162,14 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
 		req                 cns.PublishNetworkContainerRequest
 		returnCode          types.ResponseCode
 		returnMessage       string
-		publishStatusCode   int
 		publishResponseBody []byte
 		publishErrorStr     string
 		isNetworkJoined     bool
 	)
+
+	// publishing is assumed to succeed unless some other error handling sets it
+	// otherwise
+	publishStatusCode := http.StatusOK
 
 	err := service.Listener.Decode(w, r, &req)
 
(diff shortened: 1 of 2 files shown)


cns/restserver/api.go
```diff
@@ -1245,14 +1247,21 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
 		returnCode = types.UnsupportedVerb
 	}
 
+	// this is an ugly hack because DNC depends on checking this status code,
+	// even though it's no longer necessary for it to do so. It does not need to
+	// handle retries because retries will be handled by the nmagent client in
+	// CNS. However, there are versions of DNC out there that still rely on this
+	// body being present.
+	publishResponseBody := fmt.Sprintf(`{"httpStatusCode":"%d"}`, publishStatusCode)
```
ramiro-gamarra, line 1255: It seems DNC also expects a similar response on the unpublish path

timraymond: Fixed in b4443fd

ramiro-gamarra (approved):

PR #1764, opened 2023-01-19
Make binaries cross-os

pjohnst5:
Similarly to how we have it in DNC, making the build of binaries Cross-OS compatible (windows) so that we can build the binaries on windows machines as well


thatmattlong:
This LGTM, though the github diff view is showing some addtions/removals at some line endings. Could you double check that these are still using LF line endings instead of CRLF?

thatmattlong (approved):
approving, contingent on the comment I posted in the general discussion

rbtr:
Go trivially cross-compiles (set GOOS=windows), and I think trying to make all of the _tooling_ multi-OS is unnecessary and adds quite a lot of complexity and maintenance burden. Is there a strong requirement for building on Windows?

rbtr (changes requested):
blocking for discussion

thatmattlong:
This PR enables building the linux binaries from a Windows dev box. Windows/powershell doesn't handle this syntax of placing a variable assignment in front of a command, so we can use `make`'s support for per-directive environment variables instead.

pjohnst5:
The one thing I need this for is to build cns in a service fabric pipeline.
Service fabric is primarily windows, so all of the jobs up until now are run on windows machines.
Is adding just the one variable export ok?

pjohnst5:
I checked the line endings, they are LF, the reason there's a change is that there was trailing whitespace before:
image

rbtr:
> The one thing I need this for is to build cns in a service fabric pipeline. Service fabric is primarily windows, so all of the jobs up until now are run on windows machines. Is adding just the one variable export ok?

I would prefer the "special case" be in the SF pipeline, adding a Linux runner has basically no maintenance overhead or complexity, vs the complication being added to our Makefile where we need to maintain it.

pjohnst5:
We already support windows make targets in our makefile such as npm-windows-image and cns-windows-image, which are run on windows vms in our CI/CD

Since we already have windows targets, I think we should allow this change, as it keeps Linux working still

I would have to add a Linux job to the service faction pipeline, to change to GOOS to windows, to then copy those binaries from a Linux job to windows job, which just seems like more convolution than this change

What do you think @rbtr ?

@thatmattlong what do you think?

PR #1765, opened 2023-01-20
feat: allow the CNI conflist generation settings to be configured via…

thatmattlong:
… cmd line args

**Reason for Change**:

Allow the CNI conflist generation settings to be overriden with env vars and command line flags

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/service/main.go
```diff
@@ -78,6 +78,8 @@ const (
 	// 720 * acn.FiveSeconds sec sleeps = 1Hr
 	maxRetryNodeRegister = 720
 	initCNSInitalDelay   = 10 * time.Second
+
+	envVarDisableCNIConflistGeneration = "CNS_DISABLE_CNI_CONFLIST_GENERATION"
```
rbtr, line 82: is making this opt-out instead of opt-in the safest way to roll it out?

thatmattlong: hm yeah I suppose we could make it opt-in, the real purpose of this is to just override the option in config so we can enable in some places and disable in others, so I think it's probably the same either way?

thatmattlong: actually no that's not true, if we want to only start to roll the feature when CNS supports this env var option, we should leave it disabled in config and then explicitly override in the env var. Will change

rbtr (approved):

PR #1767, opened 2023-01-24
Fix the NC ID error when get container

csfmomo:
fix: repair the NC ID mismatch when get network container.

**Reason for Change**:
 Fix the mismatch between NC version returned from nmagent and CNS state. Add unit test to gate it.

**Issue Fixed**:
CNS is onboard with nmagent nc version api 2. NC ID returned from nmagent doesn't include the Swift_ prefix but pure uuid. However, the NC ID in the request and cns state has the Swift_ prefix. This mismatch result in the nc version update failure.

**Requirements**:

- [ ] includes documentation
- [ X] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/util.go
+++ b/cns/restserver/util.go
@@ -394,9 +394,8 @@ func (service *HTTPRestService) getNetworkContainerResponse(
 		}
 		nmaNCs := map[string]string{}
 		for _, nc := range ncVersionListResp.Containers {
-			nmaNCs[nc.NetworkContainerID] = nc.Version
+			nmaNCs[cns.SwiftPrefix+nc.NetworkContainerID] = nc.Version
 		}
-
 		if exists && !skipNCVersionCheck {
 			// If the goal state is available with CNS, check if the NC is pending VFP programming
 			waitingForUpdate, getNetworkContainerResponse.Response.ReturnCode, getNetworkContainerResponse.Response.Message = service.isNCWaitingForUpdate(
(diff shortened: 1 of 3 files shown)


cns/restserver/util.go
```diff
@@ -394,9 +394,8 @@ func (service *HTTPRestService) getNetworkContainerResponse(
 		}
 		nmaNCs := map[string]string{}
 		for _, nc := range ncVersionListResp.Containers {
-			nmaNCs[nc.NetworkContainerID] = nc.Version
+			nmaNCs[cns.SwiftPrefix+nc.NetworkContainerID] = nc.Version
```
ashvindeodhar, line 397: can you fix this in other places for nmaNCs?
I found one more where this bug still exists -
https://github.com/Azure/azure-container-networking/blob/ddb595435169094b3a178f89078e595957ae462a/cns/restserver/util.go#L563

ashvindeodhar (changes requested):

cns/restserver/util.go
ashvindeodhar: spoke with Shufang. Resolving this.

ashvindeodhar (approved):

pjohnst5 (dismissed):

pjohnst5 (approved):

ashvindeodhar (dismissed):

ashvindeodhar (approved):

PR #1768, opened 2023-01-24
set mellanox reg key

rajvinar:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/service/main.go
```diff
@@ -1245,3 +1249,112 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 
 	return nil
 }
+
+func SetMellanoxRegKey() {
```
rbtr, line 1253: this should not be in main, it should be in a platform specific package

rbtr (commented):
This seems like underlying infra config, and I'm not sure it should be CNS's job to configure prerequisites like this. Shouldn't this responsibility lie with the orchestrator which expects to use CNS/CNI?

cns/service/main.go
```diff
@@ -1252,3 +1253,108 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+		cmd = fmt.Sprintf(" Get-NetAdapterAdvancedProperty | Where-Object { $_.RegistryKeyword -like \"%s\" -and $_.Name -eq \"%s\" } | Select-Object -ExpandProperty RegistryValue", priorityVLANTagIdentifier, adapatername)
+		regvalue, err := platform.ExecutePowershellCommand(cmd)
+		if err != nil {
+			logger.Printf("Error while checking registery value for PriorityVLANTag for adapater: %v", err)
+		}
+		if regvalue != "3" {
+			logger.Printf("Failure while setting Mellanox Network Adapter: \"%s\" with \"%s\" property value as 3", adapatername, priorityVLANTagIdentifier)
+		} else {
+			logger.Printf("Successfully set Mellanox Network Adapter: \"%s\" with \"%s\" property value as 3", adapatername, priorityVLANTagIdentifier)
+		}
```
ashvindeodhar, line 1303: pls add return here so that we can avoid 'else'. We still need the code under else but no need for 'else'

cns/service/main.go
```diff
@@ -1252,3 +1253,108 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+			logger.Printf("Error while setting up registery value for PriorityVLANTag for adapater: %v", err)
+		}
+		//check if value is set or not
+		cmd = fmt.Sprintf(" Get-NetAdapterAdvancedProperty | Where-Object { $_.RegistryKeyword -like \"%s\" -and $_.Name -eq \"%s\" } | Select-Object -ExpandProperty RegistryValue", priorityVLANTagIdentifier, adapatername)
+		regvalue, err := platform.ExecutePowershellCommand(cmd)
+		if err != nil {
+			logger.Printf("Error while checking registery value for PriorityVLANTag for adapater: %v", err)
+		}
+		if regvalue != "3" {
+			logger.Printf("Failure while setting Mellanox Network Adapter: \"%s\" with \"%s\" property value as 3", adapatername, priorityVLANTagIdentifier)
```
ashvindeodhar, line 1300: need to return err here

cns/service/main.go
```diff
@@ -1252,3 +1253,108 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+		} else {
+			logger.Printf("Successfully set Mellanox Network Adapter: \"%s\" with \"%s\" property value as 3", adapatername, priorityVLANTagIdentifier)
+		}
+	} else {
+		//looks like its version 3 so vlantag not available so need to follow other route
+		logger.Printf("Could not find \"%s\" in adapter's advanced properties", priorityVLANTagIdentifier)
+		logger.Printf("Proceeding in a different way")
```
ashvindeodhar, line 1307: having 'Proceeding in a different way' in telemetry doesn't help anyone.

cns/service/main.go
```diff
@@ -1252,3 +1253,108 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+		logger.Printf("Searching through CIM instances for Network devices with \"%s\" in the name", mellanoxSearchString)
```
ashvindeodhar, line 1308: no need for multiple logger.Printf. consolidate into single statement

ashvindeodhar: same applies to other places below

cns/service/main.go
```diff
@@ -1252,3 +1253,108 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+		if deviceid != "" {
+			logger.Printf("Device ID found: %s", deviceid)
+			logger.Printf("Getting Pnp Device properites for %s", deviceid)
+			cmd = fmt.Sprintf("Get-PnpDeviceProperty -InstanceId \"%s\" | Where-Object KeyName -EQ \"DEVPKEY_Device_Driver\" | Select-Object -ExpandProperty Data", deviceid)
+			registryKeySuffix, err := platform.ExecutePowershellCommand(cmd)
+			if err != nil {
+				logger.Printf("Error while executing pwsh command to get device id of %s: %v", adapatername, err)
+				return
+			}
+			logger.Printf("Registry key suffix found: %s", registryKeySuffix)
```
ashvindeodhar, line 1324: look to reduce the number of these logs and see if we can only log important / things which can be used to debug issues from logs.

cns/service/main.go
```diff
@@ -1252,3 +1253,108 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+			registryKeyFullPath := registryKeyPrefix + registryKeySuffix
+			logger.Printf("Registry key full path: %s", registryKeyFullPath)
+			logger.Printf("Updating %s to be 3", priorityVLANTagIdentifier)
```
ashvindeodhar, line 1327: no need to log this upfront. It can be logged as error if unable to update to 3

ashvindeodhar (changes requested):

cns/service/main.go
rajvinar: Ok I will move this function to platform/os_windows.go ... make sense to me and thanks for example Ashvin.

rajvinar: I will also follow the pattern of this function  SetSdnRemoteArpMacAddres and call it from main.. let me know if any concerns?

cns/service/main.go
```diff
@@ -13,6 +13,7 @@ import (
 	"net/http/pprof"
```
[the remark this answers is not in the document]
rajvinar: Done

cns/service/main.go
```diff
@@ -1245,3 +1249,112 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 
 	return nil
 }
+
+func SetMellanoxRegKey() {
+	var mellanoxSearchString = "*Mellanox*"
```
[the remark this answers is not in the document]
rajvinar: Sure will fix

cns/service/main.go
```diff
@@ -1245,3 +1249,112 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+	var priorityVLANTagIdentifier = "*PriorityVLANTag"
+	var registryKeyPrefix = "HKLM:\\System\\CurrentControlSet\\Control\\Class\\"
+
+	//get adapter name
+	cmd := fmt.Sprintf("Get-NetAdapter | Where-Object { $_.InterfaceDescription -like \"%s\" } | Select-Object -ExpandProperty Name", mellanoxSearchString)
+	adapatername, err := platform.ExecutePowershellCommand(cmd)
+
+	if err != nil {
+		log.Printf("Error while executing pwsl command to get net adapter list: %v", err)
+		return
```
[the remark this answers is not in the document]
rajvinar: ok adding error statement and logging them if we get into mellanox issue... I will let you review my next update and get more review

cns/service/main.go
```diff
@@ -1245,3 +1249,112 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+	var mellanoxSearchString = "*Mellanox*"
+	var priorityVLANTagIdentifier = "*PriorityVLANTag"
+	var registryKeyPrefix = "HKLM:\\System\\CurrentControlSet\\Control\\Class\\"
+
+	//get adapter name
+	cmd := fmt.Sprintf("Get-NetAdapter | Where-Object { $_.InterfaceDescription -like \"%s\" } | Select-Object -ExpandProperty Name", mellanoxSearchString)
+	adapatername, err := platform.ExecutePowershellCommand(cmd)
+
+	if err != nil {
+		log.Printf("Error while executing pwsl command to get net adapter list: %v", err)
```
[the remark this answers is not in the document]
rajvinar: powershell.. let me edit my logs

cns/service/main.go
```diff
@@ -1245,3 +1249,112 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+		return
+	}
+	if adapatername == "" {
+		log.Printf("No Network adapter found with \"%s\" in description", mellanoxSearchString)
+		return
```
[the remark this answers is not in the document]
rajvinar: just logging it

cns/service/main.go
```diff
@@ -1245,3 +1249,112 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+	}
+
+	log.Printf("Name of Mellanox adapter : %v", adapatername)
+
+	//Find if adapter with Priority vlan tag (version 4 or up) or not(version 3)
+	cmd = fmt.Sprintf("Get-NetAdapterAdvancedProperty | Where-Object { $_.RegistryKeyword -like \"%s\" -and $_.Name -eq \"%s\" } | Select-Object -ExpandProperty Name", priorityVLANTagIdentifier, adapatername)
+	log.Printf("%v", cmd)
+	adapterNamewithVLANTag, err := platform.ExecutePowershellCommand(cmd)
```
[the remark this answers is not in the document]
rajvinar: adapter name having vlan tag in advance property. So for version 4 it will return name while for version 3 it will be blank since no *priorityvlantag reg key found

cns/service/main.go
rajvinar: sure

cns/service/main.go
rajvinar: cleaned up unnecessary comments

cns/service/main.go
rajvinar: sure

cns/service/main.go
rajvinar: sure

cns/service/main.go
rajvinar: sure.. I cleaned and kept only imp logs with imp information

cns/service/main.go
rajvinar: sure

cns/service/main.go
```diff
@@ -1245,3 +1249,112 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+	if err != nil {
+		log.Printf("Error while executing pwsh command to get VLAN Tag advance property of %s: %v", adapatername, err)
+		return
+	}
+	if adapterNamewithVLANTag != "" {
+		log.Printf("Found %s in adapter's advanced properties", priorityVLANTagIdentifier)
+
+		//set reg value for Priorityvlantag of adapater to 3
```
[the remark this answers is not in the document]
rajvinar: adding comment:
// reg key value for PriorityVLANTag = 3  --> Packet priority and VLAN enabled
// for more details goto https://learn.microsoft.com/en-us/windows-hardware/drivers/network/standardized-inf-keywords-for-ndis-qos

thatmattlong:
Does mellanox driver require reboot? I agree with @rbtr that we probably shouldn't be setting this in CNS, but if it requires a reboot then we definitely **cannot** set this with CNS.

rajvinar:
> This seems like underlying infra config, and I'm not sure it should be CNS's job to configure prerequisites like this. Shouldn't this responsibility lie with the orchestrator which expects to use CNS/CNI?

This script/code to set PriorityVLANTag already been added to Atlas code (https://msazure.visualstudio.com/One/_git/SF-SeaBreeze?path=%2Fsrc%2Fscripts%2FAtlasVMSetup.20220112.ps1&_a=contents&version=GBmain) but it seems it gets executed at starting of cluster but eventually value get reset so we want to make it part of CNS to make sure mellanox adapter's value is always set

rajvinar:
> Does mellanox driver require reboot? I agree with @rbtr that we probably shouldn't be setting this in CNS, but if it requires a reboot then we definitely **cannot** set this with CNS.

yes for version 3 it requires reboot. Let me confirm with Paul again. May I know why we cannot execute this in CNS if it requires reboot. Just curious.

rajvinar:
> @microsoft-github-policy-service agree [company="{your company}"]

@microsoft-github-policy-service agree [company="Microsoft"]

rajvinar:
> @microsoft-github-policy-service agree company="Microsoft"

pjohnst5:
> This seems like underlying infra config, and I'm not sure it should be CNS's job to configure prerequisites like this. Shouldn't this responsibility lie with the orchestrator which expects to use CNS/CNI?

That's a great point, we have tried this, but the setting gets un-set eventually, either due to live migration of the node underneath the VM, or otherwise, so we can't rely on it being set once and done, we'll have to keep setting it in CNS to make sure

pjohnst5:
And VM hosting team is unwilling / unable to set it to the values we need by default across Azure
Vivek Bhanu is his name

pjohnst5:
> Does mellanox driver require reboot? I agree with @rbtr that we probably shouldn't be setting this in CNS, but if it requires a reboot then we definitely **cannot** set this with CNS.

A reboot of the network adapter is needed on one older version Mellanox x3

The VM does not need to be rebooted

pjohnst5:
Hi @ashvindeodhar , I was thinking to put the setmellanoxadapter function call into the CreateorUpdateNetworkContainer
Reason being, we would hit the same issue here, of live migration (or any other reason that the priorityvlantag gets reset by something out of our stack), failing the datapath

If we set the mellanox adapter each time, then there's a chance when the priorityvlantag is reset by some live-migration or anything else, createNC call would fix it right away, instead of restarting CNS

platform/os_windows.go
```diff
@@ -189,6 +203,96 @@ func SetSdnRemoteArpMacAddress() error {
...
+		cmd = fmt.Sprintf(" Get-NetAdapterAdvancedProperty | Where-Object { $_.RegistryKeyword -like \"%s\" -and $_.Name -eq \"%s\" } | Select-Object -ExpandProperty RegistryValue", priorityVLANTagIdentifier, adapatername)
+		regvalue, err := ExecutePowershellCommand(cmd)
+		if err != nil {
+			return fmt.Errorf("error while checking registery value for PriorityVLANTag for adapater: %w", err)
+		}
+		if regvalue != fmt.Sprintf("%d", regvalueForVLANTag) {
+			return fmt.Errorf("failure while setting Mellanox Network Adapter: \"%s\" with \"%s\" property value to %d", adapatername, priorityVLANTagIdentifier, regvalueForVLANTag)
+		}
+		log.Printf("Successfully set Mellanox Network Adapter: \"%s\" with \"%s\" property value as %d", adapatername, priorityVLANTagIdentifier, regvalueForVLANTag)
+	} else {
```
ashvindeodhar, line 248: remove this else by adding return after line 247.

platform/os_windows.go
```diff
@@ -189,6 +203,96 @@ func SetSdnRemoteArpMacAddress() error {
...
+		//Adapter is version 3 or less as PriorityVLANTag was not found in advanced properties of mellanox adpater
+		log.Printf("Searching through CIM instances for Network devices with \"%s\" in the name", mellanoxSearchString)
+		cmd = fmt.Sprintf("Get-CimInstance -Namespace root/cimv2 -ClassName Win32_PNPEntity | Where-Object PNPClass -EQ \"Net\" | Where-Object { $_.Name -like \"%s\" } | Select-Object -ExpandProperty DeviceID", mellanoxSearchString)
+		deviceid, err := ExecutePowershellCommand(cmd)
+
```
ashvindeodhar, line 253: can you look at other files for consistent coding style? I see in some places you have spaces before if, in other places, there are no spaces

platform/os_windows.go
```diff
@@ -189,6 +203,96 @@ func SetSdnRemoteArpMacAddress() error {
...
+		if err != nil {
+			return fmt.Errorf("error while executing powershell command to get device id of %s: %w", adapatername, err)
+		}
+		if deviceid == "" {
+			log.Printf("No network device found with \"%s\" in description", mellanoxSearchString)
+			return nil
```
ashvindeodhar, line 259: is this not an error?

platform/os_windows.go
```diff
@@ -189,6 +203,96 @@ func SetSdnRemoteArpMacAddress() error {
...
+		}
+
+		log.Printf("Devcie ID found and Getting PNP device properites for %s", deviceid)
+		cmd = fmt.Sprintf("Get-PnpDeviceProperty -InstanceId \"%s\" | Where-Object KeyName -EQ \"DEVPKEY_Device_Driver\" | Select-Object -ExpandProperty Data", deviceid)
+		registryKeySuffix, err := ExecutePowershellCommand(cmd)
+		if err != nil {
+			return fmt.Errorf("error while executing powershell command to get device id of %s: %w", adapatername, err)
```
ashvindeodhar, line 266: should this be deviceID or adapatername?

platform/os_windows.go
```diff
@@ -189,6 +203,96 @@ func SetSdnRemoteArpMacAddress() error {
...
+
+		if err != nil {
+			return fmt.Errorf("error while executing powershell command to get device id of %s: %w", adapatername, err)
+		}
+		if deviceid == "" {
+			log.Printf("No network device found with \"%s\" in description", mellanoxSearchString)
+			return nil
+		}
+
+		log.Printf("Devcie ID found and Getting PNP device properites for %s", deviceid)
```
ashvindeodhar, line 262: deviceID

platform/os_windows.go
ashvindeodhar, line 266: adapterName. Please check the var naming style in other places

platform/os_windows.go
```diff
@@ -189,6 +203,96 @@ func SetSdnRemoteArpMacAddress() error {
...
+		//Adapter is version 3 or less as PriorityVLANTag was not found in advanced properties of mellanox adpater
+		log.Printf("Searching through CIM instances for Network devices with \"%s\" in the name", mellanoxSearchString)
+		cmd = fmt.Sprintf("Get-CimInstance -Namespace root/cimv2 -ClassName Win32_PNPEntity | Where-Object PNPClass -EQ \"Net\" | Where-Object { $_.Name -like \"%s\" } | Select-Object -ExpandProperty DeviceID", mellanoxSearchString)
+		deviceid, err := ExecutePowershellCommand(cmd)
+
+		if err != nil {
+			return fmt.Errorf("error while executing powershell command to get device id of %s: %w", adapatername, err)
+		}
+		if deviceid == "" {
+			log.Printf("No network device found with \"%s\" in description", mellanoxSearchString)
```
ashvindeodhar, line 258: remove \\" when printing string in the log throughout this file.

ashvindeodhar (dismissed):

ashvindeodhar:
> Hi @ashvindeodhar , I was thinking to put the setmellanoxadapter function call into the CreateorUpdateNetworkContainer Reason being, we would hit the same issue here, of live migration (or any other reason that the priorityvlantag gets reset by something out of our stack), failing the datapath
> 
> If we set the mellanox adapter each time, then there's a chance when the priorityvlantag is reset by some live-migration or anything else, createNC call would fix it right away, instead of restarting CNS

@pjohnst5 yes this is why I had added the sdn arp mac reg key at the time of getNC call to CNS. https://github.com/Azure/azure-container-networking/commit/a1f804a912f0f1975fd82a574f74d7afbae1fc83 @tamilmani1989 any idea why that was moved to main func in CNS?

cns/service/main.go
```diff
@@ -667,6 +667,13 @@ func main() {
 		return
 	}
 
+	err = platform.SetMellanoxRegKeyPriorityVLANTag()
+	if err != nil {
+		logger.Errorf("Failed to set reg key for mellanox adapter: %v", err)
+	} else {
+		logger.Printf("Successfuly set reg key for existing Mellanox adapter.")
```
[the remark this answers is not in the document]
rajvinar: actually, it will return the error in this case... 
`if adapatername == "" {
		return fmt.Errorf("no network adapter found with \"%s\" in description", mellanoxSearchString)
}`

cns/service/main.go
```diff
@@ -667,6 +667,13 @@ func main() {
 		return
 	}
 
+	err = platform.SetMellanoxRegKeyPriorityVLANTag()
```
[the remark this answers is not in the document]
rajvinar: sure will do, also Paul suggested to place this call in CreateorUpdateNetworkContainer, reason stated above in Paul's comments, can we move there?

platform/os_windows.go
```diff
@@ -65,6 +65,20 @@ const (
...
+	// PriorityVlanTag reg key for adapter
+	priorityVLANTagIdentifier = "*PriorityVLANTag"
+
+	// Registry key Path Prefix
+	registryKeyPrefix = "HKLM:\\System\\CurrentControlSet\\Control\\Class\\"
+
+	//Value for reg key: PriorityVLANTag for adapter
+	// reg key value for PriorityVLANTag = 3  --> Packet priority and VLAN enabled
+	// for more details goto https://learn.microsoft.com/en-us/windows-hardware/drivers/network/standardized-inf-keywords-for-ndis-qos
+	regvalueForVLANTag = 3
```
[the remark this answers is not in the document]
rajvinar: Ok implementing.

platform/os_windows.go
```diff
@@ -189,6 +203,96 @@ func SetSdnRemoteArpMacAddress() error {
...
 }
 
+// Set Mellanox adapter's PriorityVLANTag value to 3 if adapter exists
+// reg key value for PriorityVLANTag = 3  --> Packet priority and VLAN enabled
+// for more details goto https://learn.microsoft.com/en-us/windows-hardware/drivers/network/standardized-inf-keywords-for-ndis-qos
+func SetMellanoxRegKeyPriorityVLANTag() error {
+	//get mellanox adapter name
+	cmd := fmt.Sprintf("Get-NetAdapter | Where-Object { $_.InterfaceDescription -like \"%s\" } | Select-Object -ExpandProperty Name", mellanoxSearchString)
+	adapatername, err := ExecutePowershellCommand(cmd)
+
```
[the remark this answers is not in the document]
rajvinar: sure.

platform/os_windows.go
rajvinar: yes I should make it an error and return that error for logging in

platform/os_windows.go
rajvinar: yes device id and also message changed now.. since it was to get registry suffix for device id

platform/os_windows.go
rajvinar: yes thank you.. correcting it

tamilmani1989:
Does setting this registry key affect singletenancy scenarios? I see its set for all scenarios.

tamilmani1989:
The downside of setting this in CNI and not by orchestrator bootup script is that if in future something changes like registry value changes for different version of Mellanox, then we have to deal with supporting all versions based on what used.

tamilmani1989:
> > Hi @ashvindeodhar , I was thinking to put the setmellanoxadapter function call into the CreateorUpdateNetworkContainer Reason being, we would hit the same issue here, of live migration (or any other reason that the priorityvlantag gets reset by something out of our stack), failing the datapath
> > If we set the mellanox adapter each time, then there's a chance when the priorityvlantag is reset by some live-migration or anything else, createNC call would fix it right away, instead of restarting CNS
> 
> @pjohnst5 yes this is why I had added the sdn arp mac reg key at the time of getNC call to CNS. [a1f804a](https://github.com/Azure/azure-container-networking/commit/a1f804a912f0f1975fd82a574f74d7afbae1fc83) @tamilmani1989 any idea why that was moved to main func in CNS?

putting in createorupdatenc indirectly dependent on CNI to execute sequentially. Now that we removed lock acquiring at beginning of cni process, it can execute parallel which may call setting registry key multiple times which leads to hns restarts multiple times

platform/os_windows.go
rajvinar: sure

rbtr:
I remain opposed to this kind of underlying infrastructure configuration being CNS's responsibility at all. 

However, if it were, it must be in a startup/preflight and guaranteed to only execute once per process. If the config becomes unset such that we can't do our job anymore, we should crash out so that it is reset during reinitialization.

> > This seems like underlying infra config, and I'm not sure it should be CNS's job to configure prerequisites like this. Shouldn't this responsibility lie with the orchestrator which expects to use CNS/CNI?
> 
> That's a great point, we have tried this, but the setting gets un-set eventually, either due to live migration of the node underneath the VM, or otherwise, so we can't rely on it being set once and done, we'll have to keep setting it in CNS to make sure

That some registry flags get unset during a live migration kinda sounds like a bug with live migration, no? And if something else is randomly resetting the value, we need to understand who and why and ask them to stop, maybe? Otherwise we just fight over the reg key.

pjohnst5:
Rajvi is following up with Vivek Bhanu and Michael Nanakul as to why the registry key is reset in the first place

platform/os_windows.go
```diff
@@ -189,6 +206,170 @@ func SetSdnRemoteArpMacAddress() error {
...
+	}
+	log.Printf("Name of Mellanox adapter : %v", adapterName)
+	return true
+}
+
+// Regularly monitors the Mellanox PriorityVLANGTag registry value and sets it to desired value if needed
+func MonitorAndSetMellanoxRegKeyPriorityVLANTag() {
+	for {
+		adapterName, err := getMellanoxAdapterName()
+		if err == nil && adapterName != "" {
```
ramiro-gamarra, line 226: some general guidance: try to keep happy path code the least aligned (good explanation here: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88). this means try to return from functions or scope as early as possible.

platform/os_windows.go
```diff
@@ -189,6 +206,170 @@ func SetSdnRemoteArpMacAddress() error {
...
+			err := SetMellanoxPriorityVLANTag(adapterName)
+			if err != nil {
+				log.Errorf("error while monitoring and setting Mellanox Reg Key value: %v", err)
+			}
+			time.Sleep(mellanoxPriorityVLANTagMonitorInterval)
```
ramiro-gamarra, line 231: better to use a `time.Ticker` than sleeping in a loop

platform/os_windows.go
```diff
@@ -189,6 +206,170 @@ func SetSdnRemoteArpMacAddress() error {
 	return nil
 }
 
+func HasMellanoxAdapater() bool {
+	adapterName, err := getMellanoxAdapterName()
+	if err != nil {
+		log.Errorf("Error while getting mellanox adapter name: %v", err)
+		return false
+	} else if adapterName == "" {
```
ramiro-gamarra, line 214: if you return early, `else` statements become unnecessary.

platform/os_windows.go
```diff
@@ -189,6 +206,170 @@ func SetSdnRemoteArpMacAddress() error {
...
+func getMellanoxAdapterName() (string, error) {
+	//get mellanox adapter name
+	cmd := fmt.Sprintf("Get-NetAdapter | Where-Object { $_.InterfaceDescription -like \"%s\" } | Select-Object -ExpandProperty Name", mellanoxSearchString)
+	adapterName, err := ExecutePowershellCommand(cmd)
+	if err != nil {
+		return "", fmt.Errorf("error while executing powershell command to get net adapter list: %w", err)
+	}
+	if adapterName == "" {
+		log.Printf("no network adapter found with %s in description", mellanoxSearchString)
+	}
```
ramiro-gamarra, line 247: return an error if adapter is not found

platform/os_windows.go
```diff
@@ -189,6 +206,170 @@ func SetSdnRemoteArpMacAddress() error {
...
+// for more details goto https://docs.nvidia.com/networking/display/winof2v230/Configuring+the+Driver+Registry+Keys#ConfiguringtheDriverRegistryKeys-GeneralRegistryKeysGeneralRegistryKeys
+func SetMellanoxPriorityVLANTag(adapterName string) error {
+	//Find if adapter has property PriorityVLANTag (version 4 or up) or not (version 3)
+	cmd := fmt.Sprintf("Get-NetAdapterAdvancedProperty | Where-Object { $_.RegistryKeyword -like \"%s\" -and $_.Name -eq \"%s\" } | Select-Object -ExpandProperty Name", priorityVLANTagIdentifier, adapterName)
+	adapterNameWithVLANTag, err := ExecutePowershellCommand(cmd)
+	if err != nil {
+		return fmt.Errorf("error while executing powershell command to get VLAN Tag advance property of %s: %w", adapterName, err)
+	}
+
+	if adapterNameWithVLANTag != "" {
```
ramiro-gamarra, line 262: if you flip the condition and return early, most of this function will be unindented, which is more readable.

platform/os_windows.go
```diff
@@ -189,6 +206,170 @@ func SetSdnRemoteArpMacAddress() error {
...
+
+		log.Printf("Device ID found and Getting PNP device properites for %s", deviceid)
+		cmd = fmt.Sprintf("Get-PnpDeviceProperty -InstanceId \"%s\" | Where-Object KeyName -EQ \"DEVPKEY_Device_Driver\" | Select-Object -ExpandProperty Data", deviceid)
+		registryKeySuffix, err := ExecutePowershellCommand(cmd)
+		if err != nil {
+			return fmt.Errorf("error while executing powershell command to get registery suffix of device id %s: %w", deviceid, err)
+		}
+
+		registryKeyFullPath := registryKeyPrefix + registryKeySuffix
+		success, err := checkMellanoxPriorityVLANTagValueForV3(registryKeyFullPath, adapterName, regValueForVLANTag)
```
ramiro-gamarra, line 310: if this function is returning an error, is a `success` bool still necessary?

platform/os_windows.go
```diff
@@ -189,6 +206,170 @@ func SetSdnRemoteArpMacAddress() error {
...
+	}
+
+	return nil
+}
+
+// Checks if a Mellanox adapter's PriorityVLANTag value
+// for version 4 and up is set to the given expected value
+func checkMellanoxPriorityVLANTagValueForV4(adapterName string, expectedValue int) (bool, error) {
+	cmd := fmt.Sprintf(
+		"Get-NetAdapterAdvancedProperty | Where-Object { $_.RegistryKeyword -like \"%s\" -and $_.Name -eq \"%s\" } | Select-Object -ExpandProperty RegistryValue",
```
ramiro-gamarra, line 349: use string literals to avoid escape characters
```suggestion
		`Get-NetAdapterAdvancedProperty | Where-Object { $_.RegistryKeyword -like "%s" -and $_.Name -eq "%s" } | Select-Object -ExpandProperty RegistryValue`,
```

platform/os_windows.go
```diff
@@ -189,6 +206,170 @@ func SetSdnRemoteArpMacAddress() error {
...
+		priorityVLANTagIdentifier, adapterName)
+
+	regvalue, err := ExecutePowershellCommand(cmd)
+	if err != nil {
+		return false, err
+	}
+
+	return regvalue != fmt.Sprintf("%d", expectedValue), nil
```
ramiro-gamarra, line 357: this check seems odd. better to
- assert that the exec is returning a number, otherwise return an error
- return true only if the 2 ints match

ramiro-gamarra (commented):
Some early feedback, not exhaustive. Please feel free to set some time up if you'd like to discuss

platform/os_windows.go
rajvinar: sure let me check

platform/os_windows.go
rajvinar: I will make this change

platform/os_windows.go
rajvinar: this if condition branches off on Mellanox adapter version so we cannot return early even if we flip the condition. In both if and else, we attempt to set Mellanox version through different PowerShell commands.

platform/os_windows.go
rajvinar: sure will do

platform/os_windows.go
rajvinar: Yes because the bool denotes if the priority vlan tag value is set to the given expected value or not, whereas the error denotes the case when we were not able to read the value.

platform/os_windows.go
rajvinar: cool yes I implemented this change using "strconv.Atoi("")

rbtr:
@rajvinar are you not a member of our GitHub team? You will need to join that so that the workflows will run.

rajvinar:
> @rajvinar are you not a member of our GitHub team? You will need to join that so that the workflows will run.

ohh let me do that.. thanks for info

platform/os_windows.go
```diff
@@ -189,6 +208,188 @@ func SetSdnRemoteArpMacAddress() error {
 	return nil
 }
 
+func HasMellanoxAdapater() bool {
```
ramiro-gamarra, line 211:
```suggestion
func HasMellanoxAdapter() bool {
```

platform/os_windows.go
```diff
@@ -189,6 +208,188 @@ func SetSdnRemoteArpMacAddress() error {
...
+	interval := defaultMellanoxMonitorInterval
+	if intervalSecs > 0 {
+		interval = time.Duration(intervalSecs) * time.Second
+	}
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			log.Printf("context cancelled, stopping Mellanox Monitoring:", ctx.Err())
```
ramiro-gamarra, line 232:
```suggestion
			log.Printf("context cancelled, stopping Mellanox Monitoring: %v", ctx.Err())
```

platform/os_windows.go
```diff
@@ -189,6 +208,188 @@ func SetSdnRemoteArpMacAddress() error {
...
+	}
+	if deviceid == "" {
+		return fmt.Errorf("no network device found with %s in description", mellanoxSearchString)
+	}
+
+	log.Printf("Device ID found and Getting PNP device properites for %s", deviceid)
+	cmd = fmt.Sprintf(`Get-PnpDeviceProperty -InstanceId "%s" | Where-Object KeyName -EQ "DEVPKEY_Device_Driver" | Select-Object -ExpandProperty Data`, deviceid)
+	registryKeySuffix, err := ExecutePowershellCommand(cmd)
+	if err != nil {
+		return fmt.Errorf("error while executing powershell command to get registery suffix of device id %s: %w", deviceid, err)
```
ramiro-gamarra, line 360: type on `registry`. check on all other log lines as well

platform/os_windows.go
```diff
@@ -189,6 +208,188 @@ func SetSdnRemoteArpMacAddress() error {
...
+	}
+
+	registryKeyFullPath := registryKeyPrefix + registryKeySuffix
+
+	currentVLANTagValue, err := getMellanoxPriorityVLANTagValueForV3(registryKeyFullPath, adapterName)
+	if err != nil {
+		log.Errorf("Error while checking registery value for PriorityVLANTag for adapater, continuing to reset the value: %v", err)
```
ramiro-gamarra, line 367: typo on `adapter`. check on all other log lines as well

ramiro-gamarra, line 367: is there a reason to not continue on v4 if there is an error getting the value, but continue on v3?

platform/os_windows.go
rajvinar: Made changes to code to correct this typo

platform/os_windows.go
```diff
@@ -189,6 +208,188 @@ func SetSdnRemoteArpMacAddress() error {
...
+		interval = time.Duration(intervalSecs) * time.Second
+	}
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			log.Printf("context cancelled, stopping Mellanox Monitoring:", ctx.Err())
+			return
+		case <-ticker.C:
```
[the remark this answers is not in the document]
rajvinar: that's fine I think... we dont need to create or call anything outside of this ticker

platform/os_windows.go
rajvinar: yes corrected 3 lines where this typo was there.. thanks

platform/os_windows.go
rajvinar: fix typo in all comments where this typo was.

rajvinar: agree.. this should be return statement and should not log and continue... just like v4.. I am correcting this..

platform/os_windows.go
```diff
@@ -189,6 +208,188 @@ func SetSdnRemoteArpMacAddress() error {
...
+
+	log.Printf("Restarting Mellanox network adapter for regkey change to take effect")
+	cmd = fmt.Sprintf(`Restart-NetAdapter -Name "%s"`, adapterName)
+	_, err = ExecutePowershellCommand(cmd)
+	if err != nil {
+		return fmt.Errorf("error while executing powershell command to restart net adapter  %s: %w", adapterName, err)
+	}
+	log.Printf("Successfully restarted Mellanox network adapter")
+
+	log.Printf("For Mellanox CX-3 adapters, the reg key set to %d", desiredRegValueForVLANTag)
```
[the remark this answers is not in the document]
rajvinar: corrected

platform/windows/adapter/mellanox/mellanox.go
```diff
@@ -0,0 +1,233 @@
...
+	GetAdapaterName() (string, error)
+
+	// Get PriorityVLANTag returns PriorityVLANTag value for Mellanox Adapter (both version 3 and version 4)
+	GetPriorityVLANTag() (int, error)
+
+	// Set Mellanox adapter's PriorityVLANTag value to desired value if adapter exists
+	SetPriorityVLANTag(int) error
+}
+
+type MellanoxImpl struct{}
```
ramiro-gamarra, line 45: `Impl` is a bit of a naming anti-pattern in Go, because it implies things are created for the purpose of implementing an interface, therefore coupling them. Given the package, I think naming this type `Controller` seems appropriate. A few more things:
- The interface should be defined where it's used, not where it's implemented.
- Curious about the file structure. Can this package just be under `mellanox` instead of `platform/windows/adapter/mellanox`?

timraymond: Definitely agree with `Impl`, but I actually think the interface name is the problem here. What does it mean for a set of behavior (i.e. an interface) to be a "Mellanox"? I think instead, something like "NetworkAdapter" is better... things that claim to be an network adapter should be able to tell you their name and get/set priority VLAN tags. This struct should be "Mellanox" or "MellanoxAdapter".

platform/os_windows.go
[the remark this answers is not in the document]
rajvinar: Ok I added one time call before ticker starts

platform/windows/adapter/mellanox/mellanox.go
```diff
@@ -0,0 +1,233 @@
...
+var (
+	errorMellanoxAdapterNotFound = fmt.Errorf("no network adapter found with %s in description", mellanoxSearchString)
+	errorMellanoxDeviceNotFound  = fmt.Errorf("no network device found with %s in description", mellanoxSearchString)
+	errorPowershellNotFound      = fmt.Errorf("failed to find powershell executable")
+)
+
+type Mellanox interface {
+	// GetAdapter returns name of Mellanox adapter if found
+	// Must return errorMellanoxAdapterNotFound if adapter is not found or adapter name empty
+	GetAdapaterName() (string, error)
```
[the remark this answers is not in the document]
rajvinar: got it thanks

platform/windows/adapter/mellanox/mellanox.go
rajvinar: thanks @ramiro-gamarra and @timraymond for this feedback... makes sense.. so I will name Interface as NetworkAdapter and will move it under adapter/networkadapter.go and will rename implementation as Mellanox instead of MellanoxImpl which will make more sense.... 
and the reason I have file structure as windows/adapter... since I am specifically talking about windows here.. not sure if you have anyother suggestions for it....

pjohnst5:
I think this looks good, will wait for Ramiro to sign off on it as well before giving an official approval

ramiro-gamarra (dismissed):

pjohnst5 (changes requested):
Not really requesting changes, but want to block this until we have 1 more question cleared up

Question is, should we only run this, `if config.ChannelMode == cns.Direct` ?

Reason I ask is, this is only a known issue for multitenant orchestrators (using windows)

AKS doesn't need this change (unless I'm missing something, where they do have multitenant windows scenarios)

So perhaps better to only run this if cns is in direct mode, where we know the issue arises

pjohnst5:
Yeah so I confirmed with @thatmattlong offline that AKS doesn't use multitenant windows scenarios, so they don't need this change

So the only confirmed place we need this is if `config.ChannelMode == cns.Direct`, I think we should add that 'if'

ramiro-gamarra (approved):

PR #1777, opened 2023-01-30
Initial delete NC changes

smittal22:
**Reason for Change**:

To enable AZR support, we need to pass two extra params `azID` (uint) and `enableAZR` (boolean) as part of body to NMAgent. Since we are not passing body to NMAgent as of now, this PR adds a body to the request just like we have for publish NC.

Whenever NC has `enableAZR = true` in NC config, we will pass body like: `{"azID":<value>,"enableAZR":true}`
For already existing NC (or if NC has `enableAZR = false` in NC config), we will pass body like: `{"azID": 0,"enableAZR":false}`

**Issue Fixed**:

https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/16187805

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**: e2e flow has been tested by creating and deleting NCs with both `enableAZR` set to `true` and `false`


smittal22:
> eed to pass two extra params `azID` (uint) and `enableAZR` (boolean) as part of body to NMAgent. Since we are not passing body to NMAgent as of now, this PR adds a body to the request just like we have for publish NC.
> 
> Whenever NC has `enableAZR = true` in NC config, we will pass body like: `{"azID":<value>,"enableAZR":true}` For already existing NC (or if NC has `enableAZR = false` in NC config), we will pass body like: `{"azID": 0,"enableAZR":false}`
> 
> **Issue Fixed**:
> 
> https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/16187805

I'll probably open a new PR once the pipeline issue is resolved. Will take a look at this today.

PR #1784, opened 2023-02-03
Adding azure-ipam changes for dualstack

rjdenney:
**Reason for Change**:

Adds changes to the azure-ipam plugin to support dualstack integration with Cilium

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


azure-ipam/ipam.go
```diff
@@ -72,7 +73,7 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 	p.logger.Debug("Making request to CNS")
 	// if this fails, the caller plugin should execute again with cmdDel before returning error.
 	// https://www.cni.dev/docs/spec/#delegated-plugin-execution-procedure
-	resp, err := p.cnsClient.RequestIPAddress(context.TODO(), req)
+	resp, err := p.cnsClient.RequestIPs(context.TODO(), req)
```
ashvindeodhar, line 76: we need to call the old API if this fails

azure-ipam/ipam.go
```diff
@@ -70,12 +71,31 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
...
 	p.logger.Debug("Making request to CNS")
+	var e = "unsupported protocol scheme"
 	// if this fails, the caller plugin should execute again with cmdDel before returning error.
 	// https://www.cni.dev/docs/spec/#delegated-plugin-execution-procedure
-	resp, err := p.cnsClient.RequestIPAddress(context.TODO(), req)
+	resp, err := p.cnsClient.RequestIPs(context.TODO(), req)
 	if err != nil {
-		p.logger.Error("Failed to request IP address from CNS", zap.Error(err), zap.Any("request", req))
-		return cniTypes.NewError(ErrRequestIPConfigFromCNS, err.Error(), "failed to request IP address from CNS")
+		// if we fail a request with a 404 error try using the old API
```
rjdenney, line 79: This is where I am trying to do the 404

azure-ipam/ipam.go
```diff
@@ -29,7 +29,8 @@ type IPAMPlugin struct {
 
 type cnsClient interface {
 	RequestIPAddress(context.Context, cns.IPConfigRequest) (*cns.IPConfigResponse, error)
```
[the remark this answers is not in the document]
rjdenney: I think this would make more sense with the current naming conventions. The only reason I didn't was because they would also look very similar and could confuse people. I'm fine with either way though

Are we also planning on removing the `RequestIAddress` function at some point?

azure-ipam/ipam.go
```diff
@@ -72,10 +74,29 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 	p.logger.Debug("Making request to CNS")
 	// if this fails, the caller plugin should execute again with cmdDel before returning error.
 	// https://www.cni.dev/docs/spec/#delegated-plugin-execution-procedure
-	resp, err := p.cnsClient.RequestIPAddress(context.TODO(), req)
+	resp, err := p.cnsClient.RequestIPs(context.TODO(), req)
 	if err != nil {
-		p.logger.Error("Failed to request IP address from CNS", zap.Error(err), zap.Any("request", req))
-		return cniTypes.NewError(ErrRequestIPConfigFromCNS, err.Error(), "failed to request IP address from CNS")
+		// if we fail a request with a 404 error try using the old API
+		p.logger.Error("Failed to request IPs using RequestIPs from CNS, going to try RequestIPAddress", zap.Error(err), zap.Any("request", req))
```
thatmattlong, line 80: I think this log line belongs inside the `if` clause that starts on the next line

azure-ipam/ipam.go
```diff
@@ -85,17 +106,25 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
...
-			},
-		},
+	for _, IPNet := range *podIPNet {
+		p.logger.Debug("Parsed pod IP", zap.String("podIPNet", IPNet.String()))
+	}
+
+	cniResult := &types100.Result{}
+	for _, IPNet := range *podIPNet {
+		ipConfig := &types100.IPConfig{}
+		if net.ParseIP(IPNet.Addr().String()).To4() != nil {
```
thatmattlong, line 116: I would probably use `if IPNet.Addr().Is4() {` instead of parsing and checking for `nil`

azure-ipam/ipam.go
```diff
@@ -85,17 +106,25 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
...
-					Mask: net.CIDRMask(podIPNet.Bits(), 32), // nolint
-				},
-			},
-		},
+	for _, IPNet := range *podIPNet {
+		p.logger.Debug("Parsed pod IP", zap.String("podIPNet", IPNet.String()))
+	}
+
+	cniResult := &types100.Result{}
+	for _, IPNet := range *podIPNet {
```
thatmattlong, line 114:
```suggestion
	for _, ipNet := range *podIPNet {
```
I don't care as much what this variable is named as long as it is sane, but we shouldn't export it (aka shouldn't capitalize it)

azure-ipam/ipam.go
```diff
@@ -85,17 +106,25 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
...
-		IPs: []*types100.IPConfig{
-			{
-				Address: net.IPNet{
-					IP:   net.ParseIP(podIPNet.Addr().String()),
-					Mask: net.CIDRMask(podIPNet.Bits(), 32), // nolint
-				},
-			},
-		},
+	for _, IPNet := range *podIPNet {
+		p.logger.Debug("Parsed pod IP", zap.String("podIPNet", IPNet.String()))
```
thatmattlong, line 110: this should be in the loop starting on L114 and this loop should be removed - no need to iterate over the same thing twice esp. for debug logs only

azure-ipam/ipam.go
```diff
@@ -85,17 +106,25 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
...
+				IP:   net.ParseIP(IPNet.Addr().String()),
+				Mask: net.CIDRMask(IPNet.Bits(), 32), // nolint
+			}
+		} else {
+			ipConfig.Address = net.IPNet{
+				IP:   net.ParseIP(IPNet.Addr().String()),
+				Mask: net.CIDRMask(IPNet.Bits(), 128), // nolint
+			}
+		}
+		cniResult.IPs = append(cniResult.IPs, ipConfig)
```
thatmattlong, line 127: instead of appending, if you know the length of your slice ahead of time (which I think you do in this case), you can just `make` the slice with the correct length first and insert directly into the slice via indexing. Since this slice will only ever be lengths 0, 1, or 2, it's not a big deal, but something to get into the habit of.

azure-ipam/ipam.go
rjdenney: Moved it inside so that it now only prints when we get 404

azure-ipam/ipam.go
rjdenney: updated the if statement

azure-ipam/ipam.go
rjdenney: I never made the exact connection that capitalization was reserved for variables being exported from another package. I'll make sure to keep this in mind and updated it.

azure-ipam/ipam.go
rjdenney: Removed the loop and moved the statement to the start of the following loop.

azure-ipam/ipam.go
rjdenney: Added a `make(...)` for `cniResult.IPs`. I'm guessing the purpose of this is to strictly claim the exact amount of memory you need instead of having a slice with an undefined length.

azure-ipam/ipam.go
rjdenney: The 404 check is below

PR #1787, opened 2023-02-06
reset the http status code to 500 in case we encounter an error

smittal22:
**Reason for Change**:

During NC publish and unpublish, we do not reset the httpStatusCode that we return back to DNC. As a result, even if we encounter an error we send back HTTP 200 back to DNC

**Issue Fixed**:

https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/17035213

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -1239,6 +1239,10 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
 		returnCode = types.UnsupportedVerb
 	}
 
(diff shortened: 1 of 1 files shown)


ashvindeodhar (dismissed):

smittal22:
Since we do not want CNS to make any decisions and just pass the encoded response from NMA to DNC, we should not handle this bug in the manner this PR does. Closing this PR. It will be handled here: #1799

PR #1789, opened 2023-02-07
[NOT READY] windows multitenancy pipeline

paulyufan:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


Makefile
```diff
@@ -295,6 +295,15 @@ acncli-image: ## build cni-manager container image.
 		IMAGE=$(ACNCLI_IMAGE) \
 		TAG=$(ACNCLI_PLATFORM_TAG)
 
+acncli-image-windows: ## build the cni image for windows
```
rbtr, line 298: probably better to test using dropgz in windows flavor, as that's what we will use in production

cni/build/cni.windows.DockerFile
```diff
@@ -0,0 +1,29 @@
...
+
+SHELL ["cmd", "/S", "/C"]
+## build azure cni for windows
+WORKDIR /azure-container-networking/cni/network/plugin/
+RUN go build -a -o azure-vnet.exe -ldflags "-X main.version=%VERSION%"
+
+WORKDIR /azure-container-networking
+COPY /cni/scripts/installcni.ps1 .
+
+ENTRYPOINT ["powershell.exe", ".\\installcni.ps1"]
```
rbtr, line 29: pls configure your editor to add newlines at the end of the file on save

rbtr (dismissed):

Makefile
pjohnst5: I did just peek at `dropgz`, Paul and I don't know quite how it works, could we meet to talk about how a windows flavor would work?

pjohnst5 (commented):
Can you add new lines Paul Yu like Evan said?

Makefile
rbtr: We're actually trying to move conflist responsibilities to CNS - CNS is the component which knows when the Network is actually ready, and this has implications for NodeReady and Pod SLAs in AKS since the presence of the conflist is used to indicate NodeNetworkReady

pjohnst5: That's neat, okay, so we'd need to make one more generator in cns's conflist generator, and set the config flag to true, awsome!

dropgz/build/cni.windows.Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM mcr.microsoft.com/windows/servercore:ltsc2019 as Builder
```
rbtr, line 1: this is great ❤️

dropgz/build/cni.windows.Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.19 AS dropgz
+ARG VERSION
+WORKDIR /dropgz
+COPY --from=compressor /dropgz .
+RUN CGO_ENABLED=0 go build -a -o bin/dropgz -trimpath -ldflags "-X github.com/Azure/azure-container-networking/dropgz/internal/buildinfo.Version="$VERSION"" -gcflags="-dwarflocationlists=true" main.go
+
+FROM scratch
+COPY --from=dropgz /dropgz/bin/dropgz /dropgz
+ENTRYPOINT [ "/dropgz" ]
```
rbtr, line 31: but you still don't have newlines at the end of the files 🙁

timraymond: @paulyufan You should set your editor to do this so you never have to think about it again. That's what I do :)

rbtr (changes requested):
I've done some rework of the multiplat image pipeline in https://github.com/Azure/azure-container-networking/pull/1820 and I think we should wait for it and update this PR to take advantage of the improvements it has

rbtr:
@paulyufan #1820 merged, can you update this to follow the patterns established there for multiplatform Windows images?

paulyufan2:
DualStack CNS and CNI codes are merged to master branch. Close this PR.

PR #1791, opened 2023-02-08
fix: multitenantnetwork reconciler should check errors correctly

thatmattlong:
**Reason for Change**:

Fixes a bug in multitenantreonciler that has been present since 1.4.13, preventing upgrade to latest CNS version. The bug is caused by comparing an error string with `strings.EqualFold`, where the error string is `response: UnknownContainerID` but the string comparison checks for `UnknownContainerID`. Since we have appropriately typed errors now, we can check the response code directly instead of doing a string comparison.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


cns/multitenantcontroller/mockclients/Makefile
```diff
@@ -5,7 +5,7 @@ MOCKGEN = $(TOOLS_BIN_DIR)/mockgen
 
 generate: $(MOCKGEN) ## Generate mock clients
 	$(MOCKGEN) -source=$(REPO_ROOT)/cns/multitenantcontroller/multitenantoperator/multitenantcrdreconciler.go -package=mockclients cnsRESTservice > cnsrestservice_generated.go
-	$(MOCKGEN) -source=$(REPO_ROOT)/vendor/sigs.k8s.io/controller-runtime/pkg/client/interfaces.go -package=mockclients Client > kubeclient_generated.go
+	$(MOCKGEN) -source=$(REPO_ROOT)/vendor/sigs.k8s.io/controller-runtime/pkg/client/interfaces.go -package=mockclients Client,SubResourceWriter > kubeclient_generated.go
```
rbtr, line 8: can you provide some context on this?

rbtr, line 8: can you provide some context on this?

rbtr (dismissed):

cns/multitenantcontroller/mockclients/Makefile
thatmattlong: yes, previously our tests only used a mock of the `client.Client` but now we also need a mock of the `client.Status()`'s `client.SubResourceWriter`. So this is just being explicit about which symbols we want to generate.

rbtr: sounds good ty

cns/multitenantcontroller/multitenantoperator/multitenantcrdreconciler_test.go
```diff
@@ -168,7 +232,7 @@ var _ = Describe("multiTenantCrdReconciler", func() {
 				NamespacedName: namespacedName,
 			})
 			Expect(err).NotTo(BeNil())
-			Expect(err.Error()).To(ContainSubstring("UnknownContainerID"))
```
neaggarwMS, line 171: oh wow we were validating the unexpected error code too

neaggarwMS (approved):

rbtr (approved):

PR #1792, opened 2023-02-08
Update install-cni-plugin.sh

jayunit100:
disambiguate cni version from plugin version , since cni version implies something totally different

### DIFF
--- a/scripts/install-cni-plugin.sh
+++ b/scripts/install-cni-plugin.sh
@@ -11,7 +11,8 @@ CNI_NETCONF_DIR=/etc/cni/net.d
 function usage
 {
     printf "Installs azure-vnet CNI plugins.\n"
-    printf "Usage: install-cni-plugin version [cniVersion]\n"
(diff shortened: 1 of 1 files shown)


rbtr (approved):

scripts/install-cni-plugin.sh
```diff
@@ -11,7 +11,8 @@ CNI_NETCONF_DIR=/etc/cni/net.d
 function usage
 {
     printf "Installs azure-vnet CNI plugins.\n"
-    printf "Usage: install-cni-plugin version [cniVersion]\n"
+    printf "See https://github.com/Azure/azure-container-networking releases for versions"
```
rbtr, line 14:
```suggestion
    printf "See https://github.com/Azure/azure-container-networking/releases for versions\n"
```

rbtr (changes requested):

rbtr (approved):

PR #1800, opened 2023-02-10
fix: Adding a defer func to connecttoTelemetryservice() to prevent CNI fro…

AzureAhai:
Adding a defer func to connecttoTelemetryservice() 

**Reason for Change**:
to prevent CNI from stucking when the telemetry service is unable to start.

### DIFF
--- a/telemetry/telemetrybuffer.go
+++ b/telemetry/telemetrybuffer.go
@@ -20,7 +20,6 @@ import (
 	"github.com/Azure/azure-container-networking/common"
 	"github.com/Azure/azure-container-networking/log"
 	"github.com/Azure/azure-container-networking/platform"
-	"github.com/pkg/errors"
 )
 
 // TelemetryConfig - telemetry config read by telemetry service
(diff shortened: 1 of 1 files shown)


telemetry/telemetrybuffer.go
```diff
@@ -328,32 +328,42 @@ func (tb *TelemetryBuffer) ConnectCNIToTelemetryService(telemetryNumRetries, tel
...
+	}
+	return nil
+}
+
+// This function is getting called from ConnectCNIToTelemetryService() in each attempt inside for loop
+// This function has been created to be able to add defer within the for loop
+func (tb *TelemetryBuffer) ConnectCNIToTelemetryServiceAttempt(telemetryNumRetries, telemetryWaitTimeInMilliseconds int, netPlugin *cni.Plugin, path string, args []string) error {
+	if err := tb.Connect(); err != nil {
+		log.Logf("Connection to telemetry socket failed: %v", err)
+		if runtime.GOOS == "windows" {
```
tamilmani1989, line 343: can you also place defer under this if..so no need to check if windows in 2 places

tamilmani1989 (changes requested):

telemetry/telemetrybuffer.go
```diff
@@ -328,32 +328,42 @@ func (tb *TelemetryBuffer) ConnectCNIToTelemetryService(telemetryNumRetries, tel
...
-					log.Logf("lock acquire error: %v", err)
-					return errors.Wrap(err, "lock acquire error")
-				}
-			}
-			if err = tb.Cleanup(FdName); err != nil {
-				return errors.Wrap(err, "cleanup failed")
-			}
-			if err = StartTelemetryService(path, args); err != nil {
-				return errors.Wrap(err, "StartTelemetryService failed")
+		if err := tb.ConnectCNIToTelemetryServiceAttempt(telemetryNumRetries, telemetryWaitTimeInMilliseconds, netPlugin, path, args); err != nil {
```
tamilmani1989, line 331: can we rename this as `startAndConnectTelemetryService`

tamilmani1989 (changes requested):

telemetry/telemetrybuffer.go
AzureAhai: addressed in the new commit.

telemetry/telemetrybuffer.go
AzureAhai: addressed in the new commit

tamilmani1989 (approved):
lgtm

tamilmani1989 (approved):

PR #1804, opened 2023-02-14
fix: [NPM] update endpointcache after remove policy

ck319:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/npm/pkg/dataplane/dataplane.go
+++ b/npm/pkg/dataplane/dataplane.go
@@ -294,7 +294,7 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
 		klog.Infof("[DataPlane] Policy %s is not found. Might been deleted already", policyKey)
 		return nil
 	}
-	dp.updatePodCache.Lock()
+	dp.endpointCache.Lock()
 
 	// Use the endpoint list saved in cache for this network policy to remove
 	err := dp.policyMgr.RemovePolicy(policy.PolicyKey)
@@ -312,7 +312,7 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
 		}
 	}
 
-	dp.updatePodCache.Unlock()
+	dp.endpointCache.Unlock()
 
 	// Remove references for Rule IPSets first
 	err = dp.deleteIPSetsAndReferences(policy.RuleIPSets, policy.PolicyKey, ipsets.NetPolType)
--- a/npm/pkg/dataplane/dataplane.go
+++ b/npm/pkg/dataplane/dataplane.go
@@ -286,8 +286,8 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
 	policy, ok := dp.policyMgr.GetPolicy(policyKey)
 	endpoints := make(map[string]string, len(policy.PodEndpoints))
 
-	for podIP, endpointId := range policy.PodEndpoints {
-		endpoints[podIP] = endpointId
+	for podIP, endpointID := range policy.PodEndpoints {
+		endpoints[podIP] = endpointID
 	}
 
 	if !ok {
@@ -305,7 +305,7 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
 	for podIP := range endpoints {
 		// if the endpoint is not in the policy's endpoint list, delete policy reference from cache
 		if _, ok := policy.PodEndpoints[podIP]; !ok {
-			//check if the endpoint is in the cache
+			// check if the endpoint is in the cache
 			if endpoint, ok := dp.endpointCache.cache[podIP]; ok {
 				delete(endpoint.netPolReference, policyKey)
 			}
--- a/npm/pkg/dataplane/dataplane.go
+++ b/npm/pkg/dataplane/dataplane.go
@@ -284,12 +284,16 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
 	// because policy Manager will remove from policy from cache
 	// keep a local copy to remove references for ipsets
 	policy, ok := dp.policyMgr.GetPolicy(policyKey)
-	endpoints := policy.PodEndpoints
+	endpoints := make(map[string]string, len(policy.PodEndpoints))
+	
+	for podIP,endpointId := range policy.PodEndpoints {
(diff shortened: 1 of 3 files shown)


npm/pkg/dataplane/dataplane.go
```diff
@@ -284,15 +284,29 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
 	// because policy Manager will remove from policy from cache
 	// keep a local copy to remove references for ipsets
 	policy, ok := dp.policyMgr.GetPolicy(policyKey)
+	endpoints := policy.PodEndpoints
```
huntergregory, line 287: will need to make a deep copy

npm/pkg/dataplane/dataplane.go
```diff
@@ -284,15 +284,29 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
...
 	// Use the endpoint list saved in cache for this network policy to remove
 	err := dp.policyMgr.RemovePolicy(policy.PolicyKey)
 	if err != nil {
 		return fmt.Errorf("[DataPlane] error while removing policy: %w", err)
 	}
+
+	for npmEndpoint := range endpoints {
+		// if the endpoint is not in the policy's endpoint list, delete from cache
+		if _, ok := policy.PodEndpoints[npmEndpoint]; !ok {
+			delete(dp.endpointCache.cache, npmEndpoint)
```
huntergregory, line 304: general structure is good, but we're looking to delete the policy reference from the given endpoint the endpoint cache is also indexed by IP

The endpoint cache is also indexed by IP. If Policy.Endpoints doesn't have an IP, then the lookup may not be efficient but that's ok

huntergregory (changes requested):
noting changes from other day's discussion offline

npm/pkg/dataplane/dataplane.go
```diff
@@ -284,15 +284,34 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
...
 	err := dp.policyMgr.RemovePolicy(policy.PolicyKey)
 	if err != nil {
 		return fmt.Errorf("[DataPlane] error while removing policy: %w", err)
 	}
+
+	for podIP := range endpoints {
+		// if the endpoint is not in the policy's endpoint list, delete policy reference from cache
+		if _, ok := policy.PodEndpoints[podIP]; !ok {
+			endpoint := dp.endpointCache.cache[podIP]
+			delete(endpoint.netPolReference, policyKey)
```
huntergregory, line 309: endpoint may not exist (chance of nil pointer exception)

huntergregory (changes requested):

npm/pkg/dataplane/dataplane.go
ck319: updated

npm/pkg/dataplane/dataplane.go
ck319: updated

npm/pkg/dataplane/dataplane.go
ck319: updated

npm/pkg/dataplane/dataplane.go
```diff
@@ -284,15 +284,36 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
...
+
+	for podIP, endpointID := range policy.PodEndpoints {
+		endpoints[podIP] = endpointID
+	}
+
 	if !ok {
 		klog.Infof("[DataPlane] Policy %s is not found. Might been deleted already", policyKey)
 		return nil
 	}
+	dp.updatePodCache.Lock()
```
huntergregory, line 297: lock the endpoint cache instead

huntergregory (changes requested):
lgtm after locking endpointCache instead of updatePodCache

npm/pkg/dataplane/dataplane.go
ck319: updated

huntergregory (approved):

npm/pkg/dataplane/dataplane.go
```diff
@@ -284,15 +284,36 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
...
+
+	for podIP, endpointID := range policy.PodEndpoints {
+		endpoints[podIP] = endpointID
+	}
+
 	if !ok {
 		klog.Infof("[DataPlane] Policy %s is not found. Might been deleted already", policyKey)
 		return nil
 	}
+	dp.endpointCache.Lock()
```
vakalapa, line 297: Why cant we take endpointCache after the RemovePolicy ? if a new ep is created or existing one is deleted we will be fine right ?

npm/pkg/dataplane/dataplane.go
```diff
@@ -284,15 +284,36 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
...
+
 	// Use the endpoint list saved in cache for this network policy to remove
 	err := dp.policyMgr.RemovePolicy(policy.PolicyKey)
 	if err != nil {
 		return fmt.Errorf("[DataPlane] error while removing policy: %w", err)
 	}
+
```
vakalapa, line 304: Please check dp.ShouldUpdatePod() or something which is always true for windows and then execute this section, this is unnecessary for linux.

npm/pkg/dataplane/dataplane.go
vakalapa: @huntergregory

huntergregory: discussed offline:
The goal of this was to not block NetPol controller on making SetPolicy SysCalls while UpdatePod() is running. 

New Proposal:
1. Put endpoint lock below removePolicy()
2. Address the blocking in another PR that makes an endpoint-level lock (instead of locking the whole cache)

ck319: updated 1

npm/pkg/dataplane/dataplane.go
ck319: updated

huntergregory (approved):

PR #1815, opened 2023-02-20
wip: replaced docker@2 tasks in build pipeline with buildx script

estebancams:
**Reason for Change**:

This replaces the way cns and npm windows amd64 images for e2e testing in PR pipeline are generated. Before we used docker build in windows, but there is an intermittent permissions issues in windows docker agent that we could not figure out with 1es and obp team.
This replaces docker build with docker buildx to produce windows images in Linux, where we don't have that issue.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/containers/container-template-windows.yaml
+++ b/.pipelines/containers/container-template-windows.yaml
@@ -15,29 +15,23 @@ steps:
     command: 'login'
     addPipelineData: false
 
-- task: Docker@2
+- script: |
+    set -x
+    docker version
+    echo tag: $TAG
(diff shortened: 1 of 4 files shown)


rbtr (changes requested):
The tooling is already parameterized such that it supports buildx, you don't need to add any of this.
Also:
> This replaces docker build with docker buildx to produce windows images in Linux, where we don't have that issue.

is a mischaracterization of this change - we previously used buildx on Windows, now we're trying buildx on Linux targeting Windows

PR #1817, opened 2023-02-21
WIP Fix publish to return 200/Success w/ 401 from NMA

timraymond:
For legacy reasons, the PublishStatusCode is actually the status code from Wireserver. In the situation where NMAgent returns a 401, it's expected that the PublishStatusCode is set to 200 and the code is a Success. The embedded status from NMAgent must be a 401.

This minimally makes the failing test pass, but the logic is convoluted and needs to be refactored.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


timraymond:
Superceded by #1825

PR #1819, opened 2023-02-23
Revert "fix: Adding a defer func to connecttoTelemetryservice() to prevent CNI fro…"

AzureAhai:
Revert:
We found race condition issues with the new CNI and IPAM lock fix.
The related PRs will get reverted until the new design roll out.


tamilmani1989 (approved):
lgtm

PR #1824, opened 2023-02-27
Revert replacing NMAgent client in CNS (#1643)

timraymond:
This reverts commit 9cc0b0ba3e60aea9986ad5bc365c07324edc5d17.

Performing these changes uncovered unknown dependencies that clients had on specific formations of responses from CNS. Until these can be properly documented and understood, the legacy implementation must remain.

A subsequent approach may instead introduce alternative endpoints with known properties that clients can migrate to piecemeal.


timraymond:
Lints will need to be bypassed for this, since it is, as much as possible, a pure revert. The goal is restoring the old behavior, warts and all. Lints can be noted and applied in a subsequent cleanup PR, but we should try to preserve the atomicity of this PR by adding to it as little as we can.

Also, beware looking at this patch through diff goggles. This code is not de novo logic--it already existed in the codebase prior to the revert. I may not be able to answer questions about why things are the way they are. It is also not as good as it could be, so we may need to suspend our better taste in favor of getting a clean release.

rbtr:
no objection as longs as the functional checks pass

timraymond:
It was decided that we will not revert, and instead add functionality in #1825 to try to repair the regressions. Closing.

PR #1826, opened 2023-02-28
Include body for unpublishNC calls to support AZR

smittal22:
**Reason for Change**:

To enable AZR support, we need to pass two extra params `azID (uint)` and `enableAZR (boolean)` as part of body to NMAgent. Since we are not passing body to NMAgent as of now, this PR adds a body to the request just like we have for publish NC.

Whenever NC has `enableAZR = true` in NC config, we will pass body like: `{"azID":<value>,"enableAZR":true}`
For already existing NC (or if NC has `enableAZR = false` in NC config), we will pass body like: `{"azID": 0,"enableAZR":false}`

**Issue Fixed**:

https://msazure.visualstudio.com/DefaultCollection/One/_workitems/edit/16187805

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:


cns/wireserver/proxy.go
```diff
@@ -57,10 +57,10 @@ func (p *Proxy) PublishNC(ctx context.Context, ncParams cns.NetworkContainerPara
 	return resp, nil
 }
 
-func (p *Proxy) UnpublishNC(ctx context.Context, ncParams cns.NetworkContainerParameters) (*http.Response, error) {
+func (p *Proxy) UnpublishNC(ctx context.Context, ncParams cns.NetworkContainerParameters, payload []byte) (*http.Response, error) {
 	reqURL := fmt.Sprintf(unpublishNCURLFmt, p.Host, ncParams.AssociatedInterfaceID, ncParams.NCID, ncParams.AuthToken)
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, reqURL, bytes.NewBufferString(`""`))
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, reqURL, bytes.NewBuffer(payload))
```
ramiro-gamarra, line 63: This change is not backwards compatible with DNC.

Current versions of DNC will not send base64 encoded bytes for the added fields, but this doesn't mean an error will occur during deserialization. Instead, the field will be considered `nil`, and a buffer with no bytes will be used here, instead of `""` which is used on purpose.

ramiro-gamarra (changes requested):
This change needs to be backwards compatible with DNC.

ramiro-gamarra (approved):

PR #1828, opened 2023-03-01
add func GetNumEndpointsInNetNs() to support to get number of endpoints for deletion

paulyufan:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


network/manager_mock.go
```diff
@@ -114,3 +114,17 @@ func (nm *MockNetworkManager) FindNetworkIDFromNetNs(netNs string) (string, erro
...
 	return "", errNetworkNotFound
 }
+
+// GetNumEndpointsInNetNs mock
+func (nm *MockNetworkManager) GetNumEndpointsInNetNs(netNs string) int {
+	// based on the GetAllEndpoints func above, it seems that this mock is only intended to be used with
+	// one network, so just return the number of endpoints if network exists
+	numEndpoints := 0
+	for _, network := range nm.TestNetworkInfoMap {
+		if _, err := nm.GetAllEndpoints(network.Id); err == nil {
```
ashvindeodhar, line 124: we should assume and return one endpoint per network. So this code should be - 
for _, network := range nm.TestNetworkInfoMap {
    numEndpoints++
}

network/network_test.go
```diff
@@ -257,4 +257,51 @@ var _ = Describe("Test Network", func() {
...
+		Context("When network exists", func() {
+			It("Should return two endpoints", func() {
+				netNs := "989c079b-45a6-485f-8f9e-88b05d6c55c5"
+				networkID := "byovnetbridge-vlan1-10-128-8-0_23"
+				nm := &networkManager{
+					ExternalInterfaces: map[string]*externalInterface{
+						networkID: {
+							Name: networkID,
+							Networks: map[string]*network{
+								"byovnetbridge-vlan1-10-128-8-0_23": {
```
ashvindeodhar, line 271: to closely mimic the existing usecase, can you add one endpoint in each network instead of 2 endpoints in single network? so need 2 networks, each having one endpoint with given netNs

ashvindeodhar (changes requested):

cni/network/network.go
```diff
@@ -941,77 +941,85 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-	if err != nil {
-		log.Printf("[cni-net] Failed to extract network name from network config. error: %v", err)
-		// If error is not found error, then we ignore it, to comply with CNI SPEC.
-		if !network.IsNetworkNotFoundError(err) {
+	// Loop through all the networks that are created for the given Netns. In case of multi-nic scenario ( currently supported
+	// scenario is dual-nic ), single container may have endpoints created in multiple networks. As all the endpoints are
+	// deleted, getNetworkName will return error of the type NetworkNotFoundError which will result in nil error as compliance
+	// with CNI SPEC as mentioned below.
+
+	numEndpointsToDelete := plugin.nm.GetNumEndpointsInNetNs(args.Netns)
```
tamilmani1989, line 949: lets call this api only for multitenancy and keep  default value of `numEndpointsToDelete` to 1, otherwise this would regress a scenario where endpoint not created by azure cni but azure-ipam allocated ip for container id

tamilmani1989 (changes requested):

network/network_test.go
paulyufan: this test case is kept and new test case (two networks + two endpoints) is added

network/manager_mock.go
paulyufan: Looks like this mock is only intended to be used with one network and 0 endpoint... so we return 0. Check with GetNumberOfEndpoints()

cni/network/network.go
paulyufan: solved

cni/network/network.go
```diff
@@ -941,77 +941,89 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
 
 		if !nwCfg.MultiTenancy {
-			// attempt to release address associated with this Endpoint id
-			// This is to ensure clean up is done even in failure cases
-			log.Printf("[cni-net] Failed to query endpoint %s: %v", endpointID, err)
-			logAndSendEvent(plugin, fmt.Sprintf("Release ip by ContainerID (endpoint not found):%v", args.ContainerID))
-			if err = plugin.ipamInvoker.Delete(nil, nwCfg, args, nwInfo.Options); err != nil {
-				return plugin.RetriableError(fmt.Errorf("failed to release address(no endpoint): %w", err))
+			// Call into IPAM plugin to release the endpoint's addresses.
+			for i := range epInfo.IPAddresses {
```
tamilmani1989, line 1011: how about this?
`for _, ipAddr := range epInfo.IPAddresses {`

paulyufan: Hi Tamilmani, I like this style. I think it's old code there and if we change this to for _, ipAddr type, it will have the linter issue => "Implicit memory aliasing in for loop", so I suggest we can keep this as it is.......

tamilmani1989 (approved):
lgtm. Address ashvin's comment as well

ashvindeodhar (approved):

PR #1829, opened 2023-03-01
instantiate a default cns config if loading fails

### LINKED ISSUE #1775: CNS fails if cns_config.json does not exist
**What happened:** 
We are using older version of CNS (v1.1.0), and we do not have a config file (no /opt/cni/bin/cns_config.json). Recently, when we used an upgraded CNS binary (v.1.4.39), the service would not start. The errors seen are given below:
```
Jan 27 12:18:14 cee5794f2f31 azure-cns[359795]: panic: runtime error: invalid memory address or nil pointer dereference
Jan 27 12:18:14 cee5794f2f31 azure-cns[359795]: [signal SIGSEGV: segmentation violation code=0x1 addr=0xd8 pc=0x15a6bf6]
Jan 27 12:18:14 cee5794f2f31 azure-cns[359795]: goroutine 1 [running]:
Jan 27 12:18:14 cee5794f2f31 azure-cns[359795]: github.com/Azure/azure-container-networking/cns/configuration.SetCNSConfigDefaults(0xc00025c060?)
Jan 27 12:18:14 cee5794f2f31 azure-cns[359795]:         /mnt/vss/_work/1/s/cns/configuration/configuration.go:172 +0x16
Jan 27 12:18:14 cee5794f2f31 azure-cns[359795]: main.main()
Jan 27 12:18:14 cee5794f2f31 azure-cns[359795]:         /mnt/vss/_work/1/s/cns/service/main.go:483 +0x23cb
```

Similar to previous versions, the expectation was for the service to start anyway, using the default config values instead of panicking and failing. 

**How to reproduce it:**  
Starting the azure-cns process without having the config file (/opt/cni/bin/cns_config.json) present will reproduce the above errors.

Docker.
```
root [ /home/xxxxxxx ]# docker version
Client:
 Version:           20.10.12
 API version:       1.41
 Go version:        go1.18.8
 Git commit:        e91ed5707e038b02af3b5120fa0835c5bedfd42e
 Built:             Thu Nov 10 09:53:16 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.14
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.8
  Git commit:       87a90dc
  Built:            Thu Nov 10 09:48:38 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.6
  GitCommit:        10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc:
  Version:          1.1.2
  GitCommit:        a916309fff0f838eb94e928713dbc3c0d0ac7aa4
 docker-init:
  Version:          0.19.0
  GitCommit:
```

Linux (CBL-Mariner 2.0)

```
root [ /home/xxxxxx ]# uname -a
Linux cee5794f2f31 5.15.82.1-1.cm2 #1 SMP Thu Dec 15 02:12:38 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

In the old version we were using, a config object was returned [here](https://github.com/Azure/azure-container-networking/blob/v1.1.0/cns/configuration/configuration.go#L48). 

In the newer versions, this now returns a pointer ([code ref](https://github.com/Azure/azure-container-networking/blob/v1.4.39/cns/configuration/configuration.go#L106)). In the subsequent calls, there is an attempt to access members of this nil pointer. ([this code](https://github.com/Azure/azure-container-networking/blob/v1.4.39/cns/service/main.go#L483) calls [this code](https://github.com/Azure/azure-container-networking/blob/b0efb9c94dcd89da0ae24fa75bb4e2f00fc0ae09/cns/configuration/configuration.go#L172)).

Currently, the hack to overcome this was to create an empty cns_config.json file. Could we please ensure that CNS process does not panic, and to use the default config in case it does not find the cns_config.json file?

rbtr:
**Reason for Change**:

Currently if CNS does not find a config it returns a nil pointer and an error, but does not stop executing and continues to attempt to use that nil config pointer (immediately, in the next line...).
This change modifies the behavior such that if we fail to find a config and return a nil pointer, we instantiate a default struct and assign it to make the subsequent pointer refs valid.

**Issue Fixed**:

closes #1775

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/service/main.go
```diff
@@ -503,6 +503,7 @@ func main() {
 	cnsconfig, err := configuration.ReadConfig(cmdLineConfigPath)
 	if err != nil {
 		logger.Errorf("[Azure CNS] Error reading cns config: %v", err)
+		cnsconfig = &configuration.CNSConfig{}
```
ramiro-gamarra, line 506: We should make it apparent in the error log line that a default will be used. I'm also curious if we should only use a default in cases where the file is missing. If, for example, the file is malformed, maybe it's better to still crash such that orchestrators take action.

rbtr: agree and updated

ramiro-gamarra (approved):

PR #1831, opened 2023-03-02
fix: [NPM-WIN] do not refresh endpoints if pod cache is empty

ck319:
**Reason for Change**:

Not refreshing when not necessary improves NPM efficiency

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/pkg/dataplane/dataplane.go
+++ b/npm/pkg/dataplane/dataplane.go
@@ -216,6 +216,11 @@ func (dp *DataPlane) ApplyDataPlane() error {
 
 	// NOTE: ideally we won't refresh Pod Endpoints if the updatePodCache is empty
 	if dp.shouldUpdatePod() {
+		// do not refresh endpoints if the updatePodCache is empty
+		if len(dp.updatePodCache.cache) == 0 {
+			return nil
+		}
+
 		err := dp.refreshPodEndpoints()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


npm/pkg/dataplane/dataplane.go
```diff
@@ -216,6 +216,11 @@ func (dp *DataPlane) ApplyDataPlane() error {
 
 	// NOTE: ideally we won't refresh Pod Endpoints if the updatePodCache is empty
 	if dp.shouldUpdatePod() {
+		// do not refresh endpoints if the updatePodCache is empty
+		if len(dp.updatePodCache.cache) == 0 {
```
huntergregory, line 220: will need to lock the cache

huntergregory (changes requested):

npm/pkg/dataplane/dataplane.go
ck319: updated

huntergregory: Locking the updatePodCache while refreshing Endpoints would be a regression in the sense that it would block Pod controller from doing any dataplane operations (e.g. AddToSets() and applyIPSets()).

ck319: reconfigured locks

huntergregory (approved):
logic lgtm

vakalapa (approved):

huntergregory (approved):

PR #1820, opened 2023-02-23
Add multiplat Windows 2019 and 2022 image support to tooling and pipelines, use for CNS/NPM

rbtr:
**Reason for Change**:

Updates the container build process to support multiple WS versions in the multiplat manifest.
- Adds WS19 image builds
- Adds WS19/WS22 specific `--os-version` metadata to the multiplat manifest
- Moves the Windows image builds off of Windows runners (which are unreliable, see #1815), and builds *from* Linux *targeting* Windows using multistage Dockerfile, Go GOOS crosscompiling, and Docker BuildX magic
- Removes Windows-specific branching from Pipelines

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
closes #1808
closes #1815
impacts #1789

PR #1835, opened 2023-03-03
remove vendor again

rbtr:
**Reason for Change**:

The vendor dir was added back to try to resolve issues with Windows builds - this was addressed via a different solution in #1820 (we only build on Linux and set the GOOS to crosscompile appropriately), and we can remove the vendor again and let go mod do its thing.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr:
just padding my delta LOC

timraymond (approved):
The most beautiful thing I've seen all day

plusminus

PR #1841, opened 2023-03-07
Skip removing vnet namespace in Linux CNI Multitenancy

tamilmani1989:
**Reason for Change**:

Its not required to tear down vnet namespace if there no endpoints on that vnet and then create back again later

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


network/transparent_vlan_endpointclient_linux.go
```diff
@@ -513,15 +513,17 @@ func (client *TransparentVlanEndpointClient) DeleteEndpointsImpl(ep *endpoint, g
 
 	log.Printf("[transparent vlan] There are %d routes remaining after deletion", routesLeft)
 
-	if routesLeft <= numDefaultRoutes {
-		// Deletes default arp, default routes, vlan veth; there are two default routes
-		// so when we have <= numDefaultRoutes routes left, no containers use this namespace
```
vipul-21, line 518: We are skipping deletion of vnet namespace if there are no containers but when there are new containers for that namespace, we won't need to create that namespace again ? Is that correct ?

Wanted to know if there is a scenario in which the namespaces are getting accumulated since they were not deleted.

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -513,15 +513,17 @@ func (client *TransparentVlanEndpointClient) DeleteEndpointsImpl(ep *endpoint, g
 
 	log.Printf("[transparent vlan] There are %d routes remaining after deletion", routesLeft)
 
-	if routesLeft <= numDefaultRoutes {
```
vipul-21, line 516: Just curious, will commenting this out won't affect the unit test cases we run for it ? https://github.com/Azure/azure-container-networking/blob/master/network/transparent_vlan_endpointclient_linux_test.go#L467

If so, do we need to update the unit test cases as well ?

tamilmani1989: correct updated it

tamilmani1989: good question. these namespaces are created per vlanid. max vlan id support is 4096 and max network namespace that user can create is 27K. so at max we wont go above creating 4K namespace per vm. so should be safe even if we dont remove namespace that we created.

root@aks-nodepool1-36637793-vmss000000:/# cat /proc/sys/user/max_net_namespaces
27653

network/transparent_vlan_endpointclient_linux.go
tamilmani1989: good question. cni creates netns per vlanid and at max vlanid can range from 1-4096 and also we dont support more than 150 containers in multitenancy. so not deleting netns should not cause any issue

vmss000000:/# cat /proc/sys/user/max_net_namespaces
27653

network/transparent_vlan_endpointclient_linux.go
vipul-21: But this does limit the user's ability to go beyond 23k( approximately) even if the vlan is not at max i.e 4k 
I understand that this is safe as we don't have the requirement or a user that has that many namespace, but why are we doing it? Isn't it is safer to delete and make sure it is created when needed, instead of letting it be there.

OR Does this help in reducing the latency since we don't need to create a namespace again ? 
(Do correct me if I am understanding it wrong)

tamilmani1989: so this will be a problem when we parallelize the execution of cni in future.. one doing add and other doing delete in same namespace based on number of default routes

vipul-21 (approved):

Issue #1729, opened 2022-12-16
Edge Case for memory-starved Windows Server 2022 Nodes with NPM


huntergregory:
In memory-starved Windows Server 2022 nodes, there's a rare issue where a Pod may have unexpected connectivity.

### Scenario
- Pod A exists
- At the same time:
  - Pod A is deleted/restarts
  - NPM restarts
  - Pod B comes up with the same IP

Controller events can be jumbled. The possible sequences are:
- Pod A create --> Pod A cleanup --> Pod B create
- Pod A create --> Pod B create --> Pod A cleanup
- Pod B create --> Pod A create --> Pod A cleanup

The first sequence is typical, but the second and third need extra logic.

### Mitigation
If a Pod encounters this issue, restore proper connectivity by bumping the Pod (deleting/recreating).

### Code Fix
This happens because the newly created NPM doesn't know about Pod A's original HNS Endpoint. So NPM can incorrectly associate an Endpoint with a deleted Pod/IP pair.

#### Fix
Use this trait about current code: throws an error if an Endpoint was previously assigned to a Pod, but another Pod claims to have the same IP/Endpoint. The `UpdatePod` will remain in the dirty cache and be retried later.

If a Pod is marked for delete (i.e. when the Pod Metadata has `nodeName=""`), do the following:
- if the Pod is associated with the IP's Endpoint:
    - reset the Endpoint's NPM ACLs
    - mark the Endpoint's Pod as unspecified
- always (this code already exists):
    - remove that Pod Key's `UpdatePod` from the dirty cache so we don't retry with it anymore

##### Sequences 
- *Pod A create --> Pod A cleanup --> Pod B create*
    - Pod A create: Endpoint's Pod is unspecified. Pod A incorrectly assigned to endpoint. Any ACLs for Pod A are added.
    - Pod A cleanup: Reset ACLs for the Endpoint. Mark the Endpoint's Pod as unspecified.
    - Pod B create: Endpoint's Pod is unspecified. Pod B correctly assigned to endpoint. Any ACLs for Pod B are added.
- *Pod A create --> Pod B create --> Pod A cleanup*
    - Pod A create: Endpoint's Pod is unspecified. Pod A incorrectly assigned to endpoint. Any ACLs for Pod A are added.
    - Pod B create: failure since B has the same IP/Endpoint as Pod A. Keep B's `UpdatePod` in the cache, which will be retried next `ApplyDataplane()`
[...]

PR #1856, opened 2023-03-17
fix: [WIN-NPM] process updatePods in fifo order

huntergregory:
**Reason for Change**:
Fully solve #1729. When the issue’s scenario occurs, ordering of Pods matters for UpdatePod() within ApplyDataPlane().

UTs have been failing for sequence 1 and 2 of this issue, and the chance of failure is higher when applying IPSets in background.

**Issue Fixed**:

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Notes**:
### Solution
Handle `updatePod` objects in FIFO order, the same order as the control plane.

### Results
Update Pod ACLs in same order as the control plane, e.g., the first Pod created is the first Pod to have proper connectivity.


huntergregory:
Closing because this issue would be extremely rare. All of these rare events would have to coincide.
1. Sequence 2 of #1729
2. Pods in question are on the same node
3. RefreshEndpoints failure

huntergregory:
> Closing because this issue would be extremely rare. All of these rare events would have to coincide.
> 
> 1. Sequence 2 of [Edge Case for memory-starved Windows Server 2022 Nodes with NPM #1729](https://github.com/Azure/azure-container-networking/issues/1729)
> 2. Pods in question are on the same node
> 3. RefreshEndpoints failure

The chance of this occurring is higher when applying ipsets in background (bullet point 3 is no longer required).

Have now seen sequence 1 fail too.

npm/pkg/dataplane/dataplane.go
```diff
@@ -149,6 +174,9 @@ func (dp *DataPlane) AddToSets(setNames []*ipsets.IPSetMetadata, podMetadata *Po
 			klog.Infof("[DataPlane] {AddToSet} pod key %s not found in updatePodCache. creating a new obj", podMetadata.PodKey)
 			updatePod = newUpdateNPMPod(podMetadata)
 			dp.updatePodCache.cache[podMetadata.PodKey] = updatePod
+
+			// add to queue only if not in the cache/queue already
+			dp.updatePodCache.queue = append(dp.updatePodCache.queue, podMetadata.PodKey)
```
vakalapa, line 179: Where are we checking if the key is not in queue ? we will need to have a 2nd check here,

npm/pkg/dataplane/dataplane.go
```diff
@@ -177,6 +205,9 @@ func (dp *DataPlane) RemoveFromSets(setNames []*ipsets.IPSetMetadata, podMetadat
 			klog.Infof("[DataPlane] {RemoveFromSet} pod key %s not found in updatePodCache. creating a new obj", podMetadata.PodKey)
 			updatePod = newUpdateNPMPod(podMetadata)
 			dp.updatePodCache.cache[podMetadata.PodKey] = updatePod
+
+			// add to queue only if not in the cache/queue already
```
vakalapa, line 209: need to add a secondary check here if the key is in queue or not,

npm/pkg/dataplane/dataplane.go
```diff
@@ -236,9 +268,18 @@ func (dp *DataPlane) ApplyDataPlane() error {
 		// lock updatePodCache while driving goal state to kernel
 		// prevents another ApplyDataplane call from updating the same pods
 		dp.updatePodCache.Lock()
-		defer dp.updatePodCache.Unlock()
+		defer func() {
+			dp.updatePodCache.removeDeletedItemsFromQueue()
+			dp.updatePodCache.Unlock()
```
vakalapa, line 273: May be create a datastructure called ordered Map ? and use that as the update pod cache ? 

example: https://github.com/iancoleman/orderedmap/blob/master/orderedmap.go

npm/pkg/dataplane/types.go
```diff
@@ -76,3 +78,70 @@ func (npmPod *updateNPMPod) updateIPSetsToRemove(setNames []*ipsets.IPSetMetadat
...
+		cache:           make(map[string]*updateNPMPod),
+		queue:           make([]string, 0, initialCapacity),
+		initialCapacity: initialCapacity,
+	}
+}
+
+// enqueue adds a pod to the queue if necessary and returns the pod object used
+func (c *updatePodCache) enqueue(m *PodMetadata) *updateNPMPod {
+	pod, ok := c.cache[m.PodKey]
+	if !ok {
```
vakalapa, line 105: What if the nodename got updated ? we will need to check if the pod objs are same if not update the element in map and move the pod obj to the end ?

npm/pkg/dataplane/types.go
```diff
@@ -76,3 +78,70 @@ func (npmPod *updateNPMPod) updateIPSetsToRemove(setNames []*ipsets.IPSetMetadat
...
+	}
+
+	return pod
+}
+
+// dequeue returns the first pod in the queue and removes it from the queue.
+// Caller should ensure the queue is not empty.
+// Otherwise, the following will occur: "panic: runtime error: index out of range [0] with length 0"
+func (c *updatePodCache) dequeue() *updateNPMPod {
+	pod := c.cache[c.queue[0]]
```
vakalapa, line 120: check if 0th element is present before calling for it

vakalapa (changes requested):

npm/pkg/dataplane/types.go
huntergregory: discussed offline. caller must lock updatePodCache and check isEmpty() before calling dequeue()

vakalapa (approved):

PR #1861, opened 2023-03-20
timeout after 15 minutes waiting for the NNC reconciler to start

rbtr:
**Reason for Change**:

Currently we wait indefinitely for the NNC reconciler to receive an NNC and run the first reconcile before continuing execution (starting the CNS webserver, IPAM pool monitor, etc).
Instead, if the NNC Reconciler does not get an NNC and successfully run a reconcile within 15 minutes, timeout and fail to make it evident that CNS is encountering a fatal error/stuck.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -1272,9 +1272,12 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 		}
 	}()
 	logger.Printf("initialized NodeNetworkConfig reconciler")
-	// wait for the Reconciler to run once on a NNC that was made for this Node
-	if started := nncReconciler.Started(ctx); !started {
-		return errors.Errorf("context cancelled while waiting for reconciler start")
+	// wait for the Reconciler to run once on a NNC that was made for this Node.
+	// the nncReadyCtx has a timeout of 15 minutes, after which we will consider
+	// this false and the NNC Reconciler stuck/failed.
+	nncReadyCtx, _ := context.WithTimeout(ctx, 15*time.Minute) //nolint // it will time out and not leak
+	if started, err := nncReconciler.Started(nncReadyCtx); !started {
+		return errors.Wrapf(err, "context cancelled while waiting for reconciler start")
 	}
 	logger.Printf("started NodeNetworkConfig reconciler")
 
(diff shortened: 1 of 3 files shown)


cns/service/main.go
```diff
@@ -1255,31 +1255,38 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
-		return errors.Errorf("context cancelled while waiting for reconciler start")
+	logger.Printf("Initialized controller-manager.")
+	for {
+		logger.Printf("Waiting for NodeNetworkConfig reconciler to start.")
+		// wait for the Reconciler to run once on a NNC that was made for this Node.
+		// the nncReadyCtx has a timeout of 15 minutes, after which we will consider
+		// this false and the NNC Reconciler stuck/failed, log and retry.
+		nncReadyCtx, _ := context.WithTimeout(ctx, 15*time.Minute) //nolint // it will time out and not leak
+		if started, err := nncReconciler.Started(nncReadyCtx); !started {
+			log.Errorf("NNC reconciler has not started, does the NNC exist? err: %v", err)
```
csfmomo, line 1280: Is this remind debugger that the failure could be nnc already exist?

rbtr: the failure could be that the NNC does _not_ exist

cns/service/main.go
```diff
@@ -1255,31 +1255,38 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
-			time.Sleep(time.Second)
+			time.Sleep(time.Second) // TODO(rbtr): make this exponential backoff
 		}
 	}()
-	logger.Printf("initialized NodeNetworkConfig reconciler")
-	// wait for the Reconciler to run once on a NNC that was made for this Node
-	if started := nncReconciler.Started(ctx); !started {
-		return errors.Errorf("context cancelled while waiting for reconciler start")
+	logger.Printf("Initialized controller-manager.")
+	for {
```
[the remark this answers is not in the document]
rbtr: we were already waiting indefinitely, because the call to `nncReconciler.Started` would not timeout (the context was indefinite) and so it would only return when the Reconciler has started. 
This change makes it so that we periodically time-out, log that we haven't started yet, and retry.

csfmomo (approved):

PR #1866, opened 2023-03-21
Dualstack Overlay CNI 

rjdenney:
**Reason for Change**:

This PR is to provide dualStack CNI changes to create a NC with both ipv4 and ipv6 address in dualStackOverlay mode on both Windows&Linux OS

The PR includes following UTs:
(1) RequestIPs() => request ipv4 & ipv6 ip addresses and make sure dual-stack NC creation 
(2) RequestIPAddress() => request only ipv4 and make sure single-stack NC is working fine
(3) ReleaseIPs() => release ipv4 & ipv6 ip addresses and make sure they are released
(4) ReleaseIPAddress() => release ipv4 address only and make sure this ipv4 address can be released in single-tenancy

Tests we did manually(on both Linux and Windows):
(1) Network(ipv4+ipv6):
    pod -> pod (one node and different nodes)
    pod -> host 
    host -> pod
    pod -> internet
    host -> internet

(2) Launched dual stack overlay cluster in different regions: eastus, eastuseuap, centraluseuap
(3) Test Different Linux OSs (including Ubuntu 19, Ubunt22 and CLB-Mariner)
(4) Test different Linux processors (ARM/AMDH&I)
(5) NMAgent tests
(6) HNS tests: https://microsoftapc-my.sharepoint.com/personal/ppereira_microsoft_com/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fppereira%5Fmicrosoft%5Fcom%2FDocuments%2FDocuments%2FProjects%2FAKS%2DIPV6%2DSupport&ct=1680642656649&or=Teams%2DHL&ga=1

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


network/network_windows.go
```diff
@@ -185,6 +190,55 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
 	return nw, nil
 }
 
+// add ipv4 and ipv6 routes (if dual stack mode) to windows Node
+func (nm *networkManager) addNewNetRules(nwInfo *NetworkInfo) error {
```
AzureAhai, line 194: Is there any unit test for this function and also its Windows version?

cni/network/network_windows.go
```diff
@@ -385,7 +385,10 @@ func getNATInfo(nwCfg *cni.NetworkConfig, ncPrimaryIPIface interface{}, enableSn
 			ncPrimaryIP = ncPrimaryIPIface.(string)
 		}
 
-		natInfo = append(natInfo, []policy.NATInfo{{VirtualIP: ncPrimaryIP, Destinations: []string{networkutils.AzureDNS}}, {Destinations: []string{networkutils.AzureIMDS}}}...)
+		// only append natInfo if ncPrimaryIP is ipv4
```
AzureAhai, line 388: you can add a scenario on network_windows_test.go for this condition.

cni/network/network.go
```diff
@@ -577,12 +578,20 @@ func (plugin *NetPlugin) cleanupAllocationOnError(
 	options map[string]interface{},
 ) {
 	if result != nil && len(result.IPs) > 0 {
-		if er := plugin.ipamInvoker.Delete(&result.IPs[0].Address, nwCfg, args, options); er != nil {
+		addresses := make([]*net.IPNet, len(result.IPs))
```
AzureAhai, line 581: Please add a comment about this section and also the ipv6 one.

cni/network/network.go
```diff
@@ -651,6 +662,22 @@ func (plugin *NetPlugin) createNetworkInternal(
 		ServiceCidrs:                  ipamAddConfig.nwCfg.ServiceCidrs,
 	}
 
+	ipv4Subnet := network.SubnetInfo{
```
AzureAhai, line 665: Do we need to have a similar if statement like IPv6?

cni/network/network_test.go
```diff
@@ -76,7 +76,7 @@ func GetTestResources() *NetPlugin {
 	plugin.report = &telemetry.CNIReport{}
 	mockNetworkManager := acnnetwork.NewMockNetworkmanager()
 	plugin.nm = mockNetworkManager
-	plugin.ipamInvoker = NewMockIpamInvoker(false, false, false)
+	plugin.ipamInvoker = NewMockIpamInvoker(true, false, false) // enable ipv6 flag for dualstack test cases
```
AzureAhai, line 79: Do we also have a test case for IPv4 dualstack?

cni/network/invoker.go
```diff
@@ -17,7 +17,7 @@ type IPAMInvoker interface {
 	Add(IPAMAddConfig) (IPAMAddResult, error)
 
 	// Delete calls to the invoker source, and returns error. Returning an error here will fail the CNI Delete call.
-	Delete(address *net.IPNet, nwCfg *cni.NetworkConfig, args *cniSkel.CmdArgs, options map[string]interface{}) error
+	Delete(addresses []*net.IPNet, nwCfg *cni.NetworkConfig, args *cniSkel.CmdArgs, options map[string]interface{}) error
```
vipul-21, line 20: Shouldn't this be backward compatible ?
Just thinking out loud, we are changing the data type from string to array. Are we sure that this change in function definition is readily acceptable wherever IPAM invoker is used.

cni/network/invoker_azure.go
```diff
@@ -65,7 +70,11 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 	defer func() {
 		if err != nil {
 			if len(addResult.ipv4Result.IPs) > 0 {
-				if er := invoker.Delete(&addResult.ipv4Result.IPs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
+				addresses := []*net.IPNet{}
+				for _, ip := range addResult.ipv4Result.IPs {
```
vipul-21, line 74: Here we are appending from ipv4 result. And in the delete function we are deleting the ipv6 addresses as well. Correct me if am wrong but shouldn't we have the ipv6 addresses in the list too ?
OR am i missing something ? we can add a comment describing what the intent of this function is.

cni/network/invoker_cns.go
```diff
@@ -70,94 +73,115 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
+			}
 
-	ncgw := net.ParseIP(info.ncGatewayIPAddress)
-	if ncgw == nil {
-		if invoker.ipamMode != util.V4Overlay {
-			return IPAMAddResult{}, errors.Wrap(errInvalidArgs, "%w: Gateway address "+info.ncGatewayIPAddress+" from response is invalid")
+			if net.ParseIP(info.podIPAddress).To4() != nil {
+				ncgw = net.ParseIP(overlayGatewayV4IP)
+			} else {
+				ncgw = net.ParseIP(overlayGatewayV6IP)
```
vipul-21, line 116: Should we add a condition to check that the ipv6 address is returned ? or we are sure that if not ipv4 then ipv6 is the case.

cni/network/invoker_cns.go
```diff
@@ -70,94 +73,115 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
-	addResult.hostSubnetPrefix = *hostIPNet
+		addResult.hostSubnetPrefix = *hostIPNet
 
-	// set subnet prefix for host vm
-	// setHostOptions will execute if IPAM mode is not v4 overlay
-	if invoker.ipamMode != util.V4Overlay {
-		if err := setHostOptions(ncipnet, addConfig.options, &info); err != nil {
-			return IPAMAddResult{}, err
+		// set subnet prefix for host vm
+		// setHostOptions will execute if IPAM mode is not v4 overlay
```
vipul-21, line 173: Need to update this comment based on the `if` condition

cni/network/invoker_cns.go
```diff
@@ -229,20 +253,42 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
...
+		}
 	} else {
 		log.Printf("CNS invoker called with empty IP address")
 	}
 
-	if err := invoker.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
-		return errors.Wrap(err, fmt.Sprintf("failed to release IP %v with err ", address)+"%w")
+	if err := invoker.cnsClient.ReleaseIPs(context.TODO(), req); err != nil {
+		// if we fail a release with a 404 error try using the old API
+		if errors.Is(err, cnscli.ErrAPINotFound) {
```
vipul-21, line 273: In this case, what if the API fails a single time with 404 and we call the old API. ? Do we intend to call the old API for the subsequent calls ?
OR the old API is serving just as back up ? ( will probably produce the same results ?)

cni/network/invoker_mock.go
```diff
@@ -79,18 +79,20 @@ func (invoker *MockIpamInvoker) Add(opt IPAMAddConfig) (ipamAddResult IPAMAddRes
...
 }
 
-func (invoker *MockIpamInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConfig, _ *skel.CmdArgs, options map[string]interface{}) error {
+func (invoker *MockIpamInvoker) Delete(addresses []*net.IPNet, nwCfg *cni.NetworkConfig, _ *skel.CmdArgs, options map[string]interface{}) error {
 	if invoker.v4Fail || invoker.v6Fail {
 		return errDeleteIpam
 	}
 
-	if address == nil || invoker.ipMap == nil {
+	if len(addresses) <= 0 || invoker.ipMap == nil {
```
vipul-21, line 87: Is there a case where  length can be < 0 ?

vipul-21 (commented):
Can we add a bit more description for the PR as well. 
Like Reason for change, design doc etc.

network/network_windows.go
```diff
@@ -185,6 +190,55 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+	if !strings.Contains(nwInfo.MasterIfName, ifNamePrefix) {
+		ifName = fmt.Sprintf("%s (%s)", ifNamePrefix, nwInfo.MasterIfName)
+	}
+
+	// add ipv4 and ipv6 new net rules to windows node
+	for _, subnet := range nwInfo.Subnets {
+		prefix := subnet.Prefix.String()
+		gateway := subnet.Gateway.String()
+
+		ip, _, _ := net.ParseCIDR(prefix)
```
vipul-21, line 211: Any reason why we are not acknowledging the error from parseCIDR

network/network_windows.go
```diff
@@ -185,6 +190,55 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+		if ip.To4() != nil {
+			// netsh interface ipv4 add route $subnetV4 $hostInterfaceAlias "0.0.0.0" metric=270
+			netshV4DefaultRoute := fmt.Sprintf(netRouteCmd, "ipv4", prefix, ifName, ipv4DefaultHop, "270")
```
vipul-21, line 214: Nit: Should we make the metric as const as well ? I see we are using the same for ipv6.

network/network_windows.go
paulyufan2: check is added

network/network_windows.go
paulyufan2: Need to confirm with HNS team about this

cni/network/invoker_mock.go
paulyufan2: Change this to len(addresses) == 0

cni/network/invoker_cns.go
paulyufan2: comments updated

cni/network/invoker_cns.go
paulyufan2: we do not have that case

cni/network/invoker.go
vipul-21: As per Paul, they did all tests in all scenarios on both Linux and Windows and it works fine.

cni/network/invoker_cns.go
```diff
@@ -70,94 +72,143 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
+			ncPrimaryIP:        response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.IPSubnet.IPAddress,
+			ncGatewayIPAddress: response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.GatewayIPAddress,
+			hostSubnet:         response.PodIPInfo[i].HostPrimaryIPInfo.Subnet,
+			hostPrimaryIP:      response.PodIPInfo[i].HostPrimaryIPInfo.PrimaryIP,
+			hostGateway:        response.PodIPInfo[i].HostPrimaryIPInfo.Gateway,
+		}
 
-	log.Printf("[cni-invoker-cns] Received info %+v for pod %v", info, podInfo)
+		// set the NC Primary IP in options
+		addConfig.options[network.SNATIPKey] = info.ncPrimaryIP
```
ashvindeodhar, line 127: this value will get overwritten with the for loop. Where is this used? what happens if it gets ipv6 ip?

cni/network/invoker_cns.go
ashvindeodhar: +1 on @vipul-21 comment. We should validate that it's ipv6 IP in 'else if' followed by 'esle' which returns error

cni/network/invoker_cns.go
```diff
@@ -70,94 +72,143 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
-		if err != nil {
-			return IPAMAddResult{}, err
+		ncgw := net.ParseIP(info.ncGatewayIPAddress)
+		if ncgw == nil {
+			if (invoker.ipamMode != util.V4Overlay) && (invoker.ipamMode != util.DualStackOverlay) {
+				return IPAMAddResult{}, errors.Wrap(errInvalidArgs, "%w: Gateway address "+info.ncGatewayIPAddress+" from response is invalid")
+			}
+
+			if net.ParseIP(info.podIPAddress).To4() != nil {
+				ncgw, err = getOverlayGateway(ncIPNet)
```
ashvindeodhar, line 144: @paulyufan we get the GW from the pod subnet in case of ipv4. We are using hardcoded value for ipv6. Why is that?

cni/network/invoker_cns.go
```diff
@@ -70,94 +72,143 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
+					},
+				},
+				Routes: []*cniTypes.Route{
+					{
+						Dst: network.Ipv4DefaultRouteDstPrefix,
+						GW:  ncgw,
+					},
+				},
+			}
+		} else {
```
ashvindeodhar, line 174: let's check if this is ipv6 IP here

cni/network/invoker_cns.go
```diff
@@ -229,20 +280,42 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
...
-	if err := invoker.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
-		return errors.Wrap(err, fmt.Sprintf("failed to release IP %v with err ", address)+"%w")
+	if err := invoker.cnsClient.ReleaseIPs(context.TODO(), req); err != nil {
+		// if we fail a release with a 404 error try using the old API
+		if errors.Is(err, cnscli.ErrAPINotFound) {
+			ipconfigRequest := cns.IPConfigRequest{
+				OrchestratorContext: orchestratorContext,
+				PodInterfaceID:      GetEndpointID(args),
+				InfraContainerID:    args.ContainerID,
+				DesiredIPAddress:    req.DesiredIPAddresses[0],
```
ashvindeodhar, line 305: req.DesiredIPAddresses[0] can run into nil pointer deref?
what if len(addresses) check on line 289 fails? You will never initialize the slice on line 290.
Has this been tested with UT?

cni/network/invoker_cns.go
```diff
@@ -229,20 +280,42 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
...
+			}
+			log.Errorf("Failed to release IPs using ReleaseIPs from CNS, going to try ReleaseIPAddress. error: %v request: %v", err, req)
```
ashvindeodhar, line 307: same comment as request func. Move this up and revisit the error message.

cni/network/invoker_cns_test.go
```diff
@@ -28,6 +28,14 @@ func getTestIPConfigRequest() cns.IPConfigRequest {
 	}
 }
 
+func getTestIPConfigsRequest() cns.IPConfigsRequest {
```
ashvindeodhar, line 31: will review rest of the code tomorrow morning.

ashvindeodhar (changes requested):

cni/network/invoker_cns.go
paulyufan2: Hi Ashvin, there are two reasons:
(1) We get this default ipv6 gateway "fe80::1234:5678:9abc" from Jearyn at the beginning and we are using this as default ipv6 gateway.
(2) Jaeryn made changes recently for this reason:
```
Fixing overlay gateway bug that prevented cloud-node-manager-windows from initializing node because it couldn't reach IMDS. An additional VFP rule was implicitly plumbed by HNS to add 169.254.1.1/16 to the exception list. This is most likely based off the RouteConfiguration in the "azure" HNS network.  This additional VFP rule was skipping SNAT for any IP address in this range 169.254.0.0-169.254.255.255. We'll set the first .1 IP of podcidr as the gateway now.
```
Do we need to use this to reach IMDS?

cni/network/invoker_azure.go
```diff
@@ -131,35 +141,38 @@ func (invoker *AzureIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkCo
...
+			if err := invoker.plugin.DelegateDel(nwCfg.IPAM.Type, nwCfg); err != nil {
+				log.Printf("Failed to release ipv4 address: %v", err)
+				return invoker.plugin.Errorf("Failed to release ipv4 address: %v with error: ", nwCfg.IPAM.Address, err)
+			}
+		} else if len(address.IP.To16()) == bytesSize16 {
+			nwCfgIpv6 := *nwCfg
+			nwCfgIpv6.IPAM.Environment = common.OptEnvironmentIPv6NodeIpam
+			nwCfgIpv6.IPAM.Type = ipamV6
+			nwCfgIpv6.IPAM.Address = address.IP.String()
+			if len(invoker.nwInfo.Subnets) > 1 {
```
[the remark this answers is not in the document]
paulyufan2: This is not guranteed, we do need to check which one is for ipv4 and which one is for ipv6

cni/network/network.go
```diff
@@ -570,19 +575,28 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
 func (plugin *NetPlugin) cleanupAllocationOnError(
 	result, resultV6 *cniTypesCurr.Result,
 	nwCfg *cni.NetworkConfig,
 	args *cniSkel.CmdArgs,
 	options map[string]interface{},
 ) {
 	if result != nil && len(result.IPs) > 0 {
-		if er := plugin.ipamInvoker.Delete(&result.IPs[0].Address, nwCfg, args, options); er != nil {
+		addresses := make([]*net.IPNet, len(result.IPs))
+		for i, ip := range result.IPs {
```
tamilmani1989, line 587: instead of doing this. we should keep same delete method signature and iterate over result.IPs and pass that

tamilmani1989: btw why we change to array. are we expecting more than 1 ipv4/ipv6 addresses here?

cni/network/network.go
```diff
@@ -651,6 +667,22 @@ func (plugin *NetPlugin) createNetworkInternal(
...
+	nwInfo.Subnets = append(nwInfo.Subnets, ipv4Subnet)
+
+	if ipamAddResult.ipv6Result != nil && len(ipamAddResult.ipv6Result.IPs) > 0 {
+		ipv6Subnet := network.SubnetInfo{
+			Family:  platform.AfINET6,
+			Prefix:  *podSubnetV6Prefix,
+			Gateway: ipamAddResult.ipv6Result.IPs[0].Gateway,
+		}
+		nwInfo.Subnets = append(nwInfo.Subnets, ipv6Subnet)
+	}
```
tamilmani1989, line 684: can we move this to separate function?

cni/network/network.go
```diff
@@ -651,6 +667,22 @@ func (plugin *NetPlugin) createNetworkInternal(
...
+
 	setNetworkOptions(ipamAddResult.ncResponse, &nwInfo)
 
 	addNatIPV6SubnetInfo(ipamAddConfig.nwCfg, ipamAddResult.ipv6Result, &nwInfo)
```
tamilmani1989, line 688: is this still required? i see anyway you are adding ipv6 subnet in nwinfo above?

network/network_windows.go
AzureAhai: TODO will be added

AzureAhai:
lgtm

cni/network/network_windows.go
```diff
@@ -385,7 +385,10 @@ func getNATInfo(nwCfg *cni.NetworkConfig, ncPrimaryIPIface interface{}, enableSn
...
+		if net.ParseIP(ncPrimaryIP).To4() != nil {
```
tamilmani1989, line 389: snat required for dns and imds. isn't there a scenario where user can use ipv6 dns ip (ipv6 equivalent of 168.63.129.16)?

network/network.go
```diff
@@ -25,6 +25,8 @@ const (
 const (
 	// ipv6 modes
 	IPV6Nat = "ipv6nat"
+	// dual stack mode
+	DualStack = "dualStackOverlay"
```
tamilmani1989, line 29: i see you are using this only in network_windows.go. can you move it there? also update name of variable to "dualStackOverlay"

network/network_windows.go
```diff
@@ -185,6 +190,58 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
 	return nw, nil
 }
 
+// add ipv4 and ipv6 routes (if dual stack mode) to windows Node
+func (nm *networkManager) addNewNetRules(nwInfo *NetworkInfo) error {
```
tamilmani1989, line 194: can you add detailed description on what this function does and what routes it configures?

network/network_windows.go
```diff
@@ -185,6 +190,58 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+		out string
+	)
+
+	// get interface name of VM adapter
+	ifName := nwInfo.MasterIfName
+	if !strings.Contains(nwInfo.MasterIfName, ifNamePrefix) {
+		ifName = fmt.Sprintf("%s (%s)", ifNamePrefix, nwInfo.MasterIfName)
+	}
+
+	// add ipv4 and ipv6 new net rules to windows node
```
tamilmani1989, line 206: add description on those new net rules and why its required?

network/network_windows.go
```diff
@@ -185,6 +190,58 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+			}
+
+			// netsh interface ipv4 add route $subnetV4 $hostInterfaceAlias $gatewayV4 metric=300
+			netshV4GatewayRoute := fmt.Sprintf(netRouteCmd, "ipv4", prefix, ifName, gateway, "300")
+			if out, err = nm.plClient.ExecuteCommand(netshV4GatewayRoute); err != nil {
+				return fmt.Errorf("[net] Adding ipv4 gateway route failed: %v:%v", out, err)
+			}
+		} else {
+			// netsh interface ipv6 add route $subnetV6 $hostInterfaceAlias "::" metric=270
+			netshV6DefaultRoute := fmt.Sprintf(netRouteCmd, "ipv6", prefix, ifName, ipv6DefaultHop, "270")
```
tamilmani1989, line 229: why are we using netsh calls? isn't there hnsapi support for this?

network/network_windows.go
```diff
@@ -185,6 +190,58 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+			if out, err = nm.plClient.ExecuteCommand(netshV6DefaultRoute); err != nil {
```
tamilmani1989, line 230: also how do you make sure these calls are idempotent and what if one of the calls fails and returns?

tamilmani1989 (changes requested):

cni/network/invoker_azure.go
paulyufan2: Hi, I missed this comment, yes, we do need delete ipv6 addresses here as well

cni/network/invoker_cns.go
paulyufan2: Hi, I am not sure about this since ncPrimaryIP is from PodIPInfo IPAddress, can we discuss this later on

cni/network/invoker_cns.go
```diff
@@ -70,94 +72,143 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
-		log.Printf("Failed to get IP address from CNS with error %v, response: %v", err, response)
-		return IPAMAddResult{}, errors.Wrap(err, "Failed to get IP address from CNS with error: %w")
+		// if RequestIPs call fails, we may receive API not Found error as new CNS is not supported, then try old API RequestIPAddress
+		log.Errorf("RequestIPs not supported by CNS. Invoking RequestIPAddress API. Request: %v", ipconfig)
+		if errors.Is(err, cnscli.ErrAPINotFound) {
+			ipconfigRequest := cns.IPConfigRequest{
+				OrchestratorContext: orchestratorContext,
+				PodInterfaceID:      GetEndpointID(addConfig.args),
+				InfraContainerID:    addConfig.args.ContainerID,
+			}
```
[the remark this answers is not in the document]
paulyufan2: they are using different structure:
IPConfigRequest and IPConfigsRequest

cni/network/invoker_cns.go
paulyufan2: yes I added this fix

cni/network/invoker_cns.go
paulyufan2: yes, we do call old API if new API does not work

cni/network/network.go
```diff
@@ -384,7 +385,11 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 			res.Print()
 		}
 
-		log.Printf("[cni-net] ADD command completed for pod %v with IPs:%+v err:%v.", k8sPodName, ipamAddResult.ipv4Result.IPs, err)
+		if ipamAddResult.ipv4Result.IPs != nil {
```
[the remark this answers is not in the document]
paulyufan2: We need to check here, otherwise without line 387 log, if ipv4Result object is empty, then try to get IPs: ipamAddResult.ipv4Result.IPs, the program will get crashed and exited

cni/network/network.go
```diff
@@ -740,6 +772,8 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
 	}
 
 	if opt.resultV6 != nil {
+		// inject routes to linux pod
+		epInfo.IPV6Mode = network.DualStack
```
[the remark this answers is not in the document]
paulyufan2: Because we need to set epInfo.IPV6Mode to inject routes to Linux, otherwise Linux pod is missing routes.

network/network_windows.go
paulyufan2: HNS will support it later on. Currently CNI push these routes to the windows pod as discussed.

cni/network/network.go
```diff
@@ -1013,12 +1047,14 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
 		if !nwCfg.MultiTenancy {
 			// Call into IPAM plugin to release the endpoint's addresses.
+			addresses := make([]*net.IPNet, len(epInfo.IPAddresses))
 			for i := range epInfo.IPAddresses {
-				logAndSendEvent(plugin, fmt.Sprintf("Release ip:%s", epInfo.IPAddresses[i].IP.String()))
-				err = plugin.ipamInvoker.Delete(&epInfo.IPAddresses[i], nwCfg, args, nwInfo.Options)
-				if err != nil {
-					return plugin.RetriableError(fmt.Errorf("failed to release address: %w", err))
-				}
+				addresses[i] = &epInfo.IPAddresses[i]
```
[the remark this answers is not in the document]
paulyufan2: yes, Delete() api accepts a list of IPAddresses

cni/network/network_windows.go
paulyufan2: I do not think we need nat for ipv6

network/network.go
paulyufan2: variable name has changed, it's also used in network.go to add routes

cni/network/network.go
paulyufan2: need to discuss this

paulyufan2: This is going to add ipv6 subnet when it's ipv6nat mode, so it should be there since we did not add ipv6routes as ipv6nat mode

cni/network/network.go
paulyufan2: this will append ipv6subnet to nwInfo Subnets, so I think it should be in this func as it's more clear?

cni/network/invoker_cns.go
```diff
@@ -229,20 +283,37 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
...
 		log.Printf("CNS invoker called with empty IP address")
 	}
 
-	if err := invoker.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
-		return errors.Wrap(err, fmt.Sprintf("failed to release IP %v with err ", address)+"%w")
+	if err := invoker.cnsClient.ReleaseIPs(context.TODO(), ipConfig); err != nil {
+		// if ReleaseIPs call fails, we may receive API not Found error as new CNS is not supported, then try old API ReleaseIPAddress
+		log.Errorf("ReleaseIPs not supported by CNS. Invoking ReleaseIPAddress API. Request: %v", ipConfig)
+		if cnscli.IsUnsupportedAPI(err) {
+			ipConfigs := cns.IPConfigRequest{
```
tamilmani1989, line 302: can resuse the same variable defined in 286

cni/network/invoker_cns.go
```diff
@@ -229,20 +283,37 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 		return errEmptyCNIArgs
 	}
 
-	req := cns.IPConfigRequest{
+	ipConfig := cns.IPConfigsRequest{
```
tamilmani1989, line 286: can be `ipconfigs`

cni/network/network.go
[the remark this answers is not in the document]
tamilmani1989: isn't that behaviour same even before? we didn't see any crash before..

cni/network/network.go
tamilmani1989: it does the same thing. lets remove that.anyway we are not supporting ipv6nat mode

cni/network/network.go
tamilmani1989: the function is growing bigger.. you can define internal function like constructNetworkInfo and move initializing network info and assigning these subnets under that?

cni/network/network.go
```diff
@@ -740,6 +764,8 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
 	}
 
 	if opt.resultV6 != nil {
+		// inject routes to linux pod
+		epInfo.IPV6Mode = network.DualStackOverlay
```
tamilmani1989, line 768: why ipv6Mode is concluded as DualstackOverlay and how do we handle if we support podsubnet(vnet) for ipv6 tomorrow?

cni/network/network.go
```diff
@@ -1013,8 +1039,8 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 
 		if !nwCfg.MultiTenancy {
 			// Call into IPAM plugin to release the endpoint's addresses.
+			logAndSendEvent(plugin, fmt.Sprintf("Releasing ips:%+v", epInfo.IPAddresses))
```
tamilmani1989, line 1042: i would prefer not changing this. anyway it will add one more extra log which is fine imo.. it will give more clarity on ip has been released and what not

tamilmani1989 (changes requested):

network/network_windows.go
```diff
@@ -337,6 +397,13 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *NetworkInfo, extIf *extern
 			}
 
 			log.Printf("[net] Successfully created hcn network with response: %+v", hnsResponse)
+
+			// only add net rules if it's dualStackOverlay mode and hnsNetwork is created at first time
+			if util.DualStackOverlay == DualStackOverlay {
```
tamilmani1989, line 402: trying to understand the significance of this check.. isn't these comparing 2 constant variables?

cni/network/network.go
tamilmani1989: also where we are using this field?

tamilmani1989 (changes requested):

cni/network/invoker_azure.go
```diff
@@ -122,7 +131,7 @@ func (invoker *AzureIPAMInvoker) deleteIpamState() {
 	}
 }
 
-func (invoker *AzureIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConfig, _ *cniSkel.CmdArgs, options map[string]interface{}) error {
+func (invoker *AzureIPAMInvoker) Delete(addresses []*net.IPNet, nwCfg *cni.NetworkConfig, _ *cniSkel.CmdArgs, options map[string]interface{}) error { //nolint
```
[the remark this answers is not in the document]
paulyufan2: fixed

cni/network/invoker_cns.go
```diff
@@ -70,94 +72,146 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
 	log.Printf("Requesting IP for pod %+v using ipconfig %+v", podInfo, ipconfig)
-	response, err := invoker.cnsClient.RequestIPAddress(context.TODO(), ipconfig)
+	response, err := invoker.cnsClient.RequestIPs(context.TODO(), ipconfig)
 	if err != nil {
-		log.Printf("Failed to get IP address from CNS with error %v, response: %v", err, response)
-		return IPAMAddResult{}, errors.Wrap(err, "Failed to get IP address from CNS with error: %w")
+		// if RequestIPs call fails, we may receive API not Found error as new CNS is not supported, then try old API RequestIPAddress
+		log.Errorf("RequestIPs not supported by CNS. Invoking RequestIPAddress API with infracontainerid %s", ipconfig.InfraContainerID)
+		if cnscli.IsUnsupportedAPI(err) {
+			ipconfigs := cns.IPConfigRequest{
```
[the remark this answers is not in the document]
paulyufan2: because they are using different structure:
ipConfigs: IPConfigsRequest
ipConfig: IPConfigRequest

cni/network/network.go
paulyufan2: we are using this field in network_windows.go:
// only add net rules if it's dualStackOverlay mode and hnsNetwork is created at first time
			if util.DualStackOverlay == DualStackOverlay {
				if err := nm.addNewNetRules(nwInfo); err != nil {
					return nil, err
				}
			}

cni/network/invoker_cns.go
```diff
@@ -70,94 +72,146 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
+				return IPAMAddResult{}, errors.Wrap(errInvalidArgs, "%w: Gateway address "+info.ncGatewayIPAddress+" from response is invalid")
+			}
+
+			if net.ParseIP(info.podIPAddress).To4() != nil { //nolint:gocritic
+				ncgw, err = getOverlayGateway(ncIPNet)
+				if err != nil {
+					return IPAMAddResult{}, err
+				}
+			} else if net.ParseIP(info.podIPAddress).To16() != nil {
+				ncgw = net.ParseIP(overlayGatewayV6IP)
```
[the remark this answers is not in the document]
paulyufan2: yes, we got this fixed ipv6 gateway from HNS team

cni/network/invoker_azure.go
```diff
@@ -66,7 +71,11 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 		if err != nil {
 			if len(addResult.ipv4Result.IPs) > 0 {
 				if er := invoker.Delete(&addResult.ipv4Result.IPs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
-					err = invoker.plugin.Errorf("Failed to clean up IP's during Delete with error %v, after Add failed with error %w", er, err)
+					err = invoker.plugin.Errorf("Failed to clean up IPv4 during Delete with error %v, after Add failed with error %w", er, err)
+				}
+			} else if len(addResult.ipv6Result.IPs) > 0 {
+				if er := invoker.Delete(&addResult.ipv6Result.IPs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
+					err = invoker.plugin.Errorf("Failed to clean up IPv6 during Delete with error %v, after Add failed with error %w", er, err)
```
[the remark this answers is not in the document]
paulyufan2: thanks Tamilmani, we do not need to change this in legacy ipam.

cni/network/invoker_cns.go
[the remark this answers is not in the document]
paulyufan2: yes, HNS team gave this and we do not have this requirement for now because Jaeryn fixed this before was to fix overlay gateway bug that prevented cloud-node-manager-windows from initializing node because it couldn't reach IMDS. An additional VFP rule was implicitly plumbed by HNS to add 169.254.1.1/16 to the exception list. This is most likely based off the RouteConfiguration in the "azure" HNS network.  This additional VFP rule was skipping SNAT for any IP address in this range 169.254.0.0-169.254.255.255. We'll set the first .1 IP of podcidr as the gateway now.

cni/network/invoker_cns_test.go
```diff
@@ -186,30 +196,59 @@ func TestCNSIPAMInvoker_Add(t *testing.T) {
...
+					{
+						Dst: network.Ipv6DefaultRouteDstPrefix,
+						GW:  net.ParseIP("fe80::1234:5678:9abc"),
 					},
 				},
 			},
-			want1:   nil,
 			wantErr: false,
 		},
+		{
```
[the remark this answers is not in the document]
paulyufan2: test case -- name: "fail to request IP addresses from cns" -- includes this failure

cni/network/invoker_azure_test.go
```diff
@@ -272,7 +274,29 @@ func TestAzureIPAMInvoker_Delete(t *testing.T) {
...
+				},
+				nwCfg: &cni.NetworkConfig{
+					IPAM: cni.IPAM{
+						Address: "2001:db8:abcd:0015::0/64",
+					},
+				},
+			},
+		},
+		{
+			name: "delete happy path ipv4+ipv6",
```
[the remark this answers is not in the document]
paulyufan2: since the delete() signature is not changed, so no case to for deleting both ipv4 and ipv6 addresses. Failed test cases are there:
"error on v6 delete"
"error on v4 delete"
"error address is nil"

cni/network/invoker_cns_test.go
```diff
@@ -310,13 +581,136 @@ func TestCNSIPAMInvoker_Delete(t *testing.T) {
...
+			if tt.wantErr {
+				require.Error(err)
+			} else {
+				require.NoError(err)
+			}
+		})
+	}
+}
+
+func TestCNSIPAMInvoker_Delete_NotFound(t *testing.T) {
```
ashvindeodhar, line 649: same comment as above. Not sure what this test is testing.

cni/network/multitenancy_test.go
```diff
@@ -50,30 +70,68 @@ type cnsAPIName string
 
 const (
 	GetAllNetworkContainers cnsAPIName = "GetAllNetworkContainers"
+	RequestIPs              cnsAPIName = "RequestIPs"
+	ReleaseIPs              cnsAPIName = "ReleaseIPs"
 )
 
 var (
 	errUnsupportedAPI             = errors.New("Unsupported API")
+	errNoRequestIPFound           = errors.New("No Request IP Found")
```
ashvindeodhar, line 79: what do these mean?

cni/network/network.go
```diff
@@ -483,7 +484,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		// Issue link: https://github.com/kubernetes/kubernetes/issues/57253
 
 		if nwInfoErr == nil {
-			log.Printf("[cni-net] Found network %v with subnet %v.", networkID, nwInfo.Subnets[0].Prefix.String())
+			log.Printf("[cni-net] Found network %v.", networkID)
```
ashvindeodhar, line 487: why is this updated?

cni/network/network.go
```diff
@@ -653,10 +657,9 @@ func (plugin *NetPlugin) createNetworkInternal(
 		ServiceCidrs:                  ipamAddConfig.nwCfg.ServiceCidrs,
 	}
 
+	constructNetworkInfo(ipamAddResult, podSubnetPrefix, podSubnetV6Prefix, nwInfo)
```
ashvindeodhar, line 660: the func name should indicate that it is adding the subnets to networkInfo. The current name of the func doesn't give any idea on what is being contructed

cni/network/network.go
```diff
@@ -742,6 +764,10 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
 	}
 
 	if opt.resultV6 != nil {
+		// get IPAM mode from conflist file and add ipv6 routes to linux pod if needed
```
ashvindeodhar, line 767: what does "add ipv6 routes to linux pod if needed" mean?

cni/network/network_test.go
```diff
@@ -1102,3 +1102,15 @@ func TestGetPodSubnetNatInfo(t *testing.T) {
 		require.Empty(t, natInfo, "linux podsubnet natInfo should be empty")
 	}
 }
+
+func TestGetPodSubnetNatInfoV6(t *testing.T) {
+	ncPrimaryIP := "2001:2002:2003::1"
+	nwCfg := &cni.NetworkConfig{ExecutionMode: string(util.V4Swift)}
+	natInfo := getNATInfo(nwCfg, ncPrimaryIP, false)
+	// should not add any natInfo to policy if ncPrimaryIP is ipv6
```
ashvindeodhar, line 1110: this is incorrect. check the implementation of getNATInfo, ncPrimaryIP doesn't matter.

cni/network/network_test.go
ashvindeodhar, line 79: I believe this shouldn't be just set to true for all the tests.
you should update 
func GetTestResources() *NetPlugin { to receive a flag for ipv6 and then pass it tp NewMockIpamInvoker

network/network_windows.go
```diff
@@ -185,6 +190,61 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+	if !strings.Contains(nwInfo.MasterIfName, ifNamePrefix) {
+		ifName = fmt.Sprintf("%s (%s)", ifNamePrefix, nwInfo.MasterIfName)
+	}
+
+	// iterate subnet and add ipv4 and ipv6 default route and gateway
+	for _, subnet := range nwInfo.Subnets {
+		prefix := subnet.Prefix.String()
+		gateway := subnet.Gateway.String()
+
+		ip, _, er := net.ParseCIDR(prefix)
```
ashvindeodhar, line 213: nit: please change er to err

network/network_windows.go
```diff
@@ -185,6 +190,61 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
 	return nw, nil
 }
 
+// add ipv4 and ipv6 routes (if dual stack mode) to windows Node
+// in dualstack mode, pods are created from different subnets on different nodes, gateway has to be node ip if pods want to communicate with each other
+// add routes to make node understand pod IPs come from different subnets and VFP will take decisions based on these routes to forward traffic and avoid Natting
+func (nm *networkManager) addNewNetRules(nwInfo *NetworkInfo) error {
+	var (
+		err error
```
ashvindeodhar, line 198: why do we need to declare these here upfront?

network/network_windows.go
```diff
@@ -185,6 +190,61 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+	}
+
+	// iterate subnet and add ipv4 and ipv6 default route and gateway
+	for _, subnet := range nwInfo.Subnets {
+		prefix := subnet.Prefix.String()
+		gateway := subnet.Gateway.String()
+
+		ip, _, er := net.ParseCIDR(prefix)
+		if er != nil {
+			return fmt.Errorf("[net] failed to parse prefix %s", prefix) // nolint
```
ashvindeodhar, line 215: include err in the return value

network/network_windows.go
```diff
@@ -185,6 +190,61 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+		}
+		if ip.To4() != nil {
+			// netsh interface ipv4 add route $subnetV4 $hostInterfaceAlias "0.0.0.0"
+			netshV4DefaultRoute := fmt.Sprintf(netRouteCmd, "ipv4", prefix, ifName, ipv4DefaultHop)
+			if out, err = nm.plClient.ExecuteCommand(netshV4DefaultRoute); err != nil {
```
ashvindeodhar, line 220: dont declare out and err at the beginning of the file. use := here

network/network_windows.go
```diff
@@ -185,6 +190,61 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+				log.Printf("[net] Adding ipv4 default route failed: %v:%v", out, err)
+			}
+
+			// netsh interface ipv4 add route $subnetV4 $hostInterfaceAlias $gatewayV4
+			netshV4GatewayRoute := fmt.Sprintf(netRouteCmd, "ipv4", prefix, ifName, gateway)
+			log.Printf("[net] Adding ipv4 gateway route failed: %v:%v", out, err)
```
ashvindeodhar, line 226: this log line is just incorrect. Looks like copy paste error

network/network_windows.go
```diff
@@ -185,6 +190,61 @@ func (nm *networkManager) newNetworkImplHnsV1(nwInfo *NetworkInfo, extIf *extern
...
+			if out, err = nm.plClient.ExecuteCommand(netshV4GatewayRoute); err != nil {
+				log.Printf("[net] Adding ipv4 gateway route failed: %v:%v", out, err)
+			}
+		} else {
```
ashvindeodhar, line 230: let's check 
ip.To16() != nil {

network/network_windows.go
```diff
@@ -337,6 +397,13 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *NetworkInfo, extIf *extern
 			}
 
 			log.Printf("[net] Successfully created hcn network with response: %+v", hnsResponse)
+
+			// only add net rules if it's dualStackOverlay mode and hnsNetwork is created at first time
+			if string(util.DualStackOverlay) != "" {
```
ashvindeodhar, line 402: how is this check ensuring that CNI is operating in dualstack overlay mode?

network/network_windows.go
```diff
@@ -337,6 +397,13 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *NetworkInfo, extIf *extern
...
+				if err := nm.addNewNetRules(nwInfo); err != nil { // nolint
+					return nil, err
```
ashvindeodhar, line 404: let's please log error here

network/network_windows.go
ashvindeodhar: +1 these are not idempotent. If the route is already present recreating the route will fail as it complains saying the object is already present.
@paulyufan2 as discussed earlier this week, you would need to check if the route exists before creating it.

ashvindeodhar (changes requested):

cni/network/network.go
```diff
@@ -742,6 +764,10 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
...
+		epInfo.IPV6Mode = string(util.IpamMode(opt.nwCfg.IPAM.Mode))
```
ashvindeodhar, line 768: I think you should only set the epInfo.IPV6Mode to that value if the string(util.IpamMode(opt.nwCfg.IPAM.Mode)) is dualStackOverlay. This will prevent setting the IPV6Mode if we are not operating in dualstack mode

network/network_windows.go
paulyufan2: ip, _, er := net.ParseCIDR(prefix)
ParseCIDR already handled whether prefix is valid or not, so we do not need check ip.To16() here

network/network_windows.go
paulyufan2: use same format as appIPV6RouteEntry() here

network/network_windows.go
paulyufan2: if it's not DualStackOverlay mode (util.DualStackOverlay == "")

cni/network/invoker_cns_test.go
tamilmani1989: i guess its testing failure case when both new and old api fails. @paulyufan you should name it as "Test Delete with both new and old api to release ip fails"

cni/network/invoker_cns_test.go
```diff
@@ -66,26 +74,114 @@ func TestCNSIPAMInvoker_Add(t *testing.T) {
...
+						Dst: network.Ipv4DefaultRouteDstPrefix,
+						GW:  net.ParseIP("10.0.0.1"),
+					},
+				},
+			},
+			want1:   nil,
+			wantErr: false,
+		},
+		{
+			name: "Test happy CNI add ipv4 and ipv6",
```
tamilmani1989, line 138: can we add test for happy delete path as well.?

cni/network/multitenancy_test.go
tamilmani1989: i dont think we are testing actual code paths but creating these mock errors and returning from that in mock function. @paulyufan is that right?

cni/network/network.go
```diff
@@ -628,19 +630,21 @@ func (plugin *NetPlugin) createNetworkInternal(
...
 
+	// parse the ipv6 address and only add it to nwInfo if it's dual stack mode
+	var podSubnetV6Prefix *net.IPNet
+	if ipamAddResult.ipv6Result != nil && len(ipamAddResult.ipv6Result.IPs) > 0 {
+		_, podSubnetV6Prefix, err = net.ParseCIDR(ipamAddResult.ipv6Result.IPs[0].Address.String())
+		if err != nil {
+			return nwInfo, fmt.Errorf("Failed to ParseCIDR for pod subnet IPv6 prefix: %w", err)
+		}
+	}
+
```
tamilmani1989, line 641: can we move this and ipv4 parse as well to `addSubnetToNetworkInfo` function?

cni/network/network.go
```diff
@@ -742,6 +764,13 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
 	}
 
 	if opt.resultV6 != nil {
+		// inject ipv6 routes to Linux pod
+		ipamMode := string(util.IpamMode(opt.nwCfg.IPAM.Mode))
+		if ipamMode == "dualStackOverlay" {
+			epInfo.IPV6Mode = ipamMode
```
tamilmani1989, line 770: i guess instead of checking ipammode, we should set executionmode to dualstackswift and set that in `epInfo.ipamMode`

dropgz/build/cniTest.Dockerfile
```diff
@@ -20,6 +20,7 @@ COPY --from=azure-ipam /azure-ipam/*.conflist pkg/embed/fs
 COPY --from=azure-ipam /azure-ipam/bin/* pkg/embed/fs
 COPY --from=azure-vnet /azure-container-networking/cni/azure-$OS-swift.conflist pkg/embed/fs/azure-swift.conflist
 COPY --from=azure-vnet /azure-container-networking/cni/azure-$OS-swift-overlay.conflist pkg/embed/fs/azure-swift-overlay.conflist
+COPY --from=azure-vnet /azure-container-networking/cni/azure-$OS-swift-overlay-dualstack.conflist pkg/embed/fs/azure-swift-overlay-dualstack.conflist
```
tamilmani1989, line 23: we are moving away from copying conflist from dropgz instead cns should write conflist after NC creation

network/network_windows.go
tamilmani1989: so confusing as you define multiple variables with dualstack. why don't you use `epInfo.Ipv6mode`

tamilmani1989 (changes requested):

paulyufan2:
do not review this PR. Please use this PR: https://github.com/Azure/azure-container-networking/pull/1925/files

PR #1871, opened 2023-03-23
ci:[CNI] Load testing for cilium cni

vipul-21:
**Reason for Change**:
Add the load testing for cilium cni with aks cluster.
There are couple of scripts that does the testing.
1. deployment script - it creates around 240 pods on each node and scale down to zero and repeat it 10 times.
2. Stress test script - it checks the ips assigned to the pods are same as that in azure_endpoints json and cilium endpoints. It also restarts the systemd netword to check the connectivity works after that.

**Requirements**:

- [ X] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/scripts/scale_deployment.sh
```diff
@@ -0,0 +1,24 @@
...
+total_num_of_run=10
+echo "Total num of run $total_num_of_run"
+for ((i=1; i <= total_num_of_run; i++))
+do 
+    echo "Current Run: $i"
+    scale_up_of_pods=2400
+    scale_down_pods=0
+    echo "Scaling pods to : $scale_up_of_pods"
+    kubectl scale deployment container --replicas $scale_up_of_pods
+    sleep 2m
```
tamilmani1989, line 15: instead of sleep, check deployment status to confirm all pods are up

cni/scripts/scale_deployment.sh
```diff
@@ -0,0 +1,24 @@
...
+
+    echo "Scaling down pods to : $scale_down_pods"
+    kubectl scale deployment container --replicas $scale_down_pods
+    sleep 2m
```
tamilmani1989, line 19: same here

cni/scripts/scale_deployment.sh
```diff
@@ -0,0 +1,24 @@
...
+done
+
+scale_up_of_pods=2400
+kubectl scale deployment container --replicas $scale_up_of_pods
+sleep 10m
```
tamilmani1989, line 24: same

tamilmani1989 (changes requested):

cni/scripts/stress_test.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo "Current : $node"
+    node_name="${node##*/}"
+    node_ip=$(kubectl get "$node"  -o jsonpath='{$.status.addresses[?(@.type=="InternalIP")].address}')
+    echo "Node internal ip: $node_ip"
+    privileged_pod=$(kubectl get pods -n kube-system -o wide | grep privileged | grep $node_name | awk '{print $1}')
+    echo "privileged pod : $privileged_pod"
+    until cat azure_endpoints.json
+    do
+        kubectl exec -i $privileged_pod -n kube-system -- bash -c "cat /var/run/azure-cns/azure-endpoints.json" > azure_endpoints.json
+        sleep 10
```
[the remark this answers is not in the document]
vipul-21: The kubectl commands fails sometimes with i/o timeout and does not respond immediately so i added a buffer between two calls( failure and hopefully the next successful call)

cni/scripts/stress_test.sh
```diff
@@ -0,0 +1,101 @@
...
+        if [[ $? -eq 1 ]]; then
+            printf "Error: %s Not found in the cilium_endpoints.json" "$ip" >&2
+            exit 1
+        fi
+    done
+
+    kubectl exec -i $privileged_pod -n kube-system -- bash -c "chroot /host /bin/bash -c 'systemctl restart systemd-networkd'"
+    rm -rf cilium_endpoints.json
+    rm -rf azure_endpoints.json
+
```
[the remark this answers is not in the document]
vipul-21: Should I add this to the stress test ? Or as a separate one

cni/scripts/scale_deployment.sh
vipul-21: Updated the PR

cni/scripts/scale_deployment.sh
vipul-21: Updated the PR

cni/scripts/stress_test.sh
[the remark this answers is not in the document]
vipul-21: Will be added as a separate stage so created a different task for it.

.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -0,0 +1,147 @@
...
+                set -ex
+                az extension add --name aks-preview
+                az group create --location $(LOCATION) --name $(RESOURCE_GROUP) --subscription $(SUBSCRIPTION_ID)
+                az account set -s $(SUBSCRIPTION_ID)
+                # Create a vnet
+                az network vnet create -g $(RESOURCE_GROUP) -n overlayvnet --address-prefixes 10.0.0.0/8 --output none
+                subnet=$(az network vnet subnet create -g $(RESOURCE_GROUP) --vnet-name overlayvnet --name nodenet --address-prefix 10.10.0.0/16 -o none --query '{id: id}' --output tsv)
+                # create AKS Cluster
+                az aks create -n $(RESOURCE_GROUP) -g $(RESOURCE_GROUP) -l $(LOCATION) --node-count 10 --network-plugin azure --network-plugin-mode overlay --pod-cidr 192.168.0.0/16 --vnet-subnet-id $subnet -s Standard_D8_v3 --enable-cilium-dataplane --no-ssh-key --yes
+                az aks get-credentials -n $(RESOURCE_GROUP) -g $(RESOURCE_GROUP)
```
rbtr, line 37: can you replace these infra scripts with usage of the [hack Makefile](https://github.com/Azure/azure-container-networking/blob/master/hack/swift/Makefile)? This helps us centralize and standardize how clusters are created and reduces maintenance burden when there are changes such as the flag `--enable-cilium-dataplane` -> `--network-dataplane=cilium`

.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -0,0 +1,147 @@
...
+  - stage: validate_state
+    dependsOn: pod_deployment
+    displayName: "Validate State"
+    jobs:
+      - job: validate_state
+        steps:
+          - script: |
+              set -ex        
+              echo "installing kubectl"
+              curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
rbtr, line 67: Do the runners not have kubectl installed already?

.pipelines/cni/cilium/cilium-cni-load-test.yaml
rbtr: ref https://github.com/Azure/azure-container-networking/blob/master/.pipelines/singletenancy/cilium/cilium-e2e-step-template.yaml, this pipeline should look very similar to that one

.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -0,0 +1,147 @@
...
+              sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+              sudo apt-get install jq
```
rbtr, line 69: same question

rbtr (changes requested):

.pipelines/cni/cilium/cilium-cni-load-test.yaml
vipul-21: Sure, let me update it.

.pipelines/cni/cilium/cilium-cni-load-test.yaml
vipul-21: I don't think there is.
I see it is being installed here as well: https://github.com/Azure/azure-container-networking/blob/master/.pipelines/singletenancy/cilium/cilium-e2e-step-template.yaml#L43-L45
Let me check again though.

cni/scripts/validate_state.sh
```diff
@@ -0,0 +1,96 @@
...
+        sleep 10
+    done
+
+    total_pods=$(kubectl get pods --all-namespaces -o wide --field-selector spec.nodeName=$node_name,status.phase=Running --output json)
+
+    echo "Checking if there are any pods with no ips"
+    pods_with_no_ip=$(echo $total_pods | jq -j '(.items[] | select(.status.podIP == "" or .status.podIP == null))')
+    if [ "$pods_with_no_ip" != "" ]; then
+        echo "There are some pods with no ip assigned."
+        kubectl get pods -A
```
wedaly, line 41: maybe `-o wide` here too so we can see the node + IP

cni/scripts/validate_state.sh
```diff
@@ -0,0 +1,96 @@
...
+    for ip in "${pod_ips[@]}"
+    do
+        find_in_array "$cilium_endpoint_ips" "$ip" "cilium_endpoints.json"
+        if [[ $? -eq 1 ]]; then
+            printf "Error: %s Not found in the cilium_endpoints.json" "$ip" >&2
+            exit 1
+        fi
+    done
+
+    kubectl exec -i $privileged_pod -n kube-system -- bash -c "chroot /host /bin/bash -c 'systemctl restart systemd-networkd'"
```
wedaly, line 93: Maybe add a comment here explaining (1) why we're restarting systemd-networkd (link to Cilium issue https://github.com/cilium/cilium/issues/18706 ) and (2) that there's a later step in the load test pipeline that runs a connectivity test

.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -0,0 +1,147 @@
...
+              inlineScript: |
+                set -ex
+                az extension add --name aks-preview
+                az group create --location $(LOCATION) --name $(RESOURCE_GROUP) --subscription $(SUBSCRIPTION_ID)
+                az account set -s $(SUBSCRIPTION_ID)
+                # Create a vnet
+                az network vnet create -g $(RESOURCE_GROUP) -n overlayvnet --address-prefixes 10.0.0.0/8 --output none
+                subnet=$(az network vnet subnet create -g $(RESOURCE_GROUP) --vnet-name overlayvnet --name nodenet --address-prefix 10.10.0.0/16 -o none --query '{id: id}' --output tsv)
+                # create AKS Cluster
+                az aks create -n $(RESOURCE_GROUP) -g $(RESOURCE_GROUP) -l $(LOCATION) --node-count 10 --network-plugin azure --network-plugin-mode overlay --pod-cidr 192.168.0.0/16 --vnet-subnet-id $subnet -s Standard_D8_v3 --enable-cilium-dataplane --no-ssh-key --yes
```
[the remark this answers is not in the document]
wedaly: also, please replace `--enable-cilium-dataplane` with `--network-dataplane=cilium` (both flags work, but the first one is deprecated now)

cni/scripts/validate_state.sh
vipul-21: Make sense, sure will add it. Thanks!

cni/scripts/validate_state.sh
vipul-21: Okay

cni/scripts/validate_state.sh
```diff
@@ -0,0 +1,96 @@
...
+    until cat azure_endpoints.json
+    do
+        kubectl exec -i $privileged_pod -n kube-system -- bash -c "cat /var/run/azure-cns/azure-endpoints.json" > azure_endpoints.json
+        sleep 10
+    done
+
+    cilium_agent=$(kubectl get pod -l k8s-app=cilium -n kube-system -o wide | grep $node_name | awk '{print $1}')
+    echo "cilium agent : $cilium_agent"
+    
+    until cat cilium_endpoints.json
```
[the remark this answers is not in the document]
vipul-21: Not an expert too.
1.  Yes, it will create an empty file.
2. Yes, you are right it will continue even if the file is empty. I'll add a condition to check if the file is empty or not.

.pipelines/cni/cilium/cilium-cni-load-test.yaml
vipul-21: You are right we don't need it.

.pipelines/cni/cilium/cilium-cni-load-test.yaml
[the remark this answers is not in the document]
vipul-21: Update it in the make file

hack/swift/Makefile
```diff
@@ -94,11 +96,11 @@ overlay-byocni-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster
 
 overlay-cilium-up: rg-up overlay-net-up ## Brings up an Overlay AzCNI cluster
 	$(AZCLI) aks create -n $(CLUSTER) -g $(GROUP) -l $(REGION) \
-		--node-count 2 \
+		--node-count $(NODE_COUNT) \
 		--node-vm-size $(VM_SIZE) \
 		--load-balancer-sku basic \
 		--network-plugin azure \
-		--enable-cilium-dataplane \
+		--network-dataplane cilium \
```
tamilmani1989, line 103: did this work? i think azcli support not released for --network-dataplane

vipul-21: Yes it works.

hack/scripts/validate_state.sh
```diff
@@ -0,0 +1,105 @@
...
+            num_of_pod_ips=$((num_of_pod_ips+1))
+        fi
+    done
+    echo "Number of pods running with ip assigned $num_of_pod_ips"
+
+    num_of_azure_endpoint_ips=$( cat azure_endpoints.json | jq -r '[.Endpoints | .[] | .IfnameToIPMap.eth0.IPv4[0].IP] | length' )
+    azure_endpoint_ips=$( cat azure_endpoints.json | jq -r '(.Endpoints | .[] | .IfnameToIPMap.eth0.IPv4[0].IP) ' )
+    echo "Number of azure endpoint ips : $num_of_azure_endpoint_ips"
+
+    if [ "$num_of_pod_ips" != "$num_of_azure_endpoint_ips" ]; then
```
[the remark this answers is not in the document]
vipul-21: We need an update in CNS to output json.

hack/swift/Makefile
rbtr: can you update the other up targets with these changes also? keep them all consistent

.pipelines/cni/cilium/cilium-cni-load-test.yaml
rbtr: still wondering ^ is jq not on the runner already?

hack/swift/Makefile
vipul-21: Sure

hack/scripts/validate_state.sh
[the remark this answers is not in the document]
vipul-21: So the cns in memory command use the API which we can use directly. The API returns in json so we could use that. Will add that as a separate task.
cc: @rbtr @tamilmani1989

.pipelines/cni/cilium/cilium-cni-load-test.yaml
vipul-21: Yeah, it is already on the runner. Removed it.

tamilmani1989 (dismissed):

tamilmani1989:
please make sure you address/resolve other comments as well

hack/scripts/validate_state.sh
```diff
@@ -0,0 +1,105 @@
...
+    cilium_agent=$(kubectl get pod -l k8s-app=cilium -n kube-system -o wide | grep "$node_name" | awk '{print $1}')
+    echo "cilium agent : $cilium_agent"
+    
+    while ! [ -f "cilium_endpoints.json" ]
+    do
+        echo "trying to get the cilium_endpoints"
+        kubectl exec -i "$cilium_agent" -n kube-system -- bash -c "cilium endpoint list -o json" > cilium_endpoints.json
+        if [ -f "cilium_endpoints.json" ] && [ ! -s "cilium_endpoints.json" ]; then
+            rm cilium_endpoints.json
+        fi
```
wedaly, line 39: I think you could simplify this by moving the "is file empty" check into the `while` condition:
```
while [ ! -f "cilium_endpoints.json" ] || [ ! -s "cilium_endpoints.json" ]; do
    ...
    # no need to `rm` cilium_endpoints.json, kubectl will overwrite it on retry anyway
done
```

hack/scripts/validate_state.sh
```diff
@@ -0,0 +1,105 @@
...
+        if [[ $? -eq 1 ]]; then
+            printf "Error: %s Not found in the cilium_endpoints.json" "$ip" >&2
+            exit 1
+        fi
+    done
+
+    #We are restarting the systmemd network and checking that the connectivity works after the restart. For more details: https://github.com/cilium/cilium/issues/18706
+    kubectl exec -i "$privileged_pod" -n kube-system -- bash -c "chroot /host /bin/bash -c 'systemctl restart systemd-networkd'"
+    rm -rf cilium_endpoints.json
+    rm -rf azure_endpoints.json
```
wedaly, line 104: nit: can combine these into one line
```
rm -rf cilium_endpoints.json azure_endpoints.json
```

wedaly (approved):

.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -0,0 +1,120 @@
...
+              scriptLocation: "inlineScript"
+              scriptType: "bash"
+              addSpnToEnvironment: true
+              inlineScript: |
+                set -ex
+                if [ "$(DELETE_RESOURCES)" ]
+                then
+                  echo "Deleting Cluster and resource group"
+                  make -C ./hack/swift set-kubeconf AZCLI=az CLUSTER=${RESOURCE_GROUP}
+                  make -C ./hack/swift azcfg AZCLI=az REGION=$(LOCATION)
```
rbtr, line 111: don't need these to bring the cluster down

.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -0,0 +1,120 @@
...
+        steps:
+          - script: |
+              echo "install cilium CLI"
+              CILIUM_CLI_VERSION=$(curl -s https://raw.githubusercontent.com/cilium/cilium-cli/master/stable.txt)
+              CLI_ARCH=amd64
+              if [ "$(uname -m)" = "aarch64" ]; then CLI_ARCH=arm64; fi
+              curl -L --fail --remote-name-all https://github.com/cilium/cilium-cli/releases/download/${CILIUM_CLI_VERSION}/cilium-linux-${CLI_ARCH}.tar.gz{,.sha256sum}
+              sha256sum --check cilium-linux-${CLI_ARCH}.tar.gz.sha256sum
+              sudo tar xzvfC cilium-linux-${CLI_ARCH}.tar.gz /usr/local/bin
+              rm cilium-linux-${CLI_ARCH}.tar.gz{,.sha256sum}
```
rbtr, line 73: their script needs to be flexible, but we know what arch and version we want - fix all of these variables so that we get the same thing every time

hack/manifests/pod.yaml
```diff
@@ -0,0 +1,21 @@
...
+  template:
+    metadata:
+      labels:
+        app: container
+    spec:
+      containers:
+        - name: ubuntu
+          image: mcr.microsoft.com/mirror/docker/library/ubuntu:22.04
+          imagePullPolicy: Always
+          command: ["/bin/sleep", "3650d"]
```
rbtr, line 19: if you just want a no-op pod, use `mcr.microsoft.com/oss/kubernetes/pause:3.6`. it will pull faster, doesn't need a sleep command, and only does that one thing

hack/scripts/scale_deployment.sh
```diff
@@ -0,0 +1,37 @@
...
+    echo "Scaling pods to : $scale_up_of_pods"
+    kubectl scale deployment container --replicas $scale_up_of_pods
+    check_deployment $scale_up_of_pods
+    echo "Scaling down pods to : $scale_down_pods"
+    kubectl scale deployment container --replicas $scale_down_pods
+    check_deployment $scale_down_pods
+done
+
+kubectl scale deployment container --replicas $scale_up_of_pods
+check_deployment $scale_up_of_pods
```
rbtr, line 37: configure your editor to add newlines to the end of files on save

hack/swift/Makefile
```diff
@@ -134,11 +137,11 @@ swift-byocni-up: rg-up swift-net-up ## Bring up a SWIFT BYO CNI cluster
 
 swift-cilium-up: rg-up swift-net-up ## Bring up a SWIFT Cilium cluster
 	$(AZCLI) aks create -n $(CLUSTER) -g $(GROUP) -l $(REGION) \
-		--node-count 2 \
+		--node-count $(NODE_COUNT) \
 		--node-vm-size $(VM_SIZE) \
 		--load-balancer-sku basic \
 		--network-plugin azure \
-		--enable-cilium-dataplane \
+		--network-dataplane cilium \
```
rbtr, line 144: thanks for making these changes in all of the targets

hack/scripts/validate_state.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
rbtr, line 1: it may be late to bring this up, but...
bash has a high barrier to entry and isn't really suited to any complex control flow. is there a reason you scripted this instead of adding it to the Go tests in `/test/integration`, or copying that pattern?

vipul-21: I agree to this and we thought of going that path. 
We have a task to add it as Go tests but since we started this PR when we were going for GA and based on the timeline the effort seemed less as a bash script.

I'll check with tamilmani if we should work on the pipeline as go tests now since GA has been pushed.
Thanks for bringing it up.

hack/manifests/pod.yaml
vipul-21: Yeah just wanted to have no-op. Thanks for the image.

.pipelines/cni/cilium/cilium-cni-load-test.yaml
vipul-21: You mean hardcode the cli version and arch version ?

hack/scripts/validate_state.sh
vipul-21: @rbtr So had a chat with tamilmani. We have a task for adding the tests for CNI as go tests as mentioned and we will be working on them in the April sprint. 
If it is not a hard blocker, can move ahead with the bash script approach ?

.pipelines/cni/cilium/cilium-cni-load-test.yaml
rbtr: you _might_ need the azcfg target, but you definitely don't need the kubeconfig

.pipelines/cni/cilium/cilium-cni-load-test.yaml
rbtr: yes, we should pin those so that the CI is consistent

hack/scripts/validate_state.sh
rbtr: not a blocker

rbtr (approved):

rbtr:
needs rebase

hack/scripts/validate_state.sh
```diff
@@ -0,0 +1,98 @@
...
+
+    cilium_agent=$(kubectl get pod -l k8s-app=cilium -n kube-system -o wide | grep "$node_name" | awk '{print $1}')
+    echo "cilium agent : $cilium_agent"
+    
+    while ! [ -s "cilium_endpoints.json" ]
+    do
+        echo "trying to get the cilium_endpoints"
+        kubectl exec -i "$cilium_agent" -n kube-system -- bash -c "cilium endpoint list -o json" > cilium_endpoints.json
+        sleep 10
+    done
```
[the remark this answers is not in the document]
vipul-21: The default timeout is 60 mins.

hack/manifests/hostprocess.yaml
```diff
@@ -0,0 +1,38 @@
...
+    metadata:
+      labels:
+        app: privileged-daemonset
+    spec:
+      hostNetwork: true
+      hostPID: true
+      containers:
+        - name: privileged-container
+          image: mcr.microsoft.com/dotnet/runtime-deps:6.0
+          command: ["/bin/sleep", "3650d"]
```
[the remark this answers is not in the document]
vipul-21: Will need to check, let me try to use the image and see if that works.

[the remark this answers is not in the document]
vipul-21: I tried to use the image but it fails with the error of bash not in PATH. 
` "bash": executable file not found in $PATH:`

wedaly (approved):

.pipelines/cni/cilium/cilium-cni-load-test.yaml
vipul-21: I checked removing the set kubeconfig and it failed with:
`error: cannot delete cluster cilium-vipul-test, not in /home/vsts/.kube/config`

Issue #1614, opened 2022-09-14
Azure-NPM Windows: Occasional Latency when Applying Network Policies


huntergregory:
This is a known issue in HNS as seen with other implementations of Network Policy in Windows. It can take one minute or more to apply Policy rules to the kernel.

Two fixes:
- HNS improvements
- NPM optimizations including either async or parallel calls to HNS.

huntergregory:
update: 
- Investigations underway for how to address limitations in HNS
- Plans made for NPM optimizations (first improvement ETA January)

huntergregory:
Windows HNS has made a fix. Will close this issue once released

huntergregory:
NPM and HNS improvements have been released

PR #1875, opened 2023-03-28
perf: [WIN-NPM] apply ipsets in background

huntergregory:
**Reason for Change**:
Makes IPSet SysCalls 96% (23x) faster when there are no NetworkPolicies to process. Further optimized in #1900.

**Issue Fixed**:
First of three PRs addressing Windows limitations in #1614.

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
### Experiment
1.5K Pods, 10 NetworkPolicies

### Results
- Old NPM: 3 hours to steady state
- This PR: 8 minutes to steady state


npm/azure-npm.yaml
```diff
@@ -148,6 +148,9 @@ metadata:
 data:
   azure-npm.json: |
     {
+        "WindowsNetworkName": "azure",
```
vakalapa, line 151: We can keep this network name for simplicity.

npm/config/config.go
```diff
@@ -16,8 +16,10 @@ const (
 
 // DefaultConfig is the guaranteed configuration NPM can run in out of the box
 var DefaultConfig = Config{
-	WindowsNetworkName:    util.AzureNetworkName,
-	ResyncPeriodInMinutes: defaultResyncPeriod,
+	WindowsNetworkName:                   util.AzureNetworkName,
+	ApplyDataPlaneMaxBatches:             0,
```
vakalapa, line 20: make sure these default are real numbers, if config map is not applied properly we should not be erroring out.

npm/pkg/dataplane/dataplane.go
```diff
@@ -464,3 +564,7 @@ func (dp *DataPlane) deleteIPSetsAndReferences(sets []*ipsets.TranslatedIPSet, n
 	}
 	return nil
 }
+
+func (dp *DataPlane) configuredToApplyInBackground() bool {
+	return util.IsWindowsDP() && dp.ApplyDataPlaneMaxBatches > 0 && dp.ApplyDataPlaneInterval > 0
```
vakalapa, line 569: not a fan of this, we should add a new flag which defines this new behavior.

vakalapa (commented):
Overall looks good, some comments on user experience, also, we will need to thoroughly test these changes

huntergregory:
> also, we will need to thoroughly test these changes
Conformance and Cyclonus have succeeded for a few rounds (excluding Windows issues). Also finished multiple rounds of testing using these scripts: https://github.com/Azure/azure-container-networking/tree/master/test/scale

npm/pkg/dataplane/dataplane.go
```diff
@@ -451,3 +547,7 @@ func (dp *DataPlane) deleteIPSetsAndReferences(sets []*ipsets.TranslatedIPSet, n
 	}
 	return nil
 }
+
+func (dp *DataPlane) configuredToApplyInBackground() bool {
+	return util.IsWindowsDP() && dp.ApplyInBackground
```
vakalapa, line 552: Instead of checking for isWindowsDP everytime we need to applyDP, can we assign a global DP variable which we can set once in INIT and then only keep checking that value ?

vakalapa (changes requested):

npm/azure-npm.yaml
```diff
@@ -151,11 +151,15 @@ data:
...
+        "ApplyIntervalInMilliseconds": 500,
         "Toggles": {
             "EnablePrometheusMetrics": true,
             "EnablePprof":             true,
             "EnableHTTPDebugAPI":      true,
             "EnableV2NPM":             true,
-            "PlaceAzureChainFirst":    true
+            "PlaceAzureChainFirst":    true,
+            "ApplyIPSetsOnNeed":       false,
+            "ApplyInBackground": true
```
vakalapa, line 163: <nit> formatting of space.

vakalapa (approved):

PR #1883, opened 2023-03-31
chore:[Telemetry] Prevent telemetry connect if no binary path

jpayne3506:
**Reason for Change**:

Prevents telemetry connection if the binary does not exist.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/telemetry/aiwrapper.go
+++ b/telemetry/aiwrapper.go
@@ -3,6 +3,7 @@ package telemetry
 
 import (
 	"fmt"
+
 	"github.com/Azure/azure-container-networking/aitelemetry"
 	"github.com/Azure/azure-container-networking/log"
 )
@@ -24,7 +25,7 @@ func CreateAITelemetryHandle(aiConfig aitelemetry.AIConfig, disableAll, disableM
 
 	if disableAll {
 		log.Printf("Telemetry is disabled")
-		return fmt.Errorf("Telmetry disabled")
+		return fmt.Errorf("Telemetry disabled")
 	}
 
 	th, err = aitelemetry.NewAITelemetry("", aiMetadata, aiConfig)
--- a/telemetry/telemetrybuffer.go
+++ b/telemetry/telemetrybuffer.go
@@ -303,8 +303,15 @@ func ReadConfigFile(filePath string) (TelemetryConfig, error) {
 
 // ConnectToTelemetryService - Attempt to spawn telemetry process if it's not already running.
 func (tb *TelemetryBuffer) ConnectToTelemetryService(telemetryNumRetries, telemetryWaitTimeInMilliseconds int) {
-	path, dir := getTelemetryServiceDirectory()
+
+	path, dir, fileExists := getTelemetryServiceDirectory()
 	args := []string{"-d", dir}
+
+	if fileExists {
+		log.Logf("Skip starting telemetry service as file didn't exist")
+		return
+	}
+
 	for attempt := 0; attempt < 2; attempt++ {
 		if err := tb.Connect(); err != nil {
 			log.Logf("Connection to telemetry socket failed: %v", err)
@@ -319,14 +326,15 @@ func (tb *TelemetryBuffer) ConnectToTelemetryService(telemetryNumRetries, teleme
 	}
 }
 
-func getTelemetryServiceDirectory() (path string, dir string) {
+func getTelemetryServiceDirectory() (path string, dir string, fileExists bool) {
 	path = fmt.Sprintf("%v/%v", CniInstallDir, TelemetryServiceProcessName)
+	fileExists = true
 	if exists, _ := platform.CheckIfFileExists(path); !exists {
 		ex, _ := os.Executable()
 		exDir := filepath.Dir(ex)
 		path = fmt.Sprintf("%v/%v", exDir, TelemetryServiceProcessName)
 		if exists, _ = platform.CheckIfFileExists(path); !exists {
-			log.Logf("Skip starting telemetry service as file didn't exist")
+			fileExists = false
 			return
 		}
 		dir = exDir


telemetry/telemetrybuffer.go
```diff
@@ -303,8 +303,15 @@ func ReadConfigFile(filePath string) (TelemetryConfig, error) {
...
 func (tb *TelemetryBuffer) ConnectToTelemetryService(telemetryNumRetries, telemetryWaitTimeInMilliseconds int) {
-	path, dir := getTelemetryServiceDirectory()
+
+	path, dir, fileExists := getTelemetryServiceDirectory()
 	args := []string{"-d", dir}
+
+	if fileExists {
+		log.Logf("Skip starting telemetry service as file didn't exist")
+		return
+	}
```
rbtr, line 313: is this conditional backwards?

telemetry/telemetrybuffer.go
```diff
@@ -319,14 +326,15 @@ func (tb *TelemetryBuffer) ConnectToTelemetryService(telemetryNumRetries, teleme
 	}
 }
 
-func getTelemetryServiceDirectory() (path string, dir string) {
+func getTelemetryServiceDirectory() (path string, dir string, fileExists bool) {
```
rbtr, line 329: I think it should be the responsibility of the caller to check if the returned path exists, and this method should only be responsible for composing the path from the config

telemetry/telemetrybuffer.go
```diff
@@ -303,8 +303,15 @@ func ReadConfigFile(filePath string) (TelemetryConfig, error) {
 
 // ConnectToTelemetryService - Attempt to spawn telemetry process if it's not already running.
 func (tb *TelemetryBuffer) ConnectToTelemetryService(telemetryNumRetries, telemetryWaitTimeInMilliseconds int) {
-	path, dir := getTelemetryServiceDirectory()
+
+	path, dir, fileExists := getTelemetryServiceDirectory()
 	args := []string{"-d", dir}
+
+	if fileExists {
```
tamilmani1989, line 310: it should not be added here.. its after connect failure and before starting new telemetry service process

tamilmani1989 (changes requested):

telemetry/telemetrybuffer.go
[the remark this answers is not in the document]
vipul-21: nit: Also, it would be helpful to add a comment that describes what the function does. Since it is checking the file at couple of places, would be nice to have an  explanation.

telemetry/telemetrybuffer.go
jpayne3506: It is backwards. I will change it.

telemetry/telemetrybuffer.go
```diff
@@ -303,8 +303,15 @@ func ReadConfigFile(filePath string) (TelemetryConfig, error) {
 
 // ConnectToTelemetryService - Attempt to spawn telemetry process if it's not already running.
 func (tb *TelemetryBuffer) ConnectToTelemetryService(telemetryNumRetries, telemetryWaitTimeInMilliseconds int) {
-	path, dir := getTelemetryServiceDirectory()
+
```
vipul-21, line 306: Let's remove the extra lines(whitespaces) I guess that's why linter is complaining.

telemetry/aiwrapper.go
```diff
@@ -24,7 +25,7 @@ func CreateAITelemetryHandle(aiConfig aitelemetry.AIConfig, disableAll, disableM
 
 	if disableAll {
 		log.Printf("Telemetry is disabled")
-		return fmt.Errorf("Telmetry disabled")
+		return fmt.Errorf("Telemetry disabled")
```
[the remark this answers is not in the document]
jpayne3506: I corrected. Saw it as I was going through call hierarchies.

telemetry/telemetrybuffer.go
jpayne3506: It was supposed to be !fileExists. Does that resolve this?

telemetry/telemetrybuffer.go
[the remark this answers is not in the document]
jpayne3506: > Would it possible to write U.T for this function only ? I don't see any dependency on telemetry here and this would increase the code coverage as well.

I can give it bad data to ensure it fails, but I do not believe I can test happy paths without mock files in the proper directories.

telemetry/telemetrybuffer.go
jpayne3506: Will do.

telemetry/telemetrybuffer.go
jpayne3506: Talked to Evan and implemented agnostic calls for "Sprintf" and "platform". Changed the implementation to reflect method naming to prevent over responsibility.

telemetry/telemetrybuffer.go
jpayne3506: Moved check for binary to match request.

rbtr (changes requested):
use of error needs a little adjustment, everything else looks great

telemetry/aiwrapper.go
```diff
@@ -2,16 +2,19 @@
...
 var (
-	aiMetadata     string
-	th             aitelemetry.TelemetryHandle
-	gDisableTrace  bool
-	gDisableMetric bool
+	aiMetadata           string
+	th                   aitelemetry.TelemetryHandle
+	gDisableTrace        bool
+	gDisableMetric       bool
+	ErrTelemetryDisabled = errors.New("Telemetry is disabled")
```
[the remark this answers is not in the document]
jpayne3506: Changed

telemetry/aiwrapper.go
```diff
@@ -24,7 +27,7 @@ func CreateAITelemetryHandle(aiConfig aitelemetry.AIConfig, disableAll, disableM
 
 	if disableAll {
 		log.Printf("Telemetry is disabled")
-		return fmt.Errorf("Telmetry disabled")
+		return fmt.Errorf("%w", ErrTelemetryDisabled)
```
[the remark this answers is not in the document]
jpayne3506: Changed to 2 other calls to reflect WrapF preference.

telemetry/aiwrapper.go
```diff
@@ -2,16 +2,18 @@
...
 var (
-	aiMetadata     string
-	th             aitelemetry.TelemetryHandle
-	gDisableTrace  bool
-	gDisableMetric bool
+	aiMetadata           string
+	th                   aitelemetry.TelemetryHandle
+	gDisableTrace        bool
+	gDisableMetric       bool
+	ErrTelemetryDisabled = errors.New("telemetry is disabled")
```
vipul-21, line 16: The variable name should be `errTelemetryDisabled`

vipul-21 (changes requested):

telemetry/aiwrapper.go
rbtr: probably not, because then I can't compare to it with `errors.Is/As` from outside of this package

vipul-21: Oh got it. Thanks @rbtr

jpayne3506: Reverted back to ErrTelemetryDisabled

tamilmani1989 (approved):

vipul-21 (approved):

rbtr (approved):

PR #1885, opened 2023-04-03
Adding Pprof register method

ramiro-gamarra:
**Reason for Change**:

This PR enables pprof endpoints to be registered on the multi tenancy path.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/client/client_test.go
+++ b/cns/client/client_test.go
@@ -170,8 +170,8 @@ func TestMain(m *testing.M) {
 	config := common.ServiceConfig{}
 
 	httpRestService, err := restserver.NewHTTPRestService(&config, &fakes.WireserverClientFake{}, &fakes.WireserverProxyFake{}, &fakes.NMAgentClientFake{}, nil, nil, nil)
-	svc = httpRestService.(*restserver.HTTPRestService)
-	svc.Name = "cns-test-server"
+	svc = httpRestService
+	httpRestService.Name = "cns-test-server"
 	fakeNNC := v1alpha.NodeNetworkConfig{
 		TypeMeta:   metav1.TypeMeta{},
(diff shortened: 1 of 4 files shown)


timraymond (dismissed):
Seems reasonable to me

rbtr:
is it feasible/should we make this dynamically registerable? if we're trying to catch a memory leak, it might be nice to be able to turn the pprof handlers on, take the heap dump, and turn them back off without restarting the process.

ramiro-gamarra:
> is it feasible/should we make this dynamically registerable? if we're trying to catch a memory leak, it might be nice to be able to turn the pprof handlers on, take the heap dump, and turn them back off without restarting the process.

I believe we can still attach routes to the mux even after the server has started listening. Are you proposing:
1. client calls a CNS endpoint to enable pprof handlers, client calls pprof endpoint, client calls CNS endpoint to turn off pprof handlers.
2. client calls a CNS endpoint to collect pprof data, where CNS can internally use the pprof package without needing handlers.
3. something else.

I think while we decide on a more dynamic approach, it might be simpler to suggest to teams that use multi tenancy to run CNS with this turned on, since they own the ingra and there should be less risk for abuse of these endpoints.

timraymond:
@rbtr @ramiro-gamarra Why do we need to turn the pprof handlers on and off dynamically? There's no performance penalty until you actually collect a profile.

rbtr:
but there is _significant_ performance penalty when you do so best practice is to not leave pprof on in prod where someone could naively hit a dump endpoint and freeze the process for a minute

timraymond (approved):

PR #1887, opened 2023-04-04
fix: [WIN-NPM] allow readiness probes

### LINKED ISSUE #1881: Liveness/readiness probes time out with default deny all network policy on Windows 2022
**What happened:**  
There seems to be a difference between Linux and Windows in how liveness/readiness probes are allowed through NPM. We have an application that consists of both Windows and Linux services. Our cluster consists of multiple node pools and generally divided into two categories: Windows 2022 and Linux. 

We apply the following _default_ network policy in the namespace (that contains both Windows and Linux pods) to deny all traffic by default:

```yaml
kind: NetworkPolicy
metadata:
  name: default
spec:
  podSelector: { }
  policyTypes:
    - Ingress
    - Egress
```

Probes on Linux nodes work without any issue. However, on Windows the probes time out. When we remove the default network policy, the probes start to work on Windows as well.

I am not sure if the behaviour is wrong on Linux or Windows, but I cannot find any information on how to whitelist liveness/readiness probes in NPM configuration, so I am under the assumption that probes should work without any configuruation in a NetworkPolicy.

Probes work in a consistent manner between OS types. E.g. they should either work or not work on both Windows and Linux when a default deny policy is set. 

**How to reproduce it:**  
Add the following default NetworkPolicy to a namespace that has Windows pods:

```yaml
kind: NetworkPolicy
metadata:
  name: default
spec:
  podSelector: { }
  policyTypes:
    - Ingress
    - Egress
```

Add a liveness/readiness probe to the container in the namespace and observe probe time outs.

**Kubernetes Version:**  
v1.25.5

**Kernel (e.g. `uname -a`):**  
Not sure about the kernel version, but we are running the following node images:

Linux: AKSUbuntu-2204gen2containerd-2023.01.26
Windows: AKSWindows-2022-containerd-20348.1547.230223

huntergregory:
**Reason for Change**:
Allows readiness probes no matter the NetPol configuration. This is in-line with other NetworkPolicy managers including:
- Linux NPM
- Calico (Linux and Windows)
- Cilium

**Issue Fixed**:
Fixes #1881

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Notes**:


huntergregory:
## Testing
- 100 windows nodes
- 1 cx Pod on each node
- NPM image: acnpublic.azurecr.io/azure-npm:d4269faf

Add deny-all policy for these Pods.

1. NPM finds node IP correctly
2. NPM adds deny-all policy
3. Pods continue to run

npm/pkg/dataplane/policies/policymanager_windows.go
```diff
@@ -348,7 +349,7 @@ func getEPPolicyReqFromACLSettings(settings []*NPMACLPolSettings) (hcn.PolicyEnd
 }
 
 func getSettingsFromACL(policy *NPMNetworkPolicy) ([]*NPMACLPolSettings, error) {
-	hnsRules := make([]*NPMACLPolSettings, len(policy.ACLs))
+	hnsRules := make([]*NPMACLPolSettings, len(policy.ACLs)+1)
```
vakalapa, line 352: Can you add a comment here on why the +1

npm/pkg/dataplane/policies/policymanager_windows.go
```diff
@@ -357,6 +358,16 @@ func getSettingsFromACL(policy *NPMNetworkPolicy) ([]*NPMACLPolSettings, error)
 		}
 		hnsRules[i] = rule
 	}
+
+	// fixes #1881
+	hnsRules[len(policy.ACLs)] = &NPMACLPolSettings{
+		Id:        policy.ACLPolicyID,
+		Action:    hcn.ActionTypeAllow,
```
vakalapa, line 365: You are allowing all rules, you should add the source ip as node ip here right ? 
Add a comment on this too .

vakalapa (changes requested):

vakalapa (approved):

npm/pkg/dataplane/policies/policymanager.go
```diff
@@ -92,6 +94,11 @@ func (pMgr *PolicyManager) Bootup(epIDs []string) error {
 		// update Prometheus metrics on success
 		metrics.IncNumACLRulesBy(numLinuxBaseACLRules)
 	}
+
+	if util.IsWindowsDP() && pMgr.NodeIP == "" {
+		return npmerrors.Errorf(npmerrors.BootupPolicyMgr, false, "policy manager must have a configured nodeIP in Windows")
```
vakalapa, line 99: This return error might be fatal, but should be okay for time being. We will need to make our getnodeip logic robust in case we see any failures. 

Can put up a alert on this error message from telemetry.

npm/util/util.go
```diff
@@ -338,3 +338,16 @@ func IsIPV4(ip string) bool {
 
 	return address.Is4()
 }
+
+// Get preferred outbound ip of this machine
+// source: https://stackoverflow.com/questions/23558425/how-do-i-get-the-local-ip-address-in-go
+func NodeIP() (string, error) {
+	conn, err := net.Dial("udp", "8.8.8.8:80")
```
vakalapa, line 345: What if 8.8.8.8 is blocked by some NSGs or other way , will this still work ?

huntergregory: from the [stack overflow post](https://stackoverflow.com/questions/23558425/how-do-i-get-the-local-ip-address-in-go), it says that this actually does not actually create a connection since it is the udp protocol. Any IP address would work here

huntergregory: tested in the following way:
1. Made sure NPM gets the right node IP in 100-node cluster
2. On local Windows machine, got node IP by dialing 20 million different IPs
3. Added deny rules for 8.8.8.8:80 in AKS NSG

vakalapa (approved):

PR #1888, opened 2023-04-04
Remove redundant API in CNS that already in use in DNC

bohuini:
**Reason for Change**:
Initially, the API `GetAllNCsFromCns()` and `PostAllNetworkContainers()` were intended to add on CNS side. But later, because of some feature was abandoned on DNC side, these two API were added and used on DNC side eventually.  Hence, remove these from CNS.

**Issue Fixed**:
Removed unused api in CNS.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


timraymond:
I actually don't think leaving this is in place is a _bad_ thing. The API exists in CNS and is used... we just implemented the same methods on another client in addition to this one. Leaving these methods in place would allow CNS clients to _use_ the API that we already support and provide.

I would think differently if this was an API that was deprecated, or if this was the handler for some unused API. Neither is true though, so I think we're just removing functionality that we've already paid for in time & effort.

bohuini:
@timraymond  Make sense to me. I'll abandon this PR.

PR #1889, opened 2023-04-04
Fix incorrect error "UnsupportedAPI" returned by CNI

rjdenney:
**Reason for Change**:

This PR is to fix DualNIC log issue:
(1) If new CNS API(GetAllNetworkContainers) is failed to get ncConfigs, then it will try old CNS API(GetNetworkContainer)
(2) If old CNS API is also failed to get called, then it should return corresponding error instead of "Unsupported API" error

This PR addes three UT test cases:
1. TestGetMultiTenancyCNIResultUnsupportedAPI() => This will test if new CNS API is not supported and old CNS API can handle to get ncConfig with "Unsupported API" error
2. TestGetMultiTenancyCNIResultNotFound() => this test case includes two sub test cases:
   // 1. CNS supports new API and it does not have orchestratorContext info
   // 2. CNS does not support new API and it does not have orchestratorContext info

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/network/multitenancy_test.go
```diff
@@ -66,6 +70,18 @@ func (c *MockCNSClient) GetNetworkContainer(ctx context.Context, orchestratorCon
 }
 
 func (c *MockCNSClient) GetAllNetworkContainers(ctx context.Context, orchestratorContext []byte) ([]cns.GetNetworkContainerResponse, error) {
+	// check if getAllNetworkContainersConfiguration orchestratorContext is nil using GetAllNetworkContainers()
+	if c.getAllNetworkContainersConfiguration.orchestratorContext == nil {
+		// if orchestratorContext is same, attach CNSClientError code and error with Unsupported API and try old CNS API GetNetworkContainer() later
```
ashvindeodhar, line 75: if c.getAllNetworkContainersConfiguration.orchestratorContext is nil why is it being compared to orchestratorContext below?

cni/network/multitenancy_test.go
```diff
@@ -473,3 +489,272 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
...
+			if tt.wantErr {
+				require.Error(err)
+			}
+			require.NoError(err)
+			require.Exactly(tt.want, got[0].ncResponse)
+		})
+	}
+}
+
+func TestGetMultiTenancyCNIResultNotUnsupportedAPIError(t *testing.T) {
```
ashvindeodhar, line 628: what's the difference between this and above? The names of the funcs are not proving that clarity.

ashvindeodhar (changes requested):

tamilmani1989:
@paulyufan  Please add description to your PR describing what's the issue and what you are fixing. That is **must** before we do any review

rjdenney:
@tamilmani1989 Description has been updated.

cni/network/multitenancy_test.go
```diff
@@ -473,3 +500,288 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
...
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := tt.args.plugin.multitenancyClient.GetAllNetworkContainers(
+				tt.args.ctx,
+				tt.args.nwCfg,
+				tt.args.k8sPodName,
+				tt.args.k8sNamespace,
+				tt.args.ifName)
+			if (err != nil) != tt.wantErr {
```
tamilmani1989, line 781: the test case missing validating the actual error. we should validate if GetAllNetworkcontainers returning expected error message

paulyufan2: fixed

cni/network/multitenancy_test.go
```diff
@@ -42,7 +46,19 @@ type getAllNetworkContainersConfigurationHandler struct {
...
+	GetAllNetworkContainers cnsAPIName = "GetAllNetworkContainers"
+)
+
+var (
+	errUnsupportedAPI             = errors.New("Unsupported API")
+	errNoOrchestratorContextFound = errors.New("No CNI OrchestratorContext Found")
+)
+
 type MockCNSClient struct {
+	unsupportedAPIs                      map[cnsAPIName]bool
```
ashvindeodhar, line 61: why do we need bool? just empty struct would do, no?

ashvindeodhar: all we need to check if the key exists. We don;t really need a true / false value in the map

cni/network/multitenancy_test.go
```diff
@@ -473,3 +489,272 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
 		})
 	}
 }
+
+func TestGetMultiTenancyCNIResultUnsupportedAPI(t *testing.T) {
```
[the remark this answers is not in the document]
ashvindeodhar: correct godoc should be -
// TestGetMultiTenancyCNIResultUnsupportedAPI tests ...

cni/network/multitenancy_test.go
```diff
@@ -473,3 +500,292 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
...
+			{
+				IPAddress:        "10.1.0.0/16",
+				GatewayIPAddress: "10.1.0.1",
+			},
+		},
+	}
+
+	// set new CNS API is not supported
+	unsupportedAPIs := make(map[cnsAPIName]bool)
+	unsupportedAPIs["GetAllNetworkContainers"] = true
```
ashvindeodhar, line 541: just need to add empty struct struct{}{}

cni/network/multitenancy_test.go
```diff
@@ -473,3 +500,292 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
...
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.args.plugin.multitenancyClient.GetAllNetworkContainers(
+				tt.args.ctx,
+				tt.args.nwCfg,
+				tt.args.k8sPodName,
+				tt.args.k8sNamespace,
+				tt.args.ifName)
+			if (err != nil) != tt.wantErr {
```
ashvindeodhar, line 632: can you pls add more comments indicating what is expected, etc?

ashvindeodhar (changes requested):

cni/network/multitenancy_test.go
```diff
@@ -473,3 +500,293 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
...
+				k8sNamespace: "testnamespace1",
+				ifName:       "eth0",
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := tt.args.plugin.multitenancyClient.GetAllNetworkContainers(
```
ashvindeodhar, line 776: I think this is wrong - 
if you call this func with GetAllNetworkContainers func in unsupported list, it will return ApiNotSupported error.

cni/network/multitenancy_test.go
```diff
@@ -473,3 +500,293 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
...
+				tt.args.nwCfg,
+				tt.args.k8sPodName,
+				tt.args.k8sNamespace,
+				tt.args.ifName)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("GetAllNetworkContainers() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			if errors.Is(err, errNoOrchestratorContextFound) {
```
ashvindeodhar, line 787: what if the err is not errNoOrchestratorContextFound, it will still pass the test, right? Which is what is happening for the first comment I made

ashvindeodhar (changes requested):

cni/network/multitenancy_test.go
```diff
@@ -473,3 +500,293 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
...
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.args.plugin.multitenancyClient.GetAllNetworkContainers(
+				tt.args.ctx,
+				tt.args.nwCfg,
+				tt.args.k8sPodName,
+				tt.args.k8sNamespace,
+				tt.args.ifName)
+			if (err != nil) != tt.wantErr {
```
timraymond, line 632: This is a bit of a puzzling construction for future readers. [Clear is better than clever](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=14m35s). Prefer this:

```go
if err != nil && !tt.WantErr {
       t.Fatal("unexpected error: err:", err)
}

if err == nil && tt.WantErr {
       t.Fatal("expected an error but none received")
}
```

The first can be read literally: "if the error is not present and I expected one..." the second "if no error was present but I did expect one...".

cni/network/multitenancy_test.go
```diff
@@ -473,3 +500,293 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
...
+				// expect an error that GetAllNetworkContainers() fails to get all network containers
+				t.Errorf("GetAllNetworkContainers() error = %v, wantErr %v", err, tt.wantErr)
+				return
```
timraymond, line 635: Use `t.Fatal` instead of `t.Error` if you want to stop the test from proceeding. It has the same effect through some special runtime tricks. `return` is always a smell in tests.

cni/network/multitenancy_test.go
```diff
@@ -473,3 +500,293 @@ func TestGetMultiTenancyCNIResult(t *testing.T) {
...
+			}
+			if tt.wantErr {
+				require.Error(err)
+			}
+			require.NoError(err)
```
timraymond, line 640: Use either `t.Fatal` / `t.Error` or `require` in tests... seeing both can be confusing for future readers.

cni/network/multitenancy_test.go
```diff
@@ -61,12 +77,23 @@ func (c *MockCNSClient) ReleaseIPAddress(_ context.Context, ipconfig cns.IPConfi
...
 func (c *MockCNSClient) GetAllNetworkContainers(ctx context.Context, orchestratorContext []byte) ([]cns.GetNetworkContainerResponse, error) {
-	c.require.Exactly(c.getAllNetworkContainersConfiguration.orchestratorContext, orchestratorContext)
+	if _, isUnsupported := c.unsupportedAPIs[GetAllNetworkContainers]; isUnsupported {
+		e := &client.CNSClientError{}
+		e.Code = types.UnsupportedAPI
+		e.Err = errUnsupportedAPI
+		return nil, e
+	}
+
+	if !reflect.DeepEqual(c.getAllNetworkContainersConfiguration.orchestratorContext, orchestratorContext) {
```
timraymond, line 94: `reflect.DeepEqual` is rarely the correct func in a `*_test.go` file these days. The trouble is that it compares everything--unexported fields included. This is fine if you don't _have_ any unexported fields, but that may only be true today, at the time you're committing the code. It makes the tests more brittle because they make it more difficult for compared structs to have internals that are invisible to outside observers. Use `github.com/google/go-cmp/cmp` and the `cmp.Equal` func instead (which is a drop-in replacement). The upside is that you can provide diffs when they _aren't_ the same using `cmp.Diff`. That makes it much easier to hunt down broken things.

timraymond (commented):
I have some strongly suggested suggestions.

cni/network/multitenancy_test.go
paulyufan2: I think this is fine. Logic is:
(1) Use new API GetAllNetworkContainers() => it goes to UnsupportedAPI
(2) Use old API GetNetworkContainer() => orchestratorContext is not same => errored.

cni/network/multitenancy_test.go
paulyufan2: fixed, use cmp.Equal

cni/network/multitenancy_test.go
paulyufan2: Fixed, use t.Fatal() and t.Fatalf() instead

cni/network/multitenancy_test.go
paulyufan2: Fixed with this style

cni/network/multitenancy_test.go
paulyufan2: Fixed, use t.Fatal() and t.Fatalf() instead

ashvindeodhar (approved):

PR #1899, opened 2023-04-06
add logs for home az monitor

ZetaoZhuang:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/restserver/homeazmonitor.go
```diff
@@ -144,17 +144,33 @@ func (h *HomeAzMonitor) Populate(ctx context.Context) {
...
 			Message:    msg,
 		},
 		HomeAzResponse: homeAzResponse,
 	}
+	h.logHomeAzResponse(resp)
 	h.updateCacheValue(resp)
 }
 
+// logHomeAzResponse logs HomeAz Response
+func (h *HomeAzMonitor) logHomeAzResponse(resp cns.GetHomeAzResponse) {
```
ashvindeodhar, line 159: Hey Zetao, thinking more on this, I think we should log the HomeAzResponse everytime ( not just when it gets updated ). That's because, we may lose the AZ log if it doesn't change ( which it shouldn't ) for long time. Having that log would be beneficial.

cns/restserver/homeazmonitor.go
```diff
@@ -144,17 +144,33 @@ func (h *HomeAzMonitor) Populate(ctx context.Context) {
...
 func (h *HomeAzMonitor) update(code types.ResponseCode, msg string, homeAzResponse cns.HomeAzResponse) {
-	log.Debugf(msg)
 	resp := cns.GetHomeAzResponse{
 		Response: cns.Response{
 			ReturnCode: code,
 			Message:    msg,
 		},
 		HomeAzResponse: homeAzResponse,
 	}
+	h.logHomeAzResponse(resp)
```
ashvindeodhar, line 154: In the log line can you prepend with [HomeAzMonitor] ?

cns/restserver/homeazmonitor.go
ZetaoZhuang: that is right. I will update this

ashvindeodhar: we don't need the func body below. We should just print the resp struct here.

ZetaoZhuang: but in this way we can log it as error when there have errors in refreshing home az

ashvindeodhar: I would keep the logging simple. Since we already have logic in DNC to identify mismatch on the home AZ we can add alerting there

ashvindeodhar (approved):

PR #1906, opened 2023-04-10
Set constant mac for host veth interface in transparent vlan mode

tamilmani1989:
**Reason for Change**:

This PR adds constant mac (aa:aa:aa:aa:aa:aa) for host veth interface. This prevents kernel to allocate unique mac for each host veth interface. Transparent mode already has this change and doing the same for transparent vlan mode here. Also from ubuntu22, we saw mac address getting changed after veth creation and arp entry set to wrong mac due to this. This prevents that if transparent-vlan mode used in ubuntu22 or later.

- Validate netwrok namespace creation and if its same as VM namespace, delete and recreate it
- Check for interface exists after creating vlan interface
- make sure calls are idempotent. even if it fails, retry should succeed
- delete route before adding new route for same ip (only for transparent vlan)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -466,10 +466,11 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 
 	// iterate ipamAddResults and program the endpoint
 	for i := 0; i < len(ipamAddResults); i++ {
+		var networkID string
 		ipamAddResult = ipamAddResults[i]
 
 		options := make(map[string]any)
-		networkID, err := plugin.getNetworkName(args.Netns, &ipamAddResult, nwCfg)
+		networkID, err = plugin.getNetworkName(args.Netns, &ipamAddResult, nwCfg)
 
 		endpointID := GetEndpointID(args)
 		policies := cni.GetPoliciesFromNwCfg(nwCfg.AdditionalArgs)
@@ -557,7 +558,8 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 			natInfo:          natInfo,
 		}
 
-		epInfo, err := plugin.createEndpointInternal(&createEndpointInternalOpt)
+		var epInfo network.EndpointInfo
+		epInfo, err = plugin.createEndpointInternal(&createEndpointInternalOpt)
 		if err != nil {
 			log.Errorf("Endpoint creation failed:%w", err)
 			return err
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -953,7 +953,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 	numEndpointsToDelete := 1
 	// only get number of endpoints if it's multitenancy mode
 	if nwCfg.MultiTenancy {
-		numEndpointsToDelete = plugin.nm.GetNumEndpointsInNetNs(args.Netns)
+		numEndpointsToDelete = plugin.nm.GetNumEndpointsByContainerID(args.ContainerID)
 	}
 
 	log.Printf("[cni-net] number of endpoints to be deleted %d", numEndpointsToDelete)
@@ -973,23 +973,21 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 		}
 		// Query the network.
 		if nwInfo, err = plugin.nm.GetNetworkInfo(networkID); err != nil {
-			if !nwCfg.MultiTenancy {
-				log.Printf("[cni-net] Failed to query network:%s: %v", networkID, err)
-				// Log the error but return success if the network is not found.
-				// if cni hits this, mostly state file would be missing and it can be reboot scenario where
-				// container runtime tries to delete and create pods which existed before reboot.
-				err = nil
-				return err
-			}
+			// Log the error but return success if the network is not found.
+			// if cni hits this, mostly state file would be missing and it can be reboot scenario where
+			// container runtime tries to delete and create pods which existed before reboot.
+			log.Printf("[cni-net] Failed to query network:%s: %v", networkID, err)
+			err = nil
+			return err
 		}
 
 		endpointID := GetEndpointID(args)
 		// Query the endpoint.
 		if epInfo, err = plugin.nm.GetEndpointInfo(networkID, endpointID); err != nil {
+			log.Printf("[cni-net] GetEndpoint for endpointID: %s returns: %v", endpointID, err)
 			if !nwCfg.MultiTenancy {
 				// attempt to release address associated with this Endpoint id
 				// This is to ensure clean up is done even in failure cases
-				log.Printf("[cni-net] Failed to query endpoint %s: %v", endpointID, err)
 				logAndSendEvent(plugin, fmt.Sprintf("Release ip by ContainerID (endpoint not found):%v", args.ContainerID))
 				if err = plugin.ipamInvoker.Delete(nil, nwCfg, args, nwInfo.Options); err != nil {
 					return plugin.RetriableError(fmt.Errorf("failed to release address(no endpoint): %w", err))
--- a/netlink/mocknetlink.go
+++ b/netlink/mocknetlink.go
@@ -13,9 +13,13 @@ func newErrorMockNetlink(errStr string) error {
 	return fmt.Errorf("%w : %s", ErrorMockNetlink, errStr)
 }
 
+type routeValidateFn func(route *Route) error
+
 type MockNetlink struct {
-	returnError bool
-	errorString string
+	returnError   bool
+	errorString   string
+	deleteRouteFn func(route *Route) error
+	addRouteFn    func(route *Route) error
 }
 
 func NewMockNetlink(returnError bool, errorString string) *MockNetlink {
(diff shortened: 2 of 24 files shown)


network/endpoint_linux.go
```diff
@@ -304,6 +302,23 @@ func addRoutes(nl netlink.NetlinkInterface, netioshim netio.NetIOInterface, inte
 			Table:     route.Table,
 		}
 
+		if removeRouteBeforeAdd {
```
vipul-21, line 305: Is this a part of adding constant mac ? 
 This is used by transport vlan mode only, shouldn't we delete first as separate call and then add routes. Since the function now is adding and deleting routes. (separation of concerns)

vipul-21 (changes requested):

network/endpoint_linux.go
tamilmani1989: yep agreed. let me fix it

network/snat/snat_linux.go
```diff
@@ -81,6 +81,16 @@ func (client *Client) CreateSnatEndpoint() error {
 		return err
 	}
 
+	nuc := networkutils.NewNetworkUtils(client.netlink, client.plClient)
+	// Enabling proxy arp on bridge allows bridge to respond to arp requests it receives with its own mac otherwise arp requests are not getting forwarded and responded.
+	if client.enableProxyArpOnBridge {
+		// Enable proxy arp on bridge
+		if err := nuc.SetProxyArp(SnatBridgeName); err != nil {
+			log.Printf("Enabling proxy arp failed with error %v", err)
```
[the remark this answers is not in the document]
tamilmani1989: yes to allow internet connectivity via apipa bridge. this is existing scenario and not to regress it with transparent-vlan mode

network/snat/snat_linux.go
```diff
@@ -81,6 +81,16 @@ func (client *Client) CreateSnatEndpoint() error {
 		return err
 	}
 
+	nuc := networkutils.NewNetworkUtils(client.netlink, client.plClient)
```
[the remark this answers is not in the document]
tamilmani1989: epc is old package name (epcommon). nuc is networkutilsclient and used in multiple places

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -423,7 +430,7 @@ func (client *TransparentVlanEndpointClient) GetVnetRoutes(ipAddresses []net.IPN
 // to the virtual gateway ip on linkToName device interface
 // Route 1: 169.254.1.1 dev <linkToName>
 // Route 2: default via 169.254.1.1 dev <linkToName>
-func (client *TransparentVlanEndpointClient) AddDefaultRoutes(linkToName string, table int) error {
+func (client *TransparentVlanEndpointClient) addDefaultRoutes(linkToName string, table int) error {
```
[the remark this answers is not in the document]
tamilmani1989: this is not called outside package

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -191,7 +191,14 @@ func (client *TransparentVlanEndpointClient) PopulateVM(epInfo *EndpointInfo) er
...
 
-	if err = client.netUtilsClient.CreateEndpoint(client.vnetVethName, client.containerVethName, nil); err != nil {
+	// Get the default constant host veth mac
+	mac, err := net.ParseMAC(defaultHostVethHwAddr)
+	if err != nil {
+		log.Printf("[net] Failed to parse the mac addrress %v", defaultHostVethHwAddr)
+	}
+
+	// Create veth pair
+	if err = client.netUtilsClient.CreateEndpoint(client.vnetVethName, client.containerVethName, mac); err != nil {
```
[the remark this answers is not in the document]
tamilmani1989: yep good catch. let me fix it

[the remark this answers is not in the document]
tamilmani1989: can you point me to test case which need to be updated?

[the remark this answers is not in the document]
tamilmani1989: UT abstract creation of host veth interface..so need to update UT for this

cni/network/network.go
```diff
@@ -466,10 +466,11 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 
 	// iterate ipamAddResults and program the endpoint
 	for i := 0; i < len(ipamAddResults); i++ {
+		var networkID string
 		ipamAddResult = ipamAddResults[i]
 
 		options := make(map[string]any)
-		networkID, err := plugin.getNetworkName(args.Netns, &ipamAddResult, nwCfg)
+		networkID, err = plugin.getNetworkName(args.Netns, &ipamAddResult, nwCfg)
```
vipul-21, line 473: Why are we changing the way variable are defined ?

cni/network/network.go
```diff
@@ -973,23 +975,21 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
 		if nwInfo, err = plugin.nm.GetNetworkInfo(networkID); err != nil {
-			if !nwCfg.MultiTenancy {
-				log.Printf("[cni-net] Failed to query network:%s: %v", networkID, err)
-				// Log the error but return success if the network is not found.
-				// if cni hits this, mostly state file would be missing and it can be reboot scenario where
-				// container runtime tries to delete and create pods which existed before reboot.
-				err = nil
-				return err
-			}
+			// Log the error but return success if the network is not found.
```
vipul-21, line 978: Why did we remove the check ?

netlink/mocknetlink.go
```diff
@@ -13,9 +13,13 @@ func newErrorMockNetlink(errStr string) error {
...
 }
 
+type routeValidateFn func(route *Route) error
+
 type MockNetlink struct {
-	returnError bool
-	errorString string
+	returnError   bool
+	errorString   string
+	deleteRouteFn func(route *Route) error
```
vipul-21, line 21: The type should be `routeValidateFn`

netlink/mocknetlink.go
```diff
@@ -13,9 +13,13 @@ func newErrorMockNetlink(errStr string) error {
...
+	addRouteFn    func(route *Route) error
```
vipul-21, line 22: same

network/endpoint_linux.go
```diff
@@ -353,15 +350,15 @@ func deleteRoutes(nl netlink.NetlinkInterface, netioshim netio.NetIOInterface, i
 			}
 
 			ifIndex = devIf.Index
-		} else {
+		} else if interfaceName != "" {
```
vipul-21, line 353: Was this failing in a case where `interfaceName` is empty ?
Then should we add another else and log the error or something.

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -500,24 +567,16 @@ func (client *TransparentVlanEndpointClient) DeleteEndpoints(ep *endpoint, delet
...
 
 // getNumRoutesLeft is a function which gets the current number of routes in the namespace. Namespace: Vnet
-func (client *TransparentVlanEndpointClient) DeleteEndpointsImpl(ep *endpoint, getNumRoutesLeft func() (int, error), deleteHostVeth bool) error {
+func (client *TransparentVlanEndpointClient) DeleteEndpointsImpl(ep *endpoint, _ func() (int, error)) error {
 	routeInfoList := client.GetVnetRoutes(ep.IPAddresses)
 	if err := deleteRoutes(client.netlink, client.netioshim, client.vnetVethName, routeInfoList); err != nil {
 		return errors.Wrap(err, "failed to remove routes")
 	}
 
-	routesLeft, err := getNumRoutesLeft()
```
vipul-21, line 509: Why are we removing the `getNumRoutesLeft` ?

And if we need to remove the func, we can remove from the fucn parameters as well then

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -379,16 +441,21 @@ func (client *TransparentVlanEndpointClient) ConfigureVnetInterfacesAndRoutesImp
...
-	if err = client.AddDefaultRoutes(client.vlanIfName, 0); err != nil {
+	if err = client.addDefaultRoutes(client.vlanIfName, 0); err != nil {
 		return errors.Wrap(err, "failed vnet ns add default/gateway routes (idempotent)")
 	}
 	if err = client.AddDefaultArp(client.vlanIfName, azureMac); err != nil {
 		return errors.Wrap(err, "failed vnet ns add default arp entry (idempotent)")
 	}
+
+	// Delete old route if any for this IP
+	err = deleteRoutes(client.netlink, client.netioshim, "", routeInfoList)
```
vipul-21, line 452: If there is a route with this IP, how is this IP is assigned then ?
Also, Is this safe to assume that the Route we are deleting is no longer used ?

vipul-21 (commented):
Can we update the description with the other changes in the PR as well? like using containerId now to fetch the endpoint to delete, updating how namespaces are being created etc

network/endpoint_linux.go
tamilmani1989: it would not fail but skip deleting the route if interface not provided but the requirement is to delete route even if interface name not provided

network/transparent_vlan_endpointclient_linux.go
tamilmani1989: not required.. but i wanted to reevaluate this in future. so keeping it

vipul-21 (approved):
Let's update the description of PR as well with the new changes added in the PR.

vipul-21 (approved):

PR #1909, opened 2023-04-11
ci: validate pods and systemd-networkd restart for PRs

camrynl:
**Reason for Change**:

- A new cilium load test pipeline exists to validate that pod IPs are the same across azure endpoints and cilium endpoints. Adding a case to check CNS endpoints in memory as well. 
- Calling the load test script in the PR pipeline for cilium pod subnet and cilium overlay to strengthen test stages.
- Also added an exit case for when privileged_pod is not found because in testing if the pod didn't come up, the script would be stuck trying to grab azure_endpoints
- Modified cilium install from helm to kubectl apply cilium ds
- --> Applying the daemonset used by AKS so we can apply the init container that fixed systemd-networkd restart issues. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
verified script manually and on both ACN PR and load test pipeline.
Load Test Pipeline run: https://msazure.visualstudio.com/One/_build/results?buildId=71640453&view=logs&j=aafe78ca-85dd-54d8-c96a-840101a5fad4


hack/scripts/validate_state.sh
```diff
@@ -17,6 +17,9 @@ do
     echo "Node internal ip: $node_ip"
     privileged_pod=$(kubectl get pods -n kube-system -l app=privileged-daemonset -o wide | grep "$node_name" | awk '{print $1}')
     echo "privileged pod : $privileged_pod"
+    if [ "$privileged_pod" == '' ]; then
+        exit 1
```
vipul-21, line 21: Did we encounter such case during testing ?

Can we add the status of privileged pod deployment then. ( `kubectl describe daemonset privileged-daemonset -n kube-system`

camrynl: sure I'll add that, here you can see this run just got stuck in the loop. I had to manually cancel it https://dev.azure.com/msazure/One/_build/results?buildId=71680459&view=logs&j=4ea62961-c456-50ab-e773-f15fbc744993&t=6637b73f-d7ef-5d5e-d4d4-eb0bbec757cb&s=c689f5d8-16f1-5a52-95fe-f6a4e6a9e7fe

vipul-21: Sorry I meant if we fail to get the privileged pod then it will be good check the status of daemon set. Basically before exiting we can have that as I see from the pipeline run, `daemonset.apps/privileged-daemonset` created.

Ideally we should be waiting for the deployment to be complete before proceeding i think.

vipul-21 (dismissed):

test/integration/manifests/cilium/cilium-config.yaml
```diff
@@ -0,0 +1,93 @@
...
+    meta.helm.sh/release-namespace: kube-system
+  creationTimestamp: "2023-04-17T23:00:12Z"
+  labels:
+    app.kubernetes.io/managed-by: Helm
+    helm.toolkit.fluxcd.io/name: cilium-adapter-helmrelease
+    helm.toolkit.fluxcd.io/namespace: 643dcea81aff3b00014098f5
+  name: cilium-config
+  namespace: kube-system
+  resourceVersion: "993"
+  uid: 31ccec09-0511-4f18-b21a-10cf4033ce06
```
rbtr, line 93: got some extra metadata here

rbtr (approved):

PR #1913, opened 2023-04-12
CNS ipam and client changes to address comments

rjdenney:
**Reason for Change**:

Adding changes that @rbtr suggested in the last PR for CNS dualstack. Also adds two additional unit tests in case we try removing/adding IPs when only some exist in the IPAM. This results in a unsuccessful request/release with error and puts the IPs back where they originally were.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


rbtr (approved):
needs a lint fixed, otherwise lgtm

rbtr (approved):

PR #1914, opened 2023-04-13
fix: skip failure to call webhook during NNC CRD installation

jaer-tsun:
skip failed to call webhook failure to avoid deadlock when trying to install NNC CRD after cluster start (after stop)

**Reason for Change**:

We discovered a bug where a deadlock occurs due to k8s admission policy for CREATE, UPDATE, DELETE on CRDs. The circular dependency is when DNC-RC tries to install NNC and fails to do so because the webhook pod has no endpoints/IPs available.
Konnectivity tunnel is also broken in this scenario which also prevents API Server from reaching the webhook pod.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
- tested with Overlay by default on aks-rp master rp validation pipeline
- tested with single scenario pipeline for cluster start stop with kyverno webhook
- tested with AKS-Swift checkin validation

error logs
{"level":"error","msg":"Adding Node Network Config CRD erroring: failed to create nnc crd: Internal error occurred: failed calling webhook \"validate.kyverno.svc-ignore\": failed to call webhook: Post \"https://kyverno-svc.kyverno.svc:443/validate/ignore?timeout=10s\": no endpoints available for service \"kyverno-svc\"","orchestratorID":"6435ac854f6e910001a8ff12","time":"2023-04-11T19:15:21Z"}
{"error":"failed to create nnc crd: Internal error occurred: failed calling webhook \"validate.kyverno.svc-ignore\": failed to call webhook: Post \"https://kyverno-svc.kyverno.svc:443/validate/ignore?timeout=10s\": no endpoints available for service \"kyverno-svc\"","level":"error","msg":"error building cluster manager, will retry","orchestratorID":"6435ac854f6e910001a8ff12","time":"2023-04-11T19:15:21Z"}
{"level":"info","msg":"ensuring Node Network Config CRD definition exists on cluster","orchestratorID":"6435ac854f6e910001a8ff12","time":"2023-04-11T19:15:22Z"}

### DIFF
--- a/crd/clustersubnetstate/client.go
+++ b/crd/clustersubnetstate/client.go
@@ -3,6 +3,7 @@ package clustersubnetstate
 import (
 	"context"
 	"reflect"
+	"strings"
 
 	"github.com/Azure/azure-container-networking/crd"
 	"github.com/Azure/azure-container-networking/crd/clustersubnetstate/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


jaer-tsun:
do we need to add the same for cluster subnet state?

rbtr (changes requested):
do not merge

rbtr:
The CRD client is not the appropriate place to include this logic. The caller should be responsible for checking the error returned by this client and failing gracefully.

jaer-tsun:
> The CRD client is not the appropriate place to include this logic. The caller should be responsible for checking the error returned by this client and failing gracefully.

ok gonna add in DNC-RC side

PR #1916, opened 2023-04-14
ci: [CNI] Add restart node in stage in the load test of cni pipeline

vipul-21:
**Reason for Change**:

- Added the test of restarting the nodes in the CNI load testing pipeline. It scales down the deployment and then restart the nodes without any delay.
- Update the `scale_deployment` script to include different parameters for CNI load testing

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests - No, tested in the pipeline CNI Load testing - https://msazure.visualstudio.com/One/_build/results?buildId=71712524&view=results

**Notes**:

### DIFF
--- a/.pipelines/cni/cilium/cilium-cni-load-test.yaml
+++ b/.pipelines/cni/cilium/cilium-cni-load-test.yaml
@@ -36,6 +36,8 @@ stages:
                 set -ex
                 az extension add --name aks-preview
                 make -C ./hack/swift set-kubeconf AZCLI=az CLUSTER=${RESOURCE_GROUP}
+                kubectl apply -f hack/manifests/pod.yaml
+                kubectl apply -f hack/manifests/hostprocess.yaml
                 bash hack/scripts/scale_deployment.sh
   - stage: validate_state
     dependsOn: pod_deployment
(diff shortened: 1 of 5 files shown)


camrynl (approved):

.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -56,8 +58,37 @@ stages:
...
+              azureSubscription: $(TEST_SUB_SERVICE_CONNECTION)
+              scriptLocation: "inlineScript"
+              scriptType: "bash"
+              addSpnToEnvironment: true
+              inlineScript: |
+                echo "Scale up the pods and immediated restart the nodes"
+                make -C ./hack/swift set-kubeconf AZCLI=az CLUSTER=${RESOURCE_GROUP}
+                make -C ./hack/swift azcfg AZCLI=az REGION=$(LOCATION)
+                echo "Scaling the pods down to 100 per node"
+                bash ./hack/scripts/scale_deployment.sh -n 0 -u 1000 -s
```
tamilmani1989, line 78: is it scaling down to 1000 or 100? also number of times is 0?

.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -56,8 +58,37 @@ stages:
...
+                echo "Restarting the nodes"
+                vmss_name=$(az vmss list -g MC_${RESOURCE_GROUP}_${RESOURCE_GROUP}_$(LOCATION) --query "[].name" -o tsv)
+                make -C ./hack/swift restart-vmss AZCLI=az CLUSTER=${RESOURCE_GROUP} REGION=$(LOCATION) VMSS_NAME=$vmss_name
+                bash ./hack/scripts/scale_deployment.sh -n 0 -u 1000 -c
```
tamilmani1989, line 82: why are we calling scale script again?

.pipelines/cni/cilium/cilium-cni-load-test.yaml
vipul-21: The script scale up and down as a by default run. So by specifying `-n` as 0 we are telling it not to do that.
The flag `-s`  is used to define scale up operation. We are scaling down to 1000 pods (i.e 100 pods per node) from 240 pods per node.

.pipelines/cni/cilium/cilium-cni-load-test.yaml
vipul-21: Similarly `-c` will wait until the deployment of 1000 pods is done. The same script is used.
I added the description of what the script can do. And checking the deployment status and deploying pods(scaling pods) can be done via passing flags to the script.
If we don't wait for the deployment to finish after the restart, the validate state script will fail as few of pods( sometimes) are still being created

PR #1921, opened 2023-04-18
[Vnet Scale - CNS]: Flattening CIDR ranges for Node NNC to a list

aggarwal0009:
**Reason for Change**:
*Introduces VNETBlock NCType in CNS
*Reads CIDR IP blocks from VNETBlock NC and updates secondaryIP config accordingly

VNETScale mode NNC example: 
https://microsoft-my.sharepoint.com/personal/asn_microsoft_com/_layouts/15/onedrive.aspx?id=%2Fpersonal%2Fasn%5Fmicrosoft%5Fcom%2FDocuments%2FMicrosoft%20Teams%20Chat%20Files%2FSampleVnetScaleNNC%2Eyaml&parent=%2Fpersonal%2Fasn%5Fmicrosoft%5Fcom%2FDocuments%2FMicrosoft%20Teams%20Chat%20Files&ga=1

This PR introduces support for VNET Scale mode in CNS. The NNC will have CIDR blocks for secondary IPs. Also primaryIP will also be a CIDR block where first IP will be used as PrimaryIP for NC and rest of the IPs will be assigned to secondary IP for pods.

**Issue Fixed**:

**Requirements**:

- [*] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [*] includes documentation
- [*] adds unit tests

**Notes**:

### DIFF
--- a/cns/ipampool/monitor.go
+++ b/cns/ipampool/monitor.go
@@ -133,7 +133,7 @@ func (pm *Monitor) Start(ctx context.Context) error {
 				if nc.Type == v1alpha.VNETBlock {
 					primaryPrefix, err := netip.ParsePrefix(nc.PrimaryIP)
 					if err != nil {
-						return errors.Wrapf(err, "IP: %s", nc.PrimaryIP)
+						return errors.Wrapf(err, "unable to parse ip prefix: %s", nc.PrimaryIP)
 					}
 					pm.metastate.primaryIPAddresses[primaryPrefix.Addr().String()] = struct{}{}
 				}
--- a/cns/ipampool/monitor.go
+++ b/cns/ipampool/monitor.go
@@ -3,6 +3,7 @@ package ipampool
 import (
 	"context"
 	"fmt"
+	"net/netip"
 	"strconv"
 	"sync"
 	"time"
@@ -124,9 +125,17 @@ func (pm *Monitor) Start(ctx context.Context) error {
 			// Add Primary IP to Map, if not present.
 			// This is only for Swift i.e. if NC Type is vnet.
 			for i := 0; i < len(nnc.Status.NetworkContainers); i++ {
-				if nnc.Status.NetworkContainers[i].Type == "" ||
-					nnc.Status.NetworkContainers[i].Type == v1alpha.VNET {
-					pm.metastate.primaryIPAddresses[nnc.Status.NetworkContainers[i].PrimaryIP] = struct{}{}
+				nc := nnc.Status.NetworkContainers[i]
+				if nc.Type == "" || nc.Type == v1alpha.VNET {
+					pm.metastate.primaryIPAddresses[nc.PrimaryIP] = struct{}{}
+				}
+
+				if nc.Type == v1alpha.VNETBlock {
+					primaryPrefix, err := netip.ParsePrefix(nc.PrimaryIP)
+					if err != nil {
+						return errors.Wrapf(err, "IP: %s", nc.PrimaryIP)
+					}
+					pm.metastate.primaryIPAddresses[primaryPrefix.Addr().String()] = struct{}{}
 				}
 			}
 
--- a/cns/kubecontroller/nodenetworkconfig/conversion.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion.go
@@ -90,7 +90,7 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 
 	req, err := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error while creatng NC request from static NC")
+		return nil, errors.Wrapf(err, "error while creating NC request from static NC")
 	}
 	return req, err
 }
--- a/cns/kubecontroller/nodenetworkconfig/conversion.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion.go
@@ -89,5 +89,8 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 	}
 
 	req, err := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while creatng NC request from static NC")
+	}
 	return req, err
 }
--- a/cns/kubecontroller/nodenetworkconfig/conversion.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion.go
@@ -73,7 +73,10 @@ func CreateNCRequestFromDynamicNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwo
 //
 //nolint:gocritic //ignore hugeparam
 func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetworkContainerRequest, error) {
-	nc.Version = 0 // fix for NMA always giving us version 0 for Overlay NCs
+	if nc.Type == v1alpha.Overlay {
+		nc.Version = 0 // fix for NMA always giving us version 0 for Overlay NCs
+	}
+
 	primaryPrefix, err := netip.ParsePrefix(nc.PrimaryIP)
 	if err != nil {
 		return nil, errors.Wrapf(err, "IP: %s", nc.PrimaryIP)
--- a/cns/kubecontroller/nodenetworkconfig/conversion.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion.go
@@ -88,6 +88,6 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 		PrefixLength: uint8(subnetPrefix.Bits()),
 	}
 
-	req := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
-	return req, nil
+	req, err := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
+	return req, err
 }
--- a/cns/kubecontroller/nodenetworkconfig/conversion_test.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion_test.go
@@ -10,19 +10,27 @@ import (
 )
 
 const (
-	uuid               = "539970a2-c2dd-11ea-b3de-0242ac130004"
-	defaultGateway     = "10.0.0.2"
-	ipIsCIDR           = "10.0.0.1/32"
-	ipMalformed        = "10.0.0.0.0"
-	ncID               = "160005ba-cd02-11ea-87d0-0242ac130003"
-	primaryIP          = "10.0.0.1"
-	overlayPrimaryIP   = "10.0.0.1/30"
-	subnetAddressSpace = "10.0.0.0/24"
-	subnetName         = "subnet1"
-	subnetPrefixLen    = 24
-	testSecIP          = "10.0.0.2"
-	version            = 1
-	nodeIP             = "10.1.0.5"
+	uuid                        = "539970a2-c2dd-11ea-b3de-0242ac130004"
+	defaultGateway              = "10.0.0.2"
+	ipIsCIDR                    = "10.0.0.1/32"
+	ipMalformed                 = "10.0.0.0.0"
+	ncID                        = "160005ba-cd02-11ea-87d0-0242ac130003"
+	primaryIP                   = "10.0.0.1"
+	overlayPrimaryIP            = "10.0.0.1/30"
+	subnetAddressSpace          = "10.0.0.0/24"
+	subnetName                  = "subnet1"
+	subnetPrefixLen             = 24
+	testSecIP                   = "10.0.0.2"
+	version                     = 1
+	nodeIP                      = "10.1.0.5"
+	vnetBlockPrimaryIP          = "10.224.0.4"
+	vnetBlockPrimaryIPPrefix    = "10.224.0.4/30"
+	vnetBlockSubnetAddressSpace = "10.224.0.0/14"
+	vnetBlockSubnetPrefixLen    = 14
+	vnetBlockNodeIP             = "10.228.0.6"
+	vnetBlockDefaultGateway     = "10.224.0.1"
+	vnetBlockCIDR1              = "10.224.0.8/30"
+	vnetBlockCIDR2              = "10.224.0.12/30"
 )
 
 var invalidStatusMultiNC = v1alpha.NodeNetworkConfigStatus{
@@ -85,6 +93,88 @@ var validOverlayNC = v1alpha.NetworkContainer{
 	SubnetName:         subnetName,
 	SubnetAddressSpace: subnetAddressSpace,
 	Version:            version,
+}
+
+var validVNETBlockNC = v1alpha.NetworkContainer{
+	ID:             ncID,
+	AssignmentMode: v1alpha.Static,
+	Type:           v1alpha.VNETBlock,
+	IPAssignments: []v1alpha.IPAssignment{
+		{
+			Name: uuid,
+			IP:   vnetBlockCIDR1,
+		},
+		{
+			Name: uuid,
+			IP:   vnetBlockCIDR2,
+		},
+	},
+	NodeIP:             vnetBlockNodeIP,
+	PrimaryIP:          vnetBlockPrimaryIPPrefix,
+	SubnetName:         subnetName,
+	SubnetAddressSpace: vnetBlockSubnetAddressSpace,
+	DefaultGateway:     vnetBlockDefaultGateway,
+	Version:            version,
+}
+
+var validVNETBlockRequest = &cns.CreateNetworkContainerRequest{
+	Version: strconv.FormatInt(version, 10),
+	IPConfiguration: cns.IPConfiguration{
+		GatewayIPAddress: vnetBlockDefaultGateway,
+		IPSubnet: cns.IPSubnet{
+			PrefixLength: uint8(vnetBlockSubnetPrefixLen),
+			IPAddress:    vnetBlockPrimaryIP,
+		},
+	},
+	NetworkContainerid:   ncID,
+	NetworkContainerType: cns.Docker,
+	// Ignore first IP in first CIDR Block, i.e. 10.224.0.4
+	SecondaryIPConfigs: map[string]cns.SecondaryIPConfig{
+		"10.224.0.5": {
+			IPAddress: "10.224.0.5",
+			NCVersion: version,
+		},
+		"10.224.0.6": {
+			IPAddress: "10.224.0.6",
+			NCVersion: version,
+		},
+		"10.224.0.7": {
+			IPAddress: "10.224.0.7",
+			NCVersion: version,
+		},
+		"10.224.0.8": {
+			IPAddress: "10.224.0.8",
+			NCVersion: version,
+		},
+		"10.224.0.9": {
+			IPAddress: "10.224.0.9",
+			NCVersion: version,
+		},
+		"10.224.0.10": {
+			IPAddress: "10.224.0.10",
+			NCVersion: version,
+		},
+		"10.224.0.11": {
+			IPAddress: "10.224.0.11",
+			NCVersion: version,
+		},
+		"10.224.0.12": {
+			IPAddress: "10.224.0.12",
+			NCVersion: version,
+		},
+		"10.224.0.13": {
+			IPAddress: "10.224.0.13",
+			NCVersion: version,
+		},
+		"10.224.0.14": {
+			IPAddress: "10.224.0.14",
+			NCVersion: version,
+		},
+		"10.224.0.15": {
+			IPAddress: "10.224.0.15",
+			NCVersion: version,
+		},
+	},
 }
 
 func TestCreateNCRequestFromDynamicNC(t *testing.T) {
@@ -267,6 +357,41 @@ func TestCreateNCRequestFromStaticNC(t *testing.T) {
 					},
 				},
 				SubnetAddressSpace: "10.0.0.0", // not a cidr range
+			},
+			wantErr: true,
+		},
+		// VNET Block test cases
+		{
+			name:    "valid VNET Block",
+			input:   validVNETBlockNC,
+			wantErr: false,
+			want:    validVNETBlockRequest,
+		},
+		{
+			name: "PrimaryIP is not CIDR",
+			input: v1alpha.NetworkContainer{
+				AssignmentMode:     v1alpha.Static,
+				Type:               v1alpha.VNETBlock,
+				PrimaryIP:          vnetBlockPrimaryIP,
+				ID:                 ncID,
+				SubnetAddressSpace: "10.224.0.0/14",
+			},
+			wantErr: true,
+		},
+		{
+			name: "IP assignment is not CIDR",
+			input: v1alpha.NetworkContainer{
+				AssignmentMode: v1alpha.Static,
+				Type:           v1alpha.VNETBlock,
+				PrimaryIP:      vnetBlockPrimaryIPPrefix,
+				ID:             ncID,
+				IPAssignments: []v1alpha.IPAssignment{
+					{
+						Name: uuid,
+						IP:   "10.224.0.4",
+					},
+				},
+				SubnetAddressSpace: "10.224.0.0/14",
 			},
 			wantErr: true,
 		},
--- a/cns/kubecontroller/nodenetworkconfig/conversion_windows.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion_windows.go
@@ -21,13 +21,17 @@ func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPre
 	// as secondary IPs
 	startingAddr := primaryIPPrefix.Masked().Addr() // the masked address is the 0th IP in the subnet
 	if nc.DefaultGateway == "" && nc.Type == v1alpha.Overlay {
+		// assign 0th IP to the default gateway
 		nc.DefaultGateway = startingAddr.String()
 		startingAddr = startingAddr.Next()
 	} else if nc.DefaultGateway == "" && nc.Type == v1alpha.VNETBlock {
+		// skipping 0th IP for the Primary IP of NC
 		startingAddr = startingAddr.Next()
+		// assign next IP to default gateway
 		nc.DefaultGateway = startingAddr.String()
 		startingAddr = startingAddr.Next()
 	} else if nc.Type == v1alpha.VNETBlock {
+		// skipping 0th IP for the Primary IP of NC
 		startingAddr = startingAddr.Next()
 	}
 
--- a/cns/kubecontroller/nodenetworkconfig/conversion_windows.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion_windows.go
@@ -13,7 +13,7 @@ import (
 // secondary IPs. If the gateway is not empty, it will not reserve the 0th IP and add it as a secondary IP.
 //
 //nolint:gocritic //ignore hugeparam
-func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPrefix netip.Prefix, subnet cns.IPSubnet) *cns.CreateNetworkContainerRequest {
+func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPrefix netip.Prefix, subnet cns.IPSubnet) (*cns.CreateNetworkContainerRequest, error) {
 	secondaryIPConfigs := map[string]cns.SecondaryIPConfig{}
 
 	// if NC DefaultGateway is empty, set the 0th IP to the gateway and add the rest of the IPs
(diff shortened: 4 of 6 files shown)


cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,13 +91,38 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+			for blockIndex, ipAssignment := range nc.IPAssignments {
+				cidrPrefix, err := netip.ParsePrefix(ipAssignment.IP)
+				if err != nil {
+					return nil, errors.Wrapf(ErrInvalidCidrIPBlock, "CIDR IP Block: %s", ipAssignment.IP)
+				}
+				// iterate through all IP addresses in the CIDR block described by cidrPrefix and
+				// add them to the request as secondary IPConfigs.
+				for addr := cidrPrefix.Masked().Addr(); cidrPrefix.Contains(addr); addr = addr.Next() {
+					if blockIndex == 0 {
+						addr = addr.Next() // # Block 1 CIDR. Exclude 1st IP from pod assignment
```
miguelgoms, line 106: I think this check belongs outside the inner for-loop, otherwise you are skipping more than the first IP on the first block (besides checking the condition redundantly). for that matter, initializing "addr" explicitly before the loop seems more readable if we need to handle at least this special case.

aggarwal0009: fixed

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,13 +91,41 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+		// iterate through all IP addresses in the subnet described by primaryPrefix and
+		// add them to the request as secondary IPConfigs.
+		for addr := primaryPrefix.Masked().Addr(); primaryPrefix.Contains(addr); addr = addr.Next() {
+			secondaryIPConfigs[addr.String()] = cns.SecondaryIPConfig{
+				IPAddress: addr.String(),
+				NCVersion: int(nc.Version),
+			}
 		}
+	case v1alpha.VNET:
+		return nil, errors.New("vnet mode not supported for static NC")
```
rbtr, line 128: we don't need to check this, this function is only called when we have already checked that the AssignmentMode is `Static`

rbtr: VNETScale involves adding/removing SecondaryIPs to the NNC based on demand, though. Isn't that actually `Dynamic`?

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,13 +91,41 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+			for blockIndex, ipAssignment := range nc.IPAssignments {
+				cidrPrefix, err := netip.ParsePrefix(ipAssignment.IP)
+				if err != nil {
+					return nil, errors.Wrapf(ErrInvalidCidrIPBlock, "CIDR IP Block: %s", ipAssignment.IP)
+				}
+
+				zeroAddr := cidrPrefix.Masked().Addr()
+				if blockIndex == 0 {
+					zeroAddr = zeroAddr.Next() // # Block 1 CIDR. Exclude 1st IP from pod assignment
+				}
```
rbtr, line 106: @nairashu 
I am making an assumption on intent here, but I read this as blocking the first address of the first Secondary IP Assignment Block for use as the NC Primary IP. 

If that is correct:
I'm concerned that we're blocking the Primary IP from the "first" Block in the NC Secondary IP Assignments with an expectation of consistent ordering when that is really not guaranteed.

Without changing too much, maybe we could make the "PrimaryIP" a CIDR block like any of the Secondary IP Assignments, and use the 0th address of that block as the Primary IP for the NC while adding the rest of the block as pod-assignable IPs, as if they were Secondary assignments (this is what we do in Overlay). Then, any additional blocks that we receive through the Secondary IP Assignments do not need special-case handling like this.

rbtr (commented):
I could use some more detail on the intent of these specific changes (it would be helpful if you populated the description of the PR with enough information for someone with no context to be able to review it), but I think there are some gaps here

cns/ipampool/monitor.go
```diff
@@ -123,9 +124,17 @@ func (pm *Monitor) Start(ctx context.Context) error {
...
+				if nc.Type == "" || nc.Type == v1alpha.VNET {
+					pm.metastate.primaryIPAddresses[nc.PrimaryIP] = struct{}{}
+				}
+
+				if nc.Type == v1alpha.VNETBlock {
+					primaryPrefix, err := netip.ParsePrefix(nc.PrimaryIP)
+					if err != nil {
+						return errors.Wrapf(err, "IP: %s", nc.PrimaryIP)
+					}
+					pm.metastate.primaryIPAddresses[primaryPrefix.Addr().String()] = struct{}{}
```
rbtr, line 137: We use `.Masked()` in the conversion.go to get the 0th address, should we do that here also? If we get `192.168.0.5/28`, is the primary the `.5` address, or the `.0` address?

cns/kubecontroller/nodenetworkconfig/conversion.go
nairashu: Had an offline discussion with Evan on this front. Few clarifying points:

1. Vnet Scale is a static allocation for the node as there will be no scaling, so all the blocks are assigned during node creation.
2. Evan brought up the concern that slices intrinsically do not have ordering, especially since we are converting them from JSON back and forth.
3. Will have to change the NNC where the primary IP CIDR block is not included in the IP asssignments section when being sent.

@aggarwal0009 : Change logic to:

1. We will need to process the primary IP CIDR block from the primary IP section where the 0th IP is excluded and the remaining IPs are added to the secondary IP slice.
2. All assignments in IP assignments can now be added directly to the secondary IP slice after flattening.

cns/kubecontroller/nodenetworkconfig/conversion.go
nairashu: Vnet Scale will be static allocation mode

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -18,9 +18,12 @@ var (
 	ErrInvalidSecondaryIP = errors.New("invalid secondary IP")
 	// ErrUnsupportedNCQuantity indicates that the node has an unsupported nummber of Network Containers attached.
 	ErrUnsupportedNCQuantity = errors.New("unsupported number of network containers")
+	// ErrInvalidCidrIPBlock indicates the NC cidr ip block is invalid.
+	ErrInvalidCidrIPBlock = errors.New("invalid CIDR IP block")
```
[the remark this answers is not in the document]
nairashu: +1

cns/kubecontroller/nodenetworkconfig/conversion.go
aggarwal0009: My understanding was we assign all CIDR blocks in NNC during initialization and then there are no updates.
@nairashu Can u confirm?

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,13 +91,41 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
 
-	// iterate through all IP addresses in the subnet described by primaryPrefix and
-	// add them to the request as secondary IPConfigs.
-	for addr := primaryPrefix.Masked().Addr(); primaryPrefix.Contains(addr); addr = addr.Next() {
-		secondaryIPConfigs[addr.String()] = cns.SecondaryIPConfig{
-			IPAddress: addr.String(),
-			NCVersion: int(nc.Version),
+	switch nc.Type {
+	case v1alpha.VNETBlock:
+		{
```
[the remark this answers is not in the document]
aggarwal0009: updated

cns/kubecontroller/nodenetworkconfig/conversion.go
aggarwal0009: Updated to remove unnecessary case.

cns/kubecontroller/nodenetworkconfig/conversion.go
aggarwal0009: Updated

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -86,15 +90,44 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+		// iterate through all IP addresses in the subnet described by primaryPrefix and
+		// add them to the request as secondary IPConfigs.
+		// Skip first IP in the primary ip CIDR block during secondary ip assignments
+		zeroAddr := primaryPrefix.Masked().Addr().Next()
+		for addr := zeroAddr; primaryPrefix.Contains(addr); addr = addr.Next() {
+			secondaryIPConfigs[addr.String()] = cns.SecondaryIPConfig{
+				IPAddress: addr.String(),
+				NCVersion: int(nc.Version),
+			}
+		}
```
rbtr, line 103: This looks suspiciously like the entire contents of the `else` case. What if you pulled that code out in front of the `if` block, since it needs to happen in both branches? Then you can drop the else entirely, and there's less code in the `if`. Significantly simpler control flow, imo.

rbtr (changes requested):

cns/ipampool/monitor.go
aggarwal0009: In that case, i believe primary ip is .5 I guess shouldn't have used masked in conversion.go. 
Updating

thatmattlong: would we ever get 192.168.0.5/28? That is an invalid CIDR, I would expect it to fail parsing before this code block

cns/kubecontroller/nodenetworkconfig/conversion.go
aggarwal0009: Sounds fine to me.
Only thing is, since in VNETBlock mode we exclude first IP from primaryIP CIDR and dont mask primaryIP,  I will have to use two if statements. One for  startAddr initialization and one for IP assignments. from CIDR blocks

cns/ipampool/monitor.go
aggarwal0009: I am not enforcing that validation in CNS. I was assuming DNC provides a valid primaryIP CIDR. FYI @nairashu

nairashu: DNC should be passing down correct CIDR ranges. Maybe we should error out a metric if the range is invalid

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -18,9 +18,12 @@ var (
 	ErrInvalidSecondaryIP = errors.New("invalid secondary IP")
 	// ErrUnsupportedNCQuantity indicates that the node has an unsupported nummber of Network Containers attached.
 	ErrUnsupportedNCQuantity = errors.New("unsupported number of network containers")
+	// ErrInvalidCIDRBlock indicates the NC CIDR block is invalid.
+	ErrInvalidCIDRBlock = errors.New("invalid CIDR block")
 )
 
 // CreateNCRequestFromDynamicNC generates a CreateNetworkContainerRequest from a dynamic NetworkContainer.
+//
```
nairashu, line 26: nit: Unnecessary comment line

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -69,6 +72,7 @@ func CreateNCRequestFromDynamicNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwo
 }
 
 // CreateNCRequestFromStaticNC generates a CreateNetworkContainerRequest from a static NetworkContainer.
+//
```
nairashu, line 75: nit: Unnecessary comment line

cns/kubecontroller/nodenetworkconfig/conversion_test.go
```diff
@@ -10,19 +10,27 @@ import (
...
+	ncID                        = "160005ba-cd02-11ea-87d0-0242ac130003"
+	primaryIP                   = "10.0.0.1"
+	overlayPrimaryIP            = "10.0.0.1/30"
+	subnetAddressSpace          = "10.0.0.0/24"
+	subnetName                  = "subnet1"
+	subnetPrefixLen             = 24
+	testSecIP                   = "10.0.0.2"
+	version                     = 1
+	nodeIP                      = "10.1.0.5"
+	vnetBlockPrimaryIP          = "10.224.0.4"
```
nairashu, line 26: Let's use the primary IP block as 10.224.0.32/27 and other IP blocks for the node as 10.224.0.64/27 and 10.224.0.128/27

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -86,15 +90,44 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 	}
 
 	secondaryIPConfigs := map[string]cns.SecondaryIPConfig{}
+	if nc.Type == v1alpha.VNETBlock {
```
nairashu, line 93:
```suggestion
	if nc.Type == v1alpha.VNETBlock {
		// iterate through all IP addresses in the subnet described by primaryPrefix and
		// add them to the request as secondary IPConfigs.
		// Skip first IP in the primary ip CIDR block during secondary ip assignments
		startAddr := primaryPrefix.Addr().Next()
		for addr := startAddr; primaryPrefix.Contains(addr); addr = addr.Next() {
			secondaryIPConfigs[addr.String()] = cns.SecondaryIPConfig{
				IPAddress: addr.String(),
				NCVersion: int(nc.Version),
			}
		}

		for _, ipAssignment := range nc.IPAssignments {
			cidrPrefix, err := netip.ParsePrefix(ipAssignment.IP)
			if err != nil {
				return nil, errors.Wrapf(ErrInvalidCIDRBlock, "CIDR IP Block: %s", ipAssignment.IP)
			}

			// iterate through all IP addresses in the CIDR block described by cidrPrefix and
			// add them to the request as secondary IPConfigs.
			for addr := cidrPrefix.Addr(); cidrPrefix.Contains(addr); addr = addr.Next() {
				secondaryIPConfigs[addr.String()] = cns.SecondaryIPConfig{
					IPAddress: addr.String(),
					NCVersion: int(nc.Version),
				}
			}
		}
	} else {
		// iterate through all IP addresses in the subnet described by primaryPrefix and
		// add them to the request as secondary IPConfigs.
		for addr := primaryPrefix.Masked().Addr(); primaryPrefix.Contains(addr); addr = addr.Next() {
			secondaryIPConfigs[addr.String()] = cns.SecondaryIPConfig{
				IPAddress: addr.String(),
				NCVersion: int(nc.Version),
			}
		}
	}

	return &cns.CreateNetworkContainerRequest{
		SecondaryIPConfigs:   secondaryIPConfigs,
		NetworkContainerid:   nc.ID,
		NetworkContainerType: cns.Docker,
		Version:              strconv.FormatInt(nc.Version, 10), //nolint:gomnd // it's decimal
		IPConfiguration: cns.IPConfiguration{
			IPSubnet:         subnet,
			GatewayIPAddress: nc.DefaultGateway,
		},
	}, nil
}
```

cns/kubecontroller/nodenetworkconfig/conversion.go
aggarwal0009: removed

cns/kubecontroller/nodenetworkconfig/conversion.go
aggarwal0009: removed

cns/ipampool/monitor.go
rbtr: `192.168.0.5/28` IS valid CIDR notation, and it parses fine.
The question is, IF we get such an address, do we take *that* address (`192.168.0.5`) to be the NC Primary IP, or do we take the network address of the block (the 0th address, `192.168.0.0`)?
In Overlay, we have taken the 0th address of the block, but now I'm wondering if that's guaranteed.

nairashu: So the IP defined in the primary IP field will be taken by PubSub as the Primary IP address by removing the / from the CIDR notation. We will have to also adhere to that contract. That being said we should only be passing down an address where the /27 CIDR is the 0th address of the CIDR block.

nairashu: We should have validations in DNC code for this purpose during CIDR allocations to the NC.

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -95,6 +99,33 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 			NCVersion: int(nc.Version),
 		}
 	}
+
+	// Add IPs from CIDR block to the secondary IPConfigs
+	if nc.Type == v1alpha.VNETBlock {
+		// Delete the IPs upto(including) first IP of primary prefix
+		startAddr := primaryPrefix.Masked().Addr()
+		endAddr := primaryPrefix.Addr()
+		for addr := startAddr; addr != endAddr.Next() && primaryPrefix.Contains(addr); addr = addr.Next() {
```
nairashu, line 108: Why is there a loop here as it will only be one IP that needs to be removed always and secondaryIPConfigs is a map where you can directly access it using the address string.

aggarwal0009: ok let me explain my understanding with an example.
Lets say PrimaryIP is 10.224.0.4/27. 
With code above this if block(which is same for overlay), we add IPs starting from 10.224.0.0(not 10.224.0.4) to 10.224.0.31 to secondaryIP config.
Now we need to remove not only 10.224.0.4 but  10.224.0.0, 10.224.0.1, 10.224.0.2 &  10.224.0.3 as well.

cns/ipampool/monitor.go
rbtr: @thatmattlong is this true for Overlay also?

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -95,6 +99,33 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+
+		for _, ipAssignment := range nc.IPAssignments {
+			cidrPrefix, err := netip.ParsePrefix(ipAssignment.IP)
+			if err != nil {
+				return nil, errors.Wrapf(ErrInvalidCIDRBlock, "CIDR Block: %s", ipAssignment.IP)
+			}
+
+			// iterate through all IP addresses in the CIDR block described by cidrPrefix and
+			// add them to the request as secondary IPConfigs.
+			for addr := cidrPrefix.Addr(); cidrPrefix.Contains(addr); addr = addr.Next() {
```
rbtr, line 120: should start with the zeroth address in these CIDRs also
```suggestion
			for addr := cidrPrefix.Masked().Addr(); cidrPrefix.Contains(addr); addr = addr.Next() {
```

cns/kubecontroller/nodenetworkconfig/conversion.go
aggarwal0009: ok, updated

cns/ipampool/monitor.go
aggarwal0009: Synced with Ashish. Seems like I need to use Primary address as is(without masking) since in NMAgent the programming will be to remove the / from the IP mentioned for Primary when NC is created. We will however be passing down address where /27 CIDR is the 0th address of the block.  The validation will be in DNC though.

@rbtr Pls resolve this if your concern has been addressed. :)

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -95,6 +99,29 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+	// Add IPs from CIDR block to the secondary IPConfigs
+	if nc.Type == v1alpha.VNETBlock {
+		// Delete primary IP reserved for Primary IP for NC
+		delete(secondaryIPConfigs, primaryPrefix.Addr().String())
+
+		for _, ipAssignment := range nc.IPAssignments {
+			cidrPrefix, err := netip.ParsePrefix(ipAssignment.IP)
+			if err != nil {
+				return nil, errors.Wrapf(ErrInvalidCIDRBlock, "CIDR Block: %s", ipAssignment.IP)
+			}
```
rbtr, line 112: it looks like this `err` is not included in the return, so we are losing the cause of the failure

rbtr (changes requested):
one remaining minor issue with the error returns, otherwise lgtm

cns/kubecontroller/nodenetworkconfig/conversion.go
aggarwal0009: updated

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -95,6 +97,29 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+
+	// Add IPs from CIDR block to the secondary IPConfigs
+	if nc.Type == v1alpha.VNETBlock {
+		// Delete primary IP reserved for Primary IP for NC
+		delete(secondaryIPConfigs, primaryPrefix.Addr().String())
+
+		for _, ipAssignment := range nc.IPAssignments {
+			cidrPrefix, err := netip.ParsePrefix(ipAssignment.IP)
+			if err != nil {
+				return nil, errors.Wrapf(err, "Invalid CIDR Block: %s", ipAssignment.IP)
```
rbtr, line 109: nit: error strings should not be sentence-case, since they get embedded in other strings

rbtr (dismissed):

rbtr (approved):

rbtr:
@paulyufan2 @aggarwal0009 
please don't "Update branch", it adds commits to your branch that are just noise: 
image 
You should rebase instead - conceptually, this means your changes will always happen "after" the latest master (which is our common true "latest" revision), instead of your changes and the concurrent changes from master being intermingled.
Usually, you can use this button:
image
But because you have mangled your branch history with merges already, you will need to do it manually via the command line, and resolve your conflicts locally.

rbtr (changes requested):
blocking for clean commit history

aggarwal0009:
> @paulyufan2 @aggarwal0009 please don't "Update branch", it adds commits to your branch that are just noise: image You should rebase instead - conceptually, this means your changes will always happen "after" the latest master (which is our common true "latest" revision), instead of your changes and the concurrent changes from master being intermingled. Usually, you can use this button: image But because you have mangled your branch history with merges already, you will need to do it manually via the command line, and resolve your conflicts locally.

Updated

aggarwal0009:
> blocking for clean commit history

Updated

rbtr (approved):

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -21,6 +21,7 @@ var (
 )
 
 // CreateNCRequestFromDynamicNC generates a CreateNetworkContainerRequest from a dynamic NetworkContainer.
+//
```
nairashu, line 24: I think these got reintorduced

nairashu (dismissed):

nairashu (dismissed):

cns/kubecontroller/nodenetworkconfig/conversion_windows.go
```diff
@@ -6,14 +6,15 @@ import (
...
 // createNCRequestFromStaticNCHelper generates a CreateNetworkContainerRequest from a static NetworkContainer.
 // If the NC's DefaultGateway is empty, it will set the 0th IP as the gateway IP and all remaining IPs as
 // secondary IPs. If the gateway is not empty, it will not reserve the 0th IP and add it as a secondary IP.
 //
 //nolint:gocritic //ignore hugeparam
-func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPrefix netip.Prefix, subnet cns.IPSubnet) *cns.CreateNetworkContainerRequest {
+func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPrefix netip.Prefix, subnet cns.IPSubnet) (*cns.CreateNetworkContainerRequest, error) {
 	secondaryIPConfigs := map[string]cns.SecondaryIPConfig{}
 
 	// if NC DefaultGateway is empty, set the 0th IP to the gateway and add the rest of the IPs
```
nairashu, line 20: Let's put this code of Matt to have the extra check for the subnet type to be Overlay specifically and change the logic according for Vnet Block.

aggarwal0009: so not default gateway config needed if it is empty for vnetscale?

cns/kubecontroller/nodenetworkconfig/conversion_windows.go
```diff
@@ -32,6 +33,37 @@ func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPre
 			NCVersion: int(nc.Version),
 		}
 	}
+
+	if nc.Type == v1alpha.VNETBlock {
```
nairashu, line 37: Let's rework the logic to

if DefaultGw == "" && Overlay {
    // set 0th IP to GW
} else if DefaultGw == "" && VnetBlock {
    // set 0th IP as Primary
    // set 1st IP as GW
} else if Vnet Block {
    // set 0th IP as Primary
}

// block to add all remaining IPs from primary block for assignment

// Add all IPs from secondary block for assignment

aggarwal0009: updated

cns/kubecontroller/nodenetworkconfig/conversion_windows.go
```diff
@@ -6,22 +6,29 @@ import (
...
 
 	// if NC DefaultGateway is empty, set the 0th IP to the gateway and add the rest of the IPs
 	// as secondary IPs
 	startingAddr := primaryIPPrefix.Masked().Addr() // the masked address is the 0th IP in the subnet
-	if nc.DefaultGateway == "" {
+	if nc.DefaultGateway == "" && nc.Type == v1alpha.Overlay {
 		nc.DefaultGateway = startingAddr.String()
 		startingAddr = startingAddr.Next()
+	} else if nc.DefaultGateway == "" && nc.Type == v1alpha.VNETBlock {
+		startingAddr = startingAddr.Next()
```
nairashu, line 27: Leave a comment on why you are skipping the 0th address block here for primary IP

aggarwal0009: done

aggarwal0009: done

nairashu (approved):

cns/ipampool/monitor.go
```diff
@@ -124,9 +125,17 @@ func (pm *Monitor) Start(ctx context.Context) error {
...
-					pm.metastate.primaryIPAddresses[nnc.Status.NetworkContainers[i].PrimaryIP] = struct{}{}
+				nc := nnc.Status.NetworkContainers[i]
+				if nc.Type == "" || nc.Type == v1alpha.VNET {
+					pm.metastate.primaryIPAddresses[nc.PrimaryIP] = struct{}{}
+				}
+
+				if nc.Type == v1alpha.VNETBlock {
+					primaryPrefix, err := netip.ParsePrefix(nc.PrimaryIP)
+					if err != nil {
+						return errors.Wrapf(err, "IP: %s", nc.PrimaryIP)
```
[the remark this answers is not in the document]
timraymond: the convention I use is to answer the question: "what were you trying to do when you got the error?" The answer here would be: "parsing ip prefix %q". If done consistently it makes a nice consistent chain of implication:

`processing request: cranking widget: connecting to spline reticulator: parsing ip prefix: some error`

cns/kubecontroller/nodenetworkconfig/conversion_windows.go
```diff
@@ -6,22 +6,33 @@ import (
...
-	if nc.DefaultGateway == "" {
+	if nc.DefaultGateway == "" && nc.Type == v1alpha.Overlay {
+		// assign 0th IP to the default gateway
 		nc.DefaultGateway = startingAddr.String()
 		startingAddr = startingAddr.Next()
+	} else if nc.DefaultGateway == "" && nc.Type == v1alpha.VNETBlock {
+		// skipping 0th IP for the Primary IP of NC
+		startingAddr = startingAddr.Next()
+		// assign next IP to default gateway
+		nc.DefaultGateway = startingAddr.String()
```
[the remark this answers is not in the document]
aggarwal0009: Updated

cns/ipampool/monitor.go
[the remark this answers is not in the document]
aggarwal0009: fixed

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,6 +87,6 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 		PrefixLength: uint8(subnetPrefix.Bits()),
 	}
 
-	req := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
-	return req, nil
+	req, err := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
```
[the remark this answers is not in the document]
aggarwal0009: done

cns/kubecontroller/nodenetworkconfig/conversion_test.go
```diff
@@ -87,6 +95,88 @@ var validOverlayNC = v1alpha.NetworkContainer{
...
+	},
+	NodeIP:             vnetBlockNodeIP,
+	PrimaryIP:          vnetBlockPrimaryIPPrefix,
+	SubnetName:         subnetName,
+	SubnetAddressSpace: vnetBlockSubnetAddressSpace,
+	DefaultGateway:     vnetBlockDefaultGateway,
+	Version:            version,
+}
+
+var validVNETBlockRequest = &cns.CreateNetworkContainerRequest{
```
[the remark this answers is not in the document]
aggarwal0009: Synced with Tim offline. Will keep this as it is for now but keep in mind going forward and try to resolve this in next PR.

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,6 +87,9 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while creatng NC request from static NC")
```
timraymond, line 92: Typo:
```suggestion
		return nil, errors.Wrapf(err, "error while creating NC request from static NC")
```

nairashu (dismissed):

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,6 +87,9 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 		PrefixLength: uint8(subnetPrefix.Bits()),
 	}
 
-	req := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
-	return req, nil
+	req, err := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while creating NC request from static NC")
```
thatmattlong, line 92: since you're not using a format string you should use `errors.Wrap` and not `Wrapf`

thatmattlong (approved):
LGTM with suggestion to not use `errors.Wrapf` when there is no format string

cns/kubecontroller/nodenetworkconfig/conversion.go
aggarwal0009: fixed

nairashu (approved):

PR #1922, opened 2023-04-20
fix: [CNI] NAT hostPort mapping for HNSv2

### LINKED ISSUE #1863: Windows Nodes cannot access pod's local PortMappings
**What happened:** 
When node attempts to access its own local hostPort mapped to pod, traffic will never reach container networking stack but instead be dropped by TCP/IP stack.

I would expect traffic to be routed from TCP/IP stack to container networking stack. To achieve this, we need to substitute the [hcsshim.NatPolicy](
https://github.com/Azure/azure-container-networking/blob/75e1239132c31d6555baa967fb8f7b1f893e4d87/network/policy/policy_windows.go#L140)

This should be switched from V1 HNS to V2 HCN API and set LocalRoutedVIP flag. This will instrument the OS to add a route to the host which will make this traffic reach vSwitch.

Here is some sample code to replace the `hcsshim.NatPolicy`:
```
	natPolicy := hcn.PortMappingPolicySetting{
		InternalPort: portMappingPolicy.InternalPort,
		ExternalPort: portMappingPolicy.ExternalPort,
		Flags:        hcn.NatFlags(hcn.NatFlagsLocalRoutedVip),
	}
	switch portMappingPolicy.Protocol {
	case ProtocolTcp:
		natPolicy.Protocol = "TCP"
	case ProtocolUdp:
		natPolicy.Protocol = "UDP"
	}
```

**How to reproduce it:**  
1. Deploy a Windows Node *without* NodePort service pods running on the node.
2. Deploy a pod with hostPort mapping.
3. From the Node, curl node_ip:hostPort
4. Observe traffic will be dropped by TCP/IP

Kubernetes

Windows Server 2022 + Windows Server 2019

jpayne3506:
**Reason for Change**:

Enables the use of hostport mapping with HNSv2 by setting a flag in PortMappingPolicySetting. This reflects in hnsendpoint by showing a new policy for the route which is created by the OS. 

**Issue Fixed**:

Fixes #1863 

**Requirements**:

- [ x ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ x ] includes documentation
- [ x ] adds unit tests

**Notes**:
You can confirm the policy and route through the node by using 
`get-hnsendpoint | ? IPAddress -Like "node ip" | Convertto-json -d 16`
for exclusively policy use 
`get-hnsendpoint | ? IPAddress -Like "node ip" | Select Policies | Convertto-json -d 16`

Internal port / containerPort must match an already listening port on the pod. If not using the default service port 80, user will need to create and open port.


cni/network/network_windows.go
```diff
@@ -257,6 +257,7 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig) []policy.Policy {
 			InternalPort: uint16(mapping.ContainerPort),
 			VIP:          mapping.HostIp,
 			Protocol:     protocol,
+			Flags:        hnsv2.NatFlagsLocalRoutedVip, // iota'd, NatFlagsLocalRoutedVip = 1, uint32
```
vipul-21, line 260: Is this a constant value ? What does this mean ? ( Just curious)

jpayne3506: It is a constant value set here https://github.com/microsoft/hcsshim/blob/298b31d151ab799c1d7686f8ee9eec1cb4213926/hcn/hcnpolicy.go#L82 . 
Was trying to match other code snippets in cni where they match the TCP/UDP to constant values as well.

tamilmani1989: can you add a comment above this change saying "its to support hostport policy mapping"

vipul-21 (dismissed):

daschott (approved):
LGTM

tamilmani1989 (approved):

vipul-21 (approved):

PR #1924, opened 2023-04-21
perf: Store Lock timeout for CNI in Windows is increased to 1 minute.

AzureAhai:
perf: Performance Improvements of Azure CNI by increasing store Lock timeout for CNI in Windows to 1 minute.

**Reason for Change**:
The current value of lock timeout for Azure CNI is 10 second. This value will cause a lot of CNI add call timeout specially on large scale scenarios in Windows. To prevent this add call failure in Windows, we are increasing the value to 1 minutes on Windows to prevent subsequent delete request form Containerd for the failed add request.
**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/aitelemetry/telemetrywrapper.go
+++ b/aitelemetry/telemetrywrapper.go
@@ -116,9 +116,12 @@ func getMetadata(th *telemetryHandle) {
 		debugLog("[AppInsights] Error initializing kvs store: %v", err)
 		return
 	}
-
-	err = kvs.Lock(store.DefaultLockTimeout)
-	if err != nil {
+	// Acquire store lock. For windows 1m timeout is used while for Linux 10s timeout is assigned.
+	var lockTimeoutValue time.Duration = store.DefaultLockTimeout
+	if runtime.GOOS == "windows" {
+		lockTimeoutValue = store.DefaultLockTimeoutWindows
+	}
+	if err = kvs.Lock(lockTimeoutValue); err != nil {
 		log.Errorf("getMetadata: Not able to acquire lock:%v", err)
 		return
 	}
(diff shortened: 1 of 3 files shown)


aitelemetry/telemetrywrapper.go
```diff
@@ -116,9 +116,12 @@ func getMetadata(th *telemetryHandle) {
 		debugLog("[AppInsights] Error initializing kvs store: %v", err)
 		return
 	}
-
-	err = kvs.Lock(store.DefaultLockTimeout)
-	if err != nil {
+	// Acquire store lock. For windows 1m timeout is used while for Linux 10s timeout is assigned.
+	var lockTimeoutValue time.Duration = store.DefaultLockTimeout
```
tamilmani1989, line 120: l dont think we should change for telemetry service.. lets keep the same

tamilmani1989 (changes requested):

aitelemetry/telemetrywrapper.go
AzureAhai: I recall when I left the telemetry timeout unchanged the result got weird!

store/json.go
```diff
@@ -195,7 +197,7 @@ func (kvs *jsonFileStore) Lock(timeout time.Duration) error {
 		return errors.Wrap(err, "processLock acquire error")
 	}
 
-	log.Printf("Acquired process lock")
+	log.Printf("Acquired process lock with timeout value of %v ", timeout)
```
tamilmani1989, line 200: can you add unit for timeout in log

AzureAhai: I have tested the log it will print as 1m if it was less than a minute it will show 30s.

tamilmani1989 (approved):
lgtm

PR #1928, opened 2023-04-24
ci: cilium load test uses cns write conflist

camrynl:
**Reason for Change**:

Now that azure-cns:v1.4.47 is released, we can use CNS to write the cilium conflist in the load test pipeline.
Also updated deployment of overlay-cni to match that in the ACN PR pipeline in order to deploy the systemd-networkd init container bug fix. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
load test pipeline validation - https://msazure.visualstudio.com/One/_build/results?buildId=72132977&view=results


test/integration/manifests/cilium/cns-write-ovly.yaml
```diff
@@ -0,0 +1,228 @@
...
+          "InfrastructureNetworkID": "",
+          "NodeID": "",
+          "NodeSyncIntervalInSeconds": 30
+      },
+      "ChannelMode": "CRD",
+      "InitializeFromCNI": false,
+      "ManageEndpointState": true,
+      "ProgramSNATIPTables": false,
+      "EnableCNIConflistGeneration": true,
+      "CNIConflistFilepath": "/etc/cni/net.d/cilium.conflist",
```
vipul-21, line 226: Should this be `05-cilium.conflist` ?

test/integration/manifests/cilium/cns-write-ovly.yaml
```diff
@@ -0,0 +1,228 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
camrynl: maybe in the future we can move to that, for now let's use this separate ds as it doesn't require external variables to set.

test/integration/manifests/cilium/cns-write-ovly.yaml
camrynl: corrected ✅

.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -16,11 +16,48 @@ stages:
...
+                echo "install Cilium onto Overlay Cluster"
+                kubectl apply -f test/integration/manifests/cilium/cilium-agent
+                kubectl apply -f test/integration/manifests/cilium/cilium-operator
+                kubectl get po -owide -A
+                echo "deploy ip-masq-agent for overlay"
+                kubectl create -f test/integration/manifests/ip-masq-agent/ip-masq-agent.yaml --validate=false
+                cd test/integration/manifests/ip-masq-agent/
+                kubectl create configmap config-custom.yaml
+                kubectl create configmap config-reconcile.yaml
+                cd ../../../..
```
vipul-21, line 57: Wondering if we need to go back to the root directory ? Does the pipeline in different stages maintain that ?

vipul-21 (approved):

.pipelines/cni/cilium/cilium-cni-load-test.yaml
camrynl: hmm, maybe we don't need to. I'm not sure if the directory passes through the different stages. I grabbed this from the overlay-cilium e2e template, and I believe when I wrote that it was a precautionary measure.

PR #1929, opened 2023-04-24
Add log to happy path

bohuini:
**Reason for Change**:
DNC requiring to refresh CNS NC associations is not expected to happen too often, but when it does, we need to log all pertinent details. Added the log in CNS should to show refresh detected and initiated by DNC.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -925,15 +925,17 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
 	logger.Response(service.Name, getNetworkContainerResponses[0], getNetworkContainerResponses[0].Response.ReturnCode, err)
 }
 
-// getOrRefreshNetworkContainers is to check whether refresh association is needed.
+// getOrRefreshNetworkContainers is to check whether refresh association is needed. The state file in CNS will get updated if it is lost.
 // If received  "GET": Return all NCs in CNS's state file to DNC in order to check if NC refresh is needed
 // If received "POST": Store all the NCs (from the request body that client sent) into CNS's state file
 func (service *HTTPRestService) getOrRefreshNetworkContainers(w http.ResponseWriter, r *http.Request) {
 	switch r.Method {
 	case http.MethodGet:
+		logger.Printf("[Azure CNS] getOrRefreshNetworkContainers received GET")
 		service.handleGetNetworkContainers(w)
 		return
 	case http.MethodPost:
+		logger.Printf("[Azure CNS] getOrRefreshNetworkContainers received POST")
 		service.handlePostNetworkContainers(w, r)
 		return
 	default:


rbtr:
@bohuini pls read this https://github.com/Azure/azure-container-networking/pull/1921#issuecomment-1530069378

cns/restserver/api.go
```diff
@@ -919,11 +919,15 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
 // If received  "GET": Return all NCs in CNS's state file to DNC in order to check if NC refresh is needed
 // If received "POST": Store all the NCs (from the request body that client sent) into CNS's state file
 func (service *HTTPRestService) getOrRefreshNetworkContainers(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] NC refresh is initiated by DNC. The state file in CNS will get updated if it is lost.")
```
rbtr, line 922: unnecessary to add an informational log here if all branches of the switch have a log

cns/restserver/api.go
```diff
@@ -919,11 +919,15 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
+
 	switch r.Method {
 	case http.MethodGet:
+		logger.Printf("[Azure CNS] DNC is attempting to check if the state file in CNS is intact.")
```
rbtr, line 926: or, `GET from getOrRefreshNetworkContainers`

cns/restserver/api.go
```diff
@@ -919,11 +919,15 @@ func (service *HTTPRestService) getNetworkContainerByOrchestratorContext(w http.
...
 		service.handleGetNetworkContainers(w)
 		return
 	case http.MethodPost:
+		logger.Printf("[Azure CNS] The state file in CNS is lost. CNS will store all the NCs sent from DNC into state file.")
```
rbtr, line 930: or, `POST to getOrRefreshNetworkContainers`

rbtr (changes requested):
the log should not speculate on the _intent_ of what is happening, the log should describe what is happening concisely and in enough detail to be useful during debugging

cns/restserver/api.go
rbtr: you don't have to use my example, but it provides just as much info as yours without the potential of lying to me if that's not why this method was called

timraymond:
I'm with @rbtr here... I think concise logs are better because they tend to tell fewer lies about what's going on as things change over time. Log just the facts, given what's known at the time the log is generated.  Given what I see, it would likely be better to have a middleware that logs requests (path, method, some request ID), and responses (HTTP response code, end-to-end time, the same ID for correlation). All of those should be structured using zap. Something like:

```go
func logHTTP(next http.Handler) http.Handler {
	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
		logger.Info("request", zap.String("uri", r.URL.RequestURI()), zap.String("method", r.Method))
		// need to wrap the ResponseWriter in something that can capture the response code
		wrw := WrappedResponseWriter{rw} // this doesn't exist and needs to be written
		next(wrw, r)
		logger.Info("response", zap.Int("code", wrw.ResponseCode)) // etc.
	})
}

bohuini:
@rbtr Revised the log and rebased branch. Please review again when you get a chance. Thanks!

bohuini:
@timraymond That's a good point and I do agree. But for this time I am just making it simple for integration testing purposes.

cns/restserver/api.go
bohuini: Removed.

cns/restserver/api.go
bohuini: Make sense. Addressed.

cns/restserver/api.go
bohuini: Addressed.

timraymond:
LGTM 👍 Awaiting @rbtr to hit the approval button.

rbtr (approved):

PR #1934, opened 2023-04-28
Update to cilium v1.12.8

camrynl:
**Reason for Change**:

Pipelines will now test with cilium v1.12.8

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr:
what's the plan for larger migrations? should we parameterize this out into a(nother) test matrix so we can do [1.12.8, 1.13.x] etc?

rbtr (dismissed):

camrynl:
> what's the plan for larger migrations? should we parameterize this out into a(nother) test matrix so we can do [1.12.8, 1.13.x] etc?

For now, we just test the next version against our load tests and the pr pipeline. Do you mean to have a separate pipeline where we can test multiple versions at once?

rbtr:
I imagine it would be straightforward to run this pipeline as a matrix with the version parameterized. Like how we do ARM/AMD, so that we can transition in to new versions of the dependency without a hard cutover.

camrynl:
> I imagine it would be straightforward to run this pipeline as a matrix with the version parameterized. Like how we do ARM/AMD, so that we can transition in to new versions of the dependency without a hard cutover.

I see, I will move towards parameterized versions in future states

PR #1940, opened 2023-05-03
fix: generate cni conflist if NCs already exist in state

thatmattlong:
**Reason for Change**:
Fixes a bug in CNS where it only generates the CNI conflist if a new NC is discovered and programmed. Instead it should also generate the CNI conflist if the NC is already present in CNS state.

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


camrynl (dismissed):
looks good to me ✔️

rbtr (approved):

rbtr (approved):

PR #1946, opened 2023-05-04
CNS - Remove vnet join cache check during NC publish

ramiro-gamarra:
**Reason for Change**:

This PR reverts back a change introduced in #1825, where, during an NC publish, we check if a vnet join has been performed against nmagent and skip the call if so. This change introduced a regression due to nmagent behavior: if nmagent crashes while subscribed to a vnet with no NCs published, it will not re-subscribe once restarted, making the CNS cache invalid and resulting in new NCs published erroring with unauthorized codes. Removing this cache check will make it such that the vnet is always joined before any NC is published.

We should consider reintroducing this check (which greatly minimizes wireserver calls) once we can correctly remove a vnet from the cache when it has no NC.

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -1206,41 +1206,39 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
 
 	ctx := r.Context()
 
-	if !service.isNetworkJoined(req.NetworkID) {
-		joinResp, err := service.wsproxy.JoinNetwork(ctx, req.NetworkID) //nolint:govet // ok to shadow
-		if err != nil {
-			resp := cns.PublishNetworkContainerResponse{
-				Response: cns.Response{
-					ReturnCode: types.NetworkJoinFailed,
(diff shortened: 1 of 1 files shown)


cns/restserver/api.go
```diff
@@ -1206,42 +1206,40 @@ func (service *HTTPRestService) publishNetworkContainer(w http.ResponseWriter, r
...
-				Response: cns.Response{
-					ReturnCode: types.NetworkJoinFailed,
-					Message:    fmt.Sprintf("failed to join network %s: %v", req.NetworkID, err),
-				},
-				PublishErrorStr: err.Error(),
-			}
-			respondJSON(w, http.StatusOK, resp) // legacy behavior
-			logger.Response(service.Name, resp, resp.Response.ReturnCode, err)
-			return
+	joinResp, err := service.wsproxy.JoinNetwork(ctx, req.NetworkID) //nolint:govet // ok to shadow
```
msvik, line 1209: Is there a way to add unit test to prevent it from happening again by accident?

PR #1965, opened 2023-05-17
ci: disable kube-proxy for test clusters

camrynl:
**Reason for Change**:

- creates new hackfile targets to spin up byocni clusters without kube-proxy for cilium pipeline tests
- includes small fix for load test after node restart, bug was found while testing this pr in load test pipeline.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/cni/cilium/cilium-cni-load-test.yaml
+++ b/.pipelines/cni/cilium/cilium-cni-load-test.yaml
@@ -120,7 +120,7 @@ stages:
             name: "RestartNodes"
             displayName: "Restart Nodes"
           - script: |
-              bash hack/scripts/validate_state.sh
+              bash hack/scripts/validate_state.sh -r true
             name: "ValidateState"
             displayName: "Validate State"
             retryCountOnTaskFailure: 3
--- a/.pipelines/cni/cilium/cilium-cni-load-test.yaml
+++ b/.pipelines/cni/cilium/cilium-cni-load-test.yaml
@@ -15,9 +15,10 @@ stages:
               addSpnToEnvironment: true
               inlineScript: |
                 set -ex
-                make -C ./hack/swift config-dir
                 make -C ./hack/swift azcfg AZCLI=az REGION=$(LOCATION)
                 make -C ./hack/swift overlay-byocni-up AZCLI=az REGION=$(LOCATION) SUB=$(SUBSCRIPTION_ID) CLUSTER=${RESOURCE_GROUP} NODE_COUNT=10 VM_SIZE=Standard_DS4_v2
+                echo "Update cluster and remove kube-proxy"
+                az aks update -n CLUSTER=${{ parameters.clusterName }}-$(make revision) -g CLUSTER=${RESOURCE_GROUP} --kube-proxy-config ./hack/swift/kube-proxy.json
             name: "CreateAksCluster"
             displayName: "Create AKS Cluster"
   - stage: install_cilium
--- a/.pipelines/cni/cilium/cilium-cni-load-test.yaml
+++ b/.pipelines/cni/cilium/cilium-cni-load-test.yaml
@@ -17,8 +17,6 @@ stages:
                 set -ex
                 make -C ./hack/swift azcfg AZCLI=az REGION=$(LOCATION)
                 make -C ./hack/swift overlay-no-kube-proxy-up AZCLI=az REGION=$(LOCATION) SUB=$(SUBSCRIPTION_ID) CLUSTER=${RESOURCE_GROUP} NODE_COUNT=10 VM_SIZE=Standard_DS4_v2
-                echo "Update cluster and remove kube-proxy"
-                az aks update -n ${RESOURCE_GROUP} -g ${RESOURCE_GROUP} --kube-proxy-config ./hack/swift/kube-proxy.json
             name: "CreateAksCluster"
             displayName: "Create AKS Cluster"
   - stage: install_cilium
--- a/.pipelines/cni/cilium/cilium-cni-load-test.yaml
+++ b/.pipelines/cni/cilium/cilium-cni-load-test.yaml
@@ -18,7 +18,7 @@ stages:
                 make -C ./hack/swift azcfg AZCLI=az REGION=$(LOCATION)
                 make -C ./hack/swift overlay-byocni-up AZCLI=az REGION=$(LOCATION) SUB=$(SUBSCRIPTION_ID) CLUSTER=${RESOURCE_GROUP} NODE_COUNT=10 VM_SIZE=Standard_DS4_v2
                 echo "Update cluster and remove kube-proxy"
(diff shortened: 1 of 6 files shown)


.pipelines/cni/cilium/cilium-cni-load-test.yaml
```diff
@@ -17,6 +17,8 @@ stages:
                 set -ex
                 make -C ./hack/swift azcfg AZCLI=az REGION=$(LOCATION)
                 make -C ./hack/swift overlay-byocni-up AZCLI=az REGION=$(LOCATION) SUB=$(SUBSCRIPTION_ID) CLUSTER=${RESOURCE_GROUP} NODE_COUNT=10 VM_SIZE=Standard_DS4_v2
+                echo "Update cluster and remove kube-proxy"
+                az aks update -n ${RESOURCE_GROUP} -g ${RESOURCE_GROUP} --kube-proxy-config ./hack/swift/kube-proxy.json
```
vipul-21, line 21: I saw the note regarding the kube-proxy not being read by the make shell. What was the error message ? 
"-C" provides the directory and it should be able to read any file(given read permission)

camrynl: Testing locally, I was able to map the json config and use to bring up the cluster. I mounted the file using -v and also tried with --mount in the AZCLI docker run cmd. I also have verfied that the current directory is correct. 
However the same commands in the pipeline fails with **'kube-proxy.json is not valid file, or not accessible'**

I've been testing with different paths to the config, but azp still can't pick it up. Open to other suggestions if you have any.

vipul-21: Not an expert in docker but doesn't `--rm` remove the directories ? 
When you say verified the directory, you mean you were able to see the kube-proxy file but it was not read by pipeline.

camrynl: yes, but the contents are mounted each time we run AZCLI so it should be there. I found in azp, the config file goes to **root/tmpsrc/hack/swift**, so that is path I used to list directory and to call in the create cluster command. In this run you can see that the file is there in that directory, but shows the error when I try to use it.

https://msazure.visualstudio.com/One/_build/results?buildId=73729842&view=logs&j=3713fd4c-b129-5ac5-2b38-c7386192ca24&t=bd94de76-3239-5e11-ae5b-ed0fdc2178e5&l=140

When I run this locally, the config file is directly in **root/tmpsrc** and I'm able to get it via `--kube-proxy-config root/tmpsrc/kube-proxy.json`

vipul-21: Is the docker command being used ? As I see `AZCLI`  variable being passed as `az`. So if we are running the make file from the `hack/swift` dir and the file should be accessed via `./kube-proxy.json`. Did we try that already?
I think to use docker we would need to remove the `AZCLI` argument from the pipeline yaml call.

camrynl: the docker command is used, and `az` is containerized. I did try to access the file with just` ./kube-proxy.json`, but it also failed.

vipul-21: Tried with cni load test pipeline: https://msazure.visualstudio.com/One/_build/results?buildId=73752234&view=logs&j=3713fd4c-b129-5ac5-2b38-c7386192ca24&t=bd94de76-3239-5e11-ae5b-ed0fdc2178e5&l=79, it seems to work. Maybe the az cli version was different for you ?

camrynl: interesting.. just tried this in the pipeline as well and it's working. Does this work for you when you run byocni-up from your dev? I get the error locally

vipul-21: I tried locally but I think my preview version is outdated.(Need to check).
Getting the error invalid argument `kube-proxy-config`

camrynl: hmm, I tried updating to the latest preview and I still see the error. For now I have made separate no-kube-proxy targets that we can use specifically for the pipeline. If we need to use for a cluster in our own setup, we will just have to modify the config path to /tmpsrc/kube-proxy.json instead

hack/scripts/validate_state.sh
```diff
@@ -17,7 +17,7 @@ do
     echo "Node internal ip: $node_ip"
     # Check pod count after restarting nodes, statefile does not exist after restart
     echo "checking whether the node has any pods deployed to it or not"
-    pod_count=$(kubectl get pods -A -o wide | grep "$node_name" -c)
+    pod_count=$(kubectl get pods -n kube-system -o wide | grep "$node_name" -c)
```
vipul-21, line 20: I think we are deploying the pods to the `default` ns and validating their ips with the state file.

vipul-21 (changes requested):

hack/scripts/validate_state.sh
vipul-21: If we want to check all namespaces. `-A` should have worked. Why are we restricting it to `kube-system`. Why is getting skipped ?
The idea of the check is to have something deployed after the restart as well so that the state file exists to validate.

camrynl: Previously the check was only looking in default, I changed it to `-A` a couple weeks back in a different PR.
Not restricting to kube-system permanently, I was testing and trying to see what was happening in the restart node check. 
In all the runs I've had where the clusters successfully created without kube-proxy, the restart node check has been failing to validate state, it gets stuck `trying to get the azure_endpoints` 
This seems like the validate state check is trying to validate when the nodes still don't have pods yet. How can we refine this pod count command to get the correct # pods on the node?

vipul-21: We know that azure-endpoints won't exists when we restart the nodes if no new deployments happened after the restart. We can pass the variables `restart=true` in the pipeline yaml when we call the `validate_test.sh script` something like `./validate_test.sh --restart`
Then we can remove the pod check if condition altogether and add a limit to the for loop for getting the azure_endpoints.json( like try 5 times instead of forever). If we still fail to get the endpoints files, there are 2 cases to handle:
1. Restart node case - Check if we passed the `restart` variable, if so it is safe to assume that file didn't get created as no pods deployed to that node. Continue to next node.
2. Other cases - Fail in this case, as we are not in the `restart` phase and expect the azure_endpoints file to be present.

Let me know if you want to chat over this.

camrynl: update -- load test has been fixed with the necessary changes

tamilmani1989 (approved):

tamilmani1989 (approved):

PR #1969, opened 2023-05-18
perf: [NPM] [LINUX] add NetPols in background

huntergregory:
**Reason for Change**:
NPM must use nftables on Ubuntu 22 (k8s 1.25+). NPM SysCalls into nftables take much longer than into legacy iptables. Customers with ~100+ NetPols can see: 
- Application timeouts (due to increased latencies while xtables lock is held). 
- Delay in applying NetPols.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
## Latency Comparison
*Ran 2 experiments for each version. Used b-series burstable VMs.*

Cluster has 640 services with 2 Pods each.

| Version                      | Applying 640 NetPols  | Max iptables Restore Latency | List all rules (once at bootup)   |
|------------------------|--------------------------|---------------------------------|-----------------------------------|
| Legacy (Ubuntu 18)   | 4 minutes                     |  0.4 seconds                           | 0.01 seconds                             |
| nftables (Ubuntu 22) | 9-22 hours                    | 3-5 minutes                           | 5-11 minutes                             |
| This PR (for nftables) | 13-17 minutes               | 3-5 minutes                           | 5-11 minutes (same codepath) |

## Fix: Long Term
[nftables do not scale for services · Issue #96018 · kubernetes/kubernetes (github.com)](https://github.com/kubernetes/kubernetes/issues/96018#issuecomment-1193092167)

iptables v1.8.8 has scale improvements. However, AKS is on iptables v1.8.4. 

## Fix: Near Term (this PR)
Optimizing for ADD NetworkPolicy.

### Current Design
ADD NetworkPolicy:
1. Create all IPSets at once via ipset restore.
2. Create all iptables rules/chains at once via iptables-restore.
Current Design for Linux NPM

### This PR
Since iptables SysCalls take so long, add iptables rules for *multiple NetworkPolicies at once*. 
This change will only take effect if `nftables` is present on machine & NPM's ConfigMap has `NetPolInBackground=true`.
Redesign for Linux NPM

### Pipelines
- Stop running pipelines for NPM v1.
- Enable Linux Scale test.

### Metrics
Adds the following metrics:
1. `npm_linux_iptables_restore_latency_seconds`
4. `npm_linux_iptables_delete_latency_seconds_bucket`
```


# HELP npm_linux_iptables_delete_latency_seconds Latency in seconds to delete an iptables rule
# TYPE npm_linux_iptables_delete_latency_seconds histogram
npm_linux_iptables_delete_latency_seconds_bucket{le="0.016"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="0.032"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="0.064"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="0.128"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="0.256"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="0.512"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="1.024"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="2.048"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="4.096"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="8.192"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="16.384"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="32.768"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="65.536"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="131.072"} 0
npm_linux_iptables_delete_latency_seconds_bucket{le="+Inf"} 0
npm_linux_iptables_delete_latency_seconds_sum 0
npm_linux_iptables_delete_latency_seconds_count 0
# HELP npm_linux_iptables_restore_latency_seconds Latency in seconds to restore iptables rules by operation label (add/delete NetPol)
# TYPE npm_linux_iptables_restore_latency_seconds histogram
npm_linux_iptables_restore_latency_seconds_bucket{operation="create",le="0.016"} 0
...
```


npm/config/config.go
```diff
@@ -4,6 +4,8 @@ import "github.com/Azure/azure-container-networking/npm/util"
 
 const (
 	defaultResyncPeriod         = 15
+	defaultNetPolMaxBatches     = 100
```
vakalapa, line 7: Any reason we cannot reuse the defaultApplyMaxBatches and defaultApplyInterval ?

npm/pkg/dataplane/dataplane.go
```diff
@@ -145,6 +154,29 @@ func (dp *DataPlane) RunPeriodicTasks() {
...
+			case <-dp.stopChannel:
+				return
+			case <-ticker.C:
+				// locks policy manager but can be interrupted
+				dp.netPolInfo.Lock()
+				if dp.netPolInfo.numBatches == 0 {
+					dp.netPolInfo.Unlock()
+					return
+				}
+				dp.netPolInfo.numBatches = 0
```
vakalapa, line 172: why are we setting this to 0 ?

npm/pkg/dataplane/policies/policymanager_linux.go
```diff
@@ -41,10 +42,92 @@ Known errors that we should retry on:
...
+			}
+		}
+
+		if ops[len(ops)-1] == add {
+			toAdd = append(toAdd, policy)
+		}
+	}
+
+	if len(toRemove) > 0 {
+		if err := pMgr.removeAllPolicies(toRemove); err != nil {
```
vakalapa, line 87: need to check if there any race conditions here ? 

1. If a policy is added and delete again within a same batch call. -> i think we are safe since the key wont be present in ou local cache
2. what if multiple updates come for a same policy -> safe for that too !? 

any other senarios ?

npm/pkg/dataplane/policies/policymanager_linux.go
```diff
@@ -41,10 +42,92 @@ Known errors that we should retry on:
...
+	if len(pMgr.policyMap.linuxDirtyCache) == 0 {
+		return nil
+	}
+
+	toRemove := make([]*NPMNetworkPolicy, 0)
+	toAdd := make([]*NPMNetworkPolicy, 0)
+	for key, ops := range pMgr.policyMap.linuxDirtyCache {
+		policy, ok := pMgr.policyMap.cache[key]
+		if !ok {
+			metrics.SendErrorLogAndMetric(util.IptmID, "error: failed to find dirty policy in cache. key: %s", key)
```
vakalapa, line 70: error might be excessive, just a warn should be good. This will be more than common scenario ? If the dirtycache is not in cache it means we need to/ or alredy deleted it ?

npm/pkg/dataplane/policies/policymanager_linux.go
```diff
@@ -41,10 +42,92 @@ Known errors that we should retry on:
...
+			continue
+		}
+
+		for _, op := range ops {
+			if op == remove {
+				toRemove = append(toRemove, policy)
```
vakalapa, line 76: How will you get the previous version of the network policy to be deleted ? we would have overwritten the cache with latest copy right ?

npm/pkg/dataplane/policies/policymanager_linux.go
```diff
@@ -63,18 +180,43 @@ func (pMgr *PolicyManager) addPolicy(networkPolicy *NPMNetworkPolicy, _ map[stri
...
-	// 1. Delete jump rules from ingress/egress chains to ingress/egress policy chains.
-	// We ought to delete these jump rules here in the foreground since if we add an NP back after deleting, iptables-restore --noflush can add duplicate jump rules.
-	deleteErr := pMgr.deleteOldJumpRulesOnRemove(networkPolicy)
-	if deleteErr != nil {
-		return npmerrors.SimpleErrorWrapper("failed to delete jumps to policy chains", deleteErr)
+	for _, networkPolicy := range networkPolicies {
+		// 1. Delete jump rules from ingress/egress chains to ingress/egress policy chains.
+		// We ought to delete these jump rules here in the foreground since if we add an NP back after deleting, iptables-restore --noflush can add duplicate jump rules.
+		deleteErr := pMgr.deleteOldJumpRulesOnRemove(networkPolicy)
+		if deleteErr != nil {
```
huntergregory, line 217: need to ignore errors here

npm/pkg/dataplane/dataplane.go
```diff
@@ -433,7 +472,11 @@ func (dp *DataPlane) RemovePolicy(policyKey string) error {
 		return err
 	}
 
-	return dp.applyDataPlaneNow(contextApplyDP)
+	if err := dp.applyDataPlaneNow(contextApplyDP); err != nil {
+		return err
+	}
+
+	return dp.incrementBatchAndReconcileDirtyNetPolsIfNeeded(contextDelNetPol)
```
huntergregory, line 479: can actually just return nil here, skip error propagation to NetPol controller

npm/config/config.go
huntergregory: discussed offline to have separate threads/config for windows and Linux

npm/pkg/dataplane/dataplane.go
huntergregory: now setting num batches to 0 after successful reconcile

npm/pkg/dataplane/policies/policymanager_linux.go
huntergregory: TODO: UTs for all these edge cases

npm/pkg/dataplane/policies/policymanager_linux.go
huntergregory: now track `deletedState` for what we need:
1. policy key
2. namespace
3. pod selector list

npm/pkg/dataplane/policies/policymanager_linux.go
huntergregory: we actually already ignore `doesNotExistErrorCode` in case the rule doesn't exist

npm/azure-npm.yaml
```diff
@@ -148,19 +148,21 @@ metadata:
...
-        "ApplyMaxBatches":             100,
-        "ApplyIntervalInMilliseconds": 500,
-        "MaxBatchedACLsPerPod":        30,
+        "ResyncPeriodInMinutes":          15,
+        "ListeningPort":                  10091,
+        "ListeningAddress":               "0.0.0.0",
+        "ApplyMaxBatches":                100,
+        "ApplyIntervalInMilliseconds":    500,
+        "MaxBatchedACLsPerPod":           30,
+        "IPTablesIntervalInMilliseconds": 500,
```
vakalapa, line 157: Maxiptables config missing

npm/pkg/dataplane/dataplane.go
```diff
@@ -547,9 +601,92 @@ func (dp *DataPlane) deleteIPSetsAndReferences(sets []*ipsets.TranslatedIPSet, n
...
+		m := dp.netPolInfo.toDeleteSelectorReferences
+		if referenceType == ipsets.NetPolType {
+			m = dp.netPolInfo.toDeleteNetPolReferences
+		}
+		m[policyKey] = append(m[policyKey], prefixName)
+	}
+}
+
+func temporaryReference(policyKey string) string {
+	return "TMP-" + policyKey
```
vakalapa, line 691: make this TMP(or some delimiter) a utils const.

npm/pkg/dataplane/policies/policymanager.go
```diff
@@ -39,11 +40,38 @@ type PolicyManagerCfg struct {
...
+	direction       Direction
+	podSelectorList []SetInfo
+	wasInKernel     bool
 }
 
 type PolicyMap struct {
 	sync.RWMutex
-	cache map[string]*NPMNetworkPolicy
+	cache           map[string]*NPMNetworkPolicy
+	linuxDirtyCache map[string][]*event
```
vakalapa, line 73: can you add a comment here on what the key ?

npm/pkg/dataplane/dataplane.go
```diff
@@ -547,9 +601,92 @@ func (dp *DataPlane) deleteIPSetsAndReferences(sets []*ipsets.TranslatedIPSet, n
 				return npmerrors.Errorf(npmErrorString, false, fmt.Sprintf("[DataPlane] failed to RemoveFromList in deleteIPSetReferences with err: %s", err.Error()))
 			}
 		}
-
-		// Try to delete these IPSets
-		dp.ipsetMgr.DeleteIPSet(set.Metadata.GetPrefixName(), false)
```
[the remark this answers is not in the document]
huntergregory: need to change this logic in restore file:
`-D 1.2.3.4/32 nomatch` --> `-D 1.2.3.4/32`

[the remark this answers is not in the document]
huntergregory: updated

npm/examples/windows/azure-npm.yaml
```diff
@@ -140,19 +140,22 @@ metadata:
...
         "Toggles": {
             "EnablePrometheusMetrics": true,
             "EnablePprof":             true,
             "EnableHTTPDebugAPI":      true,
             "EnableV2NPM":             true,
-            "PlaceAzureChainFirst":    true,
+            "PlaceAzureChainFirst":    false,
             "ApplyIPSetsOnNeed":       false,
-            "ApplyInBackground":       true
+            "ApplyInBackground":       true,
```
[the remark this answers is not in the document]
huntergregory: this is for Windows only right now. Includes:
1. IPSets
2. NetPols based on UpdatePodCache

npm/pkg/dataplane/dataplane.go
```diff
@@ -548,8 +602,97 @@ func (dp *DataPlane) deleteIPSetsAndReferences(sets []*ipsets.TranslatedIPSet, n
...
+		m := dp.netPolInfo.toDeleteSelectorReferences
+		if referenceType == ipsets.NetPolType {
+			m = dp.netPolInfo.toDeleteNetPolReferences
+		}
+		m[policyKey] = append(m[policyKey], prefixName)
+	}
+}
+
+func temporaryReference(policyKey string) string {
+	return util.TmpIPSetReferenceDelimiter + policyKey
```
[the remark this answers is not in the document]
huntergregory: resolving this since we won't use tmp references anymore

npm/azure-npm.yaml
```diff
@@ -148,19 +148,22 @@ metadata:
...
             "EnablePrometheusMetrics": true,
             "EnablePprof":             true,
             "EnableHTTPDebugAPI":      true,
             "EnableV2NPM":             true,
-            "PlaceAzureChainFirst":    true,
+            "PlaceAzureChainFirst":    false,
             "ApplyIPSetsOnNeed":       false,
-            "ApplyInBackground":       true
+            "ApplyInBackground":       true,
+            "NetPolInBackground":    true
```
huntergregory, line 167: todo: fix whitespace

npm/pkg/dataplane/dataplane.go
```diff
@@ -24,48 +23,43 @@ const (
...
 	ApplyInterval     time.Duration
+	// NetPolInBackground is currently used in Linux to apply NetPol controller Add events in the background
+	NetPolInBackground bool
+	MaxPendingNetPols  int
+	NetPolInterval     time.Duration
 	*ipsets.IPSetManagerCfg
 	*policies.PolicyManagerCfg
 }
 
-type endpointCache struct {
```
huntergregory, line 40: moved all types to types.go

npm/pkg/dataplane/policies/policymanager_linux.go
```diff
@@ -174,12 +181,14 @@ func (pMgr *PolicyManager) deleteJumpRule(policy *NPMNetworkPolicy, direction Un
 	}
 
 	specs = append([]string{baseChainName}, specs...)
+	timer := metrics.StartNewTimer()
 	errCode, err := pMgr.runIPTablesCommand(util.IptablesDeletionFlag, specs...)
+	metrics.RecordIPTablesDeleteLatency(timer)
 	// if this actually happens (don't think it should), could use ignoreErrorsAndRunIPTablesCommand instead with: "Bad rule (does a matching rule exist in that chain?)"
-	if err != nil && errCode != doesNotExistErrorCode {
+	if err != nil && errCode != doesNotExistErrorCode && errCode != couldntLoadTargetErrorCode {
```
huntergregory, line 188: add couldntLoadTargetErrorCode in case the netpol chain we're deleting from somehow doesn't exist

npm/pkg/dataplane/policies/chain-management_linux.go
```diff
@@ -384,6 +394,7 @@ func (pMgr *PolicyManager) creatorForBootup(currentChains map[string]struct{}) *
 	// Step 2.1 in bootup() comment: cleanup old NPM chains, and configure base chains and their rules
 	// To leave NPM deactivated, don't specify any rules for AZURE-NPM chain.
 	creator := pMgr.newCreatorWithChains(chainsToCreate)
+	creator.Verbose()
```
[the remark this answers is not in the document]
huntergregory: removed

npm/pkg/dataplane/ipsets/ipsetmanager_linux.go
```diff
@@ -478,6 +481,7 @@ func (iMgr *IPSetManager) fileCreatorForApplyWithSaveFile(maxTryCount int, saveF
 // NOTE: duplicate code in the first step in this function and fileCreatorForApplyWithSaveFile
 func (iMgr *IPSetManager) fileCreatorForApply(maxTryCount int) *ioutil.FileCreator {
 	creator := ioutil.NewFileCreator(iMgr.ioShim, maxTryCount, ipsetRestoreLineFailurePattern) // TODO make the line failure pattern into a definition constant eventually
+	creator.Verbose()
```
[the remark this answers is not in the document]
huntergregory: removed this & all changes to ipsets/ pkg

npm/pkg/dataplane/policies/policymanager_linux.go
huntergregory: irrelevant. moved to a new design

npm/pkg/dataplane/policies/policymanager_linux.go
huntergregory: irrelevant. moved to a new design

huntergregory:
should we just use the existing Prometheus metrics instead? We have one for adding NetPols:
```
npm_add_iptables_rule_exec_time{quantile="0.5"} NaN
npm_add_iptables_rule_exec_time{quantile="0.9"} NaN
npm_add_iptables_rule_exec_time{quantile="0.99"} NaN
npm_add_iptables_rule_exec_time_sum 7.775247
npm_add_iptables_rule_exec_time_count 2
```

huntergregory:
TODO: only apply netpol in background if nf-tables is present

npm/pkg/dataplane/dataplane.go
```diff
@@ -366,52 +394,151 @@ func (dp *DataPlane) applyDataPlaneNow(context string) error {
...
-	if dp.inBootupPhase() {
-		// During bootup phase, the Pod controller will not be running.
-		// We don't need to worry about adding Policies to Endpoints, so we don't need IPSets in the kernel yet.
-		// Ideally, we get all NetworkPolicies in the cache before the Pod controller starts
-		err = dp.incrementBatchAndApplyIfNeeded(contextAddNetPol)
+// addPoliciesWithRetry tries adding all policies. If this fails, it tries adding policies one by one.
+// The caller must lock netPolQueue.
+func (dp *DataPlane) addPoliciesWithRetry(context string) {
+	netPols := dp.netPolQueue.dump()
+	klog.Infof("[DataPlane] adding policies %+v", netPols)
```
vakalapa, line 422: Can we make this a debug statement ?

npm/pkg/dataplane/dataplane.go
```diff
@@ -366,52 +394,151 @@ func (dp *DataPlane) applyDataPlaneNow(context string) error {
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// 2. get Endpoints
+	var err error
+	var endpointList map[string]string
+	if !dp.inBootupPhase() {
```
vakalapa, line 502: Is this condition correct ? the only condition for wasInBootPhase to be true is if dp.inBootupPhase() returns true. 

here you are checking the opposite and wasInBootPhase , i think we will never hit this condition ? unless i am missing something ?

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -292,8 +292,17 @@ func (dp *DataPlane) getSelectorIPSets(policy *policies.NPMNetworkPolicy) map[st
...
 
-func (dp *DataPlane) getEndpointsToApplyPolicy(policy *policies.NPMNetworkPolicy) (map[string]string, error) {
-	selectorIPSets := dp.getSelectorIPSets(policy)
+func (dp *DataPlane) getEndpointsToApplyPolicies(netPols []*policies.NPMNetworkPolicy) (map[string]string, error) {
+	if len(netPols) != 1 {
+		return nil, ErrIncorrectNumberOfNetPols
+	}
+
+	var netPol *policies.NPMNetworkPolicy
+	for _, n := range netPols {
```
vakalapa, line 301: why this loop, just say netpol := netpols[0]

npm/pkg/dataplane/dataplane.go
huntergregory: Seems like we don't have debug logging capability via klog

npm/pkg/dataplane/dataplane.go
huntergregory: The idea is to protect against a race:
1. DP in bootup phase.
2. dp.incrementBatchAndApplyIfNeeded() is called in last iteration of above loop.
3. dp.FinishBootupPhase() in another thread

Few more details in comment below

npm/pkg/dataplane/dataplane.go
```diff
@@ -89,32 +85,38 @@ func NewDataPlane(nodeName string, ioShim *common.IOShim, cfg *Config, stopChann
...
 
 	err := dp.BootupDataplane()
 	if err != nil {
 		klog.Errorf("Failed to reset dataplane: %v", err)
 		return nil, err
 	}
+
+	// Prevent netpol in background unless we're in Linux and using nftables.
+	// This step must be performed after bootupDataplane() because it calls util.DetectIptablesVersion(), which sets the proper value for util.Iptables
+	dp.netPolInBackground = cfg.NetPolInBackground && !util.IsWindowsDP() && (strings.Contains(util.Iptables, "nft") || dp.debug)
```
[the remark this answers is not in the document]
huntergregory: discussed offline to only do this for nft. We will keep codepath the same for legacy, which doesn't need the performance boost

npm/pkg/dataplane/dataplane.go
```diff
@@ -364,18 +393,69 @@ func (dp *DataPlane) applyDataPlaneNow(context string) error {
...
+		klog.Infof("[DataPlane] [%s] applying now since reached maximum batch count: %d", contextAddNetPol, newCount)
+		dp.addPoliciesWithRetry(contextAddNetPol)
+	}
+	return nil
+}
+
+// addPoliciesWithRetry tries adding all policies. If this fails, it tries adding policies one by one.
+// The caller must lock netPolQueue.
+func (dp *DataPlane) addPoliciesWithRetry(context string) {
+	netPols := dp.netPolQueue.dump()
```
matmerr, line 420: storing a queue of netpols as a map in background, but when applied we convert from map back to slice, any reason to not store background queue as a slice and then avoid this conversion?

huntergregory: having the map will help with `dp.RemovePolicy()`, where we remove the NetPol from the queue

vakalapa (approved):

matmerr (approved):

huntergregory:
> @huntergregory why not update the iptables binary that is used in the docker image?

Hi @xanecs, the limitation seems to be in the iptables version installed on the AKS node

PR #1976, opened 2023-05-23
fix: don't use 0th ip as gateway for ipv6

thatmattlong:
**Reason for Change**:
For overlay on windows, we are setting the gateway IP to the 0th IP of the the block assigned to each node. For ipv6, we want to skip this behavior and pass empty gateway to CNI. This PR fixes that.

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/conversion_test.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion_test.go
@@ -2,6 +2,7 @@ package nodenetworkconfig
 
 import (
 	"strconv"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


ashvindeodhar (approved):

rbtr (approved):

ashvindeodhar:
Let's wait on this. We may want to abandon this as this seems like a desired behavior for ipv6 as well. @paulyufan2 to validate the master and confirm

thatmattlong:
Abandoning since we do want this behavior for ipv6 as well

PR #1977, opened 2023-05-23
Making consolidated 'overlay' option to cns ipam mode

pjohnst5:
This is to make a third, consolidated, 'overlay' cns ipam mode

Once this is merged, we will do the following to safely migrate towards this option:
- First, add third "overlay" option (keeps v4overlay and dualstack options still valid)
- Switch dualstack to "overlay" in AKS-RP
- Switch v4overlay to "overlay" in AKS-RP
- Remove "v4overlay" and "dualstack" options in CNS

See our teams channel chat about this:
> Ashvin:
> Matthew Long do you see any issue with updating the the cns ipam mode in CNI conflist to just 'overlay' for v4 overlay?
> Matthew Long
> Nope no issue, maybe we’ll keep both for a little while in case of any rollbacks for backwards compatibility but eventually we can just remove the v4 part
> Neha Aggarwal
> I remember it was required to keep different handling in CNI too. Tamilmani Manoharan do you remember it? 
> Tamilmani Manoharan
> Paul is right.. cni requires mode only to differentiate podsubnet or overlay but not needed for v4 and dualstack. instead of mode, based on ipaddress cni can detect if dualstack or not.. so just keeping overlay makes sense.. have to update cni code accordingly (shouldn't be a big change)...


cni/network/network_test.go
```diff
@@ -1084,9 +1084,21 @@ func TestGetNetworkName(t *testing.T) {
...
 }
 
-func TestGetOverlayNatInfo(t *testing.T) {
+func TestGetV4OverlayNatInfo(t *testing.T) {
 	nwCfg := &cni.NetworkConfig{ExecutionMode: string(util.V4Swift), IPAM: cni.IPAM{Mode: string(util.V4Overlay)}}
 	natInfo := getNATInfo(nwCfg, nil, false)
+	require.Empty(t, natInfo, "v4overlay natInfo should be empty")
+}
+
+func TestGetDualStackOverlayNatInfo(t *testing.T) {
```
ashvindeodhar, line 1093: I believe we don't call these NAT info funcs for dualstackoverlay so no need to add these. I would suggest keeping the original one

cns/service/main.go
```diff
@@ -536,6 +537,8 @@ func main() {
 		switch scenario := cniConflistScenario(scenarioString); scenario {
 		case scenarioV4Overlay:
 			conflistGenerator = &cniconflist.V4OverlayGenerator{Writer: writer}
+		case scenarioOverlay:
+			conflistGenerator = &cniconflist.OverlayGenerator{Writer: writer}
```
ashvindeodhar, line 541: should we be replacing the existing v4overlay generator with this new one?

ashvindeodhar: As a follow up to this PR, I believe we need to make change in agent baker test files to reflect the latest mode ( not making that change won't break anything AFAIK but it's just to reflect the up-to-date file contents )

ashvindeodhar (changes requested):

cns/service/main.go
pjohnst5: > should we be replacing the existing v4overlay generator with this new one?

Yeah I think that would be best, I thought Matt wanted to keep it in case of roll back
@thatmattlong do you want me to keep this v4overlay value or remove it?

pjohnst5: Hey @thatmattlong from what I can see in AKS-RP, there is nowhere this feature (CNS generating cni conflist for overlay) is being used, is that correct?
In that case, I could remove the "v4overlay" value in this PR safely

cni/network/invoker_cns.go
```diff
@@ -135,7 +135,7 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 
 		ncgw := net.ParseIP(info.ncGatewayIPAddress)
 		if ncgw == nil {
-			if (invoker.ipamMode != util.V4Overlay) && (invoker.ipamMode != util.DualStackOverlay) {
+			if (invoker.ipamMode != util.V4Overlay) && (invoker.ipamMode != util.DualStackOverlay) && (invoker.ipamMode != util.Overlay) {
```
paulyufan2, line 138: we probably need to also take care of those files with consolidated flag too:

(1) network_windows.go under network folder:
Line 431:
if nwInfo.IPV6Mode == string(util.DualStackOverlay) => we assign this ipamMode to IPV6Mode to add ipv6 net routes to windows node

Only it's dualstack mode, then we add net routes to windows node; if it's v4 overlay, then we do not add these ipv6 net rules.

(2) network.go under cni/network folder:
Line 777:
// inject ipv6 routes to Linux pod
epInfo.IPV6Mode = string(util.IpamMode(opt.nwCfg.IPAM.Mode))
if IPAM mode becomes overlay, also need to check if it's dualstack mode to inject ipv6 routes to Linux pod.

pjohnst5: Ah yes, these two cases mean that there is a difference between v4overlay and dualstackoverlay ipam modes @ashvindeodhar 

So I think we need to keep them separate

Shall I abandon this draft or change "v4overlay" to "overlay" ? I don't think we can consolidate "v4overlay" and "dualstackoverlay" to just "overlay", there needs to be a separation

paulyufan2: yes, we probably can use "executionMode" in the conflist file to differentiate it if we use consolidated "overlay" flag, need design this

pjohnst5: I will check with HNS team and Neha if these routes (for windows) are needed also in v4 then follow up here

cns/service/main.go
pjohnst5: Nvm, aks DOES use the conflist generator feature of CNS, so we'd leave 'v4overlay' as an option

cni/network/invoker_cns.go
pjohnst5: Yes, HNS said we want the 1 ipv4 route for both dualstack and v4overlay, which we can do eventually but not in this PR

pjohnst5: Removing 1 redundant route in this PR:
https://github.com/Azure/azure-container-networking/pull/2002/files

paulyufan2, line 138: With my recent changes: https://github.com/Azure/azure-container-networking/pull/1989/files (this PR gets rid of IPv6Mode already, so we do not check ipam mode to see whether it's dualstack overlay or not). Instead we just check whether ipv6Result obj is nil or not.

Hence, we can get rid of util.DualStackOverlay flag, and even util.V4Overlay

cni/network/invoker_cns.go
```diff
@@ -199,7 +199,7 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 
 		// set subnet prefix for host vm
 		// setHostOptions will execute if IPAM mode is not v4 overlay and not dualStackOverlay mode
-		if (invoker.ipamMode != util.V4Overlay) && (invoker.ipamMode != util.DualStackOverlay) {
+		if (invoker.ipamMode != util.V4Overlay) && (invoker.ipamMode != util.DualStackOverlay) && (invoker.ipamMode != util.Overlay) {
```
paulyufan2, line 202: same here

cni/network/invoker_cns.go
pjohnst5: This (according to previous logic) should run when in v4overlay mode

If we switch it to say "if ipv6Result is nil or not", then we change the logic, and it won't run for v4overlay mode

Maybe I'm missing something, but I believe we don't want to change it to be that logic, is that right?

pjohnst5: Okay, I believe those checks you mentioned (inject ipv6 routes to Linux, or netsh rules to windows), happen elsewhere, in `cni/network/network.go` and `network/network_windows.go` respectively

But here, we don't want to change the logic, I thought

Is that right @ashvindeodhar ?

pjohnst5: Discussed offline, this is correct for these two if statements, we will add a third overlay option, change it in AKS-RP to just 'overlay', then come back and remove 'v4overlay' and 'dualstackoverlay' options

cni/network/invoker_cns.go
pjohnst5: See above

Makefile
```diff
@@ -66,6 +66,7 @@ CNI_SWIFT_BUILD_DIR = $(BUILD_DIR)/cni-swift
 CNI_OVERLAY_BUILD_DIR = $(BUILD_DIR)/cni-overlay
 CNI_BAREMETAL_BUILD_DIR = $(BUILD_DIR)/cni-baremetal
 CNI_DUALSTACK_BUILD_DIR = $(BUILD_DIR)/cni-dualstack
+CNI_OVERLAY_CONSOLIDATED_BUILD_DIR = $(BUILD_DIR)/cni-overlay-consolidated
```
rbtr, line 69: I am opposed to adding a new build artifact like this, as we will then need to migrate _off_ of it in the future and back to the bare "overlay" artifact. 
Can we not simply consolidate on to "cni-overlay" immediately?

pjohnst5: I think that is a good point, one concern was backwards compatibility, to still have 'v4overlay' and 'dualstack' be valid options until we know 'overlay' works properly in AKS

So being able to roll back and still support 'dualstack' and 'v4overlay' was the major concern

@ashvindeodhar wdyt?

pjohnst5: @thatmattlong what do you think as well?

pjohnst5: @rbtr what do you think, about having that overlap period in AKS (where all three options are viable)?

It maybe simpler to consolidate immediately, we would just need to keep these changes right in sync with AKS-RP changes and a new CNS version, might miss something

rbtr: I was specifically referring to the tgz build artifacts, but looking further I think maybe too many things are changing in this single PR (or they should be reordered).
Should the changes actually be:
- release CNI that supports the consolidated `"executionMode":"swift"` and ipam mode `overlay`
  - this should be end-to-end and should collapse v4/dualstack overlay in to the single `cni-overlay` build artifact
- release dropgz with ^
- release CNS that supports the consolidated `overlay` conflist generation
- migrate AKS to consolidate `overlay` config argument everywhere
- deprecate and remove v4overlay/dualstack overlay everywhere

vs the updates you have proposed, this way we should never need to update anything to an intermediate "overlay-consolidated" that we later need to change back.

pjohnst5: Oh I didn't know about `"executionMode": "swift"` conslidation, is there someone working on that currently?

> release dropgz with ^

After releasing dropgz, would we update AKS-RP anywhere?

> release CNS that supports the consolidated overlay conflist generation

Or do we wait till after releasing CNS to make any AKS-RP changes?

pjohnst5: We update azure-cni binary and conflist at once, by virtue of tarball
	> Check this new binary and conflist don't break V4
        > Maybe add pipeline step for overlay cni
       > what aks checks are there?
       > Still would release dualstack version for backcompat

Relesae dropgz with ^
Rollout this everywhere, ~4 weeks (CNI is everywhere, and supports 'overlay' mode)
       Can make this faster by cutting new Agent baker image for dropgz right after tarball is release in our repo
Add a third 'overlay' generator, until cns image is rolled out (4 weeks)
Change the configmap, to use 'overlay'
Remove 'dualstack' and 'v4overlay' in CNS and CNI

Unknowns:
executionMode v4swift, what is it? Do we need it in v4overlay?

pjohnst5:
We're going to follow this strategy now to update to overlay
https://github.com/Azure/azure-container-networking/pull/2008

PR #1979, opened 2023-05-24
Always use 0 for NC version in Overlay

rbtr:
**Reason for Change**:

If DNC fails to program an Overlay NC on the first attempt, the version may get incremented. However the programmed version from the Host will always show up as `0`. This temp fix ignores the NC version from the NNC so that the programmed Host version of 0 correctly moves the IPs from the NNC to "Available" and allows them to be assigned to Pods.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -92,7 +94,7 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 	for addr := primaryPrefix.Masked().Addr(); primaryPrefix.Contains(addr); addr = addr.Next() {
 		secondaryIPConfigs[addr.String()] = cns.SecondaryIPConfig{
 			IPAddress: addr.String(),
-			NCVersion: int(nc.Version),
+			NCVersion: 0,
```
neaggarwMS, line 97: There is also a Version we stick in NC goal state and not on the IPConfig.

neaggarwMS: Also this can impact VNETPrefix that will also use same mode but we need version check there

neaggarwMS (changes requested):

cns/kubecontroller/nodenetworkconfig/conversion.go
neaggarwMS: return &cns.CreateNetworkContainerRequest{
		NetworkContainerid:   nc.ID,
		**Version:              strconv.FormatInt(nc.Version, 10), //nolint:gomnd // it's decimal**

nairashu:
Why is the host sending a version of 0 always? Shouldn't DNC be publishing the updated version of the NC to NMAgent via the PubSub.

nairashu (approved):

cns/kubecontroller/nodenetworkconfig/conversion.go
rbtr: updated, setting the nc.Version = 0 and leaving everything else as is so that I don't have to track down every usage of it

nairashu:
> Why is the host sending a version of 0 always? Shouldn't DNC be publishing the updated version of the NC to NMAgent via the PubSub.

Discussed in offline chat that for Static allocation of Vnet Prefix this should adhere to same behavior as Overlay and for Dynamic behavior we will have to rework Vnet Prefix code if and when we introduce scaling of CIDR blocks.

neaggarwMS (approved):

thatmattlong (approved):

neaggarwMS (approved):

neaggarwMS (approved):

PR #1985, opened 2023-05-30
Check Ubuntu version to perform correct DNS resolution

paulyufan2:
**Reason for Change**:

This issue is to fix this ICM: 390860988

**Issue Fixed**:

The root cause is partner team upgrades Linux ubuntu version from 18 to 22; so DNS-related commands on Ubuntu are changed. So to make this change to get compatible with different ubuntu versions

### DIFF
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -256,10 +256,10 @@ func readDNSInfo(ifName string) (DNSInfo, error) {
 
 	out, err = p.ExecuteCommand(resolveCtlCmd)
 	if err != nil {
-		log.Errorf("resolvectl cmd failed with:%+v",err)
+		log.Errorf("resolvectl cmd failed with:%v",err)
 		out, err = p.ExecuteCommand(systemdResolveCmd)
 		if err != nil {
-			return dnsInfo, fmt.Errorf("failed to execute command %s", systemdResolveCmd)
+			return dnsInfo, errors.New("failed to execute systemdResolveCmd")
 		}
 	}
 
@@ -365,7 +365,7 @@ func applyDNSConfig(extIf *externalInterface, ifName string) error {
 			systemdResolveCmd := fmt.Sprintf("systemd-resolve --interface=%s%s", ifName, "--set-dns="+strings.TrimSpace(setDnsList))
 
 			_, err = p.ExecuteCommand(resolveCtlCmd)
-			log.Errorf("resolvectl cmd failed with:%+v",err)
+			log.Errorf("resolvectl cmd failed with:%v",err)
 			if err != nil {
 				_, err = p.ExecuteCommand(systemdResolveCmd)
 				return err
@@ -379,9 +379,8 @@ func applyDNSConfig(extIf *externalInterface, ifName string) error {
 			systemdResolveCmd := fmt.Sprintf("systemd-resolve --interface=%s --set-domain=%s", ifName, extIf.DNSInfo.Suffix)
 
 			if _, err = p.ExecuteCommand(resolveCtlCmd); err != nil {
-				log.Errorf("resolvectl cmd failed with:%+v",err)
+				log.Errorf("resolvectl cmd failed with:%v",err)
 				_, err = p.ExecuteCommand(systemdResolveCmd)
-				return err
 			}
 		}
 	}
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -31,7 +31,6 @@ const (
 	dnsServersStr     = "DNS Servers"
 	dnsDomainStr      = "DNS Domain"
 	ubuntuVersion17   = 17
-	ubuntuVersion22   = 22
 	// OptVethName key for veth name option
 	OptVethName = "vethname"
 	// SnatBridgeIPKey key for the SNAT bridge
@@ -216,12 +215,11 @@ func getMajorVersion(version string) (int, error) {
 	return 0, fmt.Errorf("[net] Error getting major version")
 }
 
-// getUbuntuVersion() gets Linux ubuntu version; it will return 0 if it's failed
-func getUbuntuVersion() int {
+func isGreaterOrEqaulUbuntuVersion(versionToMatch int) bool {
 	osInfo, err := platform.GetOSDetails()
 	if err != nil {
 		log.Printf("[net] Unable to get OS Details: %v", err)
-		return 0
+		return false
 	}
 
 	log.Printf("[net] OSInfo: %+v", osInfo)
@@ -234,36 +232,35 @@ func getUbuntuVersion() int {
 		retrieved_version, err := getMajorVersion(version)
 		if err != nil {
 			log.Printf("[net] Not setting dns. Unable to retrieve major version: %v", err)
-			return 0
+			return false
 		}
 
-		return retrieved_version
+		if retrieved_version >= versionToMatch {
+			return true
+		}
 	}
 
-	return 0
+	return false
 }
 
-func isGreaterOrEqualUbuntuVersion(versionToMatch int) bool {
-	ubuntuVerion := getUbuntuVersion()
-	return ubuntuVerion >= versionToMatch
-}
-
-func readDNSInfo(ifName string, osVersion int) (DNSInfo, error) {
+func readDNSInfo(ifName string) (DNSInfo, error) {
 	var (
 		dnsInfo DNSInfo
-		cmd     string
+		out     string
+		err     error
 	)
 
 	p := platform.NewExecClient()
-	if osVersion >= ubuntuVersion22 {
-		cmd = fmt.Sprintf("resolvectl status %s", ifName)
-	} else {
-		cmd = fmt.Sprintf("systemd-resolve --status %s", ifName)
-	}
+	resCtlCmd := fmt.Sprintf("resolvectl status %s", ifName)
+	sysResCtlCmd := fmt.Sprintf("systemd-resolve --status %s", ifName)
 
-	out, err := p.ExecuteCommand(cmd)
+	out, err = p.ExecuteCommand(resCtlCmd)
 	if err != nil {
-		return dnsInfo, err
+		log.Printf("[net] trying command %s", sysResCtlCmd)
+		out, err = p.ExecuteCommand(sysResCtlCmd)
+		if err != nil {
+			return dnsInfo, err
+		}
 	}
 
 	log.Printf("[net] console output for above cmd: %s", out)
@@ -302,8 +299,8 @@ func readDNSInfo(ifName string, osVersion int) (DNSInfo, error) {
 	return dnsInfo, nil
 }
 
-func saveDNSConfig(extIf *externalInterface, osVersion int) error {
-	dnsInfo, err := readDNSInfo(extIf.Name, osVersion)
+func saveDNSConfig(extIf *externalInterface) error {
+	dnsInfo, err := readDNSInfo(extIf.Name)
 	if err != nil || len(dnsInfo.Servers) == 0 || dnsInfo.Suffix == "" {
 		log.Printf("[net] Failed to read dns info %+v from interface %v: %v", dnsInfo, extIf.Name, err)
 		return err
@@ -344,12 +341,10 @@ func (nm *networkManager) applyIPConfig(extIf *externalInterface, targetIf *net.
 	return nil
 }
 
-func applyDNSConfig(extIf *externalInterface, ifName string, ubuntuVersion int) error {
+func applyDNSConfig(extIf *externalInterface, ifName string) error {
 	var (
 		setDnsList string
 		err        error
-		buf        string
-		cmd        string
 	)
 	p := platform.NewExecClient()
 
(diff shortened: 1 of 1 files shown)


network/network_linux.go
```diff
@@ -215,11 +216,13 @@ func getMajorVersion(version string) (int, error) {
 	return 0, fmt.Errorf("[net] Error getting major version")
 }
 
-func isGreaterOrEqaulUbuntuVersion(versionToMatch int) bool {
+// checkUbuntuVersion() checks ubuntu version and whether it's equal or greater than versionToMatch
+// return ubuntu version if it's equal or greater than versionToMatch; otherwise return 0
+func checkUbuntuVersion(versionToMatch int) (int, bool) {
```
tamilmani1989, line 221: can we split this to api to 2? one to get ubuntu version and other to compare version

tamilmani1989 (changes requested):

tamilmani1989:
@paulyufan2 can you add PR description

network/network_linux.go
```diff
@@ -232,22 +234,37 @@ func isGreaterOrEqaulUbuntuVersion(versionToMatch int) bool {
...
+		cmd     string
+	)
 
 	p := platform.NewExecClient()
-	cmd := fmt.Sprintf("systemd-resolve --status %s", ifName)
+	if osVersion >= ubuntuVersion22 {
+		cmd = fmt.Sprintf("resolvectl status %s", ifName)
+	} else {
+		cmd = fmt.Sprintf("systemd-resolve --status %s", ifName)
+	}
```
rbtr, line 266: to me it seems like the more reliable and generic way of accomplishing this would be to check for the presence/status of `resolvectl` and `systemd-resolved` and use the appropriate one based on that result.

rbtr: although if I'm reading the [ubuntu manpage](https://manpages.ubuntu.com/manpages/impish/man1/resolvectl.1.html) correctly, it seems like invoking `resolvectl` via `systemd-resolve` should be backwards compatible and I'm wondering why we even need this?

paulyufan2: Hi Evan, I mean "resolvectl" command is not backwards compatible with Ubuntu18 and I tried it on one Ubuntu18.04 machine:
Image
And "systemd-resolve" is not able to be used on Ubuntu22

tamilmani1989:
> To be clear, I consider this blocking:
> 
> > to me it seems like the more reliable and generic way of accomplishing this would be to check for the presence/status of resolvectl or systemd-resolved and use the appropriate one based on that result.
> 
> We shouldn't check the Ubuntu version and make assumptions based on that. We are _assuming_ `systemd-resolve` or `resolvectl` is available depending on that version, but that may become incorrect and doesn't scale (say we need to support a different Linux distro).
> 
> We should _directly check_ for the dependencies we need, and do the appropriate thing based on what we detect is available.

I like this idea. @paulyufan2  for ubuntu versions >=1.18, lets check if systemd-resolve works and if not try resolvectl

rbtr:
> ubuntu versions >=1.18, lets check if systemd-resolve works and if not try resolvectl

this doesn't need to be tied to the ubuntu version at all. 
check for resolvectl and if it exists use it. if no resolvectl, check for systemd-resolve and if it exists use it. if no systemd-resolve, fail out.

network/network_linux.go
```diff
@@ -243,14 +243,24 @@ func isGreaterOrEqaulUbuntuVersion(versionToMatch int) bool {
...
+	var (
+		dnsInfo DNSInfo
+		out     string
+		err     error
+	)
 
 	p := platform.NewExecClient()
-	cmd := fmt.Sprintf("systemd-resolve --status %s", ifName)
-	out, err := p.ExecuteCommand(cmd)
+	resCtlCmd := fmt.Sprintf("resolvectl status %s", ifName)
```
tamilmani1989, line 254: Lets name it as `resolveCtlCmd` and `systemdResolveCmd`

network/network_linux.go
```diff
@@ -243,14 +243,24 @@ func isGreaterOrEqaulUbuntuVersion(versionToMatch int) bool {
...
+	sysResCtlCmd := fmt.Sprintf("systemd-resolve --status %s", ifName)
+
+	out, err = p.ExecuteCommand(resCtlCmd)
 	if err != nil {
-		return dnsInfo, err
+		log.Printf("[net] trying command %s", sysResCtlCmd)
```
tamilmani1989, line 259: no need for this log.. `ExecuteCommand` function anyway logs it.. instead can we log error returned by resovleCtlcmd?

network/network_linux.go
```diff
@@ -345,23 +355,32 @@ func applyDnsConfig(extIf *externalInterface, ifName string) error {
...
 		if extIf.DNSInfo.Suffix != "" {
-			cmd := fmt.Sprintf("systemd-resolve --interface=%s --set-domain=%s", ifName, extIf.DNSInfo.Suffix)
-			_, err = p.ExecuteCommand(cmd)
-		}
+			// example command on Ubuntu22: resolvectl domain azure0 dlw5dhyl2njevcuzgmfubi2oid.bx.internal.cloudapp.net
+			resCtlCmd := fmt.Sprintf("resolvectl domain %s %s", ifName, extIf.DNSInfo.Suffix)
+			// example command on Ubuntu18: systemd-resolve --interface=azure0 --set-domain=dlw5dhyl2njevcuzgmfubi2oid.bx.internal.cloudapp.net
+			sysResCtlCmd := fmt.Sprintf("systemd-resolve --interface=%s --set-domain=%s", ifName, extIf.DNSInfo.Suffix)
 
+			if _, err = p.ExecuteCommand(sysResCtlCmd); err != nil {
```
tamilmani1989, line 380: lets do in same order as above.. first try resolveCtl and systemdCtl.. can we log error returned by resolveCtl cmd?

tamilmani1989 (changes requested):

network/network_linux.go
```diff
@@ -243,14 +243,24 @@ func isGreaterOrEqaulUbuntuVersion(versionToMatch int) bool {
...
 	p := platform.NewExecClient()
-	cmd := fmt.Sprintf("systemd-resolve --status %s", ifName)
-	out, err := p.ExecuteCommand(cmd)
+	resolveCtlCmd := fmt.Sprintf("resolvectl status %s", ifName)
+	systemdResolveCmd := fmt.Sprintf("systemd-resolve --status %s", ifName)
+
+	out, err = p.ExecuteCommand(resolveCtlCmd)
 	if err != nil {
-		return dnsInfo, err
+		log.Errorf("%+v",err)
```
tamilmani1989, line 259: can you log "resolvectl cmd failed with:%+v"

network/network_linux.go
```diff
@@ -345,23 +355,35 @@ func applyDnsConfig(extIf *externalInterface, ifName string) error {
...
 		if setDnsList != "" {
-			cmd := fmt.Sprintf("systemd-resolve --interface=%s%s", ifName, setDnsList)
-			_, err = p.ExecuteCommand(cmd)
+			// example command on Ubuntu22: resolvectl dns azure0 168.63.129.16
+			resolveCtlCmd := fmt.Sprintf("resolvectl dns %s%s", ifName, setDnsList)
+			// example command on Ubuntu18: systemd-resolve --interface=azure0 --set-dns=168.63.129.16
+			systemdResolveCmd := fmt.Sprintf("systemd-resolve --interface=%s%s", ifName, "--set-dns="+strings.TrimSpace(setDnsList))
+
+			_, err = p.ExecuteCommand(resolveCtlCmd)
+			log.Errorf("%+v",err)
```
tamilmani1989, line 368: log only if error happens and log as "resolveCtl cmd failed with"

network/network_linux.go
```diff
@@ -345,23 +355,35 @@ func applyDnsConfig(extIf *externalInterface, ifName string) error {
...
 		if extIf.DNSInfo.Suffix != "" {
-			cmd := fmt.Sprintf("systemd-resolve --interface=%s --set-domain=%s", ifName, extIf.DNSInfo.Suffix)
-			_, err = p.ExecuteCommand(cmd)
+			// example command on Ubuntu22: resolvectl domain azure0 dlw5dhyl2njevcuzgmfubi2oid.bx.internal.cloudapp.net
+			resolveCtlCmd := fmt.Sprintf("resolvectl domain %s %s", ifName, extIf.DNSInfo.Suffix)
+			// example command on Ubuntu18: systemd-resolve --interface=azure0 --set-domain=dlw5dhyl2njevcuzgmfubi2oid.bx.internal.cloudapp.net
+			systemdResolveCmd := fmt.Sprintf("systemd-resolve --interface=%s --set-domain=%s", ifName, extIf.DNSInfo.Suffix)
+
+			if _, err = p.ExecuteCommand(resolveCtlCmd); err != nil {
+				log.Errorf("%+v",err)
```
tamilmani1989, line 382: same as above

tamilmani1989 (approved):

network/network_linux.go
```diff
@@ -243,14 +243,24 @@ func isGreaterOrEqaulUbuntuVersion(versionToMatch int) bool {
...
+		err     error
+	)
 
 	p := platform.NewExecClient()
-	cmd := fmt.Sprintf("systemd-resolve --status %s", ifName)
-	out, err := p.ExecuteCommand(cmd)
+	resolveCtlCmd := fmt.Sprintf("resolvectl status %s", ifName)
+	systemdResolveCmd := fmt.Sprintf("systemd-resolve --status %s", ifName)
+
+	out, err = p.ExecuteCommand(resolveCtlCmd)
```
rbtr, line 257:
```suggestion
	out, err := p.ExecuteCommand(resolveCtlCmd)
```

network/network_linux.go
```diff
@@ -243,14 +243,24 @@ func isGreaterOrEqaulUbuntuVersion(versionToMatch int) bool {
...
 }
 
-func readDnsInfo(ifName string) (DNSInfo, error) {
-	var dnsInfo DNSInfo
+func readDNSInfo(ifName string) (DNSInfo, error) {
+	var (
+		dnsInfo DNSInfo
+		out     string
+		err     error
+	)
```
rbtr, line 251: predeclaring variables (unnecessarily) is an antipattern
```suggestion
	var dnsInfo DNSInfo
```

network/network_linux.go
```diff
@@ -345,23 +355,34 @@ func applyDnsConfig(extIf *externalInterface, ifName string) error {
...
-			_, err = p.ExecuteCommand(cmd)
+			// example command on Ubuntu22: resolvectl domain azure0 dlw5dhyl2njevcuzgmfubi2oid.bx.internal.cloudapp.net
+			resolveCtlCmd := fmt.Sprintf("resolvectl domain %s %s", ifName, extIf.DNSInfo.Suffix)
+			// example command on Ubuntu18: systemd-resolve --interface=azure0 --set-domain=dlw5dhyl2njevcuzgmfubi2oid.bx.internal.cloudapp.net
+			systemdResolveCmd := fmt.Sprintf("systemd-resolve --interface=%s --set-domain=%s", ifName, extIf.DNSInfo.Suffix)
+
+			if _, err = p.ExecuteCommand(resolveCtlCmd); err != nil {
+				log.Errorf("resolvectl cmd failed with:%v",err) //nolint
+				_, err = p.ExecuteCommand(systemdResolveCmd)
+			}
```
rbtr, line 384: i doubt this is safe in scenarios where both systemd-resolve and resolvectl are available, and the first one has a legitimate error

rbtr (changes requested):
this is good in that it no longer depends on OS version, but the test for which command to use is missing, and that makes the implementation unnecessarily complex.

the first thing to do is _check which commands are available_.
`$ which resolvectl` errors if `resolvectl` is not available, and then `$ which systemd-resolve` errors if `systemd-resolve` is not available.

then, we KNOW which command we should use. we don't need to try `resolvectl` again, if we have already determined that it is not present. similarly, if we have determined that we _do_ have `resolvectl`, we should not try to fall through to `systemd-resolved`.

IMO this should look like this:

```go
domainResolutionCmd := fmt.Sprintf("resolvectl domain %s %s", ifName, extIf.DNSInfo.Suffix)

if _, err := p.ExecuteCommand("which resolvectl") {
    domainResolutionCmd = fmt.Sprintf("systemd-resolve --interface=%s --set-domain=%s", ifName, extIf.DNSInfo.Suffix)
}

if extIf.DNSInfo.Suffix != "" {
    p.ExecuteCommand(domainResolutionCmd)
}
```

we test which command is available, and that immediately tells us which command we are going to execute later. by working backwards, our later code does not need any `if`s because we have inlined all the conditional logic _in_ the check for which command is available.

network/network_linux.go
```diff
@@ -345,23 +353,35 @@ func applyDnsConfig(extIf *externalInterface, ifName string) error {
 				continue
 			}
 
-			buf := fmt.Sprintf("--set-dns=%s", server)
-			setDnsList = setDnsList + " " + buf
+			setDnsList = setDnsList + " " + server
```
rbtr, line 356: is it expected for `extIf.DNSInfo.Servers` to have >1 item? I think systemd-resolve will fail the way the commands are currently written with this "setDnsList".

it seems like `resolvectl` can handle this:
```bash
resolvectl help
[...]
  dns [LINK [SERVER...]]       Get/set per-interface DNS server address
```
`$ resolvectl dns azure0 192.168.1.1 192.168.1.2` should work.


but it doesn't look like `systemd-resolve` can, and it only expects a single address when invoked like this:
```bash
systemd-resolve help
[...]
     --set-dns=SERVER       Set per-interface DNS server address
```
`$ systemd-resolve --interface=azure0 --set-dns=192.168.1.1 192.168.1.2` definitely will not work.

cc @tamilmani1989 can you clarify above?

paulyufan2:
Li agrees to move to transparent mode

rayburgemeestre:
Hi @paulyufan2

May I ask what that means? And why this was closed? :sweat_smile: 
Is there another way to make Ubuntu 22.04 work that doesn't involve moving from `systemd-resolve` to `resolvectl`?

rbtr:
@rayburgemeestre if this is a problem for you, can you open an issue with some info?

Issue #1986, opened 2023-05-31
CNI not setting the Flags in endpoint policies to use IPV6 in case of DualStack cluster


princepereira:
**What happened:** 

In a DualStack Cluster, the HostPortMapping feature is not working only for IPV6.

Root cause is, CNI is not setting the flags in following code to use NatFlagsIPv6 while creating endpoint policy.

```      
  rawPolicy, _ := json.Marshal(&hnsv2.PortMappingPolicySetting{
			ExternalPort: uint16(mapping.HostPort),
			InternalPort: uint16(mapping.ContainerPort),
			VIP:          mapping.HostIp,
			Protocol:     protocol,
			**Flags:        hnsv2.NatFlagsLocalRoutedVip,**
		})
```

Without the Ipv6 enum being set, Hns considers it as not an IPV6 portmapping and creates Hairpin rules only for IPV4.

**What you expected to happen:**  

Once the IPV6 flag is set for endpoint policy creation, HNS will use the flag to plumb Hairpin policy for IPV6 as well.
This will enable the HostPortMapping feature for IPV6 as well in dualstack.

**How to reproduce it:**  

- Create 2 node cluster
- Create a pod with following yaml
```
apiVersion: v1
kind: Pod
metadata:
  name: tcpserver-hostport
  namespace: demo
spec:
  containers:
    - name: tcpserver-hostport
      image: princepereira/tcp-client-server:WS2022
      command: ["server.exe", "-p", "4444", "-pr", "All", "-swt", "5", "-awt", "5"]
      imagePullPolicy: Always
      ports:
        - containerPort: 4444
          hostPort: 4488
  terminationGracePeriodSeconds: 1000
  nodeSelector:
        kubernetes.io/os: windows
```
- Connect to the pod from the neighbouring node with HostPort and IPV6, IPv4 node Ips
- Connectivity will work only for IPV4

**Orchestrator and Version (e.g. Kubernetes, Docker):**  
Kubernetes : 1.26.3
Containerd

**Operating System (Linux/Windows):**  
Windows

**Kernel (e.g. `uanme -a` for Linux or `$(Get-ItemProperty -Path "C:\windows\system32\hal.dll").VersionInfo.FileVersion` for Windows):**
10.0.20348.558 (WinBuild.160101.0800)

**Anything else we need to know?:**
[Miscellaneous information that will assist in solving the issue.]

tamilmani1989:
yes the fix is merged. closing this

PR #1989, opened 2023-05-31
add v6 port mapping policy to dualstack overlay

paulyufan2:
**Reason for Change**:

This PR is to fix this issue: https://github.com/Azure/azure-container-networking/issues/1986

**Issue Fixed**:

Fix has been verified by HNS team, but looking for another issue(incorrect windows CNI conflist file) to be resolved.

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -40,6 +40,7 @@ const (
 	// Supported IP version. Currently support only IPv4
 	ipamV6                = "azure-vnet-ipamv6"
 	defaultRequestTimeout = 15 * time.Second
+	dualStackOverlay      = "dualStackOverlay"
 )
 
 // CNI Operation Types
@@ -637,13 +638,11 @@ func (plugin *NetPlugin) createNetworkInternal(
 		NetNs:                         ipamAddConfig.args.Netns,
 		Options:                       ipamAddConfig.options,
 		DisableHairpinOnHostInterface: ipamAddConfig.nwCfg.DisableHairpinOnHostInterface,
-		IPV6Mode:                      ipamAddConfig.nwCfg.IPV6Mode,
+		IPV6Mode:                      ipamAddConfig.nwCfg.IPV6Mode, //TODO: check if this can be removed.
 		IPAMType:                      ipamAddConfig.nwCfg.IPAM.Type,
 		ServiceCidrs:                  ipamAddConfig.nwCfg.ServiceCidrs,
+		IsIPv6Enabled:                 ipamAddResult.ipv6Result != nil,
 	}
-
-	// set IPv6Mode to dualStackOverlay mode
-	nwInfo.IPV6Mode = ipamAddConfig.nwCfg.IPAM.Mode
 
 	if err = addSubnetToNetworkInfo(ipamAddResult, &nwInfo); err != nil {
 		log.Printf("[cni-net] Failed to add subnets to networkInfo due to %+v", err)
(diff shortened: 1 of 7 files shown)


cni/network/network_windows.go
```diff
@@ -267,13 +268,39 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig) []policy.Policy {
...
 			Type: policy.EndpointPolicy,
 			Data: hnsv2Policy,
 		}
-		log.Printf("[net] Creating port mapping policy: %+v", policy)
 
-		policies = append(policies, policy)
+		log.Printf("[net] Creating port mapping policy: %+v", policyv4)
+		policies = append(policies, policyv4)
+
+		// add port mapping policy for v6 if it's dualstack overlay mode
```
ashvindeodhar, line 279: I would suggest drop "if it's dualstack overlay mode". We want to set this for ipv6

cni/network/network.go
```diff
@@ -637,14 +638,12 @@ func (plugin *NetPlugin) createNetworkInternal(
 		NetNs:                         ipamAddConfig.args.Netns,
 		Options:                       ipamAddConfig.options,
 		DisableHairpinOnHostInterface: ipamAddConfig.nwCfg.DisableHairpinOnHostInterface,
-		IPV6Mode:                      ipamAddConfig.nwCfg.IPV6Mode,
+		IPV6Mode:                      ipamAddConfig.nwCfg.IPV6Mode, //TODO: check if this can be removed.
```
ashvindeodhar, line 641: nit: can you reword this to 'deprecated' pls? It would provide better context on if we can get rid of the whole mechanism around usage of IPV6Mode

cni/network/network.go
```diff
@@ -40,6 +40,7 @@ const (
 	// Supported IP version. Currently support only IPv4
 	ipamV6                = "azure-vnet-ipamv6"
 	defaultRequestTimeout = 15 * time.Second
+	dualStackOverlay      = "dualStackOverlay"
```
ashvindeodhar, line 43: why do we need this? IMO we don't need this

network/network_windows.go
```diff
@@ -427,8 +426,8 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *NetworkInfo, extIf *extern
 	if err != nil {
 		// if network not found, create the HNS network.
 		if errors.As(err, &hcn.NetworkNotFoundError{}) {
-			// in dualStackOverlay mode, add net routes to windows node
-			if nwInfo.IPV6Mode == string(util.DualStackOverlay) {
+			// add net routes to windows node in dualStackOverlay mode
```
ashvindeodhar, line 429: nit: update the comment. we don't want to tie this to dualstackoverlay mode. We set these v6 routes if we have v6 enabled.

cni/network/network.go
```diff
@@ -774,7 +774,7 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
 
 	if opt.resultV6 != nil {
 		// inject ipv6 routes to Linux pod
-		epInfo.IPV6Mode = string(util.IpamMode(opt.nwCfg.IPAM.Mode))
+		epInfo.IPV6Mode = dualStackOverlay // TODO: can this IPV6Mode be deprecated and can we add IsIPv6Enabled flag for generic working
```
ashvindeodhar, line 777: why do we need to make this change?

ashvindeodhar (changes requested):

ashvindeodhar (approved):

PR #1994, opened 2023-06-02
Initial getHomeAZ 404 changes

smittal22:
**Reason for Change**: If NMA returns 404 for GetHomeAZ API, we should mark CNS' response as `isSupported: false` in this case.
This will only happen for regions that do not support AZs, hence, will not have any AZ info. NMA in this case will return 404.

**Issue Fixed**:
Fixes https://dev.azure.com/msazure/One/_workitems/edit/23883668

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/cns/restserver/homeazmonitor.go
+++ b/cns/restserver/homeazmonitor.go
@@ -126,6 +126,12 @@ func (h *HomeAzMonitor) Populate(ctx context.Context) {
 				h.update(returnCode, returnMessage, cns.HomeAzResponse{IsSupported: true})
 				return
 
+			case http.StatusNotFound:
+				returnMessage := fmt.Sprintf("[HomeAzMonitor] region does not support AZs probably, failed with StatusCode: %d, error: %v", apiError.StatusCode(), err)
+				returnCode := types.NotFound
+				h.update(returnCode, returnMessage, cns.HomeAzResponse{IsSupported: false})
+				return
+
 			default:
 				returnMessage := fmt.Sprintf("[HomeAzMonitor] failed with StatusCode: %d", apiError.StatusCode())
 				returnCode := types.UnexpectedError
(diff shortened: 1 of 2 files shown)


cns/restserver/homeazmonitor.go
```diff
@@ -126,6 +126,12 @@ func (h *HomeAzMonitor) Populate(ctx context.Context) {
 				h.update(returnCode, returnMessage, cns.HomeAzResponse{IsSupported: true})
 				return
 
+			case http.StatusNotFound:
+				returnMessage := fmt.Sprintf("[HomeAzMonitor] region does not support AZs probably, failed with StatusCode: %d, error: %v", apiError.StatusCode(), err)
```
ashvindeodhar, line 130: nit: remove 'probably'
instead of saying this 'failed' ( Which is not the case ) IMO we should simply indicate that NMA returned apiError.StatusCode()

cns/restserver/homeazmonitor.go
```diff
@@ -126,6 +126,12 @@ func (h *HomeAzMonitor) Populate(ctx context.Context) {
...
+				returnCode := types.NotFound
```
ashvindeodhar, line 131: do we use this returnCode to classify this as 'error' and potentially 'retry'?
If yes, we don't want to do treat this as an error. We should still treat it as success where this is equivalent to NMA not supporting homeAZ API

nmagent/client_test.go
```diff
@@ -675,6 +675,15 @@ func TestGetHomeAz(t *testing.T) {
...
 			true,
 		},
+		{
+			"404 from NMA",
+			nmagent.AzResponse{},
+			"/machine/plugins?comp=nmagent&type=GetHomeAz%2Fapi-version%2F1",
+			map[string]interface{}{
+				"httpStatusCode": "404",
+			},
+			true,
```
ashvindeodhar, line 685: unrelatd to this PR but I don't believe 
if gotPath != test.expPath { on line 717 is adding any value. Why would NMA API path be different?

ashvindeodhar (changes requested):

cns/restserver/homeazmonitor.go
smittal22: Addressed

cns/restserver/homeazmonitor.go
smittal22: Addressed

nmagent/client_test.go
smittal22: Removed

ashvindeodhar (approved):

PR #1999, opened 2023-06-05
feat: [NPM] metric for total Pod IPs

huntergregory:
**Reason for Change**:
Windows has scale limits in terms of total Pod IPs in the cluster (on either Windows or Linux nodes).

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Notes**:
Also removes unnecessary metric for max ipset members. Total Pod IPs is a more accurate/relevant metric.

### Metric Preview
```
# HELP npm_pods_watched Number of Pods NPM tracks across the cluster including Linux and Windows nodes
# TYPE npm_pods_watched gauge
npm_pods_watched 22
```


npm/pkg/controlplane/controllers/v2/podController.go
```diff
@@ -369,6 +377,7 @@ func (c *PodController) syncAddedPod(podObj *corev1.Pod) error {
 	// Create npmPod and add it to the podMap
 	npmPodObj := common.NewNpmPod(podObj)
 	c.podMap[podKey] = npmPodObj
+	metrics.AddPod()
```
huntergregory, line 380: Goal: call `metrics.AddPod()` once per Pod/IP pair (similar for `metrics.RemovePod()`)


- If a Pod has an update to its IP, follow Pod DELETE then Pod CREATE.
- For Pod DELETE: `c.cleanupDeletePod()` is only called when `podKey` exists in `c.podMap`
- For Pod CREATE: `c.syncAddedPod()` is only called when `podKey` does not exist in `c.podMap`. Otherwise, another codepath is taken.
- Keys are only added/removed from `c.podMap` in `c.syncAddedPod()` and `c.cleanupDeletePod()`

vakalapa (approved):

PR #2000, opened 2023-06-06
Sync host nc version

rsagasthya:
**Reason for Change**:

There are cases when the NC Version reported by NMAgent is lower than what is the actual version. This means, that the NC has regressed.
When such a regression is spotted, the CNS can simply return a panic with an appropriate message. 
When CNS is restarted, the NC versions should sync. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -253,7 +253,7 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
 		}
 		if localNCVersion > nmaNCVersion {
 			// This means that NMAgent reported a regression in the NC Version.
-			panic(fmt.Errorf("NC version from NMA is decreasing: have %d, got %d", localNCVersion, nmaNCVersion))
+			panic(fmt.Sprintf("NC version from NMA is decreasing: have %d, got %d", localNCVersion, nmaNCVersion))
 		}
 		if channelMode == cns.CRD {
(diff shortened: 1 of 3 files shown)


cns/restserver/internalapi.go
```diff
@@ -253,7 +253,8 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
 		}
 		if localNCVersion > nmaNCVersion {
 			// This means that NMAgent reported a regression in the NC Version.
-			panic(fmt.Errorf("NC version from NMA is decreasing: have %d, got %d", localNCVersion, nmaNCVersion))
+			err := fmt.Sprintf("NC version from NMA is decreasing: have %d, got %d", localNCVersion, nmaNCVersion)
+			panic("NC Version Regression: " + err)
```
rbtr, line 257: this is not an error and shouldn't be called `err` since that's misleading. but since it's just a string, can we inline it?

```suggestion
			panic(fmt.Sprintf("NC version from NMA has decreased from previously recorded: have %d, got %d", localNCVersion, nmaNCVersion))
```

rsagasthya: I had initially added this inline. But golint complained. I will update the variable name.

rbtr (approved):
lgtm once the lint is addressed

PR #2002, opened 2023-06-06
Remove redundant routes in dualstackoverlay for windows nodes

pjohnst5:
Removing a redundant route added to Windows VMs for dualstackoverlay

Redundant route being removed:
`netsh interface ipv4 add route "<overlay-subnet-prefix>" "<interface-name>" "<overlay-subnet-gateway>"`

Example:
`netsh interface ipv4 add route "10.244.0.0/16" "vEthernet (Ethernet 2)" "10.244.0.1"`

This route is redundant, because prior to it being added, there is another route added with the **same matching-prefix**:
Route being added before redundant route:
`netsh interface ipv4 add route "10.244.0.0/16" "vEthernet (Ethernet 2)" "0.0.0.0"`

So the second route is rendered useless, and we don't need it

This was confirmed by @princepereira, so removing the redundant route (both for IPV4 and IPV6)

### DIFF
--- a/network/network_windows.go
+++ b/network/network_windows.go
@@ -227,7 +227,6 @@ func (nm *networkManager) addNewNetRules(nwInfo *NetworkInfo) error {
 		if subnet.Gateway == nil {
 			return fmt.Errorf("[net] failed to get subnet gateway") // nolint
 		}
-		gateway := subnet.Gateway.String()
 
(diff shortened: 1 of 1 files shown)


paulyufan2 (approved):
lgtm

pjohnst5:
@princepereira could you review as well please? Thanks! Maybe mention the tests you did with this version of CNI

princepereira (approved):
LGTM

PR #2006, opened 2023-06-08
Add wireserver and metadata connecitivity tests to cilium pipeline

tamilmani1989:
**Reason for Change**:

This PR adds test coverage for wireserver and metadata connectivity from pod. Pod should not have connectivity to wireserver and can access metadata server

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
```diff
@@ -140,7 +140,32 @@ steps:
       bash hack/scripts/validate_state.sh
     name: "validatePods"
     displayName: "Validate Pods"
-  
+
+  - script: |
```
vipul-21, line 144: Can we add this a script to the `hack` folder so that we don't have to change the script at multiple places. And call that script here.

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
```diff
@@ -140,7 +140,32 @@ steps:
...
+      echo "Run wireserver and metadata connectivity Tests"
+      cat test/integration/manifests/jobs/wget-job.yaml | sed 's/{address}/168.63.129.16/g; s/{path}/machine\/plugins?comp=nmagent\&type=getinterfaceinfov1/g'|kubectl apply -f -
+      kubectl wait --for=condition=complete --timeout=6s job wget-test
+      if [ $? -eq 0 ]; then
+        echo "wireserver connectivity test succeeded instead of failing as expected"
+        kubectl logs job/wget-test
+        kubectl get job wget-test -o yaml
+        exit 1
+      fi
```
vipul-21, line 153: Should we add an else condition to it incase it fails ?

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
```diff
@@ -140,7 +140,32 @@ steps:
...
+      kubectl delete job wget-test
+      sleep 1
+      cat test/integration/manifests/jobs/wget-job.yaml | sed 's/{address}/169.254.169.254/g; s/{path}/metadata\/instance?api-version=2021-02-01/g'| kubectl apply -f -
```
vipul-21, line 156: I feel the same step(script) is testing couple of cases. Should we add these two as separate cases. Easier to debug IMO. ?

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
```diff
@@ -140,7 +140,32 @@ steps:
       bash hack/scripts/validate_state.sh
     name: "validatePods"
     displayName: "Validate Pods"
-  
+
+  - script: |
+      echo "Run wireserver and metadata connectivity Tests"
+      cat test/integration/manifests/jobs/wget-job.yaml | sed 's/{address}/168.63.129.16/g; s/{path}/machine\/plugins?comp=nmagent\&type=getinterfaceinfov1/g'|kubectl apply -f -
```
vipul-21, line 146: I understand that how we are using the wget. But can you please explain `s/{address}/168.63.129.16/g; s/{path}/machine\/plugins?comp=nmagent\&type=getinterfaceinfov1/g`  what is the usecase address we are trying to hit and what is the response ?

vipul-21 (changes requested):

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
rbtr: +1 but put it in `build/scripts/` not `hack/`

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
rbtr: when the sed strings have `/` in them, you can use a different separator:
```suggestion
      cat test/integration/manifests/jobs/wget-job.yaml | sed 's,{address},168.63.129.16,g; s,{path},machine/plugins?comp=nmagent\&type=getinterfaceinfov1,g'|kubectl apply -f -
```
might be easier to read

rbtr (changes requested):
a `Job` will be automatically restarted on failure. Maybe we can use a bare `Pod` instead to better effect.

A bare pod also lets us inline the deployment and significantly simplify these scripts:

```bash
kubectl run wget -it --rm --image busybox --restart Never -- wget --timeout=3 --header=Metadata:true "http://169.254.169.254/machine/plugins?comp=nmagent&type=getinterfaceinfov1"
if [ $? -eq 0 ]; then
        // [...] pass/fail based on error
fi

kubectl run wget -it --rm --image busybox --restart Never -- wget --timeout=3 --header=Metadata:true "http://169.254.169.254/metadata/instance?api-version=2021-02-01"
if [ $? -eq 0 ]; then
        // [...] pass/fail based on error
fi
```

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
tamilmani1989: instead of build, imo it should be placed in test/network.. i will create one

rbtr (approved):

vipul-21 (approved):

PR #2007, opened 2023-06-08
test:[NPM] Adding service deployment for the scale test

vipul-21:
**Reason for Change**:
Adding the option to deploy service for each set of deployments for scale test. Each service will be mapped to the respective deployment(having certain number of pods per deployment)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


camrynl (approved):

test/scale/test-scale.sh
```diff
@@ -312,10 +318,25 @@ generateDeployments() {
     done
 }
 
+generateServices() {
+    local numServices=$1
+    local numDeployments=$2
+    local serviceKind=$3
+
+    for i in $(seq -f "%05g" 1 $numServices); do
```
huntergregory, line 326: to make `--num-services` optional, I think you'll need to set a default value of 0 somewhere

test/scale/test-scale.sh
```diff
@@ -406,6 +427,7 @@ set -x
 $KUBECTL $KUBECONFIG_ARG create ns scale-test
 $KUBECTL $KUBECONFIG_ARG apply -f generated/kwok-nodes/
 $KUBECTL $KUBECONFIG_ARG apply -f generated/deployments/real/
+$KUBECTL $KUBECONFIG_ARG apply -f generated/services/real/
```
huntergregory, line 430: We'll need to apply only if `numServices > 0`. Like here:
https://github.com/Azure/azure-container-networking/pull/1975/files/24467195c5ac9edd90793a3befd95ebe029098f4#diff-7f8f18cf7c38d320fefc80745a20c806a8b4f626e8fb27bf7d009a0daf96de4f

huntergregory (changes requested):
This is awesome. Minor comments

test/scale/test-scale.sh
huntergregory, line 1: Could we also modify these NPM calculations:
- increase number of NPM IPSets by numRealDeployments when numServices > 0
- increase number of IPSet members by numRealPods

test/scale/test-scale.sh
vipul-21: Cool, will add it.

test/scale/test-scale.sh
vipul-21: So for:
- numIPSetsAddedByNPM : I can add the `numRealDeployments` if  `numServices > 0`
- numIPSetMembersAddedByNPM: Add `numRealPods` if `numServices > 0` ? @huntergregory

huntergregory: ya exactly

huntergregory (dismissed):

huntergregory (approved):

PR #2008, opened 2023-06-08
Moving v4overlay and dualstackoverlay to just 'overlay' in CNI

pjohnst5:
Here's the general idea:
- We change the conflists to 'overlay' for both dualstack and v4overlay, as well as make CNI capable of handling this 'overlay' option
- We release this CNI and conflist change to AKS together (as always, since it's in a .tar file)
- Update AKS to use this new CNI (dropgz) version
- Roll that out to all of AKS
---- This is where all AKS clusters, have the same conflist now (basically) ----
    - The only clusters with different conflists, would be v4overlay linux using generator: true, but that would work the same as now

Final cleanup stage
- In CNS, add a third 'overlay' generator, which generates 'overlay' conflist (without v4swift execution mode)
- Release it in AKS
- Change CNS configmap in AKS to use 'overlay' conflist generation
- Remove 'dualstack' and 'v4overlay' options completely in CNS and CNI (generators, conflists, if statements, all of it)


cni/network/invoker_cns.go
```diff
@@ -135,7 +135,7 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 
 		ncgw := net.ParseIP(info.ncGatewayIPAddress)
 		if ncgw == nil {
-			if (invoker.ipamMode != util.V4Overlay) && (invoker.ipamMode != util.DualStackOverlay) {
+			if (invoker.ipamMode != util.V4Overlay) && (invoker.ipamMode != util.DualStackOverlay) && (invoker.ipamMode != util.Overlay) {
```
tamilmani1989, line 138: can you add a TODO to note V4Overlay and DualStackOverlay should be removed later after CNS and AKS-RP rollout

cni/util/const.go
```diff
@@ -15,4 +15,13 @@ type IpamMode string
 const (
 	V4Overlay        IpamMode = "v4overlay"
 	DualStackOverlay IpamMode = "dualStackOverlay"
+	Overlay          IpamMode = "overlay" // Nothing changes between 'v4overlay' and 'dualStackOverlay' mode, so consolidating to one
 )
+
+// Overlay consolidation plan
```
tamilmani1989, line 21: can you also final step is to remove these modes from CNI

pjohnst5:
- Will also add CNS generator in this PR (even thoug it won't be used yet in AKS, until CNI is updated)

cni/network/invoker_cns.go
pjohnst5: Yes

pjohnst5:
Tested this on a v4overlay cluster and pod to pod connectivity worked for both linux and windows
[overlay-notes-and-testing.zip](https://github.com/Azure/azure-container-networking/files/11709443/overlay-notes-and-testing.zip)

rbtr (commented):
lgtm but I will defer to @tamilmani1989 to approve

tamilmani1989 (approved):
lgtm

PR #2010, opened 2023-06-09
'Overlay' conflist generator for CNS

pjohnst5:
Add third 'overlay' option to CNS conflist generator
Will be used once #2008 is rolled out in AKS


cns/cniconflist/generator_linux.go
```diff
@@ -32,11 +32,10 @@ func (v *V4OverlayGenerator) Generate() error {
 			cni.NetworkConfig{
 				Type:              overlaycniType,
 				Mode:              cninet.OpModeTransparent,
-				ExecutionMode:     string(util.V4Swift),
```
pjohnst5, line 35: @rbtr Do we want to adjust how v4overlay generator works in this PR? Or leave it as is

Reason I ask, is there's a `testdata/fixtures/azure-linux-swift-overlay.conflist` file already, so I would have to make a new `testdata/fixtures/azure-linux-swift-overlay-consolidated.conflist` if we want a separate PR

rbtr: no, and instead of that move the existing testfile to -v4 and change which files the tests are using

pjohnst5: I am leaning towards leaving v4overlay generator untouched (adding another file) but would like to know what you think

pjohnst5: K

rbtr (approved):
lgtm

rbtr:
@pjohnst5 can you backport this change to 1.4? I know that 1.4 doesn't support dualstack, but it is still using the "v4overlay" key and I would like to be able to simplify the usage in AKS to only "overlay".

PR #2014, opened 2023-06-12
fix: Updating the vmsize for e2e cilium to avoid resource scarcity

vipul-21:
**Reason for Change**:

- Updating the vm size used for e2e test for cilium

**Issue Fixed**:

- Improves the pipeline failure due to unavailability of resources(memory ) in case of cilium e2e

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
+++ b/.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
@@ -32,7 +32,7 @@ steps:
         mkdir -p ~/.kube/
         echo "Create AKS Overlay cluster"
         make -C ./hack/swift azcfg AZCLI=az REGION=$(REGION_OVERLAY_CLUSTER_TEST)
-        make -C ./hack/swift overlay-no-kube-proxy-up AZCLI=az REGION=$(REGION_OVERLAY_CLUSTER_TEST) SUB=$(SUB_AZURE_NETWORK_AGENT_TEST) CLUSTER=${{ parameters.clusterName }}-$(make revision)
+        make -C ./hack/swift overlay-no-kube-proxy-up AZCLI=az REGION=$(REGION_OVERLAY_CLUSTER_TEST) SUB=$(SUB_AZURE_NETWORK_AGENT_TEST) CLUSTER=${{ parameters.clusterName }}-$(make revision) VM_SIZE=Standard_B2ms
         echo "Cluster successfully created"
     displayName: Create Overlay cluster
     condition: succeeded()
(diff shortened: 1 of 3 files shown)


.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
```diff
@@ -32,7 +32,7 @@ steps:
         mkdir -p ~/.kube/
         echo "Create AKS Overlay cluster"
         make -C ./hack/swift azcfg AZCLI=az REGION=$(REGION_OVERLAY_CLUSTER_TEST)
-        make -C ./hack/swift overlay-no-kube-proxy-up AZCLI=az REGION=$(REGION_OVERLAY_CLUSTER_TEST) SUB=$(SUB_AZURE_NETWORK_AGENT_TEST) CLUSTER=${{ parameters.clusterName }}-$(make revision)
+        make -C ./hack/swift overlay-no-kube-proxy-up AZCLI=az REGION=$(REGION_OVERLAY_CLUSTER_TEST) SUB=$(SUB_AZURE_NETWORK_AGENT_TEST) CLUSTER=${{ parameters.clusterName }}-$(make revision) VM_SIZE=Standard_DS4_v2
```
rbtr, line 35: `Standard_DS4_v2` is a HUGE jump up in VM size and cost (12x!). I originally picked B2s as the cheapest viable option - did you evaluate any other SKUs? Maybe we could start at `Standard_B2ms` which has double the mem at only double the cost?

test/integration/manifests/goldpinger/deployment.yaml
```diff
@@ -76,3 +76,10 @@ spec:
               port: 8080
             initialDelaySeconds: 5
             periodSeconds: 5
+          resources:
```
rbtr, line 79: The resource block is omitted intentionally so that the goldpinger Pods will overprovision on the nodes. When you add this, only (node mem)/100Mi Goldpinger pods can be scheduled on the Node.

rbtr (changes requested):

test/integration/manifests/goldpinger/deployment.yaml
vipul-21: So if the node mem is 8gb( in case `Standard_B2ms`) the number of pods that can be schedule on that node would be 80. And we need to scale the pods till 100.
Do you recommend adding a smaller limit to accommodate approx ~110 pods OR remove the limit altogether so that we can overprovision on that node

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
vipul-21: The `Standard_B2ms` worked without the resource limit. @rbtr . Thanks for the help
Will try another run to be sure.

rbtr (approved):

PR #2017, opened 2023-06-14
Removing a network policy test that is unreliable

pjohnst5:
We should disable these network policy tests, as they are flaky (confirmed), and we now have npm conformance tests which are better (and are scoped to NPM related changes, not blocking other PRs related to other components)

Here's a run that succeeded, and then a different run with the same code (for both ACN, and aks-engine), that failed:
Success: 
<internal link, dm me for details>

Failed:
<internal link, dm me for details>

You may notice that on the failed run, we are now on attempt 4 to get it to work (cc Saksham Mittal) for a release

We are removing this flaky test from the aks-engine test suite, since it is logically incorrect, but has "succeded" in the past, therefore it is not reliable, and even if we "fix" it to be logically correct, it could just cause more problems in the future, and is blocking PRs / Releases

So we are removing it

Fix:
https://github.com/csfmomo/aks-engine/pull/3
https://github.com/csfmomo/aks-engine/pull/4
https://github.com/csfmomo/aks-engine/pull/5
Release:
https://github.com/csfmomo/aks-engine/releases/tag/v1.0.9.5

### DIFF
--- a/.pipelines/singletenancy/aks-engine/e2e-step-template.yaml
+++ b/.pipelines/singletenancy/aks-engine/e2e-step-template.yaml
@@ -21,7 +21,7 @@ steps:
       cd '$(modulePath)'
 
       AKS_ENGINE_FORK=csfmomo
-      AKS_ENGINE_VERSION=v1.0.9.4
+      AKS_ENGINE_VERSION=v1.0.9.5
 
       # Get source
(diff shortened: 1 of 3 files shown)


smittal22 (approved):
LGTM

tamilmani1989 (approved):

PR #2021, opened 2023-06-16
fix: updating CNI version in dropgz to 1.5.4

pjohnst5:
Updating CNI version in dropgz to CNI v1.5.4, so that the CNI conflist installed with dropgz doesn't break CNI (was using CNI binary 1.5.0 in dropgz, which doesn't support new 'overlay' CNI conflist option, breaking it)

Release [dropgz/v0.0.8](https://github.com/Azure/azure-container-networking/releases/tag/dropgz%2Fv0.0.8) is unusable/unstable, we should not share it with anyone, should I delete the release @rbtr ?

I did not realize I had to change this variable for dropgz to use a new CNI version, that is my bad
I tested this new change, with dropgz both installing the conflist, **and** installing the binary, pod creation worked, and connectivity was good:
[new-dropgz-testing.zip](https://github.com/Azure/azure-container-networking/files/11776736/new-dropgz-testing.zip)


rbtr (approved):

pjohnst5:
I deleted v0.0.8, both tag and release of dropgz

I will figure out how to purge that image from mcr

v0.0.9 of dropgz will have the fix

pjohnst5:
v0.0.8 is purged

pjohnst5:
Incorporates #2008 (cniv1.5.4 change) into v0.0.9 dropgz

PR #2022, opened 2023-06-16
fix: reconcile initial state from CRD regardless of existing podInfo

rbtr:
**Reason for Change**:

Fix for bug where CNS may incorrectly initialize after a Node reboot

**Issue Fixed**:

A bug in CNS initialization can lead to an IP being leaked and Assigned to Pods even though it has been deallocated.

- CNS starts to scale down and marks the IP as ToBeDeleted for DNC to deallocate
- the Node reboots / DNC deallocates the IP (in parallel)
- CNS restarts. During CNS initialization, since there are no Pods running, an initialization step is skipped, and the ToBeDeleted IP is incorrectly marked as Available (it is now leaked)
- the leaked IP appears Available. As Pods are started on the Node it is incorrectly Assigned to a Pod
- normal IP pool scaling operations start to fail due to the presence of the leaked IP

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -1163,34 +1163,24 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 	poolMonitor := ipampool.NewMonitor(httpRestServiceImplementation, scopedcli, clusterSubnetStateChan, &poolOpts)
 	httpRestServiceImplementation.IPAMPoolMonitor = poolMonitor
 
-	// reconcile initial CNS state from CNI or apiserver.
-	// Only reconcile if there are any existing Pods using NC ips,
-	// else let the goal state be updated using a regular NNC Reconciler loop
-	podInfoByIP, err := podInfoByIPProvider.PodInfoByIP()
-	if err != nil {
-		return errors.Wrap(err, "failed to provide PodInfoByIP")
-	}
-	if len(podInfoByIP) > 0 {
-		logger.Printf("Reconciling initial CNS state as PodInfoByIP is not empty: %d", len(podInfoByIP))
-
-		// apiserver nnc might not be registered or api server might be down and crashloop backof puts us outside of 5-10 minutes we have for
-		// aks addons to come up so retry a bit more aggresively here.
-		// will retry 10 times maxing out at a minute taking about 8 minutes before it gives up.
-		attempt := 0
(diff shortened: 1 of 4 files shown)


cns/service/main.go
```diff
@@ -1163,34 +1163,24 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 	poolMonitor := ipampool.NewMonitor(httpRestServiceImplementation, scopedcli, clusterSubnetStateChan, &poolOpts)
 	httpRestServiceImplementation.IPAMPoolMonitor = poolMonitor
 
-	// reconcile initial CNS state from CNI or apiserver.
```
tamilmani1989, line 1166: this comment is still true right? if so, lets not remove it and also update as "CNI or CNS State or apiserver"

rbtr: it's technically true, but not useful. the decision point for _how_ we're reconciling happens way earlier than this, i think this log is leftover from a much earlier iteration

tamilmani1989 (approved):

thatmattlong (approved):

cns/service/main.go
```diff
@@ -1163,34 +1163,24 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
-		err = retry.Do(func() error {
-			attempt++
-			logger.Printf("reconciling initial CNS state attempt: %d", attempt)
-			err = reconcileInitialCNSState(ctx, scopedcli, httpRestServiceImplementation, podInfoByIPProvider)
-			if err != nil {
-				logger.Errorf("failed to reconcile initial CNS state, attempt: %d err: %v", attempt, err)
-			}
-			return errors.Wrap(err, "failed to initialize CNS state")
-		}, retry.Context(ctx), retry.Delay(initCNSInitalDelay), retry.MaxDelay(time.Minute))
+	logger.Printf("Reconciling initial CNS state")
```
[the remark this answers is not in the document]
rbtr: it is logged once we get to [`reconcileNCState`](https://github.com/Azure/azure-container-networking/blob/5e67aa6b4fb26ebc75cd046f0176538ca3469b54/cns/restserver/internalapi.go#L278C57-L278C68)

PR #2023, opened 2023-06-19
[Fix][CNS] InterfaceID should be set to containerID instead of interfaceName if CNS managing state

tamilmani1989:
**Reason for Change**:

This PR fixes an issue where interfaceID is set to interfaceName instead of containerID and this leads to incorrect CNS reconcilation where CNS key is interfaceName instead of containerID. This is the PR to change CNS running in managed state to recocile based on InterfaceID instead of podname+namespace: https://github.com/Azure/azure-container-networking/pull/1811 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/ipam.go
+++ b/cns/restserver/ipam.go
@@ -655,6 +655,7 @@ func (service *HTTPRestService) GetExistingIPConfig(podInfo cns.PodInfo) ([]cns.
 		}
 	}
 
+	logger.Printf("[GetExistingIPConfig] IPConfigExists [%t] for pod [%+v]", ipConfigExists, podInfo.Key())
 	return podIPInfo, ipConfigExists, nil
 }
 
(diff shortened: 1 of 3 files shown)


neaggarwMS (approved):

neaggarwMS:
Can you add a log line here too:

GetExistingIPConfig() <<-- function too, if it happens to find an existing info for a given PodKey

rbtr (approved):

neaggarwMS (approved):

rbtr (approved):

cns/restserver/ipam.go
```diff
@@ -655,6 +655,7 @@ func (service *HTTPRestService) GetExistingIPConfig(podInfo cns.PodInfo) ([]cns.
 		}
 	}
 
+	logger.Printf("[GetExistingIPConfig] IPConfigExists [%t] for pod [%+v]", ipConfigExists, podInfo.Key())
```
neaggarwMS, line 658: can we print the whole podInfo vs just key

neaggarwMS (changes requested):

cns/restserver/ipam.go
tamilmani1989: podinfo is an interface and not a struct

rbtr: I recently implemented `String()` on the PodInfos: https://github.com/Azure/azure-container-networking/pull/2020
You can log it by using `%s` instead of `%v`

neaggarwMS (approved):

PR #2027, opened 2023-06-21
fix: setdir in CNI client

rbtr:
**Reason for Change**:

Sets the working directory when CNS invokes the CNI (only on Windows) so that the CNS reconcile from CNI finds the real CNI statefile and reinitializes the CNS state correctly.

**Issue Fixed**:

Due to the lack of [CNIRuntimePath set when CNI is running on Windows](https://github.com/Azure/azure-container-networking/blob/390977d5c2128fdd57c39c70a1fdb79035edfb70/platform/os_windows.go#LL27C21-L27C21): when CNS invokes CNI to restore state during initialization on Windows, it looks for the [statefile in the local directory](https://github.com/Azure/azure-container-networking/blob/390977d5c2128fdd57c39c70a1fdb79035edfb70/cni/plugin.go#L170) (`.\azure-vnet.json`) instead of the absolute path at `C:\k\azure-vnet.json`. No statefile exists in the current directory, so the CNI returns empty PodInfo to CNS, leading CNS to think there are no Pods scheduled and that all of its IPs are available.

Running a CNS built with this change, we can see that it finds the CNI statefile and correctly restores from it:
> 2023/06/21 18:11:58 [17776] Initializing from CNI
2023/06/21 18:11:58 [17776] Setting cmd to CNI_COMMAND=GET_ENDPOINT_STATE
2023/06/21 18:11:58 [17776] Reconciling initial CNS state as PodInfoByIP is not empty: 19
2023/06/21 18:11:58 [17776] reconciling initial CNS state attempt: 1

restoring CNS to the previous version and restarting shows that the statefile is not used:
> 2023/06/21 18:13:54 [124] Initializing from CNI
2023/06/21 18:13:54 [124] Setting cmd to CNI_COMMAND=GET_ENDPOINT_STATE
[no reconcile logs]
2023/06/21 18:13:56 [124] initialized and started IPAM pool monitor

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/client/client.go
+++ b/cni/client/client.go
@@ -14,12 +14,6 @@ import (
 	utilexec "k8s.io/utils/exec"
 )
 
-type Client interface {
-	GetEndpointState() (*api.AzureCNIState, error)
-}
-
-var _ (Client) = (*client)(nil)
-
 type client struct {
 	exec utilexec.Interface
 }
@@ -30,7 +24,7 @@ func New(exec utilexec.Interface) *client {
 
 func (c *client) GetEndpointState() (*api.AzureCNIState, error) {
 	cmd := c.exec.Command(platform.CNIBinaryPath)
-
+	cmd.SetDir(CNIExecDir)
 	envs := os.Environ()
 	cmdenv := fmt.Sprintf("%s=%s", cni.Cmd, cni.CmdGetEndpointsState)
 	log.Printf("Setting cmd to %s", cmdenv)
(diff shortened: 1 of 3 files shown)


jaer-tsun (approved):

cni/client/client.go
```diff
@@ -30,7 +24,7 @@ func New(exec utilexec.Interface) *client {
 
 func (c *client) GetEndpointState() (*api.AzureCNIState, error) {
 	cmd := c.exec.Command(platform.CNIBinaryPath)
-
+	cmd.SetDir(CNIExecDir)
```
tamilmani1989, line 27: add a comment to say this work in windows only if state exists in c:\k

tamilmani1989, line 27: if CNIExecDir is empty, SetDir will skip setting directory?

rbtr: effectively yes
https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/os/exec/exec.go;l=171-174

PR #2031, opened 2023-06-23
fix: [CNI] CNI load test failing due to namespace already created

vipul-21:
**Reason for Change**:

The load test pipeline for CNI is failing if the `load-test` namespace is already created.

**Issue Fixed**:

Fixes the load test pipeline.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/internal/k8sutils/utils_create.go
+++ b/test/internal/k8sutils/utils_create.go
@@ -169,9 +169,6 @@ func MustCreateNamespace(ctx context.Context, clienset *kubernetes.Clientset, na
 		},
 	}, metav1.CreateOptions{})
 
-	if apierrors.IsAlreadyExists(err) {
-		return errors.Wrapf(err, "namespace: %v already exists", namespace)
-	}
 	if err != nil {
 		return errors.Wrapf(err, "failed to create namespace %v", namespace)
 	}
--- a/test/integration/load/load_test.go
+++ b/test/integration/load/load_test.go
@@ -62,9 +62,17 @@ func TestLoad(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
 	defer cancel()
 
-	err = k8sutils.MustCreateNamespace(ctx, clientset, namespace)
+	// Create namespace if it doesn't exist
+	namespaceExists, err := k8sutils.NamespaceExists(ctx, clientset, namespace)
 	if err != nil {
 		t.Fatal(err)
+	}
+
+	if !namespaceExists {
+		err = k8sutils.MustCreateNamespace(ctx, clientset, namespace)
+		if err != nil {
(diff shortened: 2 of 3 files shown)


test/internal/k8sutils/utils_create.go
```diff
@@ -170,7 +170,8 @@ func MustCreateNamespace(ctx context.Context, clienset *kubernetes.Clientset, na
 	}, metav1.CreateOptions{})
 
 	if apierrors.IsAlreadyExists(err) {
-		return errors.Wrapf(err, "namespace: %v already exists", namespace)
+		log.Printf("namespace: %v already exists", namespace)
+		return nil
 	}
 	if err != nil {
 		return errors.Wrapf(err, "failed to create namespace %v", namespace)
```
jpayne3506, line 177: The function states that it must create the namespace. This change will break that convention.

jpayne3506 (changes requested):

test/internal/k8sutils/utils_create.go
vipul-21: The convention is to delete if it exists and then create a new one. ( as per other create resources func) If it is already created, then we don't need to create it as it already exists. This is a use case for us as we need the namespace for validating the state. 

We can create a new namespace after deletion if that is what you are referring to ? 
(Also this would need a func parameter to actually not delete the namespace if specified.)

jpayne3506: The function didn't create anything even though the naming convention states that it will. If the function does not create the namespace it should report with an error that it didn't. 

If we want to use the namespace even if it is already created then we can capture that in the individual tests and keep this function as is. apierrors.IsAlreadyExists() will properly read that the namespace already existed even after the error.Wrapf(). 

The convention of deleting and creating new resources isn't being followed by this function and I didn't realize that was the convention set forth by this file. I don't believe we should delete the namespace if it already exists, but I still believe it should report an error as long as the function is named MustCreate.

vipul-21: The `MustCreate/Delete` first tries to `Create` or `Delete` but in some case if it is not able to do so it should return error. This is what you meant by convention right ? 
If so, the check in either Create i.e ifAlreadyexits or Delete i.e isNotFound should be outside these functions. These functions were added as a wrap around the API which go client provides so we don't have fail when we hit alreadyexist or notfound cases. (By returning return nil in case either of the above check fails).
I think the functionality you are referring to can be achieved by just calling the k8s API.
If we keep the function as it is, then we don't need the `apierrors.IsAlreadyExists()` as it does not add value to the error.

jpayne3506 (approved):

jpayne3506:
Lgtm.

PR #2032, opened 2023-06-24
dualstack Overlay ACN pipeline 

paulyufan2:
**Reason for Change**:

This PR is the datapath for ACN Linux test and dualstack Linux node properties check and cns state file check.
To run them locally, use following commands on dualstack overlay setup:

Linux Test Commands:
go test -timeout 30m -tags load -run ^TestLoad$ -tags=load
go test -timeout 30m -tags load -cni cniv2 -run ^TestValidateState$ -tags=load
go test -timeout 30m -tags load -run ^TestDualStackProperties$ -tags=load
go test -count=1 test/integration/datapath/datapath_linux_test.go -timeout 1m -tags connection -run ^TestDatapathLinux$ -tags=connection,integration -v

Windows Test Commands
go test -timeout 30m -tags load -run ^TestLoad$ -tags=load -os=windows
go test -timeout 30m -tags load -run ^TestDualStackProperties$ -tags=load -os=windows -cni cniv2
go test -timeout 30m -tags load -cni cniv2 -run ^TestValidateState$ -tags=load -os=windows
go test -count=1 test/integration/datapath/datapath_windows_test.go -timeout 3m -tags connection -run ^TestDatapathWin$ -tags=connection -v

Logs:
https://msazure.visualstudio.com/One/_build/results?buildId=76033872&view=results

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pjohnst5 (changes requested):

test/integration/datapath/datapath_linux_test.go
```diff
@@ -0,0 +1,325 @@
...
+)
+
+/*
+This test assumes that you have the current credentials loaded in your default kubeconfig for a
+k8s cluster with a Linux nodepool consisting of at least 2 Linux nodes.
+*** The expected nodepool name is npwin, if the nodepool has a diferent name ensure that you change nodepoolSelector with:
+	-nodepoolSelector="yournodepoolname"
+
+To run the test use one of the following commands:
+go test -count=1 test/integration/datapath/datapath_linux_test.go -timeout 3m -tags connection -run ^TestDatapathLinux$ -tags=connection,integration
```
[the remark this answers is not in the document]
pjohnst5: See above comment about this, resolving this one

test/integration/datapath/datapath_linux_test.go
```diff
@@ -0,0 +1,325 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	apiv1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	LinuxDeployIPV4            = "../manifests/datapath/linux-deployment.yaml"
+	LinuxDeployIPv6            = "../manifests/datapath/linux-deployment-ipv6.yaml"
+	podLabelKey                = "app"
```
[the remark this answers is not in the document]
paulyufan2: Sure, here needs to optimize:
(1) Rename the file
(2) Create another file to define same values

test/internal/k8sutils/utils_get.go
```diff
@@ -43,9 +43,11 @@ func GetPodsIpsByNode(ctx context.Context, clientset *kubernetes.Clientset, name
 	if err != nil {
 		return nil, err
 	}
-	ips := make([]string, 0, len(pods.Items))
+	ips := make([]string, 0, len(pods.Items)*2) //nolint
 	for index := range pods.Items {
-		ips = append(ips, pods.Items[index].Status.PodIP)
+		for _, podIP := range pods.Items[index].Status.PodIPs {
+			ips = append(ips, podIP.IP)
```
jpayne3506, line 49: Why do you want N*2 space allocated?

test/integration/datapath/datapath_linux_test.go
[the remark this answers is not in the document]
jpayne3506: Today I learned! Does GO know based off of what GOOS is set as? If so, that would avoid the overwriting from occurring if the filename change happens. 

Windows clusters can be run on a linux VM. Wouldn't that break this setup? If so, perhaps they shouldn't be under the same build tag.

test/integration/datapath/datapath_linux_test.go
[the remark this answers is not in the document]
jpayne3506: Looking at make test-integration, it uses -mod=readonly, which ensures that it doesnt read the vendor directory. Would that be more beneficial here?

test/internal/k8sutils/utils_get.go
paulyufan2: Hi John, yes because here we might have 2 * times IPs in dualstack overlay scenario (each pod has two IPs, one is ipv4 and another is ipv6)

test/integration/datapath/datapath_linux_test.go
```diff
@@ -0,0 +1,325 @@
...
+						t.Fatalf("dualstack overlay pod properties check is failed due to: %v", err)
+					}
+
+					t.Log("all dualstack linux pods properties have been verified")
+				})
+			})
+		}
+
+		t.Run("all linux pods can ping each other", func(t *testing.T) {
+			clusterCheckCtx, cancel := context.WithTimeout(ctx, 3*time.Minute)
```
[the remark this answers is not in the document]
paulyufan2: yes, it's specifically used for goldpinger to get clusterState

test/integration/datapath/datapath_linux_test.go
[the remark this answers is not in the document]
rbtr: there should be no vendor directory

test/integration/datapath/datapath_linux_test.go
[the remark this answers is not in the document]
rbtr: > Does GO know based off of what GOOS is set as

yes

`*_windows.go` is equivalent to adding a build tag `//go:build windows` in the file
https://pkg.go.dev/go/build

hack/aks/Makefile
```diff
@@ -211,6 +211,31 @@ windows-cniv1-up: rg-up overlay-net-up ## Bring up a Windows CNIv1 cluster
...
+		--ip-families ipv4,ipv6 \
+		--network-plugin-mode overlay \
+		--subscription $(SUB) \
+		--no-ssh-key \
+		--yes
+	@$(MAKE) set-kubeconf
+
+dualstack-windows-up: ## Brings up windows nodes on dualstack overlay cluster
+	$(AZCLI) aks nodepool add -g $(GROUP) -n npwin \
+	    --cluster-name  $(CLUSTER) \
```
vipul-21, line 229: nit: indentation is not aligned with the other attributes.

test/integration/datapath/datapath_linux_test.go
```diff
@@ -0,0 +1,325 @@
...
+			daemonset, err = k8sutils.MustParseDaemonSet(gpDaemonset)
+			if err != nil {
+				t.Fatal(err)
+			}
+		} else {
+			daemonset, err = k8sutils.MustParseDaemonSet(gpDaemonsetIPv6)
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
```
vipul-21, line 120: nit: This can be replaced with(easier to scale if we need to add more test profiles.)
```
var gpDaemonset= map[string]string{
	"LinuxDeployIPV4":  "filename_linux",
	"DualStack"      :  "filename"   
}
...
daemonset, err = k8sutils.MustParseDaemonSet(gpDaemonset[*testProfile])
if err != nil {
	t.Fatal(err)
}
...
```

test/integration/datapath/datapath_linux_test.go
```diff
@@ -0,0 +1,325 @@
...
+	t.Log("Create Label Selectors")
+	podLabelSelector, nodeLabelSelector := createLabelSelectors()
+
+	t.Log("Get Nodes")
+	nodes, err := k8sutils.GetNodeListByLabelSelector(ctx, clientset, nodeLabelSelector)
+	if err != nil {
+		require.NoError(t, err, "could not get k8s node list: %v", err)
+	}
+
+	createPodFlag := !(apierrors.IsAlreadyExists(err))
```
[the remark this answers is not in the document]
vipul-21: i think we should be checking for namespace creation(create if not exists) as the podnamespace is a flag

test/integration/datapath/datapath_linux_test.go
```diff
@@ -0,0 +1,325 @@
...
+
+				for _, pod := range podList.Items {
+					if pod.Status.PodIP == "" {
+						return errors.New("a pod has not been allocated an IP")
+					}
+				}
+
+				return nil
+			}
+			err := defaultRetrier.Do(ctx, checkPodIPsFn)
```
vipul-21, line 231: The function `WaitForPodsRunning` does the same thing, we can use it here.

test/integration/datapath/datapath_linux_test.go
```diff
@@ -0,0 +1,325 @@
...
+							if len(podIPs) < 2 {
+								return errors.New("a pod only gets one IP")
+							}
+							if net.ParseIP(podIPs[0].IP).To4() == nil || net.ParseIP(podIPs[1].IP).To16() == nil {
+								return errors.New("a pod does not have both ipv4 and ipv6 address")
+							}
+						}
+						return nil
+					}
+					err := defaultRetrier.Do(ctx, checkPodIPsFn)
```
vipul-21, line 261: Same for this

test/integration/manifests/load/privileged-daemonset.yaml
```diff
@@ -24,15 +24,22 @@ spec:
             privileged: true
             runAsUser: 0
           volumeMounts:
-            - mountPath: /var/run/azure-cns
+            - mountPath: /var/run/azure-network
```
vipul-21, line 27: This might break the existing cilium overlay load test pipeline fi privileged pods are used to get the json file. ( I don't think they do, but good to check)

test/internal/k8sutils/utils_get.go
```diff
@@ -43,9 +43,11 @@ func GetPodsIpsByNode(ctx context.Context, clientset *kubernetes.Clientset, name
 	if err != nil {
 		return nil, err
 	}
-	ips := make([]string, 0, len(pods.Items))
+	ips := make([]string, 0, len(pods.Items)*2) //nolint
```
vipul-21, line 46: Why `//nolint` ?

test/validate/client.go
```diff
@@ -20,6 +20,8 @@ type Validator struct {
 type IValidator interface {
 	ValidateStateFile() error
 	ValidateRestartNetwork() error
+	ValidateDualStackNodeProperties() error
```
vipul-21, line 23: fyi, You will need to create an empty method for windows validator too.

test/validate/linux_validate.go
```diff
@@ -158,6 +225,31 @@ func cnsStateFileIps(result []byte) (map[string]string, error) {
...
+		return nil, errors.Wrapf(err, "failed to unmarshal azure cni endpoint list")
+	}
+
+	azureCnsPodIps := make(map[string]string)
+	for _, v := range azureCNSNonCiliumResult.AzureCniState.ExternalInterfaces {
+		for _, networks := range v.Networks {
+			for _, ip := range networks.Endpoints {
+				pod := ip.PODName
+				ipv4 := ip.IPAddresses[0].IP
+				azureCnsPodIps[ipv4] = pod
```
vipul-21, line 241: I think this would pass in both cases, but want to make sure the variable names though. In case of dual stack the first IP address is IPv4, i thought it would be IPv6 ?

test/validate/linux_validate.go
```diff
@@ -230,3 +322,39 @@ func (v *LinuxValidator) validate(stateFileIps stateFileIpsFunc, cmd []string, c
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to get node list")
+	}
+
+	for index := range nodes.Items {
+		nodeName := nodes.Items[index].ObjectMeta.Name
+		// check node status
+		nodeConditions := nodes.Items[index].Status.Conditions
+		if nodeConditions[len(nodeConditions)-1].Type != corev1.NodeReady {
+			return errors.Wrapf(err, "node %s status is not ready", nodeName)
```
vipul-21, line 338: Is this because of error ? Or we expect the node to be ready after a certain time. ?
If it is going to be ready after a certain time, should we add a retry ?

vipul-21 (changes requested):

test/validate/client.go
paulyufan2: Created for windows

test/validate/linux_validate.go
paulyufan2: yes here we need to enhance because in dual nic scenario. two ips may be all ipv4 addresses

test/validate/client.go
vipul-21: @rbtr The idea was to create a `Client` for either windows/linux to validate the statefile/ips etc. The clients will be implementing this interface to provide the functionality. 
I agree we don't use the OOP patterns in Go, but this is what I added as factory method for validating in Go.
(This is what i used as reference :https://refactoring.guru/design-patterns/factory-method/go/example .)

I am open to any suggestions(correct way) to make it "Go lang" way.

rbtr: That reference is bad. Throw it away.

> The clients will be implementing this interface to provide the functionality.

I think this is where you went wrong. In Go, we do not pre-define the interface for concrete types to implement...we don't need to. Any type which provides the _behavior_ that an interface describes implicitly satisfies that interface. It's kind of upside-down from traditional OOP "inheritance".

In this way, the business logic defines the interface which describes the behavior it needs. _Any_ type which provides that behavior implicitly satisfies the interface and can be passed into that business logic. 

We don't create an `IValidator` in our validator package - that is not a useful interface definition. Where we _use_ the behavior of a validator implementation, we define an interface which describes the (subset of) that behavior that we are using. And we probably just call it `validator` and leave it package-private, because we are not defining it for implementation elsewhere, we are defining it for our immediate use in the adjacent business logic.

For example, see [here in the NNC Reconciler](https://github.com/Azure/azure-container-networking/blob/04566d5fca97d81014f910371c1ab29b0c43b3c7/cns/kubecontroller/nodenetworkconfig/reconciler.go#L25-L35): this type depends on some behavior so interfaces are defined here to abstract us from the concrete implementations of that behavior. But the implementations don't know about this interface definition, and they don't need to. Because they provide the behavior we need, they satisfy the interface.

Stop thinking about it in the OOP/inheritance style of "type A implements Iface" and go the other direction: "Iface describes the behavior I need from A (or any other type)".

vipul-21: Okay, let me try this approach. I think I understand it, will ping if need help. Thank @rbtr

test/integration/manifests/datapath/linux-deployment.yaml
```diff
@@ -0,0 +1,86 @@
+apiVersion: apps/v1
```
ashvindeodhar, line 1: can we not reuse test\integration\manifests\goldpinger\deployment.yaml ?

test/integration/manifests/datapath/linux-deployment-ipv6.yaml
```diff
@@ -0,0 +1,88 @@
+apiVersion: apps/v1
```
ashvindeodhar, line 1: can this be added to goldpinger folder to be consistent with existing structure?

ashvindeodhar (changes requested):

test/integration/manifests/datapath/linux-deployment.yaml
paulyufan2: The original goldpinger deployment yaml does not have:
```
              - name: HOSTS_TO_RESOLVE
                value: "1.1.1.1 8.8.8.8 www.bing.com"
```
That we need to use to test host resolution

test/validate/client.go
vipul-21: https://github.com/Azure/azure-container-networking/pull/2047

.pipelines/singletenancy/dualstack-overlay/dualstackoverlay-e2e-job-template.yaml
```diff
@@ -0,0 +1,32 @@
+parameters:
```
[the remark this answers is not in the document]
vipul-21: The idea was to keep the singletenancy folder as it is. Maybe add a parent to it which tells us that this is PR pipeline code  and under that we can have different folder structure like: 
`pr_pipeline(or something better)/singletenancy/azcnioverlay/cilium`
`pr_pipeline(or something better)/singletenancy/azcnioverlay/v4/.yaml`
`pr_pipeline(or something better)/multitenany/azcnioverlay/dualstack/.yaml`

Thoughts ?

Makefile
```diff
@@ -762,6 +762,11 @@ test-azure-ipam: ## run the unit test for azure-ipam
 kind:
 	kind create cluster --config ./test/kind/kind.yaml
 
+# install azure Linux CNS and CNI dropgz images
+install-azure-images:
```
rbtr, line 766: if we're going to create standalone tooling for installing our components, let's do it correctly by writing some [small kustomize payloads](https://kustomize.io/) for them instead of hacking it with the integration test like this.

hack/aks/Makefile
```diff
@@ -211,6 +211,41 @@ windows-cniv1-up: rg-up overlay-net-up ## Bring up a Windows CNIv1 cluster
...
+		--node-vm-size $(VM_SIZE) \
+		--network-plugin none \
+		--network-plugin-mode overlay \
+		--ip-families ipv4,ipv6 \
+		--subscription $(SUB) \
+		--no-ssh-key \
+		--yes
+	@$(MAKE) set-kubeconf
+
+dualstack-windows-up: ## Brings up windows nodes on dualstack overlay cluster
```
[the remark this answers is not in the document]
rbtr: these make targets are misleading, they do not bring up a cluster, or (as far as I can tell) do anything dualstack-specific. it adds a windows nodepool. it should just be called add windows nodepool.

paulyufan2:
separate this PR to two PRs. Close this PR

PR #2037, opened 2023-06-29
feat: [NPM] cleanup command & label node

huntergregory:
**Reason for Change**:
Help AKS disable NPM on existing clusters.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
### Cleanup Command
You can use the following on the NPM container to clean the kernel up:
```
command: ['/usr/bin/azure-npm', 'cleanup']
```

### Node Labels
NPM will now add this Label to its Node: `kubernetes.azure.com/azure-npm=installed`.

When running the `cleanup` command, NPM will add this Label instead: `kubernetes.azure.com/azure-npm=removed`.


matmerr:
just curious, when would there case be that a node wouldn't have npm scheduled?

huntergregory:
> just curious, when would there be the case that a node wouldn't have npm scheduled?

In the future, AKS may allow users to disable NPM in an existing cluster, i.e., to change `--network-policy=azure` to `--network-policy=none`. This would e.g. help a customer upgrade from a cluster with NPM to a cluster with Cilium

During that upgrade, new nodes won't have NPM installed, and `azure-npm=installed` serves as a marker for nodes that used to have NPM

npm/cmd/start.go
```diff
@@ -119,6 +121,12 @@ func start(config npmconfig.Config, flags npmconfig.Flags) error {
 
 	k8sServerVersion := k8sServerVersion(clientset)
 
+	nodeName := models.GetNodeName()
+	err = labelNode(clientset, nodeName, util.InstalledLabelValue)
+	if err != nil {
+		metrics.SendErrorLogAndMetric(util.NpmID, "error: failed to label node as NPM installed. err: %s", err.Error())
```
wedaly, line 127: should this exit on error? or maybe retry with exponential backoff + jitter? worried that we won't label the node if apiserver is temporarily down.

npm/cmd/start.go
```diff
@@ -212,3 +220,24 @@ func k8sServerVersion(kubeclientset kubernetes.Interface) *k8sversion.Info {
...
+	k8sNode, err := clientset.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get k8s node. nodeName: %s. err: %w", nodeName, err)
+	}
+
+	if k8sNode.Labels == nil {
+		k8sNode.Labels = make(map[string]string)
+	}
+	k8sNode.Labels[util.NPMNodeLabelKey] = labelValue
+	_, err = clientset.CoreV1().Nodes().Update(context.TODO(), k8sNode, metav1.UpdateOptions{})
```
wedaly, line 237: Could we double-check what happens if some other process updates the node between the `Get` (line 228) and `Update`? I think client-go checks the version and errors here (to avoid overwriting a more recent update), but want to confirm.

npm/cmd/start.go
```diff
@@ -212,3 +220,24 @@ func k8sServerVersion(kubeclientset kubernetes.Interface) *k8sversion.Info {
...
+
+	k8sNode, err := clientset.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get k8s node. nodeName: %s. err: %w", nodeName, err)
+	}
+
+	if k8sNode.Labels == nil {
+		k8sNode.Labels = make(map[string]string)
+	}
+	k8sNode.Labels[util.NPMNodeLabelKey] = labelValue
```
wedaly, line 236: could we skip the `Update` call if the label already exists?

npm/cmd/start.go
huntergregory: Looks like `Update()` should have optimistic locking:

https://github.com/kubernetes/client-go/issues/150

https://github.com/kubernetes/client-go/issues/1058#issuecomment-1340234879

npm/cmd/start.go
huntergregory: good catch. Let's exit on error

huntergregory:
No worries, let me close this

PR #2038, opened 2023-06-29
Returning IPAddress instead of IPPrefix in CNI ADD response

ronak-31:
**Reason for Change**:
CNI ADD call is sending IPPrefix instead of IPAddress in the response.
For honeycomb container(baremetal containers) it is expected to send IPAddress.
For honeycomb container after container creation AtalsAgent creates a URL and executes http probe using it to check container's health. But ADD response is sending IPPrefix so the URL is getting formed incorrectly and the probe is failing.

For example, It will form this kind of url **http://[260a:10b1:90f:ac12::]:6798/Data** which will fail as it doesn't have exact IPAddress.

**Issue Fixed**:
Sending IPAddress instead of IPPrefix in ADD response to fix this.

**Requirements**:

**Notes**:


cni/network/network.go
```diff
@@ -1258,18 +1259,25 @@ func convertNnsToCniResult(
 
 			intIndex := i
 			for _, ip := range ni.Ipaddresses {
+				ipAddr := net.ParseIP(ip.Ip)
 
-				ipWithPrefix := fmt.Sprintf("%s/%s", ip.Ip, ip.PrefixLength)
-				_, ipNet, err := net.ParseCIDR(ipWithPrefix)
+				var address net.IPNet
```
tamilmani1989, line 1264: can we move this above line 1272 (before To4() check)

cni/network/network.go
```diff
@@ -1258,18 +1259,25 @@ func convertNnsToCniResult(
...
+				prefixLength, err := strconv.Atoi(ip.PrefixLength)
```
tamilmani1989, line 1265: can't we make ip.PrefixLength as integer?

tamilmani1989 (changes requested):

cni/network/network.go
ronak-31: ip.PrefixLength is part of nnscontracts so it will be difficult to change it to integer.

cni/network/network.go
ronak-31: Sure, will move it above 1272

cni/network/network.go
```diff
@@ -1258,18 +1259,25 @@ func convertNnsToCniResult(
...
 			for _, ip := range ni.Ipaddresses {
+				ipAddr := net.ParseIP(ip.Ip)
 
-				ipWithPrefix := fmt.Sprintf("%s/%s", ip.Ip, ip.PrefixLength)
-				_, ipNet, err := net.ParseCIDR(ipWithPrefix)
+				prefixLength, err := strconv.Atoi(ip.PrefixLength)
 				if err != nil {
-					log.Printf("Error while converting to cni result for %s operation on pod %s. %s",
-						operationName, podName, err)
+					log.Printf("Error parsing prefixLength (%s) for %s operation on pod %s, err:%s.",
```
tamilmani1989, line 1266: we moved to zap logging meanwhile.. you can pull latest master and update it

ronak-31: Pulled latest changes and updated to zap logging.

cni/network/network.go
```diff
@@ -1341,20 +1342,28 @@ func convertNnsToCniResult(
...
 						zap.Error(err))
 					continue
 				}
 
+				var address net.IPNet
+				if ipAddr.To4() != nil {
+					address = net.IPNet{IP: ipAddr, Mask: net.CIDRMask(prefixLength, 32)}
+				} else {
+					address = net.IPNet{IP: ipAddr, Mask: net.CIDRMask(prefixLength, 128)}
+				}
```
timraymond, line 1362: It's more idiomatic to eliminate `else` where possible in Go

```suggestion
				address := net.IPNet{
					IP: ipAddr,
					Mask: net.CIDRMask(prefixLength, 128)
				}
				
				if ipAddr.To4() != nil {
					address.Mask = net.CIDRMask(prefixLength, 32)
				}
```

ronak-31: Updated pull request with suggested changes.

tamilmani1989 (dismissed):

ronak-31:

tamilmani1989:
e2e succeeded in retry https://msazure.visualstudio.com/One/_build/results?buildId=76596503&view=results

PR #2047, opened 2023-07-07
refactor: [CI-CNI] Removing the predefined interfaces from validate

vipul-21:
Removing the predefined interfaces from validate package

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/integration/load/load_test.go
+++ b/test/integration/load/load_test.go
@@ -5,7 +5,6 @@ package load
 import (
 	"context"
 	"flag"
-	"fmt"
 	"testing"
 	"time"
 
@@ -36,13 +35,6 @@ var noopDeploymentMap = map[string]string{
 	"linux":   manifestDir + "/noop-deployment-linux.yaml",
 }
 
-// Todo: Add the validation for the data path function for the linux/windows client.
-type stateValidator interface {
-	ValidateStateFile(context.Context) error
-	ValidateRestartNetwork(context.Context) error
-	// ValidateDataPath() error
-}
-
 /*
 In order to run the scale tests, you need a k8s cluster and its kubeconfig.
 If no kubeconfig is passed, the test will attempt to find one in the default location for kubectl config.
@@ -60,10 +52,10 @@ todo: consider adding the following scenarios
 - [x] Test the CNS Local cache.
 - [x] Test the Cilium state file.
 - [x] Test the Node restart.
-- [ ] Test based on operating system.
-- [ ] Test the HNS state file.
-- [ ] Parameterize the os, cni and number of iterations.
-- [ ] Add deployment yaml for windows.
+- [x] Test based on operating system.
+- [x] Test the HNS state file.
+- [x] Parameterize the os, cni and number of iterations.
+- [x] Add deployment yaml for windows.
 */
 func TestLoad(t *testing.T) {
 	clientset, err := k8sutils.MustGetClientset()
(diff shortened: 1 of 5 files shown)


test/validate/client.go
```diff
@@ -2,39 +2,105 @@ package validate
...
+		}
+		err = k8sutils.WaitForPodsRunning(v.ctx, v.clientset, "", "")
+		if err != nil {
+			return errors.Wrapf(err, "failed to wait for pods running")
+		}
+	}
+	return nil
+}
+
+func (v *validator) validate(stateFileIps stateFileIpsFunc, cmd []string, checkType, namespace, labelSelector string) error {
```
paulyufan2, line 68: Can we rename it as validateIPs?

test/integration/load/load_test.go
```diff
@@ -131,21 +131,28 @@ func TestValidateState(t *testing.T) {
...
-	err = validator.ValidateRestartNetwork()
-	if err != nil {
-		t.Fatal(err)
+		err = linuxClient.Validate()
+		if err != nil {
+			t.Fatal(err)
+		}
 	}
+
+	t.Log("State file validated successfully")
```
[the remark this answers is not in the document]
timraymond: @jpayne3506 You shouldn't log at all when things are successful. The status of the test will tell you the same information. `testing.(*T).Log` is really a bit of a misnomer, since it's really to accumulate line-stamped log messages in a buffer that will be displayed if `testing.(*T).Fail` or `testing.(*T).FailNow` are called. It's not really a "logger" in the traditional sense.

test/integration/load/load_test.go
```diff
@@ -131,21 +131,26 @@ func TestValidateState(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
 	defer cancel()
 
-	t.Log("Validating the state file")
-	validatorClient := validate.GetValidatorClient(*osType)
-	validator := validatorClient.CreateClient(ctx, clientset, config, namespace, *cniType, *restartCase)
+	if *osType == "windows" {
+		t.Log("Validating the windows state file")
+		windowsClient := &validate.WindowsClient{}
+		windowsClient = windowsClient.CreateClient(ctx, clientset, config, namespace, *cniType, *restartCase)
```
thatmattlong, line 137: no need to create an empty struct to then populate later, just use a constructing function. Change this
```
func (*WindowsClient) CreateClient(context.Context, ...)
```
to construct the WindowsClient itself:
```
func CreateWindowsClient(context.Context, ...) (*WindowsClient) {
    return &WindowsClient{
        clientset: clientset,
        config: config,
        ...
    }
}
```

test/integration/load/load_test.go
```diff
@@ -131,21 +131,26 @@ func TestValidateState(t *testing.T) {
...
+		t.Log("Validating the linux state file and restart network scenario")
+		linuxClient := &validate.LinuxClient{}
+		linuxClient = linuxClient.CreateClient(ctx, clientset, config, namespace, *cniType, *restartCase)
 
-	//We are restarting the systmemd network and checking that the connectivity works after the restart. For more details: https://github.com/cilium/cilium/issues/18706
-	t.Log("Validating the restart network scenario")
-	err = validator.ValidateRestartNetwork()
-	if err != nil {
-		t.Fatal(err)
+		err = linuxClient.Validate()
```
thatmattlong, line 148: nit:
```
if err := linuxClient.Validate(); err != nil {
    t.Fatal(err)
}
```
combining the if statement with err initialization keeps the `err` variable scoped only to the if block, so it's better to do this when possible to not leak scope.

test/integration/load/load_test.go
vipul-21: Cool, updated it with the rebase from your PR.

test/validate/client.go
```diff
@@ -2,39 +2,105 @@ package validate
...
 
+	k8sutils "github.com/Azure/azure-container-networking/test/internal/k8sutils"
+	"github.com/pkg/errors"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
-type Validator struct {
+type validator struct {
 	ctx         context.Context
```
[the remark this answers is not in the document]
vipul-21: Rebased with the #2082

test/integration/load/load_test.go
```diff
@@ -139,39 +131,27 @@ func TestValidateState(t *testing.T) {
...
 			t.Fatal(err)
 		}
+		if err := windowsValidator.Validate(ctx); err != nil {
+			t.Fatal(err)
+		}
 	default:
 		t.Fatalf("unknown os type %s", *osType)
 	}
-
-	err = validator.ValidateStateFile(ctx)
```
thatmattlong, line 160: looks like this got removed, maybe on accident?

test/integration/load/load_test.go
```diff
@@ -139,39 +131,27 @@ func TestValidateState(t *testing.T) {
...
-		t.Fatal(err)
-	}
-
-	// We are restarting the systmemd network and checking that the connectivity works after the restart. For more details: https://github.com/cilium/cilium/issues/18706
-	t.Log("Validating the restart network scenario")
-	t.Run(fmt.Sprintf("validate network restart - %s", *osType), func(t *testing.T) {
-		if *osType == "windows" {
-			t.Skip("validate network restart not implemented on Windows")
-		}
-		if err := validator.ValidateRestartNetwork(ctx); err != nil {
```
thatmattlong, line 171: also getting removed on accident?

test/integration/load/load_test.go
vipul-21: Since this is linux specific, I moved it linux validate file instead of load test.

test/validate/linux_validate.go
```diff
@@ -64,6 +72,10 @@ type Address struct {
 	Addr string `json:"ipv4"`
 }
 
+type LinuxValidator struct {
+	validator linuxValidator
```
thatmattlong, line 76: why have a LinuxValidator and a WindowsValidator that just hold validators? I'm guessing you are trying to implement some os-specific code. I would suggest to just remove this concept of having OS-specific validators, so just get rid of LinuxValidator and WindowsValidator, and just have one (exported) Validator. Then the os-specific branching can happen in the funcs where it needs to.

vipul-21: Yeah that made sense. Removed os specifci validator.  Branching the funcs if needed based on os.

paulyufan2 (dismissed):
lgtm

PR #2051, opened 2023-07-11
Adding cni and cns deamonset specs to target managed DNC scenario

thomasricci:
…en AI)

**Reason for Change**:
We have an E2E test pipeline that is running managed dnc scenario datapath tests. Currently the version is pinned to v1.4.7 by pulling files from this repository. I'd like to have a separate set of CNI and CNS yaml that we can use for managed dnc scenarios and update whenever we need to.

**Issue Fixed**:
Updating CNI and CNS versions to the ones used by our customers.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


tools/acncli/deployment/mdnc_azure-cni-manager.yaml
```diff
@@ -0,0 +1,57 @@
...
+      nodeSelector:
+        "beta.kubernetes.io/os": linux
+      tolerations:
+        - effect: NoSchedule
+          key: node-role.kubernetes.io/master
+          operator: Exists
+      hostNetwork: true
+      containers:
+        - name: azure-cni-installer
+          image: mcr.microsoft.com/containernetworking/azure-cni-manager:v1.4.35
```
thatmattlong, line 24: confirmed this image exists? at some point we probably stopped building it

thomasricci: I do see this release: https://github.com/Azure/azure-container-networking/releases/tag/v1.4.35 and that's also in our customer logs... let's see if there's any issue with my E2E run.

thomasricci: Actually trying to pull the image manually it seems not found

tools/acncli/deployment/mdnc_azure-cni-manager.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: apps/v1
```
tamilmani1989, line 1: why dont we update and resuse this https://github.com/Azure/azure-container-networking/blob/672cce9460be6e659c15356b6213ba8485271eb9/tools/acncli/deployment/manager_swift.yaml ?

thomasricci: Because I don't know who else is consuming it... I don't want to cause side effects somewhere else.

rbtr:
I think this should probably be in `.pipelines/mdnc` or something like that instead of in the CNS directory

cns/mdnc_azure-cns.yaml
```diff
@@ -0,0 +1,207 @@
+apiVersion: v1
```
rbtr, line 1: I think this should probably be in `.pipelines/mdnc` or something like that instead of in the CNS directory

rbtr (changes requested):

thomasricci:
Got green E2E from our nightly pipeline after manually restarting cns. It seems the manual restart is solving the issue for now.
https://dev.azure.com/msazure/One/_build/results?buildId=76481273&view=logs&j=742b4e23-ddbf-50e5-ec89-a2dad8ad71b0&t=742b4e23-ddbf-50e5-ec89-a2dad8ad71b0

rbtr (approved):

PR #2055, opened 2023-07-12
ci: use team build pool

matmerr:
**Reason for Change**:

Faster than waiting in the massive pool

edit: using this pr for manual testing, do not merge, no intention to merge

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr (dismissed):

vipul-21:
@matmerr I have this PR already open with the same changes: https://github.com/Azure/azure-container-networking/pull/2036
But i think because of the ACN pr pipeline it is not merged yet.

hack/aks/Makefile
```diff
@@ -10,7 +10,7 @@ AZCLI   ?= docker run --rm -v $(AZCFG):/root/.azure -v $(KUBECFG):/root/.kube -v
 # overrideable defaults
 REGION     ?= westus2
 OS_SKU     ?= Ubuntu
-VM_SIZE	   ?= Standard_B2s
+VM_SIZE	   ?= Standard_DS3_v2
```
rbtr, line 13: i don't think the default vm or lb skus should be changed, these are way more expensive. if we need bigger nodes for say a load test, then that test should override this via the env var

rbtr (changes requested):

hack/aks/Makefile
matmerr: tbh I'm not opinionated about the changes in this pr, I'm just reusing the pipeline for quick clusters at the moment :)

hack/aks/Makefile
```diff
@@ -90,10 +90,12 @@ overlay-byocni-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster
 	$(AZCLI) aks create -n $(CLUSTER) -g $(GROUP) -l $(REGION) \
 		--node-count $(NODE_COUNT) \
 		--node-vm-size $(VM_SIZE) \
-		--load-balancer-sku basic \
+		--load-balancer-sku standard \
 		--network-plugin none \
 		--network-plugin-mode overlay \
-		--pod-cidr 192.168.0.0/16 \
```
thatmattlong, line 96: a bunch of the overlay and cilium tests implicitly depend on this value being 192.168.0.0/16, why do we need to change it?

thatmattlong (changes requested):

hack/aks/Makefile
matmerr: emphasis on the `do-not-merge` tag, I'm using this target for a convenient one-click byocni cilium overlay cluster for manual testing

I believe I can still use `refs/pull/2055/merge` as a target if pr is converted to draft

PR #2060, opened 2023-07-14
fix: skip reserved IPs

jaer-tsun:
**Reason for Change**:

We need to use .1 address of subnet as GW; skipping 0th and last IP of subnets since they are reserved from HNS POV

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/conversion_linux_test.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion_linux_test.go
@@ -35,3 +35,63 @@ var validOverlayRequest = &cns.CreateNetworkContainerRequest{
 		},
 	},
 }
+
+var validVNETBlockRequest = &cns.CreateNetworkContainerRequest{
+	Version: strconv.FormatInt(version, 10),
+	IPConfiguration: cns.IPConfiguration{
+		GatewayIPAddress: vnetBlockDefaultGateway,
+		IPSubnet: cns.IPSubnet{
+			PrefixLength: uint8(vnetBlockSubnetPrefixLen),
+			IPAddress:    vnetBlockPrimaryIP,
+		},
+	},
+	NetworkContainerid:   ncID,
+	NetworkContainerType: cns.Docker,
+	// Ignore first IP in first CIDR Block, i.e. 10.224.0.4
+	SecondaryIPConfigs: map[string]cns.SecondaryIPConfig{
+		"10.224.0.5": {
+			IPAddress: "10.224.0.5",
+			NCVersion: version,
+		},
+		"10.224.0.6": {
+			IPAddress: "10.224.0.6",
+			NCVersion: version,
+		},
+		"10.224.0.7": {
+			IPAddress: "10.224.0.7",
+			NCVersion: version,
+		},
+		"10.224.0.8": {
+			IPAddress: "10.224.0.8",
+			NCVersion: version,
+		},
+		"10.224.0.9": {
+			IPAddress: "10.224.0.9",
+			NCVersion: version,
+		},
+		"10.224.0.10": {
+			IPAddress: "10.224.0.10",
+			NCVersion: version,
+		},
+		"10.224.0.11": {
+			IPAddress: "10.224.0.11",
+			NCVersion: version,
+		},
+		"10.224.0.12": {
+			IPAddress: "10.224.0.12",
+			NCVersion: version,
+		},
+		"10.224.0.13": {
+			IPAddress: "10.224.0.13",
+			NCVersion: version,
+		},
+		"10.224.0.14": {
+			IPAddress: "10.224.0.14",
+			NCVersion: version,
+		},
+		"10.224.0.15": {
+			IPAddress: "10.224.0.15",
+			NCVersion: version,
+		},
+	},
+}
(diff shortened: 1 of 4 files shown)


rbtr (dismissed):
no objection as a quick fix for Windows but think we should only reserve the subnet's 0th address and broadcast instead of the .0 and .255 out of every /24 within the subnet

cns/kubecontroller/nodenetworkconfig/conversion_windows.go
```diff
@@ -10,28 +10,26 @@ import (
...
 		startingAddr = startingAddr.Next()
-	} else if nc.Type == v1alpha.VNETBlock {
-		// skipping 0th IP for the Primary IP of NC
-		startingAddr = startingAddr.Next()
 	}
 
 	// iterate through all IP addresses in the subnet described by primaryPrefix and
 	// add them to the request as secondary IPConfigs.
 	for addr := startingAddr; primaryIPPrefix.Contains(addr); addr = addr.Next() {
+		if addr.Is4() && addr.As4()[3] == 255 {
```
thatmattlong, line 30: it's not really the .255 address that is reserved, it's the last address in the subnet. This will work I think because the smallest subnet that can be used for overlay is /24 and every CIDR large than that's last address ends in 255, but I'd suggest computing the last address and skipping that instead of skipping the "255th" address, in case we ever decide to use smaller pod CIDRs for overlay

cns/kubecontroller/nodenetworkconfig/conversion_windows_test.go
```diff
@@ -13,15 +13,11 @@ var validOverlayRequest = &cns.CreateNetworkContainerRequest{
...
 	NetworkContainerid:   ncID,
 	NetworkContainerType: cns.Docker,
 	SecondaryIPConfigs: map[string]cns.SecondaryIPConfig{
-		"10.0.0.1": {
-			IPAddress: "10.0.0.1",
-			NCVersion: version,
-		},
 		"10.0.0.2": {
 			IPAddress: "10.0.0.2",
 			NCVersion: version,
```
thatmattlong, line 23: commenting on this line because github doesn't let me comment on an arbitrary line, but for LL25-29, I think if you make the change I suggest above where you skip the last IP in the block, you'd remove LL25-29

cns/kubecontroller/nodenetworkconfig/conversion_windows.go
jaer-tsun: HNS told us to skip all .255s but yeah we can just skip last one here if the subnet will never get larger

jaer-tsun: let me confirm with them again on 'last' vs .255 of every subnet

jaer-tsun: ok updated to skip last IP of subnet

cns/kubecontroller/nodenetworkconfig/conversion_windows_test.go
jaer-tsun: removed those lines since we're skipping last IP of subnet now

thatmattlong (approved):

cns/kubecontroller/nodenetworkconfig/conversion_windows.go
```diff
@@ -10,23 +10,19 @@ import (
...
 //
 //nolint:gocritic //ignore hugeparam
 func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPrefix netip.Prefix, subnet cns.IPSubnet) (*cns.CreateNetworkContainerRequest, error) {
 	secondaryIPConfigs := map[string]cns.SecondaryIPConfig{}
-
-	// if NC DefaultGateway is empty, set the 0th IP to the gateway and add the rest of the IPs
-	// as secondary IPs
-	startingAddr := primaryIPPrefix.Masked().Addr() // the masked address is the 0th IP in the subnet
+	// the masked address is the 0th IP in the subnet and startingAddr is the 2nd IP (*.1)
+	startingAddr := primaryIPPrefix.Masked().Addr().Next()
```
neaggarwMS, line 20: We also need to reserve .255

cns/kubecontroller/nodenetworkconfig/conversion_windows_test.go
```diff
@@ -13,21 +13,69 @@ var validOverlayRequest = &cns.CreateNetworkContainerRequest{
...
-			NCVersion: version,
-		},
 		"10.0.0.2": {
 			IPAddress: "10.0.0.2",
+			NCVersion: 0,
+		},
+	},
+}
+
+var validVNETBlockRequest = &cns.CreateNetworkContainerRequest{
```
ashvindeodhar, line 28: can you please also add a test case for ipv6?

jaer-tsun: we'll add later since this needs to go in for quick-fix

cns/kubecontroller/nodenetworkconfig/conversion_linux_test.go
```diff
@@ -35,3 +35,63 @@ var validOverlayRequest = &cns.CreateNetworkContainerRequest{
 		},
 	},
 }
+
+var validVNETBlockRequest = &cns.CreateNetworkContainerRequest{
```
timraymond, line 39: Can you inline these vars in the body of the tests that use them or turn them into factory funcs?

Despite their prevalence in ACN code, using package variables for test data is not a best practice.

PR #2063, opened 2023-07-18
adding CNS fix for requesting IPs with 0 NCs

rjdenney:
**Reason for Change**:

A bug was found when using older versions of CNI with versions of CNS 1.5 and newer where it would panic when there aren;'t any NCs on the NNC. When CNI requested IPs in this state CNS would think it requires "0" IPs to return since it is currently based on the number of NCs. When trying to index a slice of 0 we would then see the panic. 

This fix should lead to CNS returning an error in a few locations when we see that there are 0 NCs since this also implies that there aren't any IPs in the pool.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/restserver/ipam.go
```diff
@@ -88,8 +88,8 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
 		return
 	}
 
-	// This method can only return 1 IP. If we have more than one NC then we expect to need to return one IP per NC
-	if len(service.state.ContainerStatus) > 1 {
+	// This method can only return EXACTLY 1 IP. If we have more than one NC then we expect to need to return one IP per NC
+	if len(service.state.ContainerStatus) != 1 {
```
ashvindeodhar, line 92: this will now work when the len is 0. is that intended?

cns/restserver/ipam.go
```diff
@@ -127,7 +127,20 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
 		logger.ResponseEx(service.Name+operationName, ipconfigsRequest, reserveResp, reserveResp.Response.ReturnCode, err)
 		return
 	}
-
+	// Checks to make sure we return exactly 1 IP
+	if len(ipConfigsResp.PodIPInfo) != 1 {
```
ashvindeodhar, line 131: not clear why this code is required

cns/restserver/ipam.go
```diff
@@ -661,8 +674,15 @@ func (service *HTTPRestService) GetExistingIPConfig(podInfo cns.PodInfo) ([]cns.
...
 // Assigns a pod with all IPs desired
 func (service *HTTPRestService) AssignDesiredIPConfigs(podInfo cns.PodInfo, desiredIPAddresses []string) ([]cns.PodIpInfo, error) {
+	// Sets the number of desired IPs equal to the number of NCs so that we can get one IP per NC
 	numDesiredIPAddresses := len(desiredIPAddresses)
+	// Creates a slice of PodIpInfo with the size as number of NCs to hold the result for assigned IP configs
 	podIPInfo := make([]cns.PodIpInfo, numDesiredIPAddresses)
+	// if there are no NCs on the NNC there will be no IPs in the pool so return error
+	if numDesiredIPAddresses == 0 {
+		//nolint:goerr113 // return error
+		return podIPInfo, fmt.Errorf("no NCs found on the NNC so no IPs are in the pool")
```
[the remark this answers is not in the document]
rjdenney: I'll add one

cns/restserver/ipam.go
rjdenney: This is an error check. Before it was looking for when we have more than 1 NC since it will try to allocate 1 IP per NC. Now it will also hit this error when we have 0 NCs

cns/restserver/ipam.go
```diff
@@ -661,8 +675,15 @@ func (service *HTTPRestService) GetExistingIPConfig(podInfo cns.PodInfo) ([]cns.
 
 // Assigns a pod with all IPs desired
 func (service *HTTPRestService) AssignDesiredIPConfigs(podInfo cns.PodInfo, desiredIPAddresses []string) ([]cns.PodIpInfo, error) {
+	// Sets the number of desired IPs equal to the number of NCs so that we can get one IP per NC
 	numDesiredIPAddresses := len(desiredIPAddresses)
+	// Creates a slice of PodIpInfo with the size as number of NCs to hold the result for assigned IP configs
 	podIPInfo := make([]cns.PodIpInfo, numDesiredIPAddresses)
+	// if there are no NCs on the NNC there will be no IPs in the pool so return error
+	if numDesiredIPAddresses == 0 {
+		//nolint:goerr113 // return error
```
timraymond, line 684: This linter bypass is inappropriate here. The error returned below should be defined like:

```go
var errNoPoolIPs = errors.New("no NCs found on the NNC so no IPs are in the pool")
```

and consequently used like:

```go
return nil, errNoPoolIPs
```

**Note also the `nil`, instead of `podIPInfo`**. In cases where an error is returned, it should be a nil or zero value being returned first.

This also applies to all of the other instances of `//nolint:goerr113 // return error` for this func.

cns/restserver/ipam.go
```diff
@@ -661,8 +675,15 @@ func (service *HTTPRestService) GetExistingIPConfig(podInfo cns.PodInfo) ([]cns.
 
 // Assigns a pod with all IPs desired
 func (service *HTTPRestService) AssignDesiredIPConfigs(podInfo cns.PodInfo, desiredIPAddresses []string) ([]cns.PodIpInfo, error) {
+	// Sets the number of desired IPs equal to the number of NCs so that we can get one IP per NC
 	numDesiredIPAddresses := len(desiredIPAddresses)
+	// Creates a slice of PodIpInfo with the size as number of NCs to hold the result for assigned IP configs
 	podIPInfo := make([]cns.PodIpInfo, numDesiredIPAddresses)
+	// if there are no NCs on the NNC there will be no IPs in the pool so return error
+	if numDesiredIPAddresses == 0 {
```
timraymond, line 683: Where you basically want to map over `desiredIPAddresses`, it's usually more idiomatic to let the caller pass in a `[]cns.PodIpInfo` that is partially complete (with just the IP addresses). This lets the caller control the allocation, rather than having these intermediate allocations in here.

That also forces the caller of this method to handle the request rather than spreading that responsibility among many other funcs.

If you don't pursue that, this conditional should be at the beginning of the method (so we don't allocate a slice).

cns/restserver/ipam.go
```diff
@@ -762,6 +783,11 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
 	numIPsNeeded := len(service.state.ContainerStatus)
 	// Creates a slice of PodIpInfo with the size as number of NCs to hold the result for assigned IP configs
 	podIPInfo := make([]cns.PodIpInfo, numIPsNeeded)
+	// if there are no NCs on the NNC there will be no IPs in the pool so return error
+	if numIPsNeeded == 0 {
+		//nolint:goerr113 // return error
+		return podIPInfo, fmt.Errorf("No NCs found on the NNC so no IPs are in the pool")
```
timraymond, line 789: This is exactly why the error should be reified in a var--so it can be reused here (and checked by callers).

timraymond (changes requested):

cns/restserver/ipam.go
```diff
@@ -661,8 +675,15 @@ func (service *HTTPRestService) GetExistingIPConfig(podInfo cns.PodInfo) ([]cns.
...
 // Assigns a pod with all IPs desired
 func (service *HTTPRestService) AssignDesiredIPConfigs(podInfo cns.PodInfo, desiredIPAddresses []string) ([]cns.PodIpInfo, error) {
+	// Sets the number of desired IPs equal to the number of NCs so that we can get one IP per NC
 	numDesiredIPAddresses := len(desiredIPAddresses)
+	// Creates a slice of PodIpInfo with the size as number of NCs to hold the result for assigned IP configs
 	podIPInfo := make([]cns.PodIpInfo, numDesiredIPAddresses)
+	// if there are no NCs on the NNC there will be no IPs in the pool so return error
+	if numDesiredIPAddresses == 0 {
+		//nolint:goerr113 // return error
+		return podIPInfo, fmt.Errorf("no NCs found on the NNC so no IPs are in the pool")
```
ramiro-gamarra, line 685: Is the error message appropriate here? Seems oddly coupled that this method knows about NCs or NNCs. It only knows that it can't do anything meaningful when desired IPs are 0

cns/restserver/ipam.go
ramiro-gamarra: The majority of these lines were already present, so I don't think a major refactor is needed here, but agree with moving the input validation to be the first thing in the method.

cns/restserver/ipam.go
rjdenney: This check is looking for the situation that ipam assigns multiple IPs due to multiple NCs but since we are using the old API we will only return 1. This could potentially happen if we have multiple NCs but we are using the old API. In that situation IPAM will still try to assign multiple IPs and then only the first will be returned from the list of IPs assigned back to the client. 

To prevent this I added an error check so we catch when IPAM is trying to assign multiple IPs when using the old API.

cns/restserver/ipam.go
timraymond: I should have emphasized the first part as non-blocking, but moving the conditional as blocking feedback. I still like to provide feedback on the code overall in case someone wants to refactor it... It's unlikely we will ever schedule refactoring this, so PRs provide the only opportunity.

cns/restserver/ipam.go
timraymond: Agree @ramiro-gamarra . There's an urge to provide more insight when returning an error, but usually this causes the error to ultimately tell lies when the facts of that insight change over time. It's better to return an error about exactly what it knows and provide additional contextual information through error wrapping.

cns/restserver/ipam.go
timraymond: @rjdenney Can you put basically exactly what you wrote as a comment above that conditional? If @ashvindeodhar has that question now, it's certain that a future reader of the code will as well.

cns/restserver/ipam.go
rjdenney: This should actually be based on `len(service.state.ContainerStatus)` which is the number of NCs. I'll fix this and it will be more accurate

cns/restserver/ipam.go
rjdenney: yeah I'll add it

cns/restserver/ipam.go
```diff
@@ -20,6 +20,7 @@ import (
 var (
 	ErrStoreEmpty       = errors.New("empty endpoint state store")
 	ErrParsePodIPFailed = errors.New("failed to parse pod's ip")
+	ErrNoPoolIPs        = errors.New("no NCs found on the NNC so no IPs are in the pool")
```
rbtr, line 23: This error is returned when we have no NCs, and that's _all_ that it can definitively say. 
Speculating about the state of anything not immediately related to the cause of the error is bad - it's actively misleading to the DRI reading logs and trying to root cause an issue.

rbtr (changes requested):

cns/restserver/ipam.go
rjdenney: I can change it so that it only talks about the NC count but separate question: 

is it possible for there to be IPs in the pool without an NC?

rbtr: No, that shouldn't be possible without a lot of other things going wrong

cns/restserver/ipam.go
```diff
@@ -128,6 +129,21 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
...
 	}
 
+	// Checks to make sure we return exactly 1 IP
+	// If IPAM assigned more than 1 IP then we need to raise an error since this API can only return one IP and IPAM may have assigned more than one
+	if len(ipConfigsResp.PodIPInfo) != 1 {
+		// we send a response back saying that this API won't be able to return the amount of IPs needed to fulfill the request
+		reserveResp := &cns.IPConfigResponse{
+			Response: cns.Response{
+				ReturnCode: types.UnexpectedError,
+				Message:    fmt.Sprintf("request returned incorrect number of IPs. Expected %d and returned %d", len(service.state.ContainerStatus), len(ipConfigsResp.PodIPInfo)),
```
ramiro-gamarra, line 139: better to be explicit about `1` being the expected number of IPs

cns/restserver/ipam.go
```diff
@@ -661,7 +677,15 @@ func (service *HTTPRestService) GetExistingIPConfig(podInfo cns.PodInfo) ([]cns.
 
 // Assigns a pod with all IPs desired
 func (service *HTTPRestService) AssignDesiredIPConfigs(podInfo cns.PodInfo, desiredIPAddresses []string) ([]cns.PodIpInfo, error) {
+	// gets the number of NCs
+	numOfNCs := len(service.state.ContainerStatus)
+	// Check to make sure that the number
```
ramiro-gamarra, line 682: incomplete comment?

cns/restserver/ipam.go
```diff
@@ -756,12 +780,16 @@ func (service *HTTPRestService) AssignDesiredIPConfigs(podInfo cns.PodInfo, desi
 // Assigns an available IP from each NC on the NNC. If there is one NC then we expect to only have one IP return
 // In the case of dualstack we would expect to have one IPv6 from one NC and one IPv4 from a second NC
 func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([]cns.PodIpInfo, error) {
+	// Sets the number of IPs needed equal to the number of NCs so that we can get one IP per NC
```
ramiro-gamarra, line 783: Comment needs updating?

cns/restserver/ipam_test.go
```diff
@@ -1360,6 +1360,38 @@ func IPAMMarkExistingIPConfigAsPending(t *testing.T, ncStates []ncState) {
...
+	}
+	b, _ := testPod1Info.OrchestratorContext()
+	req.OrchestratorContext = b
+	req.DesiredIPAddresses = make([]string, 1)
+	req.DesiredIPAddresses[0] = testIP1
+
+	_, err := requestIPConfigsHelper(svc, req)
+	if err == nil {
+		t.Fatalf("Expected error. Should not be able to request IPs when there are no NCs")
+	}
```
ramiro-gamarra, line 1377: should check that the error is the one we expect, either with `errors.Is` (if the error is a sentinel value) or `errors.As` (if the error is wrapped but has a specific type)

cns/restserver/ipam_test.go
```diff
@@ -1360,6 +1360,38 @@ func IPAMMarkExistingIPConfigAsPending(t *testing.T, ncStates []ncState) {
...
+		PodInterfaceID:   testPod1Info.InterfaceID(),
+		InfraContainerID: testPod1Info.InfraContainerID(),
+	}
+	b, _ := testPod1Info.OrchestratorContext()
+	req.OrchestratorContext = b
+
+	_, err := requestIPConfigsHelper(svc, req)
+	if err == nil {
+		t.Fatalf("Expected error. Should not be able to request IPs when there are no NCs")
+	}
```
ramiro-gamarra, line 1392: same as above

cns/restserver/ipam.go
```diff
@@ -88,8 +89,8 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
 		return
 	}
 
-	// This method can only return 1 IP. If we have more than one NC then we expect to need to return one IP per NC
-	if len(service.state.ContainerStatus) > 1 {
+	// This method can only return EXACTLY 1 IP. If we have more than one NC then we expect to need to return one IP per NC
+	if len(service.state.ContainerStatus) != 1 {
```
ramiro-gamarra, line 93: let's rephrase the returned error from this condition to say that we `expected 1 but found %d`

cns/restserver/ipam.go
rjdenney: fixed

cns/restserver/ipam.go
rjdenney: updated

cns/restserver/ipam_test.go
rjdenney: added an assert

cns/restserver/ipam_test.go
rjdenney: added an assert

cns/restserver/ipam.go
timraymond: @rjdenney Good question to have though, since questions like that are usually what help hammer out good error messages :)

rbtr (approved):
looks good to me

ramiro-gamarra (approved):

cns/restserver/ipam.go
rjdenney: Added a `ErrNoNCs` error variable and check for it in the UTs as well.

timraymond (approved):

PR #2065, opened 2023-07-19
chore: Removing CNI logs that are not useful

tamilmani1989:
**Reason for Change**:

Removing logs that are not useful and spamming the log file. Also it accounts for cni execution time.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -153,12 +153,6 @@ func (plugin *NetPlugin) Start(config *common.PluginConfig) error {
 		zap.String("version", plugin.Version),
 		zap.String("component", "cni-net"))
 
-	log.Logger.Info("Os Info",
-		zap.String("platform", platform.GetOSInfo()),
-		zap.String("component", "cni-net"))
-	platform.PrintDependencyPackageDetails()
-	common.LogNetworkInterfaces()
-
 	// Initialize network manager. rehyrdration not required on reboot for cni plugin
 	err = plugin.nm.Initialize(config, false)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


paulyufan2 (approved):
lgtm

cni/network/network.go
```diff
@@ -153,12 +153,6 @@ func (plugin *NetPlugin) Start(config *common.PluginConfig) error {
 		zap.String("version", plugin.Version),
 		zap.String("component", "cni-net"))
 
-	log.Logger.Info("Os Info",
-		zap.String("platform", platform.GetOSInfo()),
-		zap.String("component", "cni-net"))
-	platform.PrintDependencyPackageDetails()
```
vipul-21, line 159: We can remove the function `PrintDependencyPackageDetails` itself since this is the only place is being used.

network/manager.go
```diff
@@ -214,12 +214,6 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
 	}
 
 	log.Printf("[net] Restored state")
-	for _, extIf := range nm.ExternalInterfaces {
```
vipul-21, line 217: Why do we need to remove this part ? Is this not used anymore?
My understanding is that number of endpoints after restoring the state is displayed in the logs.

camrynl (approved):

cni/network/network.go
tamilmani1989: i wanted to keep for sometime incase if we need this to be logged

network/manager.go
tamilmani1989: moved to a place where we create and delete endpoints.. this is causing spam for windows mulitenancy case where customer create network per vlan id and this log getting looped for all networks each time on add/del

PR #2072, opened 2023-07-22
ci:[CNI] Remove load balancer tests and provide kube-system restarts between windows tests in PR pipeline

jpayne3506:
**Reason for Change**:

Windows e2e test require a taint on linux nodes to ensure that the tests run on the windows nodepool. This has led to kube-system pods restarting in the middle of tests without enough time to restart. Resulting in the tests unable to start or recover in-between steps. If there is a more systematic failure, timeout early instead of reaching the job timeout limit of 60 minutes. 

Removing load balancer tests as they are not required for CNI e2e testing. 

**Issue Fixed**:

**Requirements**:

- [ x ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ x ] relevant PR labels added

**Notes**:


.pipelines/cni/k8s-e2e/k8s-e2e-step-template.yaml
```diff
@@ -25,6 +25,7 @@ steps:
       # Taint Linux nodes so that windows tests do not run on them
       if ${{ lower(eq(parameters.os, 'windows')) }}
       then
+        kubectl rollout status -n kube-system deployment/konnectivity-agent --timeout=3m
```
vipul-21, line 28: Is this just for windows, or we need to wait for the deployment in linux as well ?

jpayne3506: Just for windows due to the linux node being tainted. Stopping scheduling while windows tests are running.

vipul-21 (approved):

PR #2074, opened 2023-07-24
test: create nginx deployments in scale test

camrynl:
**Reason for Change**:

Adapting the NPM scale test to use for future Cilium scale pipeline. Creating a secondary deployment template for nginx, this deployment is an optional parameter. Cilium will use the nginx deployment for throughput testing.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/scale/test-scale.sh
```diff
@@ -41,6 +41,7 @@ OPTIONAL PARAMETERS:
     --debug-exit-after-print-counts       skip scale test. Just print out counts of things to be created and counts of IPSets/ACLs that NPM would create
     --num-real-services                   cluster ip service for the real deployments scheduled. Each svc will point to the respective deployment(having <num-real-replicas> pods) Default is 0
     --debug-exit-after-generation         skip scale test. Exit after generating templates
+    --num-real-nginx-deployments          create real-nginx-deployments, real-deployments should be set to 0 when this is in use. Default is 0
```
huntergregory, line 44: Like the idea. Few thoughts:
1. Probably should prevent possibility of setting both nginx and agnhost deployments.
2. There are some existing dependencies in `numRealPods` and `numRealDeployments`.

May be cleaner to set something like `real-pod-type=agnhost|nginx`? Then `numRealPods` would be the number of nginx Pods, etc.

test/scale/test-scale.sh
```diff
@@ -293,13 +299,21 @@ mkdir -p generated/networkpolicies/applied
...
 mkdir -p generated/kwok-nodes
 mkdir -p generated/deployments/real/
+mkdir -p generated/deployments/real-nginx/
 mkdir -p generated/deployments/kwok/
 mkdir -p generated/services/real/
 
 generateDeployments() {
-    local numDeployments=$1
+    if [[ $numRealNginxDeployments -gt 0 ]]; then
+        local numDeployments=$numRealNginxDeployments
```
huntergregory, line 308: this feels a bit hacky. Could we keep numDeployments and depKind as parameters?

test/scale/test-scale.sh
```diff
@@ -338,7 +352,11 @@ generateServices() {
         outFile=generated/services/$serviceKind/$name.yaml
 
         sed "s/TEMP_NAME/$name/g" templates/$serviceKind-service.yaml > $outFile
-        sed -i "s/TEMP_DEPLOYMENT_NAME/$serviceKind-dep-$i/g" $outFile
+        if [[ $numRealNginxDeployments -gt 0 ]]; then
+            sed -i "s/TEMP_DEPLOYMENT_NAME/$serviceKind-nginx-dep-$i/g" $outFile
```
huntergregory, line 356: here too, feels a bit hacky

test/scale/test-scale.sh
```diff
@@ -441,6 +459,9 @@ fi
 if [[ $numRealPods -gt 0 ]]; then
     $KUBECTL $KUBECONFIG_ARG apply -f generated/deployments/real/
 fi
+if [[ $numRealNginxDeployments -gt 0 ]]; then
+    $KUBECTL $KUBECONFIG_ARG apply -f generated/deployments/real-nginx/
```
huntergregory, line 463: Perhaps something like:
```
if [[ $numRealPods -gt 0 ]]; then
    $KUBECTL $KUBECONFIG_ARG apply -f generated/deployments/$realKind/
fi
```

huntergregory (changes requested):

test/scale/test-scale.sh
camrynl: Thanks, took this into account and made the modifications so that we keep numRealPods and numRealDeployments unchanged. Added realPodType to change the deployment types

test/scale/templates/label-nodes.sh
huntergregory, line 1: nitpick: may be better placed in test/scale/ with the other scripts

huntergregory: Is retry_kubectl used elsewhere?

huntergregory (commented):
looks good! Had some q's about the label script. Sorry to comment later on it

test/scale/templates/label-nodes.sh
camrynl: oops.. meant to put it in test/scale/, and retry_kubectl isn't used anywhere else so I removed that

huntergregory (approved):

PR #2077, opened 2023-07-25
ci:[CNI] ACN PR Pipeline speed improvements

jpayne3506:
**Reason for Change**:

Improves the time it takes to run the ACN PR Pipeline by roughly 30 minutes by:
- Moving cluster creation out of the individual E2E steps ~15 minutes
- Lowering Scaleup for windows load test ~ 4 minutes
- Decreasing the retry time for faster updates ~ 2 minutes
- Removing echo-external-node deployment after Cilium connectivity test ~ 9 minutes (17 minutes total for the individual step)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
To re-run a failed E2E stage, start from the appropriate cluster creation stage.


.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
```diff
@@ -158,6 +149,9 @@ steps:
   - script: |
       echo "validate pod IP assignment and check systemd-networkd restart"
       kubectl get pod -owide -A
+      # Deleting echo-external-node deployment until version matches TODO. Upstream TODO has change. - Replace todo with vipul work/PR in upstream
+      # Saves 17 minutes
```
vipul-21, line 153: https://github.com/cilium/cilium-cli/issues/67

.pipelines/pipeline.yaml
```diff
@@ -322,24 +322,39 @@ stages:
...
       name: "cilium_overlay_cilium_e2e"
       displayName: Cilium on AKS Overlay
       pipelineBuildImage: "$(BUILD_IMAGE)"
       testDropgz: ""
+      clusterType: overlay-no-kube-proxy-up
       clusterName: "overlaye2e"
+      vmSize: Standard_B2ms
+      k8sVersion: ""
+      windowsOsSku: ""
+      dependsOn: 'containerize'
```
vipul-21, line 343: Show we depend on `validate1` so that we know the change is upto date with master

.pipelines/singletenancy/aks/e2e-cluster-template.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - task: AzureCLI@1
+            inputs:
+              azureSubscription: $(AZURE_TEST_AGENT_SERVICE_CONNECTION)
+              scriptLocation: "inlineScript"
+              scriptType: "bash"
+              addSpnToEnvironment: true
+              inlineScript: |
+                set -e
+                mkdir -p ~/.kube/
+                # echo "Create AKS cluster"
```
vipul-21, line 30: We can remove these.

.pipelines/pipeline.yaml
```diff
@@ -322,24 +322,39 @@ stages:
       displayName: Cilium
       pipelineBuildImage: "$(BUILD_IMAGE)"
       testDropgz: ""
+      clusterType: swift-no-kube-proxy-up
       clusterName: "ciliume2e"
+      vmSize: Standard_B2ms
+      k8sVersion: ""
```
vipul-21, line 328: If we are sending anything in the parameters, we can leave them out ?

vipul-21:
There is a caveat to this approach in terms of usability. We no longer can simply rerun the E2E test in the pipeline(in case it fails due to transient issue, which it does a lot with goldpinger and stuff). To restart we will need to rerun the cluster step associated with that E2E test. 
Maybe we pull out the delete cluster part as well ? ( won't be testing on new setup on retry though)

.pipelines/pipeline.yaml
jpayne3506: If we don't pass anything the container template fails. The ${{ parameters.* }} calls don't result in NULL as expected if the corresponding parameter doesn't pass anything into the make "cluster-type" command. 

A work around for this to prevent clutter would be to have various make "cluster-type" command based on os, arch, ect. That would prevent us having to pass in k8sVersion ect. when t is not being used.

.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
```diff
@@ -3,14 +3,34 @@ parameters:
...
   osSku: ""
+  vmSize: ""
+  k8sVersion: ""
+  windowsOsSku: ""
+  dependsOn: ""
+
+
+
 stages:
+  - template: ../aks/e2e-cluster-template.yaml
```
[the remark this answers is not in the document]
jpayne3506: Considering that it is using/creating aks clusters from the aks/makefile, wouldn't its current location be more appropriate? Maybe a name change from e2e-cluster-template -> cluster-template to ensure it is known that it is available for reuse, but tied to aks/makefile?

.pipelines/singletenancy/aks-swift/e2e-step-template.yaml
```diff
@@ -30,26 +30,17 @@ steps:
...
-      go get github.com/onsi/ginkgo/ginkgo
-      go get github.com/onsi/gomega/...
-      go install github.com/onsi/ginkgo/ginkgo@latest
-      go install sigs.k8s.io/kubetest2@latest
-      go install sigs.k8s.io/kubetest2/kubetest2-noop@latest
-      go install sigs.k8s.io/kubetest2/kubetest2-tester-ginkgo@latest
-      wget https://storage.googleapis.com/pub/gsutil.tar.gz
-      tar xfz gsutil.tar.gz
-      sudo mv gsutil /usr/local/bin
+        make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(make revision)
```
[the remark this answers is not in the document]
jpayne3506: Removing OSSKU=${{ parameters.osSku }} did not affect anything as the default is Ubuntu, and with the cluster-template it was removed as it was adding nothing of value to the cluster name. This was the only E2E stage that used that parameter.

.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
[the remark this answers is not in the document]
camrynl: I think a name change + folder name change could help make it more clear

test/internal/k8sutils/utils.go
camrynl, line 1: What's the reason for changing to 90 retries instead of 30 here?

jpayne3506:
> won't be testing on new setup on retry though)

In another PR we could test using ["counter" ](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#counter)as a mechanism for people that are re-running tests often within the pipeline. Would need fine tuning, but I believe that deletion should be a priority as the speedups/optimizations are worthless if the cluster sits around for any additional time (possibly 15 days) than needed.

vipul-21:
> > won't be testing on new setup on retry though)
> 
> In another PR we could test using ["counter" ](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#counter)as a mechanism for people that are re-running tests often within the pipeline. Would need fine tuning, but I believe that deletion should be a priority as the speedups/optimizations are worthless if the cluster sits around for any additional time (possibly 15 days) than needed.

I agree completely that delete is priority 1. To clarify by “pulling out delete cluster part” i meant removing from respective e2e step template and into a separate template like creat cluster. (And not move out of pipeline)

test/internal/k8sutils/utils.go
jpayne3506: Wanted to give the same amount of time coverage for the retry. If I kept it at 30 then we went from 15 minutes of time to 5 minutes.

The retries are giving us updates for scale testing, decreasing the delay provided a ~15 second speed up due to it capturing the replica status closer to when it completed on each scale up/down.

.pipelines/pipeline.yaml
jpayne3506: As long as the respective stages are dependent on the same stages that validate1 and validate2 are then we know those respective stages are up to date with master. The cluster creation stage matches with validate1 and E2E matches with validate2 with the additional dependency that cluster creation for the respective E2E stage is complete.

.pipelines/pipeline.yaml
```diff
@@ -322,24 +322,38 @@ stages:
       displayName: Cilium
       pipelineBuildImage: "$(BUILD_IMAGE)"
       testDropgz: ""
+      clusterType: swift-no-kube-proxy-up
```
tamilmani1989, line 325: can clusterType be `cilium-podsubet-up`

.pipelines/pipeline.yaml
```diff
@@ -322,24 +322,38 @@ stages:
...
+      windowsOsSku: ""
+      dependsOn: 'containerize'
 
   - template: singletenancy/cilium-overlay/cilium-overlay-e2e-job-template.yaml
     parameters:
       name: "cilium_overlay_cilium_e2e"
       displayName: Cilium on AKS Overlay
       pipelineBuildImage: "$(BUILD_IMAGE)"
       testDropgz: ""
+      clusterType: overlay-no-kube-proxy-up
```
tamilmani1989, line 338: cilium-overlay-up

.pipelines/pipeline.yaml
```diff
@@ -322,24 +322,38 @@ stages:
...
+      windowsOsSku: ""
+      dependsOn: 'containerize'
 
   - template: singletenancy/aks-swift/e2e-job-template.yaml
     parameters:
       name: "aks_swift_e2e"
       displayName: AKS Swift Ubuntu
       pipelineBuildImage: "$(BUILD_IMAGE)"
       testDropgz: ""
+      clusterType: byocni-up
```
tamilmani1989, line 351: swift-podsubnet-up

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
```diff
@@ -158,6 +149,9 @@ steps:
   - script: |
       echo "validate pod IP assignment and check systemd-networkd restart"
       kubectl get pod -owide -A
+      # Deleting echo-external-node deployment until cilium version matches TODO. https://github.com/cilium/cilium-cli/issues/67 is addressing the change.
```
tamilmani1989, line 152: what this test is for?

jpayne3506: It tries to echo (curl in this case) to an external, non-cilium, node. The deployment is stuck in pending due to its node selector looking for "cilium.io/no-schedule": "true". We cannot remove or skip the test at this point.

.pipelines/pipeline.yaml
jpayne3506: This parameter `clusterType` is calling the respective commands from [hack/aks/makefile](https://github.com/Azure/azure-container-networking/blob/cc251ca58a0fe6009a7c80c167ca52887169dce3/hack/aks/Makefile#L156). We can change them to whatever we want.

.pipelines/pipeline.yaml
camrynl: We can go ahead and rename the first two to `cilium-podsubnet-up` and `cilium-overlay-up`. I think we could keep this one as `byocni-up`, because `swift-podsubnet-up` might be misleading. We have an existing target called `swift-up` which brings up a swift cluster. 

If you don't want to keep `byocni-up` -- Maybe we could go with `swift-byocni-up`, to clearly indicate that it's byocni but just need to make sure it matches in all other places.

.pipelines/pipeline.yaml
camrynl: let's change the clustertypes to match the new hackfile commands

.pipelines/pipeline.yaml
tamilmani1989: `swift-byocni-up` makes sense

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
tamilmani1989: for now its fine, we should contribute to cilium add skip check to this test case @vipul-21

.pipelines/cni/k8s-e2e/k8s-e2e-job-template.yaml
```diff
@@ -20,11 +20,11 @@ jobs:
           addSpnToEnvironment: true
           inlineScript: |
             set -e
-            make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(make revision)
```
camrynl, line 23: we will need to keep the `-$(make revision)` in the cluster name in order to create unique clusters in each pipeline run. we won't be able to run parallel builds without it.

camrynl (changes requested):

.pipelines/cni/k8s-e2e/k8s-e2e-job-template.yaml
jpayne3506: It is passed in within the aks/e2e-job template.

.pipelines/submodules-pipeline.yaml
```diff
@@ -244,32 +245,80 @@ stages:
...
+    jobs:
+      - job: delete
+        displayName: Delete Cluster
+        pool:
+          name: "$(BUILD_POOL_NAME_DEFAULT)"
+        strategy:
+          matrix:
+            cilium_e2e:
+              name: cilium_e2e
+              clusterName: 'ciliume2e'
```
camrynl, line 299: why is the clusterName for delete and the test stages different? 'submodciliume2e' vs 'ciliume2e', same for the overlay and swfit stages

jpayne3506: Good catch. Just needed to be changed.

.pipelines/submodules-pipeline.yaml
```diff
@@ -244,32 +245,80 @@ stages:
...
+    displayName: Delete Clusters
+    condition: always()
+    dependsOn:
+      - setup
+      - "aks_swift_e2e"
+      - "cilium_e2e"
+      - "cilium_overlay_cilium_e2e"
+    variables:
+      commitID: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.commitID'] ]
+    jobs:
```
vipul-21, line 290: So if one of the stages fails, and we restart the cluster creation just for that step, this will run for every stage and will try to delete the already deleted cluster ?

.pipelines/submodules-pipeline.yaml
```diff
@@ -244,32 +245,80 @@ stages:
...
+        strategy:
+          matrix:
+            cilium_e2e:
+              name: cilium_e2e
+              clusterName: 'submodciliume2e'
+              region: $(REGION_AKS_CLUSTER_TEST)
+            cilium_overlay_cilium_e2e:
+              name: cilium_overlay_cilium_e2e
+              clusterName: 'submodoverlaye2e'
+              region: $(REGION_OVERLAY_CLUSTER_TEST)
```
vipul-21, line 304: What's the default value for the `REGION_AKS_CLUSTER_TEST` and `REGION_OVERLAY_CLUSTER_TEST` ? we are using different regions due to quota reason ?

.pipelines/pipeline.yaml
```diff
@@ -313,34 +314,49 @@ stages:
...
 
   - template: singletenancy/aks-swift/e2e-job-template.yaml
     parameters:
       name: "aks_swift_e2e"
       displayName: AKS Swift Ubuntu
-      pipelineBuildImage: "$(BUILD_IMAGE)"
       testDropgz: ""
+      clusterType: swift-byocni-up
       clusterName: "swifte2e"
-      osSku: "Ubuntu"
```
vipul-21, line 343: This  might be needed as we add windows swift cluster tests as well ?

.pipelines/singletenancy/aks-swift/e2e-step-template.yaml
```diff
@@ -31,25 +30,17 @@ steps:
       addSpnToEnvironment: true
       inlineScript: |
         set -e
-        mkdir -p ~/.kube/
-        echo "Create AKS cluster"
-        make -C ./hack/aks azcfg AZCLI=az REGION=$(REGION_AKS_CLUSTER_TEST)
-        make -C ./hack/aks byocni-up AZCLI=az REGION=$(REGION_AKS_CLUSTER_TEST) SUB=$(SUB_AZURE_NETWORK_AGENT_TEST) CLUSTER=${{ parameters.clusterName }}-${{ parameters.osSku }}-$(make revision) OSSKU=${{ parameters.osSku }}
```
vipul-21, line 37: This might be a typo in the original code for variable: `OS_SKU` . Is this needed ? by default it will be `Linux`(which satisfies our pipeline but was this added to address the future windows e2e stages for swift ?

jpayne3506: Referring to documentation, the only values that can be accepted are for az aks nodepool:`AzureLinux, CBLMariner, Mariner, Ubuntu, Windows2019, Windows2022` for az aks create:`AzureLinux, CBLMariner, Mariner, Ubuntu` Considering we are only supporting Ubuntu and Windows2022 I don't see the need to keep it or use it as we already have flow setup in case of Windows2022 being passed.

.pipelines/submodules-pipeline.yaml
jpayne3506: Originally I thought that Overlay was something that may be region specific on our subscription. Since that is not the case, quota for `REGION_OVERLAY_CLUSTER_TEST : westcentralus` is actually quite low for everything. I will change this to default to `REGION_AKS_CLUSTER_TEST : eastus`

.pipelines/submodules-pipeline.yaml
jpayne3506: Yes.

camrynl: Will the delete fail if it tries deleting the already deleted cluster?

.pipelines/cni/k8s-e2e/k8s-e2e-job-template.yaml
camrynl: thanks for clarifying

.pipelines/submodules-pipeline.yaml
jpayne3506: No, the delete succeeds as the makefile exit code does not get read if set -e is not present.

hack/aks/Makefile
```diff
@@ -99,97 +99,98 @@ overlay-byocni-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster
...
 	@$(MAKE) set-kubeconf
 
-swift-up: rg-up swift-net-up ## Bring up a SWIFT AzCNI cluster
+cilium-podsubnet-up: rg-up swift-net-up ## Bring up a SWIFT BYO CNI cluster without kube-proxy for Cilium
 	$(AZCLI) aks create -n $(CLUSTER) -g $(GROUP) -l $(REGION) \
 		--node-count $(NODE_COUNT) \
 		--node-vm-size $(VM_SIZE) \
 		--load-balancer-sku basic \
-		--network-plugin azure \
+		--network-plugin none \
```
vipul-21, line 188: Why the name has `cilium` since BYO CNI can have any CNI ?

hack/aks/Makefile
```diff
@@ -99,97 +99,98 @@ overlay-byocni-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster
...
 		--vnet-subnet-id /subscriptions/$(SUB)/resourceGroups/$(GROUP)/providers/Microsoft.Network/virtualNetworks/$(VNET)/subnets/nodenet \
 		--pod-subnet-id /subscriptions/$(SUB)/resourceGroups/$(GROUP)/providers/Microsoft.Network/virtualNetworks/$(VNET)/subnets/podnet \
 		--no-ssh-key \
-		--os-sku $(OS_SKU) \
-		--kube-proxy-config ./kube-proxy.json \
 		--yes
 	@$(MAKE) set-kubeconf
 
-swift-cilium-up: rg-up swift-net-up ## Bring up a SWIFT Cilium cluster
+cilium-overlay-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster without kube-proxy for Cilium
```
vipul-21, line 169: Same comment as cilium podsubnet up

hack/aks/Makefile
vipul-21: Also why the comment without kubeproxy ? We are providing the file `kubeproxy.json`

camrynl: The `kubeproxy.json` file is set to disable kube proxy when the cluster comes up. If we do not include the parameter, the cluster will have kube-proxy installed. As for the naming, @tamilmani1989 suggested in earlier comments to rename as such. We typically use these byocni clusters for cilium in most cases, so the name should be okay

jpayne3506: > Also why the comment without kubeproxy ? We are providing the file `kubeproxy.json`

`"enabled": false` is preventing the daemonset from being deployed. https://learn.microsoft.com/en-us/azure/aks/configure-kube-proxy#configurable-options

camrynl (dismissed):

.pipelines/singletenancy/aks/e2e-step-template.yaml
```diff
@@ -1,14 +1,8 @@
 parameters:
   name: ""
-  clusterType: ""
   clusterName: ""
-  nodeCount: ""
```
vipul-21, line 5: Since we removed these from step template, we should remove passing these attributes from jon template too ?

camrynl (approved):

vipul-21 (approved):

rbtr:
I just discovered this change because I was not getting the clusters I was expecting out of this Makefile anymore. 

@jpayne3506 it is _incredibly_ confusing and misleading to have `overlay-cilium-up` and `cilium-overlay-up`, and `swift-cilium-up` and `cilium-podsubnet-up`. 
Especially since these new targets are not even Cilium (`--network-plugin none`, no `--network-dataplane` arg) and are simply BYO CNI with the addition of the kube-proxy config. 
You probably could have modified the existing BYO CNI targets to add the kube-proxy flag if an env var was set. Short of that, these must be clearly named as "byocni-kubeproxy" flavors of the existing targets. Lmk if you want to fix this, otherwise we can revert it.
cc @camrynl and @vipul-21 since you approved

vipul-21:
@rbtr we did raise the same questions during the PR review: https://github.com/Azure/azure-container-networking/pull/2077#discussion_r1287485256, but afaik @tamilmani1989 wanted to use the `cilium` name.

rbtr:
`cilium` in the target name is lying, and you should push back on that. ([TM was okay with "byocni".](https://github.com/Azure/azure-container-networking/pull/2077#discussion_r1276958668))
 
But again, we might not have needed separate make targets for this anyway since the difference is just the addition of the kube-proxy flag. This is the kind of modification that can be done off of an env var

camrynl:
@jpayne3506 is going to update the makefile with more explicit names for each target

rbtr:
thanks!

PR #2079, opened 2023-07-26
CNS - consider multiple IPs for a pod in reconciliation after restart

ramiro-gamarra:
**Reason for Change**:

Reconciliation after CNS restart is currently not handling dual stack scenarios correctly. There's 2 bugs:
- the information fetched from CNI currently filters out ipv6 addresses.
- reconciliation after restart processes multiple NCs separately, but pod info in CNS (or at least one representation of it) is not able to be patched after first insertion, meaning, a pod will be reconciliated with ipv4 first, and then skip the update with ipv6.

This PR modifies the main reconciliation function such that it takes multiple NCs as input, and processes all IPs for a single pod key (interface or name+namespace, depending on reconcile flow) at the same time, such that both ipv4 and ipv6 can be inserted once as `DesiredIPAddresses`.

### DIFF
--- a/cns/cnireconciler/podinfoprovider.go
+++ b/cns/cnireconciler/podinfoprovider.go
@@ -51,19 +51,25 @@ func newCNIPodInfoProvider(exec exec.Interface) (cns.PodInfoByIPProvider, error)
 }
 
 // cniStateToPodInfoByIP converts an AzureCNIState dumped from a CNI exec
-// into a PodInfo map, using the first endpoint IP as the key in the map.
+// into a PodInfo map, using the endpoint IPs as keys in the map.
+// for pods with multiple IPs (such as in dualstack cases), this means multiple keys in the map
+// will point to the same pod information.
 func cniStateToPodInfoByIP(state *api.AzureCNIState) (map[string]cns.PodInfo, error) {
 	podInfoByIP := map[string]cns.PodInfo{}
 	for _, endpoint := range state.ContainerInterfaces {
-		if _, ok := podInfoByIP[endpoint.IPAddresses[0].IP.String()]; ok {
-			return nil, errors.Wrap(cns.ErrDuplicateIP, endpoint.IPAddresses[0].IP.String())
+		for _, epIP := range endpoint.IPAddresses {
+			ipKey := epIP.IP.String()
+			if _, ok := podInfoByIP[ipKey]; ok {
+				return nil, errors.Wrap(cns.ErrDuplicateIP, ipKey)
+			}
+
+			podInfoByIP[ipKey] = cns.NewPodInfo(
+				endpoint.ContainerID,
+				endpoint.PodEndpointId,
+				endpoint.PodName,
+				endpoint.PodNamespace,
+			)
 		}
-		podInfoByIP[endpoint.IPAddresses[0].IP.String()] = cns.NewPodInfo(
-			endpoint.ContainerID,
-			endpoint.PodEndpointId,
-			endpoint.PodName,
-			endpoint.PodNamespace,
-		)
 	}
 	return podInfoByIP, nil
 }
--- a/cns/cnireconciler/podinfoprovider.go
+++ b/cns/cnireconciler/podinfoprovider.go
@@ -58,17 +58,14 @@ func cniStateToPodInfoByIP(state *api.AzureCNIState) (map[string]cns.PodInfo, er
 	podInfoByIP := map[string]cns.PodInfo{}
 	for _, endpoint := range state.ContainerInterfaces {
 		for _, epIP := range endpoint.IPAddresses {
+			podInfo := cns.NewPodInfo(endpoint.ContainerID, endpoint.PodEndpointId, endpoint.PodName, endpoint.PodNamespace)
+
 			ipKey := epIP.IP.String()
-			if _, ok := podInfoByIP[ipKey]; ok {
-				return nil, errors.Wrap(cns.ErrDuplicateIP, ipKey)
+			if prevPodInfo, ok := podInfoByIP[ipKey]; ok {
+				return nil, errors.Wrapf(cns.ErrDuplicateIP, "duplicate ip %s found for different pods: pod: %+v, pod: %+v", ipKey, podInfo, prevPodInfo)
 			}
 
-			podInfoByIP[ipKey] = cns.NewPodInfo(
-				endpoint.ContainerID,
-				endpoint.PodEndpointId,
-				endpoint.PodName,
-				endpoint.PodNamespace,
-			)
+			podInfoByIP[ipKey] = podInfo
 		}
 	}
 	return podInfoByIP, nil
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
@@ -316,37 +317,43 @@ func (service *HTTPRestService) ReconcileIPAMState(ncReqs []*cns.CreateNetworkCo
 	// to pod IPs indexed by interface:
 	//
 	//   {
-	//     "aaa-eth0": podIPs{IPs:["10.0.0.1","fbac:::::1"]}
+	//     "aaa-eth0": podIPs{IPs:["10.0.0.1","fe80::1"]}
 	//   }
 	//
 	// such that we can iterate over pod interfaces, and assign all IPs for it at once.
-	ifaceToPodIPs := newInterfaceToPodIPsMap(podInfoByIP)
+	ifaceToPodIPs, err := newInterfaceToPodIPsMap(podInfoByIP)
+	if err != nil {
+		logger.Errorf("could not transform pods indexed by IP address to pod IPs indexed by interface: %v", err)
+		return types.UnexpectedError
+	}
 
 	for interfaceID, podIPs := range ifaceToPodIPs {
 		var (
 			desiredIPs []string
 			ncIDs      []string
 		)
-		for _, ip := range podIPs.IPs {
-			if ncReq, ok := allSecIPsIdx[ip]; ok {
+
+		for _, ip := range []net.IP{podIPs.v4IP, podIPs.v6IP} {
+			if ncReq, ok := allSecIPsIdx[ip.String()]; ok {
 				logger.Printf("secondary ip %s is assigned to pod %+v, ncId: %s ncVersion: %s", ip, podIPs, ncReq.NetworkContainerid, ncReq.Version)
-				desiredIPs = append(desiredIPs, ip)
+				desiredIPs = append(desiredIPs, ip.String())
 				ncIDs = append(ncIDs, ncReq.NetworkContainerid)
 			} else {
-				// todo: confirm if this is a valid scenario for system pods
+				// it might still be possible to see host networking pods here (where ips are not from ncs) if we are restoring using the kube podinfo provider
+				// todo: once kube podinfo provider reconcile flow is removed, this line will not be necessary/should be removed.
 				logger.Errorf("ip %s assigned to pod %+v but not found in any nc", ip, podIPs)
 			}
+		}
+
+		if len(desiredIPs) == 0 {
+			// this may happen for system pods
+			continue
 		}
 
 		jsonContext, err := podIPs.OrchestratorContext()
 		if err != nil {
 			logger.Errorf("Failed to marshal KubernetesPodInfo, error: %v", err)
 			return types.UnexpectedError
-		}
-
-		if len(desiredIPs) == 0 {
-			// this may happen for system pods
-			continue
 		}
 
 		ipconfigsRequest := cns.IPConfigsRequest{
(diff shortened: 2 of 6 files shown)


cns/restserver/internalapi.go
```diff
@@ -274,45 +274,62 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
 	return len(programmedNCs), nil
 }
 
-// This API will be called by CNS RequestController on CRD update.
-func (service *HTTPRestService) ReconcileNCState(ncRequest *cns.CreateNetworkContainerRequest, podInfoByIP map[string]cns.PodInfo, nnc *v1alpha.NodeNetworkConfig) types.ResponseCode {
-	logger.Printf("Reconciling NC state with CreateNCRequest: [%v], PodInfo [%+v], NNC: [%+v]", ncRequest, podInfoByIP, nnc)
-	// check if ncRequest is null, then return as there is no CRD state yet
-	if ncRequest == nil {
+func (service *HTTPRestService) ReconcileNCState(ncReqs []*cns.CreateNetworkContainerRequest, podInfoByIP map[string]cns.PodInfo, nnc *v1alpha.NodeNetworkConfig) types.ResponseCode {
```
ashvindeodhar, line 277: IMO we should rename this func to indicate it's reconciling the CNS ipam state. The previous implementation was reconciling one NC at a time but that's not the case anymore

cns/restserver/internalapi.go
```diff
@@ -274,45 +274,62 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
+	logger.Printf("Reconciling NC state with nc requests: [%+v], PodInfo [%+v], NNC: [%+v]", ncReqs, podInfoByIP, nnc)
```
ashvindeodhar, line 278: it would be good to update the log indicating this is reconciling the CNS ipam state

cns/restserver/internalapi.go
```diff
@@ -274,45 +274,62 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
+			desiredIPs []string
+			ncIDs      []string
+		)
+		for _, ip := range podInfo.IPs {
+			if ncReq, ok := allSecIPsIdx[ip]; ok {
+				logger.Printf("secondary ip %s is assigned to pod %+v, ncId: %s ncVersion: %s", ip, podInfo, ncReq.NetworkContainerid, ncReq.Version)
+				desiredIPs = append(desiredIPs, ip)
+				ncIDs = append(ncIDs, ncReq.NetworkContainerid)
+			} else {
+				// this is an IP CNI knows about but DNC doesn't. what to do?
```
ashvindeodhar, line 313: IMO as we discussed yesterday, we should log the error but continue marking other IPs as assigned if any. Ideally we should raise an alert as this may break datapath for the pod that has IP that CNS / vfp doesn't know about

cns/restserver/internalapi.go
```diff
@@ -325,6 +342,24 @@ func (service *HTTPRestService) ReconcileNCState(ncRequest *cns.CreateNetworkCon
 	return 0
 }
 
+func podInfoByIPToPodInfoExtByInterfaceID(podInfoByIP map[string]cns.PodInfo) map[string]podInfoExt {
```
ashvindeodhar, line 345: can you add comment explaining the purpose of this func?

cns/restserver/internalapi.go
```diff
@@ -325,6 +342,24 @@ func (service *HTTPRestService) ReconcileNCState(ncRequest *cns.CreateNetworkCon
...
+
+		xt := out[k]
+		xt.PodInfo = podInfo
+		xt.IPs = append(xt.IPs, ip)
+		out[k] = xt
+	}
+	return out
+}
+
+type podInfoExt struct {
```
ashvindeodhar, line 358: podIPInfo?

cns/restserver/internalapi.go
ashvindeodhar, line 345: is it possible to rename this func to something else that indicates the func?
something like GeneratePodInterfaceToPodIPsInfoMap ?

cns/service/main.go
```diff
@@ -1059,17 +1068,15 @@ func reconcileInitialCNSState(ctx context.Context, cli nodeNetworkConfigGetter,
...
-		// Call cnsclient init cns passing those two things.
-		if err := restserver.ResponseCodeToError(ncReconciler.ReconcileNCState(ncRequest, podInfoByIP, nnc)); err != nil {
-			return errors.Wrap(err, "failed to reconcile NC state")
-		}
+		ncReqs = append(ncReqs, ncRequest)
 	}
+
+	// Call cnsclient init cns passing those two things.
+	if err := restserver.ResponseCodeToError(ncReconciler.ReconcileNCState(ncReqs, podInfoByIP, nnc)); err != nil {
+		return errors.Wrap(err, "failed to reconcile NC state")
```
ashvindeodhar, line 1077: aligning with my other comments, this should either say NNC state or CNS ipam state

cns/restserver/internalapi_test.go
```diff
@@ -73,16 +73,20 @@ func TestReconcileNCStatePrimaryIPChangeShouldFail(t *testing.T) {
...
 	}
 
-	// now try to reconcile the state where the NC primary IP has changed
-	resp := svc.ReconcileNCState(&cns.CreateNetworkContainerRequest{
-		NetworkContainerid: ncID,
-		IPConfiguration: cns.IPConfiguration{
-			IPSubnet: cns.IPSubnet{
-				IPAddress:    "10.0.2.0", // note this IP has changed
-				PrefixLength: 24,
+	ncReqs := []*cns.CreateNetworkContainerRequest{
```
ashvindeodhar, line 76: can you add UT for testing this reconcile logic if it;s completely missing?

ashvindeodhar (changes requested):

cns/restserver/internalapi.go
ramiro-gamarra: Pod "IP" info feels a little incomplete because it holds multiple IPs. I wanted to just extend `cns.PodInfo` at first but that turns out to be a much larger change (not sure why it's an interface, it has a single implementation as far as I could see). This is just a small type that extends `cns.PodInfo`, thus the suffix.

cns/restserver/internalapi.go
ramiro-gamarra: yeah, it's a mouthful 😅. it's basically a "merge concat" operation (merging by same interface, concatenating all IPs for the same interface), but maybe a constructor like "newInterfaceToPodIPsInfoMap"?

cns/restserver/internalapi_test.go
```diff
@@ -537,13 +545,13 @@ func TestReconcileNCWithSystemPods(t *testing.T) {
 	req := generateNetworkContainerRequest(secondaryIPConfigs, uuid.New().String(), "-1")
 
 	expectedAssignedPods := make(map[string]cns.PodInfo)
-	expectedAssignedPods["10.0.0.6"] = cns.NewPodInfo("", "", "customerpod1", "PodNS1")
+	expectedAssignedPods["10.0.0.6"] = cns.NewPodInfo("some-guid-1", "abc-eth0", "customerpod1", "PodNS1")
 
 	// Allocate non-vnet IP for system  pod
-	expectedAssignedPods["192.168.0.1"] = cns.NewPodInfo("", "", "systempod", "kube-system")
+	expectedAssignedPods["192.168.0.1"] = cns.NewPodInfo("some-guid-2", "def-eth0", "systempod", "kube-system")
```
ramiro-gamarra, line 551: @neaggarwMS @rbtr is this a valid scenario? when would we see a host networking pod in the reconcile flow of CNS?

cns/restserver/internalapi.go
```diff
@@ -274,45 +274,91 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
+			desiredIPs []string
+			ncIDs      []string
+		)
+		for _, ip := range podIPs.IPs {
+			if ncReq, ok := allSecIPsIdx[ip]; ok {
+				logger.Printf("secondary ip %s is assigned to pod %+v, ncId: %s ncVersion: %s", ip, podIPs, ncReq.NetworkContainerid, ncReq.Version)
+				desiredIPs = append(desiredIPs, ip)
+				ncIDs = append(ncIDs, ncReq.NetworkContainerid)
+			} else {
+				// todo: confirm if this is a valid scenario for system pods
```
rbtr, line 336: if this should say "hostnetwork" pods, those are not stored in the CNI statefile so we won't learn about them from the CNI podinfo provider (they might bypass the CNI entirely since the pod is created in the host network namespace). It might still be possible to see this if we are restoring using the kube podinfo provider, so we note that this special case handling is safe to delete once that is removed.

cns/restserver/internalapi_test.go
rbtr: > Was this perhaps simulating the reconcile from kubernetes directly? 

yes, the podinfo is constructed using the first two fields or the last two fields depending on if we are rebuilding state from k8s or from CNI.

> Will we still support that scenario?

we have to keep it until we roll the cni-installer init container out to swift, then we will [remove it completely](https://github.com/Azure/azure-container-networking/pull/2087)

cns/restserver/internalapi_test.go
```diff
@@ -475,6 +485,188 @@ func TestReconcileNCWithExistingState(t *testing.T) {
 	validateNCStateAfterReconcile(t, req, expectedNcCount+1, expectedAssignedPods, nil)
 }
 
+func TestReconcileCNSIPAMWithDualStackPods(t *testing.T) {
```
ashvindeodhar, line 488: can we add test case for single stack v4 and with system pods IPs?

rbtr: i know some of the code/comments say "system pods" but that doesn't actually mean anything. do you mean hostnetwork pods?

ashvindeodhar: yes you are right. I meant hostnetworking pods

ashvindeodhar (dismissed):

cns/cnireconciler/podinfoprovider.go
```diff
@@ -51,19 +51,25 @@ func newCNIPodInfoProvider(exec exec.Interface) (cns.PodInfoByIPProvider, error)
...
+// will point to the same pod information.
 func cniStateToPodInfoByIP(state *api.AzureCNIState) (map[string]cns.PodInfo, error) {
 	podInfoByIP := map[string]cns.PodInfo{}
 	for _, endpoint := range state.ContainerInterfaces {
-		if _, ok := podInfoByIP[endpoint.IPAddresses[0].IP.String()]; ok {
-			return nil, errors.Wrap(cns.ErrDuplicateIP, endpoint.IPAddresses[0].IP.String())
+		for _, epIP := range endpoint.IPAddresses {
+			ipKey := epIP.IP.String()
+			if _, ok := podInfoByIP[ipKey]; ok {
+				return nil, errors.Wrap(cns.ErrDuplicateIP, ipKey)
```
thatmattlong, line 63: useful to add `podInfoByIP[ipKey]` to the error message too?

cns/restserver/internalapi.go
```diff
@@ -274,57 +275,156 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
+
+	for ipStr, podInfo := range podInfoByIP {
+		id := podInfo.Key()
+
+		ips, ok := podKeyToPodIPs[id]
+		if !ok {
+			ips.PodInfo = podInfo
+		}
+
+		ip := net.ParseIP(ipStr)
```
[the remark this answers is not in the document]
ramiro-gamarra: The `To4` and `To16` methods handle this case (they return nil if it's not valid for each family), such that if `ip` is not a valid v4 or v6 IP, we hit the default case.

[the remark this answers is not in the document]
ramiro-gamarra: On a further read, even if it's handled, it may not be quickly apparent since the error case is not the first one. I'll move the nil case check closer to `ParseIP`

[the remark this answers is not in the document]
thatmattlong: oh interesting, I was expecting the call to To4 or To6 to throw the NPE, but you're right, those funcs handle the case of nil receiver, so I'm good either way

cns/restserver/internalapi.go
```diff
@@ -274,57 +275,156 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
-				OrchestratorContext: jsonContext,
-				InfraContainerID:    podInfo.InfraContainerID(),
-				PodInterfaceID:      podInfo.InterfaceID(),
+	for podKey, podIPs := range podKeyToPodIPs {
+		var (
+			desiredIPs []string
+			ncIDs      []string
+		)
+
+		for _, ip := range []net.IP{podIPs.v4IP, podIPs.v6IP} {
```
rbtr, line 336: the way this range is built, i think we are wasting some cycles when the pods are not dualstack. i think we want a podIPs.IPs slice (or two, that we append(...)) so that we are only iterating over populated values

rbtr (changes requested):

rbtr (approved):

PR #2085, opened 2023-07-27
ci: cilium scale pipeline

camrynl:
**Reason for Change**:

This PR creates a new cilium scale pipeline to test a cluster at 1000 nodes. It adopts NPM's existing scale test to create nginx deployments and services, and it uses the existing connectivity test. It uses apachebench and netperf to test the throughput and gathers cpu memory usage. The results are published as pipeline artifacts. We will reuse the same cluster for all builds, and at the end of each run the cluster is scaled back down to 5 nodes.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,228 @@
...
+              scriptType: "bash"
+              addSpnToEnvironment: true
+              inlineScript: |
+                set -ex
+                az aks get-credentials --resource-group ${CLUSTER} --name ${CLUSTER}
+                echo "Scaling to 1000 nodes"
+                az aks nodepool scale --name nodepool1 --cluster-name ${CLUSTER} --resource-group ${CLUSTER} --node-count 1000
+            name: "ScaleUp1000"
+            displayName: "Scale up to 1000 Nodes"
+            timeoutInMinutes: 0
```
jpayne3506, line 55: Same here.

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,228 @@
...
+                kubectl get node 
+                kubectl get pod -A
+            name: "UpdateCiliumVersion"
+            displayName: "Update Cilium Version"
+  - stage: scale_up_cluster
+    displayName: "Scale Up Cluster"
+    jobs:
+      - job: scale_up1000
+        pool:
+          name: "$(BUILD_POOL_ACN)"
```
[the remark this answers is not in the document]
camrynl: it's the same, I can update the variable for consistency

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,228 @@
...
+        steps:
+          - task: AzureCLI@1
+            inputs:
+              azureSubscription: $(TEST_SUB_SERVICE_CONNECTION)
+              scriptLocation: "inlineScript"
+              scriptType: "bash"
+              addSpnToEnvironment: true
+              inlineScript: |
+                az aks get-credentials --resource-group ${CLUSTER} --name ${CLUSTER}
```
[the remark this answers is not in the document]
camrynl: We only have one cluster that we use for the scale tests, and as of now we expect to use the same one each time. Multiple builds is not a concern for now. I set the cluster name as default for the var so we don't have to change it at each run, but it is mutable in case we ever do need to use a different cluster.

.pipelines/cni/cilium/cilium-scale-test.yaml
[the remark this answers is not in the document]
jpayne3506: Awesome! it will allow us to use variable groups more effectively in the future.

.pipelines/cni/scale-test-templates/apache.yaml
```diff
@@ -0,0 +1,18 @@
...
+    matchLabels:
+      app: apachebench
+  template:
+    metadata:
+      labels:
+        app: apachebench
+    spec:
+      containers:
+      - name: ubuntu-tools
+        image: tamilmani1989/ubuntu18-tools
```
vipul-21, line 16: Can we see if we can a substitute for this image ?
Also @tamilmani1989 should we consider ubuntu 22 image in that case ?

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,226 @@
...
+          name: "$(BUILD_POOL_NAME_DEFAULT)"
+        steps:
+        - task: AzureCLI@1
+          inputs:
+            azureSubscription: $(TEST_SUB_SERVICE_CONNECTION)
+            scriptLocation: "inlineScript"
+            scriptType: "bash"
+            addSpnToEnvironment: true
+            inlineScript: |
+              echo "scale deployment and test network policies"
```
vipul-21, line 220: Seems incorrect based on what stage is doing.

.pipelines/cni/scale-test-templates/apache.yaml
camrynl: Last week we confirmed that we can continue to use this image

.pipelines/cni/cilium/cilium-scale-test.yaml
camrynl: thanks, good catch

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,226 @@
...
+              addSpnToEnvironment: true
+              inlineScript: |
+                set -ex
+                az aks get-credentials --resource-group ${CLUSTER} --name ${CLUSTER}
+                echo "Scaling to 5 nodes"
+                vmss_name=$(az vmss list -g MC_${CLUSTER}_${CLUSTER}_$(LOCATION) --query "[].name" -o tsv)
+                make -C ./hack/aks scale-vmss AZCLI=az CLUSTER=${CLUSTER} REGION=$(LOCATION) VMSS_NAME=$vmss_name NODE_COUNT=5
+                kubectl get node
+            name: "ScaleDown"
+            displayName: "Scale down to 5 Nodes"
```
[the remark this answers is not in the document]
camrynl: sure, I can write in a condition based on that stage

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,226 @@
...
+            inputs:
+              azureSubscription: $(TEST_SUB_SERVICE_CONNECTION)
+              scriptLocation: "inlineScript"
+              scriptType: "bash"
+              addSpnToEnvironment: true
+              inlineScript: |
+                set -ex
+                az aks get-credentials --resource-group ${CLUSTER} --name ${CLUSTER}
+                echo "Scaling to 1000 nodes"
+                az aks nodepool scale --name nodepool1 --cluster-name ${CLUSTER} --resource-group ${CLUSTER} --node-count 1000
```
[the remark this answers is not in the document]
camrynl: sounds good, the node count and other test input values have been updated as pipeline variables that can be altered

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,228 @@
...
+                echo "Deploy apachebench pod and run test"
+                cd hack/manifests
+                kubectl apply -f apache.yaml
+                echo "wait for pod to become ready"
+                sleep 5s
+                kubectl get pod -owide
+                mkdir results
+                cd results
+                AB_POD=$(kubectl get pod -l app=apachebench | grep apachebench | awk '{print $1}')
+                kubectl exec -it $AB_POD -- ab -n ${AB_REQUESTS} -c ${AB_CONCURRENCY} -r http://real-svc-00001.scale-test/ >> "ab_${AB_REQUESTS}requests_${NUM_REAL_REPLICAS_LB}kpods.log"
```
jpayne3506, line 163: Any reason to not use rollout for this deployment? It would ensure that there are no timing issues for exec call.

camrynl: updated to use rollout status instead :)

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,228 @@
...
+                kubectl rollout status deployment apachebench --timeout=30s
+                kubectl get pod -owide
+                mkdir results
+                cd results
+                AB_POD=$(kubectl get pod -l app=apachebench | grep apachebench | awk '{print $1}')
+                kubectl exec -it $AB_POD -- ab -n ${AB_REQUESTS} -c ${AB_CONCURRENCY} -r http://real-svc-00001.scale-test/ >> "ab_${AB_REQUESTS}requests_${NUM_REAL_REPLICAS_LB}kpods.log"
+                echo "running k top node" 
+                kubectl top node >> "node_cpu_and_mem.log"
+                echo "running k top pod"
+                kubectl top pod -A | grep cilium >> "pod_cpu_and_mem.log"
```
tamilmani1989, line 167: lets get pod cpu and mem at each step before and after network policy scale and after lb scale

.pipelines/cni/scale-test-templates/apache.yaml
tamilmani1989: this is fine.. this is just pod image

tamilmani1989 (approved):

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,367 @@
...
+                envsubst '${CILIUM_VERSION_TAG},${CILIUM_IMAGE_REGISTRY}' < test/integration/manifests/cilium/daemonset.yaml | kubectl apply -f -
+                envsubst '${CILIUM_VERSION_TAG},${CILIUM_IMAGE_REGISTRY}' < test/integration/manifests/cilium/deployment.yaml | kubectl apply -f -
+                kubectl apply -f test/integration/manifests/cilium/cilium-agent
+                kubectl apply -f test/integration/manifests/cilium/cilium-operator
+                echo "Keep CNS version up to date, grabbing pipeline parameter"
+                CNS_IMAGE=${CNS_IMAGE}
+                sed -i '/containers:/{n;n;s/\(image\).*/\1: '"${CNS_IMAGE//\//\\/}"'/}' test/integration/manifests/cns/daemonset.yaml 
+                kubectl apply -f test/integration/manifests/cns/daemonset.yaml 
+                vmss_name=$(az vmss list -g MC_${CLUSTER}_${CLUSTER}_$(LOCATION) --query "[].name" -o tsv)
+                make -C ./hack/aks restart-vmss AZCLI=az CLUSTER=${CLUSTER} REGION=$(LOCATION) VMSS_NAME=$vmss_name
```
vipul-21, line 34: nit: This would restart the first vmss only. I think we don't have more than one right now, but would be nice to add a loop around the restart vmss. Something like: https://github.com/Azure/azure-container-networking/blob/master/.pipelines/singletenancy/aks/e2e-step-template.yaml#L68-L70

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,367 @@
...
+                ARTIFACT_DIR=$(Build.ArtifactStagingDirectory)/before_netpol_cpu_and_mem/
+                echo $ARTIFACT_DIR
+                sudo rm -rf $ARTIFACT_DIR 
+                sudo mkdir $ARTIFACT_DIR
+                cd ..
+                sudo cp ./before_netpol_cpu_and_mem/* $ARTIFACT_DIR
+                echo "scale deployment and to prep for network policies test"
+                cd test/scale
+                chmod +x test-scale.sh
+                ./test-scale.sh --max-kwok-pods-per-node=0 --num-kwok-deployments=0 --num-kwok-replicas=0 --max-real-pods-per-node=${REAL_PODS_PER_NODE_NETPOL} --num-real-deployments=${NUM_REAL_DEPLOYMENTS_NETPOL} --num-real-replicas=${NUM_REAL_REPLICAS_NETPOL} --num-network-policies=${APPLIED_NETPOL} --num-unapplied-network-policies=${UNAPPLIED_NETPOL} --num-unique-labels-per-pod=0 --num-unique-labels-per-deployment=5 --num-shared-labels-per-pod=3 --num-real-services=${NUM_REAL_SVC_NETPOL} --delete-labels
```
vipul-21, line 112: Would recommend getting the cpu/mem usage after the test-scale too if possible. Since it captures a snapshot of cpu/mem usage of the cluster, it would be nice to have one just for the scaling part( along with "after connectivity" test)
Might be valid for other tests in the pipeline. Just want to make sure we have data on the specific runs(scale pods/svcs etc)

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,367 @@
...
+              artifactName: after_lb_cpu_and_mem
+              pathtoPublish: "$(Build.ArtifactStagingDirectory)/after_lb_cpu_and_mem"
+            condition: always()
+            name: "PublishResults2"
+            displayName: "Result Artifacts"
+  - stage: netperf_tests
+    displayName: "Run netperf tests"
+    jobs:
+      - job: netperf
+        timeoutInMinutes: 120
```
jpayne3506, line 270: What is the expected runtime?

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -0,0 +1,367 @@
...
+            condition: always()
+            name: "PublishResults"
+            displayName: "Result Artifacts"
+  - stage: test_network_policies_connectivity
+    displayName: "Test Network Policies"
+    jobs:
+      - job: network_policies
+        pool:
+          name: "$(BUILD_POOL_NAME_DEFAULT)"
+        timeoutInMinutes: 120
```
jpayne3506, line 128: Same here

test/scale/label-nodes.sh
```diff
@@ -17,11 +18,17 @@ fi
...
 do
+    node_count=$((node_count + 1))
+    echo $node_count
     echo "Current : $node"
     node_name="${node##*/}"
     echo "Apply label to the node"
     kubectl label node $node_name connectivity-test=true
     kubectl label node $node_name scale-test=true
+    if [ $node_count -lt 3 ]; then
+        kubectl label node $node_name netperf=true
```
vipul-21, line 29: Since we are using the same cluster for scale-test, are we removing these labels ? This is a one time thing ?

camrynl: The nodes that get removed after scaling will need to be relabeled after the 1000 are reprovisioned

.pipelines/cni/cilium/cilium-scale-test.yaml
camrynl: about 1 hour, but I adjusted the timeouts in case it goes a little over, or if we want to change the test input with different cases

.pipelines/cni/cilium/cilium-scale-test.yaml
camrynl: this test completes quickly, so I removed the timeout here

hack/scripts/netperf.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+# find the nodes with netperf pods and assign test vars
+node_found=0
+for node in $(kubectl get nodes -o name);
```
vipul-21, line 4: We can limit this result to three if we use the label selector `netperf=true`?

vipul-21 (approved):

jpayne3506 (approved):
Great work Camryn! Lgtm.

hack/scripts/netperf.sh
camrynl: yes, in label-nodes.sh I cut off the `netperf=true` labeling once 3 nodes have it

PR #2086, opened 2023-07-27
ci: Add k8s network conformance tests to makefile

jpayne3506:
**Reason for Change**:

Allows for users to test cluster network conformance from the command line and give a single command to test network conformance in future CI.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests
-  [x] relevant PR labels added

**Notes**:
Example commands:
 - make -C ./hack/aks/ test-k8se2e CLUSTER=yourClusterName GROUP=resourceGroup OS=windows TYPE=basic
 - make -C ./hack/aks/ test-k8se2e CLUSTER=yourClusterName GROUP=resourceGroup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ azure-ipam: azure-ipam-binary azure-ipam-archive
 
 revision: ## print the current git revision
 	@echo $(REVISION)
-	
+
 version: ## prints the root version
 	@echo $(ACN_VERSION)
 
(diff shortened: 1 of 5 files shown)


camrynl:
it might be better to add the k8s tests to the main makefile where other test targets live since we primarily use the hackfile for provisioning clusters

Makefile
```diff
@@ -762,6 +762,15 @@ test-azure-ipam: ## run the unit test for azure-ipam
 kind:
 	kind create cluster --config ./test/kind/kind.yaml
 
+test-k8se2e: test-k8se2e-build test-k8se2e-only ## Alias to run build and test
+
+test-k8se2e-build: ## Build k8s e2e test suite
+	cd hack/scripts && bash ./k8se2e.sh $(GROUP) $(CLUSTER)
```
vipul-21, line 768: Will this work? Given we don't have variables `Group` or `Cluster` in the file ?

jpayne3506: Yes, these scripts were heavily reliant on those arguments being present in command to begin with. Having a default did not provide any additional benefit.

camrynl (approved):

PR #2075, opened 2023-07-25
feat: [dropgz] Dropgz for windows

vipul-21:
Dropgz support for windows.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [x] relevant PR labels added

**Notes**:

PR #2087, opened 2023-07-27
Remove Kubernetes PodInfoProvider and default to CNI initialization

rbtr:
**Reason for Change**:

Removes the (long deprecated) Kubernetes reconcile branch from the CNS PodInfo initialization flow. The default is now the CNI reconcile flow, with the preference to restore from CNS own managed endpoint state if that is enabled.

Slightly repackages these components to be clearer:
both are now in pkg `podprovider` instead of `cnireconciler`
CNI Pod info provider -> `podprovider/cni`
CNS endpoint state pod infor provider moved from cnireconciler -> `podprovider/cns`

Since CNI reconciliation is becoming the default, the version check can be dropped as there is no longer another option to fall back to. We simply attempt it, and verbosely error.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Because we don't _always_ install the CNI with CNS yet, this is not safe to release until all AKS variants are using the CNI installer init container.


thatmattlong:
shall we wait to merge this until podsubnet CNI is installed with dropgz? my fear is that someone will forget that it's not ready to release until after that, and we'll break some clusters

rbtr:
+do-no-merge +work-in-progress to be more clear that this shouldn't be merged until cni installer is default everywhere

rbtr:
depends on #2075

rbtr:
obsoletes/removes changes introduced in #2078

PR #2090, opened 2023-07-27
fix: overlay IPAM not reporting version

### LINKED ISSUE #2081: overlay IPAM not reporting version
**What happened:** 
when on newly deployed node, checking ipam version shows unknown

image

using
mcr.microsoft.com/containernetworking/cni-dropgz:v0.0.9
mcr.microsoft.com/containernetworking/azure-cns:v1.5.5

maybe @rbtr?

nddq:
**Reason for Change**:

**Issue Fixed**:

Fixes #2081

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr:
@nddq what's the output of `./azure-ipam -v` with this change?

azure-ipam/internal/buildversion/buildversion.go
```diff
@@ -1,6 +1,6 @@
-package buildinfo
+package buildversion
```
thatmattlong, line 1: why renaming and moving these around?

Makefile
```diff
@@ -173,7 +173,7 @@ zapai-version: ## prints the zapai version
 
 # Build the delegated IPAM plugin binary.
 azure-ipam-binary:
-	cd $(AZURE_IPAM_DIR) && CGO_ENABLED=0 go build -v -o $(AZURE_IPAM_BUILD_DIR)/azure-ipam$(EXE_EXT) -ldflags "-X main.version=$(AZURE_IPAM_VERSION)" -gcflags="-dwarflocationlists=true"
+	cd $(AZURE_IPAM_DIR) && CGO_ENABLED=0 go build -v -o $(AZURE_IPAM_BUILD_DIR)/azure-ipam$(EXE_EXT) -ldflags "-X buildversion.BuildVersion=$(AZURE_IPAM_VERSION)" -gcflags="-dwarflocationlists=true"
```
thatmattlong, line 176: could just change this to `buildinfo.Version` without renaming/moving the files

nddq: I was trying to match and see if [this](https://github.com/containernetworking/plugins/blob/v1.2.0/pkg/utils/buildversion/buildversion.go) work but no luck. I did tried populating `buildversion.BuildVersion` directly before but no luck either. The only way I managed to get the version printed out is to have the `version` variable inside `main` then populate it through `main.Version` like CNI and CNS does right now. Is this the only way or I'm I missing something?

nddq:
> @nddq what's the output of `./azure-ipam -v` with this change?

image
This is what I got currently. Should azure-ipam have a release tag like v1.xxx+commit like CNI and CNS or is this intended? I remember that we discussed about wanting tagging each plugin and release them individually last summer so it would be great if we could follow up on this was well.

Makefile
rbtr: what you're missing is that `main.version` works in _our_ linker args because `main` is the top level package and we have the `version` var defined there. to inject a value more deeply, we need to fully-qualify the path. 

read the comments in the file you linked and check the mentioned "linker script", you'll see how we need to pass the [full path to the `buildversion` package and variable](https://github.com/containernetworking/plugins/blob/v1.2.0/scripts/release.sh#L10C46-L10C129) for the injection to work. we should not need to change anything except the "-X [...]=$(AZURE_IPAM_VERSION)". our Makefile will automatically set that runtime env to either the bare tag if the current revision is tagged, or the latest tag + dirty hash if the current rev is not tagged.

azure-ipam/internal/buildversion/buildversion.go
rbtr: this file should be deleted entirely

Makefile
nddq: yup, working now

azure-ipam/main.go
```diff
@@ -4,6 +4,7 @@ import (
 	"log"
 	"os"
 
+	"github.com/Azure/azure-container-networking/azure-ipam/internal/buildinfo"
```
rbtr, line 7: should we just use the cni buildinfo pkg instead?

nddq: I don't think azure cni has a buildinfo package currently, unless you are talking about a different cni?

rbtr: `github.com/containernetworking/plugins/pkg/utils/buildversion.BuildVersion`

nddq: hmm when I replaced with that pkg, the version became an empty string

rbtr (approved):

matmerr (approved):

PR #2101, opened 2023-08-02
perf: 🚀 Increase CNI Lock Timeout to 30 seconds for Linux AKS.

behzad-mir:
perf: Performance Improvements 🚀

Increasing the 

**Reason for Change**:
The Current Azure CNI lock timeout is set to 10 second while The CNS IPRequest Timeout is 15 sec. 
We are just increasing the CNI timeout value to 30 seconds so it would not be lesser than CNS API timeout. 

The relevent load tests has been done under Swift posdubnet and overlay to make sure that the increase is not braking anything.


behzad-mir:
> Just curious, if the CNS IPRequest timeout is 15 sec, why don't we have 15 secs or maybe to accommodate the request fully 20 secs or so. How did we come up with the 30 seconds ?

We did not have an accurate calculation for this tbh. Previously it was 10 and we just wanted to increase it a bit in case CNS get stuck. the performance result for 10 and 30 were not that much different!

store/json.go
```diff
@@ -23,7 +23,7 @@ const (
 	LockExtension = ".lock"
 
 	// DefaultLockTimeout - lock timeout in milliseconds
-	DefaultLockTimeout        = 10000 * time.Millisecond
+	DefaultLockTimeout        = 30000 * time.Millisecond
```
tamilmani1989, line 26: can we name this `DefaultLockTimeoutLinux`?

behzad-mir: I will add another var name "DefaultLockTimeoutLinux", since 
The DefaultLockTimeout is being used by telemetry

tamilmani1989 (approved):

vipul-21 (approved):

PR #2102, opened 2023-08-03
feat: 🌈 StatelessCNI: Adding getEndpoint and UpdateEndpoint API to CNS

behzad-mir:
**Reason for Change**:

In the stateless CNI design two new APIs on CNS side with respect of endpoint state management is needed.
GetEndpoint() is used to return the state of a given endpointId
UpdateEndpoint() is used to update the HNSId and VethName field within the endpoint state. 

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -36,6 +36,7 @@ const (
 	PathDebugRestData                        = "/debug/restdata"
 	NumberOfCPUCores                         = NumberOfCPUCoresPath
 	NMAgentSupportedAPIs                     = NmAgentSupportedApisPath
+	EndpointsPath                            = EndpointPath
 )
 
 // NetworkContainer Prefixes
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -36,7 +36,7 @@ const (
 	PathDebugRestData                        = "/debug/restdata"
 	NumberOfCPUCores                         = NumberOfCPUCoresPath
 	NMAgentSupportedAPIs                     = NmAgentSupportedApisPath
-	EndpointsPath                            = EndpointPath
+	EndpointApi                              = EndpointPath
 )
 
 // NetworkContainer Prefixes
--- a/cns/api.go
+++ b/cns/api.go
@@ -34,7 +34,7 @@ const (
 	NmAgentSupportedApisPath      = "/network/nmagentsupportedapis"
 	V1Prefix                      = "/v0.1"
 	V2Prefix                      = "/v0.2"
-	EndpointPath                  = "/network/endpoints"
+	EndpointPath                  = "/network/endpoints/"
 )
 
 // HTTPService describes the min API interface that every service should have.
@@ -351,12 +351,6 @@ type GetHomeAzResponse struct {
 
 // UpdateEndpoint describes request to update the state.
 type EndpointRequest struct {
-	EndpointID    string `json:"endpointID"`
 	HnsEndpointID string `json:"hnsEndpointID"`
 	HostVethName  string `json:"hostVethName"`
-}
-
-// UpdateEndpointResponse describes response to the UpdateEndpoint request.
-type UpdateEndpointResponse struct {
-	Response Response `json:"response"`
 }
--- a/cns/api.go
+++ b/cns/api.go
@@ -34,6 +34,7 @@ const (
 	NmAgentSupportedApisPath      = "/network/nmagentsupportedapis"
 	V1Prefix                      = "/v0.1"
 	V2Prefix                      = "/v0.2"
+	EndpointPath                  = "/network/endpoints"
 )
 
 // HTTPService describes the min API interface that every service should have.
@@ -346,4 +347,21 @@ type HomeAzResponse struct {
 type GetHomeAzResponse struct {
 	Response       Response       `json:"response"`
 	HomeAzResponse HomeAzResponse `json:"homeAzResponse"`
+}
+
+// UpdateEndpoint describes request to update the state.
+type UpdateEndpoint struct {
+	EndpointID string `json:"EndpointID"`
+	HnsID      string `json:"HnsID"`
+	VethName   string `json:"VethName"`
+}
+
+// UpdateEndpointResponse describes response to the UpdateEndpoint request.
+type UpdateEndpointResponse struct {
+	Response Response `json:"Response"`
+}
+
+// EndpointID describes request to get the state.
+type GetEndpoint struct {
+	EndpointID string `json:"EndpointID"`
 }
--- a/cns/api.go
+++ b/cns/api.go
@@ -351,9 +351,9 @@ type GetHomeAzResponse struct {
 
 // UpdateEndpoint describes request to update the state.
 type EndpointRequest struct {
-	EndpointID string `json:"endpointID"`
-	HnsID      string `json:"hnsID"`
-	VethName   string `json:"vethName"`
+	EndpointID    string `json:"endpointID"`
+	HnsEndpointID string `json:"hnsEndpointID"`
+	HostVethName  string `json:"hostVethName"`
 }
 
 // UpdateEndpointResponse describes response to the UpdateEndpoint request.
--- a/cns/api.go
+++ b/cns/api.go
@@ -350,18 +350,13 @@ type GetHomeAzResponse struct {
 }
 
 // UpdateEndpoint describes request to update the state.
-type UpdateEndpoint struct {
-	EndpointID string `json:"EndpointID"`
-	HnsID      string `json:"HnsID"`
-	VethName   string `json:"VethName"`
+type EndpointRequest struct {
+	EndpointID string `json:"endpointID"`
+	HnsID      string `json:"hnsID"`
+	VethName   string `json:"vethName"`
 }
 
 // UpdateEndpointResponse describes response to the UpdateEndpoint request.
 type UpdateEndpointResponse struct {
-	Response Response `json:"Response"`
-}
-
-// EndpointID describes request to get the state.
-type GetEndpoint struct {
-	EndpointID string `json:"EndpointID"`
+	Response Response `json:"response"`
 }
--- a/cns/api.go
+++ b/cns/api.go
@@ -349,7 +349,7 @@ type GetHomeAzResponse struct {
 	HomeAzResponse HomeAzResponse `json:"homeAzResponse"`
 }
 
-// UpdateEndpoint describes request to update the state.
+// Used by EndpointHandler API to update endpoint state.
 type EndpointRequest struct {
 	HnsEndpointID string `json:"hnsEndpointID"`
 	HostVethName  string `json:"hostVethName"`
--- a/cns/client/client.go
+++ b/cns/client/client.go
@@ -1047,7 +1047,7 @@ func (c *Client) GetEndpoint(ctx context.Context, endpointID string) (*restserve
 	var response restserver.GetEndpointResponse
 	err = json.NewDecoder(res.Body).Decode(&response)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to decode IPConfigResponse")
+		return nil, errors.Wrap(err, "failed to decode GetEndpointResponse")
 	}
 
 	if response.Response.ReturnCode != 0 {
@@ -1092,7 +1092,7 @@ func (c *Client) UpdateEndpoint(ctx context.Context, endpointID, hnsID, vethName
 	err = json.NewDecoder(res.Body).Decode(&response)
 
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to decode IPConfigResponse")
+		return nil, errors.Wrap(err, "failed to decode GetEndpointResponse")
 	}
 
 	if response.ReturnCode != 0 {
--- a/cns/client/client.go
+++ b/cns/client/client.go
@@ -45,6 +45,7 @@ var clientPaths = []string{
 	cns.DeleteNetworkContainer,
 	cns.NetworkContainersURLPath,
 	cns.GetHomeAz,
+	cns.EndpointsPath,
 }
 
 type do interface {
@@ -1020,4 +1021,90 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error) 
 	}
 
 	return &getHomeAzResponse, nil
+}
+
+// GetEndpoint calls the GetEndpoint in CNS
+func (c *Client) GetEndpoint(ctx context.Context, endpointID string) (*restserver.GetEndpointResponse, error) {
+	var err error
+	getEndpoint := cns.GetEndpoint{
+		EndpointID: endpointID,
+	}
+	var body bytes.Buffer
+	err = json.NewEncoder(&body).Encode(getEndpoint)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to encode getEndpoint")
+	}
+
+	u := c.routes[cns.EndpointsPath]
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), &body)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to build request")
+	}
+	req.Header.Set(headerContentType, contentTypeJSON)
+	res, err := c.client.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "http request failed")
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		return nil, errors.Errorf("http response %d", res.StatusCode)
+	}
+
+	var response restserver.GetEndpointResponse
+	err = json.NewDecoder(res.Body).Decode(&response)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to decode IPConfigResponse")
+	}
+
+	if response.Response.ReturnCode != 0 {
+		return nil, errors.New(response.Response.Message)
+	}
+
+	return &response, nil
+}
+
+// UpdateEndpoint calls the UpdateEndpoint in CNS
+func (c *Client) UpdateEndpoint(ctx context.Context, endpointID, hnsID, vethName string) (*cns.UpdateEndpointResponse, error) {
+	var err error
+	updateEndpoint := cns.UpdateEndpoint{
+		EndpointID: endpointID,
+		HnsID:      hnsID,
+		VethName:   vethName,
+	}
+	var body bytes.Buffer
+	err = json.NewEncoder(&body).Encode(updateEndpoint)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to encode updateEndpoint")
+	}
+
+	u := c.routes[cns.EndpointsPath]
+	req, err := http.NewRequestWithContext(ctx, http.MethodPatch, u.String(), &body)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to build request")
+	}
+	req.Header.Set(headerContentType, contentTypeJSON)
+	res, err := c.client.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "http request failed with error from server")
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		return nil, errors.Errorf("http response %d", res.StatusCode)
+	}
+
+	var response cns.UpdateEndpointResponse
+	err = json.NewDecoder(res.Body).Decode(&response)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to decode IPConfigResponse")
+	}
+
+	if response.Response.ReturnCode != 0 {
+		return nil, errors.New(response.Response.Message)
+	}
+
+	return &response, nil
 }
--- a/cns/client/client.go
+++ b/cns/client/client.go
@@ -1023,19 +1023,12 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error) 
 	return &getHomeAzResponse, nil
 }
 
-// GetEndpoint calls the EndpointHandlerAPI in CNS to retrive the state of a given EndpointID
+// GetEndpoint calls the EndpointHandlerAPI in CNS to retrieve the state of a given EndpointID
 func (c *Client) GetEndpoint(ctx context.Context, endpointID string) (*restserver.GetEndpointResponse, error) {
-	getEndpoint := cns.EndpointRequest{
-		EndpointID: endpointID,
-	}
-	var body bytes.Buffer
-
-	if err := json.NewEncoder(&body).Encode(getEndpoint); err != nil {
-		return nil, errors.Wrap(err, "failed to encode getEndpoint")
-	}
-
+	// build the request
 	u := c.routes[cns.EndpointsPath]
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), &body)
+	uString := u.String() + endpointID
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, uString, http.NoBody)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to build request")
 	}
@@ -1064,10 +1057,11 @@ func (c *Client) GetEndpoint(ctx context.Context, endpointID string) (*restserve
 	return &response, nil
 }
 
-// UpdateEndpoint calls the EndpointHandlerAPI in CNS to update the state of a given EndpointID with either HNSEndpointID or HostVethName
-func (c *Client) UpdateEndpoint(ctx context.Context, endpointID, hnsID, vethName string) (*cns.UpdateEndpointResponse, error) {
+// UpdateEndpoint calls the EndpointHandlerAPI in CNS
+// to update the state of a given EndpointID with either HNSEndpointID or HostVethName
+func (c *Client) UpdateEndpoint(ctx context.Context, endpointID, hnsID, vethName string) (*cns.Response, error) {
+	// build the request
 	updateEndpoint := cns.EndpointRequest{
-		EndpointID:    endpointID,
 		HnsEndpointID: hnsID,
 		HostVethName:  vethName,
 	}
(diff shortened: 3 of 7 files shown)


cns/client/client.go
```diff
@@ -1009,3 +1010,89 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error)
 
 	return &getHomeAzResponse, nil
 }
+
+// GetEndpoint calls the GetEndpoint in CNS
+func (c *Client) GetEndpoint(ctx context.Context, endpointID string) (*restserver.GetEndpointResponse, error) {
+	var err error
```
jaer-tsun, line 1016: can remove this and update L#1021 with err :=

cns/client/client.go
```diff
@@ -1009,3 +1010,89 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error)
...
+	getEndpoint := cns.GetEndpoint{
+		EndpointID: endpointID,
+	}
+	var body bytes.Buffer
+	err = json.NewEncoder(&body).Encode(getEndpoint)
```
jaer-tsun, line 1021: after removing L#1016, could look like

if err := json.NewEncoder(&body).Encode(getEndpoint); err != nil {...}

cns/api.go
```diff
@@ -347,3 +348,20 @@ type GetHomeAzResponse struct {
 	Response       Response       `json:"response"`
 	HomeAzResponse HomeAzResponse `json:"homeAzResponse"`
 }
+
+// UpdateEndpoint describes request to update the state.
+type UpdateEndpoint struct {
```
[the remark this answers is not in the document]
AzureAhai: Good suggestion. I'll name it EndpointRequest or something like that.

cns/client/client.go
```diff
@@ -1009,3 +1010,89 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error)
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to encode getEndpoint")
+	}
+
+	u := c.routes[cns.EndpointsPath]
```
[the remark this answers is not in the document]
AzureAhai: you mean line L#38 on NetworkContainerContract look like path = EndpointPath ?

cns/api.go
```diff
@@ -347,3 +348,20 @@ type GetHomeAzResponse struct {
...
+	EndpointID string `json:"EndpointID"`
+	HnsID      string `json:"HnsID"`
+	VethName   string `json:"VethName"`
```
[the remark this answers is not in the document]
AzureAhai: Thanks. I was just copy pasting the field there and forgot to make them lowercase.

cns/restserver/restserver.go
```diff
@@ -95,6 +95,8 @@ type EndpointInfo struct {
 	PodName       string
 	PodNamespace  string
 	IfnameToIPMap map[string]*IPInfo // key : interface name, value : IPInfo
+	HnsID         string
```
AzureAhai, line 98: @tamilmani1989 @rbtr 
I added the field to the existing endpoint state on CNS which is being used for Cilium as well. This should not break anything. But do we want to test cilium cni with this change on CNS?

cns/restserver/ipam.go
```diff
@@ -891,3 +893,181 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+
+	// Read any persisted state.
+	err := service.store.Read(storeKey, &service.state)
+	if err != nil {
+		if errors.Is(err, store.ErrKeyNotFound) {
+			// Nothing to restore.
+			logger.Printf("[GetEndpointState]  No state to restore.\n")
+		} else {
+			logger.Errorf("[GetEndpointState]  Failed to restore state, err:%v", err)
+			service.store.Remove()
```
tamilmani1989, line 975: does that cleanup entire statefile? why do we remove store on failure?

cns/restserver/ipam.go
```diff
@@ -891,3 +893,181 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+				ReturnCode: types.InvalidRequest,
+				Message:    fmt.Sprintf("[Azure CNS] updateEndpoint failed with error: %s", err.Error()),
+			},
+		}
+		err = service.Listener.Encode(w, &response)
+		logger.Response(service.Name, response, response.Response.ReturnCode, err)
+		return
+	}
+
+	err = service.UpdateEndpointHelper(req.EndpointID, req.HnsID, req.VethName)
```
tamilmani1989, line 1024: can the argument be passing reference to `req` ?

cns/restserver/ipam.go
```diff
@@ -891,3 +893,181 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+	}
+}
+
+// UpdateEndpointHelper updates the state of the given endpointId with HNSId or VethName
+func (service *HTTPRestService) UpdateEndpointHelper(endpointID, hnsID, vethName string) error {
+	if service.EndpointStateStore == nil {
+		return ErrStoreEmpty
+	}
+	logger.Printf("[UpdateEndpointState] Updating endpoint state for infra container %s", endpointID)
+	if service.Options[common.OptManageEndpointState] == true {
```
tamilmani1989, line 1052: can it be ` if service.Options[common.OptManageEndpointState] {`? and update for Get call as well

also do we need to check if service.Options is empty or nil ?

cns/restserver/ipam.go
```diff
@@ -891,3 +893,181 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+			logger.Printf("[UpdateEndpointState] Found existing endpoint state for infra container %s", endpointID)
+			if hnsID != "" {
+				service.EndpointState[endpointID].HnsID = hnsID
+				logger.Errorf("[UpdateEndpointState] update the endpoint %s with HNSID  %s", endpointID, hnsID)
+				return nil
+			} else if vethName != "" {
+				service.EndpointState[endpointID].VethName = vethName
+				logger.Errorf("[UpdateEndpointState] update the endpoint %s with vethName  %s", endpointID, vethName)
+				return nil
+			}
```
tamilmani1989, line 1063: convert to if

cns/restserver/ipam.go
```diff
@@ -891,3 +893,181 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+// UpdateEndpointHelper updates the state of the given endpointId with HNSId or VethName
+func (service *HTTPRestService) UpdateEndpointHelper(endpointID, hnsID, vethName string) error {
+	if service.EndpointStateStore == nil {
+		return ErrStoreEmpty
+	}
+	logger.Printf("[UpdateEndpointState] Updating endpoint state for infra container %s", endpointID)
+	if service.Options[common.OptManageEndpointState] == true {
+		if endpointInfo, ok := service.EndpointState[endpointID]; ok {
+			logger.Printf("[UpdateEndpointState] Found existing endpoint state for infra container %s", endpointID)
+			if hnsID != "" {
```
tamilmani1989, line 1055: can we add a check if both empty and add a warning log?

cns/restserver/restserver.go
tamilmani1989: yes

tamilmani1989 (changes requested):

cns/restserver/ipam.go
```diff
@@ -891,3 +893,181 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+			service.store.Remove()
+		}
+
+		return nil, errors.Wrap(err, "[GetEndpointState]  Failed to retrieve state")
+	}
+
+	logger.Printf("[GetEndpointState]  retrieve state, %+v\n", service.state)
+
+	if service.Options[common.OptManageEndpointState] == true {
+		err := service.EndpointStateStore.Read(EndpointStoreKey, &service.EndpointState)
```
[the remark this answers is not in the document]
behzad-mir: I forgot to tag you and @rbtr for this. I was able to get the state from both store and endpointStateStore. Wanted to make sure I'm not using the wrong one.

cns/restserver/ipam.go
behzad-mir: Forgot to delete this. It was something I added during my tests

cns/restserver/ipam.go
behzad-mir: I put else if in here to make it impossible for the client to put both vethname and hnsID in a statefile since we only have one of them for Linux and Windows.

cns/restserver/ipam.go
behzad-mir: oops. Forgot this!

cns/restserver/ipam.go
behzad-mir: It is a true false field provide in cns conf yaml file. We don't have a nil scenario AFIK

cns/restserver/ipam.go
```diff
@@ -891,3 +893,181 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
 	}
 	return nil
 }
+
+// EndpointHandlerAPI forwards the endpoint related APIs to GetEndpointHandler or UpdateEndpointHandler based on the http method
+func (service *HTTPRestService) EndpointHandlerAPI(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] EndpointHandlerAPI")
```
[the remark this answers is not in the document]
behzad-mir: The logger does not have any println func!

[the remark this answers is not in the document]
nddq: oops, didn't know this was called from our logger haha

cns/client/client.go
```diff
@@ -1009,3 +1010,87 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error)
...
+		HostVethName:  vethName,
+	}
+	var body bytes.Buffer
+
+	if err := json.NewEncoder(&body).Encode(updateEndpoint); err != nil {
+		return nil, errors.Wrap(err, "failed to encode updateEndpoint")
+	}
+
+	u := c.routes[cns.EndpointsPath]
+	req, err := http.NewRequestWithContext(ctx, http.MethodPatch, u.String(), &body)
```
rbtr, line 1069: PATCH is not the right verb here. [[MDN]](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH):
> A PATCH request is considered a set of instructions on how to modify a resource. Contrast this with [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT); which is a complete representation of a resource.  

This is a full representation of the resource, so it's either PUT or POST, depending on exactly what the desired effect is (wrt idempotency):
> The difference between PUT and [POST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST) is that **_PUT is idempotent_**: calling it once or several times successively has the same effect (that is no side effect), whereas successive identical [POST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST) requests may have additional effects, akin to placing an order several times.

Here, I think you want to PUT.

rbtr: Additionally, building on the RESTful semantics context I previously mentioned, PUTing a resource in this collection should probably be done at:
`PUT /network/endpoints/{{endpointID}}`

with the non-"uniquely identifying" content of the object in the Body.

cns/restserver/ipam.go
```diff
@@ -891,3 +893,170 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+				Message:    fmt.Sprintf("[Azure CNS] GetEndpoint failed with error: %s", err.Error()),
+			},
+		}
+		err = service.Listener.Encode(w, &response)
+		logger.Response(service.Name, response, response.Response.ReturnCode, err)
+	}
+	response := GetEndpointResponse{
+		Response: Response{
+			ReturnCode: types.Success,
+			Message:    "[Azure CNS] GetEndpoint retruned successfully",
```
rbtr, line 947:
```suggestion
			Message:    "[Azure CNS] GetEndpoint returned successfully",
```

cns/restserver/ipam.go
```diff
@@ -891,3 +893,170 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+func (service *HTTPRestService) GetEndpointHelper(endpointID string) (*EndpointInfo, error) {
+	logger.Printf("[GetEndpointState] Get endpoint state for infra container %s", endpointID)
+
+	// Skip if a store is not provided.
+	if service.EndpointStateStore == nil {
+		logger.Printf("[GetEndpointState]  store not initialized.")
+		return nil, ErrStoreEmpty
+	}
+
+	if service.Options[common.OptManageEndpointState] == true {
```
rbtr, line 966: Instead of checking this every time this method is called, what if we made the decision _when we registered the handler_?
We can let the HTTP server 404 if the endpoint is unhandled, but if it is, we know we're running in the correct configuration. No need to keep checking it everywhere.

cns/restserver/ipam.go
```diff
@@ -891,3 +893,170 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+		err = service.Listener.Encode(w, &response)
+		logger.Response(service.Name, response, response.Response.ReturnCode, err)
+	}
+	response := cns.UpdateEndpointResponse{
+		Response: cns.Response{
+			ReturnCode: types.Success,
+			Message:    "[Azure CNS] updateEndpoint retruned successfully",
+		},
+	}
+	err = service.Listener.Encode(w, &response)
```
rbtr, line 1025: Consider that we can communicate success without sending an explicit "Success!" response body just by using the HTTP code `200 OK` (or sometimes a [`201 Created`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201)?).
If we're sending anything back in the body, it should be the updated object.

cns/restserver/ipam.go
```diff
@@ -891,3 +893,170 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+// UpdateEndpointHelper updates the state of the given endpointId with HNSId or VethName
+func (service *HTTPRestService) UpdateEndpointHelper(req cns.EndpointRequest) error {
+	if service.EndpointStateStore == nil {
+		return ErrStoreEmpty
+	}
+	logger.Printf("[UpdateEndpointState] Updating endpoint state for infra container %s", req.EndpointID)
+	if service.Options[common.OptManageEndpointState] == true {
+		if endpointInfo, ok := service.EndpointState[req.EndpointID]; ok {
+			logger.Printf("[UpdateEndpointState] Found existing endpoint state for infra container %s", req.EndpointID)
+			if req.HostVethName == "" && req.HnsEndpointID != "" {
```
rbtr, line 1038: Validate the request is well-formed immediately, before anything else.

cns/restserver/restserver.go
```diff
@@ -119,6 +121,12 @@ type Response struct {
 	Message    string
 }
 
+// GetEndpointResponse describes response from the The GetEndpoint API.
+type GetEndpointResponse struct {
+	Response     Response     `json:"response"`
+	EndpointInfo EndpointInfo `json:"endpointInfo"`
+}
+
```
rbtr, line 129: I think this type is redundant, the response to a GET on `/network/endpoints/<endpoint>` should just be the `<endpoint>`. Leverage headers and response codes!

cns/restserver/restserver.go
```diff
@@ -289,6 +297,7 @@ func (service *HTTPRestService) Init(config *common.ServiceConfig) error {
 	listener.AddHandler(cns.V2Prefix+cns.DeleteHostNCApipaEndpointPath, service.deleteHostNCApipaEndpoint)
 	listener.AddHandler(cns.V2Prefix+cns.NmAgentSupportedApisPath, service.nmAgentSupportedApisHandler)
 	listener.AddHandler(cns.V2Prefix+cns.GetHomeAz, service.getHomeAz)
+	listener.AddHandler(cns.V2Prefix+cns.EndpointPath, service.EndpointHandlerAPI)
```
rbtr, line 300: I doubt we need this in both the v1 and v2 subpaths, especially since you're writing the client for it in the same change

cns/restserver/ipam.go
```diff
@@ -891,3 +893,170 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+				Message:    fmt.Sprintf("[Azure CNS] updateEndpoint failed with error: %s", err.Error()),
+			},
+		}
+		err = service.Listener.Encode(w, &response)
+		logger.Response(service.Name, response, response.Response.ReturnCode, err)
+	}
+	response := cns.UpdateEndpointResponse{
+		Response: cns.Response{
+			ReturnCode: types.Success,
+			Message:    "[Azure CNS] updateEndpoint retruned successfully",
```
rbtr, line 1022:
```suggestion
			Message:    "[Azure CNS] updateEndpoint returned successfully",
```

rbtr (changes requested):
some nitpicks and misspellings that are nbd, but i'm blocking on the HTTP/REST structure. we should be pushing best practices and standards when adding new APIs because they're basically impossible to change once they're published

cns/restserver/ipam.go
```diff
@@ -891,3 +893,170 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+	if service.Options[common.OptManageEndpointState] == true {
+		if endpointInfo, ok := service.EndpointState[req.EndpointID]; ok {
+			logger.Printf("[UpdateEndpointState] Found existing endpoint state for infra container %s", req.EndpointID)
+			if req.HostVethName == "" && req.HnsEndpointID != "" {
+				logger.Warnf("[UpdateEndpointState] No HnsEndpointID or HostVethName has been provided")
+				return nil
+			}
+			if req.HnsEndpointID != "" {
+				service.EndpointState[req.EndpointID].HnsEndpointID = req.HnsEndpointID
+				logger.Errorf("[UpdateEndpointState] update the endpoint %s with HNSID  %s", req.EndpointID, req.HnsEndpointID)
```
vipul-21, line 1044: Why is this `logger.errorf` ?

cns/restserver/ipam.go
```diff
@@ -891,3 +893,170 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+				return nil
+			}
+			if req.HostVethName != "" {
+				service.EndpointState[req.EndpointID].HostVethName = req.HostVethName
+				logger.Errorf("[UpdateEndpointState] update the endpoint %s with vethName  %s", req.EndpointID, req.HostVethName)
```
vipul-21, line 1049: Same

cns/restserver/ipam.go
```diff
@@ -891,3 +893,170 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+		err := service.EndpointStateStore.Read(EndpointStoreKey, &service.EndpointState)
+		if err != nil {
+
+			if errors.Is(err, store.ErrKeyNotFound) {
+				// Nothing to retrieve.
+				logger.Printf("[GetEndpointState]  No endpoint state to retrieve.\n")
+			} else {
+				logger.Errorf("[GetEndpointState]  Failed to retrieve state, err:%v", err)
+			}
+			return nil, errors.Wrap(err, "[GetEndpointState]  Failed to retrieve state")
```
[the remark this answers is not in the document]
rbtr: "no endpoint state to retrieve" is a subclass of "failed to retrieve state", don't you think?

cns/restserver/ipam.go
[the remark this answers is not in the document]
behzad-mir: @vipul-21 I will only using this format when we are entering the func. Will remove [Azure CNS] from other places.

cns/NetworkContainerContract.go
```diff
@@ -35,6 +35,7 @@ const (
 	PathDebugRestData                        = "/debug/restdata"
 	NumberOfCPUCores                         = NumberOfCPUCoresPath
 	NMAgentSupportedAPIs                     = NmAgentSupportedApisPath
+	EndpointsPath                            = EndpointPath
```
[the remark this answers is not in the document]
behzad-mir: We need to have two const in here for correct routing based on my understanding of the code, one on api.go and one on networkcontainercontract. I was getting errors on client side when I was not using the contract. That's why you are seeing these two different constant. I am open to other name suggestion.

cns/restserver/ipam.go
```diff
@@ -891,3 +893,170 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+		err = service.Listener.Encode(w, &response)
+		logger.Response(service.Name, response, response.Response.ReturnCode, err)
+		return
+	}
+
+	endpointInfo, err := service.GetEndpointHelper(req.EndpointID)
+	if err != nil {
+		response := GetEndpointResponse{
+			Response: Response{
+				ReturnCode: types.InvalidRequest,
```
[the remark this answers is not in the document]
behzad-mir: Changed it to notfound

cns/restserver/ipam.go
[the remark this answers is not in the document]
vipul-21: So the concern, was not related to subclass actaully. (sorry about that)
What I wanted to ask is `No endpoint state to retrieve` is just a `printf` so do we want to fail on it  ? if so, the `logger.Errof` makes more sense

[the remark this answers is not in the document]
behzad-mir: I was not sure to print that no endpoint found as a logger error. There might be a situation that containerd create another delete call and the endpoint was already deleted. open to suggestions...

cns/client/client.go
behzad-mir: @rbtr Sorry I missed this comment. I'm wondering what is the resource in here? Is it the the full state which has:
EndpointID, POdName, POdNamespace, IPAddress, VethName, HNSID .... or the individual attributes?
In this API we are only assigning values to two attributes within the State. 
So the question is that is this consider a modification of state resource (PATCH) or changing the complete representation of HNSID (a PUT) ?

cns/restserver/ipam.go
behzad-mir: addressed in the commit.

cns/restserver/ipam.go
behzad-mir: I agree with you on this. However, all of the client server code on the CNS is written in this way (maybe we want to have our own specific cns response codes?!) I have removed the UpdateEndpointResponse struct and using the cns.Response instead.

cns/restserver/ipam.go
behzad-mir: addressed in the commit.

cns/client/client.go
```diff
@@ -1009,3 +1010,87 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error)
...
+		EndpointID: endpointID,
+	}
+	var body bytes.Buffer
+
+	if err := json.NewEncoder(&body).Encode(getEndpoint); err != nil {
+		return nil, errors.Wrap(err, "failed to encode getEndpoint")
+	}
+
+	u := c.routes[cns.EndpointsPath]
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), &body)
```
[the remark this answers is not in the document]
behzad-mir: addressed in the commit.

rbtr (changes requested):
there are unaddressed comments on this still, and some additional nits, but largely looks good to me

cns/service/main.go
```diff
@@ -643,13 +644,19 @@ func main() {
 		}
 		defer endpointStoreLock.Unlock() // nolint
 
-		err = platform.CreateDirectory(endpointStoreLocation)
+		endpointStorePath := endpointStoreLocation
+		if runtime.GOOS == "windows" {
+			endpointStorePath = os.Getenv("CNSStoreFilePath")
```
rbtr, line 649: conventionally envvars are ALL_CAPS, and I think this should be part of the CNS config, but it looks like the Linux one is a const so it's okay for now.

rbtr: @nddq can you make a note and create a task to make this endpoint store more automatically cross-platform - we shouldn't even need to clutter up the business logic with `if GOOS` branches

rbtr:
@behzad-mir i'm unblocking, but you have some unaddressed comments from other reviewers pending on this still

cns/client/client.go
```diff
@@ -1021,3 +1022,82 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error)
...
+	defer res.Body.Close()
+	if res.StatusCode != http.StatusOK {
+		return nil, errors.Errorf("http response %d", res.StatusCode)
+	}
+
+	var response cns.Response
+	err = json.NewDecoder(res.Body).Decode(&response)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to decode GetEndpointResponse")
```
vipul-21, line 1095: Is the error message correct ? Or should it be failed to decode `CNSResponse`

cns/client/client_test.go
```diff
@@ -2765,3 +2765,227 @@ func TestGetHomeAz(t *testing.T) {
...
+					httpStatusCodeToReturn: http.StatusOK,
+				},
+			},
+			&cns.EndpointRequest{
+				EndpointID: "foo",
+			},
+			false,
+		},
+		{
+			"with EndpointID",
```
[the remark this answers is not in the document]
vipul-21: Two test cases have same name `with EndpointID` ?

cns/client/client.go
behzad-mir: good catch. it should change since the response type has changed from previous commit.

cns/client/client_test.go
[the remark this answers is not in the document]
tamilmani1989: Did you fix this?

cns/NetworkContainerContract.go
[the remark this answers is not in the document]
tamilmani1989: wasn't this addressed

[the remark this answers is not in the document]
behzad-mir: added to the next commit

PR #2106, opened 2023-08-07
ci: Enable /azp run for submodule pipeline

jpayne3506:
**Reason for Change**:

Enables the use of [comment triggers](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#comment-triggers) for the submodules pipeline. `.pipelines/submodules-pipeline.yaml` was excluded from the triggers and prevented ` /azp run Azure Container Networking PR - submodules ` from being called to run the pipeline.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/submodules-pipeline.yaml
+++ b/.pipelines/submodules-pipeline.yaml
@@ -8,6 +8,7 @@ pr:
     - "azure-ipam/*"
     - "dropgz/*"
     - "cni/*"
+    - ".pipelines/submodules-pipeline.yaml" # Enable /azp run Azure Container Networking PR - submodules
     exclude:
     - "*"
 


camrynl (approved):

.pipelines/submodules-pipeline.yaml
```diff
@@ -7,6 +7,7 @@ pr:
     - "zapai/*"
     - "azure-ipam/*"
     - "dropgz/*"
+    - ".pipelines/submodules-pipeline.yaml" # Enable /azp run Azure Container Networking PR - submodules
```
vipul-21, line 10: I think there are directories under `.pipelines` that submodules uses. Those contains the actual tests which are more likely to get updated than he parent file `submodules`. Should we consider `.pipelines/singletenancy` as a trigger too ?

jpayne3506: Adding `.pipelines/cni` would capture everything, but it would also capture "infrastructure" templates. ex.` .pipelines/cni/k8s-e2e` and `.pipelines/cni/load-test-templates`. If we combined `.pipelines/singletenancy` into `.pipelines/cni/singletenancy` and ensure that it contained all of our templates, then I would say it would be covered and future-proof. To add, I would add the multitenancy work into its respective `.pipelines/cni/multitenancy` directory and include that into the triggers later.

While cniv1 changes may trigger the pipeline, most of our work going forward will involve the use of dropgz, which should be tested thoroughly on both a stable and latest CNI/CNS. If we really wanted to avoid extra pipeline runs we could add a cniv1 and cniv2 directory within singletenancy, but that might be excessive.

vipul-21 (approved):

vipul-21 (approved):

camrynl (approved):

PR #2107, opened 2023-08-07
put back the config to write logs to the file instead of only stdout logs

paulyufan2:
**Reason for Change**:

This PR is to put back thye config to write to the file instead of only stdout logs

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/plugin/main.go
+++ b/cni/network/plugin/main.go
@@ -15,9 +15,10 @@ import (
 	"github.com/Azure/azure-container-networking/aitelemetry"
 	"github.com/Azure/azure-container-networking/cni"
 	"github.com/Azure/azure-container-networking/cni/api"
-	"github.com/Azure/azure-container-networking/cni/log"
+	zaplog "github.com/Azure/azure-container-networking/cni/log"
 	"github.com/Azure/azure-container-networking/cni/network"
 	"github.com/Azure/azure-container-networking/common"
+	"github.com/Azure/azure-container-networking/log"
 	"github.com/Azure/azure-container-networking/nns"
 	"github.com/Azure/azure-container-networking/platform"
 	"github.com/Azure/azure-container-networking/store"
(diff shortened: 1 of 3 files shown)


cni/network/plugin/main.go
```diff
@@ -295,14 +296,21 @@ func main() {
...
 	}
 
-	loggerCfg := &log.Config{
+	log.SetName(name)
+	log.SetLevel(log.LevelInfo)
+	if err := log.SetTargetLogDirectory(log.TargetLogfile, ""); err != nil {
+		fmt.Printf("Failed to setup cni logging: %v\n", err)
+		return
+	}
+
```
tamilmani1989, line 305: can we move this before `NewPlugin` api is called?

tamilmani1989, line 305: call defer log.Close()

tamilmani1989 (changes requested):

cni/network/plugin/main.go
paulyufan2: it was added on line 306

paulyufan2: Hi Tamilmani, can you please explain more about this? This is already before NewPluigin(rootExecute()) is called.

vipul-21:
@paulyufan2 make sure you run the `Azure Container Networking PR - submodules` pipeline. It is getting skipped as the criteria to trigger the pipeline didn't match.( Will change that in separate PR)

paulyufan2:
> @paulyufan2 make sure you run the `Azure Container Networking PR - submodules` pipeline. It is getting skipped as the criteria to trigger the pipeline didn't match.( Will change that in separate PR: #2108 )

This is the submodule pipeline log:

cni/network/plugin/main.go
```diff
@@ -15,9 +15,10 @@ import (
 	"github.com/Azure/azure-container-networking/aitelemetry"
 	"github.com/Azure/azure-container-networking/cni"
 	"github.com/Azure/azure-container-networking/cni/api"
-	"github.com/Azure/azure-container-networking/cni/log"
+	zaplog "github.com/Azure/azure-container-networking/cni/log"
```
ramiro-gamarra, line 18: is this change needed? i'm not sure why the import alias is prefixed with `zap` (seems like an implementation detail). less changed lines if you just alias the new conflicting import?

paulyufan2: Hi Ramiro, because we want to rename the import packet since they are all using "log." to avoid conflicts

cni/network/plugin/main.go
tamilmani1989: ok then its fine

tamilmani1989 (approved):

PR #2111, opened 2023-08-08
[Multitenancy] : Add MT PodNetworkConfig CRD

aggarwal0009:
**Reason for Change**:

New CRD MultitenantPodNetworkConfig added to enable VNET multitenancy – which will be watched and managed by the control plane. 

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:


crd/multitenantpodnetworkconfig/client.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	spec.DeepCopyInto(&mpnc.Spec)
+	if err := c.cli.Update(ctx, mpnc); err != nil {
+		return nil, errors.Wrap(err, "failed to update mpnc")
+	}
+	return mpnc, nil
+}
+
+// SetOwnerRef sets the controller of the MultitenantPodNetworkConfig to the given object atomically, using HTTP Patch.
+// Deprecated: SetOwnerRef is deprecated, use the more correctly named SetControllerRef.
```
thatmattlong, line 133: Let's not add a new function that's already deprecated

crd/multitenantpodnetworkconfig/client.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Get returns the MultitenantPodNetworkConfig identified by the NamespacedName.
+func (c *Client) Get(ctx context.Context, key types.NamespacedName) (*v1alpha1.MultitenantPodNetworkConfig, error) {
+	multitenantPodNetworkConfig := &v1alpha1.MultitenantPodNetworkConfig{}
+	err := c.cli.Get(ctx, key, multitenantPodNetworkConfig)
+	return multitenantPodNetworkConfig, errors.Wrapf(err, "failed to get mpnc %v", key)
+}
+
+// PatchSpec performs a server-side patch of the passed MultitenantPodNetworkConfigSpec to the MultitenantPodNetworkConfig specified by the NamespacedName.
+func (c *Client) PatchSpec(ctx context.Context, key types.NamespacedName, spec *v1alpha1.MultitenantPodNetworkConfigSpec, fieldManager string) (*v1alpha1.MultitenantPodNetworkConfig, error) {
```
thatmattlong, line 109: is this necessary? MTPNC will get created with known values for spec and I don't expect the spec to need to get updated

crd/multitenantpodnetworkconfig/client.go
```diff
@@ -0,0 +1,161 @@
...
+	obj.Spec = *spec
+	if err := c.cli.Patch(ctx, obj, client.Apply, client.ForceOwnership, client.FieldOwner(fieldManager)); err != nil {
+		return nil, errors.Wrap(err, "failed to patch mpnc")
+	}
+	return obj, nil
+}
+
+// UpdateSpec does a fetch, deepcopy, and update of the MultitenantPodNetworkConfig with the passed spec.
+// Deprecated: UpdateSpec is deprecated and usage should migrate to PatchSpec.
+func (c *Client) UpdateSpec(ctx context.Context, key types.NamespacedName, spec *v1alpha1.MultitenantPodNetworkConfigSpec) (*v1alpha1.MultitenantPodNetworkConfig, error) {
```
thatmattlong, line 120: also I think this func is not needed for same reason as above

thatmattlong (changes requested):
let's not add the helper funcs to the client which mutate the spec until we know we need them

crd/multitenantpodnetworkconfig/client.go
aggarwal0009: removed

crd/multitenantpodnetworkconfig/client.go
aggarwal0009: removed

crd/multitenantpodnetworkconfig/client.go
aggarwal0009: removed

miguelgoms (dismissed):

crd/multitenantpodnetworkconfig/README.md
```diff
@@ -0,0 +1,3 @@
+# MultitenantPodNetworkConfig CRDs
+
+MPNC objects represent the network configuration goal state for a pod running a multitenant networked container and are created and managed by control plane as part of the network configuration, during Pod lifecycle events.
```
neaggarwMS, line 3: I would also add that it is created by AKS Customers. So any change in the CRD definition has to be backward compatible.

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -0,0 +1,62 @@
...
+	// name of PNI object from requesting cx pod
+	PodNetworkInstance string `json:"podNetworkInstance,omitempty"`
+	// name of the requesting cx pod
+	PodName string `json:"podName,omitempty"`
+}
+
+// MultitenantPodNetworkConfigStatus defines the observed state of PodNetworkConfig
+type MultitenantPodNetworkConfigStatus struct {
+	// network container id
+	UUID string `json:"uuid,omitempty"`
```
neaggarwMS, line 53: Lets call it NCId only or CompartmentId

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -0,0 +1,62 @@
...
+	// ip allocated to the network container
+	PrimaryIP string `json:"primaryIP,omitempty"`
```
neaggarwMS, line 55: You also need to pass the IPSubnet details

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -0,0 +1,62 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []MultitenantPodNetworkConfig `json:"items"`
+}
+
+// MultitenantPodNetworkConfigSpec defines the desired state of PodNetworkConfig
+type MultitenantPodNetworkConfigSpec struct {
+	// name of PNI object from requesting cx pod
+	PodNetworkInstance string `json:"podNetworkInstance,omitempty"`
+	// name of the requesting cx pod
+	PodName string `json:"podName,omitempty"`
```
neaggarwMS, line 47: Can we call PodContext? It is podName + PodNamespace, no?

neaggarwMS (changes requested):

crd/multitenantpodnetworkconfig/README.md
miguelgoms: MPNC object will be created by the swift control plane, though - the crds created by the AKS customers will be on a separate PR. The point on backward compat is valid, but the interface is between CNS and control plane.

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
miguelgoms: I think we removed the pod-namespace on the component design review, because it was redundant information (the MPNC would have an ownerref to the pod, so it would always be on the same namespace)

thatmattlong: correct, the MTPNC namespace matches the Pod's namespace, don't need pod namespace in CRD

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
aggarwal0009: Waiting for Jaeryn to update the spec if this is needed

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
aggarwal0009: updated

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -0,0 +1,62 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Important: Run "make" to regenerate code after modifying this file
+
+// +kubebuilder:object:root=true
+
+// MultitenantPodNetworkConfig is the Schema for the multitenantpodnetworkconfigs API
+// +kubebuilder:resource:scope=Namespaced
+// +kubebuilder:resource:shortName=mpnc
```
thatmattlong, line 16: suggest mtpnc for shotname since that's what we've been saying out loud

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
thatmattlong: For now let's just add `Gateway string `json:"gateway,omitempty"`. Will update the spec after we close the routing design discussion, and if we need more than gateway we can add it in a future PR. Don't want to block RC dev until closing routing design and that design will impact this. So let's just add this one property and close this.

thatmattlong (approved):

rbtr (approved):

neaggarwMS (approved):

PR #2112, opened 2023-08-09
feat: stub CNS Pod watcher

rbtr:
**Reason for Change**:

Adds config and Reconciler for watching Pods. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/reconciler.go
+++ b/cns/kubecontroller/nodenetworkconfig/reconciler.go
@@ -164,6 +164,9 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, node *v1.Node) error {
 		WithEventFilter(predicate.Funcs{
 			// check that the generation is the same - status changes don't update generation.
 			UpdateFunc: func(ue event.UpdateEvent) bool {
+				if ue.ObjectOld == nil || ue.ObjectNew == nil {
+					return false
+				}
 				return ue.ObjectOld.GetGeneration() == ue.ObjectNew.GetGeneration()
 			},
 		}).
--- a/cns/kubecontroller/podwatcher/podwatcher.go
+++ b/cns/kubecontroller/podwatcher/podwatcher.go
@@ -1,0 +1,66 @@
+package podwatcher
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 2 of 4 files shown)


cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -164,6 +164,9 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, node *v1.Node) error {
 		WithEventFilter(predicate.Funcs{
 			// check that the generation is the same - status changes don't update generation.
 			UpdateFunc: func(ue event.UpdateEvent) bool {
+				if ue.ObjectOld == nil || ue.ObjectNew == nil {
```
thatmattlong, line 167: is this possible in an update event?

cns/kubecontroller/podwatcher/podwatcher.go
```diff
@@ -0,0 +1,66 @@
...
+func New(nodename string, listeners ...podListener) *podWatcher { //nolint:revive // private struct to force constructor
+	return &podWatcher{
+		listeners: listeners,
+		listOpt:   &client.ListOptions{FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": nodename})},
+	}
+}
+
+func (p *podWatcher) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	podList := &v1.PodList{}
+	if err := p.cli.List(ctx, podList, p.listOpt); err != nil {
```
thatmattlong, line 39: is calling List for each pod event necessary? could we just keep track of each pod individually, and if CNS doesn't know about a pod when a request comes in, we just fail the request?

cns/kubecontroller/nodenetworkconfig/reconciler.go
rbtr: I copied this from the [`GenerationChangedPredicate`](https://github.com/kubernetes-sigs/controller-runtime/blob/5aa6a71fc79d4058b26355eb58bdf2430fce423e/pkg/predicate/predicate.go#L161-L172) Update handler so I think the answer is yes

cns/kubecontroller/podwatcher/podwatcher.go
rbtr: I wrote this keeping in mind that the podwatcher could be reused by regular swift scenario in the ipam redesign...
We don't need to List in v2, wiring up the reconciler is sufficient, so I can make this a little more middleware-ish and dynamically inject the whole reconcile func maybe?

rbtr: did some abstracting here - I went ahead and wrote some boilerplate for future use, but the net current behavior is that reconcile events are discarded for V2 and we are only using this to populate the cache.

nddq (approved):
lgtm

rbtr:
bypassing submodules pipeline as this code is not exercised in it

PR #2114, opened 2023-08-09
feat: CNS RequestIPAddress branching for MT/V2

nddq:
**Reason for Change**:

This is my initial proposal to modify the current IPAM flow for SWIFT V2, pending changes from other PBI.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/middlewares/multitenant.go
```diff
@@ -0,0 +1,50 @@
...
+
+type IPConfigValidator func(ipConfigsRequest *cns.IPConfigsRequest) (types.ResponseCode, string)
+
+// Middleware interface for testing later on
+type Middleware interface {
+	Validator() IPConfigValidator
+	GetMultitenantIPConfig(podInfo cns.PodInfo) (*cns.PodIpInfo, error)
+}
+
+type MultitenantMiddleware struct {
```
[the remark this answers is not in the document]
nddq: I was actually talking about about this with Evan 🙂. I'm thinking changing this to `SwiftV2Middleware` but I'm open to suggestions.

cns/api.go
```diff
@@ -48,8 +48,17 @@ type HTTPService interface {
 	GetPendingReleaseIPConfigs() []IPConfigurationStatus
 	GetPodIPConfigState() map[string]IPConfigurationStatus
 	MarkIPAsPendingRelease(numberToMark int) (map[string]IPConfigurationStatus, error)
+	AttachSWIFTv2Middleware(middleware SWIFTv2Middleware)
```
rbtr, line 51: do we need this in the service interface? seems like an implementation detail

cns/middlewares/SWIFTv2.go
```diff
@@ -0,0 +1,111 @@
...
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var ErrMTPNCNotReady = errors.New("mtpnc is not ready")
+
+type SWIFTv2Middleware struct {
+	// TODO: implement
+	// need cached scoped client for pods
+	// need client for MTPNC CRD for x-ref pods
```
rbtr, line 22: stale comment?

cns/middlewares/SWIFTv2.go
```diff
@@ -0,0 +1,111 @@
...
+func NewSWIFTv2Middleware(cli client.Client) *SWIFTv2Middleware {
+	return &SWIFTv2Middleware{
+		cli: cli,
+	}
+}
+
+// Return the validator function for the middleware
+func (m *SWIFTv2Middleware) Validator() cns.IPConfigValidator {
+	return m.validateMultitenantIPConfigsRequest
+}
```
rbtr, line 35: why do we need this extra layer of indirection? it seems like we could just have a `Validate(req)` method?

cns/middlewares/SWIFTv2.go
```diff
@@ -0,0 +1,111 @@
...
+func (m *SWIFTv2Middleware) validateMultitenantIPConfigsRequest(req *cns.IPConfigsRequest) (respCode types.ResponseCode, message string) {
+	// Retrieve the pod from the cluster
+	podInfo, err := cns.UnmarshalPodInfo(req.OrchestratorContext)
+	if err != nil {
+		errBuf := fmt.Sprintf("unmarshalling pod info from ipconfigs request %v failed with error %v", req, err)
+		return types.UnexpectedError, errBuf
+	}
+	podNamespacedName := k8stypes.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
+	pod := v1.Pod{}
+	err = m.cli.Get(context.TODO(), podNamespacedName, &pod)
```
rbtr, line 48: let's pass a context in (and thread it back up the call stack as far as necessary)

cns/restserver/ipam.go
```diff
@@ -35,6 +35,24 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ipconfigsRequest cn
 		}, errors.New("failed to validate ip config request")
 	}
 
+	var MTpodIPInfo cns.PodIpInfo
+
+	// Request is for multitenant pod
+	if podInfo.IsMultitenant() {
+		// If pod is multitenant and we failed to grab its MTPNC IP config, return error immediately
+		podIPInfo, err := service.MultitenantMiddleware.GetSWIFTv2IPConfig(podInfo)
```
rbtr, line 43: it's probably not the responsibility of the multitenant "middleware" to do this and would be better organized as a different component

cns/restserver/restserver.go
```diff
@@ -349,3 +352,9 @@ func (service *HTTPRestService) MustGenerateCNIConflistOnce() {
 		}
 	})
 }
+
+func (service *HTTPRestService) AttachSWIFTv2Middleware(middleware cns.SWIFTv2Middleware) {
+	service.MultitenantMiddleware = middleware
+	// add multitenant ipconfig validator function
+	service.ipConfigsValidators = append(service.ipConfigsValidators, middleware.Validator())
```
rbtr, line 359: continuing my thought from above that we don't need this extra `Validator` method, and could have a `Validate(req) (...)` instead - here, that would simply look like:
```suggestion
	service.ipConfigsValidators = append(service.ipConfigsValidators, middleware.Validate)
```

rbtr (dismissed):
some nitpicks and suggestions, but lgtm as a proof of concept for further validation

cns/restserver/restserver.go
nddq: yeah you are right, honestly I don't remember why I even had that method 🙂

cns/restserver/ipam.go
nddq: my thinking is that this middleware will handle all related SWIFT v2 business, so it makes sense that it will handle returning the SWIFTv2 config as well..

cns/api.go
nddq: yeah, because in `InitializeCRDState` we are atually passing in the `HTTPService` interface, and then convert that interface type to implementation type.

cns/middlewares/SWIFTv2.go
nddq: we are not passing down any context in the current RequestIPConfigs flow, how far up the call stack are we talking about?

rbtr: At least up to the http.Handler - the context should exist in the scope of a single request tx

nddq: would it make more sense to open another PR for this? Since we have to do refractor in other places as well that are not related to this PR

rbtr: that's fine

rbtr (approved):

cns/NetworkContainerContract.go
```diff
@@ -77,6 +77,12 @@ const (
 	MultiTenantCRD = "MultiTenantCRD"
 )
 
+// Address Types
+const (
+	Default   = "Default"
+	Secondary = "Secondary"
+)
```
thatmattlong, line 84: suggest making this an "enum" like
```
type AddressType string

const (
    AddressTypeDefault = "Default"
    AddressTypeSecondary = "Secondary"
)

File: cns/NetworkContainerContract.go [hunk 9]
```diff
@@ -239,6 +248,10 @@ func (p *podInfo) OrchestratorContext() (json.RawMessage, error) {
 	return jsonContext, nil
 }
 
+func (p *podInfo) IsMultitenant() bool {
+	return p.Multitenant
```
thatmattlong, line 252: do we need a Getter for this? The struct member is already public

cns/configuration/env.go
```diff
@@ -12,12 +12,20 @@ const (
 	// EnvNodeIP is the IP of the node running this CNS binary
 	EnvNodeIP = "NODE_IP"
 	// LabelSwiftV2 is the Node label for Swift V2
-	LabelSwiftV2 = "kubernetes.azure.com/podnetwork-multi-tenancy"
+	LabelSwiftV2   = "kubernetes.azure.com/podnetwork-multi-tenancy"
```
thatmattlong, line 15: This label is `kubernetes.azure.com/podnetwork-multi-tenancy-enabled`

cns/middlewares/SWIFTv2.go
```diff
@@ -0,0 +1,103 @@
+package middlewares
```
thatmattlong, line 1: filenames should be all lowercase

cns/middlewares/SWIFTv2.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/Azure/azure-container-networking/crd/multitenancy/api/v1alpha1"
+	v1 "k8s.io/api/core/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var ErrMTPNCNotReady = errors.New("mtpnc is not ready")
+
+type SWIFTv2Middleware struct {
+	cli client.Client
```
thatmattlong, line 20: export this and drop the constructor, the constructor isn't doing anything but assigning this. We only use constructors in Go when we need to set some default values or run some logic

cns/middlewares/SWIFTv2.go
```diff
@@ -0,0 +1,103 @@
...
+	// Retrieve the pod from the cluster
+	podInfo, err := cns.UnmarshalPodInfo(req.OrchestratorContext)
+	if err != nil {
+		errBuf := fmt.Sprintf("unmarshalling pod info from ipconfigs request %v failed with error %v", req, err)
+		return types.UnexpectedError, errBuf
+	}
+	podNamespacedName := k8stypes.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
+	pod := v1.Pod{}
+	err = m.cli.Get(context.TODO(), podNamespacedName, &pod)
+	if err != nil {
```
thatmattlong, line 41: recommend combining err instantiation and check,
```
if err := m.cli.Get(context.TODO(), podNamespacedName, &pod); err != nil
```

cns/NetworkContainerContract.go
nddq: will this be replaced by `nicType` instead of `addressType` @jaer-tsun ?

cns/NetworkContainerContract.go
nddq: im following the convention for other podInfo's field getters, but yeah this makes sense.

rbtr: we do, because the object is exclusively accessed through the interface (note the struct is private `podInfo`)

nddq: Is there a reason why we are accessing `podInfo` object through `PodInfo` interface? Because I notice that the only thing that is implementing this interface is `podInfo` itself.

cns/restserver/ipam.go
```diff
@@ -71,6 +90,18 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ipconfigsRequest cn
 		}
 	}
 
+	// Adding MTNpodIPInfo to the response, if it is not empty
+	if MTpodIPInfo.PodIPConfig.IPAddress != "" {
+		defaultGateway := podIPInfo[0].NetworkContainerPrimaryIPConfig.GatewayIPAddress
+		// Populate the non-multitenant routes with the default gateway
+		for i := range MTpodIPInfo.Routes {
+			if MTpodIPInfo.Routes[i].GatewayIPAddress == "" {
+				MTpodIPInfo.Routes[i].GatewayIPAddress = defaultGateway
```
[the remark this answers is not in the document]
nddq: So beside the `GatewayIPAddress `for the MT route that should already been set to this point, the two route for `PodCIDR` and `ServiceCIDR` `GatewayIPAddress `haven't been set yet, so this loop essentially setting that up for them.

cns/middlewares/mock/mockClient.go
```diff
@@ -0,0 +1,127 @@
...
+			*o = *mtpnc
+		} else {
+			return errMTPNCNotFound
+		}
+	}
+	return nil
+}
+
+// List implements client.Client.List.
+func (c *MockClient) List(_ context.Context, _ client.ObjectList, _ ...client.ListOption) error {
```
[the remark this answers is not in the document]
nddq: huh that's neat, the only method im going to use and already implemented from client.Client is `Get` anyways so that's should be fine.

cns/NetworkContainerContract.go
rbtr: IIRC it was to force usage of the `Key()` method, which does different things based on how the `podInfo` was created, rather than allow direct access to the fields

cns/NetworkContainerContract.go
```diff
@@ -406,6 +422,10 @@ type PodIpInfo struct {
 	PodIPConfig                     IPSubnet
 	NetworkContainerPrimaryIPConfig IPConfiguration
 	HostPrimaryIPInfo               HostIPInfo
+	NICType                         NICType // default, secondary
```
neaggarwMS, line 425: lets remove comment as //default, secondary. We are changing names to be {Infra, Secondary, Management and IB}

cns/configuration/env.go
```diff
@@ -12,12 +12,20 @@ const (
 	// EnvNodeIP is the IP of the node running this CNS binary
 	EnvNodeIP = "NODE_IP"
 	// LabelSwiftV2 is the Node label for Swift V2
-	LabelSwiftV2 = "kubernetes.azure.com/podnetwork-multi-tenancy"
+	LabelSwiftV2   = "kubernetes.azure.com/podnetwork-multi-tenancy-enabled"
```
neaggarwMS, line 15: Do we need to call it as "kubernetes.azure.com/podnetwork-enabled" to check more secondary interfaces (SR IOV or multitenant)

cns/configuration/env.go
```diff
@@ -12,12 +12,20 @@ const (
...
+	EnvPodCIDR     = "POD_CIDR"
+	EnvServiceCIDR = "SERVICE_CIDR"
```
neaggarwMS, line 17: What about DNS?

cns/NetworkContainerContract.go
```diff
@@ -78,6 +78,14 @@ const (
 	MultiTenantCRD = "MultiTenantCRD"
 )
 
+type NICType string
+
+// NIC Types
+const (
+	NICTypeDefault   NICType = "Default"
```
neaggarwMS, line 85: Lets call them "Infra" and "Secondary"

cns/NetworkContainerContract.go
```diff
@@ -173,6 +181,8 @@ type PodInfo interface {
 	Equals(PodInfo) bool
 	// String implements string for logging PodInfos
 	String() string
+	// IsMultitenant returns true if the pod is multitenant
+	IsMultitenant() bool
```
neaggarwMS, line 185: Instead can we set IsSecondaryInterfacesSet?

cns/NetworkContainerContract.go
```diff
@@ -188,6 +198,7 @@ type podInfo struct {
 	PodInfraContainerID string
 	PodInterfaceID      string
 	Version             podInfoScheme
+	Multitenant         bool
```
neaggarwMS, line 201: Lets rename it to SecondaryInterfacesExist?

cns/NetworkContainerContract.go
```diff
@@ -278,6 +293,7 @@ func NewPodInfoFromIPConfigsRequest(req IPConfigsRequest) (PodInfo, error) {
 	}
 	p.(*podInfo).PodInfraContainerID = req.InfraContainerID
 	p.(*podInfo).PodInterfaceID = req.PodInterfaceID
+	p.(*podInfo).Multitenant = req.Multitenant
```
neaggarwMS, line 296: You are never going to have Multitenant set on IpConfigRequest from CNI.

cns/restserver/ipam.go
```diff
@@ -24,7 +25,7 @@ var (
 )
 
 // requestIPConfigHandlerHelper validates the request, assigns IPs, and returns a response
-func (service *HTTPRestService) requestIPConfigHandlerHelper(ipconfigsRequest cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
+func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context, ipconfigsRequest cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
 	podInfo, returnCode, returnMessage := service.validateIPConfigsRequest(ipconfigsRequest)
```
neaggarwMS, line 29: Add a comment here what is it doing with special case for swift 2.0

cns/NetworkContainerContract.go
```diff
@@ -429,6 +449,7 @@ type IPConfigsRequest struct {
 	InfraContainerID    string          `json:"infraContainerID"`
 	OrchestratorContext json.RawMessage `json:"orchestratorContext"`
 	Ifname              string          `json:"ifname"` // Used by delegated IPAM
+	Multitenant         bool            `json:"multitenant"`
 }
 
 // IPConfigResponse is used in CNS IPAM mode as a response to CNI ADD
```
neaggarwMS, line 455: Extend this to also set isSecondaryInterfacesSet

neaggarwMS, line 455: Extend IpConfigsResponses to add SecondaryInterfacesExist

cns/configuration/env.go
nddq: I didn't see DNS mentioned when we were discussing routing a few weeks ago, how/what would the route for DNS look like?

cns/configuration/env.go
nddq: @thatmattlong @rbtr

cns/NetworkContainerContract.go
```diff
@@ -182,6 +182,8 @@ type PodInfo interface {
 	Equals(PodInfo) bool
 	// String implements string for logging PodInfos
 	String() string
+	// IsSecondaryInterfacesSet returns true if there exist a secondary interface for this pod
+	IsSecondaryInterfacesSet() bool
```
jaer-tsun, line 186: could be called SecondaryInterfaceExists too

cns/configuration/env.go
jaer-tsun, line 15: was this label used before?

cns/middlewares/swiftV2.go
```diff
@@ -0,0 +1,96 @@
...
+	mtpncNamespacedName := k8stypes.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
+	if err := m.Cli.Get(ctx, mtpncNamespacedName, &mtpnc); err != nil {
+		return cns.PodIpInfo{}, fmt.Errorf("failed to get pod's mtpnc from cache : %w", err)
+	}
+
+	// Check if the MTPNC CRD is ready. If one of the fields is empty, return error
+	if mtpnc.Status.PrimaryIP == "" || mtpnc.Status.MacAddress == "" || mtpnc.Status.NCID == "" || mtpnc.Status.GatewayIP == "" {
+		return cns.PodIpInfo{}, ErrMTPNCNotReady
+	}
+	podIPInfo := cns.PodIpInfo{}
```
jaer-tsun, line 59: could combine like this too

podIPInfo := cns.PodIpInfo{
    PodIPConfig: cns.IPSubnet{IPAddress: mtpnc.Status.PrimaryIP,
    MacAddress: mtpnc.Status.MacAddress,
    ...
}

cns/middlewares/swiftV2.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	podCIDR, err := configuration.PodCIDR()
+	if err != nil {
+		return cns.PodIpInfo{}, fmt.Errorf("failed to get pod CIDR from environment : %w", err)
+	}
+
+	podCIDRRoute := cns.Route{
+		IPAddress:      podCIDR,
+		InterfaceToUse: "eth0",
```
jaer-tsun, line 81: are we using InterfaceToUse? We're not using on CNI side

cns/middlewares/swiftV2.go
```diff
@@ -0,0 +1,96 @@
...
+		GatewayIPAddress: mtpnc.Status.GatewayIP,
+		InterfaceToUse:   "eth1",
+	}
+
+	podCIDR, err := configuration.PodCIDR()
+	if err != nil {
+		return cns.PodIpInfo{}, fmt.Errorf("failed to get pod CIDR from environment : %w", err)
+	}
+
+	podCIDRRoute := cns.Route{
```
jaer-tsun, line 79: we'd need the gateway too, for linux transparent mode we just use 169.254.1.1

cns/middlewares/swiftV2.go
```diff
@@ -0,0 +1,96 @@
...
+	podIPInfo := cns.PodIpInfo{}
+	podIPInfo.PodIPConfig = cns.IPSubnet{
+		IPAddress: mtpnc.Status.PrimaryIP,
+	}
+	podIPInfo.MacAddress = mtpnc.Status.MacAddress
+	podIPInfo.NICType = cns.DelegatedVMNIC
+	podIPInfo.SkipDefaultRoutes = true
+	// podIPInfo.InterfaceName is empty for DelegatedVMNIC
+
+	defaultRoute := cns.Route{
```
jaer-tsun, line 68: for default route it can just be

ip route add default via 0.0.0.0 dev eth1

cns/middlewares/swiftV2.go
```diff
@@ -0,0 +1,96 @@
...
+	serviceCIDR, err := configuration.ServiceCIDR()
+	if err != nil {
+		return cns.PodIpInfo{}, fmt.Errorf("failed to get service CIDR from environment : %w", err)
+	}
+
+	serviceCIDRRoute := cns.Route{
+		IPAddress:      serviceCIDR,
+		InterfaceToUse: "eth0",
+	}
+	podIPInfo.Routes = []cns.Route{defaultRoute, podCIDRRoute, serviceCIDRRoute}
```
jaer-tsun, line 93: these routes should be on different PodIPInfo (unless this PodIPInfo struct is internal) since CNI loops through the PodIPInfo slice in response

cns/middlewares/swiftV2.go
nddq: hmm so where would the MTPNC's IP address and MTPNC's gateway address be used in this case?

cns/middlewares/swiftV2.go
nddq: The default gateway will be added in `requestIPConfigHandlerHelper`, after this function exit.

cns/middlewares/swiftV2.go
nddq: which routes should be on what `PodIPInfo`? Assuming that currently with SWIFT v2, CNI will get 3 `PodIPInfo `back

cns/middlewares/mock/mockSWIFTv2.go
```diff
@@ -0,0 +1,92 @@
...
+		errBuf := fmt.Sprintf("unmarshalling pod info from ipconfigs request %v failed with error %v", req, err)
+		return types.UnexpectedError, errBuf
+	}
+	podNamespacedName := k8types.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
+	pod, ok := m.mtPodState[podNamespacedName.String()]
+	if !ok {
+		errBuf := fmt.Sprintf("failed to get pod %v with error %v", podNamespacedName, err)
+		return types.UnexpectedError, errBuf
+	}
+	// check the pod labels for Swift V2, enrich the request with the multitenant flag. TBD on the label
```
miguelgoms, line 58: Seems we are mixing logic for the labels on the Node which is "multitenant-enabled" (secondary nic exists), versus labels for the pod *requiring* multitenancy. Pod labels for swift-v2 were expected to be: 
kubernetes.azure.com/pod-network ==> this marks the pod as requiring multitenancy
kubernetes.azure.com/pod-network-instance

cns/middlewares/swiftV2.go
```diff
@@ -0,0 +1,119 @@
...
+		return types.UnexpectedError, errBuf
+	}
+	podNamespacedName := k8stypes.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
+	pod := v1.Pod{}
+	if err := m.Cli.Get(context.TODO(), podNamespacedName, &pod); err != nil {
+		errBuf := fmt.Sprintf("failed to get pod %v with error %v", podNamespacedName, err)
+		return types.UnexpectedError, errBuf
+	}
+
+	// check the pod labels for Swift V2, set the request's SecondaryInterfaceSet flag to true.
```
miguelgoms, line 49: same comment as on the mock.

miguelgoms (changes requested):

cns/middlewares/mock/mockSWIFTv2.go
nddq: So the label `kubernetes.azure.com/podnetwork-multi-tenancy-enabled` should only be used for the node which I get, but should the pod label be `kubernetes.azure.com/pod-network` or `kubernetes.azure.com/pod-network-instance`? I'm kinda confused about your comment formatting. Do you mean `kubernetes.azure.com/pod-network` should be used to label swift-v2 pods and `kubernetes.azure.com/pod-network-instance` should be used to label pods that have its secondary nic(s) exist?

cns/configuration/env.go
```diff
@@ -12,12 +12,24 @@ const (
 	// EnvNodeIP is the IP of the node running this CNS binary
 	EnvNodeIP = "NODE_IP"
 	// LabelSwiftV2 is the Node label for Swift V2
-	LabelSwiftV2 = "kubernetes.azure.com/podnetwork-multi-tenancy"
+	LabelSwiftV2   = "kubernetes.azure.com/podnetwork-multi-tenancy-enabled"
+	EnvPodV4CIDR   = "POD_V4_CIDR"
+	EnvPodV6CIDR   = "POD_V6_CIDR"
+	EnvServiceCIDR = "SERVICE_CIDR"
```
tyler-lloyd, line 18: what about ipv6 service cidr?

nddq: added

miguelgoms (approved):

rbtr (approved):

cns/configuration/env.go
```diff
@@ -11,13 +11,30 @@ const (
 	EnvNodeName = "NODENAME"
 	// EnvNodeIP is the IP of the node running this CNS binary
 	EnvNodeIP = "NODE_IP"
-	// LabelSwiftV2 is the Node label for Swift V2
-	LabelSwiftV2 = "kubernetes.azure.com/podnetwork-multi-tenancy"
+	// LabelNodeSwiftV2 is the Node label for Swift V2
+	LabelNodeSwiftV2 = "kubernetes.azure.com/podnetwork-multi-tenancy-enabled"
+	LabelPodSwiftV2  = "kubernetes.azure.com/pod-network"
+	EnvPodV4CIDR     = "POD_V4_CIDR"
```
tyler-lloyd, line 17: just out of curiosity, what's the reason for having these as env vars instead of in the cns configMap? these CIDRs should be shared between all nodes so it feels like they are fine to have in the configMap? cc @rbtr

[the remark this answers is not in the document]
nddq: If we support multiple CIDRs in the future, then wouldn't it be better to have two separated env `POD_V4_CIDR` and `POD_V4_CIDRs`?

miguelgoms (dismissed):

cns/configuration/env.go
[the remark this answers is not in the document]
rbtr: no, it would be better to solve the general case, call it CIDRs, and then only put a single item in the list

rbtr: I probably said we could push them in to the env like downward API at some point and we just ran with that. 
I don't think there's a practical difference, in an ideal world (ie with viper) the env and the configmap would be equal and interchangeable :( do you have a preference? We could still put them in both places

tyler-lloyd: maybe weak preference towards configMap since I don't think we'd ever want different CNSes to have different CIDR sets and 2) if anything else in the future needed to read these values for whatever reason it's easier to read a cm than parse a ds spec for env vars.

really not going to die on this hill though so if we're too far into this to change that's fine with me.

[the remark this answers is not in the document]
rbtr: should be fine to allow that, wdyt @nddq?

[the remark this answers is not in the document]
nddq: yeah if k8s and AKS is doing it then it should be fine. I would still need to split them to create routes down the line though.

cns/restserver/util.go
```diff
@@ -787,6 +783,18 @@ func (service *HTTPRestService) validateIPConfigsRequest(
 	return podInfo, types.Success, ""
 }
 
+// validateDefaultIPConfigsRequest validates the request for default IP configs request
+func (service *HTTPRestService) validateDefaultIPConfigsRequest(_ context.Context, ipConfigsRequest *cns.IPConfigsRequest) (respCode types.ResponseCode, message string) {
```
jaer-tsun, line 787: should we rename to validateInfraIPConfigsRequest to match contract?

cns/restserver/util.go
```diff
@@ -766,17 +766,13 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ctx context.Context,
...
-) (cns.PodInfo, types.ResponseCode, string) {
-	if service.state.OrchestratorType != cns.KubernetesCRD && service.state.OrchestratorType != cns.Kubernetes {
-		return nil, types.UnsupportedOrchestratorType, "ReleaseIPConfig API supported only for kubernetes orchestrator"
-	}
-
-	if ipConfigsRequest.OrchestratorContext == nil {
-		return nil,
-			types.EmptyOrchestratorContext,
-			fmt.Sprintf("OrchastratorContext is not set in the req: %+v", ipConfigsRequest)
+func (service *HTTPRestService) validateIPConfigsRequest(ctx context.Context, ipConfigsRequest cns.IPConfigsRequest) (cns.PodInfo, types.ResponseCode, string) {
```
jaer-tsun, line 769: should we rename to validateSecondaryIPConfigsRequest or validateDelegatedIPConfigsRequest?

miguelgoms (approved):

miguelgoms (dismissed):

cns/restserver/util.go
nddq: this is the parent validateIPConfigsRequest function, if you look below, it will loop through all the validators, which are the infra one and the delegated one (if added)

cns/restserver/util.go
nddq: fwiw this validator will run on all ipconfigs requests, so I think that leaving it as "default" should suffice.

thatmattlong (dismissed):

tyler-lloyd (approved):

miguelgoms (approved):

thatmattlong (dismissed):

miguelgoms (approved):

tyler-lloyd (approved):

nddq:
Merged [2300](https://github.com/Azure/azure-container-networking/pull/2300)

PR #2115, opened 2023-08-09
fix: don't delete an existing conflist

thatmattlong:
**Reason for Change**:

Instead of moving an existing CNI conflist to <filename>-old immediately, this will copy it to <filename>-old instead. This means we will still generate the new conflist on top of the existing one, but without ever deleting the existing conflist. This will prevent nodes from toggling between Ready->NotReady->Ready when CNS gets restarted (either due to crash or normal rollout).

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:
not really sure how to write a good test for this since the logic is in `main`, need to think about it some more

### DIFF
--- a/internal/fs/atomic.go
+++ b/internal/fs/atomic.go
@@ -2,6 +2,7 @@ package fs
 
 import (
 	"io"
+	"io/fs"
 	"os"
 	"path"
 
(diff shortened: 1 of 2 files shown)


rbtr:
This feels like we're going backwards wrt our deployment strategy and being confident of the versions/config that are really deployed. I would like that this still upgraded previous conflists on disk.
Could we instead update the flow to:
wait for NC ->
write conflist to tmp ->
rename to final destination

I think this solves the problem of Unready'ing the Node, since if a conflist is already present it will not be deleted, and there is no gap where a conflist is not present. It also _maintains_ our ability to upgrade the conflists since CNS will write the new one over the old one during startup.

Based on the `os.Rename` docs, it seems like this should work:
> Rename renames (moves) oldpath to newpath. _If newpath already exists and is not a directory, Rename replaces it._

This should get us to continuous conflist existence on Linux.

On Windows we can probably never upgrade file in-place due to the access write-lock...

thatmattlong:
I think they made os.Rename atomic even in windows these days using some convoluted syscalls so I think we can make this change for both (but windows isn't implemented yet anyway). It seems reasonable to me, let me revisit the AtomicWriter to make this change so New* doesn't move the old one but instead generates a temp file, then generate will write to the temp file and move it.

rbtr (approved):
lgtm per stated intent, but I doubt how useful keeping the `-old` file really is

nddq (approved):

PR #2136, opened 2023-08-11
fix: Change argument in load-test template

jpayne3506:
**Reason for Change**:

Fixes improper argument for `^TestScaleDeployment$`. Replaced and improved the functionality for reporting in load-test templates to be dynamic with inputs.  

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
scaleup: Now needs a value between 1-240

### DIFF
--- a/.pipelines/cni/load-test-templates/pod-deployment-template.yaml
+++ b/.pipelines/cni/load-test-templates/pod-deployment-template.yaml
@@ -1,6 +1,6 @@
 parameters:
   clusterName: ""
-  scaleup: 1000
+  scaleup: 100
   os: ""
   iterations: 4
   nodeCount: 10
--- a/.pipelines/cni/load-test-templates/pod-deployment-template.yaml
+++ b/.pipelines/cni/load-test-templates/pod-deployment-template.yaml
@@ -3,6 +3,7 @@ parameters:
   scaleup: 1000
   os: ""
   iterations: 4
+  nodeCount: 10
 
 steps:
   - task: AzureCLI@1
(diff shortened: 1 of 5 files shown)


.pipelines/cni/load-test-templates/pod-deployment-template.yaml
```diff
@@ -3,6 +3,7 @@ parameters:
   scaleup: 1000
```
vipul-21, line 3: Should be 100. Otherwise the calculation exceed max pods per node

.pipelines/cni/load-test-templates/restart-node-template.yaml
```diff
@@ -1,5 +1,7 @@
 parameters:
   clusterName: ""
+  nodeCount: 10
+  scaleup: 1000
```
vipul-21, line 4: Same here.

vipul-21 (changes requested):

.pipelines/cni/load-test-templates/restart-node-template.yaml
```diff
@@ -14,13 +16,15 @@ steps:
...
+        scale=$(( ${{ parameters.scaleup }} * ${{ parameters.nodeCount }} / 2))
+        echo "Scaling the pods down to $(( $scale / ${{ parameters.nodeCount }} )) per node"
+        go test -count 1 -timeout 30m -tags load -run ^TestScaleDeployment$ -tags=load -replicas=$scale -skip-wait=true
         cd ../../../
         echo "Restarting the nodes"
         vmss_name=$(az vmss list -g MC_${clusterName}_${clusterName}_$(LOCATION) --query "[].name" -o tsv)
         make -C ./hack/aks restart-vmss AZCLI=az CLUSTER=${clusterName} REGION=$(LOCATION) VMSS_NAME=$vmss_name
         cd test/integration/load
-        go test -count 1 -timeout 30m -tags load -run ^TestScaleDeployment$ -tags=load -replicas=1000
+        go test -count 1 -timeout 30m -tags load -run ^TestScaleDeployment$ -tags=load -replicas=$(( $scale * 2 ))
```
vipul-21, line 28: So the intention here is not to scale up actually, we should have the same value `scale=$(( ${{ parameters.scaleup }} * ${{ parameters.nodeCount }} / 2))` here as well.

That will actually wait until all the pods(`scale` number) are up before we validate the state.
Since we restarted the node in between the scale up we want to make sure the pods are up and running before we move to next step.

vipul-21 (dismissed):

camrynl (approved):
looks good 👍

camrynl (approved):

vipul-21 (approved):

PR #2138, opened 2023-08-11
docs: proposal for async pod delete handling

rbtr:

**Reason for Change**:

Design proposal to fix the synchronous Pod delete/ReleaseIP deadlock

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/feature/async-delete/readme.md
+++ b/docs/feature/async-delete/readme.md
@@ -5,7 +5,7 @@
 In AKS with Azure CNI, the Azure CNS service manages the CNI IPAM. The `azure-vnet` CNI plugin (and any CNI using delegated IPAM through `azure-ipam`) makes IP requests to the CNS API to request an IP during Pod creation or to release an IP during Pod deletion. The CNS API is a synchronous API, which means that the IP request is not completed until the IP is allocated or released in CNS internal IPAM state.
 
 There is a deadlock scenario possible when the CNS API is not available (due to daemonset rollouts or for other reason):
-If the Node is fully saturated with Pods (scheduled pods == maxPods), and CNS is not running (a CNS daemonset rollout _deletes_ the existing Pod, then schedules the upgraded Pod), the scheduler will attempt to preempt a low priority Pod to make room for the CNS Pod. However, with no CNS Pod currently running, the CNI delete call will fail, and the Pod will be stuck in the `Terminating` state since the CRI cannot clean up the netns. The scheduler will not be able to schedule the CNS Pod, and the Node will deadlock without manual intervention to decrease the Pod pressure.
+If the Node is fully saturated with Pods (scheduled pods == maxPods), and a CNS Pod is not present (for example, a CNS daemonset rollout _deletes_ the existing Pod, then schedules the upgraded Pod), the scheduler will attempt to preempt a low priority Pod to make room for the CNS Pod. However, with no CNS Pod currently running, the CNI delete call will fail, and the Pod will be stuck in the `Terminating` state since the CRI cannot clean up the netns. The scheduler will not be able to schedule the CNS Pod, and the Node will deadlock without manual intervention to decrease the Pod pressure.
 
 ### Proposal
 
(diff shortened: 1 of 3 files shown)


docs/feature/async-delete/readme.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Proposal
+
+To address this deadlock issue, the CNI calls to CNS to release an IP address from a Pod need to be made asynchronously with a failsafe in such a way that if CNS is unavailable, it can recover these events when it does eventually start.
+
+### Design
+
+The CNI plugins (`azure-vnet`, `azure-ipam`) will be modified to treat a non-response from CNS during IP release as a non-fatal error and execution will proceed. A positive error response will still be treated as a real error and returned to the CRI for retry.
+
+If the Pod IP release was not acknowledged by CNS, the CNI plugins will fall back to a file-system backed system to save these events. When the CNI does not get a response, it will write that Pod ID to a "release queue" directory/file, and proceed with cleaning up the Pod netns.
```
thatmattlong, line 18: nit: it will write that ~Pod ID~ container ID to a "release queue"

docs/feature/async-delete/readme.md
```diff
@@ -0,0 +1,22 @@
...
+
+When CNS starts, it will create a watch on the "release queue" directory/file, and process the Pod IDs in the queue. IPs for those Pods will then be released in CNS IPAM state.
```
ramiro-gamarra, line 20: How big can this queue get? Do you need to cap it at some point and fail CNI DEL calls if it's full?

rbtr: this will only exist IFF CNS is down/unresponsive. since CNS IPAM is in the create-pod hotpath, we would never see more than `maxPods` (<= 250, in AKS) items in this delete cache/queue

docs/feature/async-delete/readme.md
```diff
@@ -0,0 +1,22 @@
+## Asyc Delete
+
+### Introduction
+
+In AKS with Azure CNI, the Azure CNS service manages the CNI IPAM. The `azure-vnet` CNI plugin (and any CNI using delegated IPAM through `azure-ipam`) makes IP requests to the CNS API to request an IP during Pod createion or to release an IP during Pod deletion. The CNS API is a synchronous API, which means that the IP request is not completed until the IP is allocated or released in CNS internal IPAM state.
+
+There is a deadlock scenario possible when the CNS API is not available (due to daemonset rollouts or for other reason):
+If the Node is fully saturated with Pods (scheduled pods == maxPods), and CNS is not running (a CNS daemonset rollout _deletes_ the existing Pod, then schedules the upgraded Pod), the scheduler will attempt to preempt a low priority Pod to make room for the CNS Pod. However, with no CNS Pod currently running, the CNI delete call will fail, and the Pod will be stuck in the `Terminating` state since the CRI cannot clean up the netns. The scheduler will not be able to schedule the CNS Pod, and the Node will deadlock without manual intervention to decrease the Pod pressure.
```
tyler-lloyd, line 8: is "daemonset rollout" synonymous with "CNS restarting for any reason"? Just might be worth specifying that if the node is fully saturated this will happen if CNS is killed manually, crashes, or rolled.

tyler-lloyd, line 8: nit: the preemption of lower pri pods only happens if you have priorityClass set, right? I guess most of the time things are in the context of AKS where it is set but might not be everywhere CNS is used.

docs/feature/async-delete/readme.md
```diff
@@ -0,0 +1,22 @@
+## Asyc Delete
+
+### Introduction
+
+In AKS with Azure CNI, the Azure CNS service manages the CNI IPAM. The `azure-vnet` CNI plugin (and any CNI using delegated IPAM through `azure-ipam`) makes IP requests to the CNS API to request an IP during Pod createion or to release an IP during Pod deletion. The CNS API is a synchronous API, which means that the IP request is not completed until the IP is allocated or released in CNS internal IPAM state.
```
tyler-lloyd, line 5: nit typo
```suggestion
In AKS with Azure CNI, the Azure CNS service manages the CNI IPAM. The `azure-vnet` CNI plugin (and any CNI using delegated IPAM through `azure-ipam`) makes IP requests to the CNS API to request an IP during Pod creation or to release an IP during Pod deletion. The CNS API is a synchronous API, which means that the IP request is not completed until the IP is allocated or released in CNS internal IPAM state.
```

docs/feature/async-delete/readme.md
rbtr: those are not entirely interchangeable; I'm using "daemonset rollout" as a case for "the CNS pod is recreated". I'm not sure if it's possible for that to happen outside of a DS updated, except for someone manually doing a delete Pod? a restart of the CNS process/the same Pod will not hit this issue since it's a problem with scheduling the new Pod.

rbtr: that's true, but do we need to consider that broader case where CNS is _never_ scheduled because it can't preempt a running Pod to make room? I'm tempted to throw that in: user-error, wontfix

camrynl (approved):

PR #2141, opened 2023-08-14
ci: Add CNIv2 Linux to Load Test pipeline

jpayne3506:
**Reason for Change**:

Add CNIv2 (swift and overlay) to Load Test pipeline. 
Refactor existing load-test templates to contain less stages for further CI. 
Removed make revision in order avoid "merge to master" conflicts. 
NPM|CNI integration updated for windows and linux instead of using exclusively one branch.
Changed the order of testing as NPM was reporting false negatives after scale test. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Work from #2109, #2136, #2181 pulled from this PR in order for quicker reviews. Please look at comments from those PR's to amplify this one.

### DIFF
--- a/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
+++ b/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
@@ -99,32 +99,14 @@ stages:
             parameters:
               clusterName: ${{ parameters.clusterName }}-$(commitID)
               restartCase: "true"
-              continueOnError: true
+              cnsManagedEndpoint: true
       - job: restart_cns
         displayName: "Restart and Validate CNS"
         dependsOn: restart_nodes
         steps:
-          - task: AzureCLI@1
-            inputs:
-              azureSubscription: $(BUILD_VALIDATIONS_SERVICE_CONNECTION)
-              scriptLocation: "inlineScript"
-              scriptType: "bash"
-              addSpnToEnvironment: true
-              inlineScript: |
-                make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(commitID)
-                kubectl get pod -owide -A
-                echo "Validate pod IP assignment before CNS restart"
-                make test-validate-state
-                echo "restart CNS"
-                kubectl rollout restart ds azure-cns -n kube-system
-                kubectl rollout status ds azure-cns -n kube-system
-                sleep 15 # curl -ing after restart without a delay is flaky
-                kubectl get pod -owide -A
-                echo "Validate pod IP assignment after CNS restart"
-                make test-validate-state # Runs a curl command from CNS Pod
-            name: "restartCNS"
-            displayName: "Restart CNS and Validate pods"
-            retryCountOnTaskFailure: 3
+          - template: ../load-test-templates/restart-cns-template.yaml
+            parameters:
+              clusterName: ${{ parameters.clusterName }}-$(commitID)
       - job: cni_tests
         displayName: "Cilium Test"
         dependsOn: restart_cns
--- a/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
+++ b/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
@@ -99,7 +99,7 @@ stages:
             parameters:
               clusterName: ${{ parameters.clusterName }}-$(commitID)
               restartCase: "true"
-              cnsManagedEndpoint: true
+              cnsManagedEndpoint: "true"
       - job: restart_cns
         displayName: "Restart and Validate CNS"
         dependsOn: restart_nodes
@@ -107,6 +107,8 @@ stages:
           - template: ../load-test-templates/restart-cns-template.yaml
             parameters:
               clusterName: ${{ parameters.clusterName }}-$(commitID)
+              scaleup: ${CILIUM_SCALEUP}
+              nodeCount: ${{ parameters.nodeCount }}
       - job: cni_tests
         displayName: "Cilium Test"
         dependsOn: restart_cns
--- a/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
+++ b/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
@@ -7,10 +7,12 @@ parameters:
   vmSize: "Standard_DS4_v2"
 
 stages:
-
-  - stage: createAKScluster
-    dependsOn: ${{ parameters.dependsOn }}
-    displayName: "AKS Cluster with Cilium"
+  - stage: create_${{ parameters.name }}
+    variables:
+      commitID: $[ stagedependencies.setup.env.outputs['SetEnvVars.commitID'] ]
+    dependsOn:
+      - setup
+    displayName: "Create Cluster - ${{ parameters.clusterName }}"
     jobs:
       - job: create_aks_cluster_with_${{ parameters.name }}
         pool:
@@ -19,16 +21,22 @@ stages:
           - template: ../load-test-templates/create-cluster-template.yaml
             parameters:
               clusterType: ${{ parameters.clusterType }}
-              clusterName: ${{ parameters.clusterName }}
+              clusterName: ${{ parameters.clusterName }}-$(commitID)
               nodeCount: ${{ parameters.nodeCount }}
               vmSize: ${{ parameters.vmSize }}
-  - stage: install_cilium
-    dependsOn: createAKScluster
-    displayName: "Install Cilium on AKS Overlay"
+              region: $(LOCATION)
+
+  - stage: ${{ parameters.name }}
+    variables:
+      commitID: $[ stagedependencies.setup.env.outputs['SetEnvVars.commitID'] ]
+    pool:
+      name: "$(BUILD_POOL_NAME_DEFAULT)"
+    dependsOn:
+      - create_${{ parameters.name }}
+      - setup
+    displayName: "Cilium Test - ${{ parameters.name }}"
     jobs:
       - job: deploy_cilium_components
-        pool:
-          name: "$(BUILD_POOL_NAME_DEFAULT)"
         steps:
           - task: AzureCLI@1
             displayName: "Install Cilium, CNS, and ip-masq-agent"
@@ -40,7 +48,7 @@ stages:
               inlineScript: |
                 set -ex
                 az extension add --name aks-preview
-                make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(make revision)
+                make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(commitID)
                 ls -lah
                 pwd
                 kubectl cluster-info
@@ -64,17 +72,13 @@ stages:
                 kubectl create configmap config-reconcile.yaml
                 cd ../../../..
                 kubectl get po -owide -A
-  - stage: pod_deployment
-    dependsOn: install_cilium
-    displayName: "Pod Deployment"
-    jobs:
       - job: deploy_pods
-        pool:
-          name: "$(BUILD_POOL_NAME_DEFAULT)"
+        displayName: "Scale Test"
+        dependsOn: deploy_cilium_components
         steps:
           - template: ../load-test-templates/pod-deployment-template.yaml
             parameters:
-              clusterName: ${{ parameters.clusterName }}
+              clusterName: ${{ parameters.clusterName }}-$(commitID)
               scaleup: ${CILIUM_SCALEUP}
               os: linux
               iterations: ${CILIUM_ITERATIONS}
@@ -89,18 +93,14 @@ stages:
         steps:
           - template: ../load-test-templates/validate-state-template.yaml
             parameters:
-              clusterName: ${{ parameters.clusterName }}
-  - stage: restart_nodes
-    dependsOn: validate_state
-    displayName: "Restart Node"
-    jobs:
+              clusterName: ${{ parameters.clusterName }}-$(commitID)
       - job: restart_nodes
-        pool:
-          name: "$(BUILD_POOL_NAME_DEFAULT)"
+        displayName: "Restart Test"
+        dependsOn: deploy_pods
         steps:
           - template: ../load-test-templates/restart-node-template.yaml
             parameters:
-              clusterName: ${{ parameters.clusterName }}
+              clusterName: ${{ parameters.clusterName }}-$(commitID)
               nodeCount: ${{ parameters.nodeCount }}
               scaleup: ${CILIUM_SCALEUP}
   - stage: validate_restart_state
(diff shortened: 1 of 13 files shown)


.pipelines/cni/singletenancy/linux-cniv2-template.yaml
```diff
@@ -0,0 +1,179 @@
...
+          - task: AzureCLI@1
+            inputs:
+              azureSubscription: $(TEST_SUB_SERVICE_CONNECTION)
+              scriptLocation: "inlineScript"
+              scriptType: "bash"
+              addSpnToEnvironment: true
+              inlineScript: |
+                echo "Delete load-test Namespace"
+                make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(commitID)
+                kubectl delete ns load-test
```
vipul-21, line 138: Why do we want to delete the namespace ? Isn't it good that CNS restart is tested with certain number of pods.

.pipelines/cni/singletenancy/linux-cniv2-template.yaml
```diff
@@ -0,0 +1,179 @@
...
+                  echo "Start Integration Tests on Overlay Cluster"
+                  echo "deploy ip-masq-agent for overlay"
+                  make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(commitID)
+                  kubectl apply -f test/integration/manifests/ip-masq-agent/ip-masq-agent.yaml --validate=false
+                  cd test/integration/manifests/ip-masq-agent/
+                  kubectl apply -f config-custom.yaml
+                  kubectl apply -f config-reconcile.yaml
+                  cd ../../../..
+                  kubectl get po -owide -A
+                  sudo -E env "PATH=$PATH" make test-integration CNS_VERSION=$(cnsVersion) CNI_DROPGZ_VERSION=$(dropgzVersion) INSTALL_CNS=true INSTALL_AZURE_CNI_OVERLAY=true TEST_DROPGZ=${{ parameters.testDropgz }}
```
vipul-21, line 72: Since this is CNI load test, wouldn't it make sense to test the CNI binaries with the released CNS ?

.pipelines/cni/pipeline.yaml
```diff
@@ -12,27 +12,204 @@ stages:
...
+            cni_dropgz_test_linux_amd64:
+              arch: amd64
+              name: cni-dropgz-test
+              os: linux
+            cni_dropgz_test_windows2022_amd64:
+              arch: amd64
+              name: cni-dropgz-test
+              os: windows
+              os_version: ltsc2022
+            cns_linux_amd64:
```
vipul-21, line 43: I 100% agree that we need to test CNI with cns/latest current code base. But IMO that should be tested in PR pipeline rather than CNI release pipeline. This will run once every release or so and should be responsible for testing CNI binaries with the released version of CNS/NPM

.pipelines/cni/singletenancy/linux-cniv2-template.yaml
jpayne3506: The bug I found while working this was sometime nodes would reach max pods inconsistently with one another. Leaving one node, or more, unable to schedule the privileged pod. Should we do a 90% or 95% scale on the restart case to prevent this instead?

vipul-21: Yeah that would work.
The intension is to have more than `batchsize`(i.e. 16) pods on the node when the cns restarts.

.pipelines/cni/singletenancy/linux-cniv2-template.yaml
jpayne3506: I agree, this slipped my mind when creating the variables for cns and npm. We need to ensure that there is visibility/consistency on which version should be used. The various `make *-version` do not output a version that can be consumed properly by either acnpublic or MCR. `v1.5.10-9-*` is the current output... My quick solution is setting it within the Variable Group, but that is not precise or widely used.

.pipelines/cni/singletenancy/linux-cniv2-template.yaml
jpayne3506: As an extension, we do we need to remove the pods before k8s e2e tests.

paulyufan2 (approved):
lgtm

paulyufan2 (dismissed):

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -64,64 +72,62 @@ stages:
...
+              addSpnToEnvironment: true
+              inlineScript: |
+                make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(commitID)
+                kubectl get pod -owide -A
+                echo "Validate pod IP assignment before CNS restart"
+                make test-validate-state
+                echo "restart CNS"
+                kubectl rollout restart ds azure-cns -n kube-system
+                kubectl rollout status ds azure-cns -n kube-system
+                sleep 15 # curl -ing after restart without a delay is flaky
```
vipul-21, line 121: how did we land up on 15 ?
If CNS is taking time to restart, should we flag it to make sure we have not impacted the perf of cns.

.pipelines/cni/load-test-templates/validate-state-template.yaml
```diff
@@ -12,9 +13,18 @@ steps:
       scriptType: "bash"
       addSpnToEnvironment: true
       inlineScript: |
-        make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(make revision)
+        make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}
         kubectl get pods -A
         make test-validate-state OS=${{ parameters.os }} RESTART_CASE=${{ parameters.restartCase }} CNI_TYPE=${{ parameters.cni }}
+
+        # If this step fails, test manually
```
vipul-21, line 20: nit: should we add this to the pipeline logs ? i.e `echo` it out. So that when it fails, it shows in the pipeline log and easier to convey ?

.pipelines/cni/pipeline.yaml
```diff
@@ -12,27 +12,184 @@ stages:
...
+
   - template: singletenancy/windows-cni-load-test-template.yaml
     parameters:
       name: win_cniv1
       clusterType: "windows-cniv1-up"
       clusterName: "win-cniv1"
       nodeCount: ${WINDOWS_NODE_COUNT}
-      vmSize: Standard_B2ms
-      dependsOn: setup
+      vmSize: Standard_DS4_v2
```
vipul-21, line 120: any reason we are changing the vmsize ?

.pipelines/cni/singletenancy/windows-cni-load-test-template.yaml
```diff
@@ -198,17 +106,23 @@ stages:
...
-                  echo "Cluster and resources down"
-                else
-                  echo "Deletion of resources is False"
-                fi
-            name: "CleanUpCluster"
-            displayName: "Cleanup cluster"
-    condition: always()
+                echo "Delete load-test Namespace"
+                make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(commitID)
+                kubectl delete ns load-test
```
vipul-21, line 111: wondering is this causing problem for e2e  that is why we deleted the load test namespace.?

.pipelines/cni/singletenancy/linux-cniv2-template.yaml
```diff
@@ -0,0 +1,183 @@
...
+              clusterName: ${{ parameters.clusterName }}-$(commitID)
+              nodeCount: ${{ parameters.nodeCount }}
+              scaleup: ${LINUX_SCALEUP}
+          - template: ../load-test-templates/validate-state-template.yaml
+            parameters:
+              clusterName: ${{ parameters.clusterName }}-$(commitID)
+              os: ${{ parameters.os }}
+              cni: cniv2
+              restartCase: "true"
+      - job: restart_cns
```
vipul-21, line 129: Would be nice to have template for this too ?

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -64,64 +72,62 @@ stages:
...
           - template: ../load-test-templates/validate-state-template.yaml
             parameters:
-              clusterName: ${{ parameters.clusterName }}
+              clusterName: ${{ parameters.clusterName }}-$(commitID)
               restartCase: "true"
-  - stage: connectivity_tests
-    dependsOn: validate_restart_state
-    displayName: "Connectivity Tests"
-    jobs:
+              continueOnError: true
```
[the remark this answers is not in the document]
jpayne3506: It is mentioned in the validate state template that we need to manually test this if it fails. Currently only Cilium requires this as CNS is managing its endpoint state, so we pass in `true` for this parameter as it defaults to `false` otherwise. https://github.com/Azure/azure-container-networking/blob/10a5e23383cf9746e9b17623158183013f84f480/.pipelines/cni/load-test-templates/validate-state-template.yaml#L6C1-L6C25

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
jpayne3506: 15s was arbitrary. There is no hard coded timeframe within CNS for it to be it up. It can get removed.

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -64,64 +72,62 @@ stages:
...
+                kubectl get pod -owide -A
+                echo "Validate pod IP assignment before CNS restart"
+                make test-validate-state
+                echo "restart CNS"
+                kubectl rollout restart ds azure-cns -n kube-system
+                kubectl rollout status ds azure-cns -n kube-system
+                sleep 15 # curl -ing after restart without a delay is flaky
+                kubectl get pod -owide -A
+                echo "Validate pod IP assignment after CNS restart"
+                make test-validate-state # Runs a curl command from CNS Pod
```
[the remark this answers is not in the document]
jpayne3506: Reading further comments, Line 121 will be removed and 124 comments after the make command will be removed

.pipelines/cni/pipeline.yaml
jpayne3506: NPM Tests require a larger VMSku. Was creating false negatives.

.pipelines/cni/singletenancy/windows-cni-load-test-template.yaml
jpayne3506: It is being deleted as after this CNI k8s E2E requires space on each node to run its tests. No tests are being run after this that require the use or leverage having pods active on each node.

.pipelines/cni/singletenancy/linux-cniv2-template.yaml
jpayne3506: Make sense to me!

.pipelines/cni/load-test-templates/validate-state-template.yaml
```diff
@@ -3,7 +3,7 @@ parameters:
   os: "linux"
   restartCase: "false"
   cni: "cilium"
-
+  cnsManagedEndpoint: false
```
vipul-21, line 6: nit: space between variables and steps

.pipelines/cni/load-test-templates/validate-state-template.yaml
```diff
@@ -12,9 +12,19 @@ steps:
       scriptType: "bash"
       addSpnToEnvironment: true
       inlineScript: |
-        make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(make revision)
+        if [ ${{ parameters.cnsManagedEndpoint }} == "true" ] && [ ${{ parameters.restartCase }} == "true"]
```
vipul-21, line 15: `cnsManageEndpoint` is boolean right ? Are these getting logged in the pipeline ?

.pipelines/cni/load-test-templates/validate-state-template.yaml
```diff
@@ -12,9 +13,19 @@ steps:
       scriptType: "bash"
       addSpnToEnvironment: true
       inlineScript: |
-        make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(make revision)
+        if [ ${{ parameters.cnsManagedEndpoint }} == "true" ] && [ ${{ parameters.restartCase }} == "true" ]
+        then
+          echo If this step fails, test manually.
+          echo Only fails when CNS is managing the endpoint state and state file does not existing due to node restart with 0 pods, outside of kube-system pods, scheduled on node.
```
camrynl, line 19: maybe specify pods we're looking for instead of `outside of kube-system pods`

.pipelines/cni/load-test-templates/validate-state-template.yaml
```diff
@@ -12,9 +13,19 @@ steps:
       scriptType: "bash"
       addSpnToEnvironment: true
       inlineScript: |
-        make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}-$(make revision)
+        if [ ${{ parameters.cnsManagedEndpoint }} == "true" ] && [ ${{ parameters.restartCase }} == "true" ]
+        then
+          echo If this step fails, test manually.
+          echo Only fails when CNS is managing the endpoint state and state file does not existing due to node restart with 0 pods, outside of kube-system pods, scheduled on node.
+          echo Failure is due to slow node restart which occurs after scale down. This also misses the intent of the restart node scenario.
+          echo Timing should be: Scale down > Restart nodes during scale down > complete scale down > validate.
```
camrynl, line 21: how slow is too slow to restart? would it be possible to wait longer and avoid the manual testing?

.pipelines/cni/load-test-templates/validate-state-template.yaml
jpayne3506: I will put that we are looking for load-test pods in this test.

.pipelines/cni/load-test-templates/validate-state-template.yaml
jpayne3506: The scale down operation takes roughly 3 minutes to complete. If the restart does not happen/start in that timeframe then it would be considered too slow. As of now I do not know of a way to capture the status of a VMSS restart reliably. You get provisioned or not provisioned states along with a few other fields, but nothing really confirms how the restart is proceeding.

Ideally we could put a timeout fail after X amount of time. I am just not sure if we can put a time on what that is as the restart could happen correctly, but not complete for a while as well.

camrynl (approved):

paulyufan2 (approved):

PR #2142, opened 2023-08-14
Stateless cni test branch

behzad-mir:
This is just a draft PR for Stateless CNI test purpose and reviews for ACN data Plane team and will be eventually closed and replaced by individual PRs.


network/manager.go
```diff
@@ -355,7 +427,38 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 }
 
 // DeleteEndpoint deletes an existing container endpoint.
-func (nm *networkManager) DeleteEndpoint(networkID, endpointID string) error {
+func (nm *networkManager) DeleteEndpoint(networkID, endpointID string, epInfo *EndpointInfo) error {
+
+	if nm.IsStatelessCNIMode() {
+		nw := &network{
```
behzad-mir, line 427: @tamilmani1989 
since the nm.getNetwork(networkID) is going to return nil for delete (due to not having a restore and initialize and the network state) then I have to create a nw struct in here. Any suggestion or alternative?

PR #2151, opened 2023-08-16
[Multitenancy]: Add PodNetwork field to MTPNC spec

aggarwal0009:
**Reason for Change**:

Add PN field to MTPNC spec as discussed in Matt's PR

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
+++ b/crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
@@ -18,6 +18,7 @@ import (
 // +kubebuilder:metadata:labels=managed=
 // +kubebuilder:metadata:labels=owner=
 // +kubebuilder:printcolumn:name="PodNetworkInstance",type=string,JSONPath=`.spec.podNetworkInstance`
+// +kubebuilder:printcolumn:name="PodNetwork",type=string,JSONPath=`.spec.podNetwork`
 // +kubebuilder:printcolumn:name="PodName",type=string,JSONPath=`.spec.podName`
 // +kubebuilder:printcolumn:name="NCID",type=string,JSONPath=`.status.ncID`
 // +kubebuilder:printcolumn:name="PrimaryIP",type=string,JSONPath=`.status.primaryIP`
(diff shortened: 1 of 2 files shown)


miguelgoms (approved):

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -44,6 +45,8 @@ type MultitenantPodNetworkConfigList struct {
 type MultitenantPodNetworkConfigSpec struct {
 	// name of PNI object from requesting cx pod
 	PodNetworkInstance string `json:"podNetworkInstance,omitempty"`
+	// name of PN object from requesting cx pod
+	PodNetwork string `json:"podNetwork,omitempty"`
```
rbtr, line 49: should one of the PN or PNI be optional?

miguelgoms: PNI is optional.

aggarwal0009: Doesn't omitempty means that field is optional? Is yes, should I remove omitempty from PN here?

thatmattlong (dismissed):

crd/multitenantpodnetworkconfig/api/v1alpha1/multitenantpodnetworkconfig.go
rbtr: not exactly. `omitempty` means if the field is empty/zero-valued, that it will not even be included in the JSON representation of the object. If we had a `size int` field, if the value was `0` then `size: 0` would not even be included in the JSON.
there is a [kubebuilder annotation](https://github.com/Azure/azure-container-networking/blob/906313d49fad628ccb2eb7efc3d7702639d29a5c/crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go#L64-L65) that we should be using to mark optional fields (and maybe we should be explicit about required fields also, but I think the default is required if unspecified).

aggarwal0009: // +kubebuilder:validation:Optional  - This only seems to work when 'omitempty' tag is not present. 
What I mean by aboe is,  without the optional validation annotation and omitempty tag, if I regen the CRD manifest, the field is marked as required 
But if it is tagged with omitempty, adding above annotation has no effect to the generated CRD manifest.

aggarwal0009: So basically if I add this annotation - // +kubebuilder:validation:Optional , it doesn't have any change in generated CRD manifest. Only after omitempty is removed, it makes the field optional.

miguelgoms (approved):

rbtr (approved):

PR #2158, opened 2023-08-18
fix: Remove NLM_F_EXCL flag from Netlink Delete call (#2150)

vipul-21:
* fix: Remove unix.NLM_F_EXCL from Netlink Delete Route api call

unix.NLM_F_EXCL is not expected to set in netlink delete route calls. It's no-op in older kernel and didnt return error. From kernel 5.19+, new flag NLM_F_BULK was defined with same value and serves a purpose in delete route call. This changes breaks azure cni and netlink calls fails in 5.19 kernel and onwards.

The fix is to remove setting unix.NLM_F_EXCL in netlink delete route request.

* fix: Remove unix.NLM_F_EXCL from Netlink Delete Route api call

unix.NLM_F_EXCL is not expected to set in netlink delete route calls. It's no-op in older kernel and didnt return error. From kernel 5.19+, new flag NLM_F_BULK was defined with same value and serves a purpose in delete route call. This changes breaks azure cni and netlink calls fails in 5.19 kernel and onwards.

The fix is to remove setting unix.NLM_F_EXCL in netlink delete route request.

* Add unit tests for netlink add/delete address and add/delete routes

**Reason for Change**:

Hot fix over the released version 1.4.39

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/netlink/ip_linux.go
+++ b/netlink/ip_linux.go
@@ -43,7 +43,7 @@ func (Netlink) setIPAddress(ifName string, ipAddress net.IP, ipNet *net.IPNet, a
 		flags = unix.NLM_F_CREATE | unix.NLM_F_EXCL | unix.NLM_F_ACK
 	} else {
 		msgType = unix.RTM_DELADDR
-		flags = unix.NLM_F_EXCL | unix.NLM_F_ACK
+		flags = unix.NLM_F_ACK
 	}
 
 	req := newRequest(msgType, flags)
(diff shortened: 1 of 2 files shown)


jaer-tsun (approved):
lgtm, lets check with @behzad-mir too

behzad-mir (commented):
lgtm

behzad-mir (approved):

PR #2163, opened 2023-08-21
Add NotFound error handling to NMAgent client

debecerra:
**Reason for Change**:

This PR makes some minor changes to the NMAgent client package as a part of a larger work item. This commit adds a NotFound method to the error returned by the NMAgent client. It also adds special handling to treat a 400 BadRequest as a NotFound when returned by the DeleteNetwork API call, since this case should be interpreted as a NotFound by the caller.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/nmagent/error.go
+++ b/nmagent/error.go
@@ -11,6 +11,8 @@ import (
 	pkgerrors "github.com/pkg/errors"
 )
 
+var deleteNetworkPattern = regexp.MustCompile(`/NetworkManagement/joinedVirtualNetworks/[^/]+/api-version/\d+/method/DELETE`)
+
 // ContentError is encountered when an unexpected content type is obtained from
 // NMAgent.
 type ContentError struct {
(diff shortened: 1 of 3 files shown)


nmagent/error.go
```diff
@@ -103,3 +105,13 @@ func (e Error) StatusCode() int {
 func (e Error) Unauthorized() bool {
 	return e.Code == http.StatusUnauthorized
 }
+
+// NotFound reports whether the error was produced as a result of the resource
+// not existing.
+func (e Error) NotFound() bool {
+	deleteNetworkPattern := regexp.MustCompile(`/NetworkManagement/joinedVirtualNetworks/[^/]+/api-version/\d+/method/DELETE`)
```
rbtr, line 112: do you waint to compile this regexp every time `NotFound()` is called, or should we hoist it in to a higher scope so that it's only compiled once?

debecerra: The regex doesn't need to be compiled every time, moved it to a higher scope.

rbtr (approved):
lgtm, want @timraymond for final signoff

timraymond (approved):
It's a bit regrettable that the `die` func in there grows by another parameter, but I'm not yet certain what a better signature would look like. I had thought about something like `func die(resp *http.Response, req *http.Request)`... but that also doesn't feel entirely right.

Since I don't have a better suggestion, usually the answer is to wait and see how this evolves more over time. This looks fine for now given that :)

PR #2175, opened 2023-08-23
feat: Add NMAgent Delete Network toggle to UnpublishNetworkContainer path

debecerra:
**Reason for Change**:

NMAgent is currently rolling out a change that exposes a Delete Network endpoint to clean up vnet subscriptions. This PR is part of larger change with the end goal of invoking the NMAgent Delete Network endpoint after all NCs in a vnet have been deleted from the node. 

This change adds a `shouldDeleteNetwork` toggle to the UnpublishNetworkContainer API route which will allow DNC to tell CNS if a NMAgent Delete Network request should be made after the NC has been unpublished. Determining whether this is the final NC (meaning the vnet can/should be deleted) will be handled by DNC in a subsequent PR.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -554,6 +554,10 @@ type UnpublishNetworkContainerRequest struct {
 	JoinNetworkURL                    string
 	DeleteNetworkContainerURL         string
 	DeleteNetworkContainerRequestBody []byte
+
+	// ShouldDeleteNetwork specifies whether the virtual network should be deleted via
+	// NMAgent after the network container has been unpublished.
+	ShouldDeleteNetwork bool
 }
 
 func (u UnpublishNetworkContainerRequest) String() string {
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -555,7 +555,7 @@ type UnpublishNetworkContainerRequest struct {
 	DeleteNetworkContainerURL         string
 	DeleteNetworkContainerRequestBody []byte
 
-	// ShouldDeleteNetwork specifies whether the virtual network should be deleted via
+	// ShouldDeleteNetwork specifies whether the virtual network should be deleted from
 	// NMAgent after the network container has been unpublished.
 	ShouldDeleteNetwork bool
 }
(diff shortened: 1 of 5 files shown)


cns/NetworkContainerContract.go
```diff
@@ -554,6 +554,10 @@ type UnpublishNetworkContainerRequest struct {
 	JoinNetworkURL                    string
 	DeleteNetworkContainerURL         string
 	DeleteNetworkContainerRequestBody []byte
+
+	// ShouldDeleteNetwork specifies whether the virtual network should be deleted from
+	// NMAgent after the network container has been unpublished.
+	ShouldDeleteNetwork bool
```
ramiro-gamarra, line 560: - Is "delete network" the correct wording? Is nmagent deleting a network from pubsub or unsubscribing from it?
- How does DNC know to set this flag?

debecerra: "Delete Network" is the name of the API that is documented and exposed by NMAgent. However, based on my conversations with @ashvindeodhar, my understanding is that invoking this API essentially indicates to NMAgent that it should _unsubscribe_ from the virtual network. 

When forming the `UnpublishNetworkContainer` request to CNS, DNC will get the node entity for the network container that is currently being unpublished. From that node entity, it uses the `DeployedNetworkContainerMap` to get the count of NCs in the same vnet on the current node. If this count is 1, then the NC currently being unpublished is the final NC from that vnet on the node, so it is safe to "delete" the virtual network once the NC has been unpublished. I have a draft PR with a sketch of the DNC changes open that I can share if you'd like more info. This approach is still a work in progress and open to feedback.

ramiro-gamarra: Do you have any links to that documentation? My fear is any misunderstanding that may come from using incorrect terminology in this case. No network is being deleted, this is just an indication to nmagent that it should unsubscribe from a network (even the corresponding initial action within CNS code doesn't say "create network", it says "join network"). My suggestion would be to keep our code accurate in terms of what's actually happening, despite possibly incorrect naming from the NMagent side. 

I'll defer to @ashvindeodhar on this though, in case this has already been discussed and agreed on.

debecerra:
After discussions with the team, it's been determined that these computations (to determine if the network should be deleted from NMAgent) need to occur on the CNS side to avoid race conditions, so the current approach will be abandoned.

PR #2181, opened 2023-08-24
ci: Improve CreateValidator and restartNetworkCmd reliability 

jpayne3506:
**Reason for Change**:

CreateVaildator was erroneously using deleted privileged-daemonset pods on restart as it was succeeding on WaitForPodsRunning due to it only checking if Pods are ready, not the daemonset. Additionally, the required functionality has been added and improved the existing k8sutils package.

restartNetworkCmd was flakey based on what characters would be flagged within the POST request. Removing the single quotes removed this error.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/internal/k8sutils/utils.go
```diff
@@ -267,6 +267,47 @@ func WaitForPodDeployment(ctx context.Context, clientset *kubernetes.Clientset,
...
+	daemonsetClient := clientset.AppsV1().DaemonSets(namespace)
+	checkPodDaemonsetFn := func() error {
+		daemonset, err := daemonsetClient.Get(ctx, daemonsetName, metav1.GetOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "could not get daemonset %s", daemonsetName)
+		}
+
+		if daemonset.Status.NumberReady == 0 && daemonset.Status.DesiredNumberScheduled == 0 {
+			// Capture daemonset restart. Restart sets every numerical status to 0.
+			log.Printf("daemonset %s has %d pods in ready status, expected %d", daemonsetName, daemonset.Status.NumberReady, daemonset.Status.DesiredNumberScheduled)
```
vipul-21, line 281: Won't this always print 0 for ready and expected ?
better to log something like daemonset is in restart phase or something.

test/internal/k8sutils/utils.go
```diff
@@ -267,6 +267,47 @@ func WaitForPodDeployment(ctx context.Context, clientset *kubernetes.Clientset,
...
+			return errors.New("daemonset does not have the expected number of ready state pods")
+		}
+
+		podList, err := podsClient.List(ctx, metav1.ListOptions{LabelSelector: podLabelSelector})
+		if err != nil {
+			return errors.Wrapf(err, "could not list pods with label selector %s", podLabelSelector)
+		}
+
+		if daemonset.Status.NumberUnavailable > int32(0) {
+			return errors.New("some pods have not been running for long enough in ")
```
vipul-21, line 297: Curious to know what this tells about the pod ? Is this an error or just something in progress and the number will decrease ?

test/internal/k8sutils/utils.go
```diff
@@ -267,6 +267,47 @@ func WaitForPodDeployment(ctx context.Context, clientset *kubernetes.Clientset,
...
+		}
+
+		log.Printf("daemonset %s has %d pods in ready status, expected %d", daemonsetName, len(podList.Items), daemonset.Status.CurrentNumberScheduled)
+		if len(podList.Items) != int(daemonset.Status.CurrentNumberScheduled) {
```
vipul-21, line 301: We should compare with expected value as well. During downgrade i have seen(atleast for deployment) that the `.AvailableReplicas` is updated but pods were still present in the terminating state.

test/internal/k8sutils/utils.go
```diff
@@ -267,6 +267,47 @@ func WaitForPodDeployment(ctx context.Context, clientset *kubernetes.Clientset,
...
+		if daemonset.Status.NumberReady == 0 && daemonset.Status.DesiredNumberScheduled == 0 {
+			// Capture daemonset restart. Restart sets every numerical status to 0.
+			log.Printf("daemonset %s has %d pods in ready status, expected %d", daemonsetName, daemonset.Status.NumberReady, daemonset.Status.DesiredNumberScheduled)
+			return errors.New("daemonset has not correctly scheduled the right number of pods")
+		}
+
+		if daemonset.Status.NumberReady != daemonset.Status.DesiredNumberScheduled {
+			// Provide real-time daemonset availability to console
+			log.Printf("daemonset %s has %d pods in ready status, expected %d", daemonsetName, daemonset.Status.NumberReady, daemonset.Status.DesiredNumberScheduled)
+			return errors.New("daemonset does not have the expected number of ready state pods")
```
paulyufan2, line 288: can we log expected number of ready state pods?

test/internal/k8sutils/utils.go
```diff
@@ -267,6 +267,47 @@ func WaitForPodDeployment(ctx context.Context, clientset *kubernetes.Clientset,
...
+	checkPodDaemonsetFn := func() error {
+		daemonset, err := daemonsetClient.Get(ctx, daemonsetName, metav1.GetOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "could not get daemonset %s", daemonsetName)
+		}
+
+		if daemonset.Status.NumberReady == 0 && daemonset.Status.DesiredNumberScheduled == 0 {
+			// Capture daemonset restart. Restart sets every numerical status to 0.
+			log.Printf("daemonset %s has %d pods in ready status, expected %d", daemonsetName, daemonset.Status.NumberReady, daemonset.Status.DesiredNumberScheduled)
+			return errors.New("daemonset has not correctly scheduled the right number of pods")
```
paulyufan2, line 282: I am thinking whether we should use all errors.Wraf() here?

test/internal/k8sutils/utils.go
jpayne3506: I saw these lines as a way to output to terminal the final state before passing and should never be triggered, as the previous checks require that the pods are ready. `daemonset.Status.NumberReady` is a more comparable to status to `.AvailableReplicas`, so I will replace `CurrentNumberScheduled` with it.

test/internal/k8sutils/utils.go
jpayne3506: The line `"daemonset %s has %d pods in ready status, expected %d"` shows the number ready and the number expected.

test/internal/k8sutils/utils.go
jpayne3506: The errors are originating from here. We shouldn't use wrapf() as it will capture the empty value from any error call before it.

vipul-21 (approved):

paulyufan2 (approved):

PR #2183, opened 2023-08-24
feat: CNS/CNI async pod delete

camrynl:
**Reason for Change**:

This change is to avoid deadlock when a node is fully saturated, and CNS is down. 
CNI stores the intended deletes (named by containerIDs) in a filesystem when their IPs cannot be released because CNS is unavailable. The filesystem watcher starts up when CNS starts, and will remove any existing items in the directory. It will continue to check for new files to delete as CNS runs.  

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
tested with byocni swift, byocni cilium, and overlay + windows nodepool 

Split into 2 PRs, azure-ipam changes will come after this change is merged due to dependency conflicts. 

Async delete documentation [here](https://github.com/Azure/azure-container-networking/tree/master/docs/feature/async-delete).

### DIFF
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	cniPath = "/var/run/azure-vnet"
+	watcherPath = "/var/run/azure-vnet"
 )
 
 // IPAMPlugin is the struct for the delegated azure-ipam plugin
@@ -139,12 +139,11 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
 	// cnsClient enforces it own timeout
 	if err := p.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
 		if errors.Is(err, &cnscli.ConnectionFailureErr{}) {
-			p.logger.Info("---Found connection failure. Add file to watcher")
-			addErr := fsnotify.WatcherAddFile(args.ContainerID, cniPath)
+			addErr := fsnotify.WatcherAddFile(args.ContainerID, watcherPath)
 			if addErr != nil {
-				p.logger.Error("---Failed to Add file to watcher", zap.Error(addErr))
+				p.logger.Error("Failed to Add file to watcher", zap.Error(addErr))
 			} else {
-				p.logger.Info("---Add file success")
+				p.logger.Info("File successfully added to watcher directory")
 			}
 		} else {
 			p.logger.Error("Failed to release IP address from CNS", zap.Error(err), zap.Any("request", req))
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -19,7 +19,8 @@ import (
 )
 
 const (
-	watcherPath = "/var/run/azure-vnet"
+	watcherPath      = "/var/run/azure-vnet"
+	watcherDirectory = "/deleteIDs"
 )
 
 // IPAMPlugin is the struct for the delegated azure-ipam plugin
@@ -139,9 +140,10 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
 	// cnsClient enforces it own timeout
 	if err := p.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
 		if errors.Is(err, &cnscli.ConnectionFailureErr{}) {
-			addErr := fsnotify.WatcherAddFile(args.ContainerID, watcherPath)
+			addErr := fsnotify.WatcherAddFile(args.ContainerID, watcherPath, watcherDirectory)
 			if addErr != nil {
-				p.logger.Error("Failed to Add file to watcher", zap.Error(addErr))
+				p.logger.Error("Failed to add file to watcher", zap.Error(addErr))
+				return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "Failed to add file to watcher")
 			} else {
 				p.logger.Info("File successfully added to watcher directory")
 			}
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -9,6 +9,8 @@ import (
 	"github.com/Azure/azure-container-networking/azure-ipam/internal/buildinfo"
 	"github.com/Azure/azure-container-networking/azure-ipam/ipconfig"
 	"github.com/Azure/azure-container-networking/cns"
+	cnscli "github.com/Azure/azure-container-networking/cns/client"
+	"github.com/Azure/azure-container-networking/cns/fsnotify"
 	cniSkel "github.com/containernetworking/cni/pkg/skel"
 	cniTypes "github.com/containernetworking/cni/pkg/types"
 	types100 "github.com/containernetworking/cni/pkg/types/100"
@@ -132,8 +134,12 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
 	p.logger.Debug("Making request to CNS")
 	// cnsClient enforces it own timeout
 	if err := p.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
-		p.logger.Error("Failed to release IP address from CNS", zap.Error(err), zap.Any("request", req))
-		return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "failed to release IP address from CNS")
+		if errors.Is(err, &cnscli.ConnectionFailureErr{}) {
+			fsnotify.WatcherAddFile(args.ContainerID)
+		} else {
+			p.logger.Error("Failed to release IP address from CNS", zap.Error(err), zap.Any("request", req))
+			return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "failed to release IP address from CNS")
+		}
 	}
 
 	p.logger.Info("DEL success")
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -18,6 +18,10 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	cniPath = "/var/run/azure-vnet"
+)
+
 // IPAMPlugin is the struct for the delegated azure-ipam plugin
 // https://www.cni.dev/docs/spec/#section-4-plugin-delegation
 type IPAMPlugin struct {
@@ -135,7 +139,13 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
 	// cnsClient enforces it own timeout
 	if err := p.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
 		if errors.Is(err, &cnscli.ConnectionFailureErr{}) {
-			fsnotify.WatcherAddFile(args.ContainerID)
+			p.logger.Info("---Found connection failure. Add file to watcher")
+			addErr := fsnotify.WatcherAddFile(args.ContainerID, cniPath)
+			if addErr != nil {
+				p.logger.Error("---Failed to Add file to watcher", zap.Error(addErr))
+			} else {
+				p.logger.Info("---Add file success")
+			}
 		} else {
 			p.logger.Error("Failed to release IP address from CNS", zap.Error(err), zap.Any("request", req))
 			return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "failed to release IP address from CNS")
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -9,18 +9,11 @@ import (
 	"github.com/Azure/azure-container-networking/azure-ipam/internal/buildinfo"
 	"github.com/Azure/azure-container-networking/azure-ipam/ipconfig"
 	"github.com/Azure/azure-container-networking/cns"
-	cnscli "github.com/Azure/azure-container-networking/cns/client"
-	"github.com/Azure/azure-container-networking/cns/fsnotify"
 	cniSkel "github.com/containernetworking/cni/pkg/skel"
 	cniTypes "github.com/containernetworking/cni/pkg/types"
 	types100 "github.com/containernetworking/cni/pkg/types/100"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
-)
-
-const (
-	watcherPath      = "/var/run/azure-vnet"
-	watcherDirectory = "/deleteIDs"
 )
 
 // IPAMPlugin is the struct for the delegated azure-ipam plugin
@@ -139,18 +132,8 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
 	p.logger.Debug("Making request to CNS")
 	// cnsClient enforces it own timeout
 	if err := p.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
-		if errors.Is(err, &cnscli.ConnectionFailureErr{}) {
-			addErr := fsnotify.WatcherAddFile(args.ContainerID, watcherPath, watcherDirectory)
-			if addErr != nil {
-				p.logger.Error("Failed to add file to watcher", zap.Error(addErr))
-				return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "Failed to add file to watcher")
-			} else {
-				p.logger.Info("File successfully added to watcher directory")
-			}
-		} else {
-			p.logger.Error("Failed to release IP address from CNS", zap.Error(err), zap.Any("request", req))
-			return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "failed to release IP address from CNS")
-		}
+		p.logger.Error("Failed to release IP address from CNS", zap.Error(err), zap.Any("request", req))
+		return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "failed to release IP address from CNS")
 	}
 
 	p.logger.Info("DEL success")
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
-	"time"
 
 	"github.com/Azure/azure-container-networking/cni"
 	"github.com/Azure/azure-container-networking/cni/util"
@@ -331,16 +330,11 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 		} else {
 			var connectionErr *cnscli.ConnectionFailureErr
 			if errors.As(err, &connectionErr) {
-				go func() {
-					for {
-						addErr := fsnotify.AddFile(ipConfigs.PodInterfaceID, args.ContainerID, watcherPath)
-						if addErr != nil {
-							logger.Error("Failed to add file to watcher", zap.String("podInterfaceID", ipConfigs.PodInterfaceID), zap.String("containerID", args.ContainerID), zap.Error(addErr))
-							time.Sleep(time.Second)
-							continue
-						}
-					}
-				}()
+				addErr := fsnotify.AddFile(ipConfigs.PodInterfaceID, args.ContainerID, watcherPath)
+				if addErr != nil {
+					logger.Error("Failed to add file to watcher", zap.String("podInterfaceID", ipConfigs.PodInterfaceID), zap.String("containerID", args.ContainerID), zap.Error(addErr))
+					return errors.Wrap(addErr, fmt.Sprintf("failed to add file to watcher with containerID %s and podInterfaceID %s", args.ContainerID, ipConfigs.PodInterfaceID))
+				}
 			} else {
 				logger.Error("Failed to release IP address",
 					zap.String("infracontainerid", ipConfigs.InfraContainerID),
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -330,9 +330,9 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 		} else {
 			var connectionErr *cnscli.ConnectionFailureErr
 			if errors.As(err, &connectionErr) {
-				addErr := fsnotify.AddFile(args.ContainerID, watcherPath)
+				addErr := fsnotify.AddFile(ipConfigs.PodInterfaceID, args.ContainerID, watcherPath)
 				if addErr != nil {
-					logger.Error("Failed to add file to watcher", zap.String("containerID", args.ContainerID), zap.Error(addErr))
+					logger.Error("Failed to add file to watcher", zap.String("podInterfaceID", ipConfigs.PodInterfaceID), zap.String("containerID", args.ContainerID), zap.Error(addErr))
 				}
 			} else {
 				logger.Error("Failed to release IP address",
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/Azure/azure-container-networking/cni"
 	"github.com/Azure/azure-container-networking/cni/util"
@@ -330,10 +331,16 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 		} else {
 			var connectionErr *cnscli.ConnectionFailureErr
 			if errors.As(err, &connectionErr) {
-				addErr := fsnotify.AddFile(ipConfigs.PodInterfaceID, args.ContainerID, watcherPath)
-				if addErr != nil {
-					logger.Error("Failed to add file to watcher", zap.String("podInterfaceID", ipConfigs.PodInterfaceID), zap.String("containerID", args.ContainerID), zap.Error(addErr))
-				}
+				go func() {
+					for {
+						addErr := fsnotify.AddFile(ipConfigs.PodInterfaceID, args.ContainerID, watcherPath)
+						if addErr != nil {
+							logger.Error("Failed to add file to watcher", zap.String("podInterfaceID", ipConfigs.PodInterfaceID), zap.String("containerID", args.ContainerID), zap.Error(addErr))
+							time.Sleep(time.Second)
+							continue
+						}
+					}
+				}()
 			} else {
 				logger.Error("Failed to release IP address",
 					zap.String("infracontainerid", ipConfigs.InfraContainerID),
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Azure/azure-container-networking/cni/util"
 	"github.com/Azure/azure-container-networking/cns"
 	cnscli "github.com/Azure/azure-container-networking/cns/client"
+	"github.com/Azure/azure-container-networking/cns/fsnotify"
 	"github.com/Azure/azure-container-networking/iptables"
 	"github.com/Azure/azure-container-networking/network"
 	"github.com/Azure/azure-container-networking/network/networkutils"
@@ -24,6 +25,7 @@ var (
 	errEmptyCNIArgs    = errors.New("empty CNI cmd args not allowed")
 	errInvalidArgs     = errors.New("invalid arg(s)")
 	overlayGatewayV6IP = "fe80::1234:5678:9abc"
+	watcherPath        = "/var/run/azure-vnet/deleteIDs"
 )
 
 type CNSIPAMInvoker struct {
@@ -318,16 +320,26 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 
 			if err = invoker.cnsClient.ReleaseIPAddress(context.TODO(), ipConfig); err != nil {
 				// if the old API fails as well then we just return the error
+
 				logger.Error("Failed to release IP address from CNS using ReleaseIPAddress ",
 					zap.String("infracontainerid", ipConfigs.InfraContainerID),
 					zap.Error(err))
+
 				return errors.Wrap(err, fmt.Sprintf("failed to release IP %v using ReleaseIPAddress with err ", ipConfig.DesiredIPAddress)+"%w")
 			}
 		} else {
-			logger.Error("Failed to release IP address",
-				zap.String("infracontainerid", ipConfigs.InfraContainerID),
-				zap.Error(err))
-			return errors.Wrap(err, fmt.Sprintf("failed to release IP %v using ReleaseIPs with err ", ipConfigs.DesiredIPAddresses)+"%w")
+			var connectionErr *cnscli.ConnectionFailureErr
+			if errors.As(err, &connectionErr) {
+				addErr := fsnotify.AddFile(args.ContainerID, watcherPath)
+				if addErr != nil {
+					logger.Error("Failed to add file to watcher", zap.String("containerID", args.ContainerID), zap.Error(addErr))
+				}
+			} else {
+				logger.Error("Failed to release IP address",
+					zap.String("infracontainerid", ipConfigs.InfraContainerID),
+					zap.Error(err))
+				return errors.Wrap(err, fmt.Sprintf("failed to release IP %v using ReleaseIPs with err ", ipConfigs.DesiredIPAddresses)+"%w")
+			}
 		}
 	}
 
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -1078,7 +1078,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 					if errors.As(err, &connectionErr) {
 						addErr := fsnotify.AddFile(args.ContainerID, watcherPath)
 						if addErr != nil {
-							log.Logger.Error("Failed to add file to watcher", zap.String("containerID", args.ContainerID), zap.Error(addErr))
+							logger.Error("Failed to add file to watcher", zap.String("containerID", args.ContainerID), zap.Error(addErr))
 							return plugin.RetriableError(fmt.Errorf("failed to add file to watcher: %w", addErr))
 						}
 					} else {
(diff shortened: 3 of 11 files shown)


cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/Azure/azure-container-networking/azure-ipam/logger"
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/client"
+	"github.com/fsnotify/fsnotify"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+)
+
+const (
+	directory = "/deleteIDs"
```
neaggarwMS, line 17: Is this hardcoded? How are we going to solve it in Windows?

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,146 @@
...
+		return errors.Wrapf(err, "Error deleting file")
+	}
+	file.Close()
+	return nil
+}
+
+// call cns ReleaseIPs
+func (w *Watcher) releaseIP(containerID string) {
+	ipconfigreq := cns.IPConfigsRequest{InfraContainerID: containerID}
+	w.CnsClient.ReleaseIPs(context.Background(), ipconfigreq)
```
neaggarwMS, line 143: still need to do err validation and delete only if it returns nil

azure-ipam/ipam.go
```diff
@@ -132,8 +134,12 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
 	p.logger.Debug("Making request to CNS")
 	// cnsClient enforces it own timeout
 	if err := p.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
-		p.logger.Error("Failed to release IP address from CNS", zap.Error(err), zap.Any("request", req))
-		return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "failed to release IP address from CNS")
+		if errors.Is(err, &cnscli.ConnectionFailureErr{}) {
+			fsnotify.WatcherAddFile(args.ContainerID)
```
neaggarwMS, line 138: we still need to do err validation and return failure if not able to update the file

cni/network/network.go
```diff
@@ -1073,7 +1074,11 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 					zap.String("containerID", args.ContainerID))
 				sendEvent(plugin, fmt.Sprintf("Release ip by ContainerID (endpoint not found):%v", args.ContainerID))
 				if err = plugin.ipamInvoker.Delete(nil, nwCfg, args, nwInfo.Options); err != nil {
-					return plugin.RetriableError(fmt.Errorf("failed to release address(no endpoint): %w", err))
+					if errors.Is(err, &cnscli.ConnectionFailureErr{}) {
+						fsnotify.WatcherAddFile(args.ContainerID)
```
neaggarwMS, line 1078: same do the error validation and return failure if err != nil

neaggarwMS (changes requested):
I think i reviewed it early on, but can you ensure comments are addressed.

rbtr:
Adding requested documentation that fsnotify is widely used:
image
also by [containerd](https://github.com/containerd/containerd/blob/2a806c710a6f719ab1a497733b899f3cc23c7bd7/pkg/cri/server/cni_conf_syncer.go#L27)

azure-ipam/ipam.go
```diff
@@ -9,13 +9,19 @@ import (
 	"github.com/Azure/azure-container-networking/azure-ipam/internal/buildinfo"
 	"github.com/Azure/azure-container-networking/azure-ipam/ipconfig"
 	"github.com/Azure/azure-container-networking/cns"
+	cnscli "github.com/Azure/azure-container-networking/cns/client"
+	"github.com/Azure/azure-container-networking/cns/fsnotify"
```
tamilmani1989, line 13: can we move fsnotify package outside of cns directory?

azure-ipam/ipam.go
```diff
@@ -132,8 +138,18 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
 	p.logger.Debug("Making request to CNS")
 	// cnsClient enforces it own timeout
 	if err := p.cnsClient.ReleaseIPAddress(context.TODO(), req); err != nil {
-		p.logger.Error("Failed to release IP address from CNS", zap.Error(err), zap.Any("request", req))
-		return cniTypes.NewError(cniTypes.ErrTryAgainLater, err.Error(), "failed to release IP address from CNS")
+		if errors.Is(err, &cnscli.ConnectionFailureErr{}) {
+			p.logger.Info("---Found connection failure. Add file to watcher")
+			addErr := fsnotify.WatcherAddFile(args.ContainerID, cniPath)
+			if addErr != nil {
+				p.logger.Error("---Failed to Add file to watcher", zap.Error(addErr))
```
tamilmani1989, line 145: should we return error in this case?

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,154 @@
...
+
+type Watcher struct {
+	CnsClient *client.Client
+	logger    *zap.Logger
+}
+
+func WatchFs(w *Watcher, path string) error {
+	loggerCfg := &logger.Config{
+		Level:       "debug",
+		Filepath:    "/var/log/fsnotify-watcher.log",
```
tamilmani1989, line 28: same as neha comment. Won't work for windows

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// Create new fs watcher.
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		w.logger.Error("Error creating watcher: ", zap.Error(err))
+	}
+	defer watcher.Close()
+
+	// Start listening for events.
+	go func() {
```
tamilmani1989, line 49: add a log before watcher starts listening?

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,154 @@
...
+				w.logger.Error("Watcher Error: ", zap.Error(err))
+			}
+		}
+	}()
+
+	// Add directory where intended deletes are kept
+	dirpath := path + "/" + directory
+	err = os.MkdirAll(dirpath, 0o755)
+	if err != nil {
+		w.logger.Error("Error making directory: ", zap.Error(err))
```
tamilmani1989, line 77: shouldn't this be created in `WatcherAddFile` call? what if cni starts and call del before cns starts?

camrynl: The directory needs to be created in WatchFS before we add it to the watcher to monitor for events. If CNS is deleted the directory with intended deletes is still present and holds the containerIDs that need to be removed next

azure-ipam/ipam.go
camrynl: fsnotify is in this directory so it can act as a library to be used in CNS

common/config.go
```diff
@@ -147,4 +147,11 @@ const (
 	OptCNIConflistScenario = "cni-conflist-scenario"
 	// OptCNIConflistScenarioAlias "shorthand" for the cni conflist scenairo, see above
 	OptCNIConflistScenarioAlias = "cniconflistscenario"
+
+	// OptWatcherPath
+	OptWatcherPath      = "watcher-path"
+	OptWatcherPathAlias = "wp"
+	// OptDeleteDirectory
+	OptDeleteDirectory      = "delete-directory"
+	OptDeleteDirectoryAlias = "deletedir"
```
rbtr, line 156: we don't need flags for this, the options being in the config file is sufficient

cns/service/main.go
```diff
@@ -87,6 +89,11 @@ const (
 
 	// envVarEnableCNIConflistGeneration enables cni conflist generation if set (value doesn't matter)
 	envVarEnableCNIConflistGeneration = "CNS_ENABLE_CNI_CONFLIST_GENERATION"
+	// envVarEnableAsyncPodDelete enables CNS and CNI to use fsnotify package for async pod deletes
+	envVarEnableAsyncPodDelete = "ASYNC_POD_DELETE"
```
rbtr, line 93: same here, we don't need an env var, the config file is sufficient

cns/service/main.go
```diff
@@ -87,6 +89,11 @@ const (
...
+
+	cnsBaseURL    = "http://localhost:10090"
```
rbtr, line 95: this is the [default](https://github.com/Azure/azure-container-networking/blob/e102891d98372e28a5e39ba235e02039677e877a/cns/client/client.go#L63), so we don't need it. if we instantiate the client with `""` it will use this

rbtr: alternatively we could make the default `client.defaultURL` public -> `client.DefaultURL` and pass it in where we construct the client. then we're not polluting `main` with these vars/consts that don't really belong in the root context

cns/service/main.go
```diff
@@ -314,6 +321,20 @@ var args = acn.ArgumentList{
...
+		Type:         "string",
+		DefaultValue: "",
+	},
+	{
+		Name:         acn.OptDeleteDirectory,
+		Shorthand:    acn.OptDeleteDirectoryAlias,
+		Description:  "Watcher delete directory",
+		Type:         "string",
+		DefaultValue: "",
+	},
```
rbtr, line 337: don't need

cns/service/main.go
```diff
@@ -686,6 +707,8 @@ func main() {
 	httpRestService.SetOption(acn.OptHttpResponseHeaderTimeout, httpResponseHeaderTimeout)
 	httpRestService.SetOption(acn.OptProgramSNATIPTables, cnsconfig.ProgramSNATIPTables)
 	httpRestService.SetOption(acn.OptManageEndpointState, cnsconfig.ManageEndpointState)
+	httpRestService.SetOption(acn.OptWatcherPath, cnsconfig.WatcherPath)
+	httpRestService.SetOption(acn.OptDeleteDirectory, cnsconfig.DeleteDirectory)
```
rbtr, line 711: don't need

cns/service/main.go
```diff
@@ -805,6 +828,25 @@ func main() {
 		}
 	}
 
+	_, envEnableAsyncPodDelete := os.LookupEnv(envVarEnableAsyncPodDelete)
+	if cnsconfig.EnableAsyncPodDelete || envEnableAsyncPodDelete {
+		// Start fs watcher here
+		logger.Printf("[Azure CNS] Start fsnotify watcher for intended deletes")
+		cnsclient, err := cnsclient.New(cnsBaseURL, cnsReqTimeout) //nolint
+		if err != nil {
+			logger.Errorf("failed to initialize CNS client:%v.\n", err)
```
rbtr, line 837: should this be a fatal error?

cns/service/main.go
```diff
@@ -805,6 +828,25 @@ func main() {
...
+		}
+		w := &fsnotify.Watcher{
+			CnsClient: cnsclient,
+		}
+		watcherPath := cnsconfig.WatcherPath
+		watcherDirectory := cnsconfig.DeleteDirectory
+		err = fsnotify.WatchFs(w, watcherPath, watcherDirectory, z)
+		if err != nil {
+			logger.Errorf("Failed to start fsnotify watcher, err:%v.\n", err)
```
rbtr, line 846: same here, should we exit if this fails? i think we should be retrying, at least

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+
+type Watcher struct {
+	CnsClient cnsclient
+}
+
+func WatchFs(w *Watcher, path, directory string, logger *zap.Logger) error {
+	// Create new fs watcher.
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.Error("Error creating watcher: ", zap.Error(err))
```
rbtr, line 26: this is a structured logger, _you_ don't add kv semantics to the log strings, it does it for you.
this log line is going to look something like:
> {..."message": "Error creating watcher: ", "error": "<errstr>"}  

the message string should just be the string, like:
> > {..."message": "Error creating watcher", "error": "<errstr>"}

```suggestion
		logger.Error("Error creating watcher", zap.Error(err))
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+	// Start listening for events.
+	logger.Info("Listening for events from watcher")
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				logger.Info("Watcher: ", zap.Any("event: ", event.Name))
```
rbtr, line 39: don't use `zap.Any` when you know the type of the thing you're logging, 
the log message should be something useful, not just metadata (that should be in its own field attached to the logger)
```suggestion
				logger.Info("received event", zap.String("event", event.Name))
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+				if event.Has(fsnotify.Create) {
+					logger.Info("created file: ", zap.Any("event: ", event.Name))
```
rbtr, line 41:
```suggestion
					logger.Info("file created, triggering release", zap.String("event", event.Name))
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+					w.releaseIP(event.Name, path, directory, logger)
```
rbtr, line 42: we should log the result of this call

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+				}
+				if event.Has(fsnotify.Remove) {
+					logger.Info("removed file: ", zap.Any("event: ", event.Name))
```
rbtr, line 45:
```suggestion
					logger.Info("file deleted", zap.String("event", event.Name))
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+			logger.Info("Path to be removed from: ", zap.String("path: ", path))
+			w.releaseIP(file.Name(), path, directory, logger)
+		}
+	}
+
+	return nil
+}
+
+// WatcherAddFile creates new file using the containerID as name
+func WatcherAddFile(containerID, path, directory string) error {
```
rbtr, line 84: these methods can probably be `AddFile`/`RemoveFile` since the package is always part of the usage (ex `fsnotify.AddFile`)

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// list the directory then call ReleaseIPs
+	logger.Info("Listing directory deleteIDs: ")
+	dirContents, err := os.ReadDir(dirpath)
+	if err != nil {
+		logger.Error("Error reading deleteID directory", zap.Error(err))
+	} else {
+		for _, file := range dirContents {
+			logger.Info("File to be deleted: ", zap.String("name", file.Name()))
+			logger.Info("Path to be removed from: ", zap.String("path: ", path))
```
rbtr, line 75: collapse to single log line

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return nil
+}
+
+// WatcherAddFile creates new file using the containerID as name
+func WatcherAddFile(containerID, path, directory string) error {
+	dirpath := path + directory
+	_, err := os.Stat(dirpath)
+	if errors.Is(err, fs.ErrNotExist) {
+		return errors.Wrapf(err, "Error reading directory")
```
rbtr, line 88: don't capitalize/sentence case error strings, they get embedded in log lines and that breaks the flow when reading. the additional context on this error needs to be something like "directory must already exist"

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	filepath := dirpath + "/" + containerID
+	f, err := os.Create(filepath)
+	if err != nil {
+		return errors.Wrapf(err, "Error creating file")
+	}
+	defer f.Close()
```
rbtr, line 96: just call `f.Close()` directly instead of defering it

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,141 @@
...
+// call cns ReleaseIPs
+func (w *Watcher) releaseIP(containerID, path, directory string, logger *zap.Logger) {
+	ipconfigreq := &cns.IPConfigsRequest{InfraContainerID: containerID}
+
+	cnsReleaseErr := w.CnsClient.ReleaseIPs(context.Background(), *ipconfigreq)
+	if cnsReleaseErr != nil {
+		logger.Error("failed to release IP from watcher directory")
+	}
+
+	err := WatcherRemoveFile(containerID, path, directory)
```
rbtr, line 137: let's not remove the file _in_ the releaseIP call; rather keep releaseIP scoped to a single action, and based on the result it returns the caller may remove the file.

rbtr (changes requested):
core structure and functionality lgtm. logs need to be improved (formatting is handled by the logger, not in the strings), we've added too many ways to configure this, and I have a bunch of suggestions in Go style and best practices

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+		return errors.Wrapf(err, "error reading directory, directory must already exist")
+	}
+
+	filepath := path + "/" + containerID
+
+	// check the file exists
+	_, fileExists := os.Stat(filepath)
+	if errors.Is(fileExists, fs.ErrNotExist) {
+		return nil
+	}
```
rbtr, line 124: don't think we need to check if the file exists before we try to remove it, if it doesn't exist we should get the same `ErrNotExist` and can do this check there (and otherwise propogate the error)

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return nil
+}
+
+// AddFile creates new file using the containerID as name
+func AddFile(containerID, path string) error {
+	_, err := os.Stat(path)
+	if errors.Is(err, fs.ErrNotExist) {
+		return errors.Wrapf(err, "error reading directory, directory must already exist")
+	}
```
rbtr, line 100: probably don't need to check this existence of the dir, we can just attempt the `os.Create` and return whatever error we get if the directory does not exist already

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+
+	filepath := path + "/" + containerID
+	f, err := os.Create(filepath)
+	if err != nil {
+		return errors.Wrapf(err, "error creating file")
+	}
+	f.Close()
+	return nil
```
rbtr, line 108:
```suggestion
	return f.Close()
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+
+// call cns ReleaseIPs
+func (w *Watcher) releaseIP(containerID string) error {
+	ipconfigreq := &cns.IPConfigsRequest{InfraContainerID: containerID}
+
+	cnsReleaseErr := w.CnsClient.ReleaseIPs(context.Background(), *ipconfigreq)
+	if cnsReleaseErr != nil {
+		return errors.Wrapf(cnsReleaseErr, "error releasing IP from CNS")
+	}
+	return nil
```
rbtr, line 146: a quirk of `errors.Wrap` is that if `err` is `nil`, `errors.Wrap(err)` returns nil
```suggestion
	ipconfigreq := &cns.IPConfigsRequest{InfraContainerID: containerID}
	errors.Wrap(w.CnsClient.ReleaseIPs(context.Background(), *ipconfigreq), "error releasing IP from CNS")
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil
+	}
+
+	file, err := os.Open(filepath)
+	if err != nil {
+		return errors.Wrapf(err, "error opening file")
+	}
+
+	if err := os.Remove(filepath); err != nil {
+		return errors.Wrapf(err, "error deleting file")
```
rbtr, line 132: use `errors.Wrap`, not `errors.Wrapf`, when there is no string formatting happening.
```suggestion
		return errors.Wrap(err, "error deleting file")
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// check the file exists
+	_, fileExists := os.Stat(filepath)
+	if errors.Is(fileExists, fs.ErrNotExist) {
+		return nil
+	}
+
+	file, err := os.Open(filepath)
+	if err != nil {
+		return errors.Wrapf(err, "error opening file")
```
rbtr, line 128:
```suggestion
		return errors.Wrap(err, "error opening file")
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+	return nil
+}
+
+// call cns ReleaseIPs
+func (w *Watcher) releaseIP(containerID string) error {
+	ipconfigreq := &cns.IPConfigsRequest{InfraContainerID: containerID}
+
+	cnsReleaseErr := w.CnsClient.ReleaseIPs(context.Background(), *ipconfigreq)
+	if cnsReleaseErr != nil {
+		return errors.Wrapf(cnsReleaseErr, "error releasing IP from CNS")
```
rbtr, line 144:
```suggestion
		return errors.Wrap(cnsReleaseErr, "error releasing IP from CNS")
```

cns/service/main.go
```diff
@@ -805,6 +809,33 @@ func main() {
...
+			CnsClient: cnsclient,
+		}
+		watcherPath := cnsconfig.AsyncPodDeletePath
+
+		go func() {
+			for {
+				if err := fsnotify.WatchFs(w, watcherPath, z); err != nil {
+					logger.Errorf("Failed to start fsnotify watcher, err:%v.\n", err)
+				} else {
+					logger.Printf("Fsnotify watcher started")
```
rbtr, line 829:
```suggestion
					logger.Print("Fsnotify watcher started")
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	f.Close()
+	return nil
+}
+
+// RemoveFile removes the file based on containerID
+func RemoveFile(containerID, path string) error {
+	_, err := os.Stat(path)
+	if errors.Is(err, fs.ErrNotExist) {
+		return errors.Wrapf(err, "error reading directory, directory must already exist")
```
rbtr, line 115:
```suggestion
		return errors.Wrap(err, "error reading directory, directory must already exist")
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+func AddFile(containerID, path string) error {
+	_, err := os.Stat(path)
+	if errors.Is(err, fs.ErrNotExist) {
+		return errors.Wrapf(err, "error reading directory, directory must already exist")
+	}
+
+	filepath := path + "/" + containerID
+	f, err := os.Create(filepath)
+	if err != nil {
+		return errors.Wrapf(err, "error creating file")
```
rbtr, line 105:
```suggestion
		return errors.Wrap(err, "error creating file")
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	return nil
+}
+
+// AddFile creates new file using the containerID as name
+func AddFile(containerID, path string) error {
+	_, err := os.Stat(path)
+	if errors.Is(err, fs.ErrNotExist) {
+		return errors.Wrapf(err, "error reading directory, directory must already exist")
```
rbtr, line 99:
```suggestion
		return errors.Wrap(err, "error reading directory, directory must already exist")
```

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+			err := RemoveFile(file.Name(), path)
+			if err != nil {
+				logger.Error("failed to remove file", zap.Error(err))
+			}
+		}
+	}
+
+	return nil
+}
```
rbtr, line 93: this feels like it should be a method on the `Watcher` instead of having it passed in.
I also thing it should be a blocking call (which the caller must start the goroutine for) so that it can propagate errors from the fs library. Here's how I would thread the context (written entirely on the GH UI so the syntax might not be perfect...).
```suggestion
type Watcher struct {
	client cnsclient
	path   string
	l      *zap.Logger
}

// WatchFS starts the filesystem watcher to handle async Pod deletes.
// Blocks until the context is closed; returns underlying fsnotify errors
// if something goes fatally wrong.
func WatchFS(ctx context.Context) error {
	// Create new fs watcher.
	watcher, err := fsnotify.NewWatcher()
	if err != nil {
		logger.Error("error creating watcher", zap.Error(err))
	}
	defer watcher.Close()

	c, cancel := context.WithCancel(ctx)
	// Start listening for events.
	logger.Info("listening for events from watcher")
	go func() {
		defer cancel()
		for {
			select {
			case := <-ctx.Done():
				return
			case event, ok := <-watcher.Events:
				if !ok {
					return
				}
				logger.Info("received event", zap.String("event", event.Name))
				if event.Has(fsnotify.Create) {
					logger.Info("file created, triggering release", zap.String("event", event.Name))
					cnsReleaseErr := w.releaseIP(event.Name)
					if cnsReleaseErr != nil {
						logger.Error("failed to release IP from CNS", zap.Error(cnsReleaseErr))
					}
					deleteErr := RemoveFile(event.Name, path)
					if deleteErr != nil {
						logger.Error("failed to remove file", zap.Error(err))
					}
				}
				if event.Has(fsnotify.Remove) {
					logger.Info("file deleted", zap.String("event", event.Name))
				}
			case watcherErr, ok := <-watcher.Errors:
				if !ok {
					return
				}
				logger.Error("watcher error", zap.Error(watcherErr))
			}
		}
	}()

	// Add directory where intended deletes are kept
	err = os.MkdirAll(path, 0o755) //nolint
	if err != nil {
		logger.Error("error making directory", zap.Error(err))
	}
	err = watcher.Add(path)
	if err != nil {
		logger.Error("watcher add directory error", zap.Error(err))
	}

	// list the directory then call ReleaseIPs
	logger.Info("listing directory deleteIDs")
	dirContents, err := os.ReadDir(path)
	if err != nil {
		logger.Error("error reading deleteID directory", zap.Error(err))
	} else {
		for _, file := range dirContents {
			logger.Info("file to be deleted", zap.String("name", file.Name()))
			cnsReleaseErr := w.releaseIP(file.Name())
			if cnsReleaseErr != nil {
				logger.Error("failed to release IP from CNS", zap.Error(cnsReleaseErr))
			}
			err := RemoveFile(file.Name(), path)
			if err != nil {
				logger.Error("failed to remove file", zap.Error(err))
			}
		}
	}
	<-c.Done()
	return c.Error()
}
```

rbtr (changes requested):

cns/fsnotify/fsnotify.go
rbtr: the point of it being a blocking call is, currently we don't know if the internal goroutine is still running or has failed and exited, and if we block and use contexts to cancel we can communicate that back to the caller

cns/service/main.go
camrynl: will keep this one unchanged since logger.Print is undefined

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,147 @@
...
+				logger.Error("watcher error", zap.Error(watcherErr))
+			}
+		}
+	}()
+
+	// Add directory where intended deletes are kept
+	err = os.MkdirAll(path, 0o755) //nolint
+	if err != nil {
+		logger.Error("error making directory", zap.Error(err))
+	}
```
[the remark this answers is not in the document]
camrynl: the directory needs to be created here when the watcher starts so that there is a path to watch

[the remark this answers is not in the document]
rbtr: sure, my concern is that this may hide an issue from us: if the container is not mounting the hostpath correctly, this mkdirall will _create_ the whole path inside the containerfs. we probably don't want that, we want to expect that the path (or most of the path) already exists (because it's been mounted in), and we should fail here instead of creating the whole directory tree if it doesn't exist.

cns/fsnotify/fsnotify.go
rbtr: don't forget the 
```go 
			select {
			case := <-ctx.Done():
				return
```
in the select, it's important to exit the for/select loop when the outer context has been cancelled

cni/network/network.go
```diff
@@ -1073,7 +1076,15 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 					zap.String("containerID", args.ContainerID))
 				sendEvent(plugin, fmt.Sprintf("Release ip by ContainerID (endpoint not found):%v", args.ContainerID))
 				if err = plugin.ipamInvoker.Delete(nil, nwCfg, args, nwInfo.Options); err != nil {
-					return plugin.RetriableError(fmt.Errorf("failed to release address(no endpoint): %w", err))
+					if errors.As(err, &connectionErr) {
```
neaggarwMS, line 1079: confused, why we need to add it here? I thought it was added in other PR already (in ipam.go)

cni/network/network.go
```diff
@@ -1073,7 +1076,15 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+						addErr := fsnotify.AddFile(args.ContainerID, watcherPath)
+						if addErr != nil {
+							log.Logger.Error("Failed to add file to watcher", zap.Error(addErr))
```
neaggarwMS, line 1082: same comment as other PR, please make logs actionable by adding contextual data (like PodInfo/IPAddress for helping in debugging during livesite investigation)

cns/fsnotify/fsnotify.go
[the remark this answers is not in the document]
camrynl: so rather than mkdirall, I will update this to mkdir so the only directory that gets created is deleteIDs

cni/network/network.go
camrynl: the changes in network.go are for azure-vnet. the other PR contains the same changes for azure-ipam

rbtr: this PR is CNI/CNS, other PR is `azure-ipam` since it is a separate Go module.

rbtr, line 1079: should this be 
```suggestion
					if errors.Is(err, &cnscli.ConnectionFailureErr{}) {
```
the pointer semantics of errors.As/Is are confusing to me. but it seems wrong to compare `err` to this local `connectionErr *cnscli.ConnectionFailureErr` variable

camrynl: from what I previously gathered in search, errors.As works because it compares the actual err to type ConnectionFailureErr, and it will return true when a match is found in the error chain.

rbtr (approved):

cni/network/network.go
neaggarwMS: This is a main function that will invoke different IPAM, for now, we should only handle it inside invoker_cns.go and that too only in AKS Swift scenario as only in that scenario CNS is cleaning up those ips latebound.

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,129 @@
...
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				w.Logger.Info("received event", zap.String("event", event.Name))
+				if event.Has(fsnotify.Create) {
+					w.Logger.Info("file created, triggering release", zap.String("event", event.Name))
+					cnsReleaseErr := w.releaseIP(ctx, event.Name)
+					if cnsReleaseErr != nil {
+						w.Logger.Error("failed to release IP from CNS", zap.Error(cnsReleaseErr))
```
neaggarwMS, line 52: need to handle the error and retry

cni/network/network.go
camrynl: The initial plan was to make the change for both azure-ipam and azure-vnet. The deadlock issue can still occur in the scenario with azure-ipam, at max pod saturation the pods in delete will get stuck if CNS is down. I think it would make sense to keep the changes as is, for each respective plugin

cns/fsnotify/fsnotify.go
```diff
@@ -0,0 +1,169 @@
...
+					w.Logger.Info("attempt delete on entries in releaseIPRetry")
+					for _, entry := range releaseIPRetry {
+						retryErr := w.releaseIP(ctx, entry)
+						if retryErr != nil {
+							w.Logger.Error("failed to release IP from CNS", zap.Error(retryErr))
+						} else {
+							delete(releaseIPRetry, entry)
+							removeErr := RemoveFile(entry, w.Path)
+							if removeErr != nil {
+								w.Logger.Error("failed to remove file", zap.Error(removeErr))
```
nddq, line 96: what happen if we failed to remove the file from the fs here? I think it would make the operation doesn't feel "atomic" since even though we did successfully ReleaseIP'ed the entry, we can still see the related file in the fs, indicating that it did not get ReleaseIP'ed.

rbtr (approved):
lgtm 🚀

nddq (dismissed):
lgtm

rbtr (dismissed):

cni/network/invoker_cns.go
```diff
@@ -318,16 +320,26 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
...
 		} else {
-			logger.Error("Failed to release IP address",
-				zap.String("infracontainerid", ipConfigs.InfraContainerID),
-				zap.Error(err))
-			return errors.Wrap(err, fmt.Sprintf("failed to release IP %v using ReleaseIPs with err ", ipConfigs.DesiredIPAddresses)+"%w")
+			var connectionErr *cnscli.ConnectionFailureErr
+			if errors.As(err, &connectionErr) {
+				addErr := fsnotify.AddFile(ipConfigs.PodInterfaceID, args.ContainerID, watcherPath)
+				if addErr != nil {
+					logger.Error("Failed to add file to watcher", zap.String("podInterfaceID", ipConfigs.PodInterfaceID), zap.String("containerID", args.ContainerID), zap.Error(addErr))
```
neaggarwMS, line 335: return error and repeat if you failed ot add file, otherwise it will lead to a leak

rbtr (approved):

neaggarwMS (approved):

neaggarwMS (approved):

PR #2187, opened 2023-08-28
v4overlay windows test cases

paulyufan2:
**Reason for Change**:

This PR is to add ipv4 overlay windows test cases in ACN pipeline

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
+++ b/.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
@@ -197,8 +197,8 @@ steps:
               kubectl taint nodes $node node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
           fi
       done
-      CNS=$(make cns-version) DROPGZ=$(make cni-dropgz-version)
-      sudo -E env "PATH=$PATH" make test-load SCALE_UP=32 OS_TYPE=windows CNI_TYPE=cniv2 VALIDATE_STATEFILE=true INSTALL_CNS=true INSTALL_AZURE_CNI_OVERLAY=true VALIDATE_V4OVERLAY=true CNS_VERSION=${CNS} CNI_DROPGZ_VERSION=${DROPGZ} CLEANUP=true
+      CNS=$(make cns-version)
+      sudo -E env "PATH=$PATH" make test-load SCALE_UP=32 OS_TYPE=windows CNI_TYPE=cniv2 VALIDATE_STATEFILE=true INSTALL_CNS=true INSTALL_AZURE_CNI_OVERLAY=true VALIDATE_V4OVERLAY=true CNS_VERSION=${CNS} CNI_DROPGZ_VERSION=$(dropgzVersion) TEST_DROPGZ=${{ parameters.testDropgz }} CLEANUP=true
     name: "WindowsOverlayControlPlaneScaleTests"
     displayName: "Windows v4Overlay ControlPlane Scale Tests"
     retryCountOnTaskFailure: 3
--- a/.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
+++ b/.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
@@ -172,3 +172,41 @@ steps:
     name: "Cleanupartifactdir"
     displayName: "Cleanup artifact dir"
     condition: always()
+
+  - task: AzureCLI@2
+    inputs:
+      azureSubscription: $(BUILD_VALIDATIONS_SERVICE_CONNECTION)
+      scriptLocation: "inlineScript"
+      scriptType: "bash"
+      addSpnToEnvironment: true
+      inlineScript: |
+        set -e
+        make -C ./hack/aks windows-nodepool-up AZCLI=az SUB=$(SUB_AZURE_NETWORK_AGENT_BUILD_VALIDATIONS) CLUSTER=${{ parameters.clusterName }}
+        echo "Windows node are successfully added to v4 Overlay Cluster"
+        kubectl cluster-info
+        kubectl get node -owide
+        kubectl get po -owide -A
+    name: "Add_Windows_Node"
+    displayName: "Add windows node on v4 overlay cluster"
+
+  - script: |
+      nodeList=`kubectl get node -owide | grep Windows | awk '{print $1}'`
+      for node in $nodeList; do
+          taint=`kubectl describe node $node | grep Taints | awk '{print $2}'`
+          if [ $taint == "node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule" ]; then
+              kubectl taint nodes $node node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
+          fi
+      done
+      CNS=$(make cns-version) DROPGZ=$(make cni-dropgz-version)
+      sudo -E env "PATH=$PATH" make test-load SCALE_UP=32 OS_TYPE=windows CNI_TYPE=cniv2 VALIDATE_STATEFILE=true INSTALL_CNS=true INSTALL_AZURE_CNI_OVERLAY=true VALIDATE_V4OVERLAY=true CNS_VERSION=${CNS} CNI_DROPGZ_VERSION=${DROPGZ} CLEANUP=true
+    name: "WindowsOverlayControlPlaneScaleTests"
+    displayName: "Windows v4Overlay ControlPlane Scale Tests"
+    retryCountOnTaskFailure: 3
+
+  - script: |
+      echo "IPv4 Overlay DataPath Test"
+      cd test/integration/datapath
+      sudo -E env "PATH=$PATH" go test -count=1 datapath_windows_test.go -timeout 3m -tags connection -initbyocni true -run ^TestDatapathWin$
+    name: "WindowsV4OverlayDatapathTests"
+    displayName: "Windows v4Overlay Datapath Tests"
+    retryCountOnTaskFailure: 3
--- a/.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
+++ b/.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
@@ -206,7 +206,7 @@ steps:
   - script: |
       echo "IPv4 Overlay DataPath Test"
       cd test/integration/datapath
-      sudo -E env "PATH=$PATH" go test -count=1 datapath_windows_test.go -timeout 3m -tags connection -initbyocni true -run ^TestDatapathWin$
+      sudo -E env "PATH=$PATH" go test -count=1 datapath_windows_test.go -timeout 3m -tags connection -restartKubeproxy true -run ^TestDatapathWin$
     name: "WindowsV4OverlayDatapathTests"
     displayName: "Windows v4Overlay Datapath Tests"
     retryCountOnTaskFailure: 3
--- a/test/integration/datapath/datapath_windows_test.go
+++ b/test/integration/datapath/datapath_windows_test.go
@@ -10,21 +10,26 @@ import (
 
 	"github.com/Azure/azure-container-networking/test/internal/datapath"
 	"github.com/Azure/azure-container-networking/test/internal/kubernetes"
+	"github.com/Azure/azure-container-networking/test/validate"
 	"github.com/stretchr/testify/require"
 	apiv1 "k8s.io/api/core/v1"
 )
 
 const (
-	WindowsDeployYamlPath = "../manifests/datapath/windows-deployment.yaml"
-	podLabelKey           = "app"
-	podCount              = 2
-	nodepoolKey           = "agentpool"
+	WindowsDeployYamlPath         = "../manifests/datapath/windows-deployment.yaml"
+	IPMasqAgentYamlPath           = "../manifests/ip-masq-agent/ip-masq-agent.yaml"
+	IPMasqAgentCustomConfigMap    = "../manifests/ip-masq-agent/config-custom.yaml"
+	IPMasqAgentReconcileConfigMap = "../manifests/ip-masq-agent/config-reconcile.yaml"
+	podLabelKey                   = "app"
+	podCount                      = 2
+	nodepoolKey                   = "agentpool"
 )
 
 var (
 	podPrefix        = flag.String("podName", "datapod", "Prefix for test pods")
 	podNamespace     = flag.String("namespace", "windows-datapath-test", "Namespace for test pods")
 	nodepoolSelector = flag.String("nodepoolSelector", "npwin", "Provides nodepool as a windows Node-Selector for pods")
+	initBYOCNI       = flag.Bool("initbyocni", false, "installs ipmas agent, configmap, and restarts kubeproxy")
 )
 
 /*
@@ -48,10 +53,33 @@ Timeout context is controled by the -timeout flag.
 func setupWindowsEnvironment(t *testing.T) {
 	ctx := context.Background()
 
+	t.Log("Get REST config")
+	restConfig := kubernetes.MustGetRestConfig(t)
+
 	t.Log("Create Clientset")
 	clientset, err := kubernetes.MustGetClientset()
 	if err != nil {
 		t.Fatal(err)
+	}
+
+	if *initBYOCNI {
+		// create ipmasq configmaps
+		err := kubernetes.MustSetupConfigMap(ctx, clientset, IPMasqAgentCustomConfigMap)
+		require.NoError(t, err, "failed to setup ip masq agent custom configmap")
+		err = kubernetes.MustSetupConfigMap(ctx, clientset, IPMasqAgentReconcileConfigMap)
+		require.NoError(t, err, "failed to setup ip masq agent reconcile configmap")
+
+		// create ipmasq agent daemonset
+		ds, err := kubernetes.MustParseDaemonSet(IPMasqAgentYamlPath)
+		require.NoError(t, err, "failed to parse ipmasq agent")
+		err = kubernetes.MustCreateDaemonset(ctx, clientset.AppsV1().DaemonSets(ds.Namespace), ds)
+		require.NoError(t, err, "failed to create ipmasq daemonset")
+
+		// restart kubeproxy
+		validator, err := validate.CreateValidator(ctx, clientset, restConfig, *podNamespace, "cniv2", false, "windows")
+		require.NoError(t, err)
+		err = validator.RestartKubeProxyService(ctx)
+		require.NoError(t, err)
 	}
 
 	t.Log("Create Label Selectors")
--- a/test/integration/datapath/datapath_windows_test.go
+++ b/test/integration/datapath/datapath_windows_test.go
@@ -16,13 +16,10 @@ import (
 )
 
 const (
-	WindowsDeployYamlPath         = "../manifests/datapath/windows-deployment.yaml"
-	IPMasqAgentYamlPath           = "../manifests/ip-masq-agent/ip-masq-agent.yaml"
-	IPMasqAgentCustomConfigMap    = "../manifests/ip-masq-agent/config-custom.yaml"
-	IPMasqAgentReconcileConfigMap = "../manifests/ip-masq-agent/config-reconcile.yaml"
-	podLabelKey                   = "app"
-	podCount                      = 2
-	nodepoolKey                   = "agentpool"
+	WindowsDeployYamlPath = "../manifests/datapath/windows-deployment.yaml"
+	podLabelKey           = "app"
+	podCount              = 2
+	nodepoolKey           = "agentpool"
 )
 
 var (
--- a/test/integration/datapath/datapath_windows_test.go
+++ b/test/integration/datapath/datapath_windows_test.go
@@ -29,7 +29,7 @@ var (
 	podPrefix        = flag.String("podName", "datapod", "Prefix for test pods")
 	podNamespace     = flag.String("namespace", "windows-datapath-test", "Namespace for test pods")
 	nodepoolSelector = flag.String("nodepoolSelector", "npwin", "Provides nodepool as a windows Node-Selector for pods")
-	initBYOCNI       = flag.Bool("initbyocni", false, "installs ipmas agent, configmap, and restarts kubeproxy")
+	restartKubeproxy = flag.Bool("restartKubeproxy", false, "restarts kubeproxy on the windows node")
 )
 
 /*
@@ -62,20 +62,7 @@ func setupWindowsEnvironment(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if *initBYOCNI {
-		// create ipmasq configmaps
-		err := kubernetes.MustSetupConfigMap(ctx, clientset, IPMasqAgentCustomConfigMap)
-		require.NoError(t, err, "failed to setup ip masq agent custom configmap")
-		err = kubernetes.MustSetupConfigMap(ctx, clientset, IPMasqAgentReconcileConfigMap)
-		require.NoError(t, err, "failed to setup ip masq agent reconcile configmap")
-
-		// create ipmasq agent daemonset
-		ds, err := kubernetes.MustParseDaemonSet(IPMasqAgentYamlPath)
-		require.NoError(t, err, "failed to parse ipmasq agent")
-		err = kubernetes.MustCreateDaemonset(ctx, clientset.AppsV1().DaemonSets(ds.Namespace), ds)
-		require.NoError(t, err, "failed to create ipmasq daemonset")
-
-		// restart kubeproxy
+	if *restartKubeproxy {
 		validator, err := validate.CreateValidator(ctx, clientset, restConfig, *podNamespace, "cniv2", false, "windows")
 		require.NoError(t, err)
 		err = validator.RestartKubeProxyService(ctx)
--- a/test/integration/load/load_test.go
+++ b/test/integration/load/load_test.go
@@ -201,9 +201,11 @@ func TestValidCNSStateDuringScaleAndCNSRestart(t *testing.T) {
 	require.NoError(t, err)
 	deploymentsClient := clientset.AppsV1().Deployments(namespace)
 
-	// Create a deployment
-	err = kubernetes.MustCreateDeployment(ctx, deploymentsClient, deployment)
-	require.NoError(t, err)
+	if testConfig.Cleanup {
+		// Create a deployment
+		err = kubernetes.MustCreateDeployment(ctx, deploymentsClient, deployment)
+		require.NoError(t, err)
+	}
 
 	// Scale it up and "skipWait", so CNS restart can happen immediately after scale call is made (while pods are still creating)
 	skipWait := true
--- a/test/integration/load/load_test.go
+++ b/test/integration/load/load_test.go
@@ -21,6 +21,7 @@ type TestConfig struct {
 	Replicas          int    `env:"REPLICAS" default:"1"`
 	ValidateStateFile bool   `env:"VALIDATE_STATEFILE" default:"false"`
 	ValidateDualStack bool   `env:"VALIDATE_DUALSTACK" default:"false"`
+	ValidateV4Overlay bool   `env:"VALIDATE_V4OVERLAY" default:"false"`
 	SkipWait          bool   `env:"SKIP_WAIT" default:"false"`
 	RestartCase       bool   `env:"RESTART_CASE" default:"false"`
 	Cleanup           bool   `env:"CLEANUP" default:"false"`
@@ -107,6 +108,10 @@ func TestLoad(t *testing.T) {
 		t.Run("Validate state file", TestValidateState)
 	}
 
+	if testConfig.ValidateV4Overlay {
+		t.Run("Validate v4overlay", TestV4OverlayProperties)
+	}
+
 	if testConfig.ValidateDualStack {
 		t.Run("Validate dualstack overlay", TestDualStackProperties)
 	}
@@ -114,6 +119,8 @@ func TestLoad(t *testing.T) {
 	if testConfig.Cleanup {
 		err = kubernetes.MustDeleteDeployment(ctx, deploymentsClient, deployment)
 		require.NoError(t, err, "error deleteing load deployment")
+		err = kubernetes.WaitForPodsDelete(ctx, clientset, namespace, podLabelSelector)
+		require.NoError(t, err, "error waiting for pods to delete")
 	}
 }
 
@@ -171,7 +178,92 @@ func TestScaleDeployment(t *testing.T) {
 	if testConfig.Cleanup {
 		err = kubernetes.MustDeleteDeployment(ctx, deploymentsClient, deployment)
 		require.NoError(t, err, "error deleteing load deployment")
+		err = kubernetes.WaitForPodsDelete(ctx, clientset, namespace, podLabelSelector)
+		require.NoError(t, err, "error waiting for pods to delete")
 	}
+}
+
+func TestValidCNSStateDuringScaleAndCNSRestart(t *testing.T) {
+	clientset, err := kubernetes.MustGetClientset()
+	require.NoError(t, err)
+
+	config := kubernetes.MustGetRestConfig(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
+	defer cancel()
+
+	validator, err := validate.CreateValidator(ctx, clientset, config, namespace, testConfig.CNIType, testConfig.RestartCase, testConfig.OSType)
+	require.NoError(t, err)
+
+	err = validator.Validate(ctx)
+	require.NoError(t, err)
+
+	deployment, err := kubernetes.MustParseDeployment(noopDeploymentMap[testConfig.OSType])
+	require.NoError(t, err)
+	deploymentsClient := clientset.AppsV1().Deployments(namespace)
+
+	// Create a deployment
+	err = kubernetes.MustCreateDeployment(ctx, deploymentsClient, deployment)
+	require.NoError(t, err)
+
+	// Scale it up and "skipWait", so CNS restart can happen immediately after scale call is made (while pods are still creating)
+	skipWait := true
+	err = kubernetes.MustScaleDeployment(ctx, deploymentsClient, deployment, clientset, namespace, podLabelSelector, testConfig.ScaleUpReplicas, skipWait)
+	require.NoError(t, err)
+
+	// restart linux CNS (linux, windows)
+	err = kubernetes.RestartCNSDaemonset(ctx, clientset)
+	require.NoError(t, err)
+
+	// wait for pods to settle before checking cns state (otherwise, race between getting pods in creating state, and getting CNS state file)
+	err = kubernetes.WaitForPodDeployment(ctx, clientset, namespace, deployment.Name, podLabelSelector, testConfig.ScaleUpReplicas)
+	require.NoError(t, err)
+
+	// Validate the CNS state
+	err = validator.Validate(ctx)
+	require.NoError(t, err)
+
+	// Scale it down
+	err = kubernetes.MustScaleDeployment(ctx, deploymentsClient, deployment, clientset, namespace, podLabelSelector, testConfig.ScaleDownReplicas, skipWait)
+	require.NoError(t, err)
+
+	// restart linux CNS (linux, windows)
+	err = kubernetes.RestartCNSDaemonset(ctx, clientset)
+	require.NoError(t, err)
+
+	// wait for pods to settle before checking cns state (otherwise, race between getting pods in terminating state, and getting CNS state file)
+	err = kubernetes.WaitForPodDeployment(ctx, clientset, namespace, deployment.Name, podLabelSelector, testConfig.ScaleDownReplicas)
+	require.NoError(t, err)
+
+	// Validate the CNS state
+	err = validator.Validate(ctx)
+	require.NoError(t, err)
+
+	if testConfig.Cleanup {
+		err = kubernetes.MustDeleteDeployment(ctx, deploymentsClient, deployment)
+		require.NoError(t, err, "error deleteing load deployment")
+		err = kubernetes.WaitForPodsDelete(ctx, clientset, namespace, podLabelSelector)
+		require.NoError(t, err, "error waiting for pods to delete")
+	}
+}
+
+func TestV4OverlayProperties(t *testing.T) {
+	if !testConfig.ValidateV4Overlay {
+		return
+	}
+	clientset, err := kubernetes.MustGetClientset()
+	require.NoError(t, err)
+
+	config := kubernetes.MustGetRestConfig(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
+	defer cancel()
+
+	validator, err := validate.CreateValidator(ctx, clientset, config, namespace, testConfig.CNIType, testConfig.RestartCase, testConfig.OSType)
+	require.NoError(t, err)
+
+	// validate IPv4 overlay scenarios
+	t.Log("Validating v4Overlay node labels")
+	err = validator.ValidateV4OverlayControlPlane(ctx)
+	require.NoError(t, err)
 }
 
 func TestDualStackProperties(t *testing.T) {
--- a/test/integration/load/load_test.go
+++ b/test/integration/load/load_test.go
@@ -183,7 +183,9 @@ func TestScaleDeployment(t *testing.T) {
 	}
 }
 
-func TestValidCNSStateDuringScaleAndCNSRestart(t *testing.T) {
+// TestValidCNSStateDuringScaleAndCNSRestartToTriggerDropgzInstall
+// tests that dropgz install during a pod scaling event, does not crash cns
+func TestValidCNSStateDuringScaleAndCNSRestartToTriggerDropgzInstall(t *testing.T) {
 	clientset, err := kubernetes.MustGetClientset()
 	require.NoError(t, err)
 
--- a/test/internal/kubernetes/utils.go
+++ b/test/internal/kubernetes/utils.go
@@ -233,6 +233,24 @@ func WaitForPodsRunning(ctx context.Context, clientset *kubernetes.Clientset, na
 	return errors.Wrap(retrier.Do(ctx, checkPodIPsFn), "failed to check if pods were running")
 }
 
+func WaitForPodsDelete(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelselector string) error {
+	podsClient := clientset.CoreV1().Pods(namespace)
+
+	checkPodsDeleted := func() error {
+		podList, err := podsClient.List(ctx, metav1.ListOptions{LabelSelector: labelselector})
+		if err != nil {
+			return errors.Wrapf(err, "could not list pods with label selector %s", labelselector)
+		}
+		if len(podList.Items) != 0 {
+			return errors.Errorf("pods still present: %+v", podList.Items)
+		}
+		return nil
+	}
+
+	retrier := retry.Retrier{Attempts: RetryAttempts, Delay: RetryDelay}
+	return errors.Wrap(retrier.Do(ctx, checkPodsDeleted), "failed to wait for pods to delete")
+}
+
 func WaitForPodDeployment(ctx context.Context, clientset *kubernetes.Clientset, namespace, deploymentName, podLabelSelector string, replicas int) error {
 	podsClient := clientset.CoreV1().Pods(namespace)
 	deploymentsClient := clientset.AppsV1().Deployments(namespace)
@@ -430,4 +448,20 @@ func HasWindowsNodes(ctx context.Context, clientset *kubernetes.Clientset) (bool
 		}
 	}
 	return false, nil
+}
+
+func MustRestartDaemonset(ctx context.Context, clientset *kubernetes.Clientset, namespace, daemonsetName string) error {
+	ds, err := clientset.AppsV1().DaemonSets(namespace).Get(ctx, daemonsetName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "failed to get daemonset %s", daemonsetName)
+	}
+
+	if ds.Spec.Template.ObjectMeta.Annotations == nil {
+		ds.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
+
+	ds.Spec.Template.ObjectMeta.Annotations["kubectl.kubernetes.io/restartedAt"] = time.Now().Format(time.RFC3339)
+
+	_, err = clientset.AppsV1().DaemonSets(namespace).Update(ctx, ds, metav1.UpdateOptions{})
+	return errors.Wrapf(err, "failed to update ds %s", daemonsetName)
 }
--- a/test/internal/kubernetes/utils.go
+++ b/test/internal/kubernetes/utils.go
@@ -242,7 +242,7 @@ func WaitForPodsDelete(ctx context.Context, clientset *kubernetes.Clientset, nam
 			return errors.Wrapf(err, "could not list pods with label selector %s", labelselector)
 		}
 		if len(podList.Items) != 0 {
-			return errors.Errorf("pods still present: %+v", podList.Items)
+			return errors.Errorf("%d pods still present", len(podList.Items))
 		}
 		return nil
 	}
--- a/test/validate/linux_validate.go
+++ b/test/validate/linux_validate.go
@@ -23,12 +23,6 @@ var (
 	cnsLocalCacheCmd       = []string{"curl", "localhost:10090/debug/ipaddresses", "-d", "{\"IPConfigStateFilter\":[\"Assigned\"]}"}
 )
 
-// dualstack overlay Linux and windows nodes must have these labels
-var dualstackOverlayNodeLabels = map[string]string{
-	"kubernetes.azure.com/podnetwork-type":   "overlay",
-	"kubernetes.azure.com/podv6network-type": "overlay",
-}
-
 type stateFileIpsFunc func([]byte) (map[string]string, error)
 
 var linuxChecksMap = map[string][]check{
--- a/test/validate/validate.go
+++ b/test/validate/validate.go
@@ -21,9 +21,22 @@ var nodeSelectorMap = map[string]string{
 	"linux":   "kubernetes.io/os=linux",
 }
 
+// IPv4 overlay Linux and windows nodes must have this label
+var v4OverlayNodeLabels = map[string]string{
+	"kubernetes.azure.com/podnetwork-type": "overlay",
+}
+
+// dualstack overlay Linux and windows nodes must have these labels
+var dualstackOverlayNodeLabels = map[string]string{
+	"kubernetes.azure.com/podnetwork-type":   "overlay",
+	"kubernetes.azure.com/podv6network-type": "overlay",
+}
+
 const (
-	privilegedLabelSelector = "app=privileged-daemonset"
-	privilegedNamespace     = "kube-system"
+	privilegedLabelSelector  = "app=privileged-daemonset"
+	privilegedNamespace      = "kube-system"
+	IPv4ExpectedIPCount      = 1
+	DualstackExpectedIPCount = 2
 )
 
 type Validator struct {
@@ -154,20 +167,18 @@ func (v *Validator) validateIPs(ctx context.Context, stateFileIps stateFileIpsFu
 		if err != nil {
 			return errors.Wrapf(err, "failed to exec into privileged pod - %s", podName)
 		}
-		filePodIps, err := stateFileIps(result)
+		cnsPodIPs, err := stateFileIps(result)
 		if err != nil {
 			return errors.Wrapf(err, "failed to get pod ips from state file")
 		}
-		if len(filePodIps) == 0 && v.restartCase {
+		if len(cnsPodIPs) == 0 && v.restartCase {
 			log.Printf("No pods found on node %s", nodes.Items[index].Name)
 			continue
 		}
 		// get the pod ips
-		podIps := getPodIPsWithoutNodeIP(ctx, v.clientset, nodes.Items[index])
+		observedPodIPs := getPodIPsWithoutNodeIP(ctx, v.clientset, nodes.Items[index])
 
-		check := compareIPs(filePodIps, podIps)
-
-		if !check {
+		if err := compareIPs(cnsPodIPs, observedPodIPs); err != nil {
 			return errors.Wrapf(errors.New("State file validation failed"), "for %s on node %s", checkType, nodes.Items[index].Name)
 		}
 	}
@@ -175,12 +186,8 @@ func (v *Validator) validateIPs(ctx context.Context, stateFileIps stateFileIpsFu
 	return nil
 }
 
-func (v *Validator) validateDualStackNodeProperties(ctx context.Context) error {
-	log.Print("Validating Dualstack Overlay Node properties")
-	nodes, err := acnk8s.GetNodeListByLabelSelector(ctx, v.clientset, nodeSelectorMap[v.os])
-	if err != nil {
-		return errors.Wrapf(err, "failed to get node list")
-	}
+func validateNodeProperties(nodes *corev1.NodeList, labels map[string]string, expectedIPCount int) error {
+	log.Print("Validating Node properties")
 
 	for index := range nodes.Items {
 		nodeName := nodes.Items[index].ObjectMeta.Name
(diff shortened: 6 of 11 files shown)


test/integration/manifests/cns/daemonset-windows.yaml
```diff
@@ -0,0 +1,108 @@
...
+      tolerations:
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - operator: "Exists"
+          effect: NoExecute
+        - operator: "Exists"
+          effect: NoSchedule
+      containers:
+        - name: cns-container
+          image: acnpublic.azurecr.io/azure-cns:windows-amd64-ltsc2022-$WINDOWS_CNS_VERSION
```
rbtr, line 42: it should not be necessary to use the windows-specific tag. use the same multiplatform manifest tag that the linux tests use

test/integration/load/load_test.go
```diff
@@ -26,6 +26,7 @@ var (
 	replicas          = flag.Int("replicas", 1, "Number of replicas to scale up/down to")
 	validateStateFile = flag.Bool("validate-statefile", false, "Validate the state file")
 	validateDualStack = flag.Bool("validate-dualstack", false, "Validate the dualstack overlay")
+	validateV4Overlay = flag.Bool("validate-v4Overlay", false, "Validate the IPv4 overlay")
```
ecigar13, line 29: Looks like other flags are lower-case. Maybe we should use all lowercase.

test/integration/manifests/cnsconfig/overlaywindowsconfigmap.yaml
```diff
@@ -0,0 +1,33 @@
...
+      "EnableSubnetScarcity": false,
+      "ChannelMode": "CRD",
+      "InitializeFromCNI": true,
+      "ManageEndpointState": false,
+      "ProgramSNATIPTables" : false,
+      "MetricsBindAddress": ":10092",
+      "EnableCNIConflistGeneration": false,
+      "CNIConflistFilepath": "C:\\k\\azurecni\\netconf\\10-azure.conflist",
+      "CNIConflistScenario": "v4overlay"
+    }
```
pjohnst5, line 33: Can you add a newline at the end of this file? And others without it

pjohnst5:
Main question I have is, does it work in the pipeline ok?
Can you link a successful run with your changes?

test/validate/windows_validate.go
```diff
@@ -89,6 +98,17 @@ func hnsStateFileIps(result []byte) (map[string]string, error) {
 	return hnsPodIps, nil
 }
 
+// return windows HNS network state
+func hnsNetworkState(result []byte) ([]HNSNetwork, error) {
+	var hnsNetworkResult []HNSNetwork
+	err := json.Unmarshal(result, &hnsNetworkResult)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal hns network list")
```
ramiro-gamarra, line 106:
```suggestion
		return nil, errors.Wrap(err, "failed to unmarshal hns network list")
```
since you are not using any formatting directives. or otherwise
```suggestion
		return nil, errors.Wrapf(err, "failed to unmarshal hns network list. hns bytes: %s", string(result))
```

test/validate/validate.go
```diff
@@ -224,3 +224,94 @@ func (v *Validator) ValidateDualStackControlPlane(ctx context.Context) error {
...
+	}
+
+	return nil
+}
+
+func (v *Validator) validateHNSNetworkState(ctx context.Context) error {
+	// check HNS network state only on windows nodes
+	nodes, err := k8sutils.GetNodeListByLabelSelector(ctx, v.clientset, nodeSelectorMap["windows"])
+	if err != nil {
+		return errors.Wrapf(err, "failed to get node list")
```
ramiro-gamarra, line 275: Use `errors.Wrap` (no `f` suffix) if you are not doing any formatting in the message, or, preferably, add more context to the messages. Applies to all usages of `errors.Wrapf`.

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
```diff
@@ -161,3 +161,87 @@ steps:
...
+        echo "Upload CNI"
+        dropgz_version=$(make cni-dropgz-test-version)
+        export DROP_GZ_URL=$( make cni-dropgz-test-image-name-and-tag OS='windows' ARCH='amd64' OS_VERSION='ltsc2022' CNI_DROPGZ_VERSION=${dropgz_version})
+        envsubst < ./test/integration/manifests/cni/cni-installer-v1-windows.yaml | kubectl apply -f -
+    name: "Deploy_CNS_CNI"
+    displayName: "Deploy CNS and upload CNI on v4Overlay cluster"
+
+  - script: |
+      kubectl taint nodes aksnpwin000000  node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
+      kubectl taint nodes aksnpwin000001  node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
```
ramiro-gamarra, line 204: Is this what you mentioned was needed because of cloud node manager issues?

pjohnst5: Yeah

pjohnst5 (dismissed):
LGTM, given it passes again
I'm fine if we leave these two lines are remove them after that Sev2 incident in AKS:
```
kubectl taint nodes aksnpwin000000  node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
      kubectl taint nodes aksnpwin000001  node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
```

pjohnst5 (approved):
LGTM, given it passes again
I'm fine if we leave these two lines are remove them after that Sev2 incident in AKS:
```
kubectl taint nodes aksnpwin000000  node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
kubectl taint nodes aksnpwin000001  node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
```

test/integration/manifests/cns/daemonset-windows.yaml
```diff
@@ -0,0 +1,149 @@
...
+      hostNetwork: true
+      volumes:
+        - name: log
+          hostPath:
+            path: /k/azurecns
+            type: DirectoryOrCreate
+        - name: cns-config
+          configMap:
+            name: cns-win-config
+            defaultMode: 420
```
jaer-tsun, line 109: what does 420 mean?

test/validate/validate.go
```diff
@@ -224,3 +224,94 @@ func (v *Validator) ValidateDualStackControlPlane(ctx context.Context) error {
...
+	}
+
+	if err := v.validateHNSNetworkState(ctx); err != nil {
+		return errors.Wrap(err, "failed to validate IPv4 overlay HNS network state")
+	}
+
+	return nil
+}
+
+func (v *Validator) validateHNSNetworkState(ctx context.Context) error {
```
jaer-tsun, line 271: should this be on windows_validate?

paulyufan2: This function is to validate windows HNS network state. HNS is only for Windows.

test/integration/manifests/cns/daemonset-windows.yaml
paulyufan2: The permission of this file

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
```diff
@@ -161,3 +161,87 @@ steps:
...
+      addSpnToEnvironment: true
+      inlineScript: |
+        set -e
+        echo "Deploying windows CNS on v4 overlay cluster"
+        export WINDOWS_CNS_VERSION=$(make cns-version)
+        envsubst '${WINDOWS_CNS_VERSION}' < test/integration/manifests/cns/daemonset-windows.yaml | kubectl apply -f -
+        echo "wait for CNS pods deployment ready"
+        kubectl rollout status ds azure-cns-win -n kube-system --timeout 5m
+        kubectl get po -owide -A
+        kubectl describe ds azure-cns-win -n kube-system | grep Image
```
vipul-21, line 194: Why do we need this ?

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
```diff
@@ -161,3 +161,87 @@ steps:
...
+        echo "Upload CNI"
+        dropgz_version=$(make cni-dropgz-test-version)
+        export DROP_GZ_URL=$( make cni-dropgz-test-image-name-and-tag OS='windows' ARCH='amd64' OS_VERSION='ltsc2022' CNI_DROPGZ_VERSION=${dropgz_version})
```
vipul-21, line 197:
```suggestion
        export DROP_GZ_URL=$(make cni-dropgz-test-image-name-and-tag OS='windows' ARCH='amd64' OS_VERSION='ltsc2022' CNI_DROPGZ_VERSION=${dropgz_version})
```

test/validate/validate.go
```diff
@@ -224,3 +224,94 @@ func (v *Validator) ValidateDualStackControlPlane(ctx context.Context) error {
 
 	return nil
 }
+
+func (v *Validator) validateIPv4NodeProperties(ctx context.Context) error {
```
vipul-21, line 228: This function is similar to `validateDualStackNodeProperties`, i would say we create function in validator i.e `validateNodeProperties` and reuse the code if we can

vipul-21 (changes requested):

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
pjohnst5: I see most other places have it with a space:
image

test/validate/validate.go
pjohnst5: Good point, @paulyufan2 , you could make a function like this:
```
func validateNodeProperties(labels map[string]string, ipCount int) {
nodes, err := k8sutils.GetNodeListByLabelSelector(ctx, v.clientset, nodeSelectorMap[v.os])
	if err != nil {
		return errors.Wrap(err, "failed to get node list")
	}

	for index := range nodes.Items {
		nodeName := nodes.Items[index].ObjectMeta.Name
		// check nodes status;
		// nodes status should be ready after cluster is created
		nodeConditions := nodes.Items[index].Status.Conditions
		if nodeConditions[len(nodeConditions)-1].Type != corev1.NodeReady {
			return errors.Wrapf(err, "node %s status is not ready", nodeName)
		}

		// get node labels
		nodeLabels := nodes.Items[index].ObjectMeta.GetLabels()
		for key := range nodeLabels {
			if label, ok := labels[key]; ok {
				log.Printf("label %s is correctly shown on the node %+v", key, nodeName)
				if label != overlayClusterLabelName {
					return errors.Wrapf(err, "node %s overlay label name is wrong; expected label:%s but actual label:%s", nodeName, overlayClusterLabelName, label)
				}
			}
		}

		// check if node has two internal IPs(one is IPv4 and another is IPv6)
		internalIPCount := 0
		for _, address := range nodes.Items[index].Status.Addresses {
			if address.Type == "InternalIP" {
				internalIPCount++
			}
		}
		if internalIPCount != ipCount {
			return errors.Wrap(err, "node does not expected number of IPs")
		}
}
```

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
paulyufan2: it will print which CNS image we are using:
paulyu@paul-microsoft:/mnt/c/Users/paulyu$ kubectl describe ds azure-cns-win -n kube-system | grep Image
    Image:       mcr.microsoft.com/containernetworking/azure-cns:v1.5.11

test/validate/validate.go
```diff
@@ -224,3 +224,94 @@ func (v *Validator) ValidateDualStackControlPlane(ctx context.Context) error {
...
+			return errors.Wrap(err, "failed to exec into privileged pod")
+		}
+
+		hnsNetwork, err := hnsNetworkState(result)
+		if err != nil {
+			return errors.Wrap(err, "failed to unmarshal hns network list")
+		}
+
+		if len(hnsNetwork) == 0 {
+			return errors.Wrap(err, "windows node does not have HNS network")
```
ramiro-gamarra, line 297: `errors.Wrap` returns nil if the wrapped error variable is nil. This is incorrect since in this case you will be returning nil instead of your intended error. Please correct **all** instances of `errors.Wrap` in your PR where a nil error variable is being wrapped.

test/validate/validate.go
pjohnst5: And then call it like this:
```
func (v *Validator) validateDualStackNodeProperties(ctx context.Context) error {
    log.Print("Validating Dualstack Overlay Node properties")
    if err := validateNodeProperties(ctx, dualstackOverlayNodeLabels, 2); err != nil  {
        return err
    }
return nil
```
func (v *Validator) validateIPv4NodeProperties(ctx context.Context) error {
    log.Print("Validating IPv6 Overlay Cluster Node properties")
    if err := validateNodeProperties(ctx, v4OverlayNodeLabels, 1); err != nil  {
        return err
    }
return nil
```

```

ramiro-gamarra (changes requested):

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
```diff
@@ -161,3 +161,87 @@ steps:
...
+        envsubst < ./test/integration/manifests/cni/cni-installer-v1-windows.yaml | kubectl apply -f -
+    name: "Deploy_CNS_CNI"
+    displayName: "Deploy CNS and upload CNI on v4Overlay cluster"
+
+  - script: |
+      kubectl taint nodes aksnpwin000000  node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
+      kubectl taint nodes aksnpwin000001  node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
+      cd test/integration/load
+      echo "IPv4 Overlay Windows control plane CNS validation test"
+      sudo go test -timeout 30m -tags load -run ^TestLoad$ -os=windows
```
pjohnst5, line 207: I think we should just run all these load tests, and make load test suite have a way of installing CNS daemonset, in go-lang code, similar to how `make test-integration` does it

We could even have 1 central go-lang code, that installs CNS and Dropgz, so that both test suites (`test/ingetration` and `test/integration/load` can use the same code)

pjohnst5 (changes requested):
We should find out a better way of isntalling cns daemosnet for windows, similar to how `make test-integration` does it

test/integration/manifests/cns/daemonset-windows.yaml
```diff
@@ -0,0 +1,149 @@
+
```
jpayne3506, line 1: nit: the linux daemonset within this directory does not have a configmap section, hence the cnsconfig directory. Is there any reason not to keep that pattern?

test/validate/linux_validate.go
```diff
@@ -22,6 +22,11 @@ var (
 	cnsLocalCacheCmd      = []string{"curl", "localhost:10090/debug/ipaddresses", "-d", "{\"IPConfigStateFilter\":[\"Assigned\"]}"}
 )
 
+// IPv4 overlay Linux and windows nodes must have this label
+var v4OverlayNodeLabels = map[string]string{
```
jpayne3506, line 26: Does the node label differ if it is v4 or v6? What other variants are there?

jpayne3506 (changes requested):

test/validate/validate.go
```diff
@@ -224,3 +224,94 @@ func (v *Validator) ValidateDualStackControlPlane(ctx context.Context) error {
...
+		}
+
+		if len(hnsNetwork) == 1 {
+			return errors.Wrap(err, "HNS default ext network or azure network does not exist")
+		}
```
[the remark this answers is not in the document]
paulyufan2: if len == 0, windows does not have HNS network at all;
if len == 1, windows is missing default ext network or azure network

test/validate/linux_validate.go
paulyufan2: you can compare with dualstackOverlayNodeLabels map

test/validate/validate.go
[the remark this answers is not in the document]
jpayne3506: Shouldn't `len == 0` be impossible to reach as the `execCmd` would fail if it does not have HNS network and its respective file? What I am suggesting is that `len == 0` needs a different error or it has the same root cause as `len == 1`, the network exists, but is missing/not listed.

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
```diff
@@ -161,3 +161,87 @@ steps:
...
     condition: always()
+
+  - task: AzureCLI@2
+    inputs:
+      azureSubscription: $(AZURE_TEST_AGENT_SERVICE_CONNECTION)
+      scriptLocation: "inlineScript"
+      scriptType: "bash"
+      addSpnToEnvironment: true
+      inlineScript: |
+        make -C ./hack/aks windows-nodepool-up AZCLI=az SUB=$(SUB_AZURE_NETWORK_AGENT_BUILD_VALIDATIONS) CLUSTER=${{ parameters.clusterName }}
```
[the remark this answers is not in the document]
jpayne3506: As long as the user does not restart the test before it is deleted this is true.

test/validate/validate.go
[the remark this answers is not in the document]
paulyufan2: yes, if no HNS network, execCmd will fail, so here we only check if len == 1 or more.

test/validate/validate.go
```diff
@@ -224,3 +224,94 @@ func (v *Validator) ValidateDualStackControlPlane(ctx context.Context) error {
...
+
+		if v.cni == "overlay" {
+			for _, network := range hnsNetwork {
+				if network.State != 1 {
+					return errors.Wrap(err, "windows HNS network state is not correct")
+				}
+				if network.ManagementIP == "" {
+					return errors.Wrap(err, "windows HNS network is missing ipv4 management IP")
+				}
```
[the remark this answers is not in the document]
paulyufan2: because if v.cni == dualstackoverlay, we need to check other properties of hnsNetwork

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
paulyufan2: fixed all places

test/validate/windows_validate.go
```diff
@@ -130,3 +156,80 @@ func azureVnetIpamIps(result []byte) (map[string]string, error) {
...
+		if nodeConditions[len(nodeConditions)-1].Type != corev1.NodeReady {
+			return errors.Errorf("node %s status is not ready", nodeName)
+		}
+
+		// get node labels
+		nodeLabels := nodes.Items[index].ObjectMeta.GetLabels()
+		for key := range nodeLabels {
+			if label, ok := labels[key]; ok {
+				log.Printf("label %s is correctly shown on the node %+v", key, nodeName)
+				if label != overlayClusterLabelName {
```
vipul-21, line 178: This is accessing the labels in `linux_validate`.
I think `validateNodeProperties` is OS agnostic and should stay in `validate.go` and based on the cni type it should validate the labels.

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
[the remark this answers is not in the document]
paulyufan2: set -e

test/internal/k8sutils/utils.go
```diff
@@ -54,12 +54,12 @@ func MustGetClientset() (*kubernetes.Clientset, error) {
 	return clientset, nil
 }
 
-func MustGetRestConfig(t *testing.T) *rest.Config {
+func MustGetRestConfig() (*rest.Config, error) {
```
ramiro-gamarra, line 57: The `Must` prefix is a common idiom in Go to denote a panic will occur instead of returning an error. Can you change this to `GetRestConfig`?

test/internal/k8sutils/utils_create.go
```diff
@@ -41,6 +53,34 @@ func MustCreateDaemonset(ctx context.Context, daemonsets typedappsv1.DaemonSetIn
...
 }
 
+func MustRestartDaemonset(ctx context.Context, daemonsets typedappsv1.DaemonSetInterface, ds appsv1.DaemonSet) error {
+	data := fmt.Sprintf(`{"spec": {"template": {"metadata": {"annotations": {"kubectl.kubernetes.io/restartedAt": "%q"}}}}}`, time.Now().Format("20060102150405"))
+	_, err := daemonsets.Patch(ctx, ds.Name, types.StrategicMergePatchType, []byte(data), metav1.PatchOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to patch daemonset")
+	}
+
+	checkReadyCtx, cancel := context.WithTimeout(context.Background(), contextTimeout)
```
ramiro-gamarra, line 63: pass ctx here

test/internal/k8sutils/utils.go
pjohnst5: Yeah I can change it, it's not used anywhere but test code

pjohnst5: I'll change it back

pjohnst5: Well actually, it's the only method in here, that receives `t *testing.T` as an argument, which I thought was odd
So I'm inclined to rename it `GetRestConfig`, but still leave out the `t *testing.T` argument

test/integration/utils.go
```diff
@@ -0,0 +1,394 @@
...
+	} else {
+		log.Printf("Env %v not set to true, deploying cni.Dockerfile", envTestDropgz)
+		initImage, _ := k8sutils.ParseImageString(cns.Spec.Template.Spec.InitContainers[0].Image)
+		cns.Spec.Template.Spec.InitContainers[0].Image = k8sutils.GetImageString(initImage, dropgzVersion)
+	}
+	return cns
+}
+
+func loadAzureVnet(ctx context.Context, clientset *kubernetes.Clientset, cns appsv1.DaemonSet) (appsv1.DaemonSet, error) {
+	if installBool1 := os.Getenv(envInstallAzureVnet); installBool1 != "" {
```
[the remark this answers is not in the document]
pjohnst5: Yeah, just copied it from before, I'll rename them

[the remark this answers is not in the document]
pjohnst5: They used to be all in one method, which is why they were just "1", "2" ect. 😄

pjohnst5:
Aha! Works to install CNS daemonset for windows, as well as dropgz for windows as an initi container:
image

test/integration/utils.go
```diff
@@ -0,0 +1,499 @@
...
+		if testDropgzScenario, err := strconv.ParseBool(installFlag); err == nil && testDropgzScenario {
+			log.Printf("Env %v set to true, deploy cniTest.Dockerfile", envTestDropgz)
+			initImage, _ := k8sutils.ParseImageString("acnpublic.azurecr.io/cni-dropgz-test:latest")
+			cns.Spec.Template.Spec.InitContainers[0].Image = k8sutils.GetImageString(initImage, dropgzVersion)
+		}
+	} else {
+		log.Printf("Env %v not set to true, deploying cni.Dockerfile", envTestDropgz)
+		initImage, _ := k8sutils.ParseImageString(cns.Spec.Template.Spec.InitContainers[0].Image)
+		cns.Spec.Template.Spec.InitContainers[0].Image = k8sutils.GetImageString(initImage, dropgzVersion)
+	}
```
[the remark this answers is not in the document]
jpayne3506:
```suggestion
	installFlag := os.Getenv(envTestDropgz)
	if testDropgzScenario, err := strconv.ParseBool(installFlag); err == nil && testDropgzScenario {
		log.Printf("Env %v set to true, deploy cniTest.Dockerfile", envTestDropgz)
		initImage, _ := k8sutils.ParseImageString("acnpublic.azurecr.io/cni-dropgz-test:latest")
		cns.Spec.Template.Spec.InitContainers[0].Image = k8sutils.GetImageString(initImage, dropgzVersion)
	} else {
		log.Printf("Env %v not set to true, deploying cni.Dockerfile", envTestDropgz)
		initImage, _ := k8sutils.ParseImageString(cns.Spec.Template.Spec.InitContainers[0].Image)
		cns.Spec.Template.Spec.InitContainers[0].Image = k8sutils.GetImageString(initImage, dropgzVersion)
	}
```

Without changing the fields that are brought in to the function I think this captures the above sentiment and makes it more predictable.

Makefile
```diff
@@ -727,6 +727,11 @@ test-integration: ## run all integration tests.
 		CNS_VERSION=$(CNS_VERSION) \
 		go test -mod=readonly -buildvcs=false -timeout 1h -coverpkg=./... -race -covermode atomic -coverprofile=coverage.out -tags=integration ./test/integration...
 
+test-load: ## run all load tests
+	CNI_DROPGZ_VERSION=$(CNI_DROPGZ_VERSION) \
+		CNS_VERSION=$(CNS_VERSION) \
+			go test -timeout 30m -race -tags=load ./test/integration/load...
```
rbtr, line 733:
```suggestion
test-load: ## run all load tests
	CNI_DROPGZ_VERSION=$(CNI_DROPGZ_VERSION) \
		CNS_VERSION=$(CNS_VERSION) \
		go test -timeout 30m -race -tags=load ./test/integration/load...
```

test/integration/utils.go
```diff
@@ -0,0 +1,499 @@
...
+	// setup RBAC, Role, RoleBinding
+	if err := k8sutils.MustSetUpRBAC(ctx, clientset, cnsRolePath, cnsRoleBindingPath); err != nil {
+		return appsv1.DaemonSet{}, errors.Wrap(err, "failed to setup RBAC, Role and RoleBinding")
+	}
+
+	if err := k8sutils.MustCreateDaemonset(ctx, cnsDaemonsetClient, cns); err != nil {
+		return appsv1.DaemonSet{}, errors.Wrap(err, "failed to create daemonset")
+	}
+
+	if err := k8sutils.WaitForPodsRunning(ctx, clientset, cns.Namespace, labelSelector); err != nil {
```
jpayne3506, line 366: `k8sutils.WaitForPodDaemonset` is more resilient and has the same functionality as `k8sutils.WaitForPodsRunning`. If the daemonset already exists then it won't continue until all pods have been restarted otherwise it will proceed as intended.

test/internal/k8sutils/utils_create.go
```diff
@@ -41,8 +53,36 @@ func MustCreateDaemonset(ctx context.Context, daemonsets typedappsv1.DaemonSetIn
...
+		daemonset, err := daemonsets.Get(ctx, ds.Name, metav1.GetOptions{})
+		if err != nil {
+			return errors.Wrap(err, "failed to get daemonset")
+		}
+		numReady := daemonset.Status.NumberReady
+		desiredNumReady := daemonset.Status.DesiredNumberScheduled
+		if numReady != desiredNumReady {
+			return errors.Errorf("daemonset %s in namespace %s has %v pods ready but desired %v", ds.Name, ds.Namespace, numReady, desiredNumReady)
+		}
+		return nil
```
jpayne3506, line 75: `WaitForPodDaemonset` has this functionality with some additional flavor within utils.go

jpayne3506 (changes requested):

test/validate/validate.go
```diff
@@ -175,52 +207,72 @@ func (v *Validator) validateIPs(ctx context.Context, stateFileIps stateFileIpsFu
...
+	}
+
+	if err := validateNodeProperties(nodes, v4OverlayNodeLabels, ipv4OverlayInternalIPs); err != nil {
+		return errors.Wrap(err, "failed to validate IPv4 overlay node properties")
+	}
+
+	if v.os == "windows" {
+		if err := validateHNSNetworkState(ctx, nodes, v.clientset, v.config); err != nil {
+			return errors.Wrap(err, "failed to validate IPv4 overlay HNS network state")
+		}
```
jpayne3506, line 274: Is this an overlay exclusive HNS check? Maybe we should make a function that validates all HNS related items.

test/validate/validate.go
```diff
@@ -175,52 +207,72 @@ func (v *Validator) validateIPs(ctx context.Context, stateFileIps stateFileIpsFu
...
 				log.Printf("label %s is correctly shown on the node %+v", key, nodeName)
 				if label != overlayClusterLabelName {
-					return errors.Wrapf(err, "node %s overlay label name is wrong; expected label:%s but actual label:%s", nodeName, overlayClusterLabelName, label)
+					return errors.Errorf("node %s overlay label name is wrong; expected label:%s but actual label:%s", nodeName, overlayClusterLabelName, label)
 				}
 			}
 		}
 
-		// check if node has two internal IPs(one is IPv4 and another is IPv6)
+		// check if node has corret number two internal IPs
```
jpayne3506, line 233: `corret number two ` -> ` correct number of `

test/integration/utils.go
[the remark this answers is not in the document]
jpayne3506: Changes captured within https://github.com/Azure/azure-container-networking/pull/2227.

test/integration/load/load_test.go
```diff
@@ -124,25 +135,20 @@ func TestLoad(t *testing.T) {
...
 		t.Run("Validate dualstack overlay", TestDualStackProperties)
 	}
+
+	if *validateV4Overlay {
+		t.Run("Validate IPv4 overlay", TestIPv4OverlayProperties)
+	}
+
+	// Delete pods
+	err = k8sutils.MustDeleteDeployment(ctx, deploymentsClient, deployment)
+	require.NoError(t, err, "error deleteing load deployment")
```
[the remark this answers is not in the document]
pjohnst5: Fixed in #2229

test/integration/datapath/datapath_windows_test.go
```diff
@@ -48,12 +53,35 @@ Timeout context is controled by the -timeout flag.
...
 	clientset, err := kubernetes.MustGetClientset()
 	if err != nil {
 		t.Fatal(err)
 	}
 
+	if *initBYOCNI {
+		// create ipmasq configmaps
+		err := kubernetes.MustSetupConfigMap(ctx, clientset, IPMasqAgentCustomConfigMap)
+		require.NoError(t, err, "failed to setup ip masq agent custom configmap")
+		err = kubernetes.MustSetupConfigMap(ctx, clientset, IPMasqAgentReconcileConfigMap)
```
paulyufan2, line 69: If we deploy ipmasqagent here, let's remove the ip-masq-agent installation from pipeline:
https://github.com/Azure/azure-container-networking/blob/master/.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml

Line 72:
kubectl apply -f test/integration/manifests/ip-masq-agent/ip-masq-agent.yaml --validate=false

test/integration/datapath/datapath_windows_test.go
```diff
@@ -10,21 +10,26 @@ import (
...
+	podLabelKey                   = "app"
+	podCount                      = 2
+	nodepoolKey                   = "agentpool"
 )
 
 var (
 	podPrefix        = flag.String("podName", "datapod", "Prefix for test pods")
 	podNamespace     = flag.String("namespace", "windows-datapath-test", "Namespace for test pods")
 	nodepoolSelector = flag.String("nodepoolSelector", "npwin", "Provides nodepool as a windows Node-Selector for pods")
+	initBYOCNI       = flag.Bool("initbyocni", false, "installs ipmas agent, configmap, and restarts kubeproxy")
```
paulyufan2, line 32: nit: ip-masq-agent

test/integration/load/load_test.go
```diff
@@ -171,9 +178,94 @@ func TestScaleDeployment(t *testing.T) {
 	if testConfig.Cleanup {
 		err = kubernetes.MustDeleteDeployment(ctx, deploymentsClient, deployment)
 		require.NoError(t, err, "error deleteing load deployment")
+		err = kubernetes.WaitForPodsDelete(ctx, clientset, namespace, podLabelSelector)
+		require.NoError(t, err, "error waiting for pods to delete")
 	}
 }
 
+func TestValidCNSStateDuringScaleAndCNSRestart(t *testing.T) {
```
jpayne3506, line 186: This looks like the node restart scenario. We need to ensure that we have pods on the cluster, validate, restart CNS, validate. No need to scale down or use `skipWait` on scale up.

Or is this adding additional coverage?

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
```diff
@@ -172,3 +172,41 @@ steps:
...
+
+  - script: |
+      nodeList=`kubectl get node -owide | grep Windows | awk '{print $1}'`
+      for node in $nodeList; do
+          taint=`kubectl describe node $node | grep Taints | awk '{print $2}'`
+          if [ $taint == "node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule" ]; then
+              kubectl taint nodes $node node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
+          fi
+      done
+      CNS=$(make cns-version) DROPGZ=$(make cni-dropgz-version)
```
vipul-21, line 200: We removed the submodules pipeline so this will be change when we are testing CNI and when we are testing CNS components. example: https://github.com/Azure/azure-container-networking/blob/master/.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml#L83-L92

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
```diff
@@ -172,3 +172,41 @@ steps:
...
+      sudo -E env "PATH=$PATH" make test-load SCALE_UP=32 OS_TYPE=windows CNI_TYPE=cniv2 VALIDATE_STATEFILE=true INSTALL_CNS=true INSTALL_AZURE_CNI_OVERLAY=true VALIDATE_V4OVERLAY=true CNS_VERSION=${CNS} CNI_DROPGZ_VERSION=${DROPGZ} CLEANUP=true
```
vipul-21, line 201: Is there a reason why we chose `32` ?

test/integration/load/load_test.go
```diff
@@ -171,9 +178,94 @@ func TestScaleDeployment(t *testing.T) {
...
+	err = validator.Validate(ctx)
+	require.NoError(t, err)
+
+	deployment, err := kubernetes.MustParseDeployment(noopDeploymentMap[testConfig.OSType])
+	require.NoError(t, err)
+	deploymentsClient := clientset.AppsV1().Deployments(namespace)
+
+	// Create a deployment
+	err = kubernetes.MustCreateDeployment(ctx, deploymentsClient, deployment)
+	require.NoError(t, err)
```
jpayne3506, line 206: Lets add the `if testConfig.Cleanup` here.

test/integration/load/load_test.go
```diff
@@ -171,9 +178,94 @@ func TestScaleDeployment(t *testing.T) {
...
+
+	// Scale it up and "skipWait", so CNS restart can happen immediately after scale call is made (while pods are still creating)
+	skipWait := true
+	err = kubernetes.MustScaleDeployment(ctx, deploymentsClient, deployment, clientset, namespace, podLabelSelector, testConfig.ScaleUpReplicas, skipWait)
+	require.NoError(t, err)
+
+	// restart linux CNS (linux, windows)
+	err = kubernetes.RestartCNSDaemonset(ctx, clientset)
```
jpayne3506, line 214: can add  `WaitForPodDaemonset()` to ensure a race condition does not happen after restart.

test/integration/datapath/datapath_windows_test.go
```diff
@@ -48,12 +53,35 @@ Timeout context is controled by the -timeout flag.
...
+		require.NoError(t, err, "failed to setup ip masq agent reconcile configmap")
+
+		// create ipmasq agent daemonset
+		ds, err := kubernetes.MustParseDaemonSet(IPMasqAgentYamlPath)
+		require.NoError(t, err, "failed to parse ipmasq agent")
+		err = kubernetes.MustCreateDaemonset(ctx, clientset.AppsV1().DaemonSets(ds.Namespace), ds)
+		require.NoError(t, err, "failed to create ipmasq daemonset")
+
+		// restart kubeproxy
+		validator, err := validate.CreateValidator(ctx, clientset, restConfig, *podNamespace, "cniv2", false, "windows")
```
vipul-21, line 79: "cniv2" will be a constant ? Should this be a parameter ?

test/internal/kubernetes/utils.go
```diff
@@ -233,6 +233,24 @@ func WaitForPodsRunning(ctx context.Context, clientset *kubernetes.Clientset, na
...
+func WaitForPodsDelete(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelselector string) error {
+	podsClient := clientset.CoreV1().Pods(namespace)
+
+	checkPodsDeleted := func() error {
+		podList, err := podsClient.List(ctx, metav1.ListOptions{LabelSelector: labelselector})
+		if err != nil {
+			return errors.Wrapf(err, "could not list pods with label selector %s", labelselector)
+		}
+		if len(podList.Items) != 0 {
+			return errors.Errorf("pods still present: %+v", podList.Items)
```
jpayne3506, line 245: How does the output of podList.Items look?

test/internal/kubernetes/utils.go
```diff
@@ -431,3 +449,19 @@ func HasWindowsNodes(ctx context.Context, clientset *kubernetes.Clientset) (bool
 	}
 	return false, nil
 }
+
+func MustRestartDaemonset(ctx context.Context, clientset *kubernetes.Clientset, namespace, daemonsetName string) error {
```
jpayne3506, line 453: Does this not have a happy path? Looks like every return is resulting in an error.

test/validate/validate.go
```diff
@@ -154,77 +167,121 @@ func (v *Validator) validateIPs(ctx context.Context, stateFileIps stateFileIpsFu
 		if err != nil {
 			return errors.Wrapf(err, "failed to exec into privileged pod - %s", podName)
 		}
-		filePodIps, err := stateFileIps(result)
+		cnsPodIPs, err := stateFileIps(result)
```
jpayne3506, line 170: `validateIPs()` is used for multiple state files, so a general var name `filePodIps` was used.

test/validate/windows_validate.go
```diff
@@ -130,3 +156,40 @@ func azureVnetIpamIps(result []byte) (map[string]string, error) {
...
+		}
+
+		hnsNetwork, err := hnsNetworkState(result)
+		if err != nil {
+			return errors.Wrap(err, "failed to unmarshal HNS state file")
+		}
+
+		// check hns properties
+		if len(hnsNetwork) == 1 {
+			return errors.New("HNS default ext network or azure network does not exist")
```
vipul-21, line 182: Would be nice to log the `hnsNetwork` as the error says it is missing one of those info.

test/validate/validate.go
```diff
@@ -154,77 +167,121 @@ func (v *Validator) validateIPs(ctx context.Context, stateFileIps stateFileIpsFu
...
+
+	if err := validateNodeProperties(nodes, dualstackOverlayNodeLabels, DualstackExpectedIPCount); err != nil {
 		return errors.Wrap(err, "failed to validate dualstack overlay node properties")
 	}
 
 	return nil
 }
 
+func (v *Validator) RestartKubeProxyService(ctx context.Context) error {
+	nodes, err := acnk8s.GetNodeList(ctx, v.clientset)
```
vipul-21, line 259: We can filter based on windows here and then remove line 266 ?

test/validate/validate.go
```diff
@@ -154,77 +167,121 @@ func (v *Validator) validateIPs(ctx context.Context, stateFileIps stateFileIpsFu
...
 				log.Printf("label %s is correctly shown on the node %+v", key, nodeName)
 				if label != overlayClusterLabelName {
-					return errors.Wrapf(err, "node %s overlay label name is wrong; expected label:%s but actual label:%s", nodeName, overlayClusterLabelName, label)
+					return errors.Errorf("node %s overlay label name is wrong; expected label:%s but actual label:%s", nodeName, overlayClusterLabelName, label)
 				}
 			}
 		}
 
-		// check if node has two internal IPs(one is IPv4 and another is IPv6)
+		// check if node has corret number of internal IPs
```
jpayne3506, line 212: `corret` -> `correct`

test/validate/validate.go
vipul-21, line 170: Why `cnsPodIps` ? I think this contains all podsIps ?

test/validate/validate.go
```diff
@@ -154,77 +167,121 @@ func (v *Validator) validateIPs(ctx context.Context, stateFileIps stateFileIpsFu
...
 		internalIPCount := 0
 		for _, address := range nodes.Items[index].Status.Addresses {
 			if address.Type == "InternalIP" {
 				internalIPCount++
 			}
 		}
-		if internalIPCount != 2 { //nolint
-			return errors.Wrap(err, "node does not have two internal IPs")
+		if internalIPCount != expectedIPCount {
+			return errors.Errorf("node does not expected number of IPs %d", expectedIPCount)
```
jpayne3506, line 220: Can we return the `internalIPCount` as well?

jpayne3506 (changes requested):

test/integration/load/load_test.go
```diff
@@ -171,9 +178,94 @@ func TestScaleDeployment(t *testing.T) {
...
+	// wait for pods to settle before checking cns state (otherwise, race between getting pods in creating state, and getting CNS state file)
+	err = kubernetes.WaitForPodDeployment(ctx, clientset, namespace, deployment.Name, podLabelSelector, testConfig.ScaleUpReplicas)
+	require.NoError(t, err)
+
+	// Validate the CNS state
+	err = validator.Validate(ctx)
+	require.NoError(t, err)
+
+	// Scale it down
+	err = kubernetes.MustScaleDeployment(ctx, deploymentsClient, deployment, clientset, namespace, podLabelSelector, testConfig.ScaleDownReplicas, skipWait)
```
vipul-21, line 226: is there a reason we scale it down ?

test/integration/datapath/datapath_windows_test.go
```diff
@@ -48,12 +53,35 @@ Timeout context is controled by the -timeout flag.
...
+	restConfig := kubernetes.MustGetRestConfig(t)
+
 	t.Log("Create Clientset")
 	clientset, err := kubernetes.MustGetClientset()
 	if err != nil {
 		t.Fatal(err)
 	}
 
+	if *initBYOCNI {
+		// create ipmasq configmaps
```
vipul-21, line 66: We'll need to remove the yaml deployments from the pipeline yaml i suppose.

test/validate/validate.go
pjohnst5: That's true, I guess what I meant to say here, was "pods that CNS is managing"

test/integration/datapath/datapath_windows_test.go
pjohnst5: Hmm, that's on Linux, and I'm not sure if we should change, that, because in this PR, we're testing Windows
That was already there, if we want to remove it in a different PR, I'm OK with that, but I think for this PR, we should scope this change to Windows tests only

pjohnst5: Seems like there are a lot of places, where `ip-masq-agent` has been installed as a `kubectl apply -f yaml`
I will not install it in code here, and make a different PR after this, to change all instances of that installation into go-code, in a separate PR

test/integration/datapath/datapath_windows_test.go
```diff
@@ -10,21 +10,26 @@ import (
...
 
 const (
-	WindowsDeployYamlPath = "../manifests/datapath/windows-deployment.yaml"
-	podLabelKey           = "app"
-	podCount              = 2
-	nodepoolKey           = "agentpool"
+	WindowsDeployYamlPath         = "../manifests/datapath/windows-deployment.yaml"
+	IPMasqAgentYamlPath           = "../manifests/ip-masq-agent/ip-masq-agent.yaml"
+	IPMasqAgentCustomConfigMap    = "../manifests/ip-masq-agent/config-custom.yaml"
+	IPMasqAgentReconcileConfigMap = "../manifests/ip-masq-agent/config-reconcile.yaml"
```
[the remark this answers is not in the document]
pjohnst5: Good point, yes it is

Seems like there are a lot of places, where ip-masq-agent has been installed as a kubectl apply -f yaml
I will not install it in code here, and make a different PR after this, to change all instances of that installation into go-code, in a separate PR
image

test/integration/load/load_test.go
pjohnst5: This adds additional coverage, by restarting CNS in the middle of a scale down event
There was a bug before in CNS, that on restart of CNS, and scaling up/down, IPs could get "leaked", i.e., their Pods are gone, but the IPs in CNS are still "allocated", so now we're testing that here

test/integration/load/load_test.go
pjohnst5: We want the race condition to happen here actually, we want to test that `dropgz` as an init container to `cns` for windows nodes, does not have any issues replacing the `azure-vnet.exe` binary during a pod scaling event (and it's good, no issues)

test/internal/kubernetes/utils.go
pjohnst5: It does, the last `errors.Wrap()` will return nil, if the err coming back 
`Wrapf returns an error annotating err with a stack trace at the point Wrapf is called, and the format specifier. If err is nil, Wrapf returns nil.`

test/internal/kubernetes/utils.go
pjohnst5:
```
"pods still present: [{TypeMeta:{Kind: APIVersion:} ObjectMeta:{Name:win-load-test-6fdc4cbfc5-4mhp7 GenerateName:win-load-test-6fdc4cbfc5- Namespace:load-test SelfLink: UID:7a51ae9d-60b4-41c4-90c4-fbda0b8fd78c ResourceVersion:344945 Generation:0 CreationTimestamp:2023-09-27 10:25:11 -0700 PDT DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[load-test:true os:windows pod-template-hash:6fdc4cbfc5] Annotations:map[] OwnerReferences:[{APIVersion:apps/v1 Kind:ReplicaSet Name:win-load-test-6fdc4cbfc5 UID:edc399ee-5440-4992-ab05-08949af0a223 Controller:0xc0000ff92f BlockOwnerDeletion:0xc0000ff94f}] Finalizers:[] ManagedFields:[{Manager:kube-controller-manager Operation:Update APIVersion:v1 Time:2023-09-27 10:25:11 -0700 PDT FieldsType:FieldsV1 FieldsV1:{\"f:metadata\":{\"f:generateName\":{},\"f:labels\":{\".\":{},\"f:load-test\":{},\"f:os\":{},\"f:pod-template-hash\":{}},\"f:ownerReferences\":{\".\":{},\"k:{\\\"uid\\\":\\\"edc399ee-5440-4992-ab05-08949af0a223\\\"}\":{}}},\"f:spec\":{\"f:containers\":{\"k:{\\\"name\\\":\\\"noop\\\"}\":{\".\":{},\"f:image\":{},\"f:imagePullPolicy\":{},\"f:name\":{},\"f:ports\":{\".\":{},\"k:{\\\"containerPort\\\":80,\\\"protocol\\\":\\\"TCP\\\"}\":{\".\":{},\"f:containerPort\":{},\"f:protocol\":{}}},\"f:resources\":{},\"f:terminationMessagePath\":{},\"f:terminationMessagePolicy\":{}}},\"f:dnsPolicy\":{},\"f:enableServiceLinks\":{},\"f:nodeSelector\":{},\"f:restartPolicy\":{},\"f:schedulerName\":{},\"f:securityContext\":{},\"f:terminationGracePeriodSeconds\":{},\"f:topologySpreadConstraints\":{\".\":{},\"k:{\\\"topologyKey\\\":\\\"kubernetes.io/hostname\\\",\\\"whenUnsatisfiable\\\":\\\"ScheduleAnyway\\\"}\":{\".\":{},\"f:labelSelector\":{},\"f:maxSkew\":{},\"f:topologyKey\":{},\"f:whenUnsatisfiable\":{}}}}} Subresource:} {Manager:kubelet.exe Operation:Update APIVersion:v1 Time:2023-09-27 10:25:27 -0700 PDT FieldsType:FieldsV1 FieldsV1:{\"f:status\":{\"f:conditions\":{\"k:{\\\"type\\\":\\\"ContainersReady\\\"}\":{\".\":{},\"f:lastProbeTime\":{},\"f:lastTransitionTime\":{},\"f:status\":{},\"f:type\":{}},\"k:{\\\"type\\\":\\\"Initialized\\\"}\":{\".\":{},\"f:lastProbeTime\":{},\"f:lastTransitionTime\":{},\"f:status\":{},\"f:type\":{}},\"k:{\\\"type\\\":\\\"Ready\\\"}\":{\".\":{},\"f:lastProbeTime\":{},\"f:lastTransitionTime\":{},\"f:status\":{},\"f:type\":{}}},\"f:containerStatuses\":{},\"f:hostIP\":{},\"f:phase\":{},\"f:podIP\":{},\"f:podIPs\":{\".\":{},\"k:{\\\"ip\\\":\\\"192.168.2.28\\\"}\":{\".\":{},\"f:ip\":{}}},\"f:startTime\":{}}} Subresource:status}]} Spec:{Volumes:[{Name:kube-api-access-kf7vs VolumeSource:{HostPath:nil EmptyDir:nil GCEPersistentDisk:nil AWSElasticBlockStore:nil GitRepo:nil Secret:nil NFS:nil ISCSI:nil Glusterfs:nil PersistentVolumeClaim:nil RBD:nil FlexVolume:nil Cinder:nil CephFS:nil Flocker:nil DownwardAPI:nil FC:nil AzureFile:nil ConfigMap:nil VsphereVolume:nil Quobyte:nil AzureDisk:nil PhotonPersistentDisk:nil Projected:&ProjectedVolumeSource{Sources:[]VolumeProjection{VolumeProjection{Secret:nil,DownwardAPI:nil,ConfigMap:nil,ServiceAccountToken:&ServiceAccountTokenProjection{Audience:,ExpirationSeconds:*3607,Path:token,},},VolumeProjection{Secret:nil,DownwardAPI:nil,ConfigMap:&ConfigMapProjection{LocalObjectReference:LocalObjectReference{Name:kube-root-ca.crt,},Items:[]KeyToPath{KeyToPath{Key:ca.crt,Path:ca.crt,Mode:nil,},},Optional:nil,},ServiceAccountToken:nil,},VolumeProjection{Secret:nil,DownwardAPI:&DownwardAPIProjection{Items:[]DownwardAPIVolumeFile{DownwardAPIVolumeFile{Path:namespace,FieldRef:&ObjectFieldSelector{APIVersion:v1,FieldPath:metadata.namespace,},ResourceFieldRef:nil,Mode:nil,},},},ConfigMap:nil,ServiceAccountToken:nil,},},DefaultMode:*420,} PortworxVolume:nil ScaleIO:nil StorageOS:nil CSI:nil Ephemeral:nil}}] InitContainers:[] Containers:[{Name:noop Image:mcr.microsoft.com/oss/kubernetes/pause:3.6 Command:[] Args:[] WorkingDir: Ports:[{Name: HostPort:0 ContainerPort:80 Protocol:TCP HostIP:}] EnvFrom:[] Env:[] Resources:{Limits:map[] Requests:map[] Claims:[]} ResizePolicy:[] RestartPolicy:<nil> VolumeMounts:[{Name:kube-api-access-kf7vs ReadOnly:true MountPath:/var/run/secrets/kubernetes.io/serviceaccount SubPath: MountPropagation:<nil> SubPathExpr:}] VolumeDevices:[] LivenessProbe:nil ReadinessProbe:nil StartupProbe:nil Lifecycle:nil TerminationM...+224150 more"
```

pjohnst5: Super verbose, fixed it to just print number of pods

test/validate/validate.go
pjohnst5: Sure, I'll "corret" that 😃
<crickets..>

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
pjohnst5: it's double the default batch size of 16, so it is guaranteed to trigger a scaling event in CNS

test/validate/validate.go
pjohnst5: Do you mean, inside of `GetNodeList()` ? Or just outside on the results ? 
If outside on the results, I would argue that's just the same as skipping non-windows nodes in the for loop

test/integration/load/load_test.go
pjohnst5: Yeah there have been bugs before in CNS, when we scale down, that IPs aren't released, and then after restart, those IPs are stuck so we're testing that now

test/integration/datapath/datapath_windows_test.go
pjohnst5: Actually, removing this new code that isntalls, and leaving installation code in pipeline
If we really want to later, I can remove it all pipelines and make it only in go code in "InstallCNSDaemonset" method

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
vipul-21: Wouldn't it be better to have a number i.e 40, not a multiple of 16.?

rbtr: why would that be better?

test/integration/load/load_test.go
pjohnst5: I'll add in a "ToTriggerDropgzRestart" to function name and add comments to describe why we restart cns during a pod scaling even (to trigger dropgz)

test/integration/datapath/datapath_windows_test.go
```diff
@@ -10,21 +10,26 @@ import (
...
 	apiv1 "k8s.io/api/core/v1"
 )
 
 const (
-	WindowsDeployYamlPath = "../manifests/datapath/windows-deployment.yaml"
-	podLabelKey           = "app"
-	podCount              = 2
-	nodepoolKey           = "agentpool"
+	WindowsDeployYamlPath         = "../manifests/datapath/windows-deployment.yaml"
+	IPMasqAgentYamlPath           = "../manifests/ip-masq-agent/ip-masq-agent.yaml"
```
pjohnst5, line 20: Remove these

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
pjohnst5, line 201: Add this to line here `TEST_DROPGZ=${{ parameters.testDropgz }}`

paulyufan2 (commented):
lgtm

.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
```diff
@@ -172,3 +172,41 @@ steps:
...
+  - script: |
+      nodeList=`kubectl get node -owide | grep Windows | awk '{print $1}'`
+      for node in $nodeList; do
+          taint=`kubectl describe node $node | grep Taints | awk '{print $2}'`
+          if [ $taint == "node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule" ]; then
+              kubectl taint nodes $node node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
+          fi
+      done
+      CNS=$(make cns-version)
+      sudo -E env "PATH=$PATH" make test-load SCALE_UP=32 OS_TYPE=windows CNI_TYPE=cniv2 VALIDATE_STATEFILE=true INSTALL_CNS=true INSTALL_AZURE_CNI_OVERLAY=true VALIDATE_V4OVERLAY=true CNS_VERSION=${CNS} CNI_DROPGZ_VERSION=$(dropgzVersion) TEST_DROPGZ=${{ parameters.testDropgz }} CLEANUP=true
```
jpayne3506, line 201: Are we going to use the `dropgzVersion` var set in lines 58-67?

test/internal/kubernetes/utils.go
```diff
@@ -233,6 +233,24 @@ func WaitForPodsRunning(ctx context.Context, clientset *kubernetes.Clientset, na
...
+func WaitForPodsDelete(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelselector string) error {
+	podsClient := clientset.CoreV1().Pods(namespace)
+
+	checkPodsDeleted := func() error {
+		podList, err := podsClient.List(ctx, metav1.ListOptions{LabelSelector: labelselector})
+		if err != nil {
+			return errors.Wrapf(err, "could not list pods with label selector %s", labelselector)
+		}
+		if len(podList.Items) != 0 {
+			return errors.Errorf("%d pods still present", len(podList.Items))
```
jpayne3506, line 245: Would the verbosity of `podList.Items` give a wrong count from len?

pjohnst5: At least in my debugger, each pod is an item in the slice, so I think it's the right number

jpayne3506 (approved):
lgtm

vipul-21 (approved):

pjohnst5:
Hey @rbtr can you merge this override-style for me, since it really just barely was beat by your submodules pr, and passed in the ACN pipeline?
The cni load test checks aren't required, and are failing due to an unrelated npm build issue

PR #2192, opened 2023-08-30
CNS - Ensuring no stale NCs during NNC reconcile

ramiro-gamarra:
**Reason for Change**:
There may be occasions where node upgrades may not result in delete and re-create events for an NNC, rather a new set of NCs will be available in the NNC as an update. The current reconcile logic only adds NCs to CNS, so in these cases stale NCs would remain. This PR attempts to ensure there are no stale NCs during NNC reconciles.

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/reconciler.go
+++ b/cns/kubecontroller/nodenetworkconfig/reconciler.go
@@ -24,7 +24,7 @@ import (
 
 type cnsClient interface {
 	CreateOrUpdateNetworkContainerInternal(*cns.CreateNetworkContainerRequest) cnstypes.ResponseCode
-	EnsureNoStaleNCs(validNCIDs []string)
+	MustEnsureNoStaleNCs(validNCIDs []string)
 }
 
 type nodeNetworkConfigListener interface {
@@ -82,7 +82,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	for i := range nnc.Status.NetworkContainers {
 		validNCIDs[i] = nnc.Status.NetworkContainers[i].ID
 	}
-	r.cnscli.EnsureNoStaleNCs(validNCIDs)
+	r.cnscli.MustEnsureNoStaleNCs(validNCIDs)
 
 	// for each NC, parse it in to a CreateNCRequest and forward it to the appropriate Listener
 	for i := range nnc.Status.NetworkContainers {
--- a/cns/kubecontroller/nodenetworkconfig/reconciler.go
+++ b/cns/kubecontroller/nodenetworkconfig/reconciler.go
@@ -24,6 +24,7 @@ import (
 
 type cnsClient interface {
 	CreateOrUpdateNetworkContainerInternal(*cns.CreateNetworkContainerRequest) cnstypes.ResponseCode
+	EnsureNoStaleNCs(validNCIDs []string)
 }
 
 type nodeNetworkConfigListener interface {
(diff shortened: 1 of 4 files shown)


rbtr (approved):
lgtm, thanks for fixing so fast

ashvindeodhar (approved):

thatmattlong:
LGTM

thatmattlong (approved):

thatmattlong:
@rbtr this unblocks dualstack/overlay but we'll probably still need to fix for podsubnet, basically remove the IPs from ipampool if we decide the NC is stale

neaggarwMS (changes requested):
Holding it for review

cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -74,6 +75,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	ipAssignments := 0
 
+	// during node upgrades, an nnc may be updated with new ncs. at any given time, only the ncs
+	// that exist in the nnc are valid. any others that may have been previously created and no
+	// longer exist in the nnc should be considered stale.
+	validNCIDs := make([]string, len(nnc.Status.NetworkContainers))
```
neaggarwMS, line 81: This should be done only during CNS restart at the time it is reconciling with CNI/NNC state, not here. (or may be here too, still thinking).

cns/restserver/internalapi.go
```diff
@@ -475,6 +475,25 @@ func (service *HTTPRestService) DeleteNetworkContainerInternal(
...
+func (service *HTTPRestService) EnsureNoStaleNCs(validNCIDs []string) {
+	valid := make(map[string]struct{})
+	for _, ncID := range validNCIDs {
+		valid[ncID] = struct{}{}
+	}
+
+	service.Lock()
+	defer service.Unlock()
+
+	for ncID := range service.state.ContainerStatus {
```
neaggarwMS, line 487: Lets also make sure there is no state in the cache data. like PodIPMap, and if there is then it is not assigned to any pod.

neaggarwMS (changes requested):

cns/kubecontroller/nodenetworkconfig/reconciler.go
ramiro-gamarra: The issue arises *because* CNS doesn't restart, no?

neaggarwMS: Yes we should handle cleaning up this stale NC during initial reoncile flow and before actually starting http listener.

neaggarwMS: wait, if it is node reimage then i would expect CNS to start no?

ashvindeodhar: regardless of the CNS restart, this should be enough to take care of the NC updates in NNC and remove stale NCs, no? @rbtr ?
Do we not execute reconcile upon CNS restart too?

cns/restserver/internalapi.go
ashvindeodhar: is this possible? we won't be able to rebuild the cache with stale data since 'service.state.ContainerStatus' doesn't have the stale NC anymore, no?

cns/kubecontroller/nodenetworkconfig/reconciler.go
thatmattlong: IMO we don't need to do this at startup, we have to do it for each status update anyway so when reconciler starts, it will get an event for this status and do the cleanup

rbtr (approved):

ashvindeodhar (approved):

ashvindeodhar:
All checks are passed, the change has incorporated all the feedback from Neha along with tests. Merging this change.

PR #2193, opened 2023-08-30
Feat: Created a Network Container Status Section to be updated with the latest error code

nairashu:
Feat: Created a Network Container Status Section to be updated with the latest error code

**Reason for Change**:
This change is to introduce a section to include the latest error code section in the NNC and other NC statuses. This section will help to populate and push the subnet full event into the NNC and can be reflected further in the CNI events.

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:


crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -62,10 +62,17 @@ const (
...
-	NetworkContainers []NetworkContainer `json:"networkContainers,omitempty"`
+	AssignedIPCount          int                      `json:"assignedIPCount"`
+	Scaler                   Scaler                   `json:"scaler,omitempty"`
+	Status                   Status                   `json:"status,omitempty"`
+	NetworkContainers        []NetworkContainer       `json:"networkContainers,omitempty"`
+	NetworkContainerStatuses []NetworkContainerStatus `json:"networkContainerStatuses,omitempty"`
+}
+
+// NetworkContainerStatus defines the status of the Network Containers
+type NetworkContainerStatus struct {
```
ramiro-gamarra, line 73: any reason not to have this be a field within `NetworkContainer`?

nairashu: My argument for that is the Scaler and Status fields were not included in the Network Containers as well. My thought process is to avoid breaking change, so continued along that path.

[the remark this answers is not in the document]
nairashu: My argument for that is the Scaler and Status fields were not included in the Network Containers as well. My thought process is to avoid breaking change, so continued along that path.

nairashu: Updated to include it into the Network container with a timestamp.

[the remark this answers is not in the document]
nairashu: Updated to include it into the Network container with a timestamp.

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -68,6 +68,13 @@ type NodeNetworkConfigStatus struct {
 	NetworkContainers []NetworkContainer `json:"networkContainers,omitempty"`
 }
 
+// NetworkContainerStatus defines the status of the Network Containers
+type NetworkContainerStatus struct {
+	Timestamp       metav1.Time `json:"timestamp,omitempty"`
+	LatestStatus    string      `json:"latestStatus,omitempty"`
+	LatestErrorText string      `json:"latestErrorText,omitempty"`
```
ramiro-gamarra, line 75: - Is the "latest" prefix needed? At any given time this struct should already reflect the latest status no?
- Are you only including the error text? What about the numeric error code which is better for handling than arbitrary text?

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -68,6 +68,13 @@ type NodeNetworkConfigStatus struct {
 	NetworkContainers []NetworkContainer `json:"networkContainers,omitempty"`
 }
 
+// NetworkContainerStatus defines the status of the Network Containers
+type NetworkContainerStatus struct {
+	Timestamp       metav1.Time `json:"timestamp,omitempty"`
+	LatestStatus    string      `json:"latestStatus,omitempty"`
```
ramiro-gamarra, line 74: What are the values that can go here? Actual NC processing is a combination of both State and Status, i.e `Publish` and `Pending`

nairashu: This should only be updating the Final Status during DNC-RC reconciliation if the NC update was successful or if there was an error.

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
nairashu: I can get rid of the Latest prefix. The reason I added only the ErrorText was the SubnetFull error was only seen in the Error text and the error code is a more general failure in the DNC code. Had ErrorCode initially but then removed it. Will introduce it back.

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -68,6 +68,13 @@ type NodeNetworkConfigStatus struct {
 	NetworkContainers []NetworkContainer `json:"networkContainers,omitempty"`
 }
 
+// NetworkContainerStatus defines the status of the Network Containers
```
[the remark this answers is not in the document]
nairashu: Addressed

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -68,6 +68,14 @@ type NodeNetworkConfigStatus struct {
 	NetworkContainers []NetworkContainer `json:"networkContainers,omitempty"`
 }
 
+// NetworkContainerStatus defines the status of the Network Container
+type NetworkContainerStatus struct {
+	Timestamp metav1.Time `json:"timestamp,omitempty"`
+	Status    string      `json:"status,omitempty"`
+	ErrorCode string      `json:"errorCode,omitempty"`
+	ErrorText string      `json:"errorText,omitempty"`
```
thatmattlong, line 76: if this is just plumbing for well-known error codes, do we need an error text here? Can CNS just convert the well-known error code into some text?

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -68,6 +68,14 @@ type NodeNetworkConfigStatus struct {
 	NetworkContainers []NetworkContainer `json:"networkContainers,omitempty"`
 }
 
+// NetworkContainerStatus defines the status of the Network Container
+type NetworkContainerStatus struct {
+	Timestamp metav1.Time `json:"timestamp,omitempty"`
```
thatmattlong, line 73: what's the purpose of including a timestamp?

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
nairashu: Problem is SubnetIsFull is not a well defined error code inside DNC and currently is being sent in as a text. I would like to keep both errorcode and error text for now

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
nairashu: Updating the NC status based on the timestamp on the CNS side. You always want the latest status only to be reflected in CNS.

ramiro-gamarra: Not sure I follow. The NNC should always contain the latest NC state, including the latest status. CNS should always act on whatever latest NNC state there is. How will this timestamp be used?

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -68,6 +68,14 @@ type NodeNetworkConfigStatus struct {
...
+	Status    string      `json:"status,omitempty"`
```
ramiro-gamarra, line 74: The other thread became outdated, so following up on my previous question. What are the possible values for this field? Just `Pending`, `Complete`, and `Failed`?

nairashu: Yes just those 3 values reflecting the DNC response

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
nairashu: I think I agree here. Removing Timestamp

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
nairashu: Using the Status field to represent anything related to this NC like RequestIPBatch can be the update made by CNS. DNC-RC can update it with PendingIPAllocation, RequestFailed, RequestCompleted.

nairashu: It should be highly useful with Dual stack where one NC could have a status update while the other could be in a different state.

nairashu: I will not be using this field with CNS currently but as mentioned can be extended for the future if required. We will not have DNC-RC and CNS ever overwrite each other by extending the use of this status IMO.
Today, CNS only updates the requested IP count in the NNCSpec to request an additional batch or release a batch. Only if SubnetExhaustionFeature is enabled will CNS update the Scaler section to change batch size. Currently, subnet exhaustion feature is not enabled anywhere.

With the current design, I propose moving the status object inside the NC Status to independently represent each NC's status along with its error code and error text from. These will ideally be updated only by DNC-RC today.

Future extension proposal for this: (Future proposal not in this work item or for any current changes)
* CNS updates Status of Request/Release IPs and DNC-RC only works on the NNC update event if the status is set to CNS communication events through this status field.
* DNC-RC updates this status with the NC response statuses as today with Pending/Success/Failure. CNS reconciles only on these specific RC communication events.

jaer-tsun (dismissed):
lgtm

neaggarwMS (approved):

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
nairashu: Removed the Status and simplified this change.

neaggarwMS (approved):

neaggarwMS (approved):

rbtr (approved):

PR #2197, opened 2023-08-31
🌈 feat: StatelessCNI: Applying stateless CNI mode changes in network package.

behzad-mir:
**Reason for Change**:
Applying the required stateless CNI mode changes in network package.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -149,8 +149,7 @@ func (plugin *NetPlugin) Start(config *common.PluginConfig) error {
 	// Log platform information.
 	logger.Info("Plugin Info",
 		zap.String("name", plugin.Name),
-		zap.String("version", plugin.Version),
-		zap.String("component", "cni-net"))
+		zap.String("version", plugin.Version))
 
 	// Initialize network manager. rehyrdration not required on reboot for cni plugin
 	err = plugin.nm.Initialize(config, false)
@@ -1051,6 +1050,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 				// Log the error but return success if the network is not found.
 				// if cni hits this, mostly state file would be missing and it can be reboot scenario where
 				// container runtime tries to delete and create pods which existed before reboot.
+				// this condition will not apply to stateless CNI since the network struct will be crated on each call
 				err = nil
 				if !plugin.nm.IsStatelessCNIMode() {
 					return err
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -149,7 +149,8 @@ func (plugin *NetPlugin) Start(config *common.PluginConfig) error {
 	// Log platform information.
 	logger.Info("Plugin Info",
 		zap.String("name", plugin.Name),
-		zap.String("version", plugin.Version))
+		zap.String("version", plugin.Version),
+		zap.String("component", "cni-net"))
 
 	// Initialize network manager. rehyrdration not required on reboot for cni plugin
 	err = plugin.nm.Initialize(config, false)
@@ -486,7 +487,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		options := make(map[string]any)
 		networkID, err = plugin.getNetworkName(args.Netns, &ipamAddResult, nwCfg)
 
-		endpointID := GetEndpointID(args)
+		endpointID := plugin.nm.GetEndpointID(args.ContainerID, args.IfName)
 		policies := cni.GetPoliciesFromNwCfg(nwCfg.AdditionalArgs)
 
 		// Check whether the network already exists.
@@ -1051,11 +1052,13 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 				// if cni hits this, mostly state file would be missing and it can be reboot scenario where
 				// container runtime tries to delete and create pods which existed before reboot.
 				err = nil
-				return err
+				if !plugin.nm.IsStatelessCNIMode() {
+					return err
+				}
 			}
 		}
 
-		endpointID := GetEndpointID(args)
+		endpointID := plugin.nm.GetEndpointID(args.ContainerID, args.IfName)
 		// Query the endpoint.
 		if epInfo, err = plugin.nm.GetEndpointInfo(networkID, endpointID); err != nil {
 			logger.Info("[cni-net] GetEndpoint",
@@ -1086,7 +1089,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 			zap.String("endpointID", endpointID))
 		sendEvent(plugin, fmt.Sprintf("Deleting endpoint:%v", endpointID))
 		// Delete the endpoint.
-		if err = plugin.nm.DeleteEndpoint(networkID, endpointID); err != nil {
+		if err = plugin.nm.DeleteEndpoint(networkID, endpointID, epInfo); err != nil {
 			// return a retriable error so the container runtime will retry this DEL later
 			// the implementation of this function returns nil if the endpoint doens't exist, so
 			// we don't have to check that here
--- a/network/manager.go
+++ b/network/manager.go
@@ -9,7 +9,6 @@ import (
 	"sync"
 	"time"
 
-	cnilogger "github.com/Azure/azure-container-networking/cni/log"
 	cnms "github.com/Azure/azure-container-networking/cnms/cnmspackage"
 	cnsclient "github.com/Azure/azure-container-networking/cns/client"
 	"github.com/Azure/azure-container-networking/common"
@@ -35,6 +34,9 @@ const (
 	cnsBaseURL           = "" // fallback to default http://localhost:10090
 	cnsReqTimeout        = 15 * time.Second
 	StateLessCNIIsNotSet = "StateLess CNI mode is not enabled"
+	InfraInterfaceName   = "eth0"
+	ContainerIDLength    = 8
+	EndpointIfIndex      = 0 // Azure CNI supports only one interface
 )
 
 var Ipv4DefaultRouteDstPrefix = net.IPNet{
@@ -69,7 +71,7 @@ type EndpointClient interface {
 
 // NetworkManager manages the set of container networking resources.
 type networkManager struct {
-	StetlessCniMode    bool
+	StatelessCniMode   bool
 	CnsClient          *cnsclient.Client
 	Version            string
 	TimeStamp          time.Time
@@ -125,9 +127,9 @@ func NewNetworkManager(nl netlink.NetlinkInterface, plc platform.ExecClient, net
 func (nm *networkManager) Initialize(config *common.PluginConfig, isRehydrationRequired bool) error {
 	nm.Version = config.Version
 	nm.store = config.Store
-	if config.Stateless == true {
-		nm.SetStatelessCNIMode()
-		return nil
+	if config.Stateless {
+		err := nm.SetStatelessCNIMode()
+		return errors.Wrapf(err, "Failed to initialize stateles CNI")
 	}
 
 	// Restore persisted state.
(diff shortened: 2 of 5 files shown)


network/manager.go
```diff
@@ -354,11 +403,56 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
...
 }
 
+// UpdateEndpointState will make a call to CNS updatEndpointState API in the stateless CNI mode
+// It will add HNSEndpointID or HostVeth name to the endpoint state
+func (nm *networkManager) UpdateEndpointState(ep *endpoint) error {
+	cnilogger.Logger.Info("Calling cns updateEndpoint API with ", zap.String("containerID: ", ep.ContainerID), zap.String("HnsId: ", ep.HnsId), zap.String("HostIfName: ", ep.HostIfName))
+	response, err := nm.CnsClient.UpdateEndpoint(context.TODO(), ep.ContainerID, ep.HnsId, ep.HostIfName)
+	if err != nil {
+		cnilogger.Logger.Error("Update endpoint API returend with error", zap.Error(err))
+		return errors.Wrapf(err, "Get endpoint API returend with error")
```
vipul-21, line 413: The error description is saying `Get endpoint` instead of `Update endpoint`

network/manager.go
```diff
@@ -354,11 +403,56 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
...
+
 // DeleteEndpoint deletes an existing container endpoint.
-func (nm *networkManager) DeleteEndpoint(networkID, endpointID string) error {
+func (nm *networkManager) DeleteEndpoint(networkID, endpointID string, epInfo *EndpointInfo) error {
 	nm.Lock()
 	defer nm.Unlock()
 
+	if nm.IsStatelessCNIMode() {
+		nw := &network{
+			Id:           "azure",
```
vipul-21, line 426: We can use the `networkID` here ?

network/manager.go
```diff
@@ -382,6 +476,31 @@ func (nm *networkManager) GetEndpointInfo(networkId string, endpointId string) (
 	nm.Lock()
 	defer nm.Unlock()
 
+	if nm.IsStatelessCNIMode() {
+		cnilogger.Logger.Info("calling cns getEndpoint API")
+		endpointResponse, err := nm.CnsClient.GetEndpoint(context.TODO(), endpointId)
+		if err != nil {
+			cnilogger.Logger.Error("Get endpoint API returend with error", zap.Error(err))
+			return nil, errors.Wrapf(err, "Get endpoint API returend with error")
```
behzad-mir, line 484: @tamilmani1989 
I can check in here if the state is not found we return err=nil. But I'm afarid we might create other issues and prevent a cleanup.

network/manager.go
```diff
@@ -298,6 +340,7 @@ func (nm *networkManager) DeleteNetwork(networkID string) error {
 
 // GetNetworkInfo returns information about the given network.
 func (nm *networkManager) GetNetworkInfo(networkId string) (NetworkInfo, error) {
+
```
tamilmani1989, line 343: nit: space.. revert this change

network/manager.go
```diff
@@ -326,6 +369,7 @@ func (nm *networkManager) GetNetworkInfo(networkId string) (NetworkInfo, error)
 
 // CreateEndpoint creates a new container endpoint.
 func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epInfo *EndpointInfo) error {
+
```
tamilmani1989, line 372: nit: space.. revert

network/manager.go
```diff
@@ -336,16 +380,21 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 
 	if nw.VlanId != 0 {
 		if epInfo.Data[VlanIDKey] == nil {
+			cnilogger.Logger.Info("overriding endpoint vlanid with network vlanid")
```
tamilmani1989, line 383: why logging again? can you remove this

network/manager.go
```diff
@@ -269,7 +307,9 @@ func (nm *networkManager) CreateNetwork(nwInfo *NetworkInfo) error {
 	if err != nil {
 		return err
 	}
-
+	if nm.IsStatelessCNIMode() {
```
tamilmani1989, line 310: can we move this check inside nm.save() instead of repeating in multiple places?

network/manager.go
tamilmani1989, line 426: lets not consider `azure` here instead lets write api to get network name.. incase of multitenancy network name has to be derived from vlan id and subnet space

network/manager.go
```diff
@@ -354,11 +403,56 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
...
 	nm.Lock()
 	defer nm.Unlock()
 
+	if nm.IsStatelessCNIMode() {
+		nw := &network{
+			Id:           "azure",
+			Mode:         opModeTransparentVlan,
+			SnatBridgeIP: "",
+			extIf: &externalInterface{
+				Name:       "eth0",
```
tamilmani1989, line 430: lets not hardcode interface name as well

network/manager.go
```diff
@@ -354,11 +403,56 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
...
+				MacAddress: nil,
+			},
+		}
+
```
tamilmani1989, line 434: i would suggest to move this block to separate function

network/manager.go
```diff
@@ -354,11 +403,56 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
...
+			HostIfName:               epInfo.IfName,
+			LocalIP:                  "",
+			VlanID:                   0,
+			AllowInboundFromHostToNC: false,
+			AllowInboundFromNCToHost: false,
+			EnableSnatOnHost:         false,
+			EnableMultitenancy:       false,
+			NetworkContainerID:       epInfo.Id,
+		}
+		netlink.NewNetlink()
```
tamilmani1989, line 447: why this is required?

network/manager.go
```diff
@@ -354,11 +403,56 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
...
+		cnilogger.Logger.Info("Deleting endpoint with", zap.String("Endpoint Info: ", epInfo.PrettyString()), zap.String("HNISID : ", ep.HnsId))
+		err := nw.deleteEndpointImpl(netlink.NewNetlink(), platform.NewExecClient(), nil, ep)
+		if err != nil {
+			return err
+		}
+		return nil
```
tamilmani1989, line 453: can it be return err?

network/manager.go
```diff
@@ -382,6 +476,31 @@ func (nm *networkManager) GetEndpointInfo(networkId string, endpointId string) (
...
+	if nm.IsStatelessCNIMode() {
+		cnilogger.Logger.Info("calling cns getEndpoint API")
+		endpointResponse, err := nm.CnsClient.GetEndpoint(context.TODO(), endpointId)
+		if err != nil {
+			cnilogger.Logger.Error("Get endpoint API returend with error", zap.Error(err))
+			return nil, errors.Wrapf(err, "Get endpoint API returend with error")
+		}
+		epInfo := &EndpointInfo{
+			Id:                 endpointId,
+			IfIndex:            0, // Azure CNI supports only one interface
```
tamilmani1989, line 488: For ifindex, we should  return whatever existing cni returns.. if required we can get interface index based on interface name

tamilmani1989 (changes requested):

network/manager.go
tamilmani1989: should check if error is `endpoint not found` and then ignore.. cannot ignore all other errors

cni/network/network.go
```diff
@@ -1052,11 +1056,13 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 				// if cni hits this, mostly state file would be missing and it can be reboot scenario where
 				// container runtime tries to delete and create pods which existed before reboot.
 				err = nil
-				return err
+				if !plugin.nm.IsStatelessCNIMode() {
```
vipul-21, line 1059: nit: Can we add a comment here why we need to do this for stateless CNI ?

network/manager.go
```diff
@@ -113,6 +125,10 @@ func NewNetworkManager(nl netlink.NetlinkInterface, plc platform.ExecClient, net
 func (nm *networkManager) Initialize(config *common.PluginConfig, isRehydrationRequired bool) error {
 	nm.Version = config.Version
 	nm.store = config.Store
+	if config.Stateless == true {
+		nm.SetStatelessCNIMode()
```
vipul-21, line 129: We are bypassing the error returned by `SetStatelessCNIMode`, we should return that error

network/manager.go
```diff
@@ -61,6 +69,8 @@ type EndpointClient interface {
 
 // NetworkManager manages the set of container networking resources.
 type networkManager struct {
+	StetlessCniMode    bool
```
vipul-21, line 72:
```suggestion
	StatelessCniMode    bool
```

network/manager.go
```diff
@@ -123,6 +139,26 @@ func (nm *networkManager) Initialize(config *common.PluginConfig, isRehydrationR
 func (nm *networkManager) Uninitialize() {
 }
 
+// SetStatelessCNIMode enable the statelessCNI falg and inititlizes a CNSClient
+func (nm *networkManager) SetStatelessCNIMode() error {
+	nm.StetlessCniMode = true
```
vipul-21, line 144:
```suggestion
	nm.StatelessCniMode = true
```

network/manager.go
```diff
@@ -123,6 +139,26 @@ func (nm *networkManager) Initialize(config *common.PluginConfig, isRehydrationR
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to initialize CNS client")
+	}
+	nm.CnsClient = client
+	return nil
+}
+
+// IsStatelessCNIMode checks if the Stateless CNI mode has been enabled or not
+func (nm *networkManager) IsStatelessCNIMode() bool {
+	if nm.StetlessCniMode == true {
```
vipul-21, line 156:
```suggestion
	if nm.StatelessCniMode == true {
```

cni/network/network.go
```diff
@@ -150,16 +150,19 @@ func (plugin *NetPlugin) Start(config *common.PluginConfig) error {
 	// Log platform information.
 	log.Logger.Info("Plugin Info",
 		zap.String("name", plugin.Name),
-		zap.String("version", plugin.Version))
+		zap.String("version", plugin.Version),
+		zap.String("component", "cni-net"))
```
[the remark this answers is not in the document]
behzad-mir: addressed in the next commit

cni/network/network.go
behzad-mir: addressed in the next commit

network/manager.go
behzad-mir: addressed in the next commit.

network/manager.go
behzad-mir: addressed in the next commit.

network/manager.go
behzad-mir: addressed in the next commit

network/manager.go
behzad-mir: addressed in the next commit

behzad-mir: I am using the networkId that is passed tot he func for now.

network/manager.go
behzad-mir: addressed in the next commit

network/manager.go
behzad-mir: forgot to clean up.

network/manager.go
behzad-mir: We should still return err here back to the cni/network.go Delete func so it can just exit gracefully the same as legacy CNI.

network/manager.go
behzad-mir: This value Has been hardcoded on existing GetEndpointInfo API with 0. I just followed the same. For the new commit, I just added a constant value.

network/manager.go
behzad-mir: A constant has been added.

PR #2202, opened 2023-09-05
feat: Consume the NCStatus to be able to append subnet is full error to Pod IP reservation failures

nairashu:
Consume the NCStatus to be able to append subnet is full error to Pod IP reservation failures

**Reason for Change**:
Consume the NCStatus to be able to append subnet is full error to Pod IP reservation failures

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -65,6 +65,8 @@ type NCStatus string
 
 const (
 	NCStatusSubnetFull NCStatus = "SubnetFull"
```
vipul-21, line 67: nit: Does `NCSubnetFull` aligns more with the `NCUpdateSuccess` and `NCUpdateFailed` ?

cns/restserver/ipam.go
```diff
@@ -790,6 +791,12 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
 	if numOfNCs == 0 {
 		return nil, ErrNoNCs
 	}
+	// slice to store NCIDs
+	ncIDs := make([]string, 0)
```
vipul-21, line 795: Why not preallocate the slice with the length as `numOfNCs` ? More over can we use `service.state.ContainerStatus` directly in the loop below ?

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
nairashu: So NCSubnetFull will be more accurate representation when the Subnet is Full error is present in DNC to pass on this error. NCUpdate Success is to overwrite when the NC was updated successfully from DNC and NCUpdateFailure is for any other error Updating the NC from DNC-RC's perspective. I am using this field to pass and communicate errors between DNC and the node services like CNS and CNI

cns/restserver/ipam.go
nairashu: Changed the allocation size to number of NCs. We are already looping over Container Status not sure what you meant by that part of the comment

cns/restserver/ipam.go
```diff
@@ -816,8 +823,17 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
...
 	if len(ipsToAssign) != numOfNCs {
-		//nolint:goerr113 // return error
-		return podIPInfo, fmt.Errorf("not enough IPs available, waiting on Azure CNS to allocate more")
+		for _, ncID := range ncIDs {
+			if _, found := ipsToAssign[ncID]; found {
+				continue
+			}
+			if service.state.ContainerStatus[ncID].CreateNetworkContainerRequest.NCStatus == v1alpha.NCStatusSubnetFull {
+				//nolint:goerr113 // return error
+				return podIPInfo, fmt.Errorf("not enough IPs available for %s, waiting on Azure CNS to allocate more but the Subnet is full", ncID)
```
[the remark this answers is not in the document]
nairashu: Updated to errors.Errorf

cns/restserver/ipam.go
vipul-21: We can use the `numOfNCs` for the length of the slice ? i.e `make([]string, numOfNCs)`.  Added this because the performance(speed and memory i/o) with predefined length is better than dynamic size of the slice. (https://dev.to/jonathanlawhh/golang-slice-vs-array-and-should-i-declare-my-slice-size-4kng)

For the looping part, I meant why didn't we do this (this would eliminate the variables/slice):
```
...
		for ncId := range service.state.ContainerStatus {
		...
		}

...
```

cns/restserver/ipam.go
```diff
@@ -816,8 +823,15 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
...
 	// Checks to make sure we found one IP for each NC
 	if len(ipsToAssign) != numOfNCs {
-		//nolint:goerr113 // return error
-		return podIPInfo, fmt.Errorf("not enough IPs available, waiting on Azure CNS to allocate more")
+		for _, ncID := range ncIDs {
+			if _, found := ipsToAssign[ncID]; found {
+				continue
+			}
+			if service.state.ContainerStatus[ncID].CreateNetworkContainerRequest.NCStatus == v1alpha.NCStatusSubnetFull {
+				return podIPInfo, errors.Errorf("not enough IPs available for %s, waiting on Azure CNS to allocate more but the Subnet is full", ncID)
```
neaggarwMS, line 831: we should print the error code instead of expanding that in the error string, that way code gets updated all the way to containerd and cx can filter/take action on those error codes

cns/restserver/ipam.go
```diff
@@ -816,8 +823,15 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
 
 	// Checks to make sure we found one IP for each NC
 	if len(ipsToAssign) != numOfNCs {
```
neaggarwMS, line 825: this doesnt seem to be the right place for the check, shouldnt it be in line #856

neaggarwMS (changes requested):

cns/restserver/ipam.go
nairashu: Made an update based on these comments

cns/restserver/ipam.go
nairashu: Line #856 will be too late as over here itself we determine if there are enough IPs present to allocate to the pod from each NC. If ips to assign are not sufficient we will return right here.

cns/restserver/ipam.go
nairashu: Printing the status code directly so that the data can go all the way to containerd.

cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -86,6 +86,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	// for each NC, parse it in to a CreateNCRequest and forward it to the appropriate Listener
 	for i := range nnc.Status.NetworkContainers {
+		// if the NC status has not been updated successfully skip the reconcilliation
+		noIPs := nnc.Status.NetworkContainers[i].IPAssignments == nil || len(nnc.Status.NetworkContainers[i].IPAssignments) == 0
```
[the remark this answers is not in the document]
nairashu: Removed the changes to handle 0 ip allocations

cns/restserver/ipam.go
nairashu: Discussed offline and removed this code to facilitate iteration directly where ncIDs are required

neaggarwMS (approved):

rbtr (approved):

PR #2205, opened 2023-09-07
Windows Dropgz v4overlay tests

pjohnst5:
Windows v4overlay dropgz tests
- Introduces windows v4overlay tests
- Uses windows `dropgz` to install CNI

### DIFF
--- a/.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
+++ b/.pipelines/singletenancy/azure-cni-overlay/azure-cni-overlay-e2e-step-template.yaml
@@ -172,3 +172,41 @@ steps:
     name: "Cleanupartifactdir"
     displayName: "Cleanup artifact dir"
     condition: always()
+
+  - task: AzureCLI@2
+    inputs:
+      azureSubscription: $(BUILD_VALIDATIONS_SERVICE_CONNECTION)
+      scriptLocation: "inlineScript"
+      scriptType: "bash"
+      addSpnToEnvironment: true
+      inlineScript: |
+        set -e
+        make -C ./hack/aks windows-nodepool-up AZCLI=az SUB=$(SUB_AZURE_NETWORK_AGENT_BUILD_VALIDATIONS) CLUSTER=${{ parameters.clusterName }}
+        echo "Windows node are successfully added to v4 Overlay Cluster"
+        kubectl cluster-info
+        kubectl get node -owide
+        kubectl get po -owide -A
+    name: "Add_Windows_Node"
+    displayName: "Add windows node on v4 overlay cluster"
+
+  - script: |
+      nodeList=`kubectl get node -owide | grep Windows | awk '{print $1}'`
+      for node in $nodeList; do
+          taint=`kubectl describe node $node | grep Taints | awk '{print $2}'`
+          if [ $taint == "node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule" ]; then
+              kubectl taint nodes $node node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
+          fi
+      done
+      CNS=$(make cns-version) DROPGZ=$(make cni-dropgz-version)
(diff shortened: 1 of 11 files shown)


test/integration/manifests/cns/daemonset.yaml
```diff
@@ -131,4 +131,150 @@ spec:
...
+            privileged: true
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+          volumeMounts:
+            - mountPath: /k/azurecns
+              name: log
+            - mountPath: /etc/azure-cns
+              name: cns-config
+      initContainers:
+        - name: delete-azure-vnet-telemetry
```
pjohnst5, line 241: ## What do we know?
- In Windows, you can rename a running process (binary)
- In Windows, you **cannot**, overwrite a running process (binary)
    -  It errors with "Access is denied"
    - This is why `cp` did not work in my powershell example (if you were there), but `mv` and `Rename-Item` did

## What does this mean for us?
- azure-vnet-telemetry.exe is a long-lived process, and we cannot overwrite the `azure-vnet-telemetry.exe.old` file, the 2nd time that `dropgz` runs
    - 1st time `dropgz` runs is fine (it can rename long-running `azure-vnet-telemetry.exe` to `azure-vnet-telemetry.exe.old` just fine
    - 2nd time however, `dropgz` tries to overwrite existing `azure-vnet-telemetry.exe.old`, but `azure-vnet-telemetry.exe.old` is still running as before (since it's long-lived)

## Solution
- We kill the `azure-vnet-telemetry.exe` process, before installing new `azure-vnet-telemetry.exe` binary through `dropgz`
- Then, since `dropgz` is an init container to `cns`, `cns` restarts and also restarts `azure-vnet-telemetry.exe`

pjohnst5: - `azure-vnet.exe` is not a problem, since we can rename binaries as they are running (that is the only case we would hit with `azure-vnet.exe`, since it is short-lived)
- The only possible issue with `azure-vnet.exe`, is if we happen to update `dropgz` **twice**, right in a row, **while** an `azure-vnet.exe` call is running in the background
     - This would error once, but then dropgz would retry and succeed the next call

test/integration/manifests/cnsconfig/overlayconfigmap.yaml
```diff
@@ -29,3 +29,37 @@ data:
...
+          "NodeSyncIntervalInSeconds": 30
+      },
+      "EnableSubnetScarcity": false,
+      "ChannelMode": "CRD",
+      "InitializeFromCNI": true,
+      "ProgramSNATIPTables": false,
+      "ManageEndpointState": false,
+      "MetricsBindAddress": ":10092",
+      "EnableCNIConflistGeneration": false,
+      "CNIConflistFilepath": "C:\\k\\azurecni\\etconf\\10-azure.conflist",
```
paulyufan2, line 63: should it be C:\\k\\azurecni\\**netconf**\\10-azure.conflist

test/integration/manifests/cns/daemonset.yaml
```diff
@@ -131,4 +131,150 @@ spec:
             path: /run/xtables.lock
             type: File
       serviceAccountName: azure-cns
-
```
paulyufan2, line 134: So if we add windows daemonset to the same daemonset.yaml, when it tries to install CNS daemonset, installCNSDaemonset() just installs Linux CNS daemonset, can we set up the windows CNS conifgmap for azure cni overlay here?

test/integration/manifests/cnsconfig/overlayconfigmap.yaml
pjohnst5: Yes I think you're right

pjohnst5: One thing, is I think dropgz should be installing the conflist, at least like it does for linux today:
```
initContainers:
  - args:
    - deploy
    - azure-vnet
    - -o
    - /opt/cni/bin/azure-vnet
    - azure-vnet-telemetry
    - -o
    - /opt/cni/bin/azure-vnet-telemetry
    - azure-swift-overlay.conflist
    - -o
    - /etc/cni/net.d/15-azure-swift-overlay.conflist
    command:
    - /dropgz
```

test/integration/manifests/cns/daemonset.yaml
pjohnst5: Ah, you're right, did not see this, thanks Paul

test/integration/manifests/cnsconfig/overlayconfigmap.yaml
paulyufan2: yeah, dropgz should be able to install conflist on windows as well.

pjohnst5:
Closed in favor of #2187

PR #2207, opened 2023-09-08
feat: update contracts to support swift 2

jaer-tsun:
**Reason for Change**:
updating contracts to support swift 2
- update contract when CNI RequestIPs from CNS to include 
> - MacAddress of secondary (multitenant) interface
> - AddressType to define whether the IP config is default (existing scenarios) or secondary/multitenant (swift 2)
> - IsDefaultInterface to specify which interface traffic will go through by default
> - Routes to configure on interface
- combine default v4 and v6 results in invokers and add slice of CNIResult for secondary results
- add Default and Secondary address types
- add SecondaryInterfaces to endpoint struct to store new multitenant/secondary interface(s)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/invoker.go
+++ b/cni/network/invoker.go
@@ -37,7 +37,7 @@ type IPAMAddResult struct {
 
 type CNIResult struct {
 	ipResult           *cniTypesCurr.Result
-	addressType        string
+	nicType            string
 	macAddress         net.HardwareAddr
 	isDefaultInterface bool
 }
--- a/cni/network/invoker.go
+++ b/cni/network/invoker.go
@@ -27,8 +27,16 @@ type IPAMAddConfig struct {
 }
 
 type IPAMAddResult struct {
-	ipv4Result       *cniTypesCurr.Result
-	ipv6Result       *cniTypesCurr.Result
+	defaultCniResult CNIResult
+	cniResults       []CNIResult
 	ncResponse       *cns.GetNetworkContainerResponse
 	hostSubnetPrefix net.IPNet
+	ipv6Enabled      bool
+}
+
+type CNIResult struct {
+	ipResult           *cniTypesCurr.Result
+	addressType        string
+	macAddress         net.HardwareAddr
+	isDefaultInterface bool
 }
--- a/cni/network/invoker.go
+++ b/cni/network/invoker.go
@@ -27,17 +27,18 @@ type IPAMAddConfig struct {
 }
 
 type IPAMAddResult struct {
-	// Splitting defaultCniResult from (secondary) cniResults so we don't need to loop for default CNI result every time
-	defaultCniResult CNIResult
-	cniResults       []CNIResult
+	// Splitting defaultInterfaceInfo from secondaryInterfaceInfo so we don't need to loop for default CNI result every time
+	defaultInterfaceInfo   InterfaceInfo
+	secondaryInterfaceInfo []InterfaceInfo
+	// ncResponse is used for Swift 1.0 multitenancy
 	ncResponse       *cns.GetNetworkContainerResponse
 	hostSubnetPrefix net.IPNet
 	ipv6Enabled      bool
 }
 
-type CNIResult struct {
-	ipResult           *cniTypesCurr.Result
-	nicType            string
-	macAddress         net.HardwareAddr
-	isDefaultInterface bool
+type InterfaceInfo struct {
+	ipResult          *cniTypesCurr.Result
+	nicType           string
+	macAddress        net.HardwareAddr
+	skipDefaultRoutes bool
 }
--- a/cni/network/invoker.go
+++ b/cni/network/invoker.go
@@ -27,6 +27,7 @@ type IPAMAddConfig struct {
 }
 
 type IPAMAddResult struct {
+	// Splitting defaultCniResult from (secondary) cniResults so we don't need to loop for default CNI result every time
 	defaultCniResult CNIResult
 	cniResults       []CNIResult
 	ncResponse       *cns.GetNetworkContainerResponse
--- a/network/endpoint.go
+++ b/network/endpoint.go
@@ -87,7 +87,7 @@ type EndpointInfo struct {
 	VnetCidrs                string
 	ServiceCidrs             string
 	NATInfo                  []policy.NATInfo
-	AddressType              string
+	NICType                  string
 	IsDefaultInterface       bool
 }
 
@@ -110,7 +110,7 @@ type InterfaceInfo struct {
 	IPAddress          []net.IPNet
 	Gateways           []net.IP
 	Routes             []RouteInfo
-	AddressType        string
+	NICType            string
 	IsDefaultInterface bool
 }
 
--- a/network/endpoint.go
+++ b/network/endpoint.go
@@ -88,7 +88,7 @@ type EndpointInfo struct {
 	ServiceCidrs             string
 	NATInfo                  []policy.NATInfo
 	NICType                  string
-	IsDefaultInterface       bool
+	SkipDefaultRoutes        bool
 }
 
 // RouteInfo contains information about an IP route.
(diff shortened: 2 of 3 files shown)


neaggarwMS (approved):

cns/NetworkContainerContract.go
```diff
@@ -404,6 +410,10 @@ type PodIpInfo struct {
 	PodIPConfig                     IPSubnet
```
miguelgoms, line 410: we should add comments to the PodIpInfo interface at this point.
we're adding enough fields it's not just a tuple of other ipconfigs anymore - including an enum of non-obvious allowed string values.

jaer-tsun: sure, will update with comments

miguelgoms (dismissed):

cns/NetworkContainerContract.go
```diff
@@ -69,6 +69,12 @@ const (
 	Vxlan = "Vxlan"
 )
 
+// Address Types
+const (
+	Default   = "Default"
+	Secondary = "Secondary"
```
nddq, line 75: Did we decided to change the address type from `Multitenant` to `Secondary`?

nddq (approved):
LGTM, see comment.

cns/NetworkContainerContract.go
```diff
@@ -69,6 +69,12 @@ const (
 	Vxlan = "Vxlan"
 )
 
+// Address Types
+const (
```
tamilmani1989, line 73: what's difference between `addressType` and `isDefaultInterface`? Why require both?

tamilmani1989 (changes requested):

tamilmani1989 (changes requested):

cns/NetworkContainerContract.go
jaer-tsun: yup

rbtr (dismissed):

cns/NetworkContainerContract.go
jaer-tsun: address type makes sense since we're not getting the interface info for default, if secondary we'll get the mac address

tamilmani1989: adressType doesn't make sense to me..tommorow cni has to support IB nic and addressType will be IB for that? nicType makes more sense

jaer-tsun: ok gonna use NICType then

cni/network/invoker.go
```diff
@@ -27,8 +27,17 @@ type IPAMAddConfig struct {
...
-	ipv6Result       *cniTypesCurr.Result
+	// Splitting defaultCniResult from (secondary) cniResults so we don't need to loop for default CNI result every time
+	defaultCniResult CNIResult
+	cniResults       []CNIResult
 	ncResponse       *cns.GetNetworkContainerResponse
 	hostSubnetPrefix net.IPNet
+	ipv6Enabled      bool
+}
+
+type CNIResult struct {
```
neaggarwMS, line 38: InterfaceInfo?

cni/network/invoker.go
```diff
@@ -27,8 +27,17 @@ type IPAMAddConfig struct {
 }
 
 type IPAMAddResult struct {
-	ipv4Result       *cniTypesCurr.Result
-	ipv6Result       *cniTypesCurr.Result
+	// Splitting defaultCniResult from (secondary) cniResults so we don't need to loop for default CNI result every time
+	defaultCniResult CNIResult
+	cniResults       []CNIResult
```
neaggarwMS, line 32: Lets call them: defaultInterfacesInfo and secondaryInterfacesInfo  (@tamilmani)

cni/network/invoker.go
```diff
@@ -27,8 +27,17 @@ type IPAMAddConfig struct {
...
 	ncResponse       *cns.GetNetworkContainerResponse
```
neaggarwMS, line 33: add a comment that this ncResponse will be used in Swift 1.0 multitenancy flow

cns/NetworkContainerContract.go
neaggarwMS: agree with Tamilmani, isDefaultInterface and NicType as Default is confusing. So in Swift 2.0 my NicType = default will be eth0, but my eth1 will isDefaultInterface

cns/NetworkContainerContract.go
```diff
@@ -406,6 +412,14 @@ type PodIpInfo struct {
 	PodIPConfig                     IPSubnet
 	NetworkContainerPrimaryIPConfig IPConfiguration
 	HostPrimaryIPInfo               HostIPInfo
+	// NICType defines whether NIC is Default or Secondary
+	NICType string
+	// MacAddress of (secondary) interface
+	MacAddress string
+	// IsDefaultInterface is true if traffic should go through this interface by default
+	IsDefaultInterface bool
```
neaggarwMS, line 420: Rename it to DefaultEgressRouteRequired bool   // This signifies if this inteface will be used for defaultEgressRoute

neaggarwMS: We also need to check, only 1 interface should have this prop set.

neaggarwMS: discussed with Jareyn: call it SkipDefaultRoutes bool, if set, then honor the routes added in cns.route[] else use the defaults picked by CNI

neaggarwMS: This makes existing logic work without any change and you can special case for multitenant swift 2.0 pods

network/endpoint.go
```diff
@@ -100,6 +103,17 @@ type RouteInfo struct {
...
 
+// InterfaceInfo contains information for secondary interfaces
+type InterfaceInfo struct {
+	Name               string
+	MacAddress         net.HardwareAddr
+	IPAddress          []net.IPNet
+	Gateways           []net.IP
+	Routes             []RouteInfo
+	NICType            string
+	IsDefaultInterface bool
```
neaggarwMS, line 114: change here too

cns/NetworkContainerContract.go
tamilmani1989: can be `defaultGatewayNic` instead of `defaultNic`

nddq: also I think you should make this an enum as well
```
type NICType string

// NIC Types
const (
	NICTypeDefault   NICType = "Default"
	NICTypeSecondary NICType = "Secondary"
)
```

jaer-tsun: renamed to Infra and Secondary NICTypes

cni/network/invoker.go
```diff
@@ -27,8 +27,18 @@ type IPAMAddConfig struct {
 }
 
 type IPAMAddResult struct {
-	ipv4Result       *cniTypesCurr.Result
-	ipv6Result       *cniTypesCurr.Result
+	// Splitting defaultInterfaceInfo from secondaryInterfaceInfo so we don't need to loop for default CNI result every time
```
tamilmani1989, line 30: loop for defaultInterfaceInfo every time?

network/endpoint.go
```diff
@@ -52,6 +52,7 @@ type endpoint struct {
 	PODNameSpace             string `json:",omitempty"`
 	InfraVnetAddressSpace    string `json:",omitempty"`
 	NetNs                    string `json:",omitempty"`
+	SecondaryInterfaces      map[string]*InterfaceInfo
```
tamilmani1989, line 55: can you add a comment indicating what's the key for this map.? I guess interface Name

network/endpoint.go
```diff
@@ -100,6 +103,17 @@ type RouteInfo struct {
 	Table    int
 }
 
+// InterfaceInfo contains information for secondary interfaces
+type InterfaceInfo struct {
+	Name              string
+	MacAddress        net.HardwareAddr
+	IPAddress         []net.IPNet
+	Gateways          []net.IP
```
tamilmani1989, line 111: is Gateways Field being used. If not remove it

cni/network/invoker.go
jaer-tsun: yeah like whenever we refer to defaultInterfaceInfo, we'd have to loop through interfacesInfo (if we don't split) and determine based on NICType or something

network/endpoint.go
jaer-tsun: updated

miguelgoms (approved):

network/endpoint.go
```diff
@@ -86,6 +88,8 @@ type EndpointInfo struct {
 	VnetCidrs                string
 	ServiceCidrs             string
 	NATInfo                  []policy.NATInfo
+	NICType                  string
+	SkipDefaultRoutes        bool
```
behzad-mir, line 92: Can you add a comment about skipDefaultRout and its usage on Swift 2.0?

behzad-mir (approved):
lgtm

cni/network/invoker.go
tamilmani1989: i asked you to update comment from default CNI result to defaultInterfaceInfo

PR #2208, opened 2023-09-08
fix: serviceAccount must be specified at pod spec level in dual stack goldpinger deployments

ramiro-gamarra:
**Reason for Change**:
Goldpinger pods in dual stack tests are not able to query apiserver to discover what pod IPs to ping, due to RBAC being incorrectly set up for these deployments.


pjohnst5 (approved):
LGTM

jpayne3506 (approved):
lgtm

PR #2229, opened 2023-09-20
chore: refactor load test suite

pjohnst5:
**Reason for Change**:
Refactoring the load test suite in preparation for #2187
- Making `env` variables instead of cmd line flags
- Adding a `CLEANUP` option
- Adding `CNS` installation as an option

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -733,7 +733,7 @@ test-load: ## run all load tests
 		go test -timeout 30m -race -tags=load ./test/integration/load...
 
 test-validate-state:
-	cd test/integration/load && OS_TYPE=$(OS) go test -mod=readonly -count=1 -timeout 30m -tags load -run ^TestValidateState
+	cd test/integration/load && go test -mod=readonly -count=1 -timeout 30m -tags load -run ^TestValidateState
 	cd ../../..
 
 test-cyclonus: ## run the cyclonus test for npm.
--- a/Makefile
+++ b/Makefile
@@ -733,7 +733,7 @@ test-load: ## run all load tests
 		go test -timeout 30m -race -tags=load ./test/integration/load...
 
 test-validate-state:
-	cd test/integration/load && go test -mod=readonly -count=1 -timeout 30m -tags load -run ^TestValidateState -restart-case=$(RESTART_CASE) -os=$(OS) -cni=$(CNI_TYPE)
+	cd test/integration/load && OS_TYPE=$(OS) go test -mod=readonly -count=1 -timeout 30m -tags load -run ^TestValidateState
 	cd ../../..
 
 test-cyclonus: ## run the cyclonus test for npm.
--- a/Makefile
+++ b/Makefile
@@ -727,6 +727,11 @@ test-integration: ## run all integration tests.
 		CNS_VERSION=$(CNS_VERSION) \
 		go test -mod=readonly -buildvcs=false -timeout 1h -coverpkg=./... -race -covermode atomic -coverprofile=coverage.out -tags=integration ./test/integration...
 
+test-load: ## run all load tests
+	CNI_DROPGZ_VERSION=$(CNI_DROPGZ_VERSION) \
+		CNS_VERSION=$(CNS_VERSION) \
+		go test -timeout 30m -race -tags=load ./test/integration/load...
+
 test-validate-state:
 	cd test/integration/load && go test -mod=readonly -count=1 -timeout 30m -tags load -run ^TestValidateState -restart-case=$(RESTART_CASE) -os=$(OS) -cni=$(CNI_TYPE)
 	cd ../../..
--- a/test/integration/load/load_test.go
+++ b/test/integration/load/load_test.go
@@ -92,7 +92,7 @@ func TestLoad(t *testing.T) {
 	for i := 0; i < testConfig.Iterations; i++ {
 		t.Log("Iteration ", i)
 		t.Log("Scale down deployment")
-		err = kubernetes.MustScaleDeployment(ctx, deploymentsClient, deployment, clientset, namespace, podLabelSelector, testConfig.Iterations, testConfig.SkipWait)
+		err = kubernetes.MustScaleDeployment(ctx, deploymentsClient, deployment, clientset, namespace, podLabelSelector, testConfig.ScaleDownReplicas, testConfig.SkipWait)
 		require.NoError(t, err)
 
 		t.Log("Scale up deployment")
--- a/test/integration/load/load_test.go
+++ b/test/integration/load/load_test.go
@@ -4,32 +4,35 @@ package load
 
 import (
 	"context"
-	"flag"
 	"testing"
 	"time"
 
 	"github.com/Azure/azure-container-networking/test/internal/kubernetes"
 	"github.com/Azure/azure-container-networking/test/validate"
+	"github.com/stretchr/testify/require"
 )
+
+type TestConfig struct {
+	OSType            string `env:"OS_TYPE" default:"linux"`
+	CNIType           string `env:"CNI_TYPE" default:"cilium"`
+	Iterations        int    `env:"ITERATIONS" default:"2"`
+	ScaleUpReplicas   int    `env:"SCALE_UP" default:"10"`
+	ScaleDownReplicas int    `env:"SCALE_DOWN" default:"1"`
+	Replicas          int    `env:"REPLICAS" default:"1"`
+	ValidateStateFile bool   `env:"VALIDATE_STATEFILE" default:"false"`
+	ValidateDualStack bool   `env:"VALIDATE_DUALSTACK" default:"false"`
+	SkipWait          bool   `env:"SKIP_WAIT" default:"false"`
+	RestartCase       bool   `env:"RESTART_CASE" default:"false"`
+	Cleanup           bool   `env:"CLEANUP" default:"false"`
+}
 
 const (
 	manifestDir      = "../manifests"
 	podLabelSelector = "load-test=true"
+	namespace        = "load-test"
 )
 
-var (
-	osType            = flag.String("os", "linux", "Operating system to run the test on")
-	cniType           = flag.String("cni", "cilium", "CNI to run the test on")
-	iterations        = flag.Int("iterations", 2, "Number of iterations to run the test for")
-	scaleUpReplicas   = flag.Int("scaleup", 10, "Number of replicas to scale up to")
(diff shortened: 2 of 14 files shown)


test/integration/load/setup_test.go
```diff
@@ -0,0 +1,110 @@
...
+			return
+		}
+	} else {
+		log.Printf("Env %v not set to true, skipping", kubernetes.EnvInstallCNS)
+	}
+
+	exitCode = m.Run()
+}
+
+func LoadEnvironment(obj interface{}) {
```
jpayne3506, line 77: +1. Much better than parsing the variables at the various functions. Should alleviate or remove a lot of the bad patterns that came from that.

test/internal/kubernetes/utils_create.go
```diff
@@ -37,62 +40,12 @@ const (
 	envInstallOverlay          = "INSTALL_OVERLAY"
 	envInstallAzureCNIOverlay  = "INSTALL_AZURE_CNI_OVERLAY"
 	envInstallDualStackOverlay = "INSTALL_DUALSTACK_OVERLAY"
```
jpayne3506, line 42: Should we move this to the TestConfig struct?

test/integration/load/setup_test.go
```diff
@@ -0,0 +1,110 @@
...
+		if cnscleanup, err = kubernetes.InstallCNSDaemonset(ctx, clientset, logDir); err != nil {
+			log.Print(err)
+			exitCode = 2
+			return
+		}
+	} else {
+		log.Printf("Env %v not set to true, skipping", kubernetes.EnvInstallCNS)
+	}
+
+	exitCode = m.Run()
```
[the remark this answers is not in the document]
pjohnst5: I don't think so, because it's just a variable in go

test/internal/kubernetes/utils_create.go
pjohnst5: These variables are shared across packages, so I think they should stay here

The TestConfig is only in load

test/integration/load/load_test.go
```diff
@@ -60,142 +63,138 @@ todo: consider adding the following scenarios
...
 	t.Log("Repeating the scale up/down cycle")
-	for i := 0; i < *iterations; i++ {
+	for i := 0; i < testConfig.Iterations; i++ {
 		t.Log("Iteration ", i)
 		t.Log("Scale down deployment")
-		err = kubernetes.MustScaleDeployment(ctx, deploymentsClient, deployment, clientset, namespace, podLabelSelector, *scaleDownReplicas, *skipWait)
-		if err != nil {
-			t.Fatal(err)
-		}
+		err = kubernetes.MustScaleDeployment(ctx, deploymentsClient, deployment, clientset, namespace, podLabelSelector, testConfig.Iterations, testConfig.SkipWait)
```
jpayne3506, line 95: ` testConfig.Iterations,` -> `testConfig.ScaleDownReplicas`

jpayne3506 (changes requested):

test/internal/kubernetes/utils_create.go
```diff
@@ -272,6 +225,158 @@ func InstallCNSDaemonset(ctx context.Context, clientset *kubernetes.Clientset, l
...
+		return appsv1.DaemonSet{}, errors.Wrap(err, "failed to check daemonset running")
+	}
+
+	return cns, nil
+}
+
+func setupCNSDaemonset(ctx context.Context, clientset *kubernetes.Clientset, cns appsv1.DaemonSet, cnsScenarioMap map[string]cnsScenario, flag string) (appsv1.DaemonSet, error) {
+	cnsScenarioConfig, ok := cnsScenarioMap[flag]
+	if !ok {
+		return cns, errors.Wrap(ErrUnsupportedCNSScenario, fmt.Sprintf("%s not a supported cns scneario", flag))
```
vipul-21, line 334: nit: why not use `errors.Wrapf` ?

test/integration/load/load_test.go
```diff
@@ -41,7 +44,7 @@ In order to run the scale tests, you need a k8s cluster and its kubeconfig.
 If no kubeconfig is passed, the test will attempt to find one in the default location for kubectl config.
 Run the tests as follows:
 
-go test -timeout 30m -tags load -run ^TestLoad$ -tags=load
+go test -timeout 30m -tags load -run ^TestLoad$
```
vipul-21, line 47: Since the file still has `build` tag we will need `-tags` right ?

test/integration/load/load_test.go
```diff
@@ -60,142 +63,138 @@ todo: consider adding the following scenarios
...
+		err = kubernetes.MustDeleteDeployment(ctx, deploymentsClient, deployment)
+		require.NoError(t, err, "error deleteing load deployment")
 	}
 }
 
 func TestDualStackProperties(t *testing.T) {
-	clientset, err := kubernetes.MustGetClientset()
-	if err != nil {
-		t.Fatal(err)
+	if !testConfig.ValidateDualStack {
```
vipul-21, line 178: Why do we need this ?

test/integration/load/load_test.go
```diff
@@ -60,142 +63,138 @@ todo: consider adding the following scenarios
...
+	err = kubernetes.MustScaleDeployment(ctx, deploymentsClient, deployment, clientset, namespace, podLabelSelector, testConfig.Replicas, testConfig.SkipWait)
+	require.NoError(t, err)
+
+	if testConfig.Cleanup {
+		err = kubernetes.MustDeleteDeployment(ctx, deploymentsClient, deployment)
+		require.NoError(t, err, "error deleteing load deployment")
 	}
 }
 
 func TestDualStackProperties(t *testing.T) {
```
vipul-21, line 177: Food for thought: The `TestValidate` and this function are doing the same thing apart from this call: `validator.ValidateDualStackControlPlane(ctx)`. Wondering if we can have common code for both ?

test/integration/load/setup_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	clientset, err := kubernetes.MustGetClientset()
+	if err != nil {
+		return
+	}
+
+	ctx := context.Background()
+	installopt := os.Getenv(kubernetes.EnvInstallCNS)
+	// create dirty cns ds
+	if installCNS, err := strconv.ParseBool(installopt); err == nil && installCNS == true {
```
vipul-21, line 64: nit: `err == nil && installCNS {`

test/integration/load/setup_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+		env := fieldTyp.Tag.Get("env")
+		defaultVal := fieldTyp.Tag.Get("default")
+		envVal := os.Getenv(env)
+
+		if envVal == "" {
+			envVal = defaultVal
+		}
+
+		switch fieldVal.Kind() {
```
vipul-21, line 93: should we add a default case to handle unknown types ?

test/integration/load/setup_test.go
```diff
@@ -0,0 +1,110 @@
...
+		if r := recover(); r != nil {
+			log.Println(string(debug.Stack()))
+			exitCode = exitFail
+		}
+
+		if err != nil {
+			log.Print(err)
+			exitCode = exitFail
+		} else {
+			if cnicleanup != nil {
```
vipul-21, line 45: This is for future use case ?

test/validate/validate.go
```diff
@@ -224,3 +224,16 @@ func (v *Validator) ValidateDualStackControlPlane(ctx context.Context) error {
...
+
+func (v *Validator) Cleanup(ctx context.Context) error {
+	// deploy privileged pod
+	privilegedDaemonSet, err := acnk8s.MustParseDaemonSet(privilegedDaemonSetPathMap[v.os])
+	if err != nil {
+		return errors.Wrap(err, "unable to parse daemonset")
+	}
+	daemonsetClient := v.clientset.AppsV1().DaemonSets(privilegedNamespace)
+	if err := acnk8s.MustDeleteDaemonset(ctx, daemonsetClient, privilegedDaemonSet); err != nil {
+		return errors.Wrap(err, "unable to create daemonset")
```
vipul-21, line 236:
```suggestion
		return errors.Wrap(err, "unable to delete daemonset")
```

Makefile
```diff
@@ -727,8 +727,13 @@ test-integration: ## run all integration tests.
 		CNS_VERSION=$(CNS_VERSION) \
 		go test -mod=readonly -buildvcs=false -timeout 1h -coverpkg=./... -race -covermode atomic -coverprofile=coverage.out -tags=integration ./test/integration...
 
+test-load: ## run all load tests
```
vipul-21, line 730: Are we using this make command anywhere ?

pjohnst5: Yeah in the next PR

test/integration/load/setup_test.go
pjohnst5: Yeah it was just there before

test/internal/kubernetes/utils_create.go
pjohnst5: Good idea!

test/integration/load/load_test.go
pjohnst5: Yeah but it's already there, just before , in that same line

test/integration/load/load_test.go
pjohnst5: Ah, because now, with that `make test-load` target, it runs all tests in `load_test.go`
If we don't have this check, it will try to validate dualstack stuff, every time, but that's not the case we want, we only want this test to run if the config is set as a dualstack cluster

test/integration/load/load_test.go
pjohnst5: Just tried that, but, because in this function we are cancelling the `ctx` as a `defer`, it is meant to cancel the `ctx` at the end of the test

If I pull out the common `validator` code, the `ctx` would be cancelled prematurely

So I think we should leave it

test/integration/load/setup_test.go
pjohnst5: Done

test/integration/load/setup_test.go
pjohnst5: Done

test/validate/validate.go
pjohnst5: Done

vipul-21 (approved):

jpayne3506 (approved):

PR #2231, opened 2023-09-20
[CNI] zap logger migration for store package

paulyufan2:
**Reason for Change**:

This PR is to migrate zap logger for store package.
If functions are called by CNI packages, then use zap logger, otherwise use legacy log.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


store/json.go
```diff
@@ -83,7 +89,12 @@ func (kvs *jsonFileStore) Read(key string, value interface{}) error {
 		}
 
 		if len(b) == 0 {
-			log.Printf("Unable to read file %s, was empty", kvs.fileName)
+			if kvs.isZapLogger {
```
tamilmani1989, line 92: can decide based on logger instance nil or not instead of bool

platform/os.go
```diff
@@ -4,8 +4,9 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"log"
 	"os"
+
+	"github.com/Azure/azure-container-networking/log"
```
tamilmani1989, line 9: can you make this change when you do for platform package?

tamilmani1989 (changes requested):

platform/os.go
vipul-21: +1

store/json.go
```diff
@@ -12,12 +12,16 @@ import (
 	"sync"
 	"time"
 
+	zapLog "github.com/Azure/azure-container-networking/cni/log"
```
vipul-21, line 15: We should not need this as well since we are passing the logger during client initialization

store/json.go
```diff
@@ -83,7 +89,12 @@ func (kvs *jsonFileStore) Read(key string, value interface{}) error {
 		}
 
 		if len(b) == 0 {
-			log.Printf("Unable to read file %s, was empty", kvs.fileName)
+			if kvs.logger != nil {
+				logger.Info("Unable to read empty file", zap.String("fileName", kvs.fileName))
```
vipul-21, line 93: This would become, `kvs.logger.Info(..)`. And we remove the variable `logger` from line 23.
This is valid for all log in the file.

store/json_test.go
```diff
@@ -155,12 +155,12 @@ func TestKeyValuePairsAreWrittenAndReadCorrectly(t *testing.T) {
 
 // test case for testing newjsonfilestore idempotent
 func TestNewJsonFileStoreIdempotent(t *testing.T) {
-	_, err := NewJsonFileStore(testLockFileName, processlock.NewMockFileLock(false))
+	_, err := NewJsonFileStore(testLockFileName, processlock.NewMockFileLock(false), nil)
```
vipul-21, line 158: Should be testing the condition where it is not `nil`. In that case, `zap.logger` should be called.

vipul-21 (changes requested):

store/json.go
paulyufan2: fixed

vipul-21 (dismissed):

store/json_test.go
tamilmani1989: @paulyufan2  was this addressed?

paulyufan2: yes it's fixed

tamilmani1989 (approved):

vipul-21 (approved):

PR #1773, opened 2023-01-25
Dualstack CNS Changes

rjdenney:
**Reason for Change**:

This is the CNS side of changes for dualstack in Linux in Windows. Azure-ipam changes will follow once this is merged

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

PR #2239, opened 2023-09-22
Adding dualstack overlay support to azure-ipam plugin for Cilium

rjdenney:
**Reason for Change**:

Adds support in the azure-ipam plugin to use new APIs created for CNS to allow for passing multiple IPs on Add and Delete commands which is needed to create dualstack clusters.

Here's where the new APIs were added.
https://github.com/Azure/azure-container-networking/pull/1773

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -181,7 +181,6 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
 			}
 			err = p.cnsClient.ReleaseIPAddress(context.TODO(), ipconfigReq)
 
-			// if the old API fails as well then we just return the error
 			if err != nil {
 				if errors.As(err, &connectionErr) {
 					p.logger.Info("Failed to release IP address from CNS due to connection failure, saving to watcher to delete")
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -36,6 +36,8 @@ type IPAMPlugin struct {
 
 type cnsClient interface {
 	RequestIPAddress(context.Context, cns.IPConfigRequest) (*cns.IPConfigResponse, error)
+	RequestIPs(context.Context, cns.IPConfigsRequest) (*cns.IPConfigsResponse, error)
+	ReleaseIPs(context.Context, cns.IPConfigsRequest) error
 	ReleaseIPAddress(context.Context, cns.IPConfigRequest) error
 }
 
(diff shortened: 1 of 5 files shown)


cilium/cilium_helm_values.yaml
camrynl, line 1: let's leave this change out since we aren't using helm to install cilium on byocni. we'll probably want to configure conditions later on in the cilium config to set the values for ipv6 since we may not always want to create dualstack clusters

azure-ipam/ipam.go
```diff
@@ -130,18 +161,43 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
...
+	if err := p.cnsClient.ReleaseIPs(context.TODO(), req); err != nil {
+		// if we fail a request with a 404 error try using the old API
+		if cnscli.IsUnsupportedAPI(err) {
+			p.logger.Error("Failed to release IPs using ReleaseIPs from CNS, going to try ReleaseIPAddress", zap.Error(err), zap.Any("request", req))
+			ipconfigReq, err := ipconfig.CreateIPConfigReq(args)
+			if err != nil {
+				p.logger.Error("Failed to create CNS IP config request", zap.Error(err))
+				return cniTypes.NewError(ErrCreateIPConfigRequest, err.Error(), "failed to create CNS IP config request")
+			}
+			err = p.cnsClient.ReleaseIPAddress(context.TODO(), ipconfigReq)
```
vipul-21, line 182: Same comment for logs here as well.

cilium/cilium_helm_values.yaml
camrynl: used to be used for ci, but we moved away from that method of install after adding our own init containers to the cilium daemonset

azure-ipam/ipam.go
```diff
@@ -69,40 +71,69 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
...
+	req, err := ipconfig.CreateIPConfigsReq(args)
 	if err != nil {
-		p.logger.Error("Failed to create CNS IP config request", zap.Error(err))
-		return cniTypes.NewError(ErrCreateIPConfigRequest, err.Error(), "failed to create CNS IP config request")
+		p.logger.Error("Failed to create CNS IP configs request", zap.Error(err))
+		return cniTypes.NewError(ErrCreateIPConfigRequest, err.Error(), "failed to create CNS IP configs request")
 	}
 	p.logger.Debug("Created CNS IP config request", zap.Any("request", req))
 
 	p.logger.Debug("Making request to CNS")
```
tamilmani1989, line 81: these logs should be moved where request is made

azure-ipam/ipam.go
```diff
@@ -130,18 +161,43 @@ func (p *IPAMPlugin) CmdDel(args *cniSkel.CmdArgs) error {
...
+
+			if err != nil {
+				if errors.As(err, &connectionErr) {
+					p.logger.Info("Failed to release IP address from CNS due to connection failure, saving to watcher to delete")
+					addErr := fsnotify.AddFile(args.ContainerID, args.ContainerID, watcherPath)
+					if addErr != nil {
+						p.logger.Error("Failed to add file to watcher", zap.String("containerID", args.ContainerID), zap.Error(addErr))
+						return cniTypes.NewError(cniTypes.ErrTryAgainLater, addErr.Error(), fmt.Sprintf("failed to add file to watcher with containerID %s", args.ContainerID))
+					} else {
+						p.logger.Info("File successfully added to watcher directory")
```
tamilmani1989, line 192: can we make this change as separate PR?

tamilmani1989 (changes requested):

tamilmani1989:
@rjdenney i dont see unit tests added.. can you add for new changes you added?

cilium/cilium_helm_values.yaml
rjdenney: removed the helm values

camrynl (approved):

vipul-21 (approved):

azure-ipam/ipam_test.go
```diff
@@ -227,15 +385,25 @@ func TestCmdDel(t *testing.T) {
...
 		},
 		{
-			name:    "Fail request CNS release IP during CmdDel",
+			name:    "Fail request CNS release using ReleaseIPAddress during CmdDel",
 			args:    buildArgs("failRequestCNSReleaseIPArgs", happyPodArgs, happyNetConfByteArr),
 			wantErr: true,
 		},
+		{
+			name:    "Fail request CNS release during CmdDel",
+			args:    buildArgs("failRequestCNSReleaseIPsArgs", happyPodArgs, happyNetConfByteArr),
```
tamilmani1989, line 404: this case should validate if azure-ipam calls ReleaseIPAddress api if ReleaseIPs fails

tamilmani1989:
@rjdenney @camrynl should have test case to validate if azure-cns returns "unsupported api" error, cns should call old api

azure-ipam/ipam.go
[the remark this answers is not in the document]
rjdenney: Here's where the new APIs were added.
https://github.com/Azure/azure-container-networking/pull/1773

azure-ipam/ipam_test.go
rjdenney: Right now in the code the only time that we call `ReleaseIPAddress` is if we get a 404 error when we call `ReleaseIPs`. 

With the four tests that exist now the first happy path uses  `ReleaseIPs` and works fine since it is the happy path, second one first calls `ReleaseIPs`, fails with a 404 error, and then calls `ReleaseIPAddress` which passes, third fails does the same as second but fails at `ReleaseIPAddress`, and the fourth fails `ReleaseIPs` with an error that is not 404 so we don't call `ReleaseIPAddress`.

azure-ipam/ipam_test.go
```diff
@@ -85,6 +88,114 @@ func (c *MockCNSClient) RequestIPAddress(ctx context.Context, ipconfig cns.IPCon
...
 
+func (c *MockCNSClient) RequestIPs(ctx context.Context, ipconfig cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
+	switch ipconfig.InfraContainerID {
+	case "failRequestCNSArgs":
+		return nil, errFoo
+	case "happyArgsSingle", "failProcessCNSRespSingleIP", "failRequestCNSArgsSingleIP":
+		e := &client.CNSClientError{}
+		e.Code = types.UnsupportedAPI
+		e.Err = errUnsupportedAPI
+		return nil, e
```
rjdenney, line 99: @tamilmani1989 For `RequestIPs` each of these tests will return back the UnsupportedAPI when we call `RequestIPs` which then causes us to test `RequestIPAddress`

tamilmani1989 (approved):
lgtm

PR #2241, opened 2023-09-22
ci: Add CNIv1 Linux to Load Test Pipeline

jpayne3506:
**Reason for Change**:

Adds Linux CNIv1 to load test pipeline to test additional scenarios.
Refactors windows and linux CNIv1 into a single template to ensure that the testing coverage matches where applicable.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/cni/pipeline.yaml
+++ b/.pipelines/cni/pipeline.yaml
@@ -70,8 +70,6 @@ stages:
             artifact: npm_k8s
       - job:
         displayName: "Build k8s NPM Windows Test Suite Binary"
-        pool:
-          name: $(BUILD_POOL_NAME_DEFAULT)
         steps:
           - bash: |
               git clone https://github.com/huntergregory/kubernetes.git --depth=1 --branch=sleep-before-probing
--- a/.pipelines/cni/pipeline.yaml
+++ b/.pipelines/cni/pipeline.yaml
@@ -118,11 +118,12 @@ stages:
       clusterName: win22-cniv1
       nodeCount: ${WINDOWS_NODE_COUNT}
       vmSize: Standard_DS4_v2
-      windowsVMSize: ${WINDOWS_VM_SKU}
+      windowsVMSize: Standard_DS4_v2
+      os: windows
+      os_version: 'ltsc2022'
       arch: amd64
       scaleup: ${WINDOWS_SCALEUP}
       iterations: ${WINDOWS_ITERATIONS}
-      os_version: 'ltsc2022'
 
   - template: singletenancy/cniv1-template.yaml
     parameters:
@@ -131,6 +132,7 @@ stages:
       clusterName: linux-cniv1
       nodeCount: ${LINUX_NODE_COUNT}
       vmSize: Standard_DS4_v2
+      os: linux
       arch: amd64
       scaleup: ${LINUX_SCALEUP}
       iterations: ${LINUX_ITERATIONS}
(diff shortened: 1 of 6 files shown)


.pipelines/cni/pipeline.yaml
```diff
@@ -70,8 +70,6 @@ stages:
             artifact: npm_k8s
       - job:
         displayName: "Build k8s NPM Windows Test Suite Binary"
-        pool:
```
vipul-21, line 73: why are we removing the pool name ?

.pipelines/cni/pipeline.yaml
```diff
@@ -111,16 +109,31 @@ stages:
...
-      clusterType: "windows-cniv1-up"
-      clusterName: "win-cniv1"
+      name: win22_cniv1
+      clusterType: windows-cniv1-up
+      clusterName: win22-cniv1
       nodeCount: ${WINDOWS_NODE_COUNT}
       vmSize: Standard_DS4_v2
-      windowsVMSize: ${WINDOWS_VM_SKU}
-      arch: amd64
+      windowsVMSize: Standard_DS4_v2
```
vipul-21, line 119: Can we parameterize this ? 
As windows scale up is also parameterized we sometimes need to upgrade the vm size based on that

test/validate/linux_validate.go
```diff
@@ -18,6 +18,7 @@ var (
 	restartNetworkCmd      = []string{"bash", "-c", "chroot /host /bin/bash -c systemctl restart systemd-networkd"}
 	cnsManagedStateFileCmd = []string{"bash", "-c", "cat /var/run/azure-cns/azure-endpoints.json"}
 	azureVnetStateFileCmd  = []string{"bash", "-c", "cat /var/run/azure-vnet.json"}
+	azureVnetStateIpamCmd  = []string{"bash", "-c", "cat /var/run/azure-vnet-ipam.json"}
```
vipul-21, line 21: nit: `azureVnetIpamStateCmd`

.pipelines/cni/pipeline.yaml
jpayne3506: Because it is set at the stage level. This assigns every job below it the same pool

test/validate/linux_validate.go
jpayne3506: Also corrected the function name

.pipelines/cni/pipeline.yaml
jpayne3506: parameterized all related fields.

vipul-21 (approved):

hack/aks/Makefile
```diff
@@ -256,6 +248,17 @@ dualstack-overlay-byocni-up: rg-up overlay-net-up ## Brings up an dualstack Over
...
 	@$(MAKE) set-kubeconf
 
+windows-nodepool-up: ## Add windows node pool
+	$(AZCLI) aks nodepool add -g $(GROUP) -n npwin \
+		--node-count $(NODE_COUNT) \
+		--node-vm-size $(VM_SIZE) \
+		--cluster-name  $(CLUSTER) \
+		--os-type Windows \
+		--os-sku $(WINDOWS_OS_SKU) \
+		--max-pods 250 \
```
camrynl, line 258: should we set a var for max pods?

jpayne3506: Currently we do not test anything against the max pods on a node. This is set because the default was 30 and we needed to be able to do scale tests ect. Until we test against it I think it should stay at the maximum, so it matches what the other cluster commands default to.

camrynl (approved):

PR #2246, opened 2023-09-23
feat: installing windows cns if windows nodes present

pjohnst5:
**Reason for Change**:
Adding the capability to install windows cns if windows nodes present, and v4overlay

In preparation for #2187


test/internal/kubernetes/utils_create.go
```diff
@@ -209,157 +223,248 @@ func InstallCNSDaemonset(ctx context.Context, clientset *kubernetes.Clientset, l
...
-		if err := ExportLogsByLabelSelector(ctx, clientset, cns.Namespace, cnsLabelSelector, logDir); err != nil {
-			return errors.Wrapf(err, "failed to export logs by label selector %s", cnsLabelSelector)
+		if err := ExportLogsByLabelSelector(ctx, clientset, cnsLinux.Namespace, cnsLinuxDetails.labelSelector, logDir); err != nil {
+			return errors.Wrapf(err, "failed to export linux logs by label selector %s", cnsLinuxLabelSelector)
+		}
+		if hasWinNodes {
+			if err := ExportLogsByLabelSelector(ctx, clientset, cnsWindows.Namespace, cnsWindowsDetails.labelSelector, logDir); err != nil {
+				return errors.Wrapf(err, "failed to export windows logs by label selector %s", cnsWindowsLabelSelector)
+			}
 		}
```
jpayne3506, line 252: If I remember correctly, this will get used by a cluster that has both linux and windows CNS installed. Should we try to wrap it in both windows and linux logs or do and if/else statement? As it stands now there is a possibility that the `hasWinNodes` check will be skipped as it will return before it reaches it.

test/integration/manifests/cns/daemonset-windows.yaml
```diff
@@ -0,0 +1,134 @@
...
+            path: /k/azurecns
+            type: DirectoryOrCreate
+        - name: cns-config
+          configMap:
+            name: cns-win-config
+            defaultMode: 420
+        - name: cni-bin
+          hostPath:
+            path: /k/azurecni/bin
+            type: Directory # // TODO: add windows cni conflist when ready
```
jpayne3506, line 132: Is the conflist ready?

test/internal/kubernetes/utils_create.go
```diff
@@ -30,21 +47,18 @@ type cnsScenario struct {
...
+	envCNSVersion           = "CNS_VERSION"
+	EnvInstallCNS           = "INSTALL_CNS"
+	cnsLinuxLabelSelector   = "k8s-app=azure-cns"
+	cnsWindowsLabelSelector = "k8s-app=azure-cns-win"
 )
 
 var (
 	ErrUnsupportedCNSScenario = errors.New("unsupported CNS scenario")
 	ErrPathNotFound           = errors.New("failed to get the absolute path to directory")
+	ErrNoCNSScenarioDefined   = errors.New("no CNSScenario set to true as env var")
```
jpayne3506, line 61: Should this be changed to `CNSScenario is not defined`? Would read out `no CNSSCenario env vars set to true, must explicitly set one to true: CNSScenario is not defined` instead of `no CNSSCenario env vars set to true, must explicitly set one to true: no CNSScenario set to true as env var`

test/internal/kubernetes/utils_create.go
pjohnst5: > If I remember correctly, this will get used by a cluster that has both linux and windows CNS installed.
Correct

That's a good point, if something fails to get the linux logs, then we'd skip the windows logs all together

test/integration/manifests/cns/daemonset-windows.yaml
```diff
@@ -0,0 +1,134 @@
...
+            - name: CNS_CONFIGURATION_PATH
+              value: /etc/azure-cns/cns_config.json
+            - name: NODENAME
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: spec.nodeName
+      initContainers:
+        - name: init-cni-dropgz
+          image: acnpublic.azurecr.io/cni-dropgz:latest
```
vipul-21, line 108: Do we publish the `latest` tag for cni-dropgz ?

test/internal/kubernetes/utils_create.go
pjohnst5: Fixed

test/integration/manifests/cns/daemonset-windows.yaml
pjohnst5: No we don't, but this is just a placeholder, we update it in code through `CNI_DROPGZ_VERSION`
I'll update this here too to reflect that

vipul-21 (approved):

paulyufan2 (approved):

pjohnst5:
@microsoft-github-policy-service agree

pjohnst5:
@microsoft-github-policy-service agree

pjohnst5:
@microsoft-github-policy-service rerun

PR #2253, opened 2023-09-26
fix: set cluster auto-upgrade and node auto-upgrade in the hackfile

rbtr:
Since we occasionally generate long-lived clusters using the hackfile (either intentionally, or leaking them accidentally out of the pipelines), we can at least reduce any ongoing security scan violations by enabling all possible automatic upgrades on these clusters by default.
 

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] relevant PR labels added

**Notes**:

PR #2258, opened 2023-09-27
fix: cluster template env vars

jpayne3506:
**Reason for Change**:

**Issue Fixed**:

Fixes #2253 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


.pipelines/cni/pipeline.yaml
```diff
@@ -107,16 +107,16 @@ stages:
       clusterType: overlay-byocni-nokubeproxy-up
       clusterName: "cilium-overlay"
       nodeCount: ${CILIUM_NODE_COUNT}
-      vmSize: ${CILIUM_VM_SKU}
+      vmSize: ${CILIUM_VM_SIZE}
```
rbtr, line 110:
```suggestion
      vmSize: ${VM_SIZE_CILIUM}

```
it should follow the least-to-most-specific ordering convention so that the lexicographic sort will group variants of the same arg together

.pipelines/cni/pipeline.yaml
```diff
@@ -107,16 +107,16 @@ stages:
...
   - template: singletenancy/cniv1-template.yaml
     parameters:
       name: win22_cniv1
       clusterType: windows-cniv1-up
       clusterName: win22-cniv1
       nodeCount: ${WINDOWS_NODE_COUNT}
-      vmSize: ${WINDOWS_LINUX_VM_SKU}
-      windowsVMSize: ${WINDOWS_VM_SKU}
+      vmSize: ${WINDOWS_LINUX_VM_SIZE}
+      vmSizeWin: ${WINDOWS_VM_SIZE}
```
rbtr, line 119:
```suggestion
      vmSize: ${VM_SIZE}
      vmSizeWin: ${VM_SIZE_WIN}

```

.pipelines/cni/pipeline.yaml
rbtr, line 110:
```suggestion
      vmSize: ${VM_SIZE_CILIUM}
```
it should follow the least-to-most-specific ordering convention so that the lexicographic sort will group variants of the same arg together

.pipelines/cni/pipeline.yaml
```diff
@@ -129,7 +129,7 @@ stages:
       clusterType: linux-cniv1-up
       clusterName: linux-cniv1
       nodeCount: ${LINUX_NODE_COUNT}
-      vmSize: ${LINUX_VM_SKU}
+      vmSize: ${LINUX_VM_SIZE}
```
rbtr, line 132:
```suggestion
      vmSize: ${VM_SIZE_LINUX}

```

.pipelines/cni/pipeline.yaml
```diff
@@ -141,7 +141,7 @@ stages:
       clusterType: swift-byocni-up
       clusterName: linux-cniv2-podsubnet
       nodeCount: ${LINUX_NODE_COUNT}
-      vmSize: ${LINUX_VM_SKU}
+      vmSize: ${LINUX_VM_SIZE}
```
rbtr, line 144:
```suggestion
      vmSize: ${VM_SIZE_LINUX}

```

.pipelines/cni/pipeline.yaml
```diff
@@ -150,7 +150,7 @@ stages:
       clusterType: overlay-byocni-up
       clusterName: linux-cniv2-overlay
       nodeCount: ${LINUX_NODE_COUNT}
-      vmSize: ${LINUX_VM_SKU}
+      vmSize: ${LINUX_VM_SIZE}
```
rbtr, line 153:
```suggestion
      vmSize: ${VM_SIZE_LINUX}

```

rbtr (changes requested):

.pipelines/cni/pipeline.yaml
```diff
@@ -141,7 +141,7 @@ stages:
       clusterType: swift-byocni-up
       clusterName: linux-cniv2-podsubnet
       nodeCount: ${LINUX_NODE_COUNT}
```
vipul-21, line 143: We should change these as well ?

.pipelines/cni/pipeline.yaml
jpayne3506: For windows clusters we need a 4th var that allows us to control the linux node vmSize independently of the linux clusters, Looking at `VM_SIZE_WIN_LINUX` as a compromise.

rbtr: why can't we use the same VM size as we use in linux clusters?

jpayne3506: Its not that we can't, but when testing scale with higher node counts having a VM with 16 cores assigned to 2 nodes is overkill. Can't see an instance of the inverse being negative..

.pipelines/cni/pipeline.yaml
jpayne3506: We can, originally it was setup this way to ensure that all related variables for each cluster were sorted next to each other. With the changes above it is starting to break away for that pattern. Would rather have a singular pattern versus 2. Will change to match.

.pipelines/cni/pipeline.yaml
jpayne3506: Agreed upon `VM_SIZE_WINCLUSTER_SYSTEMPOOL` with @rbtr offline to clearly signify the purpose of the env var

rbtr (dismissed):

vipul-21 (dismissed):

rbtr (approved):

vipul-21 (approved):

PR #2268, opened 2023-10-02
chore: must* functions should panic

pjohnst5:
`must*` functions should panic, not return `err`

Targeting the `test/internal/kubernetes` package in this PR

### DIFF
--- a/test/integration/datapath/datapath_linux_test.go
+++ b/test/integration/datapath/datapath_linux_test.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"flag"
 	"net"
-	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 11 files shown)


ecigar13 (approved):

rbtr (dismissed):

pjohnst5:
These CNI Load test failures don't block, they are a different testing pipeline that John and the Cilium folks use to test CNI load tests with Cilium

Right now, I see errors about a deployment not being able to be modified because it has been altered (likely due to a duplicate run or re-run of the pipeline since I've been trying over and over), and another about quota being too low in the sub

I think we can write that off as unrelated, we actually run the same code now in the ACN PR pipeline anyways, so we know those code paths are being hit

vipul-21 (approved):

PR #2271, opened 2023-10-03
Fix missing Wireserver IP configuration in Wireserver Client

rbtr:
Previously, the Wireserver Client used a hardcoded IP, which made starting CNS locally impossible. This is because it attempted to make requests to a Wireserver that wasn't there. There is already a configuration for a WireserverIP, with the intention that that be used to redirect those requests to a mock server for local testing.

This makes the host to which those Wireserver requests are directed configurable so that this configuration is respected by the Wireserver client.

No configuration changes are necessary to end users since no configuration has changed.

To use this locally, any mocks that are in use will need to add an endpoint responding to `/machine/plugins?comp=nmagent&getinterfaceinfov1`  responding with XML in the form:

```xml
<Result>
  <Interfaces MacAddress="{{mac}}" IsPrimary="true">
    <IPSubnet Prefix="{{prefix}}">
      <IPAddress Address="{{addr}}" IsPrimary="true"></IPAddress>
    </IPSubnet>
  </Interfaces>
</Result>
```

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -678,6 +678,7 @@ func main() {
 	}
 
 	wsclient := &wireserver.Client{
+		HostPort:   cnsconfig.WireserverIP,
 		HTTPClient: &http.Client{},
 		Logger:     logger.Log,
 	}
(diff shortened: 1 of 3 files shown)


nddq (approved):
lgtm

timraymond (approved):

PR #2272, opened 2023-10-03
ci: Agent pool default set at stage level

jpayne3506:
**Reason for Change**:

Ensures that E2E templates in PR pipeline use `Networking-ContainerNetworking-VMSS` build pool by setting it at the stage level, which is passed to the individual jobs and can be overwritten. This is to prevent using the default build pool that can call old images that no longer exist or are depreciated. 

```
##[warning]An image label with the label Ubuntu16 does not exist.
,##[error]The remote provider was unable to process the request.
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
+++ b/.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
@@ -41,6 +41,8 @@ stages:
       CURRENT_VERSION: $[ stagedependencies.containerize.check_tag.outputs['CurrentTagManifests.currentTagManifests'] ]
       commitID: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.commitID'] ]
     condition: and(succeeded(), eq(variables.TAG, variables.CURRENT_VERSION))
+    pool:
+      name: $(BUILD_POOL_NAME_DEFAULT)
     jobs:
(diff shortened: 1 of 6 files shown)


.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
```diff
@@ -41,6 +41,8 @@ stages:
       CURRENT_VERSION: $[ stagedependencies.containerize.check_tag.outputs['CurrentTagManifests.currentTagManifests'] ]
       commitID: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.commitID'] ]
     condition: and(succeeded(), eq(variables.TAG, variables.CURRENT_VERSION))
+    pool:
```
vipul-21, line 44: So are these getting changed for stages even if the job has a default specified ?

jpayne3506: If the job level specifies a pool then the stage level gets overwritten.

rbtr (approved):

matmerr (approved):

vipul-21 (approved):

PR #2276, opened 2023-10-05
feat: Dev stateless cni

behzad-mir:
This PR includes all changes related to Stateless CNI in which CNS will manage and maintain the endpoint states.

- There are changes related to CNS to add two new APIs with respect of endpoint state management.
GetEndpoint() is used to return the state of a given endpointId. UpdateEndpoint() is used to update the HNSId and VethName field within the endpoint state.
- The other changes are related to CNI and network package so the ADD/Del calls can be done without the use of statefile on Store package and instead use CNS and aformenetioned APIs.

- [X] includes documentation
- [X] adds unit tests

**Notes**:
[Stateless Azure cni Design Doc- P1.docx](https://github.com/Azure/azure-container-networking/files/13620367/Stateless.Azure.cni.Design.Doc-.P1.docx)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@ ZAPAI_VERSION			?= $(notdir $(shell git describe --match "zapai*" --tags --alway
 AZURE_IPAM_DIR = $(REPO_ROOT)/azure-ipam
 CNM_DIR = $(REPO_ROOT)/cnm/plugin
 CNI_NET_DIR = $(REPO_ROOT)/cni/network/plugin
+STATELESS_CNI_NET_DIR = $(REPO_ROOT)/cni/network/stateless
 CNI_IPAM_DIR = $(REPO_ROOT)/cni/ipam/plugin
 CNI_IPAMV6_DIR = $(REPO_ROOT)/cni/ipam/pluginv6
 CNI_TELEMETRY_DIR = $(REPO_ROOT)/cni/telemetry/service
@@ -58,11 +59,13 @@ AZURE_IPAM_BUILD_DIR = $(BUILD_DIR)/azure-ipam
 IMAGE_DIR  = $(OUTPUT_DIR)/images
 CNM_BUILD_DIR = $(BUILD_DIR)/cnm
 CNI_BUILD_DIR = $(BUILD_DIR)/cni
+STATELESS_CNI_BUILD_DIR = $(CNI_BUILD_DIR)/stateless
 ACNCLI_BUILD_DIR = $(BUILD_DIR)/acncli
 CNI_MULTITENANCY_BUILD_DIR = $(BUILD_DIR)/cni-multitenancy
 CNI_MULTITENANCY_TRANSPARENT_VLAN_BUILD_DIR = $(BUILD_DIR)/cni-multitenancy-transparent-vlan
 CNI_SWIFT_BUILD_DIR = $(BUILD_DIR)/cni-swift
 CNI_OVERLAY_BUILD_DIR = $(BUILD_DIR)/cni-overlay
+STATELESS_CNI_OVERLAY_BUILD_DIR = $(CNI_OVERLAY_BUILD_DIR)/stateless
 CNI_BAREMETAL_BUILD_DIR = $(BUILD_DIR)/cni-baremetal
 CNI_DUALSTACK_BUILD_DIR = $(BUILD_DIR)/cni-dualstack
 CNS_BUILD_DIR = $(BUILD_DIR)/cns
@@ -130,7 +133,7 @@ endif
 
 # Shorthand target names for convenience.
 azure-cnm-plugin: cnm-binary cnm-archive
-azure-cni-plugin: azure-vnet-binary azure-vnet-ipam-binary azure-vnet-ipamv6-binary azure-vnet-telemetry-binary cni-archive
+azure-cni-plugin: azure-vnet-binary azure-vnet-stateless-binary azure-vnet-ipam-binary azure-vnet-ipamv6-binary azure-vnet-telemetry-binary cni-archive
 azure-cns: azure-cns-binary cns-archive
 acncli: acncli-binary acncli-archive
 azure-cnms: azure-cnms-binary cnms-archive
@@ -179,6 +182,10 @@ cnm-binary:
 # Build the Azure CNI network binary.
 azure-vnet-binary:
 	cd $(CNI_NET_DIR) && CGO_ENABLED=0 go build -v -o $(CNI_BUILD_DIR)/azure-vnet$(EXE_EXT) -ldflags "-X main.version=$(CNI_VERSION)" -gcflags="-dwarflocationlists=true"
+
+# Build the Azure CNI stateless network binary
+azure-vnet-stateless-binary:
+	cd $(STATELESS_CNI_NET_DIR) && CGO_ENABLED=0 go build -v -o $(STATELESS_CNI_BUILD_DIR)/azure-vnet$(EXE_EXT) -ldflags "-X main.version=$(CNI_VERSION)" -gcflags="-dwarflocationlists=true"
 
 # Build the Azure CNI IPAM binary.
 azure-vnet-ipam-binary:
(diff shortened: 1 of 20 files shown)


cni/network/network.go
```diff
@@ -1054,12 +1054,15 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 				// Log the error but return success if the network is not found.
 				// if cni hits this, mostly state file would be missing and it can be reboot scenario where
 				// container runtime tries to delete and create pods which existed before reboot.
+				// this condition will not apply to stateless CNI since the network struct will be crated on each call
 				err = nil
-				return err
+				if !plugin.nm.IsStatelessCNIMode() {
```
jaer-tsun, line 1059: do we need this check if it doesn't apply to stateless cni?

cni/network/stateless/main.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/Azure/azure-container-networking/platform"
+	"github.com/Azure/azure-container-networking/telemetry"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+)
+
+var logger = zapLog.CNILogger.With(zap.String("component", "cni-main"))
+
+const (
+	hostNetAgentURL = "http://168.63.129.16/machine/plugins?comp=netagent&type=cnireport"
```
jaer-tsun, line 26: some of these look common across regular cni

network/manager.go
```diff
@@ -382,11 +442,67 @@ func (nm *networkManager) DeleteEndpoint(networkID, endpointID string) error {
...
+		LocalIP:                  "",
+		VlanID:                   0,
+		AllowInboundFromHostToNC: false,
+		AllowInboundFromNCToHost: false,
+		EnableSnatOnHost:         false,
+		EnableMultitenancy:       false,
+		NetworkContainerID:       epInfo.Id,
+	}
+	logger.Info("Deleting endpoint with", zap.String("Endpoint Info: ", epInfo.PrettyString()), zap.String("HNISID : ", ep.HnsId))
+	return nw.deleteEndpointImpl(netlink.NewNetlink(), platform.NewExecClient(logger), nil, nil, ep)
```
jaer-tsun, line 469: we'd need to pass endpoint with SecondaryInterfaceInfo for deletion as well

cns/restserver/ipam.go
```diff
@@ -942,3 +946,171 @@ func validateDesiredIPAddresses(desiredIPs []string) error {
...
+		},
+		EndpointInfo: *endpointInfo,
+	}
+	w.Header().Set(cnsReturnCode, response.Response.ReturnCode.String())
+	err = service.Listener.Encode(w, &response)
+	logger.Response(service.Name, response, response.Response.ReturnCode, err)
+}
+
+// GetEndpointHelper returns the state of the given endpointId
+func (service *HTTPRestService) GetEndpointHelper(endpointID string) (*EndpointInfo, error) {
```
jaer-tsun, line 1014: we'd have to return endpoint struct which has the SecondaryInterfaces info or maybe slice of EndpointInfo

network/manager.go
behzad-mir: This is only going to cover stateless CNI for single Tenancy

behzad-mir: We are not covering Swift 2.0 at this point

cns/restserver/ipam.go
behzad-mir: This is the EndpointInfo of the restserver.go and does not have any secondary endpointInfo. The struct get filled out during the IP assignment from CNS

cni/network/network.go
behzad-mir: I added a comment on top of it :
// this condition will not apply to stateless CNI since the network struct will be crated on each call

cni/network/stateless/main.go
behzad-mir: Will be corrected on next commit

cns/restserver/ipam.go
tamilmani1989: @behzad-mir will stateless cni work for swift2.0?

cns/service/main.go
```diff
@@ -658,13 +659,18 @@ func main() {
 		}
 		defer endpointStoreLock.Unlock() // nolint
 
-		err = platform.CreateDirectory(endpointStoreLocation)
+		endpointStorePath := endpointStoreLocation
+		if runtime.GOOS == "windows" {
+			endpointStorePath = os.Getenv("CNSStoreFilePath")
+		}
+		err = platform.CreateDirectory(endpointStorePath)
```
tamilmani1989, line 666: why do we need this? who sets `CNSStoreFilePath`

behzad-mir: It's from the CNS configMap written by AKS-RP. We need it since the path for Windows is different than Linux

cns/restserver/ipam.go
behzad-mir: I think swift 2.0 is for multi tenancy right? We're only targetting single tenancy

network/manager.go
tamilmani1989: if we address swift 2.0, then we dont need to care about migration for swift2.0 later.. Not in this PR, probably a different PR

cns/service/main.go
tamilmani1989: if cns configmap, why its getting from env? shouldn't this be part of cnsconfig field?

behzad-mir: This path is not within CNS CM. We have to read it from DS. It seems this is the way to read from CNS DS confing.

behzad-mir: This variable basically getting passed as an arg on CNS DS.

tamilmani1989 (approved):

rbtr (changes requested):
few minor comments. it would be easier to review if we scoped the changes to smaller PRs container only CNS and only CNI

cns/service/main.go
```diff
@@ -81,7 +81,8 @@ const (
 	name                              = "azure-cns"
 	pluginName                        = "azure-vnet"
 	endpointStoreName                 = "azure-endpoints"
-	endpointStoreLocation             = "/var/run/azure-cns/"
+	endpointStoreLocationLinux        = "/var/run/azure-cns/"
+	endpointStoreLocationWindows      = "/k/azurecns/"
```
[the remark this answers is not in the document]
behzad-mir: I used that path at first but store package showed error and couldn't create the endpoint file.

Makefile
```diff
@@ -180,6 +184,10 @@ cnm-binary:
 azure-vnet-binary:
 	cd $(CNI_NET_DIR) && CGO_ENABLED=0 go build -v -o $(CNI_BUILD_DIR)/azure-vnet$(EXE_EXT) -ldflags "-X main.version=$(CNI_VERSION)" -gcflags="-dwarflocationlists=true"
 
+# Build the Azure CNI stateless network binary
+azure-vnet-stateless-binary:
+	cd $(STATELESS_CNI_NET_DIR) && CGO_ENABLED=0 go build -v -o $(STATELESS_CNI_BUILD_DIR)/azure-vnet$(EXE_EXT) -ldflags "-X main.version=$(CNI_VERSION)" -gcflags="-dwarflocationlists=true"
```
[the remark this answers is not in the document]
behzad-mir: Removed the makefile changes

cns/service/main.go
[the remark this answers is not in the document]
rbtr: then it needs to come from config instead of being hardcoded. we could have `config_linux.go` and `config_windows.go` to populate defaults if it is not set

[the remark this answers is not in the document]
behzad-mir: We also already made the AgentBaker changes according to k/azurecns directory since it was provided int he CNS Configmap for Windows. I simply applied the value form the configmap

tamilmani1989 (approved):

rbtr (approved):
unblocking but let's revisit the windows/linux state file path (and how it comes from config). the hardcoded paths can't stay

tamilmani1989 (approved):

PR #2277, opened 2023-10-05
ci: Add control through environment variables for CNI Load Test 

jpayne3506:
**Reason for Change**:

Provides cluster creation and test scenario control through the `CONTROL_OS, CONTROL_CNI, and CONTROL_SCENARIO` env vars. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Default value is `all` for `CONTROL_OS, CONTROL_CNI, and CONTROL_SCENARIO`. Specific variables for the environment variables are as listed:
`CONTROL_OS` : windows linux
`CONTROL_CNI` : cniv1 cniv2 cilium
`CONTROL_SCENARIO` : npm scaleTest restartNode restartCNS

I.E.
`CONTROL_OS = windows `
`CONTROL_CNI = all`
`CONTROL_SCENARIO = scaleTest`

### DIFF
--- a/.pipelines/cni/singletenancy/linux-cniv2-template.yaml
+++ b/.pipelines/cni/singletenancy/linux-cniv2-template.yaml
@@ -16,10 +16,6 @@ parameters:
 # + Pods should have ips assigned and connectivity/datapath test should be present.
 # + CNS restart and validates the state
 
-#  and( succeeded(),  or( and( contains(variables.RUN, 'cniv2'), contains(variables.RUN, ${{ parameters.os }} ) , contains(variables.RUN, 'all') ) ) )
-#  and( succeeded(),  or(  ...   , contains(variables.RUN, 'all') )
-#   and( contains(variables.RUN, 'cniv2'),
-#  contains(variables.RUN, ${{ parameters.os }} )
 stages:
   - ${{ if eq(parameters.os, 'linux') }}:
     - stage: create_${{ parameters.name }}
@@ -120,6 +116,7 @@ stages:
           dependsOn: integration
           continueOnError: true
       - job: deploy_pods
+        condition: and( and( not(canceled()), not(failed()) ), or( contains(variables.SCENARIO, 'scaleTest') , contains(variables.SCENARIO, 'all') ) )
         displayName: "Scale Test"
         dependsOn: npm_k8se2e
         steps:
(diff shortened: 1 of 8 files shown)


.pipelines/cni/singletenancy/linux-cniv2-template.yaml
```diff
@@ -17,25 +17,48 @@ parameters:
...
-            parameters:
-              clusterType: ${{ parameters.clusterType }}
-              clusterName: ${{ parameters.clusterName }}-$(commitID)
-              nodeCount: ${{ parameters.nodeCount }}
-              vmSize: ${{ parameters.vmSize }}
-              region: $(LOCATION)
-              osSKU: ${{ parameters.osSKU }}
+  - ${{ if eq(parameters.os, 'linux') }}:
+    - stage: create_${{ parameters.name }}
+      condition: and( succeeded(), or( and( contains(variables.CONTROL_CNI, 'cniv2'), contains(variables.CONTROL_OS, 'linux') ) , contains(variables.CONTROL_CNI, 'all') ) )
```
vipul-21, line 22: In case of `CONTROL_CNI` as all and CONTROL_OS as `windows` the above state is true ? Where as my intentions is to test all CNI for windows. 
Or the line 20 won't be triggered somehow ? Trying to understand this.

jpayne3506: Yes, it would be true due to me missing an interaction/check that I want down below. It will be fixed.

Line 20 intent is to control expected comparisons based on the template call (subsequently the same for line 41) so that we can match up the `CONTROL_OS` calls with `parameter.os`. This is because we cannot pass in `parameter.xxx` because they are compile time variables versus the runtime variables of `variable.xxx`. Ideally this would all be on one line and one stage block, but c'est la vie. 

As it is now, if `CONTROL_CNI = all` then all clusters will be created, which was not intended. Correct logic would be this 
`and( succeeded(),`
`and( `
`   or( contains(variables.CONTROL_CNI, 'cniv2'), contains(variables.CONTROL_CNI, 'all') ),`  
`   or( contains(variables.CONTROL_OS, 'linux'), contains(variables.CONTROL_OS, 'all') )`
`)`

jpayne3506: Expected variations are listed above in the PR description.

camrynl (approved):
The comments in the pipeline template are helpful, but we probably need to have a guide/doc on how to run the load tests and all of the different variables we can customize.

jpayne3506:
Working on that documentation as well. Will be covering intent, purpose, when to use, how to use (variable usage listed here), coverage, and the outputs expected from the pipeline.

vipul-21 (approved):

PR #2290, opened 2023-10-12
[cni]: dualstack overlay hostIP issue hotfix

paulyufan2:
**Reason for Change**:

This PR is to fix this issue: https://dev.azure.com/msazure/One/_workitems/edit/25403093/

If no hostIP is provided, then we assume to use ipv4 flag, if hostIP is provided and it's ipv6 address, then we set flag to ipv6 HNS flag.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -781,7 +781,11 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt) 
 		NATInfo:            opt.natInfo,
 	}
 
-	epPolicies := getPoliciesFromRuntimeCfg(opt.nwCfg)
+	epPolicies, err := getPoliciesFromRuntimeCfg(opt.nwCfg)
+	if err != nil {
+		logger.Error("failed to get policies from runtime configurations", zap.Error(err))
+	}
+
 	epInfo.Policies = append(epInfo.Policies, epPolicies...)
 
 	// Populate addresses.
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -781,8 +781,7 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt) 
 		NATInfo:            opt.natInfo,
 	}
 
-	isIPv6Enabled := opt.resultV6 != nil
-	epPolicies := getPoliciesFromRuntimeCfg(opt.nwCfg, isIPv6Enabled)
+	epPolicies := getPoliciesFromRuntimeCfg(opt.nwCfg)
 	epInfo.Policies = append(epInfo.Policies, epPolicies...)
 
 	// Populate addresses.
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -784,6 +784,7 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt) 
 	epPolicies, err := getPoliciesFromRuntimeCfg(opt.nwCfg)
 	if err != nil {
 		logger.Error("failed to get policies from runtime configurations", zap.Error(err))
+		return epInfo, err
 	}
 
 	epInfo.Policies = append(epInfo.Policies, epPolicies...)
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -781,7 +781,8 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt) 
 		NATInfo:            opt.natInfo,
 	}
 
-	epPolicies, err := getPoliciesFromRuntimeCfg(opt.nwCfg)
+	isIPv6Enabled := opt.resultV6 != nil
+	epPolicies, err := getPoliciesFromRuntimeCfg(opt.nwCfg, isIPv6Enabled)
 	if err != nil {
 		logger.Error("failed to get policies from runtime configurations", zap.Error(err))
 		return epInfo, err
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -785,7 +785,7 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt) 
 	epPolicies, err := getPoliciesFromRuntimeCfg(opt.nwCfg, isIPv6Enabled)
 	if err != nil {
 		logger.Error("failed to get policies from runtime configurations", zap.Error(err))
-		return epInfo, err
+		return epInfo, plugin.Errorf(err.Error())
 	}
 
 	epInfo.Policies = append(epInfo.Policies, epPolicies...)
--- a/cni/network/network_windows.go
+++ b/cni/network/network_windows.go
@@ -280,7 +280,7 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig) ([]policy.Policy, error
 		rawPolicy, err := json.Marshal(&hnsv2.PortMappingPolicySetting{
 			ExternalPort: uint16(mapping.HostPort),
 			InternalPort: uint16(mapping.ContainerPort),
-			VIP:          hostIP,
+			VIP:          mapping.HostIp,
 			Protocol:     protocol,
 			Flags:        flag,
 		})
--- a/cni/network/network_windows.go
+++ b/cni/network/network_windows.go
@@ -248,7 +248,7 @@ func getEndpointDNSSettings(nwCfg *cni.NetworkConfig, result *cniTypesCurr.Resul
 }
 
 // getPoliciesFromRuntimeCfg returns network policies from network config.
-func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig) ([]policy.Policy, error) {
+func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) ([]policy.Policy, error) {
 	logger.Info("Runtime Info", zap.Any("config", nwCfg.RuntimeConfig))
 	var policies []policy.Policy
 	var protocol uint32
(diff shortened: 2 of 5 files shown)


cni/network/network_windows.go
```diff
@@ -248,7 +248,7 @@ func getEndpointDNSSettings(nwCfg *cni.NetworkConfig, result *cniTypesCurr.Resul
 }
 
 // getPoliciesFromRuntimeCfg returns network policies from network config.
-func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) []policy.Policy {
+func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig) []policy.Policy {
```
ramiro-gamarra, line 251: There are a lot of conditions that can fail in this function, why is it not returning an error? The calling function already returns an error if something else fails.

cni/network/network_windows.go
```diff
@@ -266,30 +266,30 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) []p
...
+				Type: policy.EndpointPolicy,
+				Data: hnsv2Policy,
+			}
 
-		// add port mapping policy for v6 if we have IPV6 enabled
-		if isIPv6Enabled {
+			logger.Info("Creating port mapping policyv4",
+				zap.Any("policy", policyv4))
+			policies = append(policies, policyv4)
+		} else {
```
ramiro-gamarra, line 291: There is a lot of duplication here, but the only thing that changes is the HNS flag. This could just be like
```go
hostIP, err := netip.ParseAddr(mapping.HostIp)
// handle err

var flag hnsv2.NatFlags
switch {
case hostIP.Is4():
    flag = hnsv2.NatFlagsLocalRoutedVip
case hostIP.Is6():
    flag = hnsv2.NatFlagsIPv6
}

rawPolicy := hnsv2.PortMappingPolicySetting{
    // ...
    Flags: flag,
}
```

cni/network/network.go
```diff
@@ -781,8 +781,11 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
 		NATInfo:            opt.natInfo,
 	}
 
-	isIPv6Enabled := opt.resultV6 != nil
-	epPolicies := getPoliciesFromRuntimeCfg(opt.nwCfg, isIPv6Enabled)
+	epPolicies, err := getPoliciesFromRuntimeCfg(opt.nwCfg)
+	if err != nil {
+		logger.Error("failed to get policies from runtime configurations", zap.Error(err))
```
ramiro-gamarra, line 786: if an error happens, this should not continue

cni/network/network_windows.go
```diff
@@ -264,59 +263,52 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) []p
 			protocol = policy.ProtocolUdp
 		}
 
+		hostIP := net.ParseIP(mapping.HostIp)
```
ramiro-gamarra, line 266: Can you use `netip.ParseAddr`? It returns an error which makes it clearer that the input may have been malformed, and the return has clearer methods such as `Is4()` or `Is6()` that also make the switch below cleaner.

ramiro-gamarra (changes requested):

cni/network/network_windows.go
```diff
@@ -264,59 +263,55 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) []p
 			protocol = policy.ProtocolUdp
 		}
 
+		hostIP, err := netip.ParseAddr(mapping.HostIp)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse hostIP")
```
vipul-21, line 268: Can we pass the incorrect hostIP as well in the error ?

cni/network/network_windows.go
```diff
@@ -264,59 +263,55 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) []p
...
+		}
+
 		// To support hostport policy mapping
 		// uint32 NatFlagsLocalRoutedVip = 1
-		rawPolicy, _ := json.Marshal(&hnsv2.PortMappingPolicySetting{
+		// To support hostport policy mapping for ipv6 in dualstack overlay mode
+		// uint32 NatFlagsIPv6 = 2
+		var flag hnsv2.NatFlags
```
vipul-21, line 275: So earlier we were adding ipv6 policies based on `cniTypesCurr.Result` (which is from IPAM i suppose), but now we are directly reading the hostIP address type(v4 or v6) and based on that adding the flags ? 
What if the hostip is not provided ?(is that a valid case ?)

cni/network/network.go
```diff
@@ -781,8 +781,12 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
...
+		return epInfo, err
```
ramiro-gamarra, line 787: im surprised the linter doesn't complain about not wrapping the error

ramiro-gamarra (dismissed):

cni/network/network.go
paulyufan2: This is using zap logger here

ramiro-gamarra: it's not about logging, it's about the error returned, unless that linter is not enabled for checking that errors are wrapped? @rbtr

cni/network/network_windows.go
```diff
@@ -264,59 +263,55 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) []p
 			protocol = policy.ProtocolUdp
 		}
 
+		hostIP, err := netip.ParseAddr(mapping.HostIp)
```
tamilmani1989, line 266: is hostIP non-empty for all portmapping enabled pods?  If hostIP is empty, Is that valid case?

cni/network/network_windows.go
```diff
@@ -248,9 +248,8 @@ func getEndpointDNSSettings(nwCfg *cni.NetworkConfig, result *cniTypesCurr.Resul
 }
 
 // getPoliciesFromRuntimeCfg returns network policies from network config.
-func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) []policy.Policy {
-	logger.Info("Runtime Info",
-		zap.Any("config", nwCfg.RuntimeConfig))
+func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) ([]policy.Policy, error) {
```
tamilmani1989, line 251: if `isIPv6Enabled` not used, can we remove it?

cni/network/network.go
tamilmani1989: can we wrap this error before returning?

cni/network/network_windows.go
paulyufan2: Vipul and I thought if customer can modify ipam configuration to set isIPv6Enabled to true and create windows pod in v4Overlay cluster and it would cause the issue.

cni/network/network_windows.go
paulyufan2: it's slightly possible that if we have portmapping but hostIP is empty, for example:
```
   ports:
   - hostPort: 54321
     containerPort: 8080
     protocol: TCP
     hostIP: ""
 ```
 But it's mostly not a case, what do you think?

Comment (paulyufan2) [reply cni/network/network_windows.go:266]: Yes it's not a required field

File: cni/network/network_windows.go (hunk 9 above)
Comment (paulyufan2) [reply cni/network/network_windows.go:251]: Since hostIP is not required field, so we still need "isIPv6Enabled" flag to check if it's ipv4 or ipv6 policy.

File: cni/network/network_windows.go (hunk 8 above)
Comment (ramiro-gamarra) [reply cni/network/network_windows.go:266]: @paulyufan2 are we able to test the scenario that host IP in our pipeline?

File: cni/network/network_windows.go [hunk 10]
```diff
@@ -266,57 +265,61 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) []p
...
+			switch {
+			case hostIP.Is4():
+				flag = hnsv2.NatFlagsLocalRoutedVip
+			case hostIP.Is6() && isIPv6Enabled:
+				flag = hnsv2.NatFlagsIPv6
+			}
+		} else {
+			flag = hnsv2.NatFlagsLocalRoutedVip
+			if isIPv6Enabled {
+				flag = hnsv2.NatFlagsIPv6
```
ramiro-gamarra, line 287: if `hostIP` is empty, shouldn't we just assume ipv4? How is `isIPv6Enabled` determined?

ramiro-gamarra (dismissed):

cni/network/network_windows_test.go
vipul-21, line 1: Let's add more test cases involving ipv6 hostip as well.

cni/network/network_windows.go
vipul-21: What happens, when hostIp is v6 but ipv6Enabled is `false` ? Should that error our invalid hostIp ?

cni/network/network_windows_test.go
paulyufan2: added an ipv6 UT test

cni/network/network_windows.go
paulyufan2: If customer provides wrong ipv6 address in v4Overlay cluster which ipv6Enabled is false, then it will pass the wrong flag to policy but customer cannot create a valid pod. So this is controlled by customer and we cannot return error here

vipul-21 (changes requested):

vipul-21 (approved):

ramiro-gamarra (approved):

PR #2292, opened 2023-10-13
ci: Enable multiple runs with the same commitID

jpayne3506:
**Reason for Change**:

Having multiple runs on the same commit ID was causing conflicts from cluster creation/deletion. This change makes each run unique by giving it a time stamp of DDHHMM (DayHourMinute). Expected clusterName  `<unique cluster name>-<commitID>-<Timestamp>` | `cilium-over-ebdde6c1-132228`

Adding 7 additional characters at the end of the clusterName resulted in having to change the unique cluster names due to the max character limit being 80. Generally the max length for a cluster name should be 14 characters as 32 characters are reserved by the commit hash + timestamp and prefix/suffix can be up to 18 characters (based on region selection).

ex. `MC_ciliovercnie2e-67b2ede2-132124_ciliovercnie2e-67b2ede2-132124_centraluseuap` 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/cni/pipeline.yaml
+++ b/.pipelines/cni/pipeline.yaml
@@ -13,7 +13,7 @@ stages:
           - script: |
               echo "Setting up environment"
               go version
-              echo "##vso[task.setvariable variable=commitID;isOutput=true]$(make revision)"
+              echo "##vso[task.setvariable variable=commitID;isOutput=true]$(echo $(make revision)-$(date "+%d%H%M"))"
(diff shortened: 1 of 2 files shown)


jpayne3506:
Testing scenario: PR pipeline starts run, call /azp run (this should cancel the ongoing run and is controled by ADO), Manually create a new run based off canceled PR run.

vipul-21 (approved):

PR #2294, opened 2023-10-14
feat: Using the Node IP as the Primary IP if the feature flag is enabled for the same

nairashu:
**Reason for Change**:
Updates to CNS to program the Node IP as the Primary IP instead of one of the IPs from the secondary allocations for SNAT if the feature flag is enabled.

**Issue Fixed**:
This update was needed as we were wasting half the subnet for OpenAI to simply provision one primary IP per node as in the Vnet Scale mode we are wasting one CIDR block for Primary IP when the no. of pods is a multiple of 16

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/configuration/cns_config.json
+++ b/cns/configuration/cns_config.json
@@ -33,5 +33,8 @@
     "AZRSettings": {
         "EnableAZR": false,
         "PopulateHomeAzCacheRetryIntervalSecs": 60
+    },
+    "FeatureFlagSettings": {
+        "UseNodeIPAsNCPrimaryIP": false
     }
 }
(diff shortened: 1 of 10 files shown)


rjdenney:
Hey @Ashish,

Is this change still planned to go in at some point? If so how what are your thoughts on how this would work in a dualstack scenario? I have been going through the CNS changes and I don't think there is much that needs to change in CNS/CNI to allow for dualstack in VNET block. The only change I can think of would depend on whether or not we use the first IP in a block as the primary IP or not for v6 like we normally would for v4.

nairashu:
> Hey @nairashu ,
> 
> Is this change still planned to go in at some point? If so how what are your thoughts on how this would work in a dualstack scenario? I have been going through the CNS changes and I don't think there is much that needs to change in CNS/CNI to allow for dualstack in VNET block. The only change I can think of would depend on whether or not we use the first IP in a block as the primary IP or not for v6 like we normally would for v4.

This change is still planned to be integrated at some point. However, will be addressed in another PR created from the main repo rather than the fork as Pipelines are not building from the fork.

PR #2301, opened 2023-10-17
Change expected code to 200, since DNC returns 200

smittal22:
**Reason for Change**:
e #25350573 on ADO

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -373,7 +373,7 @@ type NodeInterrogator interface {
 }
 
 // RegisterNode - Tries to register node with DNC when CNS is started in managed DNC mode
-func registerNode(httpClient acn.HTTPClient, httpRestService cns.HTTPService, dncEP, infraVnet, nodeID string, ni NodeInterrogator) error {
+func registerNode(ctx context.Context, httpClient acn.HTTPClient, httpRestService cns.HTTPService, dncEP, infraVnet, nodeID string, ni NodeInterrogator) error {
 	logger.Printf("[Azure CNS] Registering node %s with Infrastructure Network: %s PrivateEndpoint: %s", nodeID, infraVnet, dncEP)
 
 	var (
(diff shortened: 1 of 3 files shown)


cns/service/main.go
```diff
@@ -409,7 +409,7 @@ func sendRegisterNodeRequest(httpc *http.Client, httpRestService cns.HTTPService
 	}
 	defer response.Body.Close()
 
-	if response.StatusCode != http.StatusCreated {
+	if response.StatusCode != http.StatusOK {
```
ramiro-gamarra, line 412: are there no unit tests for this? if not, some should be added to lock down the expected responses.

rbtr: @smittal22 can you add a test for this to prevent the behavior from drifting accidentally?

smittal22: Added the UTs.

common/utils.go
```diff
@@ -78,6 +79,25 @@ type metadataWrapper struct {
...
+func (c *StandardHTTPClient) Do(req *http.Request) (*http.Response, error) {
+	client := http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "http request failed")
+	}
+
+	return resp, nil
+}
+
```
rbtr, line 100: we don't need this - the same discussion was happening on [this PR](https://github.com/Azure/azure-container-networking/pull/2421#discussion_r1419463500)

cns/service/main.go
```diff
@@ -360,7 +360,7 @@ type NodeInterrogator interface {
 }
 
 // RegisterNode - Tries to register node with DNC when CNS is started in managed DNC mode
-func registerNode(httpc *http.Client, httpRestService cns.HTTPService, dncEP, infraVnet, nodeID string, ni NodeInterrogator) error {
+func registerNode(ctx context.Context, httpClient acn.HTTPClient, httpRestService cns.HTTPService, dncEP, infraVnet, nodeID string, ni NodeInterrogator) error {
```
rbtr, line 363: Instead of a global `acn.HTTPClient` interface, these methods can use a local (private) interface. Remember that in Go, anything that has the necessary methods satisfies the interface. Any type is acceptable that provides the behavior that we need, _including existing types_ like the stdlib `http.Client`

```suggestion
type httpDoer interface {
	Do(req *http.Request) (*http.Response, error)
}

func registerNode(ctx context.Context, httpClient httpDoer, httpRestService cns.HTTPService, dncEP, infraVnet, nodeID string, ni NodeInterrogator) error {
```

rbtr (changes requested):

common/utils.go
rbtr: https://github.com/Azure/azure-container-networking/pull/2421#discussion_r1412577352

smittal22: Addressed.

cns/service/main.go
smittal22: Thanks for the tip. Addressed.

rbtr (approved):

ZetaoZhuang (approved):

PR #2305, opened 2023-10-18
ci: Move E2E runtime variables from job level to stage

jpayne3506:
**Reason for Change**:

E2E did not have consistent assigning of variables. This led to a bug that affected cilium overlay E2E logging because the commitID variable was never passed to the failure logs. This was not showing as a passed run would skip the logging job. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
+++ b/.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
@@ -40,6 +40,9 @@ stages:
       TAG: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.Tag'] ]
       CURRENT_VERSION: $[ stagedependencies.containerize.check_tag.outputs['CurrentTagManifests.currentTagManifests'] ]
       commitID: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.commitID'] ]
+      GOPATH: "$(Agent.TempDirectory)/go" # Go workspace path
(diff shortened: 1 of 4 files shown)


.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
```diff
@@ -40,6 +40,9 @@ stages:
       TAG: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.Tag'] ]
       CURRENT_VERSION: $[ stagedependencies.containerize.check_tag.outputs['CurrentTagManifests.currentTagManifests'] ]
       commitID: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.commitID'] ]
+      GOPATH: "$(Agent.TempDirectory)/go" # Go workspace path
```
vipul-21, line 43: I agree that commit id was shared with other stages that is why it is job level variable. Are all the other variables also need in other stages ?

jpayne3506: This change makes it where the variables are visible to all jobs in their respective stage.

vipul-21: Understood, so the variables are visible or accessible to other jobs in the same stage. We are not using those in those jobs.

vipul-21 (approved):

PR #2311, opened 2023-10-19
chore: wait for deployment to completely delete before proceeding

pjohnst5:
The windows overlay tests kept hitting:
```
--- FAIL: TestScaleDeployment (0.30s)
    load_test.go:141: Scale deployment
panic: could not update deployment win-load-test: Operation cannot be fulfilled on deployments.apps "win-load-test": the object has been modified; please apply your changes to the latest version and try again [recovered]
	panic: could not update deployment win-load-test: Operation cannot be fulfilled on deployments.apps "win-load-test": the object has been modified; please apply your changes to the latest version and try again
```

This was most likely because the deployment was not being completely deleted, before each test run
Now we wait for the deployment to be deleted before proceeding

### DIFF
--- a/test/internal/kubernetes/utils.go
+++ b/test/internal/kubernetes/utils.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/Azure/azure-container-networking/test/internal/retry"
 	"github.com/pkg/errors"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


test/internal/kubernetes/utils.go
```diff
@@ -247,6 +249,19 @@ func WaitForPodDeployment(ctx context.Context, clientset *kubernetes.Clientset,
...
+func WaitForDeploymentToDelete(ctx context.Context, deploymentsClient typedappsv1.DeploymentInterface, d appsv1.Deployment) error {
+	assertDeploymentNotFound := func() error {
+		_, err := deploymentsClient.Get(ctx, d.Name, metav1.GetOptions{})
+		// only if the error is "isNotFound", do we say, the deployment is deleted
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return errors.New(fmt.Sprintf("expected isNotFound error when getting deployment, but got %+v", err))
+	}
+	retrier := retry.Retrier{Attempts: DeleteRetryAttempts, Delay: RetryDelay}
```
jpayne3506, line 261: nit: should we have/add/use a `DeleteRetryDelay = X `? Anything faster then 10 seconds would be nice. Can still do the math to have the total time polled equal to 60s

pjohnst5: Yes

pjohnst5:
Closing in favor of #2312 which is the same, but on a branch directly from ACN, since we are no longer allowed to use forks

PR #2320, opened 2023-10-24
Use ResolveCtl for ubuntu22.04

sheylatrudo:
**Reason for Change**:
Pipelines are failing due to updated image version. Ubuntu 22.04 does not have systemd-resolve, and instead uses resolvectl.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -39,6 +39,8 @@ const (
 	LocalIPKey = "localIP"
 	// InfraVnetIPKey key for infra vnet
 	InfraVnetIPKey = "infraVnetIP"
+	// Ubuntu Release Version for checking which command to use.
+	Ubuntu22 = "22.04"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


rbtr:
@paulyufan2 was previously working [on this](https://github.com/Azure/azure-container-networking/pull/1985), and there were several iterations on that previous PR that might make it a better solution to this issue.
In particular, we should _not_ use the OS version (probably ever, but especially not here) as it's very brittle and locks us in to this OS family. Instead, we should be interrogating the system to understand its capabilities (testing the presence of `systemd-resolve` or `resolvectl`) and doing the right thing based on what we discover.
Should we reopen that change instead?

jpayne3506:
Closed due to fork PR. See #2327 for matching PR.

PR #2321, opened 2023-10-25
fix: Add SubnetGUID to MT PodNetwork CRD

samarths-msft:
**Reason for Change**:
Multi-tenancy orchestrator needs to submit Azure Subnet ResourceGUID to PN CRD, in addition to ARM Resource ID.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/crd/multitenancy/api/v1alpha1/podnetwork.go
+++ b/crd/multitenancy/api/v1alpha1/podnetwork.go
@@ -17,7 +17,8 @@ import (
 // +kubebuilder:printcolumn:name="Status",type=string,priority=1,JSONPath=`.status.status`
 // +kubebuilder:printcolumn:name="Address Prefixes",type=string,priority=1,JSONPath=`.status.addressPrefixes`
 // +kubebuilder:printcolumn:name="Network",type=string,priority=1,JSONPath=`.spec.vnetGUID`
-// +kubebuilder:printcolumn:name="Subnet",type=string,priority=1,JSONPath=`.spec.subnetGUID`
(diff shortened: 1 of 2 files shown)


aggarwal0009 (approved):

isaac-dasan (dismissed):

crd/multitenancy/api/v1alpha1/podnetwork.go
```diff
@@ -40,8 +40,8 @@ type PodNetworkSpec struct {
 	// +kubebuilder:validation:Optional
 	// customer vnet guid
 	VnetGUID string `json:"vnetGUID,omitempty"`
-	// customer subnet id
-	SubnetResourceID string `json:"subnetResourceID,omitempty"`
+	// customer subnet guid
+	SubnetGUID string `json:"subnetGUID,omitempty"`
```
miguelgoms, line 44: I do not think we should remove the ARM resource id field. 
This is still the public identifier of the subnet, and how we'd find the scope subscription / resourcegroup / etc. 
the reason we do not include that for vnet is because the vnet arm path is already implicit on the subnet resource id.

miguelgoms (changes requested):

crd/multitenancy/api/v1alpha1/podnetwork.go
samarths-msft: Restored ARM resource ID field for subnet, added subnetGUID as additional field.

miguelgoms (approved):

aggarwal0009 (approved):

jaer-tsun (approved):

PR #2335, opened 2023-10-28
fix: enable ip forward on transparent vlan network create

QxBytes:
**Reason for Change**:

IP forwarding may not be enabled on the vm when transparent vlan endpoint client first starts (for the first time on this new vm). When namespaces are created, they would have ip forwarding disabled, leading to connectivity issues. This change runs a command to enable forwarding in each namespace on creation, as well as on the vm when the network is first created. Enabling IP forwarding on the namespace may be redundant but is useful for testing (can be removed if necessary). Does not use the existing enable ip forwarding function because that function seems to add additional iptable rules.

Adds transparent vlan multitenancy conflist to dropgz for testing.

**Issue Fixed**:

See above.

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -93,6 +93,11 @@ func (nm *networkManager) newNetworkImpl(nwInfo *NetworkInfo, extIf *externalInt
 	case opModeTransparentVlan:
 		logger.Info("Transparent vlan mode")
 		ifName = extIf.Name
+		nu := networkutils.NewNetworkUtils(nm.netlink, nm.plClient)
+		if err := nu.EnableIPV4Forwarding(); err != nil {
+			return nil, fmt.Errorf("Ipv4 forwarding failed: %w", err)
+		}
+		logger.Info("Ipv4 forwarding enabled")
 	default:
 		return nil, errNetworkModeInvalid
 	}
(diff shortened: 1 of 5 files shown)


network/network_linux.go
```diff
@@ -93,6 +93,11 @@ func (nm *networkManager) newNetworkImpl(nwInfo *NetworkInfo, extIf *externalInt
 	case opModeTransparentVlan:
 		logger.Info("Transparent vlan mode")
 		ifName = extIf.Name
+		nu := networkutils.NewNetworkUtils(nm.netlink, nm.plClient)
```
tamilmani1989, line 96: can we add UT for this code? if it returns error, cni should bubble up the err correctly

network/networkutils/networkutils_linux.go
```diff
@@ -220,6 +221,11 @@ func (nu NetworkUtils) EnableIPForwarding(ifName string) error {
 	return nil
 }
 
+func (nu NetworkUtils) EnableIPV4Forwarding() error {
+	_, err := nu.plClient.ExecuteCommand(enableIPV4ForwardCmd)
+	return errors.Wrap(err, "enable ipv4 forwarding failed")
```
vipul-21, line 226:
```suggestion
	return errors.Wrap(err, "Enable ipv4 forwarding failed")
```

vipul-21, line 226: Should log error as well similar to ipv6 ?

vipul-21 (approved):

tamilmani1989 (approved):

PR #2340, opened 2023-10-30
ci: update to VM SKU with acclnet

jshr-w:
**Reason for Change**:

PR updates VM SKUs on PR pipeline to use accelerated networking.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/pipeline.yaml
+++ b/.pipelines/pipeline.yaml
@@ -358,7 +358,7 @@ stages:
       testDropgz: ""
       clusterType: swift-byocni-nokubeproxy-up
       clusterName: "ciliume2e"
-      vmSize: Standard_B2s_v2
+      vmSize: Standard_D2s_v3
       k8sVersion: ""
       dependsOn: "containerize"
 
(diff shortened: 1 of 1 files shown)


tamilmani1989 (approved):

rbtr:
unless we are only supporting accelnet VMs, it seems like we should leave some tests on the old SKUs still?

vipul-21:
I agree with Evan. We should have tests for non acclnet skus too. 
When you said " that is intended to be done as a separate PR." - Do you mean we will be adding all the stages/tests without acclnet or a subet of stages in the PR pipeline ?

jpayne3506:
> @jshr-w I chose B SKUs for their cost; this change ~doubles the cost of the CI. Do we need to update _all_ of the tests to run on these SKUs?

We actually upgraded from B2 to B2ms to handle the test cases better which puts it in the same cost bracket as B2s_v2. 

Looking at the differences B2s_v2 v D2s_v3 ...

tamilmani1989:
we can have non acclnet sku in release pipeline.. what's the default in AKS? is it acclnet enabled or disabled

jshr-w:
AKS automatically enables acclnet if it is supported on the VM SKU

tamilmani1989:
@vipul-21 @rbtr we can have non-acclnet in release pipeline ... no strong opinion im fine with other way as well

jshr-w:
Closing, after discussion we have decided to change the VM SKUs on the release pipeline instead of the PR pipeline.

PR #2341, opened 2023-10-30
fix: wait for vnet ns to create and ensure veths are inside namespace instead of assuming

QxBytes:
**Reason for Change**:

In the transparent vlan endpoint client add endpoint path, detecting the vnet network namespace would previously assume the vlan interface was already inside the namespace. However, this was not always the case (the vlan interface may not exist or may be in the vm namespace). The fix removes any stray vlan interface pertaining to the container in the vm if it exists and also deletes the network namespace if it does not have a valid vlan interface. Then, the code can proceed as though the vnet namespace never existed. Deleting the vnet namespace will not cause existing endpoints to become inaccessible because they must have errored (and been cleaned up) previously to enter this state. 

Additionally, we also explicitly confirm the vlan interface as well as the vnet interface are successfully created and moved into the vnet namespace. If the vlan interface fails to move, we delete the vnet namespace and the vlan interface. If the vnet interface fails to move into the vnet namespace, we only delete the vnet interface.

Code changes have been made to create unit tests.

**Issue Fixed**:

See above.

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
The following were manually tested (against commit "Address feedback"):
- create pod when its vnet namespace does not exist
- create pod when its vnet namespace and vlan interface does exist
- create pod when its vnet namespace exists but the vlan interface does not exist
- create pod when its vlan interface is in the vm namespace (as if it failed to move into the vnet namespace)
- in all scenarios, we test pod-pod same vnet same vm connectivity, pod-pod same vnet different vm connectivity, pod-internet connectivity, pod-pod same vm different vnet with no connectivity

### DIFF
--- a/network/transparent_vlan_endpointclient_linux.go
+++ b/network/transparent_vlan_endpointclient_linux.go
@@ -195,7 +195,7 @@ func (client *TransparentVlanEndpointClient) setLinkNetNSAndConfirm(name string,
 	// confirm veth was moved successfully
 	err = RunWithRetries(func() error {
 		// retry checking in the namespace if the interface is not detected
-		return client.executeInNSFn(client.vnetNSName, func() error {
+		return ExecuteInNS(client.nsClient, client.vnetNSName, func() error {
 			_, ifDetectedErr := client.netioshim.GetNetworkInterfaceByName(client.vlanIfName)
 			return errors.Wrap(ifDetectedErr, "failed to get vlan veth in namespace")
 		})
--- a/network/transparent_vlan_endpointclient_linux.go
+++ b/network/transparent_vlan_endpointclient_linux.go
@@ -136,7 +136,7 @@ func (client *TransparentVlanEndpointClient) ensureCleanPopulateVM() error {
 		vlanIfNotFoundErr := ExecuteInNS(client.nsClient, client.vnetNSName, func() error {
 			// Ensure the vlan interface exists in the namespace
 			_, err := client.netioshim.GetNetworkInterfaceByName(client.vlanIfName)
-			return err
+			return errors.Wrap(err, "failed to get vlan interface in namespace")
 		})
 		if vlanIfNotFoundErr != nil {
 			logger.Info("Vlan interface doesn't exist even though network namespace exists, deleting network namespace...")
@@ -197,7 +197,7 @@ func (client *TransparentVlanEndpointClient) setLinkNetNSAndConfirm(name string,
 		// retry checking in the namespace if the interface is not detected
 		return client.executeInNSFn(client.vnetNSName, func() error {
 			_, ifDetectedErr := client.netioshim.GetNetworkInterfaceByName(client.vlanIfName)
-			return ifDetectedErr
+			return errors.Wrap(ifDetectedErr, "failed to get vlan veth in namespace")
 		})
 	}, numRetries, sleepInMs)
 	if err != nil {
@@ -280,7 +280,7 @@ func (client *TransparentVlanEndpointClient) PopulateVM(epInfo *EndpointInfo) er
 		// sometimes there is slight delay in interface creation. check if it exists
 		err = RunWithRetries(func() error {
 			_, err = client.netioshim.GetNetworkInterfaceByName(client.vlanIfName)
-			return err
+			return errors.Wrap(err, "failed to get vlan veth")
 		}, numRetries, sleepInMs)
 
 		if err != nil {
@@ -316,20 +316,19 @@ func (client *TransparentVlanEndpointClient) PopulateVM(epInfo *EndpointInfo) er
 
 	// Ensure vnet veth is created, as there may be a slight delay
 	err = RunWithRetries(func() error {
-		var getErr error
-		_, getErr = client.netioshim.GetNetworkInterfaceByName(client.vnetVethName)
-		return getErr
+		_, getErr := client.netioshim.GetNetworkInterfaceByName(client.vnetVethName)
+		return errors.Wrap(getErr, "failed to get vnet veth")
 	}, numRetries, sleepInMs)
 	if err != nil {
 		return errors.Wrap(err, "vnet veth does not exist")
 	}
 
 	// Ensure container veth is created, as there may be a slight delay
-	var containerIf *net.Interface = nil
+	var containerIf *net.Interface
 	err = RunWithRetries(func() error {
 		var getErr error
 		containerIf, getErr = client.netioshim.GetNetworkInterfaceByName(client.containerVethName)
-		return getErr
+		return errors.Wrap(getErr, "failed to get container veth")
 	}, numRetries, sleepInMs)
 	if err != nil {
 		return errors.Wrap(err, "container veth does not exist")
@@ -708,8 +707,8 @@ func ExecuteInNS(nsc NamespaceClientInterface, nsName string, f func() error) er
 	return f()
 }
 
-func RunWithRetries(f func() error, maxRuns int, sleepMs int) error {
-	var err error = nil
+func RunWithRetries(f func() error, maxRuns, sleepMs int) error {
+	var err error
 	for i := 0; i < maxRuns; i++ {
 		err = f()
 		if err == nil {
--- a/network/transparent_vlan_endpointclient_linux.go
+++ b/network/transparent_vlan_endpointclient_linux.go
@@ -123,10 +123,8 @@ func (client *TransparentVlanEndpointClient) AddEndpoints(epInfo *EndpointInfo) 
 	})
 }
 
-func (client *TransparentVlanEndpointClient) createNetworkNamespace(vmNS, numRetries int) error {
-	var isNamespaceUnique bool
-
-	for i := 0; i < numRetries; i++ {
+func (client *TransparentVlanEndpointClient) createNetworkNamespace(vmNS int) error {
+	createNsImpl := func() error {
 		vnetNS, err := client.netnsClient.NewNamed(client.vnetNSName)
 		if err != nil {
 			return errors.Wrap(err, "failed to create vnet ns")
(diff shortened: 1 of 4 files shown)


network/transparent_vlan_endpointclient_linux.go
```diff
@@ -126,32 +125,84 @@ func (client *TransparentVlanEndpointClient) AddEndpoints(epInfo *EndpointInfo)
...
+	var existingErr error
+	client.vnetNSFileDescriptor, existingErr = client.netnsClient.GetFromName(client.vnetNSName)
+	if existingErr == nil {
+		// The namespace exists
+		vlanIfNotFoundErr := ExecuteInNS(client.nsClient, client.vnetNSName, func() error {
+			// Ensure the vlan interface exists in the namespace
+			_, err := client.netioshim.GetNetworkInterfaceByName(client.vlanIfName)
+			return errors.Wrap(err, "failed to get vlan interface in namespace")
+		})
+		if vlanIfNotFoundErr != nil {
```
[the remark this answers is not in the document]
QxBytes: Good catch, the possible errors that I see based on the documentation appear to be:
- No such interface, in which case we should delete the namespace on line 143
- Some failure with the iptable or the interface name isn't valid, in which case it could be a one-off error and the vlan interface _could_ actually exist (we just can't get it right now), in which case containers may exist that use this namespace, which means the namespace should not be deleted.
- If there are other error scenarios feel free to mention
I'll update the code to match on the no such interface error.

dropgz/build/cniTest_linux.Dockerfile
```diff
@@ -20,7 +20,6 @@ COPY --from=azure-ipam /azure-ipam/*.conflist pkg/embed/fs
 COPY --from=azure-ipam /azure-ipam/bin/* pkg/embed/fs
 COPY --from=azure-vnet /azure-container-networking/cni/azure-$OS.conflist pkg/embed/fs/azure.conflist
 COPY --from=azure-vnet /azure-container-networking/cni/azure-$OS-swift.conflist pkg/embed/fs/azure-swift.conflist
-COPY --from=azure-vnet /azure-container-networking/cni/azure-$OS-multitenancy-transparent-vlan.conflist pkg/embed/fs/azure-multitenancy-transparent-vlan.conflist
```
vipul-21, line 23: Is this related ?

vipul-21 (dismissed):

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -126,32 +125,94 @@ func (client *TransparentVlanEndpointClient) AddEndpoints(epInfo *EndpointInfo)
...
+		// The namespace exists
+		vlanIfErr := ExecuteInNS(client.nsClient, client.vnetNSName, func() error {
+			// Ensure the vlan interface exists in the namespace
+			_, err := client.netioshim.GetNetworkInterfaceByName(client.vlanIfName)
+			return errors.Wrap(err, "failed to get vlan interface in namespace")
+		})
+
+		// If the vlan veth for sure doesn't exist, we delete the vnet ns, but if we can't query, we can't delete the ns (the vlan veth might exist!)
+		if vlanIfErr != nil {
+			if strings.Contains(vlanIfErr.Error(), "no such network interface") {
```
tamilmani1989, line 144: instead of checking for this error message, can cni just remove network namespace if it gets error in finding vlan device in veth namespace

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -126,32 +125,94 @@ func (client *TransparentVlanEndpointClient) AddEndpoints(epInfo *EndpointInfo)
 	})
 }
 
-func (client *TransparentVlanEndpointClient) createNetworkNamespace(vmNS, numRetries int) error {
-	var isNamespaceUnique bool
+// Called from PopulateVM, Namespace: VM and Vnet
```
tamilmani1989, line 128: its callled from AddEndpoints. update comment

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -126,32 +125,94 @@ func (client *TransparentVlanEndpointClient) AddEndpoints(epInfo *EndpointInfo)
...
+				logger.Info("Vlan interface doesn't exist even though network namespace exists, deleting network namespace...")
+				delErr := client.netnsClient.DeleteNamed(client.vnetNSName)
+				if delErr != nil {
+					return errors.Wrap(delErr, "failed to cleanup/delete ns after noticing vlan veth does not exist")
+				}
+			} else {
+				return errors.Wrap(vlanIfErr, "could not determine if vlan veth exists in vnet namespace")
+			}
+		}
+	} else {
```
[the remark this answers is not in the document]
tamilmani1989: i dont think log is required.. anyway we log in populateVM right?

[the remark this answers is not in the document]
QxBytes: Yes we do log in populateVM, but here (see above comment) we want to know if GetNetworkInterfaceByName returned an error other than "interface not found", in case the method itself fails. Again, if you've only ever seen it fail if the interface is not found, we can probably remove the log.

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -126,32 +125,94 @@ func (client *TransparentVlanEndpointClient) AddEndpoints(epInfo *EndpointInfo)
...
+	// Delete the vlan interface in the VM namespace if it exists
+	_, vlanIfInVMErr := client.netioshim.GetNetworkInterfaceByName(client.vlanIfName)
+	if vlanIfInVMErr == nil {
+		// The vlan interface exists in the VM ns because it failed to move into the network ns previously and needs to be cleaned up
+		logger.Info("Vlan interface exists on the VM namespace, deleting", zap.String("vlanIfName", client.vlanIfName))
+		if delErr := client.netlink.DeleteLink(client.vlanIfName); delErr != nil {
+			return errors.Wrap(delErr, "failed to clean up vlan interface")
+		}
+	} else {
+		logger.Info("Expecting vlan interface not to exist", zap.String("result", vlanIfInVMErr.Error()))
```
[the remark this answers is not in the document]
QxBytes: (see above comment)

network/transparent_vlan_endpointclient_linux.go
tamilmani1989: my concern is what if they changed the error message in future? Im fine with this if we have UT for catching this error message with exact string

network/transparent_vlan_endpointclient_linux.go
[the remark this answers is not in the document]
tamilmani1989: ya lets remove the log.. anyway we log in populateVM

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -171,7 +232,7 @@ func (client *TransparentVlanEndpointClient) PopulateVM(epInfo *EndpointInfo) er
 		// We assume the only possible error is that the namespace doesn't exist
 		logger.Info("No existing NS detected. Creating the vnet namespace and switching to it")
 
-		if err = client.createNetworkNamespace(vmNS, numRetries); err != nil {
+		if err = client.createNetworkNamespace(vmNS); err != nil {
```
tamilmani1989, line 235: why dont we call this also with RunwithRetries instead of calling within the method?

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -171,7 +232,7 @@ func (client *TransparentVlanEndpointClient) PopulateVM(epInfo *EndpointInfo) er
 		// We assume the only possible error is that the namespace doesn't exist
 		logger.Info("No existing NS detected. Creating the vnet namespace and switching to it")
```
tamilmani1989, line 233: can we log error here?

network/transparent_vlan_endpointclient_linux.go
QxBytes: To clarify, do you mean catching/matching when the error is _not_ "no such network interface" (ex: [SyscallError](https://cs.opensource.google/go/go/+/master:src/os/error.go;l=67;drc=ac85f2bedd405d64b54c3a248c9079b6cde57359;bpv=1;bpt=1)) and any other error will trigger deletion of the namespace?

network/transparent_vlan_endpointclient_linux.go
QxBytes: yes I will add the `existingErr` message to this `logger.Info` statement.

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -126,32 +125,94 @@ func (client *TransparentVlanEndpointClient) AddEndpoints(epInfo *EndpointInfo)
...
+		vlanIfErr := ExecuteInNS(client.nsClient, client.vnetNSName, func() error {
+			// Ensure the vlan interface exists in the namespace
+			_, err := client.netioshim.GetNetworkInterfaceByName(client.vlanIfName)
+			return errors.Wrap(err, "failed to get vlan interface in namespace")
+		})
+
+		// If the vlan veth for sure doesn't exist, we delete the vnet ns, but if we can't query, we can't delete the ns (the vlan veth might exist!)
+		if vlanIfErr != nil {
+			if strings.Contains(vlanIfErr.Error(), "no such network interface") {
+				logger.Info("Vlan interface doesn't exist even though network namespace exists, deleting network namespace...")
```
rbtr: `VLAN` or `vlan`, but never `Vlan`. and logs don't need to be sentences

QxBytes:
@tamilmani1989 The network namespace `GetNamed(...)` function doesn't have a clearly named error-- it seems like it will return some kind of syscall error (any error related to opening the file).

- Checked CNI statefile before creating pods, after creating pods, and after deleting pods, and the file looks the same as when it started.
- Routes appear to be properly deleted and there don't appear to be any straggling interfaces (other than what we expect) after deletion

tamilmani1989 (approved):

PR #2299, opened 2023-10-17
[CNI] Migrate zap logger for acn ipam package

paulyufan2:
**Reason for Change**:

This PR is to migrate zap logger for ACN ipam package

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] relevant PR labels added

**Notes**:

PR #2344, opened 2023-11-02
fix zap logger rotation issues

paulyufan2:
**Reason for Change**:

This PR is to fix log rotation issue that is crashed after windows nodes are restarted.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/ipam/plugin/main.go
+++ b/cni/ipam/plugin/main.go
@@ -10,7 +10,6 @@ import (
 	"github.com/Azure/azure-container-networking/cni"
 	"github.com/Azure/azure-container-networking/cni/ipam"
 	"github.com/Azure/azure-container-networking/common"
-	"github.com/Azure/azure-container-networking/log"
 )
 
 const name = "azure-vnet-ipam"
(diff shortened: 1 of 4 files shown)


vipul-21:
Can we add more to the description about what the rotation issue is ?

cni/network/plugin/main.go
```diff
@@ -295,15 +294,6 @@ func main() {
 		os.Exit(0)
 	}
 
-	log.SetName(name)
```
vipul-21, line 298: These were added to support a known issue(incompatilbility of loggers between cns/cni) if i remember correctly ? Is that no longer valid ?

paulyufan2: yes after we migrate all legacy logs to zap logger, we do not need this anymore and it is using legacy logs, trying to rotate log files but in the meantime, zap logger lumberjack is rotating log files as well, causing windows pods to be crashed and cannot get back. So this PR is to fix this issue.

vipul-21: Let's run CNI release pipeline with this change too

paulyufan2: yes it's running

cni/ipam/plugin/main.go
```diff
@@ -23,17 +22,6 @@ func main() {
 	var config common.PluginConfig
 	config.Version = version
 
-	logDirectory := "" // Sets the current location as log directory
```
tamilmani1989, line 26: did we validate ipam logs moved to zap

paulyufan2: yes I did with this PR: https://github.com/Azure/azure-container-networking/pull/2299
Verified azure-ipam.log

tamilmani1989 (approved):

vipul-21 (approved):

behzad-mir (commented):
plz make ssure we don't have any "github.com/Azure/azure-container-networking/log in the CNI and network related packages

behzad-mir (approved):

PR #2345, opened 2023-11-02
test: validate secondary endpoint client failure

jaer-tsun:
**Reason for Change**:
Add a test to validate that endpoint will not be created if SecondaryEndpointClient fails

Passing NIC Type var from CNS response down instead of hardcoding

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/network/endpoint_test.go
+++ b/network/endpoint_test.go
@@ -262,24 +262,25 @@ var _ = Describe("Test Endpoint", func() {
 				Expect(ep.Id).To(Equal(epInfo.Id))
 			})
 		})
-		Context("When secondary endpoint add failed", func() {
-			It("Should not be added to the network", func() {
-				_, ipnet, _ := net.ParseCIDR("0.0.0.0/0")
-				nw := &network{
-					Endpoints: map[string]*endpoint{},
-					Mode:      opModeTransparent,
-					extIf:     &externalInterface{BridgeName: "testbridge"},
-				}
-				epInfo := &EndpointInfo{
-					Id:      "768e8deb-eth1",
-					IfName:  eth0IfName,
-					NICType: cns.InfraNIC,
-				}
-				secondaryEpInfo := &EndpointInfo{
-					MacAddress: netio.BadHwAddr, // mock netlink will fail to set link state on bad eth
-					NICType:    cns.DelegatedVMNIC,
-					Routes:     []RouteInfo{{Dst: *ipnet}},
-				}
+		Context("When secondary endpoint client is used", func() {
+			_, ipnet, _ := net.ParseCIDR("0.0.0.0/0")
+			nw := &network{
+				Endpoints: map[string]*endpoint{},
+				Mode:      opModeTransparent,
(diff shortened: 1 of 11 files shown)


network/endpoint_test.go
```diff
@@ -262,6 +262,37 @@ var _ = Describe("Test Endpoint", func() {
...
+				}
+				secondaryEpInfo := &EndpointInfo{
+					MacAddress: netio.BadHwAddr, // mock netlink will fail to set link state on bad eth
+					NICType:    cns.DelegatedVMNIC,
+					Routes:     []RouteInfo{{Dst: *ipnet}},
+				}
+				ep, err := nw.newEndpointImpl(nil, netlink.NewMockNetlink(false, ""), platform.NewMockExecClient(false),
+					netio.NewMockNetIO(false, 0), nil, NewMockNamespaceClient(), []*EndpointInfo{epInfo, secondaryEpInfo})
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(Equal("SecondaryEndpointClient Error: Mock Netlink Error"))
```
vipul-21, line 286: nit: we can reuse the variable for error `netlink.ErrorMockNetlink`.

vipul-21 (approved):

network/endpoint_test.go
jaer-tsun: sure, updated

vipul-21 (dismissed):

network/endpoint_test.go
```diff
@@ -262,6 +262,37 @@ var _ = Describe("Test Endpoint", func() {
...
+					netio.NewMockNetIO(false, 0), nil, NewMockNamespaceClient(), []*EndpointInfo{epInfo, secondaryEpInfo})
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(Equal("SecondaryEndpointClient Error: " + netlink.ErrorMockNetlink.Error()))
+				Expect(ep).To(BeNil())
+
+				secondaryEpInfo.MacAddress = netio.HwAddr
+				ep, err = nw.newEndpointImpl(nil, netlink.NewMockNetlink(false, ""), platform.NewMockExecClient(false),
+					netio.NewMockNetIO(false, 0), nil, NewMockNamespaceClient(), []*EndpointInfo{epInfo, secondaryEpInfo})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(ep.Id).To(Equal(epInfo.Id))
```
nddq, line 293: If this test is testing for secondary endpoint add failed, then I don't think we should be testing for when it is success here. Maybe have a separate test for this?

nddq (changes requested):
Just a nit comment on the UT, just wondering, will the secondary endpoint client failed only on bad MAC address, or would there be other scenarios that could cause this failure also?

network/endpoint_test.go
jaer-tsun: sure, split

nddq (approved):

vipul-21 (approved):

PR #2362, opened 2023-11-07
fix: Cilium CLI version

jpayne3506:
**Reason for Change**:

Unblocks the PR pipeline as Cilium CLI update v0.15.11 -> v0.15.12 brought in new tests for features we may or may not support. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
https://github.com/cilium/cilium-cli/releases/tag/v0.15.12 
https://github.com/cilium/cilium-cli/compare/v0.15.11...v0.15.12

### DIFF
--- a/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
+++ b/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
@@ -130,7 +130,7 @@ stages:
         steps:
           - script: |
               echo "install cilium CLI"
-              CILIUM_CLI_VERSION=v0.13.2
+              CILIUM_CLI_VERSION="v0.15.11"
               CLI_ARCH=amd64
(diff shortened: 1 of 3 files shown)


vipul-21 (approved):
Should we add it as a pipeline parameter assuming we will need to eventually move to the new version ? This unblocks us too though so I am okay with it for now.

jpayne3506:
We can parameterize it in a follow on PR, but I want to lean on the original implementation. 

For the parameterization I will have it be able to consume any version that is provided and on empty use the original implementation of consuming the current stable release of Cilium CLI. 

Should we look into skipping these tests? The test infra itself is rather robust, so I think we should look more into the root cause of missing feature flags ect, instead of just skipping... https://github.com/cilium/cilium-cli/blob/6948dd10b838a883c985a4b1f5a40daa68db00bf/connectivity/suite.go#L755-L759

tamilmani1989 (approved):
i hope this is just workaround.. cilium cli option supports skipping test.. we should move to latest and skip tests if its not supported in azure

PR #2363, opened 2023-11-07
Feat: Stateless cni migration

behzad-mir:
This PR contains code changes related to Stateless CNI Migration. 
1- There has been code change to use cnireconciler (pointing to statefull CNI binary) when the managedEndpoint and Migration flag on CNS is set.
2- There has been code changes on CNI GetEndpoint function to fetch the HNS Endpoint corresponded to an IP address to cover the cases for the missing HNSEndpointID in statefile.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/cns/cnireconciler/podinfoprovider.go
+++ b/cns/cnireconciler/podinfoprovider.go
@@ -7,6 +7,7 @@ import (
 	"github.com/Azure/azure-container-networking/cni/client"
 	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/cns/restserver"
+	"github.com/Azure/azure-container-networking/platform"
 	"github.com/Azure/azure-container-networking/store"
 	"github.com/pkg/errors"
 	"k8s.io/utils/exec"
@@ -15,7 +16,11 @@ import (
 // NewCNIPodInfoProvider returns an implementation of cns.PodInfoByIPProvider
 // that execs out to the CNI and uses the response to build the PodInfo map.
 func NewCNIPodInfoProvider() (cns.PodInfoByIPProvider, error) {
-	return newCNIPodInfoProvider(exec.New())
+	return newCNIPodInfoProvider(exec.New(), platform.CNIBinaryPath)
+}
+
+func NewStatefullCNIPodInfoProvider() (cns.PodInfoByIPProvider, error) {
+	return newCNIPodInfoProvider(exec.New(), platform.StatefullCNIBinaryPath)
 }
 
 func NewCNSPodInfoProvider(endpointStore store.KeyValueStore) (cns.PodInfoByIPProvider, error) {
@@ -39,9 +44,9 @@ func newCNSPodInfoProvider(endpointStore store.KeyValueStore) (cns.PodInfoByIPPr
 	}), nil
 }
 
-func newCNIPodInfoProvider(exec exec.Interface) (cns.PodInfoByIPProvider, error) {
+func newCNIPodInfoProvider(exec exec.Interface, cniBinaryPath string) (cns.PodInfoByIPProvider, error) {
 	cli := client.New(exec)
-	state, err := cli.GetEndpointState()
+	state, err := cli.GetEndpointState(cniBinaryPath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to invoke CNI client.GetEndpointState(): %w", err)
 	}
--- a/cns/cnireconciler/podinfoprovider.go
+++ b/cns/cnireconciler/podinfoprovider.go
@@ -7,7 +7,6 @@ import (
 	"github.com/Azure/azure-container-networking/cni/client"
 	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/cns/restserver"
-	"github.com/Azure/azure-container-networking/platform"
 	"github.com/Azure/azure-container-networking/store"
 	"github.com/pkg/errors"
 	"k8s.io/utils/exec"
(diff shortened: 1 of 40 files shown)


rbtr (commented):
I'm confused about the state of the world at the point that this would run. This still seems to assume some amount of stateless CNI has happened - in that the existing CNI is expected to be renamed to `azure-vnet-stateful`, at least.

We need to stop thinking about this as part of the stateless CNI change. This is a CNI-managed state to CNS-managed state migration. It is happening entirely independently of any other feature and should not assume anything about the progress of other features. It shouldn't be part of the stateless CNI changeset (except as a prereq) or branch.

behzad-mir:
> I'm confused about the state of the world at the point that this would run. This still seems to assume some amount of stateless CNI has happened - in that the existing CNI is expected to be renamed to `azure-vnet-stateful`, at least.
> 
> We need to stop thinking about this as part of the stateless CNI change. This is a CNI-managed state to CNS-managed state migration. It is happening entirely independently of any other feature and should not assume anything about the progress of other features. It shouldn't be part of the stateless CNI changeset (except as a prereq) or branch.

At the moment that we enter into migration, statless CNI has already been deployed. There are two ways to handles this: 1) We should either duplicate the whole GetEndpointState procedure with store lock initialization from the statefull CNI into the stateless code, or 2) just invoke the statefull CNI binary.
I went with the second option since it will avoid the code duplication and having the whole statefile recovery on stateless CNI will simply spoil the whole point of "stateless" CNI. In my opinion, the migration should call into statefull CNI from design perspective since the statefile is part of that binary.

cns/cnireconciler/podinfoprovider.go
```diff
@@ -15,7 +16,11 @@ import (
 // NewCNIPodInfoProvider returns an implementation of cns.PodInfoByIPProvider
 // that execs out to the CNI and uses the response to build the PodInfo map.
 func NewCNIPodInfoProvider() (cns.PodInfoByIPProvider, error) {
-	return newCNIPodInfoProvider(exec.New())
+	return newCNIPodInfoProvider(exec.New(), platform.CNIBinaryPath)
+}
+
+func NewStatefullCNIPodInfoProvider() (cns.PodInfoByIPProvider, error) {
+	return newCNIPodInfoProvider(exec.New(), platform.CNIStateFilePath)
```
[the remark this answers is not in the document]
behzad-mir: OOps, Was a mistake

network/endpoint_windows.go
```diff
@@ -492,3 +493,28 @@ func (ep *endpoint) getInfoImpl(epInfo *EndpointInfo) {
...
+	}
+	hcnEndpoints, err := Hnsv2.ListEndpointsOfNetwork(hnsResponse.Id)
+	if err != nil {
+		return epInfo, err
+	}
+	for _, hcnEndpoint := range hcnEndpoints {
+		for _, ipConfiguration := range hcnEndpoint.IpConfigurations {
+			for _, ipAddress := range ipAddresses {
+				prefixLength, _ := ipAddress.Mask.Size()
+				if ipConfiguration.IpAddress == ipAddress.IP.String() && ipConfiguration.PrefixLength == uint8(prefixLength) {
```
tamilmani1989, line 512: is matching prefix length required? isn't that be always /32 ip?

network/endpoint.go
```diff
@@ -347,3 +347,21 @@ func GetPodNameWithoutSuffix(podName string) string {
...
+	endpointInfo, err := epInfo.GetEndpointInfoByIPImpl(ipAddresses, networkID)
+	if err != nil {
+		return nil, err
+	}
+	return endpointInfo, nil
+}
+
+// IsEndpointStateComplete returns true if both HNSEndpointID and HostVethName are missing.
+func (epInfo *EndpointInfo) IsEndpointStateIncomplete() bool {
+	if epInfo.HNSEndpointID == "" && epInfo.IfName == "" {
```
tamilmani1989, line 363: i guess `epInfo.Ifname` is not hostvethname. there would be field in epInfo explictly saying hostveth interface name

tamilmani1989 (changes requested):
Does this change include linux support.? If you intend to create separate PR for it, update the title to reflect its for windows.

network/endpoint.go
behzad-mir: We do have a HostIfName on ednpoint. DO you want me to add it to endpointInfo struct too?

cns/service/main.go
```diff
@@ -1191,10 +1191,19 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 		if err != nil {
 			if errors.Is(err, store.ErrKeyNotFound) {
 				logger.Printf("[Azure CNS] No endpoint state found, skipping initializing CNS state")
+				if os.Getenv("StatelessCNIMigration") == "true" {
```
rbtr, line 1194: can this come from the config instead of the env? and to align with the convention there, maybe`EnableManagedStateMigration`

network/endpoint_windows.go
```diff
@@ -492,3 +493,28 @@ func (ep *endpoint) getInfoImpl(epInfo *EndpointInfo) {
 func (nm *networkManager) updateEndpointImpl(nw *network, existingEpInfo *EndpointInfo, targetEpInfo *EndpointInfo) (*endpoint, error) {
 	return nil, nil
 }
+
+// GetEndpointInfoByIPImpl returns an endpointInfo with the corrsponding HNS Endpoint ID that matches an specific IP Address.
+func (epInfo *EndpointInfo) GetEndpointInfoByIPImpl(ipAddresses []net.IPNet, networkID string) (*EndpointInfo, error) {
```
vipul-21, line 498: Can we add logs to the functions when it fails/succeeds while fetching Network/Endpoints etc.

network/endpoint.go
vipul-21: This would always return `false` for linux ? As HNSendpointId is empty for linux

Issue #2333, opened 2023-10-27
CNI Installer migration


rbtr:
This is a tracking issue for the work to migrate from the single `dropgz` image to per-component `-installer` images.

Up until now, we have released the CNI installer image (`cni-dropgz`) as a versioned component separately from the CNI releases themselves. This has created a bit of versioning hell, as we struggle to maintain multiple release trains of the CNI (and thus CNI installer) and azure-ipam when the CNI/azure-ipam and installer version are not correlated. 

The proposed change is to flip the architecture of the CNI installer: treat `dropgz` as a library, and leverage it to build separate installer images for each versioned component which we need to install via container. 

We will migrate from the current state of:

|image|binaries|version|
|-|-|-|
|cni-dropgz||a.b.c|
||azure-vnet|j.k.l|
||azure-ipam|x.y.z|

to 
|image|binaries|version|
|-|-|-|
|cni-installer||j.k.l|
||azure-vnet|j.k.l|
|ipam-installer||x.y.z|
||azure-ipam|x.y.z|

Migration steps:
- [x] ~Build CNI installer images~ #2324
- [x] ~Build azure-ipam installer images~ #2339 
- [x] ~Set up MCR publishing on new images~ MCR #2917
- [x] ~Add MCR sovcloud syndication~ MCR #3016
- [x] ~Migrate pipelines to new images~ #2372
- [x] ~Add v1 scenario installation daemonset to AKS~ AKS #9774675
- [x] Migrate existing downstreams to new images
  - [x] AgentBaker https://github.com/Azure/AgentBaker/pull/4256
  - [x] AKS
  - [x] ~CAPZ~ https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/4315
- [x] Remove existing dropgz image infra (Dockerfiles, pipelines, etc) 
  - [x] ~dropgz-test images~ #2372
  - [x] dropgz images
- [x] Remove AzCNI bins from AgentBaker

rbtr:
this is done for 1.5+, and can be closed as completed once #2871 is completed and AKS migrates for k8s 1.27.

rbtr:
AKS has migrated and cni-dropgz is unused. Purging it from the VHD [here](https://github.com/Azure/AgentBaker/pull/5300) and closing this as complete 🎉

PR #2372, opened 2023-11-09
chore: migrate to azure-cni and azure-ipam from dropgz-test

rbtr:
**Reason for Change**:

As part of https://github.com/Azure/azure-container-networking/issues/2333, the migration from omnibus dropgz to per-component installer images, start using the component installer images in the test pipelines in place of dropgz-test. This is a functionally equivalent change - we will test the current rev of cni/azure-ipam which is what dropgz-test previously bundled.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


.pipelines/pipeline.yaml
```diff
@@ -377,94 +331,45 @@ stages:
...
-    parameters:
-      name: "cilium_podsubnet_cni_e2e"
-      displayName: CNI - Cilium Podsubnet
-      testDropgz: true
-      clusterType: swift-byocni-nokubeproxy-up
-      clusterName: "cilpodcnie2e"
-      vmSize: Standard_B2ms
-      k8sVersion: ""
-      dependsOn: "containerize"
+      dependsOn: "test"
```
camrynl, line 338: nice! this makes the pipeline much more efficient. what happens if the create cluster step completes before the publish stage? does the e2e test step wait?

.pipelines/cni/pipeline.yaml
```diff
@@ -16,12 +16,6 @@ stages:
               echo "##vso[task.setvariable variable=commitID;isOutput=true]$(echo $(make revision)-$(date "+%d%H%M"))"
               echo "##vso[task.setvariable variable=npmVersion;isOutput=true]$(make npm-version)"
               echo "##vso[task.setvariable variable=cnsVersion;isOutput=true]$(CNS_VERSION)"
-
-              if [ -z $(DROP_GZ_VERSION) ]; then
```
camrynl, line 20: will we still have a way to test the cni/cns for different release trains?

camrynl: or that will come by default with specifying the cns version?

.pipelines/pipeline.yaml
rbtr: yes, the e2e stages depend on the publish step, so it would wait there

.pipelines/cni/pipeline.yaml
rbtr: my expectation is that the release trains will be separate `release/v*` branches, so the pipeline only needs to test <its current codebase>. I don't need to specify a dropgz version because I am always testing CNI/CNS built from the current rev, not just dropgz built at the current rev (which had older release tags of CNI).
the pipeline caller can specify a CNI, CNS, or IPAM version to test any version skews that we think are necessary, but I have not included any now.

camrynl (approved):

.pipelines/singletenancy/aks/e2e-step-template.yaml
```diff
@@ -32,23 +32,23 @@ steps:
...
-          echo "Dropgz image: $DROP_GZ_URL"
+          export CNI_IMAGE=$( make cni-image-name-and-tag OS='windows' ARCH=${{ parameters.arch }} OS_VERSION=${{ parameters.os_version }})
+          echo "CNI image: $CNI_IMAGE"
           envsubst < ./test/integration/manifests/cni/cni-installer-v1-windows.yaml | kubectl apply -f -
           kubectl rollout status daemonset/azure-cni-windows -n kube-system
         else
-          export DROP_GZ_URL=$( make cni-dropgz-test-image-name-and-tag OS=${{ parameters.os }} ARCH=${{ parameters.arch }} CNI_DROPGZ_VERSION=${{ parameters.version }})
-          echo "Dropgz image: $DROP_GZ_URL"
+          export CNI_IMAGE=$(make cni-image-name-and-tag OS=${{ parameters.os }} ARCH=${{ parameters.arch }})
+          echo "CNI image: $CNI_IMAGE"
```
jpayne3506, line 46: Is this missing `CNI_VERSION=$(make cni-version)` to generate the correct url? Or is it defaulting correctly as we are only using a singular cni image variety now?

.pipelines/cni/pipeline.yaml
jpayne3506: What you are describing to specify a version is being deleted here. The ability to run based off a tag/rev still exists as the default value of `DROP_GZ_VERSION = ""` . 

This also exists to be able to reuse a CNI release image candidate. We do not want to generate a fresh image that could be potentially (even though it should not be) different from what we will release.

test/integration/manifests/cns/daemonset-linux.yaml
```diff
@@ -41,7 +41,7 @@ spec:
           effect: NoSchedule
       containers:
         - name: cns-container
-          image: acnpublic.azurecr.io/azure-cns:v1.5.10
+          image: acnpublic.azurecr.io/azure-cns:latest
```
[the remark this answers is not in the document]
rbtr: these manifests are not used as-is, this whole image reference is replaced by the test. consider this to be an implicit placeholder that just looks like a valid reference.

test/integration/manifests/cns/daemonset-windows.yaml
```diff
@@ -104,13 +104,13 @@ spec:
                   apiVersion: v1
                   fieldPath: spec.nodeName
       initContainers:
-        - name: init-cni-dropgz
-          image: acnpublic.azurecr.io/cni-dropgz:CNI_DROPGZ_VERSION
+        - name: cni-installer
+          image: acnpublic.azurecr.io/cni-dropgz:latest
           imagePullPolicy: Always
-          command: ["%CONTAINER_SANDBOX_MOUNT_POINT%/dropgz.exe"]
+          command: ["%CONTAINER_SANDBOX_MOUNT_POINT%/dropgz"]
```
pjohnst5, line 110: Should we keep .exe? Since this is windows?

dropgz/build/cniTest_windows.Dockerfile
```diff
@@ -1,30 +0,0 @@
-ARG ARCH
```
pjohnst5, line 1: Yay, I'm excited to just be able to use `azure-cni` as an init container, so cool! 🤩

pjohnst5 (commented):
Just one question for me

test/integration/manifests/cns/daemonset-windows.yaml
rbtr: don't need it

pjohnst5: So it's still able to invoke just `dropgz`, even though we call it `dropgz.exe` here?
https://github.com/Azure/azure-container-networking/blob/2a90a322b1d22708013cd12a3213fd1d917edeb3/dropgz/build/windows.Dockerfile#L37

rbtr: no, this is the old image and there are no changes to how it is used. the reference in the ds is to the new image

jpayne3506 (approved):

PR #2377, opened 2023-11-09
ci: Skip cilium encryption tests

jpayne3506:
**Reason for Change**:

Update to Cilium CLI brought in tests that we do not functionally need or use.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
+++ b/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
@@ -159,7 +159,7 @@ stages:
             displayName: "Get AKS Cluster"
           - script: |
               kubectl delete ns load-test
-              cilium connectivity test --connect-timeout 4s --request-timeout 30s
+              cilium connectivity test --connect-timeout 4s --request-timeout 30s --test '!pod-to-pod-encryption,!node-to-node-encryption'
             retryCountOnTaskFailure: 6
             name: "CiliumConnectivityTests"
             displayName: "Run Cilium Connectivity Tests"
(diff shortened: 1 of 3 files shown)


tamilmani1989 (approved):

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -159,7 +159,7 @@ stages:
             displayName: "Get AKS Cluster"
           - script: |
               kubectl delete ns load-test
-              cilium connectivity test --connect-timeout 4s --request-timeout 30s
+              cilium connectivity test --connect-timeout 4s --request-timeout 30s --test '!pod-to-pod-encryption,!node-to-node-encryption'
```
vipul-21, line 162: Can we add a comment specifying from which cli version this is added from.
And what cilium agent it supports. We need to keep a track of it once we are on higher versions of cilium i guess.

vipul-21 (approved):

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
jpayne3506: I can add a comment `Enable these tests when encryption feature is enabled` , but the reason why the tests were originally disabled was because we have encryption disabled. Upstream Cilium CLI update changed it to where only if the cluster is a single node cluster do these tests not run. IMO as long as we have encryption disabled we should not run these tests and explicitly defining it as such will ensure future changes to this test do not impact us.

PR #2392, opened 2023-11-15
use renamewithnetsh for nc rename

dejandundjerski:
**Reason for Change**:
Use netsh when creating NC instead of powershell to improve overall create time.

### DIFF
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -48,7 +48,10 @@ type CNSConfig struct {
 	WatchPods                   bool
 	EnableAsyncPodDelete        bool
 	AsyncPodDeletePath          string
+	EnableRenameWithNetSh       bool
 }
+
+var CnsConfig CNSConfig
 
 type TelemetrySettings struct {
 	// Flag to disable the telemetry.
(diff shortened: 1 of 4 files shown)


tamilmani1989 (approved):

ramiro-gamarra (changes requested):
No change needed per se, but need to clarify what the impact of this change is since it will be default for all CNS users.

cns/networkcontainers/networkcontainers_windows.go
```diff
@@ -160,6 +161,8 @@ func createOrUpdateWithOperation(adapterName string, ipConfig cns.IPConfiguratio
 		"true",
 		"/weakhostreceive",
 		"true",
+		"/renamewithnetsh",
+		fmt.Sprintf("%t", configuration.CnsConfig.EnableRenameWithNetSh),
```
ramiro-gamarra, line 165: Should not be consumed from a global. It should be passed as a parameter (like `setWeakHost`)

cns/configuration/configuration.go
```diff
@@ -48,8 +48,11 @@ type CNSConfig struct {
 	WatchPods                   bool
 	EnableAsyncPodDelete        bool
 	AsyncPodDeletePath          string
+	EnableRenameWithNetSh       bool
 }
 
+var CnsConfig CNSConfig
```
ramiro-gamarra, line 54: no globals please.

ramiro-gamarra (changes requested):

PR #2403, opened 2023-11-18
feat: multitenant-transparent-vlan conflist generator for CNS

pjohnst5:
This is a CNI conflist generator, for the multitenant transparent-vlan mode

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -38,6 +38,7 @@ import (
 const (
 	dockerNetworkOption = "com.docker.network.generic"
 	OpModeTransparent   = "transparent"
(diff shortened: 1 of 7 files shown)


pjohnst5:
Actually may not need this, can use the built-in multitenancy-transparent-vlan conflist already in the new CNI image:
https://github.com/Azure/azure-container-networking/blob/b30fee0e0f73127246d4ddf351e535bde5d7d02c/cni/linux.Dockerfile#L21C158-L21C158

Will test it out and see

pjohnst5:
Yeah we don't need this

PR #2412, opened 2023-11-22
ci: Adding the Vnet Scale Integration Tests to the E2E pipeline 

nairashu:
**Reason for Change**:
Added the Vnet Scale Integration Test to the E2E Pipeline and set it up with Kubernetes version 1.28 as the feature is only supported with CNS 1.5x and above which are correlated versions.

**Issue Fixed**:
N/A

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/hack/aks/Makefile
+++ b/hack/aks/Makefile
@@ -77,6 +77,11 @@ swift-net-up: ## Create vnet, nodenet and podnet subnets
 	$(AZCLI) network vnet subnet create -g $(GROUP) --vnet-name $(VNET) --name nodenet --address-prefixes 10.240.0.0/16 -o none
 	$(AZCLI) network vnet subnet create -g $(GROUP) --vnet-name $(VNET) --name podnet --address-prefixes 10.241.0.0/16 -o none
 
+vnetscale-swift-net-up: ## Create vnet, nodenet and podnet subnets for vnet scale
+	$(AZCLI) network vnet create -g $(GROUP) -l $(REGION) --name $(VNET) --address-prefixes 10.0.0.0/8 -o none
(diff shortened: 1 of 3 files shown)


hack/aks/Makefile
```diff
@@ -212,6 +217,71 @@ swift-up: rg-up swift-net-up ## Bring up a SWIFT AzCNI cluster
 		--yes
 	@$(MAKE) set-kubeconf
 
+vnetscale-swift-byocni-up: rg-up vnetscale-swift-net-up ## Bring up a Vnet Scale SWIFT BYO CNI cluster
```
sheylatrudo, line 220: We might want a warning or something that this isn't enabled automatically except for specific environments.

nairashu: Added a comment to reference the requirements to enable this feature and create Vnet Scale clusters

aggarwal0009 (approved):
LGTM

hack/aks/Makefile
```diff
@@ -77,6 +77,11 @@ swift-net-up: ## Create vnet, nodenet and podnet subnets
 	$(AZCLI) network vnet subnet create -g $(GROUP) --vnet-name $(VNET) --name nodenet --address-prefixes 10.240.0.0/16 -o none
 	$(AZCLI) network vnet subnet create -g $(GROUP) --vnet-name $(VNET) --name podnet --address-prefixes 10.241.0.0/16 -o none
```
aggarwal0009, line 79: nit:  prob add a comment that podsubnet size here leads to cluster creation in vnetscale mode?

nairashu: Not needed out here as this will change and subnet size is only for private preview. Also folks hardly ever update this piece.

rayaisaiah (approved):

PR #2415, opened 2023-11-27
only bind to interfaces with private IPs

paulyufan2:
**Reason for Change**:

This PR is to solve the security issue that:
CNS main service is bound to localhost and doesn't need to change. but it is additionally binding the metrics service on *:10092; In AKS, we already bind CNS to localhost 127.0.0.1; 
We need to only bind to the interfaces with private IPs

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/restserver/util.go
```diff
@@ -788,6 +789,37 @@ func (service *HTTPRestService) validateDefaultIPConfigsRequest(_ context.Contex
 	return types.Success, ""
 }
 
+func isPrivateIP(ip net.IP) bool {
```
rbtr, line 792: https://pkg.go.dev/net/netip#Addr.IsPrivate

cns/restserver/util.go
```diff
@@ -797,10 +829,16 @@ func (service *HTTPRestService) getPrimaryHostInterface(ctx context.Context) (*w
...
 		primary, err := wireserver.GetPrimaryInterfaceFromResult(res)
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to get primary interface from IMDS response")
 		}
+
+		if !isPrivateIP(net.ParseIP(primary.PrimaryIP)) {
+			return nil, errors.Wrap(err, "only private IPv4 or IPv6 address can be used")
+		}
+
 		service.state.primaryInterface = primary
```
rbtr, line 842: this is not the interface that metrics get bound to

rbtr (changes requested):
couple issue: we don't need to implement our own `IsPrivate`, it exists in the stdlib.
the host may have multiple interfaces, so we actually need to discard the public interfaces and bind the private ones to maintain parity with binding `0.0.0.0` while excluding public.
the metrics server address has not been changed here at all. we use the controller-runtime built-in prometheus metrics, we need to configure _that_ server

rbtr:
i'm thinking that actually, it isn't CNS's responsibility to determine which interfaces it should bind to (public, private, etc.). it's better behavior if those addresses are exposed via config for the operator to set, and CNS can simply respect the config (maybe with some sane default, maybe with an error if the required config is not set).

cns/restserver/util.go
```diff
@@ -797,10 +829,16 @@ func (service *HTTPRestService) getPrimaryHostInterface(ctx context.Context) (*w
...
 		}
+
 		primary, err := wireserver.GetPrimaryInterfaceFromResult(res)
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to get primary interface from IMDS response")
 		}
+
+		if !isPrivateIP(net.ParseIP(primary.PrimaryIP)) {
+			return nil, errors.Wrap(err, "only private IPv4 or IPv6 address can be used")
+		}
```
tamilmani1989, line 840: customer can create vnet in public ip space as well.. we shouldn't have this check

cns/restserver/util.go
tamilmani1989: this change is only for standalone scenario.. i guess for aks, cns listening only on localhost

tamilmani1989 (changes requested):
where in cns we bind only on VM's IP and localhost? I don't see changes there

PR #2419, opened 2023-11-29
fix: disable ipv6 on transparent vlan mode network create

QxBytes:
**Reason for Change**:

Ipv6 traffic from within the pod should not be able to reach eth0 on the vm from eth1 in the container. To enforce this, we disable ipv6 on all interfaces in the vm namespace. The vnet namespace has ipv6 forwarding disabled by default so no modification is required on that datapath.

**Issue Fixed**:

See above.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
- Network utils failure is already tested when we test "enable ipv4 forwarding"


paulyufan2 (approved):
Hi @vipul-21 

(1) Dualstack support is only for AKS overlay at the moment considered as single tenancy mode
(2) We are not considering ipv6 address support for multi nic for now.

tamilmani1989 (approved):

PR #2421, opened 2023-12-01
refactor: add receiver to iptables and create interface

QxBytes:
**Reason for Change**:

The iptables package is currently a set of functions without a receiver. Because of this, any code that uses the iptables package has a dependency on it, and proper unit tests cannot be written. This PR will add a receiver to the iptables methods and create an interface in the network folder so that in the future, unit tests can mock the iptables interface and methods that use iptables can be tested. This PR should not change the functionality of the codebase (for example, existing tests, if they use iptables, should still use a non-mocked version of the new iptables client).

**Issue Fixed**:

See above.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
- A new iptables interface is created in network_utils.go and used in the network package.
- Any creation of structs without a `New` function (ex: `Client{ }`) has been checked to ensure either an iptablesClient is passed in if the struct now requires an `iptablesClient`, or is part of testing code.
- TransparentVlanEndpointClient, Client (from snat_linux), and OVSEndpointClient are the clients that now require an iptablesClient.
- Functions that use iptables in networkutils linux now have receivers and the iptables client is passed in as the first parameter.


network/networkutils/iptables.go
```diff
@@ -0,0 +1,16 @@
+package networkutils
+
+import "github.com/Azure/azure-container-networking/iptables"
+
+type IPTablesClientInterface interface {
```
rbtr, line 5: we [don't declare interfaces like this](https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a) in Go.

TL;DR: Go has implicit interface implementation. If a type has the methods, it implements the interface. You don't need to explicitly declare that it implements the interface like in Java/C# or other languages.
Therefore, an interface can be satisfied by any type that implements all of the interface methods in its method set... Or, if we approach it from the other direction, an interface can be defined over a subset of the methods of an object(s).
Interfaces are for the _consumer_, not the producer. The consumer defines the interface that describes only the behavior that it needs, and any object which satisfies that behavioral contract can be passed in as the backing implementation of that interface.

rbtr (changes requested):

network/endpoint_linux.go
```diff
@@ -267,10 +271,10 @@ func (nw *network) deleteEndpointImpl(nl netlink.NetlinkInterface, plc platform.
 			epInfo := ep.getInfo()
 			if nw.Mode == opModeTransparentVlan {
 				logger.Info("Transparent vlan client")
-				epClient = NewTransparentVlanEndpointClient(nw, epInfo, ep.HostIfName, "", ep.VlanID, ep.LocalIP, nl, plc, nsc)
+				epClient = NewTransparentVlanEndpointClient(nw, epInfo, ep.HostIfName, "", ep.VlanID, ep.LocalIP, nl, plc, nsc, iptc)
 
 			} else {
-				epClient = NewOVSEndpointClient(nw, epInfo, ep.HostIfName, "", ep.VlanID, ep.LocalIP, nl, ovsctl.NewOvsctl(), plc)
+				epClient = NewOVSEndpointClient(nw, epInfo, ep.HostIfName, "", ep.VlanID, ep.LocalIP, nl, ovsctl.NewOvsctl(), plc, iptc)
```
vipul-21, line 277: Just to confirm we don't need it for `NewTransparentEndpointClient` ?

network/snat/snat_linux.go
```diff
@@ -46,6 +53,7 @@ type Client struct {
 	enableProxyArpOnBridge bool
 	netlink                netlink.NetlinkInterface
 	plClient               platform.ExecClient
+	IPTablesClient         ipTablesClientInterface
```
vipul-21, line 56: Why is this in capital case ?

QxBytes: The `IPTablesClient` variable is passed into `nuc.EnableIPForwarding` in endpoint_snatroute_linux.go (in the network package). Since network and snat are two different packages, in order to see the client, it needs to be exported (capitalized).

network/endpoint_linux.go
QxBytes: `transparent_endpointclient_linux.go` doesn't seem to use any iptables methods and doesn't have an iptables client in its struct so I don't believe we need to pass it in.

cni/network/invoker_cns.go
```diff
@@ -227,25 +227,26 @@ func setHostOptions(ncSubnetPrefix *net.IPNet, options map[string]interface{}, i
 	// we need to snat IMDS traffic to node IP, this sets up snat '--to'
 	snatHostIPJump := fmt.Sprintf("%s --to %s", iptables.Snat, info.hostPrimaryIP)
 
+	iptablesClient := iptables.NewClient()
```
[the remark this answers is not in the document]
QxBytes: My reasoning was that we're always going to call that `New` method from the `iptables` package (which only has one implementation inside it) so the statement would end up as `iptables.NewIPTablesClient`, and I recall some styling suggestion mentioning reducing stuttering so I thought the package name would provide sufficient context to know what type of client would be created.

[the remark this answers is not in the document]
rbtr: if the package only has a single `New`, it's fully identified by just saying `iptables.New`, and that should be preferred

network/endpoint_linux.go
rbtr: the compiler would flag this if it were a problem, i think

tamilmani1989 (changes requested):

network/iptables.go
```diff
@@ -0,0 +1,9 @@
+package network
```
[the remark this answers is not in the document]
rbtr: No, that's not how we do interfaces in Go. The interface exists to abstract the behavior that your code is using from the backing implementation of that behavior _at your usage_ of that behavior. This package only needs _this subset_ of behavior, so the interface should be defined in this package.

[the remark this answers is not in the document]
rbtr: https://go.dev/doc/effective_go#interfaces
> "If something can do _this_, it can be used _here_."

Don't predefine an interface that simply matches a type.

network/iptables.go
```diff
@@ -0,0 +1,9 @@
...
+
+type ipTablesClientInterface interface {
```
rbtr, line 3: simply call it `ipTablesClient`, we don't do [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation)

network/network_linux.go
```diff
@@ -651,10 +651,10 @@ func (nm *networkManager) disconnectExternalInterface(extIf *externalInterface,
 	logger.Info("Disconnected interface", zap.String("Name", extIf.Name))
 }
 
-func (*networkManager) addToIptables(cmds []iptables.IPTableEntry) error {
+func (nm *networkManager) addToIptables(cmds []iptables.IPTableEntry) error {
```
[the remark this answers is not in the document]
QxBytes: `RunCmd` is used in `network_linux.go` (addToIpTables)-- it takes in an `IPTableEntry` slice so I'm not quite sure how to refactor it (no method in iptables takes in an ip table entry struct).

network/snat/snat_linux.go
QxBytes: Addressed by moving `EnableIPForwarding` to snatClient since that is the only place it is used and the function calls iptables and has logic specific to snat.

iptables/iptables.go
```diff
@@ -87,8 +87,14 @@ type IPTableEntry struct {
 	Params  string
 }
 
+type Client struct{}
+
+func NewClient() *Client {
+	return &Client{}
+}
```
rbtr, line 94: there are no fields (at all) being set in this struct, or other initialization magic - we don't need a constructor and can make the struct directly `... := &iptables.Client{}`

network/snat/snat_linux.go
```diff
@@ -474,3 +488,20 @@ func (client *Client) addVlanDropRule() error {
...
 }
+
+// This function enables ip forwarding in VM and allow forwarding packets from the interface
+func (client *Client) EnableIPForwarding() error {
+	// Enable ip forwading on linux vm.
+	// sysctl -w net.ipv4.ip_forward=1
+	_, err := client.plClient.ExecuteCommand(enableIPForwardCmd)
+	if err != nil {
+		return errors.Wrap(err, "enable ipforwarding command failed")
+	}
```
rbtr, line 499: hm, I wish we didn't have this in the codebase already... I generally don't think that it's our responsibility to configure the host like this.

rbtr (approved):
some nits but mostly lgtm

vipul-21 (dismissed):

network/iptables.go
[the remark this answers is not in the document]
tamilmani1989: i guess iptables package used in other components like cns i guess.. if its just used in network package, then i agree

network/networkutils/networkutils_linux.go
```diff
@@ -41,6 +40,12 @@ const (
 
 var logger = log.CNILogger.With(zap.String("component", "net-utils"))
 
+type ipTablesClientInterface interface {
```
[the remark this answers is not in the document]
tamilmani1989: why do we define this interface in multiple place? can it be defined in one place and use across?

tamilmani1989 (dismissed):
talked with @QxBytes offline and clarified standard practice is to define interface in consumer side

rbtr (approved):

vipul-21 (approved):

PR #2422, opened 2023-12-01
feat: v2 swift ipampool 

rbtr:
**Reason for Change**:

v2 IPAM Pool Monitor introduces idempotent scaling math, migrates to an event-driven instead of polling architecture, and leverages the Pod watcher for IP demand instead of using the "free IPs in the pool" to trigger scaling.

Notably this change improves pool scaling performance from O(n) to O(1).

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**: not to be confused with SwiftV2, this is applicable to all dynamic Pod IP Swift scenarios

Design docs for reference: https://github.com/Azure/azure-container-networking/pull/2013

### DIFF
--- a/cns/api.go
+++ b/cns/api.go
@@ -50,6 +50,7 @@ type HTTPService interface {
 	GetPodIPConfigState() map[string]IPConfigurationStatus
 	MarkIPAsPendingRelease(numberToMark int) (map[string]IPConfigurationStatus, error)
 	AttachIPConfigsHandlerMiddleware(IPConfigsHandlerMiddleware)
+	MarkNIPsPendingRelease(n int) (map[string]IPConfigurationStatus, error)
 }
 
 // IPConfigsHandlerFunc
@@ -281,13 +282,13 @@ type NodeConfiguration struct {
 	NodeSubnet Subnet
 }
 
+// IpamPoolMonitorStateSnapshot struct to expose state values for IPAMPoolMonitor struct
 type IPAMPoolMonitor interface {
 	Start(ctx context.Context) error
 	Update(nnc *v1alpha.NodeNetworkConfig) error
 	GetStateSnapshot() IpamPoolMonitorStateSnapshot
 }
 
-// IpamPoolMonitorStateSnapshot struct to expose state values for IPAMPoolMonitor struct
 type IpamPoolMonitorStateSnapshot struct {
 	MinimumFreeIps           int64
 	MaximumFreeIps           int64
--- a/cns/ipampool/v2/adapter.go
+++ b/cns/ipampool/v2/adapter.go
@@ -1,0 +1,36 @@
+package v2
+
+import (
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/crd/nodenetworkconfig/api/v1alpha"
+	v1 "k8s.io/api/core/v1"
+)
+
+var _ cns.IPAMPoolMonitor = (*adapter)(nil)
+
+type adapter struct {
+	nncSink chan<- v1alpha.NodeNetworkConfig
+	*Monitor
+}
+
+func (m *Monitor) AsV1(nncSink chan<- v1alpha.NodeNetworkConfig) cns.IPAMPoolMonitor {
+	return &adapter{
+		nncSink: nncSink,
+		Monitor: m,
+	}
+}
+
+func (m *adapter) Update(nnc *v1alpha.NodeNetworkConfig) error {
+	m.nncSink <- *nnc
+	return nil
+}
+
+func (m *adapter) GetStateSnapshot() cns.IpamPoolMonitorStateSnapshot {
+	return cns.IpamPoolMonitorStateSnapshot{}
+}
+
+func PodIPDemandListener(ch chan<- int) func([]v1.Pod) {
+	return func(pods []v1.Pod) {
+		ch <- len(pods)
+	}
+}
--- a/cns/ipampool/v2/monitor.go
+++ b/cns/ipampool/v2/monitor.go
@@ -1,0 +1,167 @@
+package v2
+
+import (
+	"context"
+	"math"
+	"sync"
+
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/crd/clustersubnetstate/api/v1alpha1"
+	"github.com/Azure/azure-container-networking/crd/nodenetworkconfig/api/v1alpha"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+)
+
+const (
+	// DefaultMaxIPs default maximum allocatable IPs on a k8s Node.
+	DefaultMaxIPs = 250
+	// fieldManager is the field manager used when patching the NodeNetworkConfig.
+	fieldManager = "azure-cns"
+)
+
+type nodeNetworkConfigSpecUpdater interface {
+	PatchSpec(context.Context, *v1alpha.NodeNetworkConfigSpec, string) (*v1alpha.NodeNetworkConfig, error)
+}
+
+type ipStateStore interface {
+	GetPendingReleaseIPConfigs() []cns.IPConfigurationStatus
+	MarkNIPsPendingRelease(n int) (map[string]cns.IPConfigurationStatus, error)
+}
+
+type scaler struct {
+	batch     int64
+	buffer    float64
+	exhausted bool
+	max       int64
+}
+
+type Monitor struct {
+	z            *zap.Logger
+	scaler       scaler
+	nnccli       nodeNetworkConfigSpecUpdater
+	store        ipStateStore
+	demand       int64
+	request      int64
+	demandSource <-chan int
+	cssSource    <-chan v1alpha1.ClusterSubnetState
+	nncSource    <-chan v1alpha.NodeNetworkConfig
+	started      chan interface{}
+	once         sync.Once
+}
+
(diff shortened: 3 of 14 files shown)


cns/ipampool/v2/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+		case demand := <-pm.demandSource: // updated demand for IPs, recalculate request
+			pm.demand = int64(demand)
+			pm.z.Info("demand update", zap.Int64("demand", pm.demand))
+		case css := <-pm.cssSource: // received an updated ClusterSubnetState, recalculate request
+			pm.scaler.exhausted = css.Status.Exhausted
+			pm.z.Info("exhaustion update", zap.Bool("exhausted", pm.scaler.exhausted))
+		case nnc := <-pm.nncSource: // received a new NodeNetworkConfig, extract the data from it and recalculate request
+			pm.scaler.max = int64(math.Min(float64(nnc.Status.Scaler.MaxIPCount), DefaultMaxIPs))
+			pm.scaler.batch = int64(math.Min(math.Max(float64(nnc.Status.Scaler.BatchSize), 1), float64(pm.scaler.max)))
+			pm.scaler.buffer = math.Abs(float64(nnc.Status.Scaler.RequestThresholdPercent)) / 100 //nolint:gomnd // it's a percentage
```
nddq, line 83: break these parts down a bit and add more comments for readability?

nddq (commented):
Left some comments, I'm wondering if we could try and do some stress tests on an actual cluster (or you have done it already 🙂)

cns/ipampool/v2/adapter.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+var _ cns.IPAMPoolMonitor = (*adapter)(nil)
+
+type adapter struct {
+	nncSink chan<- v1alpha.NodeNetworkConfig
+	*Monitor
+}
+
+func (m *Monitor) AsV1(nncSink chan<- v1alpha.NodeNetworkConfig) cns.IPAMPoolMonitor {
```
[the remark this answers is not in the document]
rbtr: > Can it just implement the monitor interface

I specifically didn't want to do that. That architecture was created backwards and that interface is too broad.
 
In considering the long-term design, there is no reason for an interface to exist at all. The Monitor is a self-contained component that nothing else needs to directly communicate with. Thus the architecture of the v2 Monitor is totally decoupled from the rest of CNS - Pod events are pushed to an event stream (a channel), and the Monitor processes them asynchronously. No interface needs to exist, because nobody should call the Monitor synchronously - in other words data flow is unidirectional from "the thing that knows the scaling demand" (Pod Watcher) in to "the thing scales the IP pool" (Monitor), and that communication is the use case for a channel, not an interface.

rbtr:
> I'm wondering if we could try and do some stress tests on an actual cluster (or you have done it already 🙂)

I don't have numbers, but I validated that it works as expected and reaches target state faster for big swings in Pod scheduling 🙂
The really interesting results will be testing this is combination with @behzad-mir's stateless CNI changes. I expect the time-to-Pod-ready for big Pod scale-ups to be much much lower.

cns/api.go
```diff
@@ -281,13 +282,13 @@ type NodeConfiguration struct {
 	NodeSubnet Subnet
 }
 
+// IpamPoolMonitorStateSnapshot struct to expose state values for IPAMPoolMonitor struct
```
nddq, line 285: wrong place for this comment?

nddq (approved):
LGTM 🚀

cns/ipampool/v2/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+const (
+	// DefaultMaxIPs default maximum allocatable IPs on a k8s Node.
+	DefaultMaxIPs = 250
+	// fieldManager is the field manager used when patching the NodeNetworkConfig.
+	fieldManager = "azure-cns"
+)
+
+type nodeNetworkConfigSpecUpdater interface {
+	PatchSpec(context.Context, *v1alpha.NodeNetworkConfigSpec, string) (*v1alpha.NodeNetworkConfig, error)
+}
```
thatmattlong, line 24: probably out of scope of this PR, but I would like to get us to a point where the ipampool doesn't know/care that we talk through kubernetes CRDs and this interface becomes something like an `IPRequester` which doesn't `PatchSpec` but instead `Do/Make/Apply` a domain type `Request`, and the `kubecontroller` package does the translation of the domain type to the k8s type

rbtr: moving in that direction, but the change to extract all of the data out of the NNC to feed the Monitor directly would be pretty invasive to the NNC reconciler and I'm not ready to do that in this change

cns/ipampool/v2/monitor.go
```diff
@@ -0,0 +1,167 @@
...
+		select {
+		case <-ctx.Done(): // calling context has closed, we'll exit.
+			return errors.Wrap(ctx.Err(), "pool monitor context closed")
+		case demand := <-pm.demandSource: // updated demand for IPs, recalculate request
+			pm.demand = int64(demand)
+			pm.z.Info("demand update", zap.Int64("demand", pm.demand))
+		case css := <-pm.cssSource: // received an updated ClusterSubnetState, recalculate request
+			pm.scaler.exhausted = css.Status.Exhausted
+			pm.z.Info("exhaustion update", zap.Bool("exhausted", pm.scaler.exhausted))
+		case nnc := <-pm.nncSource: // received a new NodeNetworkConfig, extract the data from it and recalculate request
```
thatmattlong, line 80: along similar lines as my comment above, I'd eventually like to get us to a point where we're not operating on NNC specs/statuses in this package but instead the nncSource contains some domain type `IPState` instead of a `v1alpha.NodeNetworkConfigStatus`

cns/kubecontroller/pod/reconciler.go
```diff
@@ -2,79 +2,99 @@ package pod
...
+	z              *zap.Logger
+	cli            cli
+	reconcileFuncs []reconcile.Func
 }
 
-type PodWatcher struct {
-	cli            podcli
-	listOpt        client.ListOption
-	ReconcileFuncs []reconcile.Func
+func New(z *zap.Logger) *watcher { //nolint:revive // force usage of new by keeping the struct private
```
thatmattlong, line 28: I think it's fine to leave the struct exported so long as there is a good comment indicating that it should be constructed with `New`. However the flip side is that I think we should only export things when necessary, so if there is no reason to export, then it's fine like this

cns/kubecontroller/pod/reconciler.go
```diff
@@ -2,79 +2,99 @@ package pod
...
 }
 
-func (p *PodWatcher) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	for _, f := range p.ReconcileFuncs {
-		if _, err := f(ctx, req); err != nil {
-			return reconcile.Result{}, errors.Wrap(err, "failed to reconcile")
+func (p *watcher) Reconcile(ctx context.Context, req reconcile.Request) (ctrl.Result, error) {
+	for _, f := range p.reconcileFuncs {
+		if res, err := f(ctx, req); !res.IsZero() || err != nil {
+			return res, errors.Wrap(err, "failed to reconcile")
```
thatmattlong, line 43: I personally haven't found a good use for a non-empty `reconcile.Result`, but in my mind the intended usage is to indicate "everything was fine, but we want to reconcile this again later". I wonder if it makes sense here to return on the first non-empty `reconcile.Result` (current implementation) or allow all the funcs to run as long as there is no error, saving all the `reconcile.Result`s along the way and returning a non-zero one should one exist. Then you'd kind of have to decide which non-empty one to return which isn't ideal, but if they are indented to be used how I think, I think it doesn't make sense to return early unless `err` is `nil`.

cns/restserver/ipam.go
```diff
@@ -437,6 +438,65 @@ func (service *HTTPRestService) MarkIPAsPendingRelease(totalIpsToRelease int) (m
...
+
+	// if we can release the requested quantity, return the IPs
+	if n <= 0 {
+		maps.Copy(pendingProgrammingIPs, availableIPs)
+		return pendingProgrammingIPs, nil
+	}
+
+	// else revert changes
+	for uuid, ipConfig := range pendingProgrammingIPs { //nolint:gocritic // intentional value copy
+		_, _ = service.updateIPConfigState(uuid, types.PendingProgramming, ipConfig.PodInfo)
```
thatmattlong, line 492: might be worth leaving a comment why this can't fail and why it's safe to ignore the returned `error`

cns/ipampool/v2/monitor.go
rbtr: will do once I can delete the v1 Monitor...until then it complicates the NNC reconciler too much

thatmattlong (approved):

cns/kubecontroller/pod/reconciler.go
rbtr: Poka-yoke. If I let you construct the struct directly and you do not know that you must pass a logger reference, everything will NPE.

cns/kubecontroller/pod/reconciler.go
rbtr: hm, I have also never used a non-zero reconcile.Result. I wrote this thinking that you would use it _with_ an error to indicate to not retry immediately, but at a certain point in time in the future. Your interpretation seems more correct...

cns/restserver/ipam.go
rbtr: I'm not confident that it _is_ safe. But we don't really have a choice; If we get an error here it's because an IPID that we ***just changed while holding the state Mutex lock*** is no longer in our state. Maybe we should panic, because that probably indicates some cosmic ray/bit flip event has corrupted our internal state and we don't have any chance of resuming normal operation.

rbtr:
going to move this one forward, and consider the feedback while running it for real to refine it.

PR #2426, opened 2023-12-06
fix: use `MkdirAll` in fsnotify

tyler-lloyd:
**Reason for Change**:

On Windows, the path /var/run/azure-vnet does not exist so calling `Mkdir('/var/run/azure-vnet/deleteIDs'` will fail because the parent folder doesn't exist. Switch to MkdirAll so it creates any necessary parent folders.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


timraymond (approved):

rbtr:
i think we want to leave this as `MkDir` so that we can error out if the volume mount is wrong/the path doesn't exist - this should be a terminal error to CNS.
the goroutine leak that prompted this PR was fixed in https://github.com/Azure/azure-container-networking/pull/2427

PR #2427, opened 2023-12-06
fix: add contexts to clean up fsnotify goroutines

rbtr:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/fsnotify/fsnotify.go
```diff
@@ -164,8 +166,12 @@ func (w *watcher) readFS() error {
 // WatchFS starts the filesystem watcher to handle async Pod deletes.
 // Blocks until the context is closed; returns underlying fsnotify errors
 // if something goes fatally wrong.
-func (w *watcher) Start(ctx context.Context) error {
+func (w *watcher) Start(c context.Context) error {
+	ctx, cancel := context.WithCancel(c)
+	defer cancel()
+
 	errs := make(chan error)
```
ramiro-gamarra, line 173: probably need to buffer this channel to 2. even if the context is scoped to here and `watchPendingDelete` and `watchFS` respect it, either/both may return something (could even be `nil` error) after `Start` returns, but they won't be able to send since nothing will be consuming and the goroutine leaks.

rbtr: what do you think about the update to errgroup pattern?

cns/fsnotify/fsnotify.go
```diff
@@ -115,6 +118,24 @@ func (w *watcher) watchFS(ctx context.Context) error {
...
+	// List the directory and creates synthetic events for any existing items.
+	w.log.Info("listing directory", zap.String("path", w.path))
+	dirContents, err := os.ReadDir(w.path)
+	if err != nil {
+		w.log.Error("error reading deleteID directory", zap.String("path", w.path), zap.Error(err))
+		return errors.Wrapf(err, "failed to read %s", w.path)
+	}
+	if len(dirContents) == 0 {
+		w.log.Info("no missed deletes found")
+		return nil
```
ramiro-gamarra, line 130: I don't think you want to return here right? since the files may be created after this.

cns/fsnotify/fsnotify.go
```diff
@@ -139,51 +160,19 @@ func (w *watcher) watchFS(ctx context.Context) error {
...
-		errs <- w.watchFS(ctx)
-	}(errs)
-
-	// Read the directory to enqueue any missed deletes that are already present on-disk.
-	if err := w.readFS(); err != nil {
-		return err
-	}
-
+	g.Go(func() error { return w.watchFS(groupCtx) })
 	// block until one of the goroutines returns an error
```
ramiro-gamarra, line 174: nit: this comment might now be better with the errgroup semantics:
```
// the first error from the errgroup will trigger context cancellation for other goroutines in the group.
// this will block until all goroutines complete and return the first error.
```

rbtr: updated

cns/fsnotify/fsnotify.go
```diff
@@ -27,17 +29,18 @@ type watcher struct {
 }
 
 // Create the AsyncDelete watcher.
-func New(cli releaseIPsClient, path string, logger *zap.Logger) *watcher { //nolint
+func New(cli releaseIPsClient, path string, logger *zap.Logger) (*watcher, error) { //nolint
 	// Add directory where intended deletes are kept
-	if err := os.Mkdir(path, 0o755); err != nil { //nolint
+	if err := os.Mkdir(path, 0o755); err != nil && !errors.Is(err, fs.ErrExist) { //nolint
```
camrynl, line 34: Will the missing mount be added later or will we go with mkdirall here?

rbtr: I don't think we should mkdirall, since that will hide the problem of not mounting the right path or passing the wrong path in via config

cns/fsnotify/fsnotify.go
```diff
@@ -115,6 +118,23 @@ func (w *watcher) watchFS(ctx context.Context) error {
 	if err != nil {
 		w.log.Error("failed to add path to fsnotify watcher", zap.String("path", w.path), zap.Error(err))
 	}
```
rbtr, line 120: @camrynl is this error intentionally not returned here? it seems like we shouldn't continue if we fail to `Add`

ramiro-gamarra (approved):

cns/fsnotify/fsnotify.go
camrynl: I don't recall, but it would make sense to return here

camrynl (approved):

cns/fsnotify/fsnotify.go
```diff
@@ -139,51 +160,20 @@ func (w *watcher) watchFS(ctx context.Context) error {
...
-
-// WatchFS starts the filesystem watcher to handle async Pod deletes.
+// Start starts the filesystem watcher to handle async Pod deletes.
 // Blocks until the context is closed; returns underlying fsnotify errors
 // if something goes fatally wrong.
-func (w *watcher) Start(ctx context.Context) error {
-	errs := make(chan error)
+func (w *watcher) Start(c context.Context) error {
+	ctx, cancel := context.WithCancel(c)
+	defer cancel()
```
rbtr, line 168: this is a redundant context but can be removed later

cns/fsnotify/fsnotify.go
```diff
@@ -114,7 +117,25 @@ func (w *watcher) watchFS(ctx context.Context) error {
 	err = watcher.Add(w.path)
```
tyler-lloyd, line 117: maybe add a comment here that it's critical the watcher starts here before we do the one time read otherwise it could be possible to miss one if the watcher is started after the read.

cns/service/main.go
```diff
@@ -826,15 +826,19 @@ func main() {
...
+				}
 				if err := w.Start(rootCtx); err != nil {
 					z.Error("failed to start fsnotify watcher, will retry", zap.Error(err))
-					time.Sleep(time.Minute)
-					continue
+					return errors.Wrap(err, "failed to start fsnotify watcher, will retry")
 				}
-			}
+				return nil
+			}, retry.DelayType(retry.BackOffDelay), retry.Attempts(0), retry.Context(rootCtx))
```
tyler-lloyd, line 841: > 4.0.0
> infinity retry is possible by set Attempts(0) by PR

going to steal this for aks retry library

tyler-lloyd: maybe put a comment here that's what Attempts(0) does?

tyler-lloyd (approved):

cns/service/main.go
rbtr: commented

ramiro-gamarra (approved):

PR #2439, opened 2023-12-15
update stale workflow

rbtr:
**Reason for Change**:

lowers threshold for stale to 2 weeks
lowers threshold for close to 1 week
deletes branches for closed stale PRs

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -8,6 +8,7 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     permissions:
+      contents: write
       issues: write
       pull-requests: write
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yaml
```diff
@@ -17,11 +17,12 @@ jobs:
           ascending: true
           close-issue-message: 'Issue closed due to inactivity.'
           close-pr-message: 'Pull request closed due to inactivity.'
-          days-before-close: 14
-          days-before-stale: 60
-          exempt-issue-labels: 'blocked,must,should,keep'
+          days-before-close: 7
+          days-before-stale: 14
+          delete-branch: true
```
ramiro-gamarra, line 22: Since the new workflow uses branches from this repo instead of forks, not sure if we want to delete the branch. With forking I always have a remote with my branch, but now all my branches are going to be here. Let me know if am misunderstanding this line.

rbtr: Your understanding is correct. This was not previously enabled because it couldn't delete forked branches anyway.
I think deleting the branch if a PR is closed for being stale is appropriate. It will be 3 weeks of total inactivity - after 2 weeks, the stale label is added, and anyone can remove that to preserve the PR. There's definitely the potential to lose a branch, but there needs to be motivation to not abandon PRs.

ramiro-gamarra: Are stale branches that much of an issue? I definitely agree with closing PRs to eliminate noise, but they can always be reactivated.

rbtr: I think we could similarly argue that the official Org repo should be for actively contributing and collaborating and stale branches are also noise and don't belong there. It's straightforward enough to re-push a branch from your local, or to work primarily in a fork and only push to the upstream remote when you're ready to actually open the PR and follow through all the way to merging it.

ramiro-gamarra: It seems that even if branches are deleted, there is a restore option in the PR as well (although not sure how long that remains available after delete). Even if the branch cannot be restored, commits should still be discoverable via the closed PRs, so work won't be totally lost, which is my main concern. I think I'm ok to try this change and see if we have any issues in the next few months. It seems you will still need to add one more permission for this to work though: https://github.com/actions/stale?tab=readme-ov-file#delete-branch

rbtr: added `contents: write`

ramiro-gamarra (approved):

PR #2442, opened 2023-12-15
cleanup redundant logs in cns savestate-file, createorUpdateNetworkCo…

kmurudi:
…ntainer

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/util.go
+++ b/cns/restserver/util.go
@@ -49,21 +49,17 @@ func (service *HTTPRestService) removeNetworkInfo(networkName string) {
 
 // saveState writes CNS state to persistent store.
 func (service *HTTPRestService) saveState() error {
-	logger.Printf("[Azure CNS] saveState")
-
 	// Skip if a store is not provided.
 	if service.store == nil {
-		logger.Printf("[Azure CNS]  store not initialized.")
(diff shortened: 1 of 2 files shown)


rbtr (approved):

cns/restserver/util.go
```diff
@@ -49,21 +49,17 @@ func (service *HTTPRestService) removeNetworkInfo(networkName string) {
...
 
 	// Update time stamp.
 	service.state.TimeStamp = time.Now()
 	err := service.store.Write(storeKey, &service.state)
-	if err == nil {
-		logger.Printf("[Azure CNS]  State saved successfully.\n")
-	} else {
-		logger.Errorf("[Azure CNS]  Failed to save state., err:%v\n", err)
+	if err != nil {
+		logger.Errorf("[Azure CNS] Failed to save state, err: %v", err)
```
rbtr, line 62: maybe we don't need to log an error in this method if we're returning it and the caller can log it instead?

kmurudi: for this method saveState() there are quite a few usages (~13) & all of them do not log the error currently, may require some work there probably we can address in next PR

PR #2444, opened 2023-12-18
Enable CNS swiftv2 path for SF/Standalone scenario to program secondary nics

kmurudi:
**Reason for Change**:
L1VH/Singularity & Swiftv2 need to program secondary nics for creating pod via containerd. Changes have been supported for AKS pod scenario, but this PR adds support for windows Service fabric/standalone scenario as it gets the data from DNC goal state.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/NetworkContainerContract.go
```diff
@@ -455,7 +455,9 @@ type GetNetworkContainerResponse struct {
 type PodIpInfo struct {
 	PodIPConfig                     IPSubnet
 	NetworkContainerPrimaryIPConfig IPConfiguration
+	NetworkContainerID              string
```
kmurudi, line 458: remove this, use macaddress string in getNetworkName()

cns/wireserver/net.go
```diff
@@ -49,6 +54,49 @@ func GetPrimaryInterfaceFromResult(res *GetInterfacesResult) (*InterfaceInfo, er
...
+			continue
+		}
+
+		// skip if no subnets
+		if len(i.IPSubnet) == 0 {
+			continue
+		}
+
+		logger.Printf("i.MacAddress is %s", strings.Split(i.MacAddress, ":"))
+		newMacAddress := regexp.MustCompile(`[^a-zA-Z0-9 ]+`).ReplaceAllString(macAddress, "")
```
paulyufan2, line 71: let's use a better way to handle the newMacAddress

cns/wireserver/net.go
```diff
@@ -49,6 +54,49 @@ func GetPrimaryInterfaceFromResult(res *GetInterfacesResult) (*InterfaceInfo, er
...
+
+			return &InterfaceInfo{
+				Subnet:       s.Prefix,
+				IsPrimary:    false,
+				Gateway:      gw.String(),
+				SecondaryIPs: secondaryIPs,
+			}, nil
+		}
+	}
+	return nil, ErrNoSecondaryInterfaces
```
paulyufan2, line 97: since we will get only one secondary interface at a time as we discussed, so let's rename ErrNoSecondaryInterfaces to ErrNoSecondaryInterface

cns/wireserver/net.go
```diff
@@ -1,14 +1,19 @@
...
 
 	"github.com/pkg/errors"
 )
 
 var (
-	// ErrNoPrimaryInterface indicates the wireserver respnose does not have a primary interface indicated.
+	// ErrNoPrimaryInterface indicates the wireserver response does not have a primary interface indicated.
 	ErrNoPrimaryInterface = errors.New("no primary interface found")
+	// ErrNoSecondaryInterfaces indicates the wireserver response does not have secondary interfaces on the node
+	ErrNoSecondaryInterfaces = errors.New("no secondary interfaces found")
```
paulyufan2, line 16: rename this ErrorName

cns/restserver/ipam.go
```diff
@@ -26,7 +35,13 @@ var (
 
 // requestIPConfigHandlerHelper validates the request, assign IPs and return the IPConfigs
 func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context, ipconfigsRequest cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
-	// For SWIFT v2 scenario, the validator function will also modify the ipconfigsRequest.
+	cmdLineConfigPath := common.GetArg(common.OptCNSConfigPath).(string)
+	cnsconfig, err := configuration.ReadConfig(cmdLineConfigPath)
+	var swiftv2sf bool
```
paulyufan2, line 40: can we add some comments here to explain more about swiftv2sf scenario, for example how we can get podInfo with this flag; also I remember we will need to support the K8S, so we will still use this flag I believe?

cns/restserver/ipam.go
```diff
@@ -125,6 +166,67 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+	service.getNetworkContainerByOrchestratorContext(w, req)
+
+	err = json.NewDecoder(w.Body).Decode(&resp)
+	if err != nil {
+		return cns.PodIpInfo{}, fmt.Errorf("decoding response: %w", err)
+	}
+
+	// Check if the ncstate/ipconfig ready. If one of the fields is empty, return error
+	if resp.IPConfiguration.IPSubnet.IPAddress == "" || resp.NetworkInterfaceInfo.MACAddress == "" || resp.NetworkContainerID == "" || resp.IPConfiguration.GatewayIPAddress == "" {
+		return cns.PodIpInfo{}, errors.New("ipconfig is empty for given nc")
```
paulyufan2, line 198: let's log error here which field is empty for debugging

cns/wireserver/net.go
kmurudi: added a separate function below for this

cns/restserver/ipam.go
kmurudi: added

cns/middlewares/SFSwiftV2.go
```diff
@@ -0,0 +1,125 @@
...
+func (m *SFSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req *cns.IPConfigsRequest) (podInfo cns.PodInfo, respCode types.ResponseCode, message string) {
+	// Retrieve the pod from the cluster
+	podInfo, err := cns.UnmarshalPodInfo(req.OrchestratorContext)
+	if err != nil {
+		errBuf := errors.Wrapf(err, "failed to unmarshalling pod info from ipconfigs request %+v", req)
+		return nil, types.UnexpectedError, errBuf.Error()
+	}
+	logger.Printf("[SWIFTv2Middleware] validate ipconfigs request for pod %s", podInfo.Name())
+
+	req.SecondaryInterfacesExist = true
```
nddq, line 86: How will this validation function check whether a pod is SWIFT v2 or not? As things stand right now, all pod ipconfigs requests will come through this validation function and will be tag as secondary interface exist for them.

cns/restserver/ipam.go
```diff
@@ -176,6 +171,9 @@ func (service *HTTPRestService) requestIPConfigsHandler(w http.ResponseWriter, r
 		// Wrap the default datapath handlers with the middleware
 		wrappedHandler := service.IPConfigsHandlerMiddleware.IPConfigsRequestHandlerWrapper(service.requestIPConfigHandlerHelper, service.releaseIPConfigHandlerHelper)
 		ipConfigsResp, err = wrappedHandler(r.Context(), ipconfigsRequest)
+		if ipconfigsRequest.AddInterfacesDataToResponse {
```
nddq, line 174: Currently we are passing in ipconfigsRequest by value, so whatever changes we made to ipconfigsRequest within the wrapped handler will not stick when we exited to here. You can change the `IPConfigsHandlerFunc` function signature to taking in `IPConfigsRequest` as a pointer instead.

cns/restserver/ipam.go
```diff
@@ -192,6 +190,34 @@ func (service *HTTPRestService) requestIPConfigsHandler(w http.ResponseWriter, r
 	logger.ResponseEx(service.Name+operationName, ipconfigsRequest, ipConfigsResp, ipConfigsResp.Response.ReturnCode, err)
 }
 
+func (service *HTTPRestService) updatePodInfoWithInterfaces(ipconfigResponse *cns.IPConfigsResponse) (*cns.IPConfigsResponse, error) {
+	for _, podIpInfo := range ipconfigResponse.PodIPInfo {
+		hostPrimaryInterface, err := service.getPrimaryHostInterface(context.TODO())
```
nddq, line 195: You should pass down the context from `requestIPConfigsHandler`

cns/restserver/ipam.go
```diff
@@ -192,6 +190,34 @@ func (service *HTTPRestService) requestIPConfigsHandler(w http.ResponseWriter, r
...
+		if err != nil {
+			return &cns.IPConfigsResponse{}, err
+		}
+
+		hostSecondaryInterface, err := service.getSecondaryHostInterface(context.TODO(), podIpInfo.MacAddress)
```
nddq, line 200: Same here

cns/wireserver/net.go
```diff
@@ -49,6 +52,47 @@ func GetPrimaryInterfaceFromResult(res *GetInterfacesResult) (*InterfaceInfo, er
...
+			s := i.IPSubnet[0]
+			gw, err := calculateGatewayIP(s.Prefix)
+			if err != nil {
+				return nil, err
+			}
+
+			secondaryIP := ""
+			for _, ip := range s.IPAddress {
+				if !ip.IsPrimary {
+					secondaryIP = ip.Address
```
nddq, line 79: nit: we are expecting just one secondary IP here right? I think you can add a `break` here

cns/wireserver/net.go
```diff
@@ -49,6 +52,47 @@ func GetPrimaryInterfaceFromResult(res *GetInterfacesResult) (*InterfaceInfo, er
...
+			}
+			var secondaryIPs []string
+			secondaryIPs = append(secondaryIPs, secondaryIP)
+
+			return &InterfaceInfo{
+				Subnet:       s.Prefix,
+				IsPrimary:    false,
+				Gateway:      gw.String(),
+				SecondaryIPs: secondaryIPs,
+			}, nil
```
nddq, line 90: nit: maybe something like this would looks cleaner
```go
return &InterfaceInfo{
		Subnet:       s.Prefix,
		IsPrimary:    false,
		Gateway:      gw.String(),
		SecondaryIPs: []string{secondaryIP},
	}, nil
```

cns/wireserver/net.go
```diff
@@ -79,3 +123,10 @@ func calculateGatewayIP(cidr string) (net.IP, error) {
 	}
 	return gw, nil
 }
+
+func macAddressesEqual(macAddress1, macAddress2 string) bool {
+	macAddress1 = strings.ToLower(strings.ReplaceAll(macAddress1, ":", ""))
+	macAddress2 = strings.ToLower(strings.ReplaceAll(macAddress2, ":", ""))
+
+	return macAddress1 == macAddress2
+}
```
nddq, line 132: nit: in what scenario would directly comparing two MAC addresses string doesn't work here? Why do we need to remove all the colons and convert them all to lowercases?

cns/middlewares/SFSwiftV2.go
```diff
@@ -0,0 +1,125 @@
...
+	// which can be done only by restserver service function, hence setting this flag to do so in ipam.go
+	req.AddInterfacesDataToResponse = true
+
+	logger.Printf("[SWIFTv2Middleware] pod %s has secondary interface : %v", podInfo.Name(), req.SecondaryInterfacesExist)
+	// retrieve podinfo from orchestrator context
+	return podInfo, types.Success, ""
+}
+
+// getIPConfig returns the pod's SWIFT V2 IP configuration.
+func (m *SFSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodInfo) (cns.PodIpInfo, error) {
```
nddq, line 97: can you go over the lint errors in this function?

cns/restserver/ipam.go
kmurudi: updated to pass`r.Context()` from `requestIPConfigsHandler`

cns/restserver/ipam.go
kmurudi: updated

cns/wireserver/net.go
kmurudi: cc @paulyufan2 I've updated to break after this, let me know

cns/wireserver/net.go
kmurudi: the format for both is a bit different, one we get from cns goal state & another one is what we get from wireserver client

cns/middlewares/SFSwiftV2.go
kmurudi: if we enable cns config as swiftv2, it will program only for secondary nics i.e. no request for primary nic is expected, that's what I understand from swiftv2 scenario for us, I'll double check this

cns/restserver/ipam.go
kmurudi: good catch, I've modified it for k8s middleware too

cns/service/main.go
```diff
@@ -789,6 +789,13 @@ func main() {
 		if platform.HasMellanoxAdapter() {
 			go platform.MonitorAndSetMellanoxRegKeyPriorityVLANTag(rootCtx, cnsconfig.MellanoxMonitorIntervalSecs)
 		}
+		// if swiftv2 scenario is enabled, we need to initialize the Service Fabric (standalone) swiftv2 middleware to process IP configs requests
+		if cnsconfig.EnableSwiftV2 {
+			if cnsconfig.SWIFTV2Mode == configuration.SFSWIFTV2 {
```
nddq, line 794: we can get rid of this config IMO, since each SWIFT v2 scenario is operated in 2 different channel mode without conflict.

cns/service/main.go
```diff
@@ -1381,18 +1388,12 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
-		switch cnsconfig.SWIFTV2Mode {
-		case configuration.K8sSWIFTV2:
-			swiftV2Middleware = &middlewares.K8sSWIFTv2Middleware{Cli: manager.GetClient()}
-		case configuration.SFSWIFTV2:
-		default:
-			// default to K8s middleware for now, in a later changes we where start to pass in
-			// SWIFT v2 mode in CNS config, this should throw an error if the mode is not set.
-			swiftV2Middleware = &middlewares.K8sSWIFTv2Middleware{Cli: manager.GetClient()}
+		// here for AKS(K8s) swiftv2 middleware to process IP configs requests
+		if cnsconfig.SWIFTV2Mode == configuration.K8sSWIFTV2 {
```
nddq, line 1392: Same here, we only need to know that SWIFT v2 is enabled or not.

cns/restserver/ipam.go
```diff
@@ -29,9 +29,9 @@ var (
 )
 
 // requestIPConfigHandlerHelper validates the request, assign IPs and return the IPConfigs
-func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context, ipconfigsRequest cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
+func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context, ipconfigsRequest *cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
 	// For SWIFT v2 scenario, the validator function will also modify the ipconfigsRequest.
```
nddq, line 33: You can remove this comment since this is no longer true

cns/middlewares/SFSwiftV2.go
nddq, line 1: Still seeing linting errors in this file, can you resolve it?

cns/service/main.go
kmurudi: ack removed this config altogether, could be added if required later

cns/restserver/ipam_test.go
```diff
@@ -1717,3 +1727,170 @@ func TestIPAMGetK8sSWIFTv2IPFailure(t *testing.T) {
 		t.Fatal("Expected available ips to be 2 since we expect the IP to not be assigned")
 	}
 }
+
+//
+//func TestIPAMGetSFSWIFTv2IPSuccess(t *testing.T) {
+//	req := cns.IPConfigsRequest{
+//		PodInterfaceID:   testPod1Info.InterfaceID(),
+//		InfraContainerID: testPod1Info.InfraContainerID(),
```
kmurudi, line 1735: remove these UTs

cns/restserver/k8sSwiftV2.go
```diff
@@ -1,4 +1,4 @@
-package middlewares
```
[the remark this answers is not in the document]
kmurudi: yes, due to dependency cycle in ipam_test

cns/restserver/ipam.go
```diff
@@ -40,11 +39,9 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
 			},
 		}, errors.New("failed to validate ip config request")
 	}
-
 	// record a pod requesting an IP
 	service.podsPendingIPAssignment.Push(podInfo.Key())
-
-	podIPInfo, err := requestIPConfigsHelper(service, ipconfigsRequest)
+	podIPInfo, err := requestIPConfigsHelper(service, *ipconfigsRequest)
```
kmurudi, line 44: @nddq I'm not sure why we need to pass context here according to linter error? it's not used in the functions mentioned

nddq: ideally, we should pass the context all the way down but right now it is not necessary within this PR's scope so you can just add `// nolint:contextcheck` next to it

cns/restserver/ipam.go
```diff
@@ -176,9 +170,12 @@ func (service *HTTPRestService) requestIPConfigsHandler(w http.ResponseWriter, r
 	if service.IPConfigsHandlerMiddleware != nil {
 		// Wrap the default datapath handlers with the middleware
 		wrappedHandler := service.IPConfigsHandlerMiddleware.IPConfigsRequestHandlerWrapper(service.requestIPConfigHandlerHelper, service.releaseIPConfigHandlerHelper)
-		ipConfigsResp, err = wrappedHandler(r.Context(), ipconfigsRequest)
+		ipConfigsResp, err = wrappedHandler(r.Context(), &ipconfigsRequest)
+		if ipconfigsRequest.AddInterfacesDataToResponse {
```
nddq, line 174: now that i have taken a better look, we could have had this field `AddInterfacesDataToResponse` in the `ipConfigsResp` instead of the request. That way we would not have to modify anything related to `ipConfigsRequest`. Plus I think it makes sense that it should be the response's responsibility and not the request to tell CNS to add interfaces data to itself

cns/restserver/SFSwiftV2.go
```diff
@@ -0,0 +1,115 @@
...
+	// which can be done only by restserver service function, hence setting this flag to do so in ipam.go
+	req.AddInterfacesDataToResponse = true
+
+	logger.Printf("[SWIFTv2Middleware] pod %s has secondary interface : %v", podInfo.Name(), req.SecondaryInterfacesExist)
+	// retrieve podinfo from orchestrator context
+	return podInfo, types.Success, ""
+}
+
+// getIPConfig returns the pod's SWIFT V2 IP configuration.
+func (m *SFSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodInfo) (cns.PodIpInfo, error) {
```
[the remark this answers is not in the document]
kmurudi: i'm returning back to caller the error at L54

cns/restserver/SFSwiftV2.go
```diff
@@ -0,0 +1,115 @@
...
+			}, errors.Wrapf(err, "failed to get SWIFTv2 IP config : %v", req)
+		}
+		ipConfigsResp.PodIPInfo = append(ipConfigsResp.PodIPInfo, SWIFTv2PodIPInfo)
+		return ipConfigsResp, nil
+	}
+}
+
+// validateIPConfigsRequest validates if pod orchestrator context is unmarshalled & sets secondary interfaces true
+// nolint
+func (m *SFSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req *cns.IPConfigsRequest) (podInfo cns.PodInfo, respCode types.ResponseCode, message string) {
```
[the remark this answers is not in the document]
kmurudi: yes, I only have to verify the unmarshal

cns/restserver/ipam.go
kmurudi: should I introduce a field in PodIpInfo in that case? it doesn't seem right to add this to the current struct ipConfigsResponse, does not fit well the usages & pattern & type of fields in ipconfigsResponse currently

nddq: yes lets try that

cns/restserver/ipam_test.go
nddq: will we need these UTs in the future? if not then we should remove it instead of commenting them out

kmurudi: no i do plan to remove them

cns/restserver/ipam.go
paulyufan2: yes we should only have one single podIPInfo at a time

cns/restserver/ipam.go
```diff
@@ -176,9 +171,12 @@ func (service *HTTPRestService) requestIPConfigsHandler(w http.ResponseWriter, r
 	if service.IPConfigsHandlerMiddleware != nil {
 		// Wrap the default datapath handlers with the middleware
 		wrappedHandler := service.IPConfigsHandlerMiddleware.IPConfigsRequestHandlerWrapper(service.requestIPConfigHandlerHelper, service.releaseIPConfigHandlerHelper)
-		ipConfigsResp, err = wrappedHandler(r.Context(), ipconfigsRequest)
+		ipConfigsResp, err = wrappedHandler(r.Context(), &ipconfigsRequest)
+		if reflect.DeepEqual(ipConfigsResp.PodIPInfo[0].HostPrimaryIPInfo, cns.HostIPInfo{}) || reflect.DeepEqual(ipConfigsResp.PodIPInfo[0].HostSecondaryIPInfo, cns.HostIPInfo{}) {
```
nddq, line 175: I would advise against using `reflect` here. Can we just check whether `PrimaryIP` has been set or not?

cns/api.go
```diff
@@ -54,7 +55,7 @@ type HTTPService interface {
 }
 
 // IPConfigsHandlerFunc
-type IPConfigsHandlerFunc func(context.Context, IPConfigsRequest) (*IPConfigsResponse, error)
+type IPConfigsHandlerFunc func(context.Context, *IPConfigsRequest) (*IPConfigsResponse, error)
```
nddq, line 58: Now that we are not modifying the request anymore, I think we should revert this function back to taking `IPConfigsRequest` as a value and not a pointer.

cns/restserver/ipam.go
```diff
@@ -177,6 +171,9 @@ func (service *HTTPRestService) requestIPConfigsHandler(w http.ResponseWriter, r
 		// Wrap the default datapath handlers with the middleware
 		wrappedHandler := service.IPConfigsHandlerMiddleware.IPConfigsRequestHandlerWrapper(service.requestIPConfigHandlerHelper, service.releaseIPConfigHandlerHelper)
 		ipConfigsResp, err = wrappedHandler(r.Context(), ipconfigsRequest)
+		if ipConfigsResp.PodIPInfo[0].AddInterfacesDataToPodInfo {
```
nddq, line 174: move this to after we check err instead. If there is err then this would fail

cns/configuration/configuration.go
```diff
@@ -36,7 +36,7 @@ type CNSConfig struct {
 	EnableIPAMv2                bool
 	EnablePprof                 bool
 	EnableSubnetScarcity        bool
-	EnableSwiftV2               bool
```
nddq, line 39: Double check if we are still referencing this elsewhere.

kmurudi: checked by doing a search project-wide, we don't use it anywhere else

nddq (approved):

PR #2465, opened 2024-01-03
L1VH CNI first version

paulyufan2:
**Reason for Change**:

This implementation is to support swiftv2 L1VH CNI feature:
CNI receives goal state from CNS and create linux/windows pod on Windows VM/Node in swiftv2 L1VH path;  
The container pod is using secondary interface with its mac address and NIC type(Delegated NIC or Accelnet) to be created; 
We support up to 8 pods to be created by using different interfaces along with different hnsNetwork; CNI supports hnsNetwork creation(when pod is added to the VM) and deletion(when pod is removed from VM) to prevent leakage; 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/keyvault/azsecrets v0.12.0
 	github.com/Masterminds/semver v1.5.0
 	github.com/Microsoft/go-winio v0.6.1
-	github.com/Microsoft/hcsshim v0.11.4
+	github.com/Microsoft/hcsshim v0.12.0-rc.2
```
jpayne3506, line 11: What caused this update?

paulyufan2: we need to fetch this hcsshim version for:
hcnNetwork.Flags = hcn.DisableHostPort

jpayne3506: Considering it is a release candidate update. Can you also run the CNI Release Test Pipeline? Just want to ensure there are not additional changes that are breaking.

cni/network/invoker_cns.go
```diff
@@ -171,6 +171,10 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 				numInterfacesWithDefaultRoutes++
 			}
 
+			info.hostSubnet = response.PodIPInfo[i].HostSecondaryIPInfo.Subnet
```
kmurudi, line 174: L168 - add other nic types too for same case

network/endpoint_windows.go
```diff
@@ -406,6 +420,19 @@ func (nw *network) newEndpointImplHnsV2(cli apipaClient, epInfo *EndpointInfo) (
...
+	for i, ipconfig := range epInfo.IPAddresses {
+		ipconfigs[i] = &IPConfig{Address: ipconfig}
+	}
+
+	ep.SecondaryInterfaces[ep.Id] = &InterfaceInfo{
+		Name:       ep.Id,
+		MacAddress: ep.MacAddress,
+		IPConfigs:  ipconfigs,
+		Routes:     ep.Routes,
+		NICType:    cns.DelegatedVMNIC,
```
kmurudi, line 433: let's not hardcode this nic type here, fetch from input to function, epInfo.NICtype

cni/network/invoker_cns.go
```diff
@@ -450,6 +455,13 @@ func configureSecondaryAddResult(info *IPResultInfo, addResult *IPAMAddResult, p
 		return errors.Wrap(err, "Invalid mac address")
 	}
 
+	_, hostIPNet, err := net.ParseCIDR(info.hostSubnet)
```
tamilmani1989, line 458: why this is required?

cni/network/network.go
```diff
@@ -45,6 +45,9 @@ const (
 	ipv6FullMask          = 128
 )
 
+// Secondary interface NIC types
+var secondaryInterfaceNICType = []string{cns.BackendNICNC, string(cns.DelegatedVMNIC)}
```
tamilmani1989, line 49: we should have enum for nic types.. we should resuse from cns package

cni/network/network.go
```diff
@@ -362,12 +386,18 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		telemetry.SendCNIMetric(&cniMetric, plugin.tb)
 
 		// Add Interfaces to result.
-		defaultCniResult := convertInterfaceInfoToCniResult(ipamAddResult.defaultInterfaceInfo, args.IfName)
+		cniResult := cniTypesCurr.Result{}
+		// Add secondaryInterface info to result if it exists; otherwise add default primary interface info to result
```
tamilmani1989, line 390: can it be other way.. add secondary interface to result if default interface does not exist

cni/network/network.go
```diff
@@ -460,6 +494,14 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 				zap.Any("results", ipamAddResult))
 			return plugin.Errorf(errMsg)
 		}
+	} else if !nwCfg.MultiTenancy && nwCfg.IPAM.Type == network.AzureCNS {
```
tamilmani1989, line 497: wanted to understand why we need this change

tamilmani1989 (changes requested):

go.mod
tamilmani1989: we cannot merge with release candidate version.. have to get official version released

cni/network/invoker_cns.go
```diff
@@ -165,12 +165,17 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
+			// only handling single v4 PodIPInfo for DelegatedVMNIC and BackendNIC at the moment, will have to update once v6 gets added
 			if !info.skipDefaultRoutes {
 				numInterfacesWithDefaultRoutes++
 			}
 
+			// Add secondary interface info from podIPInfo to ipamAddResult
+			info.hostSubnet = response.PodIPInfo[i].HostSecondaryIPInfo.Subnet
+			info.hostPrimaryIP = response.PodIPInfo[i].HostSecondaryIPInfo.PrimaryIP
+			info.hostGateway = response.PodIPInfo[i].HostSecondaryIPInfo.Gateway
+
```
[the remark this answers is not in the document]
paulyufan2: it gets secondary interface info from HostSecondaryIPInfo

cni/network/invoker_cns.go
paulyufan2: because we do need "hostSubnetPrefix" to find master interface if you check findMasterInterface()

cni/network/network.go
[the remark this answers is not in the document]
paulyufan2: I think it's because they are defined in different types on CNS codes, I can update CNS code

cni/network/network.go
paulyufan2: in design, we have both default and secondary interface info in ipamAddResult;

cni/network/network.go
paulyufan2: because for swiftv2 path, the nwconfig multitenancy must not be set to true which was confused before; and the nwcfg ipam type should be azureCNS

cni/network/network.go
```diff
@@ -607,13 +639,50 @@ func (plugin *NetPlugin) createNetworkInternal(
...
-		return nwInfo, err
+		// check if masterInterface is in SecondaryInterfaces map
+		// if not, then no master interface is found
+		if SecondaryInterfaces != nil {
+			for name, secondaryInterface := range SecondaryInterfaces {
+				if secondaryInterface.macAddress == macAddress {
+					masterIfName = name
+				}
+			}
+		} else {
```
tamilmani1989, line 674: should return error if it fails to find interface by mac address

cni/network/network.go
```diff
@@ -607,13 +639,50 @@ func (plugin *NetPlugin) createNetworkInternal(
...
+		}
 	}
-	logger.Info("Found master interface", zap.String("ifname", masterIfName))
+
+	var secondaryInterfaceProperty SecondaryInterfaceProperty
+	secondaryInterfaceProperty.ipNet = ipnet.String()
+	secondaryInterfaceProperty.macAddress = macAddress
+
+	SecondaryInterfaces = make(map[string]SecondaryInterfaceProperty)
+	SecondaryInterfaces[masterIfName] = secondaryInterfaceProperty
```
tamilmani1989, line 685: this code is not required

cni/network/network.go
```diff
@@ -622,14 +691,12 @@ func (plugin *NetPlugin) createNetworkInternal(
 		return nwInfo, err
 	}
 
-	nwDNSInfo, err := getNetworkDNSSettings(ipamAddConfig.nwCfg, ipamAddResult.defaultInterfaceInfo.DNS)
+	nwDNSInfo, err := getNetworkDNSSettings(ipamAddConfig.nwCfg, interfaceInfo.DNS)
```
tamilmani1989, line 694: default has precedence.. use secondary only if there is no primary

cni/network/network.go
```diff
@@ -622,14 +691,12 @@ func (plugin *NetPlugin) createNetworkInternal(
...
 	if err != nil {
 		err = plugin.Errorf("Failed to getDNSSettings: %v", err)
 		return nwInfo, err
 	}
```
tamilmani1989, line 699: revert this change

cni/network/network.go
```diff
@@ -666,7 +733,14 @@ func (plugin *NetPlugin) createNetworkInternal(
 
 // construct network info with ipv4/ipv6 subnets
 func addSubnetToNetworkInfo(ipamAddResult IPAMAddResult, nwInfo *network.NetworkInfo) error {
```
tamilmani1989, line 735: lets pass interface info to `addSubnetToNetworkInfo`

cni/network/network.go
```diff
@@ -801,13 +875,15 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
 
 		epInfos = append(epInfos,
 			&network.EndpointInfo{
+				IfName:            epInfo.IfName,
```
tamilmani1989, line 878: check if this is set when parsing secondaryInterfaceInfo

cni/network/network.go
```diff
@@ -801,13 +875,15 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
...
 				ContainerID:       epInfo.ContainerID,
 				NetNsPath:         epInfo.NetNsPath,
 				IPAddresses:       addresses,
 				Routes:            secondaryCniResult.Routes,
 				MacAddress:        secondaryCniResult.MacAddress,
 				NICType:           secondaryCniResult.NICType,
 				SkipDefaultRoutes: secondaryCniResult.SkipDefaultRoutes,
+				Policies:          opt.policies,
```
tamilmani1989, line 886: can you check if we are setting this

cni/network/network.go
```diff
@@ -1008,7 +1085,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 
 	// Loop through all the networks that are created for the given Netns. In case of multi-nic scenario ( currently supported
 	// scenario is dual-nic ), single container may have endpoints created in multiple networks. As all the endpoints are
-	// deleted, getNetworkName will return error of the type NetworkNotFoundError which will result in nil error as compliance
+	// deleted,  will return error of the type NetworkNotFoundError which will result in nil error as compliance
```
tamilmani1989, line 1088: revert it

cni/network/network.go
```diff
@@ -1017,10 +1094,21 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 		numEndpointsToDelete = plugin.nm.GetNumEndpointsByContainerID(args.ContainerID)
 	}
 
+	endpointID := plugin.nm.GetEndpointID(args.ContainerID, args.IfName)
+
+	// Get secondary interface info from CNI state file and save them to ipamAddResult in CNI Delete call
+	secondaryInterfaceInfo, err := plugin.nm.GetNetworkSecondaryInterfaceInfo(endpointID)
+	if err != nil {
+		logger.Error("Failed to get secondaryInterfaceInfo", zap.Error(err))
```
tamilmani1989, line 1102: change level to info
log endpointID

cni/network/network.go
```diff
@@ -1017,10 +1094,21 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+	}
+
+	ipamAddResult.secondaryInterfacesInfo = []network.InterfaceInfo{}
```
tamilmani1989, line 1105: if err == nil {
    then set this
}

cni/network/network.go
```diff
@@ -1017,10 +1094,21 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 		numEndpointsToDelete = plugin.nm.GetNumEndpointsByContainerID(args.ContainerID)
 	}
 
+	endpointID := plugin.nm.GetEndpointID(args.ContainerID, args.IfName)
+
+	// Get secondary interface info from CNI state file and save them to ipamAddResult in CNI Delete call
+	secondaryInterfaceInfo, err := plugin.nm.GetNetworkSecondaryInterfaceInfo(endpointID)
```
tamilmani1989, line 1100: should not be specfic call for each interface, instead it should fetch all interface associated with that containerID

cni/network/network.go
```diff
@@ -1032,6 +1120,16 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 			err = plugin.Errorf("Failed to extract network name from network config. error: %v", err)
 			return err
 		}
+
+		// cleanup interfaces usage map for swiftv2 L1VH and delete hnsNetwork associated with this networkID
+		if &ipamAddResult != nil && ipamAddResult.secondaryInterfacesInfo != nil {
```
tamilmani1989, line 1125: change ipamAddResult to pointer otherwise

cni/network/network.go
```diff
@@ -1049,7 +1147,6 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 			}
 		}
```
tamilmani1989, line 1149: revert this change

tamilmani1989 (changes requested):

cni/network/network.go
tamilmani1989: will address in stateless cni

cni/network/network.go
paulyufan2: this line is on line 1082, just moved to the above

cni/network/network.go
paulyufan2: just remembered this code should be here because when a pod is created, then we cannot see that interface from ipconfig; so we must have a map to record which interface is being used, otherwise we cannot find master interface by macAddress because ipconfig will return nil to us;

cni/network/network.go
```diff
@@ -78,6 +78,14 @@ type NetPlugin struct {
 	multitenancyClient MultitenancyClient
 }
 
+// save secondary interface property
+// key is secondary interface name and value is its macAddress
+var SecondaryInterfaces map[string]SecondaryInterfaceProperty
+
+type SecondaryInterfaceProperty struct {
+	macAddress string
+}
```
tamilmani1989, line 87: if its not required, lets remove

cni/network/network.go
```diff
@@ -217,6 +226,14 @@ func (plugin *NetPlugin) findMasterInterface(nwCfg *cni.NetworkConfig, subnetPre
 	subnetPrefixString := subnetPrefix.String()
 	interfaces, _ := net.Interfaces()
 	for _, iface := range interfaces {
+		if subnetPrefixString == "" {
```
tamilmani1989, line 229: add a new function findMasterInterfaceByMac

cni/network/network.go
```diff
@@ -362,12 +387,17 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		telemetry.SendCNIMetric(&cniMetric, plugin.tb)
 
 		// Add Interfaces to result.
-		defaultCniResult := convertInterfaceInfoToCniResult(ipamAddResult.defaultInterfaceInfo, args.IfName)
+		cniResult := &cniTypesCurr.Result{}
+		cniResult = convertInterfaceInfoToCniResult(ipamAddResult.defaultInterfaceInfo, args.IfName)
```
tamilmani1989, line 391: can you add UT for this?

cni/network/network.go
```diff
@@ -362,12 +387,17 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+		// if defaultInterfaceInfo does not exist, then try to use secondaryInterfaceInfo
+		if cniResult == nil && hasSecondaryInterface(ipamAddResult) {
```
tamilmani1989, line 393: should append to result.Interface since its an array and not overwrite it

cni/network/network.go
```diff
@@ -434,6 +465,9 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		return fmt.Errorf("failed to create cns client with error: %w", err)
 	}
 
+	options := make(map[string]any)
+	ipamAddConfig := IPAMAddConfig{nwCfg: nwCfg, args: args, options: options}
```
tamilmani1989, line 469: if not required, lets remove it

tamilmani1989 (changes requested):

cni/network/network.go
kmurudi: how would be branch to this vs the default function used by other paths?

cni/network/network.go
paulyufan2: we get it from conflist file with outbound NAT policies that I confirmed

cni/network/network.go
paulyufan2: yes

cni/network/network.go
paulyufan2: I checked with Kshitija and we only add one NC at a time and the secondary interface is always the first element of the secondaryInterfacesInfo

go.mod
jpayne3506: This is in rc status as they are making 0.12.X dependent on containerd 2.0, which has not been released yet.

PR #2485, opened 2024-01-08
ci: Improve Validate package

jpayne3506:
**Reason for Change**:

- Adds windows CNIv2 state validation
- Moves `cnsCacheStateFileIps()` to `validate.go` as it is OS agnostic
- Adds in reporting for leaked ips that exist on state files when `compareIPs()` fails due to size difference between actual IPs on the cluster and what exists in the state file. Old functionality reported an entire map of IPs from both current IPs and state file IPs. 
- Adds in reporting when `ExecCmdPod()` has a 0 byte response
- Fixes `validateRestartNetwork()` to work in hybrid clusters and only execute on linux nodes
- Moves `RestartKubeProxyService()` to internal `kubernetes` package as it is a wrapper function for `ExecCmdPod()` and does not validate anything

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/integration/datapath/datapath_windows_test.go
+++ b/test/integration/datapath/datapath_windows_test.go
@@ -15,13 +15,10 @@ import (
 )
 
 const (
-	WindowsDeployYamlPath   = "../manifests/datapath/windows-deployment.yaml"
-	podLabelKey             = "app"
-	podCount                = 2
-	nodepoolKey             = "agentpool"
-	privilegedDaemonSetPath = "../manifests/load/privileged-daemonset-windows.yaml"
-	privilegedLabelSelector = "app=privileged-daemonset"
-	privilegedNamespace     = "kube-system"
+	WindowsDeployYamlPath = "../manifests/datapath/windows-deployment.yaml"
+	podLabelKey           = "app"
+	podCount              = 2
+	nodepoolKey           = "agentpool"
 )
 
 var (
@@ -59,15 +56,15 @@ func setupWindowsEnvironment(t *testing.T) {
 	clientset := kubernetes.MustGetClientset()
 
 	if *restartKubeproxy {
-		privilegedDaemonSet := kubernetes.MustParseDaemonSet(privilegedDaemonSetPath)
-		daemonsetClient := clientset.AppsV1().DaemonSets(privilegedNamespace)
+		privilegedDaemonSet := kubernetes.MustParseDaemonSet(kubernetes.PrivilegedDaemonSetPath)
+		daemonsetClient := clientset.AppsV1().DaemonSets(kubernetes.PrivilegedNamespace)
 		kubernetes.MustCreateDaemonset(ctx, daemonsetClient, privilegedDaemonSet)
 
 		// Ensures that pods have been replaced if test is re-run after failure
-		if err := kubernetes.WaitForPodDaemonset(ctx, clientset, privilegedNamespace, privilegedDaemonSet.Name, privilegedLabelSelector); err != nil {
+		if err := kubernetes.WaitForPodDaemonset(ctx, clientset, kubernetes.PrivilegedNamespace, privilegedDaemonSet.Name, kubernetes.PrivilegedLabelSelector); err != nil {
 			require.NoError(t, err)
 		}
-		err := kubernetes.RestartKubeProxyService(ctx, clientset, privilegedNamespace, privilegedLabelSelector, restConfig)
+		err := kubernetes.RestartKubeProxyService(ctx, clientset, kubernetes.PrivilegedNamespace, kubernetes.PrivilegedLabelSelector, restConfig)
 		require.NoError(t, err)
 	}
 
--- a/test/integration/datapath/datapath_windows_test.go
+++ b/test/integration/datapath/datapath_windows_test.go
@@ -10,16 +10,18 @@ import (
 
 	"github.com/Azure/azure-container-networking/test/internal/datapath"
 	"github.com/Azure/azure-container-networking/test/internal/kubernetes"
-	"github.com/Azure/azure-container-networking/test/validate"
 	"github.com/stretchr/testify/require"
 	apiv1 "k8s.io/api/core/v1"
 )
 
 const (
-	WindowsDeployYamlPath = "../manifests/datapath/windows-deployment.yaml"
-	podLabelKey           = "app"
-	podCount              = 2
-	nodepoolKey           = "agentpool"
+	WindowsDeployYamlPath   = "../manifests/datapath/windows-deployment.yaml"
+	podLabelKey             = "app"
+	podCount                = 2
+	nodepoolKey             = "agentpool"
+	privilegedDaemonSetPath = "../manifests/load/privileged-daemonset-windows.yaml"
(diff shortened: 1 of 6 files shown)


test/validate/windows_validate.go
```diff
@@ -13,12 +13,22 @@ import (
...
+	cnsWinLabelSelector = "k8s-app=azure-cns-win"
+)
+
 var (
 	hnsEndPointCmd      = []string{"powershell", "-c", "Get-HnsEndpoint | ConvertTo-Json"}
 	hnsNetworkCmd       = []string{"powershell", "-c", "Get-HnsNetwork | ConvertTo-Json"}
 	azureVnetCmd        = []string{"powershell", "-c", "cat ../../k/azure-vnet.json"}
 	azureVnetIpamCmd    = []string{"powershell", "-c", "cat ../../k/azure-vnet-ipam.json"}
 	restartKubeProxyCmd = []string{"powershell", "Restart-service", "kubeproxy"}
+	cnsWinLocalCacheCmd = []string{
```
vipul-21, line 26: nit: It gives only the assigned IPs ? Not the full local cache as we have a filter. Would good to rename it to something that captures that.

test/validate/windows_validate.go
```diff
@@ -100,6 +112,10 @@ func hnsStateFileIps(result []byte) (map[string]string, error) {
 	for _, v := range hnsResult {
 		if !v.IsRemoteEndpoint {
 			hnsPodIps[v.IPAddress.String()] = v.MacAddress
+			if v.IPv6Address.String() != "<nil>" {
+				hnsPodIps[v.IPv6Address.String()] = v.MacAddress
```
vipul-21, line 116: Is this for dualstack windows ? Or is it some error we found ?

test/validate/utils.go
```diff
@@ -11,14 +11,16 @@ import (
...
-	if len(expected) != len(actual) {
-		return errors.Errorf("len of expected IPs != len of actual IPs, expected: %+v, actual: %+v", expected, actual)
-	}
+	expectedLen := len(expected)
 
 	for _, ip := range actual {
 		if _, ok := expected[ip]; !ok {
 			return errors.Errorf("actual ip %s is unexpected, expected: %+v, actual: %+v", ip, expected, actual)
 		}
+		delete(expected, ip)
```
vipul-21, line 20: Not able to understand why `expected` is being manipulated here ?

jpayne3506: Manipulating expected like this allows us to get any leaked IP(s) (as it will be the only IPs left in the map after all actual IPs are deleted) if the total number of IPs dont match between actual and expected.

Per PR description - "Adds in reporting for leaked ips that exist on state files when compareIPs() fails due to size difference between actual IPs on the cluster and what exists in the state file. Old functionality reported an entire map of IPs from both current IPs and state file IPs"

test/validate/windows_validate.go
jpayne3506: This is for dualstack windows.

test/validate/windows_validate.go
jpayne3506: Matched the pre-existing naming within `linux_validate.go`, so I will change there as well.

Would `*CachedAssignedIPStateCmd` work?

test/validate/utils.go
vipul-21: I see, can we add that to the error message that these Ips might be leaked.

vipul-21 (approved):

paulyufan2 (dismissed):

test/validate/windows_validate.go
```diff
@@ -13,12 +14,22 @@ import (
...
+	cnsWinLabelSelector = "k8s-app=azure-cns-win"
+)
+
 var (
-	hnsEndPointCmd      = []string{"powershell", "-c", "Get-HnsEndpoint | ConvertTo-Json"}
-	hnsNetworkCmd       = []string{"powershell", "-c", "Get-HnsNetwork | ConvertTo-Json"}
-	azureVnetCmd        = []string{"powershell", "-c", "cat ../../k/azure-vnet.json"}
-	azureVnetIpamCmd    = []string{"powershell", "-c", "cat ../../k/azure-vnet-ipam.json"}
-	restartKubeProxyCmd = []string{"powershell", "Restart-service", "kubeproxy"}
+	hnsEndPointCmd                 = []string{"powershell", "-c", "Get-HnsEndpoint | ConvertTo-Json"}
```
vipul-21, line 22: Did we try using `-AsArray` for the ConvertTo-Json cmd ? Will that help us to avoid a switch case for object and array while reading the data.
ref: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/convertto-json?view=powershell-7.4#-asarray

jpayne3506: It is currently not possible as our daemonset is using Powershell 5.1. When trying to update the image being used by the ds for a newer version of Powershell it would still use 5.1. Further effort is required to understand what is preventing the use of 7.*

vipul-21: Not a blocking comment. So lgtm.

vipul-21 (dismissed):

test/integration/datapath/datapath_windows_test.go
```diff
@@ -10,16 +10,18 @@ import (
...
-	podLabelKey           = "app"
-	podCount              = 2
-	nodepoolKey           = "agentpool"
+	WindowsDeployYamlPath   = "../manifests/datapath/windows-deployment.yaml"
+	podLabelKey             = "app"
+	podCount                = 2
+	nodepoolKey             = "agentpool"
+	privilegedDaemonSetPath = "../manifests/load/privileged-daemonset-windows.yaml"
+	privilegedLabelSelector = "app=privileged-daemonset"
+	privilegedNamespace     = "kube-system"
```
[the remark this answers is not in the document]
vipul-21: We can just declare these constants in the utils/constant file and then use where we need ?

test/validate/linux_validate.go
```diff
@@ -212,3 +200,35 @@ func azureVnetIpamStateIps(result []byte) (map[string]string, error) {
...
+		if node.Status.NodeInfo.OperatingSystem != string(corev1.Linux) {
+			continue
+		}
+		// get the privileged pod
+		pod, err := acnk8s.GetPodsByNode(ctx, v.clientset, privilegedNamespace, privilegedLabelSelector, nodes.Items[index].Name)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get privileged pod")
+		}
+
+		privilegedPod := pod.Items[0]
```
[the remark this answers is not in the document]
jpayne3506: Included other locations that did not check for empty return from `GetPodsByNode()`

test/integration/datapath/datapath_windows_test.go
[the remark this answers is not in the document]
jpayne3506: I will go ahead and move it to `utils.go`, but upon looking further there needs to be an additional PR to address the multiple locations of `const( ... )` that reference internal file locations.

vipul-21 (approved):

PR #2490, opened 2024-01-09
ci: add OS scale timeout

jpayne3506:
**Reason for Change**:

Set a standard time for `TestLoad()` per iteration for each OS type. Setting arbitrary to a default 30 limited the amount of iterations and did not take into account the performance differences between OS.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/integration/load/load_test.go
+++ b/test/integration/load/load_test.go
@@ -40,6 +40,13 @@ var noopDeploymentMap = map[string]string{
 	"linux":   manifestDir + "/noop-deployment-linux.yaml",
 }
 
+// This map is used exclusively for TestLoad. Windows is expected to take 10-15 minutes per iteration.
+// Will change this as scale testing results are verified. This will ensure we keep a standard performance metric.
+var scaleTimeoutMap = map[string]time.Duration{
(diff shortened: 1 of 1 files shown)


test/integration/load/load_test.go
```diff
@@ -40,6 +40,13 @@ var noopDeploymentMap = map[string]string{
 	"linux":   manifestDir + "/noop-deployment-linux.yaml",
 }
 
+// This map is used exclusively for TestLoad. Windows is expected to take 10-15 minutes per iteration.
+// Will change this as scale testing results are verified. This will ensure we keep a standard performance metric.
```
vipul-21, line 44: Does this we will verify after merging or before ?

jpayne3506: After, as an ongoing process. Currently the only performance metric that I know of is for windows clusters and timed accordingly. Have not tested for linux or cilium specifically.

vipul-21 (approved):

camrynl (approved):

PR #2498, opened 2024-01-10
ci: Improve CNI|NPM integration test

jpayne3506:
**Reason for Change**:

Window NPM testing takes an increased amount of time to complete then linux due to custom changes upstream. https://github.com/kubernetes/kubernetes/compare/master...huntergregory:kubernetes:sleep-before-probing

Fixed taint being left on windows nodes if job timeout occurs.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/npm/npm-cni-integration-test.yaml
+++ b/.pipelines/npm/npm-cni-integration-test.yaml
@@ -16,6 +16,7 @@ jobs:
     displayName: "NPM k8s E2E"
     dependsOn: ${{ parameters.dependsOn }}
     condition: and( and( not(canceled()), not(failed()) ), ${{ or(contains(parameters.os_version, '2022'), eq(parameters.os, 'linux') ) }} , or( contains(variables.CONTROL_SCENARIO, 'npm') , contains(variables.CONTROL_SCENARIO, 'all') ) )
+    timeoutInMinutes: 180 # This is for testing windows, may have to keep due to the 3m between the 14 tests -> results in 42m of wasted time
     pool:
       name: $(BUILD_POOL_NAME_DEFAULT)
       demands:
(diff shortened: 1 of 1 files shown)


.pipelines/npm/npm-cni-integration-test.yaml
```diff
@@ -16,6 +16,7 @@ jobs:
     displayName: "NPM k8s E2E"
     dependsOn: ${{ parameters.dependsOn }}
     condition: and( and( not(canceled()), not(failed()) ), ${{ or(contains(parameters.os_version, '2022'), eq(parameters.os, 'linux') ) }} , or( contains(variables.CONTROL_SCENARIO, 'npm') , contains(variables.CONTROL_SCENARIO, 'all') ) )
+    timeoutInMinutes: 180 # This is for testing windows, due to the 3m between the 14 tests -> results in 42m of wasted time
```
huntergregory, line 19: Noticing that that change to the k8s fork was in July fyi

If desired, we could also create another branch with a smaller timeout. I don't know why we changed to two minutes, but I feel like it's probably unnecessary at this point (Perhaps there was a latency issue in HNS that's been fixed)

huntergregory (approved):

.pipelines/npm/npm-cni-integration-test.yaml
jpayne3506: I tested with a shorter timeout from your fork and got similar results. (Roughly 20 minutes faster). I would rather keep this in and have consistent results and avoid flakes due to HNS variability.

vipul-21: Does it actually take 180 mins to complete ? Can we pull npm test in parallel to all other tests (if not already)

vipul-21 (approved):

.pipelines/npm/npm-cni-integration-test.yaml
jpayne3506: It takes roughly 75 - 105 minutes to complete... Main problem is that it takes 140-160 minutes to fail. If a job timeout happens before failure then we lose a considerable amount of results and debugging logs.

PR #2499, opened 2024-01-10
CNI must add static route for ipv6 default gateway

paulyufan2:
**Reason for Change**:

This PR is associated with ADO 26116186; 
The PR is to fix in dualstack overlay cluster windows node that in the rare cases, windows nodes are missing ipv6 default route gateway; so from CNI side, it should check if ipv6 default route is missing; if not, then do not do anything, otherwise add the default route(::/0 and nextHop is fe80::1234:5678:9abc)
We add ipv6 default route once when HNs network is created

**Issue Fixed**:

Successfully added the routes:
Active Routes:
 If Metric Network Destination      Gateway
 10    266 ::/0                     fe80::1234:5678:9abc
 10    266 ::/1                     fe80::1234:5678:9abc

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/network_windows.go
+++ b/network/network_windows.go
@@ -327,6 +327,8 @@ func (nm *networkManager) addIPv6DefaultRoute() error {
 		err error
 		out string
 	)
+
+	logger.Info("Adding default ipv6 route to windows node")
 	// the default ipv6 route is missing sometimes due to ARP issue
 	// need to add ipv6 default route if it does not exist in dualstack overlay windows node
 	cmd := fmt.Sprintf(`Get-NetAdapter | Where-Object { $_.InterfaceDescription -like 'Hyper-V*' } | Select-Object -ExpandProperty ifIndex`)
--- a/network/network_windows.go
+++ b/network/network_windows.go
@@ -329,7 +329,7 @@ func (nm *networkManager) addIPv6DefaultRoute() error {
 	)
 
 	logger.Info("Adding default ipv6 route to windows node")
-	// the default ipv6 route is missing sometimes due to ARP issue
+	// the default ipv6 route is missing sometimes
 	// need to add ipv6 default route if it does not exist in dualstack overlay windows node
 	cmd := fmt.Sprintf(`Get-NetAdapter | Where-Object { $_.InterfaceDescription -like 'Hyper-V*' } | Select-Object -ExpandProperty ifIndex`)
 	ifIndex, err := nm.plClient.ExecutePowershellCommand(cmd)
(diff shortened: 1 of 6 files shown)


network/network_windows.go
```diff
@@ -318,6 +323,36 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
 	return hcnNetwork, nil
 }
 
+func (nm *networkManager) addIPv6DefaultRoute() error {
+	var (
+		err error
+		out string
+	)
```
ramiro-gamarra, line 330: Don't declare variables ahead of time.

network/network_windows.go
```diff
@@ -318,6 +323,36 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+
+	logger.Info("Adding default ipv6 route to windows node")
```
ramiro-gamarra, line 332: Only log that the route is being added if it's missing. This would make it seem like the route is always being added on every CNI invocation.

network/network_windows.go
```diff
@@ -318,6 +323,36 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+	// the default ipv6 route is missing sometimes
+	// need to add ipv6 default route if it does not exist in dualstack overlay windows node
+	cmd := `Get-NetAdapter | Where-Object { $_.InterfaceDescription -like 'Hyper-V*' } | Select-Object -ExpandProperty ifIndex`
+	ifIndex, err := nm.plClient.ExecutePowershellCommand(cmd)
+	if err != nil {
+		return fmt.Errorf("error while executing powershell command to get ipv6 Hyper-V interface: %w", err)
```
ramiro-gamarra, line 338: use `errors.Wrapf` for consistency.

network/network_windows.go
```diff
@@ -318,6 +323,36 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+	}
+
+	getIPv6RouteCmd := fmt.Sprintf("Get-NetRoute -DestinationPrefix %s", defaultIPv6Route)
+	if out, err = nm.plClient.ExecutePowershellCommand(getIPv6RouteCmd); err != nil {
+		logger.Info("ipv6 default route is not found, adding it to the system", zap.Any("out", out))
```
ramiro-gamarra, line 343: - if `out` is a string, then you should use `zap.String`.
- use a more descriptive name than `"out"` for the log property. `"powershellOutput"` maybe.

ramiro-gamarra (changes requested):
Some comments. We also need tests added.

PR #2327, opened 2023-10-26
fix: Pipeline agent image update from 18.04 to 22.04 | AzSecPack

jpayne3506:
**Reason for Change**:

This PR is the first to migrate our agent pool to a 1ES maintained agent pool. This is primarily to unblock the ACN PR pipeline and will be improved upon. 

Future CI:
- Check for systemd-resolved not tied to ubuntu image version
- Properly assign *auth.json to skopeo commands
- Remove reinstalling dependencies from JIT provisioning script
- Remove all uses of `sudo` where applicable

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [x] relevant PR labels added

**Notes**:

PR #2506, opened 2024-01-15
fix: joining multiple dns server lists

vakalapa:
**Reason for Change**:

Earlier command was not setting DNS list but instead it was resolving the IP causing failures in CNI ADD, new change will set the DNS specifically 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -294,7 +294,14 @@ func (nm *networkManager) addDNSServers(ifName string, dnsServers []string) (str
 	case strings.HasPrefix(osVersion, Ubuntu22):
 		cmd = fmt.Sprintf("resolvectl dns %s %s", ifName, strings.Join(dnsServers, " "))
 	default:
-		cmd = fmt.Sprintf("systemd-resolve --interface %s %s", ifName, strings.Join(dnsServers, "--set-dns "))
+		serverList := ""
+		for _, server := range dnsServers {
+			if serverList == "" {
+				serverList = " --set-dns "
+			}
+			serverList = serverList + " --set-dns " + server
+		}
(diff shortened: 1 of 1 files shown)


network/network_linux.go
```diff
@@ -289,12 +289,21 @@ func (nm *networkManager) addDNSServers(ifName string, dnsServers []string) (str
 		return osVersion, err
 	}
 
+	if len(dnsServers) == 0 {
```
vakalapa, line 292: Adding this check again because of the caller behavior changes then this could be a potential point of failure.

network/network_linux.go
```diff
@@ -289,12 +289,21 @@ func (nm *networkManager) addDNSServers(ifName string, dnsServers []string) (str
...
+
 	var cmd string
 	switch {
 	case strings.HasPrefix(osVersion, Ubuntu22):
 		cmd = fmt.Sprintf("resolvectl dns %s %s", ifName, strings.Join(dnsServers, " "))
 	default:
-		cmd = fmt.Sprintf("systemd-resolve --interface %s %s", ifName, strings.Join(dnsServers, "--set-dns "))
+		serverList := ""
+		for _, server := range dnsServers {
+			serverList = serverList + " --set-dns " + server
```
anubhabMajumdar, line 304: There will be a double space, but that works
```bash
 sudo systemd-resolve --interface eth0  --set-dns 8.8.8.8 --set-dns 1.1.1.1                                                                                                                    ok | 01:29:08
 ~                                                                                                                                                                                                ok | 01:29:09 
```

anubhabMajumdar (approved):
LGTM.

jpayne3506 (approved):

jpayne3506:
fixes #2327

PR #2507, opened 2024-01-15
Adding UTs for addDNSServers

anubhabMajumdar:
**Reason for Change**:
Adding UTs for `addDNSServers` function. 

**Issue Fixed**:

**Requirements**:
Check in after https://github.com/Azure/azure-container-networking/pull/2506 is checked in.

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


rbtr (changes requested):
I don't see the need for these to be Ginkgo, can you make them plain Go tests pls

network/network_linux_test.go
```diff
@@ -0,0 +1,98 @@
...
+func TestAddDNSServers(t *testing.T) {
+	testCases := []struct {
+		desc      string
+		osversion string
+		ifName    string
+		servers   []string
+		cmd       string
+		err       error
+	}{
+		{
```
jpayne3506, line 30: Do we know of any commonly used cmd(s) that should fail, so we can have some negative tests?

jpayne3506 (approved):

network/network_linux_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/Azure/azure-container-networking/platform"
+)
+
+func TestAddDNSServersWithErr(t *testing.T) {
```
vipul-21, line 11: This checks for empty dnsServers length case right ? Ifo so, the result of the error is `nil` so not sure if `TestAddDNSServersWithErr` make sense as we don't expect an error ?

network/network_linux_test.go
```diff
@@ -0,0 +1,98 @@
...
+			desc:      "multiple server addresses",
+			osversion: "16.0.0",
+			ifName:    "azure0",
+			servers:   []string{"1.1.1.1", "bing.com"},
+			cmd:       "systemd-resolve --interface azure0 --set-dns 1.1.1.1 --set-dns bing.com",
+			err:       nil,
+		},
+	}
+	for _, tc := range testCases {
+		t.Log("Running: ", tc.desc, " for ", tc.osversion)
```
timraymond, line 80: Please use `t.Run` for this so we can get the benefits of subtests. The name of the subtest can just be `tc.desc`, or you can concatenate the `tc.osversion` as well if you want.

network/network_linux_test.go
```diff
@@ -0,0 +1,98 @@
...
+		m.SetExecCommand(func(cmd string) (string, error) {
+			return tc.osversion, nil
+		})
+		nm := &networkManager{
+			plClient: m,
+		}
+		cmd, err := nm.addDNSServers(tc.ifName, tc.servers)
+
+		if !errors.Is(err, tc.err) {
+			t.Errorf("Expected error: %v, got: %v", tc.err, err)
```
timraymond, line 92: Typically I make these Fatal, on the grounds that if an error occurred, it's usually nonsensical to try to compare anything about the results of the thing I was trying to do when the error occurred. You just end up with extra output in the test that obscures what the actual root cause of the test failure was.

timraymond:
@anubhabMajumdar The required PR in the description has merged btw. Stalebot is awake.

PR #2508, opened 2024-01-15
add ipv6 default route to dualstack windows nodes

paulyufan2:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


network/network_windows.go
```diff
@@ -347,6 +395,13 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *NetworkInfo, extIf *extern
 		logger.Info("Network with name already exists", zap.String("name", hcnNetwork.Name))
 	}
 
+	// check if ipv6 default gateway route is missing before windows endpoint creation
+	if len(nwInfo.Subnets) >= numDualStackSubnet {
+		if err = nm.addIPv6DefaultRoute(); err != nil {
+			return nil, errors.Wrapf(err, "failed to add missing ipv6 default route to windows node active and persistent store")
+		}
+	}
+
```
tamilmani1989, line 404: did we test if hns network succeeded but adding route failed. on second call everything should succeed, since we are not cleaning up hns network if adding ipv6 route fails. lets also add UT for that specific case

network/network_windows.go
```diff
@@ -347,6 +395,13 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *NetworkInfo, extIf *extern
 		logger.Info("Network with name already exists", zap.String("name", hcnNetwork.Name))
 	}
 
+	// check if ipv6 default gateway route is missing before windows endpoint creation
+	if len(nwInfo.Subnets) >= numDualStackSubnet {
```
tamilmani1989, line 399: should it be `==`

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
 	return hcnNetwork, nil
 }
 
+func (nm *networkManager) addIPv6DefaultRoute() error {
+	// add ipv6 default route if it does not exist in dualstack overlay windows node from persistentstore
+	cmd := `((Get-NetIPInterface | where InterfaceAlias  -Like "vEthernet*").IfIndex)[0]`
+	ifIndex, err := nm.plClient.ExecutePowershellCommand(cmd)
+	if err != nil {
+		return errors.Wrapf(err, "error while executing powershell command to get ipv6 Hyper-V interface")
+	}
```
tamilmani1989, line 334: what is this check for?

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+		// run powershell cmd to add ipv6 default route
+		addCmd := fmt.Sprintf("Remove-NetRoute -DestinationPrefix %s -InterfaceIndex %s -NextHop %s -confirm:$false;New-NetRoute -DestinationPrefix %s -InterfaceIndex %s -NextHop %s -confirm:$false",
+			defaultIPv6Route, ifIndex, defaultIPv6NextHop, defaultIPv6Route, ifIndex, defaultIPv6NextHop)
+
+		// if command is failed to execute, then attempt to execute extra two times
+		for i := 0; i <= addIPv6DefaultRouteAttempts; i++ {
+			if out, err := nm.plClient.ExecutePowershellCommand(addCmd); err != nil {
+				logger.Error("Failed to add ipv6 default gateway route, retrying after error", zap.Any("out", out), zap.Error(err))
+			} else {
+				// check if ipv6 default route is added to both active and persistent store
```
tamilmani1989, line 349: whats difference between active and persistent store? why it has to be in both store?

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+				if out, err := nm.plClient.ExecutePowershellCommand(getIPv6RoutePersistentCmd); err != nil {
```
tamilmani1989, line 350: why `zap.Any` isn't out string?

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+					logger.Error("default Route is not added to windows persistent store", zap.Any("out", out))
+					return errors.Wrapf(err, "Failed to add ipv6 default gateway route")
```
tamilmani1989, line 352: can it be `Failed to get ipv6 default gateway route in persistent store`. Also update logger.Error with same message?

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+				}
+				if out, err := nm.plClient.ExecutePowershellCommand(getIPv6RouteActiveCmd); err != nil {
+					logger.Error("default Route is not added to windows active store", zap.Any("out", out))
+					return errors.Wrapf(err, "Failed to add ipv6 default gateway route")
```
tamilmani1989, line 356: Failed to get ipv6 default gateway route in active store

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+				}
+			}
+		}
+	}
+
+	if err != nil {
+		return errors.Wrapf(err, "Failed to add ipv6 default gateway route")
+	}
```
tamilmani1989, line 364: what's the purpose of this check?

network/network_windows_test.go
```diff
@@ -227,3 +227,28 @@ func TestDeleteNetworkImplHnsV1WithTimeout(t *testing.T) {
...
+	}
+
+	Hnsv2 = hnswrapper.NewHnsv2wrapperFake()
+
+	_, err := nm.newNetworkImplHnsV2(nwInfo, extInterface)
+	if err != nil {
+		fmt.Printf("+%v", err)
+		t.Fatal(err)
+	}
+}
```
tamilmani1989, line 254: can you add for cases for:
1.  where hns network creation succeed and route failed and on second attempt it should succeed without creating hns network again?
2. failure cases for each error returned and if error returned is expected

tamilmani1989 (changes requested):

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
 	return hcnNetwork, nil
 }
 
+func (nm *networkManager) addIPv6DefaultRoute() error {
+	// add ipv6 default route if it does not exist in dualstack overlay windows node from persistentstore
+	cmd := `((Get-NetIPInterface | where InterfaceAlias  -Like "vEthernet*").IfIndex)[0]`
```
princepereira, line 330: Good to have: cmd -> getIpv6IfIndexCmd

network/network_windows.go
```diff
@@ -16,6 +15,7 @@ import (
 	"github.com/Microsoft/hcsshim"
 	"github.com/Microsoft/hcsshim/hcn"
 	"github.com/google/uuid"
+	"github.com/pkg/errors"
```
princepereira, line 18: Assuming this is intentionally added and not added by the formatting commands.

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+
+	getIPv6RouteActiveCmd := fmt.Sprintf("Get-NetRoute -DestinationPrefix %s", defaultIPv6Route)
+	getIPv6RoutePersistentCmd := fmt.Sprintf("Get-NetRoute -DestinationPrefix %s -PolicyStore Persistentstore", defaultIPv6Route)
+	if out, err := nm.plClient.ExecutePowershellCommand(getIPv6RoutePersistentCmd); err != nil {
+		logger.Info("ipv6 default route is not found from persistentstore, adding default ipv6 route to the windows node", zap.Any("out", out))
+		// run powershell cmd to add ipv6 default route
+		addCmd := fmt.Sprintf("Remove-NetRoute -DestinationPrefix %s -InterfaceIndex %s -NextHop %s -confirm:$false;New-NetRoute -DestinationPrefix %s -InterfaceIndex %s -NextHop %s -confirm:$false",
+			defaultIPv6Route, ifIndex, defaultIPv6NextHop, defaultIPv6Route, ifIndex, defaultIPv6NextHop)
+
+		// if command is failed to execute, then attempt to execute extra two times
```
princepereira, line 344: Mis match in comment and implementation. Comment says 2 retries. But as per the logic it is 4 retries (0, 1, 2, 3, 4).

addIPv6DefaultRouteAttempts = 3

3 retry will be good to have. But comment and implementation should be in sync.

network/network_windows.go
princepereira, line 352: Can we have the retry logic here as well.

network/network_windows.go
paulyufan2: for adding ipv6 route, we need to find interface index; so this is to check whether it's successful or not to get the interface index

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
 	return hcnNetwork, nil
 }
 
+func (nm *networkManager) addIPv6DefaultRoute() error {
+	// add ipv6 default route if it does not exist in dualstack overlay windows node from persistentstore
+	cmd := `((Get-NetIPInterface | where InterfaceAlias  -Like "vEthernet*").IfIndex)[0]`
+	ifIndex, err := nm.plClient.ExecutePowershellCommand(cmd)
+	if err != nil {
+		return errors.Wrapf(err, "error while executing powershell command to get ipv6 Hyper-V interface")
```
ramiro-gamarra, line 333: `Wrapf` is for when you want to interpolate something in the format string. Otherwise just use `Wrap`. Applies to similar usages in PR.

network/network_windows.go
ramiro-gamarra: By default, it gets added to both, but also you can't just store something in the persistent store without it being in the active store: https://learn.microsoft.com/en-us/powershell/module/nettcpip/new-netroute?view=windowsserver2022-ps#-policystore

network/network_windows.go
```diff
@@ -318,6 +325,47 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+			} else {
+				// check if ipv6 default route is added to both active and persistent store
+				if out, err := nm.plClient.ExecutePowershellCommand(getIPv6RoutePersistentCmd); err != nil {
+					logger.Error("default Route is not added to windows persistent store", zap.Any("out", out))
+					return errors.Wrapf(err, "Failed to add ipv6 default gateway route")
+				}
+				if out, err := nm.plClient.ExecutePowershellCommand(getIPv6RouteActiveCmd); err != nil {
+					logger.Error("default Route is not added to windows active store", zap.Any("out", out))
+					return errors.Wrapf(err, "Failed to add ipv6 default gateway route")
+				}
```
ramiro-gamarra, line 357: if there's no error, how are you breaking out of this loop?

ramiro-gamarra (changes requested):

network/network_windows.go
paulyufan2: we need to use the formatted errors, so it's intentionally added here

network/network_windows.go
paulyufan2: added break

network/network_windows.go
tamilmani1989: can it be only in active store? on reboot, cni has to recreate both hns network and this route.. usually everything should come clean after reboot

network/network_windows.go
```diff
@@ -318,6 +325,45 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
 	return hcnNetwork, nil
 }
 
+func (nm *networkManager) addIPv6DefaultRoute() error {
+	// add ipv6 default route if it does not exist in dualstack overlay windows node from persistentstore
+	getIpv6IfIndexCmd := `((Get-NetIPInterface | where InterfaceAlias  -Like "vEthernet*").IfIndex)[0]`
+	ifIndex, err := nm.plClient.ExecutePowershellCommand(getIpv6IfIndexCmd)
+	if err != nil {
+		return errors.Wrap(err, "error while executing powershell command to get ipv6 Hyper-V interface")
+	}
```
tamilmani1989, line 334: can you add a comment saying "getting interface index to add ipv6 default route"?

@princepereira / @paulyufan2  can there be multiple `vEthernet*` interfaces? If so, how cni know which is the right interface?

network/network_windows.go
paulyufan2: use errors.wrap() now

network/network_windows.go
paulyufan2: Prince confirmed that if cni is recreating the network on every reboot, you can keep the route only in activestore. And no need to keep in persistentstore

network/network_windows.go
paulyufan2: as discussed, I will use (Get-NetIPAddress | where IPAddress -EQ "IPV6 Management IP").InterfaceIndex to get the interfaceIndex

network/network_windows.go
tamilmani1989: yes lets do that

network/network_windows.go
paulyufan2: Or we know only have one Ethernet interface if we do not consider swiftv2 as multiple interfaces; so above command is working; otherwise we have to use above command and add ipv6 management IP from the network;

network/network_windows.go
```diff
@@ -347,6 +393,15 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *NetworkInfo, extIf *extern
 		logger.Info("Network with name already exists", zap.String("name", hcnNetwork.Name))
 	}
 
+	logger.Info("extIf is", zap.Any("extIf", extIf))
```
ramiro-gamarra, line 396: was this added for debugging? does it need to be removed?

network/network_windows.go
```diff
@@ -318,6 +325,45 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+
+		// if command is failed to execute, then attempt to execute additional two times
+		for i := 0; i <= addIPv6DefaultRouteRetryAttempts; i++ {
+			if out, err := nm.plClient.ExecutePowershellCommand(addCmd); err != nil {
+				logger.Error("Failed to add ipv6 default gateway route, retrying after error", zap.String("out", out), zap.Error(err))
+			} else {
+				// check if ipv6 default route is added to both active and persistent store
+				if out, err := nm.plClient.ExecutePowershellCommand(getIPv6RouteActiveCmd); err != nil {
+					logger.Error("Failed to get ipv6 default gateway route in active store", zap.String("out", out), zap.Error(err))
+					return errors.Wrap(err, "Failed to get ipv6 default gateway route in active store")
```
ramiro-gamarra, line 352: you are printing similar logs twice (caller prints the error already). just condense all the information you need into the error and let caller log once:
```go
return errors.Wrapf(err, "Failed to get ipv6 default gateway route in active store. powershell output: %s", out)
```
same for return below.

network/network_windows.go
paulyufan2: as discussed we only consider single NIC scenario for now and so we can use this command: `((Get-NetIPInterface | where InterfaceAlias  -Like "vEthernet*").IfIndex)[0]`

network/network_windows_test.go
paulyufan2: 1. Yes, once hns network is created successfully, then it will attempt to execution command to add ipv6 default route; it will try 3 times to add default route; if they are all failed, just log error and it won't create hnsNetwork again;

network/network_windows.go
```diff
@@ -318,6 +325,45 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+	getIpv6IfIndexCmd := `((Get-NetIPInterface | where InterfaceAlias -Like "vEthernet*").IfIndex)[0]`
+	ifIndex, err := nm.plClient.ExecutePowershellCommand(getIpv6IfIndexCmd)
+	if err != nil {
+		return errors.Wrap(err, "error while executing powershell command to get ipv6 Hyper-V interface")
+	}
+
+	getIPv6RouteActiveCmd := fmt.Sprintf("Get-NetRoute -DestinationPrefix %s", defaultIPv6Route)
+	getIPv6RoutePersistentCmd := fmt.Sprintf("Get-NetRoute -DestinationPrefix %s -PolicyStore Persistentstore", defaultIPv6Route)
+	if out, err := nm.plClient.ExecutePowershellCommand(getIPv6RoutePersistentCmd); err != nil {
+		logger.Info("ipv6 default route is not found from persistentstore, adding default ipv6 route to the windows node", zap.String("out", out))
```
ramiro-gamarra, line 341: what does `out` contain when there's an error? seems odd that there would be any value there if an error was returned. could you share an example of this log line?

network/network_windows.go
```diff
@@ -347,6 +393,16 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *NetworkInfo, extIf *extern
 		logger.Info("Network with name already exists", zap.String("name", hcnNetwork.Name))
 	}
 
+	// check if ipv6 default gateway route is missing before windows endpoint creation
+	for i := range nwInfo.Subnets {
+		if nwInfo.Subnets[i].Family == platform.AfINET6 {
+			if err = nm.addIPv6DefaultRoute(); err != nil {
+				// should not block network creation but remind user that it's failed to add ipv6 default route to windows node
+				logger.Error("failed to add missing ipv6 default route to windows node active/persistent store", zap.Error(err))
+			}
```
ramiro-gamarra, line 402: need to add a `break` here to make clear that once the route is added we don't continue looping.

network/network_windows_test.go
```diff
@@ -227,3 +229,46 @@ func TestDeleteNetworkImplHnsV1WithTimeout(t *testing.T) {
...
+		Subnets: []string{"subnet1", "subnet2"},
+	}
+
+	Hnsv2 = hnswrapper.NewHnsv2wrapperFake()
+
+	// check if network can be successfully created
+	_, err := nm.newNetworkImplHnsV2(nwInfo, extInterface)
+	if err != nil {
+		fmt.Printf("+%v", err)
+		t.Fatal(err)
```
ramiro-gamarra, line 272: - remove the print statement. it will be already printed by the fatal call.
- add context to failure.

network/network_windows_test.go
```diff
@@ -227,3 +229,46 @@ func TestDeleteNetworkImplHnsV1WithTimeout(t *testing.T) {
...
+
+	extInterface := &externalInterface{
+		Name:    "eth0",
+		Subnets: []string{"subnet1", "subnet2"},
+	}
+
+	Hnsv2 = hnswrapper.NewHnsv2wrapperFake()
+
+	// check if network can be successfully created
+	_, err := nm.newNetworkImplHnsV2(nwInfo, extInterface)
```
ramiro-gamarra, line 269: I'm curious about how this test is run.
- Does it use powershell directly on the VM?
- If so, how do we know if the route is there or not on each run? How do we cover happy path as well as error paths in the code?

ramiro-gamarra (changes requested):

network/network_windows.go
```diff
@@ -318,6 +325,45 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+	if out, err := nm.plClient.ExecutePowershellCommand(getIPv6RoutePersistentCmd); err != nil {
+		logger.Info("ipv6 default route is not found from persistentstore, adding default ipv6 route to the windows node", zap.String("out", out))
+		// run powershell cmd to add ipv6 default route
+		addCmd := fmt.Sprintf("Remove-NetRoute -DestinationPrefix %s -InterfaceIndex %s -NextHop %s -confirm:$false;New-NetRoute -DestinationPrefix %s -InterfaceIndex %s -NextHop %s -confirm:$false",
+			defaultIPv6Route, ifIndex, defaultIPv6NextHop, defaultIPv6Route, ifIndex, defaultIPv6NextHop)
+
+		// if command is failed to execute, then attempt to execute additional two times
+		for i := 0; i <= addIPv6DefaultRouteRetryAttempts; i++ {
+			if out, err := nm.plClient.ExecutePowershellCommand(addCmd); err != nil {
+				logger.Error("Failed to add ipv6 default gateway route, retrying after error", zap.String("out", out), zap.Error(err))
```
tamilmani1989, line 349: add continue and remove else

network/network_windows.go
```diff
@@ -318,6 +325,45 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+		return errors.Wrap(err, "error while executing powershell command to get ipv6 Hyper-V interface")
+	}
+
+	getIPv6RouteActiveCmd := fmt.Sprintf("Get-NetRoute -DestinationPrefix %s", defaultIPv6Route)
+	getIPv6RoutePersistentCmd := fmt.Sprintf("Get-NetRoute -DestinationPrefix %s -PolicyStore Persistentstore", defaultIPv6Route)
+	if out, err := nm.plClient.ExecutePowershellCommand(getIPv6RoutePersistentCmd); err != nil {
+		logger.Info("ipv6 default route is not found from persistentstore, adding default ipv6 route to the windows node", zap.String("out", out))
+		// run powershell cmd to add ipv6 default route
+		addCmd := fmt.Sprintf("Remove-NetRoute -DestinationPrefix %s -InterfaceIndex %s -NextHop %s -confirm:$false;New-NetRoute -DestinationPrefix %s -InterfaceIndex %s -NextHop %s -confirm:$false",
+			defaultIPv6Route, ifIndex, defaultIPv6NextHop, defaultIPv6Route, ifIndex, defaultIPv6NextHop)
```
tamilmani1989, line 344: add a comment on why we remove and adding again

network/network_windows_test.go
```diff
@@ -227,3 +229,45 @@ func TestDeleteNetworkImplHnsV1WithTimeout(t *testing.T) {
 		t.Fatal("Failed to timeout HNS calls for deleting network")
 	}
 }
+
+func TestAddIPv6DefaultRoute(t *testing.T) {
+	_, ipnetv4, _ := net.ParseCIDR("10.240.0.0/12")
+	_, ipnetv6, _ := net.ParseCIDR("fc00::/64")
+
+	nm := &networkManager{
+		ExternalInterfaces: map[string]*externalInterface{},
```
tamilmani1989, line 238: have to use mock plclient

network/network_windows.go
```diff
@@ -318,6 +325,38 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *NetworkInfo, extIf *extern
...
+
+		// if command is failed to execute, then attempt to execute additional two times
+		for i := 0; i <= addIPv6DefaultRouteRetryAttempts; i++ {
+			if out, err := nm.plClient.ExecutePowershellCommand(addCmd); err != nil {
+				logger.Error("Failed to add ipv6 default gateway route, retrying after error", zap.String("out", out), zap.Error(err))
+				continue
+			}
+		}
```
tamilmani1989, line 354:
```suggestion
		for i := 0; i <= addIPv6DefaultRouteRetryAttempts; i++ {
			if out, err := nm.plClient.ExecutePowershellCommand(addCmd); err != nil {
				logger.Error("Failed to add ipv6 default gateway route, retrying after error", zap.String("out", out), zap.Error(err))
				continue
			}
			break
		}
```

tamilmani1989 (changes requested):

network/network_windows_test.go
```diff
@@ -227,3 +229,46 @@ func TestDeleteNetworkImplHnsV1WithTimeout(t *testing.T) {
...
+
+	extInterface := &externalInterface{
+		Name:    "eth0",
+		Subnets: []string{"subnet1", "subnet2"},
+	}
+
+	Hnsv2 = hnswrapper.NewHnsv2wrapperFake()
+
+	// check if network can be successfully created
+	_, err := nm.newNetworkImplHnsV2(nwInfo, extInterface)
```
tamilmani1989, line 270: this UT doesn't test whole code path.. it should test all code paths.. you may need to return failure and then only it will go to path where it attempts to add route

tamilmani1989 (approved):
lgtm

ramiro-gamarra (approved):
lgtm, thank you for helping out with this change!

PR #2514, opened 2024-01-16
CI: Setup Cilium with Hubble Enabled nightly run

MikeZappa87:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


MikeZappa87:
@matmerr

.pipelines/networkobservability/pipeline.yaml
```diff
@@ -0,0 +1,145 @@
+pr: none
+trigger: none
+
+schedules:
+- cron: '0 0 * * *'
```
jpayne3506, line 5: I like that we are controlling the scheduling from the yaml. Just be advised that changes within the AZP UI will override what you have here

.pipelines/networkobservability/pipeline.yaml
```diff
@@ -0,0 +1,145 @@
...
+  displayName: Daily midnight Cilium with Hubble
+  branches:
+    include:
+    - master
+
+variables:
+  clustername: ciliumhubble-$(Build.SourceBranchName)-$(Build.BuildId)
+  cilium_version: v1.14.4
```
jpayne3506, line 13: nit: Are we sure this should be hardcoded?

.pipelines/networkobservability/pipeline.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+schedules:
+- cron: '0 0 * * *'
+  displayName: Daily midnight Cilium with Hubble
+  branches:
+    include:
+    - master
+
+variables:
+  clustername: ciliumhubble-$(Build.SourceBranchName)-$(Build.BuildId)
```
jpayne3506, line 12: How many characters does this normally come out to? 26?

.pipelines/networkobservability/pipeline.yaml
```diff
@@ -0,0 +1,145 @@
...
+            then
+                CNS=$(CNS_VERSION) IPAM=$(AZURE_IPAM_VERSION) && echo "Running nightly"
+            else
+                CNS=$(make cns-version) IPAM=$(make azure-ipam-version)
+            fi
+            sudo -E env "PATH=$PATH" make test-integration AZURE_IPAM_VERSION=${IPAM} CNS_VERSION=${CNS} INSTALL_CNS=true INSTALL_OVERLAY=true
+          retryCountOnTaskFailure: 3
+          name: "aziliumTest"
+          displayName: "Install Azure-CNS and Run Azilium E2E on AKS Overlay"
+          enabled: true
```
jpayne3506, line 102: Is this pipeline/step also being used in multiple locations?

.pipelines/networkobservability/pipeline.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+        - script: |
+            kubectl apply -f test/integration/manifests/cilium/$(cilium_version)/hubble/hubble-peer-svc.yaml
+            kubectl get pods -Aowide
+            echo "verify Hubble metrics endpoint is usable"
+            go test ./test/integration/networkobservability -tags=networkobservability
+          retryCountOnTaskFailure: 3
+          name: "HubbleConnectivityTests"
+          displayName: "Run Hubble Connectivity Tests"
+
```
jpayne3506, line 132: If you want you could also add a job here for logs so that way you could look at cilium files from the nightly run. https://github.com/Azure/azure-container-networking/blob/5c1fe966129a287533cc90f596ce19c19aff65ff/.pipelines/cni/cilium/nightly-release-test.yml#L107-L118

jpayne3506 (commented):
Looks good. Might want to add `k get pods -Aowide` before tests to add extra context when looking back on failed runs.

.pipelines/networkobservability/pipeline.yaml
MikeZappa87: This is bound to the folder structure. It doesn't make sense to be dynamic (yet) until we come up with a proper strategy.

.pipelines/networkobservability/pipeline.yaml
MikeZappa87: That is a good call out. In some cases I could see it useful that we have that ability.

.pipelines/networkobservability/pipeline.yaml
MikeZappa87: In another branch I actually change this format since it gets long. The branch off this branch is cilium-$(Build.BuildId)-$(stack)

.pipelines/networkobservability/pipeline.yaml
MikeZappa87: Yes. We need to come up with a strategy here as right now we have duplicate steps/stages scattered across multiple files :-(

.pipelines/networkobservability/pipeline.yaml
MikeZappa87: Sounds reasonable and useful for troubleshooting.

.pipelines/networkobservability/pipeline.yaml
jpayne3506: If I remember right, the magic number to ensure you can use any region in AKS is 28. AKS char limit is 80 for context.

MikeZappa87: I ran into the 80 char limit earlier today which was the main driver to changing the naming strategy.

MikeZappa87: See: https://github.com/Azure/azure-container-networking/pull/2517

matmerr (approved):
can update tests in the future for require.NoError(...) in lieu of t.Fail(), but that's non-blocking for the moment

🚀

PR #2515, opened 2024-01-16
fix: block pod to wireserver port 80 traffic on windows multitenancy

QxBytes:
**Reason for Change**:

Traffic from the pod to the wireserver on port 80 should not leave the VM.

**Issue Fixed**:

See above.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Tested Windows Container on Windows 2022. After rule is added, wgets from the pod to websites such as google.com are still seen leaving the pod destined towards the 12:34... MAC. After the rule is added, traffic is still detected when pinging other pods (confirming the rule does not block pod-pod traffic). Tested routing wireserver requests to the eth0 and apipa eth inside the pod and confirmed that for hyper-v isolation containers, the vfp filter counter increases when requests to the wireserver are made (no traffic on the node is detected going to the wireserver from the pod either). Without the rule, traffic is detected on the node from the pod to the wireserver.

### DIFF
--- a/cni/azure-windows-multitenancy.conflist
+++ b/cni/azure-windows-multitenancy.conflist
@@ -42,6 +42,19 @@
                         "DestinationPrefix": "10.0.0.0/8",
                         "NeedEncap": true
                     }
+                },
+                {
+                    "Name": "EndpointPolicy",
+                    "Value": {
+                        "Type": "ACL",
+                        "Protocols": "6",
+                        "Action": "Block",
+                        "Direction": "Out",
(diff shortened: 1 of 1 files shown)


tamilmani1989 (approved):

tamilmani1989:
@QxBytes can you also get approval from @ashvindeodhar

vipul-21:
Curious to know why this is only a problem in windows multitenancy and not others?

PR #2527, opened 2024-01-18
ci: E2E Framework [Scenarios] [5/6]

matmerr:
**Reason for Change**:

Has strong dependency on:

https://github.com/Azure/azure-container-networking/pull/2528
https://github.com/Azure/azure-container-networking/pull/2529

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] relevant PR labels added

**Notes**:

PR #2529, opened 2024-01-18
ci: E2E Framework [Kubernetes steps] [3/6]

matmerr:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/e2e/framework/kubernetes/create-agnhost-statefulset.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+var ErrLabelMissingFromPod = fmt.Errorf("label missing from pod")
+
+const (
+	AgnhostHTTPPort = 80
+	AgnhostReplicas = 1
+)
+
+type CreateAgnhostStatefulSet struct {
```
vakalapa, line 24: you can drop the "Create" prefix ? Just AgnhostStatefulSet is good ?

vakalapa: Also will be easy to have a New() func ?

test/e2e/framework/kubernetes/create-kapinger-deployment.go
```diff
@@ -0,0 +1,256 @@
...
+)
+
+const (
+	KapingerHTTPPort  = 8080
+	KapingerTCPPort   = 8085
+	KapingerUDPPort   = 8086
+	MaxAffinityWeight = 100
+)
+
+type CreateKapingerDeployment struct {
```
vakalapa, line 26: Same comment, remove "Create" ?

test/e2e/framework/kubernetes/create-agnhost-statefulset.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (c *CreateAgnhostStatefulSet) Prevalidate() error {
+	return nil
+}
+
+func (c *CreateAgnhostStatefulSet) Postvalidate() error {
+	return nil
+}
+
+func (c *CreateAgnhostStatefulSet) getAgnhostDeployment() *appsv1.StatefulSet {
```
vakalapa, line 73: Also, a ping or curl command with this object should do a exec on this pod to ping or curl given IP or host, that will be nice

test/e2e/framework/kubernetes/create-agnhost-statefulset.go
matmerr: > you can drop the "Create" prefix ? Just AgnhostStatefulSet is good ?

Since is this an actional step and not an actual appsv1.StatefulSet struct, I'd prefer to keep verb prefix, that way when using intellisense and I type `kubernetes.` I'd see all of the create steps adjacent to each other in that package matching a similar verb prefix

```
kubernetes.CreateAgnhostStatefulSet
kubernetes.CreateNeworkPolicy
Kubernetes.CreateX
```

and then it's easy to track in the test when to have a corresponding step doing another crud action

```
kubernetes.DeleteAgnhostStatefulSet
kubernetes.DeleteNetworkPolicy
...
```

test/e2e/framework/kubernetes/create-agnhost-statefulset.go
matmerr: Currently the objects themselves created by steps aren't saved to the job context. I had a version that exposed struct passing within the job context using map[string]interface{}, but became a nightmare with typecasting and prevalidation. By just keeping parameters in job context, we can do prevalidation ahead of time to make sure everything is populated, and less likely to have "x is nil" passed to a step 1hr into execution 

However, accomplish the mentioned scenario, the parameters passed to the steps themselves are what exist in the job context, see sample scenario in PR #5, the parameter PodName passed to the agnhost create step is then subsequently used in the exec step to curl:

https://github.com/Azure/azure-container-networking/pull/2527/files#diff-4585a7357e679c3568ad5f63a961dc38f34a73087d796e75f718986f60812f5cR37-R60

Open to for future discussion around passing structs to subsequent steps, but then devs need to be very meticulous when authoring, and/or leverage prevalidate steps to write sample objects to context which subsequent steps can also prevalidate

test/e2e/framework/kubernetes/create-kapinger-deployment.go
matmerr: Same thoughts as above

test/e2e/framework/kubernetes/create-agnhost-statefulset.go
matmerr: > Also will be easy to have a New() func

imo adding the struct directly to the slice makes it very readable to see what parameters are passed, if using a NewObjectName(...), then it opens the door for users of the step to misplace the order of the parameters, and then might end up creating a resource with an unintended name

See: https://github.com/Azure/azure-container-networking/pull/2527/files#diff-649efb346a9d03496b0347a2dfd0214ff4f242b3ae8ef54ca3ca02b00931a43fR32-R53

vakalapa (approved):

vakalapa (dismissed):

vakalapa (approved):

PR #2545, opened 2024-01-25
feat: create or update nodeinfo crd at cns startup for multitenancy

thatmattlong:
**Reason for Change**:
CNS will write the vm unique id to the nodeinfo crd (or update it, if one already exists) for the node at startup if swift v2 is enabled. This is required because we can't reliably read the vm unique id from the node.Status.NodeInfo.SystemUniqueID property, as it can be reported by kubelet/cadvisor with the wrong endianess.

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


cns/service/main.go
```diff
@@ -1215,8 +1219,9 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 	if _, ok := node.Labels[configuration.LabelNodeSwiftV2]; ok {
 		cnsconfig.EnableSwiftV2 = true
 		cnsconfig.WatchPods = true
-		// TODO(rbtr): create the NodeInfo for Swift V2
-		// register the noop mtpnc reconciler to populate the cache
+		if nodInfoErr := createOrUpdateNodeInfoCRD(ctx, kubeConfig, node); nodInfoErr != nil {
```
aggarwal0009, line 1222: nit; nodeInfo

aggarwal0009 (approved):
LGTM

crd/multitenancy/client.go
```diff
@@ -190,3 +191,28 @@ func (i *Installer) InstallOrUpdatePodNetworkInstance(ctx context.Context) (*v1.
 	}
 	return current, nil
 }
+
+type NodeInfoClient struct {
```
nddq, line 195: Looks like we have implemented an `Installer` struct which seems abstracted enough to be used for NodeInfo CRD as well, is there a reason why we can't build upon that?

rbtr: `Installer` is for the CustomResource**Definition**, not for the instance of the CustomResource

nddq: ah i see what you mean, not familiar with CNS creating the actual CR itself 🙂

miguelgoms (dismissed):

cns/service/main.go
thatmattlong: fixed

aggarwal0009 (approved):

nddq (approved):

miguelgoms (approved):

PR #2546, opened 2024-01-26
ci: Capture privileged pod failure

jpayne3506:
**Reason for Change**:

If privileged pod was the cause of the failure in a test scenario it could potentially get missed and the error logs overwritten. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/templates/log-template.yaml
+++ b/.pipelines/templates/log-template.yaml
@@ -76,6 +76,26 @@ steps:
     displayName: Deployment Images
     condition: always()
 
+  - ${{ if eq(parameters.logType, 'failure') }}:
+    - bash: |
+        kubectl get pods -n kube-system -o custom-columns=NAME:.metadata.name,STATUS:.status.phase,NODE:.spec.nodeName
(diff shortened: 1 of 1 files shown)


.pipelines/templates/log-template.yaml
```diff
@@ -76,6 +76,26 @@ steps:
...
+          echo "Directory created: $acnLogs/${{ parameters.os }}non-ready"
+          echo "Capturing failed pods"
+          for pod in $podList; do
+            kubectl describe pod -n kube-system $pod > $acnLogs/${{ parameters.os }}non-ready/$pod.txt
+            echo "$acnLogs/${{ parameters.os }}non-ready/$pod.txt"
+          done
+        fi
+      displayName: Failure Logs
+      condition: always()
+
```
rbtr, line 98: because it was relocated and the diff shows the whole block as new, i'm struggling to see what has actually changed here

jpayne3506: The change allows us to catch if the privileged pod failed to reach ready state from any of the various ways we apply the daemonset to the cluster. 

Originally we would use kubectl to apply the daemonset again to retrieve logs from the node. In doing this we deleted the error logs and did not catch that it was the cause for the failure.

PR #2552, opened 2024-01-27
ci: Kernel Upgrade Script

jpayne3506:
**Reason for Change**:

Adds in a script that allows anyone to update their kernel to a publicly available proposed Ubuntu kernel

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Dependent on #2553


hack/scripts/kernel.sh
```diff
@@ -0,0 +1,123 @@
+# Should be used for manual validation. Does not check if correct kernel is installed or if installation failed.
+# If used within automation, error check at the end
```
vipul-21, line 2: Let's add a shebang `#!/bin/bash` and `set -e`

hack/scripts/kernel.sh
```diff
@@ -0,0 +1,123 @@
...
+
+if [ ! $UPGRADE_ONLY = "true" ] || [ -z $UPGRADE_ONLY ]; then
```
vipul-21, line 4: A comment on the top with the attributes it accept would be helpful.

hack/scripts/kernel.sh
```diff
@@ -0,0 +1,123 @@
...
+privArray=(`kubectl get pods -n kube-system -l os=linux,app=privileged-daemonset -owide --no-headers | awk '{print $1}'`)
+nodeArray=(`kubectl get pods -n kube-system -l os=linux,app=privileged-daemonset -owide --no-headers | awk '{print $7}'`)
+kubectl get pods -n kube-system -l os=linux,app=privileged-daemonset -owide
+
+i=0
+for _ in ${privArray[@]}; do
+    echo "-- Restarting Node ${nodeArray[i]} through ${privArray[i]} --"
+    kubectl exec -i -n kube-system ${privArray[i]} -- bash -c "reboot"
+    echo "-- Waiting for condition NotReady --"
+    kubectl wait --for=condition=Ready=false -n kube-system pod/${privArray[i]} --timeout=90s
```
vipul-21, line 108: how does this work if node has not restarted and pod itself is down ?  Should we check for node `AGE` as well ?

hack/scripts/kernel.sh
```diff
@@ -0,0 +1,123 @@
...
+    echo "-- Waiting for condition Ready --"
+    kubectl wait --for=condition=Ready -n kube-system pod/${privArray[i]} --timeout=90s
+    ((i = i + 1))
+    echo "Wait 10s for pods to settle"
+    sleep 10s
+done
+
+kubectl rollout status ds -n kube-system privileged-daemonset
+for pod in $privList; do
+    echo "-- Check current Ubuntu kernel --"
```
vipul-21, line 118: This should say upgraded Ubuntu kernel ? Or maybe check with the earlier version and echo out if the upgrade happened or not.

hack/scripts/kernel.sh
jpayne3506: So,
1: triggers a reboot for the node on line 106 ` bash -c "reboot"`. This triggers a restart of the node immediately.
2: waits for the priv pod on the matching node to be reconciled and be marked as not ready `--for=condition=Ready=false -n kube-system pod/${privArray[i]} `
3: waits for priv pod to be marked ready `--for=condition=Ready -n kube-system pod/${privArray[i]}`

Not sure if `AGE` will help us here as due to the reconciliation.

hack/scripts/kernel.sh
jpayne3506: I think the current string makes sense as we are not doing anything within bash to confirm that it is in fact upgraded. This script is meant to be used manually and we would be visually confirming the kernel upgraded as well.

hack/scripts/kernel.sh
```diff
@@ -0,0 +1,123 @@
...
+    echo "-- Install dummy CNI for nodes to be marked as Ready --"
+    kubectl get pods -Aowide
+    kubectl apply -f https://raw.githubusercontent.com/Azure/azure-container-networking/v1.5.3/hack/manifests/cni-installer-v1.yaml
+    kubectl rollout status ds -n kube-system azure-cni
+fi
+
+echo "-- Start privileged daemonset --"
+kubectl get pods -Aowide
+kubectl apply -f ../../test/integration/manifests/load/privileged-daemonset.yaml
+sleep 10s
```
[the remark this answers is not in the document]
jpayne3506: bash script goes to fast and rollout status catches the 0 desired and 0 ready state. We could lower it to 3s..?

hack/scripts/kernel.sh
jpayne3506: Is failing fast worth fighting through false negatives coming from kubectl exec?

hack/scripts/kernel.sh
```diff
@@ -0,0 +1,123 @@
...
+# If used within automation, error check at the end
+
+if [ ! $UPGRADE_ONLY = "true" ] || [ -z $UPGRADE_ONLY ]; then
+    if [ ! $DEFAULT = "true" ] || [ -z $DEFAULT ]; then
+        echo "User should set CLUSTER_TYPE CLUSTER_NAME REGION SUB VMSIZE DUMMY_CNI. Otherwise, pass in DEFAULT=true and SUB"
+        echo "Proceeding with variable check"
+    else
+        CLUSTER_TYPE=overlay-byocni-up
+        REGION=eastus
+        CLUSTER_NAME="jpayne-kernel-$(date "+%d%H%M")"
```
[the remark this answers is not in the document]
jpayne3506: matched the cluster name from the pipeline template.

hack/scripts/kernel.sh
```diff
@@ -0,0 +1,135 @@
...
+        echo "DUMMY_CNI is empty"
+        flag="true"
+    fi
+    if [ flag = "true" ]; then
+        echo "Please ensure that variables are set correctly."
+        echo "Exiting script gracefully"
+        exit 1
+    fi
+    echo "-- Cluster Create --"
+    make -C ../aks ${CLUSTER_TYPE} \
```
rbtr, line 66: i think this is far outside of the responsibility of the "kernel upgrade script"

hack/scripts/kernel.sh
```diff
@@ -0,0 +1,135 @@
...
+kubectl get pods -n kube-system -l os=linux,app=privileged-daemonset -owide
+privList=`kubectl get pods -n kube-system -l os=linux,app=privileged-daemonset -owide --no-headers | awk '{print $1}'`
+for pod in $privList; do
+    echo "-- Update Ubuntu Packages --"
+    # Not needed, but ensures that the correct packages exist to perform upgrade
+    kubectl exec -i -n kube-system $pod -- bash -c "apt update && apt-get install software-properties-common -y"
+
+    echo "-- Add proposed repository --"
+    kubectl exec -i -n kube-system $pod -- bash -c "add-apt-repository ppa:canonical-kernel-team/proposed -y"
+    kubectl exec -i -n kube-system $pod -- bash -c "add-apt-repository ppa:canonical-kernel-team/proposed2 -y"
```
rbtr, line 98: instead of exec'ing in to a pod for every line of bash, you could write a separate payload file, deliver it to the node via `kubectl cp`, then execute it

hack/scripts/kernel.sh
```diff
@@ -0,0 +1,135 @@
...
+for _ in ${privArray[@]}; do
+    echo "-- Restarting Node ${nodeArray[i]} through ${privArray[i]} --"
+    kubectl exec -i -n kube-system ${privArray[i]} -- bash -c "reboot"
+    echo "-- Waiting for condition NotReady --"
+    kubectl wait --for=condition=Ready=false -n kube-system pod/${privArray[i]} --timeout=90s
+    echo "-- Waiting for condition Ready --"
+    kubectl wait --for=condition=Ready -n kube-system pod/${privArray[i]} --timeout=90s
+    ((i = i + 1))
+    echo "Wait 10s for pods to settle"
+    sleep 10s
```
rbtr, line 125: an opportunity to integrate with the new Go test framework and write less bash, maybe? it's pretty unreadable to do anythnig nontrivial with just bash and kubectl

hack/scripts/kernel.sh
jpayne3506: Considering that I am not doing any error checking with this script I think this is a pretty valid point.

hack/scripts/kernel.sh
jpayne3506: I think this is a scenario that should be created within the framework as well. For now, I think with improvements (delivering payloads with `kubectl cp`, removing cluster creation) readability and useability will be improved.

rbtr (approved):

hack/scripts/cluster-kernel-upgrade.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+# Should be used for manual validation. Does not check if correct kernel is installed or if installation failed.
+
+# DUMMY_CNI     - Flag to provide a temporary CNI
+if [ $DUMMY_CNI = "true" ]; then
+    echo "-- Install dummy CNI for nodes to be marked as Ready --"
+    kubectl get pods -Aowide
+    kubectl apply -f https://raw.githubusercontent.com/Azure/azure-container-networking/v1.5.3/hack/manifests/cni-installer-v1.yaml
```
vipul-21, line 8: The version here is different from the version for privileged-daemonset. Going forward are these going to be same or we need to change them as we release new versions ? If this for testing maybe we can add version parameter to script.

PR #2577, opened 2024-02-12
fix: delete endpoint on cni state save failure

QxBytes:
**Reason for Change**:

In windows, it is possible to successfully create an HNS endpoint and then subsequently fail to write to the CNI statefile. While the CNI call fails, the HNS endpoint is never deleted, causing retries to fail because the HNS endpoint has already been created. As it is not possible to retrieve the existing endpoint on a retried ADD call, we instead opt to delete the HNS endpoint if saving the CNI state fails.

**Issue Fixed**:

See above.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
The code change affects both the windows and linux scenarios (if saving fails, deleteEndpoint will call the platform specific version for deletion of the endpoint). On windows, tested triggering a save failure and then examining HNS endpoints-- the endpoint is cleaned up which is confirmed in the logs. Without the fix, the endpoint still shows and produces the endpoint already exists (ip address claimed) error. On linux transparent-vlan mode, tested triggering a save failure, which does call into deleting the endpoint. Subsequent delete/adds for that pod run without erroring.

### DIFF
--- a/network/manager.go
+++ b/network/manager.go
@@ -402,6 +402,10 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 
 	err = nm.save()
 	if err != nil {
+		delErr := nw.deleteEndpoint(nm.netlink, nm.plClient, nm.netio, nm.nsClient, nm.iptablesClient, ep.Id)
+		if delErr != nil {
+			logger.Error("Deleting endpoint after failing to save state failed with", zap.Error(delErr))
+		}
 		return err
 	}
 
--- a/network/manager.go
+++ b/network/manager.go
@@ -395,25 +395,24 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 	if err != nil {
 		return err
 	}
-	var delEndpointErr error
 	// any error after this point should also clean up the endpoint we created above
 	defer func() {
-		if delEndpointErr != nil {
+		if err != nil {
 			delErr := nw.deleteEndpoint(nm.netlink, nm.plClient, nm.netio, nm.nsClient, nm.iptablesClient, ep.Id)
 			if delErr != nil {
-				logger.Error("Deleting endpoint after failing to save state failed with", zap.Error(delErr))
+				logger.Error("Deleting endpoint after create endpoint failure failed with", zap.Error(delErr))
 			}
 		}
 	}()
 
 	if nm.IsStatelessCNIMode() {
-		delEndpointErr = nm.UpdateEndpointState(ep)
-		return delEndpointErr
+		err = nm.UpdateEndpointState(ep)
+		return err
 	}
 
-	delEndpointErr = nm.save()
-	if delEndpointErr != nil {
-		return delEndpointErr
+	err = nm.save()
+	if err != nil {
+		return err
 	}
 
 	return nil
--- a/network/manager.go
+++ b/network/manager.go
@@ -395,18 +395,25 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 	if err != nil {
 		return err
 	}
+	var delEndpointErr error
+	// any error after this point should also clean up the endpoint we created above
+	defer func() {
+		if delEndpointErr != nil {
+			delErr := nw.deleteEndpoint(nm.netlink, nm.plClient, nm.netio, nm.nsClient, nm.iptablesClient, ep.Id)
+			if delErr != nil {
+				logger.Error("Deleting endpoint after failing to save state failed with", zap.Error(delErr))
+			}
+		}
+	}()
 
 	if nm.IsStatelessCNIMode() {
-		return nm.UpdateEndpointState(ep)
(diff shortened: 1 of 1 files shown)


network/manager.go
```diff
@@ -395,14 +395,25 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 	if err != nil {
 		return err
 	}
+	var delEndpointErr error
+	// any error after this point should also clean up the endpoint we created above
+	defer func() {
+		if delEndpointErr != nil {
+			delErr := nw.deleteEndpoint(nm.netlink, nm.plClient, nm.netio, nm.nsClient, nm.iptablesClient, ep.Id)
+			if delErr != nil {
+				logger.Error("Deleting endpoint after failing to save state failed with", zap.Error(delErr))
```
tamilmani1989, line 404:
```suggestion
				logger.Error("Deleting endpoint after create endpoint failure failed with", zap.Error(delErr))
```

network/manager.go
```diff
@@ -395,14 +395,25 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
...
 	if nm.IsStatelessCNIMode() {
-		return nm.UpdateEndpointState(ep)
+		delEndpointErr = nm.UpdateEndpointState(ep)
+		return delEndpointErr
 	}
 
-	err = nm.save()
-	if err != nil {
-		return err
+	delEndpointErr = nm.save()
```
tamilmani1989, line 414: why renaming from err to delEndpointErr

tamilmani1989 (changes requested):

network/manager.go
QxBytes: Originally, I wanted to make it clear which error would be checked (that any previous assignment of err wouldn't affect the defer, but since we cover the err != nil case right before the defer that will never occur). I'll change it to err.

network/manager.go
```diff
@@ -395,9 +395,19 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 	if err != nil {
 		return err
 	}
+	// any error after this point should also clean up the endpoint we created above
+	defer func() {
+		if err != nil {
+			delErr := nw.deleteEndpoint(nm.netlink, nm.plClient, nm.netio, nm.nsClient, nm.iptablesClient, ep.Id)
```
tamilmani1989, line 401: can we log error before calling delete.. otherwise it will be confusing when reading logs

network/manager.go
```diff
@@ -395,9 +395,20 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 	if err != nil {
 		return err
 	}
+	// any error after this point should also clean up the endpoint we created above
+	defer func() {
+		if err != nil {
+			logger.Error("Create endpoint failure failed with", zap.Error(err))
```
tamilmani1989, line 401:
```suggestion
			logger.Error("Create endpoint failure. Cleanup resources", zap.Error(err))
```

network/manager.go
```diff
@@ -395,9 +395,21 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
...
 		return err
 	}
+	// any error after this point should also clean up the endpoint we created above
+	defer func() {
+		if err != nil {
+			logger.Error("Create endpoint failure", zap.Error(err))
+			logger.Info("Cleanup resources")
+			delErr := nw.deleteEndpoint(nm.netlink, nm.plClient, nm.netio, nm.nsClient, nm.iptablesClient, ep.Id)
+			if delErr != nil {
+				logger.Error("Deleting endpoint after create endpoint failure failed with", zap.Error(delErr))
```
jpayne3506, line 405: Isn't this an error log for if deleteEndpoint() fails?

QxBytes: yes

jpayne3506: This log looks like it is for a create endpoint failure while "Deleting" is making me think it succeeding or is actively doing something. Perhaps we can reference that this delete is a cleanup of a failed create endpoint?
```suggestion
				logger.Error("Create endpoint cleanup with delete endpoint failed with", zap.Error(delErr))
```

 I don't think just stating `delete endpoint failed` is good enough here so I understand the verbosity.

network/manager.go
```diff
@@ -395,9 +395,21 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 	if err != nil {
 		return err
 	}
+	// any error after this point should also clean up the endpoint we created above
+	defer func() {
```
vipul-21, line 399: Curios to know how are we checking if `cni state save failure` is actually the error that triggered this deferred function ? I see there is stateless CNI code as well that tries to update the endpoint state. If that call fails do we want to delete the endpoint in that scenario as well ?

QxBytes: Any failure after creating the endpoint should trigger deleting the endpoint (ex: if stateless CNI fails)

Issue #2573, opened 2024-02-09
Remove CNMS Code


QxBytes:
**What happened:** 
The CNMS code is currently unused (not built, no images published with it) and should be removed.

**What you expected to happen:**  

**How to reproduce it:**  

**Orchestrator and Version (e.g. Kubernetes, Docker):**  

**Operating System (Linux/Windows):**  

**Kernel (e.g. `uanme -a` for Linux or `$(Get-ItemProperty -Path "C:\windows\system32\hal.dll").VersionInfo.FileVersion` for Windows):**

**Anything else we need to know?:**

Issue #2565, opened 2024-02-07
Remove assests from release which are not supported


tamilmani1989:
**What happened:** 
There are number of artifacts thats getting released in unsupported.  The ask is to avoid building those targets and remove artifacts from release that are not supported.

azure-vnet-cni-multitenancy-transparent-vlan-windows-amd64-v1.5.20.zip,
azure-vnet-cnm-linux-amd64-v1.5.20.tgz, 
azure-vnet-cnm-linux-arm64-v1.5.20.tgz
azure-vnet-cnm-windows-amd64-v1.5.20.zip
azure-vnet-cnm-windows-arm64-v1.5.20.zip

**What you expected to happen:**  

**How to reproduce it:**  

**Orchestrator and Version (e.g. Kubernetes, Docker):**  

**Operating System (Linux/Windows):**  

**Kernel (e.g. `uanme -a` for Linux or `$(Get-ItemProperty -Path "C:\windows\system32\hal.dll").VersionInfo.FileVersion` for Windows):**

**Anything else we need to know?:**
[Miscellaneous information that will assist in solving the issue.]

PR #2583, opened 2024-02-13
[standalone][CNS] CNS to listen on addresses that its serving the request for

paulyufan2:
**Reason for Change**:

In this PR, we are going to create two listeners; one listener is listening DNC/CNS connection as nodeListener, another listener is localListener that's listening to CNI connection;

The TLS connection is only for DNC listener if customer provides TLS settings

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cnm/ipam/ipam.go
+++ b/cnm/ipam/ipam.go
@@ -118,7 +118,7 @@ func (plugin *ipamPlugin) getCapabilities(w http.ResponseWriter, r *http.Request
 		RequiresRequestReplay: requiresRequestReplay,
 	}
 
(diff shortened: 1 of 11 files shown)


rbtr:
don't update CNM
https://github.com/Azure/azure-container-networking/issues/2573

tamilmani1989:
> don't update CNM #2573

@rbtr that `CNMS` though and not CNM.. but I think we should remove cnm as well

rbtr:
maybe I should have linked https://github.com/Azure/azure-container-networking/issues/2565

PR #2593, opened 2024-02-15
fix: no longer use pip and keyring for az devops login

jpayne3506:
**Reason for Change**:

pip and keyring have long standing issues that result in flaky logins with `az devops login` 

Moving to login via PAT env var avoids using the above and is a[ 1 of 3 suggested ways to login. ](https://learn.microsoft.com/en-us/azure/devops/cli/log-in-via-pat?view=azure-devops&tabs=windows)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.github/workflows/pipeline.yaml
+++ b/.github/workflows/pipeline.yaml
@@ -35,7 +35,7 @@ jobs:
         azcliversion: latest
         inlineScript: |
           az account show
-          export ${{ secrets.AZURE_DEVOPS_EXT_PAT }}
+          export AZURE_DEVOPS_EXT_PAT=${{ secrets.AZURE_DEVOPS_EXT_PAT }}
 
           echo "Sanity check recently triggered Merge Queue Pipeline runs"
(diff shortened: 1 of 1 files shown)


.github/workflows/pipeline.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
         azcliversion: latest
         inlineScript: |
           az account show
-          echo ${{ secrets.AZURE_DEVOPS_EXT_PAT }} | az devops login --org ${{ secrets.AZURE_PIPELINE_ORG }}
+          export ${{ secrets.AZURE_DEVOPS_EXT_PAT }}
```
rbtr, line 38: does this need to be 
```suggestion
          export AZURE_DEVOPS_EXT_PAT=${{ secrets.AZURE_DEVOPS_EXT_PAT }}
```

rbtr (approved):

PR #2606, opened 2024-02-23
Remove azr enabled from cns

ZetaoZhuang:
**Reason for Change**:
Remove EnableAZR config flag from CNS, the config flag is unnecessary.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/homeazmonitor.go
+++ b/cns/restserver/homeazmonitor.go
@@ -10,7 +10,6 @@ import (
 	"github.com/Azure/azure-container-networking/cns/logger"
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/nmagent"
-	"github.com/google/go-cmp/cmp"
 	"github.com/patrickmn/go-cache"
 	"github.com/pkg/errors"
 )
@@ -65,6 +64,7 @@ func (h *HomeAzMonitor) readCacheValue() cns.GetHomeAzResponse {
 func (h *HomeAzMonitor) Start() {
 	logger.Printf("[HomeAzMonitor] start the goroutine for refreshing homeAz")
 	go h.refresh()
+	defer h.Stop()
 }
 
 // Stop ends the refresh thread
@@ -169,7 +169,7 @@ func (h *HomeAzMonitor) update(code types.ResponseCode, msg string, homeAzRespon
 	}
 
 	// log the response and update the cache if it doesn't match with the current cached value
-	if !cmp.Equal(h.readCacheValue(), resp) {
+	if h.readCacheValue() != resp {
 		logger.Printf("[HomeAzMonitor] updating home az cache value: %+v", resp)
 		h.updateCacheValue(resp)
 	}
--- a/cns/restserver/homeazmonitor.go
+++ b/cns/restserver/homeazmonitor.go
@@ -63,7 +63,6 @@ func (h *HomeAzMonitor) readCacheValue() cns.GetHomeAzResponse {
 // Start starts a new thread to refresh home az cache
 func (h *HomeAzMonitor) Start() {
 	logger.Printf("[HomeAzMonitor] start the goroutine for refreshing homeAz")
-	defer h.Stop()
 	go h.refresh()
 }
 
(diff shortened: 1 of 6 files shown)


cns/service/main.go
```diff
@@ -669,6 +663,11 @@ func main() {
 		config.ChannelMode = cns.Managed
 	}
 
+	homeAzMonitor := restserver.NewHomeAzMonitor(nmaClient, time.Duration(cnsconfig.AZRSettings.PopulateHomeAzCacheRetryIntervalSecs)*time.Second)
+	if cnsconfig.ChannelMode == cns.Direct {
```
ashvindeodhar, line 667: can you add a comment indicating that the homeAz monitor is only required when there is a direct channel between DNC and CNS. This will prevent the monitor from unnecessarily calling NMA APIs for other scenarios such as AKS-swift, swiftv2

ashvindeodhar (changes requested):

ashvindeodhar:
In the PR description please indicate that the change is to remove the config flag from CNS. The config flag is unnecessary

ashvindeodhar (approved):

cns/restserver/homeazmonitor.go
```diff
@@ -10,13 +10,14 @@ import (
 	"github.com/Azure/azure-container-networking/cns/logger"
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/nmagent"
+	"github.com/google/go-cmp/cmp"
```
rbtr, line 13: [do not use cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp)
> It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal.

cns/restserver/homeazmonitor.go
```diff
@@ -165,8 +167,12 @@ func (h *HomeAzMonitor) update(code types.ResponseCode, msg string, homeAzRespon
 		},
 		HomeAzResponse: homeAzResponse,
 	}
-	logger.Printf("[HomeAzMonitor] updating home az cache value: %+v", resp)
-	h.updateCacheValue(resp)
+
+	// log the response and update the cache if it doesn't match with the current cached value
+	if !cmp.Equal(h.readCacheValue(), resp) {
```
rbtr, line 172: this looks like a value type (struct of value types), and could be compared with `==`.
https://go.dev/ref/spec#Comparison_operators
> Struct types are comparable if all their field types are comparable. Two struct values are equal if their corresponding non-[blank](https://go.dev/ref/spec#Blank_identifier) field values are equal. The fields are compared in source order, and comparison stops as soon as two field values differ (or all fields have been compared).

cns/service/main.go
```diff
@@ -1015,10 +1016,7 @@ func main() {
 		}
 	}
 
-	if cnsconfig.AZRSettings.EnableAZR {
-		logger.Printf("end the goroutine for refreshing homeAz")
-		homeAzMonitor.Stop()
-	}
+	homeAzMonitor.Stop()
```
rbtr, line 1019: why not `defer homeAzMonitor.Stop()` under the `Start()`?

rbtr (changes requested):

cns/restserver/homeazmonitor.go
```diff
@@ -62,6 +62,8 @@ func (h *HomeAzMonitor) readCacheValue() cns.GetHomeAzResponse {
 
 // Start starts a new thread to refresh home az cache
 func (h *HomeAzMonitor) Start() {
+	logger.Printf("[HomeAzMonitor] start the goroutine for refreshing homeAz")
+	defer h.Stop()
```
rbtr, line 66: i meant defer stop in main - if you do it here you will start and _immediately_ stop

ZetaoZhuang: ohh.Thanks

rbtr:
lgtm but want a review from @ashvindeodhar since it's changed a bit since the approval

PR #2617, opened 2024-02-29
fix: add in missing ACLs for windows multitenancy

pjohnst5:
In #2515 , an ACL was introduced to block port 80 of Wireserver
However, this also breaks all connectivity of a Windows multitenany pod
Adding in missing ACLs to fix this
Pings work just fine now with this fix:
image


pjohnst5:
I see this ACL was added in #2515 but this fails multitenancy for Windows, was this tested @QxBytes ?

pjohnst5:
The reason I say this breaks QxBytes, is if a customer/partner team were to use Kubernetes with this conflist, the pods would not be able to resolve DNS, which is done through wireserver `168.63.129.16` (because multitenant customer pods could not use coredns, since these customer pods are in a differnet network than the K8S infra)

So I suggest here we add this back in, at least for testing, and maybe even back to the prod one too, because I don't see why we need to block Wireserver IP

pjohnst5:
@tamilmani1989 also to comment on this

pjohnst5:
Validated both that outbound traffic to wireserver 80 is blocked within the pod and wireserver via pod's apipia endpoint also is blocked while seeing blocked packets increase by +1 each time

QxBytes (approved):
Paul and I confirmed that invoking a request to google.com is successful with these two added ACL rules. DNS functionality with nslookup within the container was also confirmed and successful. We also validated that the wireserver traffic is blocked via both apipa and the container's eth0 by checking the vfp rules and seeing the dropped packet counter increase when a request is issued.

tamilmani1989 (approved):

PR #2618, opened 2024-03-01
ci: Add safeguards to prevent race condition resulting from MustRestartDaemonset()

jpayne3506:
**Reason for Change**:

`MustRestartDaemonset()` did not have safeguards to ensure that `WaitForPodDaemonset()` could capture a restart. This was due to the fact that we need not ensure that a restart had actually occurred within `MustRestartDaemonset()` just that the k8sAPI request was sent and received without errors. The safeguard that was introduced is a short lived check for generation update. This is normally the following response after 200-ing the update request. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/internal/kubernetes/utils.go
+++ b/test/internal/kubernetes/utils.go
@@ -468,6 +468,7 @@ func MustRestartDaemonset(ctx context.Context, clientset *kubernetes.Clientset, 
 		ds.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
 	}
 
+	// gen represents the generation before triggering a restart
 	gen := ds.Status.ObservedGeneration
 	log.Printf("Current generation is %v", gen)
 	ds.Spec.Template.ObjectMeta.Annotations["kubectl.kubernetes.io/restartedAt"] = time.Now().Format(time.RFC3339)
--- a/test/internal/kubernetes/utils.go
+++ b/test/internal/kubernetes/utils.go
@@ -466,10 +466,31 @@ func MustRestartDaemonset(ctx context.Context, clientset *kubernetes.Clientset, 
 		ds.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
 	}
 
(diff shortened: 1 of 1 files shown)


test/internal/kubernetes/utils.go
```diff
@@ -466,10 +466,37 @@ func MustRestartDaemonset(ctx context.Context, clientset *kubernetes.Clientset,
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to update ds %s", daemonsetName)
+	}
+	checkDaemonsetGenerationFn := func() error {
+		ds, err := clientset.AppsV1().DaemonSets(namespace).Get(ctx, daemonsetName, metav1.GetOptions{})
+		if err != nil {
+			return errors.Wrapf(err, "could not get daemonset %s", daemonsetName)
+		}
+
+		if ds.Status.ObservedGeneration < gen {
```
jpayne3506, line 483: Technically if generation is less then observed then a restart has occurred, but not an intended one.

test/internal/kubernetes/utils.go
```diff
@@ -38,6 +38,8 @@ const (
 	RetryDelay              = 10 * time.Second
 	DeleteRetryAttempts     = 12
 	DeleteRetryDelay        = 5 * time.Second
+	ShortRetryAttempts      = 8
```
paulyufan2, line 41: can we know why this ShortRetryAttempts is set to 8?

jpayne3506: ShortRetry* should be used with funtions that fail fast and are short lived. 8 * (250ms + runtime) should come out to a 2.25s - 3s max time.

vipul-21 (approved):

jpayne3506:
Look into k8sapi event based solution to replace `Retrier`.

paulyufan2 (approved):

PR #2634, opened 2024-03-06
Add swiftv2 middleware support for standalone service fabric scenario (windows)

kmurudi:
**Reason for Change**:
L1VH/Singularity & Swiftv2 need to program secondary nics for creating pod via containerd. Changes have been supported for AKS pod scenario, but this PR adds support for windows Service fabric/standalone scenario as it gets the data from DNC goal state.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/middlewares/k8sSwiftV2.go
```diff
@@ -282,3 +282,7 @@ func (m *K8sSWIFTv2Middleware) setRoutes(podIPInfo *cns.PodIpInfo) error {
 	}
 	return nil
 }
+
+func (m *K8sSWIFTv2Middleware) GetMiddlewareType() configuration.SWIFTV2Mode {
+	return configuration.SFSWIFTV2
```
nddq, line 287: this should be k8sSWIFTV2

cns/restserver/ipam.go
```diff
@@ -179,8 +278,13 @@ func (service *HTTPRestService) requestIPConfigsHandler(w http.ResponseWriter, r
 
 	// Check if IPConfigsHandlerMiddleware is set
 	if service.IPConfigsHandlerMiddleware != nil {
-		// Wrap the default datapath handlers with the middleware
-		wrappedHandler := service.IPConfigsHandlerMiddleware.IPConfigsRequestHandlerWrapper(service.requestIPConfigHandlerHelper, service.releaseIPConfigHandlerHelper)
+		// Wrap the default datapath handlers with the middleware depending on middleware type
+		var wrappedHandler cns.IPConfigsHandlerFunc
+		if service.IPConfigsHandlerMiddleware.GetMiddlewareType() == configuration.K8sSWIFTV2 {
```
nddq, line 283: can use a `switch` here

cns/restserver/ipam.go
```diff
@@ -179,8 +278,13 @@ func (service *HTTPRestService) requestIPConfigsHandler(w http.ResponseWriter, r
...
+			wrappedHandler = service.IPConfigsHandlerMiddleware.IPConfigsRequestHandlerWrapper(service.requestIPConfigHandlerHelper, service.releaseIPConfigHandlerHelper)
+		} else if service.IPConfigsHandlerMiddleware.GetMiddlewareType() == configuration.SFSWIFTV2 {
+			wrappedHandler = service.IPConfigsHandlerMiddleware.IPConfigsRequestHandlerWrapper(service.requestIPConfigHandlerHelperSF, service.releaseIPConfigHandlerHelper)
```
nddq, line 286: pretty sure you can pass a `nil` here instead of `service.releaseIPConfigHandlerHelper` if SF is not gonna use it

cns/api.go
```diff
@@ -59,6 +61,7 @@ type IPConfigsHandlerFunc func(context.Context, IPConfigsRequest) (*IPConfigsRes
 // IPConfigsHandlerMiddleware
 type IPConfigsHandlerMiddleware interface {
 	IPConfigsRequestHandlerWrapper(defaultHandler IPConfigsHandlerFunc, failureHandler IPConfigsHandlerFunc) IPConfigsHandlerFunc
+	GetMiddlewareType() configuration.SWIFTV2Mode
```
nddq, line 64: I think making this `Type()` would be sufficiently verbose here?

nddq (changes requested):

cns/api.go
kmurudi: what is the usual pattern in cns/cni repo? I mapped from our internal repo dnc to atleast use a GetType() kind of name to make the intent explicit in caller function

cns/wireserver/net.go
```diff
@@ -79,3 +124,9 @@ func calculateGatewayIP(cidr string) (net.IP, error) {
 	}
 	return gw, nil
 }
+
+func macAddressesEqual(macAddress1, macAddress2 string) bool {
+	macAddress1 = strings.ToLower(strings.ReplaceAll(macAddress1, ":", ""))
+	macAddress2 = strings.ToLower(strings.ReplaceAll(macAddress2, ":", ""))
+	return macAddress1 == macAddress2
```
jaer-tsun, line 131: compare doesn't work without replacing ':'?

cns/restserver/ipam.go
```diff
@@ -91,6 +93,103 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+		},
+		PodIPInfo: ipConfigsResp.PodIPInfo,
+	}, nil
+}
+
+func (service *HTTPRestService) updatePodInfoWithInterfaces(ctx context.Context, ipconfigResponse *cns.IPConfigsResponse) (*cns.IPConfigsResponse, error) {
+	podIPInfoList := make([]cns.PodIpInfo, 0, len(ipconfigResponse.PodIPInfo))
+	for i := range ipconfigResponse.PodIPInfo {
+		// populating podIpInfo with primary & secondary interface info & updating IpConfigsResponse
+		hostPrimaryInterface, err := service.getPrimaryHostInterface(ctx)
```
jaer-tsun, line 164: looks like this can be done outside of loop

cns/restserver/ipam.go
```diff
@@ -91,6 +93,103 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+				ReturnCode: returnCode,
+				Message:    returnMessage,
+			},
+		}, errors.New("failed to validate ip config request")
+	}
+
+	// record a pod requesting an IP
+	service.podsPendingIPAssignment.Push(podInfo.Key())
+	// unmarshal & retrieve podInfo from OrchestratorContext
+	podInfo, err := cns.NewPodInfoFromIPConfigsRequest(ipconfigsRequest)
```
jaer-tsun, line 112: this err isn't checked

cns/api.go
timraymond: @kmurudi in this case, I think @nddq is correct. You have to consider the entire signature in the name. Since this can only return a `configuration.SWIFTV2Mode`, then `Type()` is reasonably concise. If ever there is a question of "what kind of Type is it?" you can refer to the return type.

cns/wireserver/net.go
```diff
@@ -49,6 +52,48 @@ func GetPrimaryInterfaceFromResult(res *GetInterfacesResult) (*InterfaceInfo, er
...
+// Gets secondary interface details for swiftv2 secondary nics scenario
+func GetSecondaryInterfaceFromResult(res *GetInterfacesResult, macAddress string) (*InterfaceInfo, error) {
+	for _, i := range res.Interface {
+		// skip if primary
+		if i.IsPrimary {
+			continue
+		}
+
+		// skip if no subnets
+		if len(i.IPSubnet) == 0 {
```
estebancams, line 64: What is the logic here? Can we add a bigger comment?

kmurudi: this is a replica of the aboce func - `GetPrimaryInterfaceFromResult` to find similar details about secondary nics. for the interfaces in list, if subnet does not exist should skip the nic incase it's not programmed or usable for IP

cns/wireserver/net.go
kmurudi: the formats of both MACs in caller func - L68, differ in one being plain & one having ":" in the mac address string

cns/wireserver/net.go
```diff
@@ -79,3 +124,9 @@ func calculateGatewayIP(cidr string) (net.IP, error) {
 	}
 	return gw, nil
 }
+
+func macAddressesEqual(macAddress1, macAddress2 string) bool {
```
timraymond, line 128: Rather than undifferentiated strings, I'd just make:

```go
type MACAddress string

func (m MACAddress) Equal(other MACAddress) bool {
       // ...everything in here
}
```

MAC addresses are things that can be compared, validated, dissected, etc., so they deserve a type.

cns/middlewares/SFSwiftV2.go
```diff
@@ -0,0 +1,31 @@
...
+func (m *SFSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(ipRequestHandler, _ cns.IPConfigsHandlerFunc) cns.IPConfigsHandlerFunc {
+	return func(ctx context.Context, req cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
+		ipConfigsResp, err := ipRequestHandler(ctx, req)
+		if err != nil {
+			ipConfigsResp.Response.ReturnCode = types.UnexpectedError
+			return ipConfigsResp, errors.Wrapf(err, "Failed to requestIPConfigs for SF from IPConfigsRequest %v", req)
+		}
+
+		// SwiftV2-SF will always request for secondaryInterfaces for a pod
+		req.SecondaryInterfacesExist = true
```
nddq, line 24: do we need this anymore?

cns/restserver/ipam.go
```diff
@@ -91,6 +92,106 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+		return &cns.IPConfigsResponse{
+			Response: cns.Response{
+				ReturnCode: returnCode,
+				Message:    returnMessage,
+			},
+		}, errors.New("failed to validate ip config request")
+	}
+
+	// record a pod requesting an IP
+	service.podsPendingIPAssignment.Push(podInfo.Key())
```
nddq, line 109: this is only applied for pods that are getting ips from the default datapath, i don't think you need this here.

cns/restserver/ipam.go
```diff
@@ -91,6 +92,106 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+		return &cns.IPConfigsResponse{}, errors.Wrapf(err, "failed to parse IPConfigsRequest %v", ipconfigsRequest)
+	}
+	orchestratorContext, err := podInfo.OrchestratorContext()
+	if err != nil {
+		return &cns.IPConfigsResponse{}, fmt.Errorf("error getting orchestrator context from PodInfo %w", err)
+	}
+	cnsRequest := cns.GetNetworkContainerRequest{OrchestratorContext: orchestratorContext}
+	resp := service.getAllNetworkContainerResponses(cnsRequest) //nolint:contextcheck // not passed in any methods, appease linter
+	// return err if nil - error should be failed due to no nc response above
+	if resp == nil {
```
nddq, line 122: `service.getAllNetworkContainerResponses` returns a list, i dont think you can check for nil here

cns/restserver/ipam.go
```diff
@@ -91,6 +92,106 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+				ReturnCode: returnCode,
+				Message:    returnMessage,
+			},
+		}, errors.New("failed to validate ip config request")
+	}
+
+	// record a pod requesting an IP
+	service.podsPendingIPAssignment.Push(podInfo.Key())
+	// unmarshal & retrieve podInfo from OrchestratorContext
+	podInfo, err := cns.NewPodInfoFromIPConfigsRequest(ipconfigsRequest)
```
nddq, line 111: you already got podInfo from `service.validateIPConfigsRequest` above

cns/restserver/util.go
```diff
@@ -790,6 +790,24 @@ func (service *HTTPRestService) getPrimaryHostInterface(ctx context.Context) (*w
...
+		res, err := service.wscli.GetInterfaces(ctx)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get interfaces from wireserver client")
+		}
+		secondary, err := wireserver.GetSecondaryInterfaceFromResult(res, macAddress)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get secondary interface from wireserver client")
+		}
+
+		service.state.secondaryInterface = secondary
```
nddq, line 806: can the cached interfaceInfo ever be outdated?

cns/restserver/ipam.go
```diff
@@ -91,6 +92,106 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+		MacAddress:                      resp[0].NetworkInterfaceInfo.MACAddress,
+		NICType:                         resp[0].NetworkInterfaceInfo.NICType,
+		SkipDefaultRoutes:               false,
+		NetworkContainerPrimaryIPConfig: resp[0].IPConfiguration,
+	}
+	ipConfigsResp := &cns.IPConfigsResponse{
+		Response: cns.Response{
+			ReturnCode: types.Success,
+		},
+		PodIPInfo: []cns.PodIpInfo{},
```
nddq, line 141: can embed the podIPInfo here right away instead of doing append below

cns/restserver/ipam.go
```diff
@@ -91,6 +92,106 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+	}
+
+	ipConfigsResp.PodIPInfo = append(ipConfigsResp.PodIPInfo, podIPInfo)
+	ipConfigsResp, err = service.updatePodInfoWithInterfaces(ctx, ipConfigsResp)
```
nddq, line 145: `service.updatePodInfoWithInterfaces` takes in a pointer to `ipConfigsResp` already. You don't need to return another pointer and can instead modify the ipConfigsResp within the function and check for any err returned.

cns/restserver/ipam.go
```diff
@@ -91,6 +92,106 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+	if err != nil {
+		return &cns.IPConfigsResponse{
+			Response: cns.Response{
+				ReturnCode: types.FailedToAllocateIPConfig,
+				Message:    fmt.Sprintf("AllocateIPConfig failed while updating pod with interfaces: %v, IP config request is %v", err, ipconfigsRequest),
+			},
+		}, err
+	}
+	return &cns.IPConfigsResponse{
```
nddq, line 154: you already have a `ipConfigsResp`, you don't have to create a new one

cns/restserver/ipam.go
```diff
@@ -91,6 +92,106 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+		Response: cns.Response{
+			ReturnCode: types.Success,
+		},
+		PodIPInfo: ipConfigsResp.PodIPInfo,
+	}, nil
+}
+
+func (service *HTTPRestService) updatePodInfoWithInterfaces(ctx context.Context, ipconfigResponse *cns.IPConfigsResponse) (*cns.IPConfigsResponse, error) {
```
nddq, line 162: can make this `func (service *HTTPRestService) updatePodInfoWithInterfaces(ctx context.Context, ipconfigResponse *cns.IPConfigsResponse) error` and do whatever we want with `ipconfigResponse` within this func since it is being passed in as a pointer already.

nddq (changes requested):

cns/restserver/ipam.go
kmurudi: needs test+discuss

cns/restserver/ipam.go
kmurudi: will add check for zero list length.

cns/restserver/ipam.go
kmurudi: confirm

cns/restserver/ipam.go
kmurudi: ack, need to return relevant errors

cns/middlewares/SFSwiftV2.go
kmurudi: removed

cns/restserver/util.go
kmurudi: removing this field as we discussed it would change for each call

cns/wireserver/net.go
kmurudi: @timraymond I agree with this if there is more use-case for this type than just for string comparison here..the main caller of this method comes from ipam.go(L160) where it passes a field  `ipconfigResponse.PodIPInfo[i].MacAddress` so I don't want to convert this to a new type MACAddress & back to strings inside for comparison. If this was used for more methods or part of a struct I could add it, but it seems more convert-to&back just for one use.

rbtr: we should absolutely not be using strings here, there is a stdlib MAC type: https://pkg.go.dev/net#HardwareAddr

kmurudi: this refactor means it would have to be refactored at the caller level too, should we change the type for MacAddress field in PodIpInfo? - 
image

kmurudi: let me try to explore this library to compare the two macs instead of this function

kmurudi: I tried to use the package& parse the mac string but unfortunately the format in which wireserver api returns the MACAddress is not parsed by this package as well - 

@rbtr do you suggest any other package that could format this for comparison

rbtr: this was recently raised actually https://github.com/golang/go/issues/66682
Generally, we should be converting this sort of thing to a type at the boundary of its usage so that we can use the types without worrying about format differences and special cases every time we use it.
If the res.Interface.MacAddress is of type net.HardwareAddr already, you don't have to parse it here at all 😉 We can consolidate any special handling of MAC parsing in to a single place and simply compare HardwareAddr == HardwareAddr here

cns/middlewares/SFSwiftV2.go
```diff
@@ -0,0 +1,29 @@
...
+
+type SFSWIFTv2Middleware struct{}
+
+// IPConfigsRequestHandlerWrapper is the middleware function for handling SWIFT v2 IP config requests for SF standalone scenario. This function wraps the default SWIFT request
+// and release IP configs handlers.
+func (m *SFSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(ipRequestHandler, _ cns.IPConfigsHandlerFunc) cns.IPConfigsHandlerFunc {
+	return func(ctx context.Context, req cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
+		ipConfigsResp, err := ipRequestHandler(ctx, req)
+		if err != nil {
+			ipConfigsResp.Response.ReturnCode = types.UnexpectedError
```
msvik, line 19: Is there a reason to overwrite the ReturnCode that is returned by ipRequestHandler?

cns/wireserver/net.go
```diff
@@ -49,6 +52,48 @@ func GetPrimaryInterfaceFromResult(res *GetInterfacesResult) (*InterfaceInfo, er
...
+
+			secondaryIP := ""
+			for _, ip := range s.IPAddress {
+				if !ip.IsPrimary {
+					secondaryIP = ip.Address
+					break
+				}
+			}
+			var secondaryIPs []string
+			secondaryIPs = append(secondaryIPs, secondaryIP)
```
msvik, line 84: if secondaryIP is empty at this point, do we still return success?

kmurudi: will add a check for this here, we should fail early.

cns/configuration/configuration.go
```diff
@@ -46,7 +39,7 @@ type CNSConfig struct {
 	MellanoxMonitorIntervalSecs int
 	MetricsBindAddress          string
 	ProgramSNATIPTables         bool
-	SWIFTV2Mode                 SWIFTV2Mode
+	SWIFTV2Mode                 cns.SWIFTV2Mode
```
kmurudi, line 42: use boolean EnableSwiftV2

cns/restserver/ipam.go
```diff
@@ -91,6 +92,94 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+	podIPInfoList := make([]cns.PodIpInfo, 0, len(ipconfigResponse.PodIPInfo))
+
+	// fetching primary host interface to use below for updating IPConfigsResponse
+	hostPrimaryInterface, err := service.getPrimaryHostInterface(ctx)
+	if err != nil {
+		return err
+	}
+	for i := range ipconfigResponse.PodIPInfo {
+		// populating podIpInfo with secondary interface info & updating IpConfigsResponse
+		hostSecondaryInterface, err := service.getSecondaryHostInterface(ctx, ipconfigResponse.PodIPInfo[i].MacAddress)
```
kmurudi, line 160: cni should check for mac/ipconfig...cns can just pass the goal state details - nmagent to check for secondaryip

cns/restserver/ipam.go
```diff
@@ -91,6 +92,94 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+		}
+
+		podIPInfoList = append(podIPInfoList, ipconfigResponse.PodIPInfo[i])
+	}
+	ipconfigResponse.PodIPInfo = podIPInfoList
+	return nil
+}
+
 // RequestIPConfigHandler requests an IPConfig from the CNS state
 func (service *HTTPRestService) RequestIPConfigHandler(w http.ResponseWriter, r *http.Request) {
```
kmurudi, line 184: this is called in SF/standalone only for swiftv2 so we dont need the label swiftv2-sfmode

cns/service/main.go
```diff
@@ -801,6 +801,12 @@ func main() {
 		if platform.HasMellanoxAdapter() {
 			go platform.MonitorAndSetMellanoxRegKeyPriorityVLANTag(rootCtx, cnsconfig.MellanoxMonitorIntervalSecs)
 		}
+
+		// if swiftv2 scenario is enabled, we need to initialize the Service Fabric (standalone) swiftv2 middleware to process IP configs requests
+		if cnsconfig.SWIFTV2Mode == cns.SFSWIFTV2 {
+			swiftV2Middleware := &middlewares.SFSWIFTv2Middleware{}
+			httpRestService.AttachIPConfigsHandlerMiddleware(swiftV2Middleware)
```
kmurudi, line 808: initialize by default

cns/middlewares/SFSwiftV2.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/types"
+	"github.com/pkg/errors"
+)
+
+type SFSWIFTv2Middleware struct{}
```
kmurudi, line 11: rename middleware

cns/wireserver/net.go
kmurudi: addressed

cns/wireserver/net.go
kmurudi: added check to return interface only if its non-empty

cns/middlewares/standaloneSwiftV2.go
```diff
@@ -0,0 +1,29 @@
...
+type StandaloneSWIFTv2Middleware struct{}
+
+// IPConfigsRequestHandlerWrapper is the middleware function for handling SWIFT v2 IP config requests for SF standalone scenario. This function wraps the default SWIFT request
+// and release IP configs handlers.
+func (m *StandaloneSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(ipRequestHandler, _ cns.IPConfigsHandlerFunc) cns.IPConfigsHandlerFunc {
+	return func(ctx context.Context, req cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
+		ipConfigsResp, err := ipRequestHandler(ctx, req)
+		if err != nil {
+			ipConfigsResp.Response.ReturnCode = types.UnexpectedError
+			return ipConfigsResp, errors.Wrapf(err, "Failed to requestIPConfigs for SF from IPConfigsRequest %v", req)
```
msvik, line 20: nit pick: StandaloneSwiftV2

cns/service/main.go
```diff
@@ -813,6 +813,11 @@ func main() {
 		if platform.HasMellanoxAdapter() {
 			go platform.MonitorAndSetMellanoxRegKeyPriorityVLANTag(rootCtx, cnsconfig.MellanoxMonitorIntervalSecs)
 		}
+
+		// if swiftv2 scenario is enabled, we need to initialize the Service Fabric (standalone) swiftv2 middleware to process IP configs requests
+		// we don't need a config check here since this middleware will be called only for swiftv2 path for standalone
```
msvik, line 818: Please mention here that it is CNI that will call this middleware for Swift v2 path. CNI has a swiftv2 flag.
For AKS swiftv2, different handler gets called.

cns/restserver/ipam.go
```diff
@@ -91,6 +92,80 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+				ReturnCode: types.FailedToAllocateIPConfig,
+				Message:    fmt.Sprintf("AllocateIPConfig failed due to not getting NC Response from statefile, IP config request is %v", ipconfigsRequest),
+			},
+		}, ErrGetAllNCResponseEmpty
+	}
+	podIPInfo := cns.PodIpInfo{
+		PodIPConfig:                     resp[0].IPConfiguration.IPSubnet,
+		MacAddress:                      resp[0].NetworkInterfaceInfo.MACAddress,
+		NICType:                         resp[0].NetworkInterfaceInfo.NICType,
+		SkipDefaultRoutes:               false,
```
msvik, line 127: just curious why is this false?

cns/NetworkContainerContract.go
```diff
@@ -477,9 +477,10 @@ type PodIpInfo struct {
 }
 
 type HostIPInfo struct {
-	Gateway   string
-	PrimaryIP string
-	Subnet    string
+	Gateway     string
+	PrimaryIP   string
+	SecondaryIP string
```
msvik, line 482: Is this needed as I don't see it being used ?

cns/restserver/util.go
```diff
@@ -756,12 +756,12 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ctx context.Context,
 }
 
 func (service *HTTPRestService) validateIPConfigsRequest(ctx context.Context, ipConfigsRequest cns.IPConfigsRequest) (cns.PodInfo, types.ResponseCode, string) {
-	if service.state.OrchestratorType != cns.KubernetesCRD && service.state.OrchestratorType != cns.Kubernetes {
+	if service.state.OrchestratorType != cns.KubernetesCRD && service.state.OrchestratorType != cns.Kubernetes && service.state.OrchestratorType != cns.ServiceFabric {
 		return nil, types.UnsupportedOrchestratorType, "ReleaseIPConfig API supported only for kubernetes orchestrator"
```
msvik, line 760: nit pick: message doesn't match the if statement

cns/NetworkContainerContract.go
paulyufan2, line 482: As we discussed, we do not need this secondaryIP right?

cns/restserver/ipam.go
```diff
@@ -179,8 +254,15 @@ func (service *HTTPRestService) RequestIPConfigsHandler(w http.ResponseWriter, r
...
 	// Check if IPConfigsHandlerMiddleware is set
 	if service.IPConfigsHandlerMiddleware != nil {
-		// Wrap the default datapath handlers with the middleware
-		wrappedHandler := service.IPConfigsHandlerMiddleware.IPConfigsRequestHandlerWrapper(service.requestIPConfigHandlerHelper, service.ReleaseIPConfigHandlerHelper)
+		// Wrap the default datapath handlers with the middleware depending on middleware type
+		var wrappedHandler cns.IPConfigsHandlerFunc
+		switch service.IPConfigsHandlerMiddleware.Type() {
+		case cns.K8sSWIFTV2:
+			wrappedHandler = service.IPConfigsHandlerMiddleware.IPConfigsRequestHandlerWrapper(service.requestIPConfigHandlerHelper, service.ReleaseIPConfigHandlerHelper)
+		case cns.StandaloneSWIFTV2:
```
paulyufan2, line 262: can we add a comment if it's standalone swiftv2, which path we should go and why we do not need to have releaseIPConfigHandlerHelper called

cns/restserver/ipam.go
```diff
@@ -27,6 +27,7 @@ var (
 	ErrNoNCs                  = errors.New("no NCs found in the CNS internal state")
 	ErrOptManageEndpointState = errors.New("CNS is not set to manage the endpoint state")
 	ErrEndpointStateNotFound  = errors.New("endpoint state could not be found in the statefile")
+	ErrGetAllNCResponseEmpty  = errors.New("failed to get all NC responses from statefile")
```
paulyufan2, line 30: NIT: errors.New("failed to get NC responses from statefile")

cns/restserver/ipam.go
```diff
@@ -91,6 +92,80 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+	if len(resp) == 0 {
+		return &cns.IPConfigsResponse{
+			Response: cns.Response{
+				ReturnCode: types.FailedToAllocateIPConfig,
+				Message:    fmt.Sprintf("AllocateIPConfig failed due to not getting NC Response from statefile, IP config request is %v", ipconfigsRequest),
+			},
+		}, ErrGetAllNCResponseEmpty
+	}
+	podIPInfo := cns.PodIpInfo{
+		PodIPConfig:                     resp[0].IPConfiguration.IPSubnet,
```
paulyufan2, line 124: why we get podIPInfo from resp with index 0? Should we check this instead of hardcoding index 0?

cns/restserver/ipam.go
```diff
@@ -91,6 +92,80 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+	// return err if returned list has no NCs
+	if len(resp) == 0 {
+		return &cns.IPConfigsResponse{
+			Response: cns.Response{
+				ReturnCode: types.FailedToAllocateIPConfig,
+				Message:    fmt.Sprintf("AllocateIPConfig failed due to not getting NC Response from statefile, IP config request is %v", ipconfigsRequest),
+			},
+		}, ErrGetAllNCResponseEmpty
+	}
+	podIPInfo := cns.PodIpInfo{
```
paulyufan2, line 123: Yes, for delegatedNIC, CNI needs macaddress, nicType, primaryIP and subnet

cns/NetworkContainerContract.go
kmurudi: yes i can remove it, was just waiting for your review & if other fields are good enough

cns/restserver/ipam.go
```diff
@@ -91,6 +92,80 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
...
+				ReturnCode: types.FailedToAllocateIPConfig,
+				Message:    fmt.Sprintf("AllocateIPConfig failed while updating pod with interfaces: %v, IP config request is %v", err, ipconfigsRequest),
+			},
+		}, err
+	}
+	return ipConfigsResp, nil
+}
+
+func (service *HTTPRestService) updatePodInfoWithInterfaces(ctx context.Context, ipconfigResponse *cns.IPConfigsResponse) error {
+	podIPInfoList := make([]cns.PodIpInfo, 0, len(ipconfigResponse.PodIPInfo))
```
paulyufan2, line 150: NIT: should be podIPInfoList := make([]cns.PodIpInfo,  len(ipconfigResponse.PodIPInfo))? len(ipconfigResponse.PodIPInfo) should be the length of this podIPInfoList? Should we put this to the capacity field?

cns/restserver/ipam.go
kmurudi: included

cns/restserver/ipam.go
kmurudi: make(cns.podinfo, 0, len) -> means it starts with an empty list but sets a capacity of len(ipconfigResponse.PodIpInfo)

cns/service/main.go
```diff
@@ -813,6 +813,11 @@ func main() {
 		if platform.HasMellanoxAdapter() {
 			go platform.MonitorAndSetMellanoxRegKeyPriorityVLANTag(rootCtx, cnsconfig.MellanoxMonitorIntervalSecs)
 		}
+
+		// if swiftv2 scenario is enabled, we need to initialize the ServiceFabric(standalone) swiftv2 middleware to process IPConfigsRequests
+		// we don't need a cns-config check here since this middleware will be called only by CNI for SwiftV2 flow
+		swiftV2Middleware := &middlewares.StandaloneSWIFTv2Middleware{}
+		httpRemoteRestService.AttachIPConfigsHandlerMiddleware(swiftV2Middleware)
```
tamilmani1989, line 820: move closer to place where k8s middleware is intialized?

cns/service/main.go
```diff
@@ -813,6 +813,11 @@ func main() {
 		if platform.HasMellanoxAdapter() {
 			go platform.MonitorAndSetMellanoxRegKeyPriorityVLANTag(rootCtx, cnsconfig.MellanoxMonitorIntervalSecs)
 		}
+
+		// if swiftv2 scenario is enabled, we need to initialize the ServiceFabric(standalone) swiftv2 middleware to process IPConfigsRequests
+		// we don't need a cns-config check here since this middleware will be called only by CNI for SwiftV2 flow
```
tamilmani1989, line 818:
```suggestion

		// if swiftv2 scenario is enabled, we need to initialize the ServiceFabric(standalone) swiftv2 middleware to process IPConfigsRequests
		// `RequestIPConfigs` will be invoked only for swiftv2 or swiftv1 k8s scenarios. For swiftv1 direct mode different api will be invoked. So initializing this middleware always under `direct` mode should not affect any other scenarios.
```

cns/restserver/util.go
```diff
@@ -756,12 +756,12 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ctx context.Context,
 }
 
 func (service *HTTPRestService) validateIPConfigsRequest(ctx context.Context, ipConfigsRequest cns.IPConfigsRequest) (cns.PodInfo, types.ResponseCode, string) {
-	if service.state.OrchestratorType != cns.KubernetesCRD && service.state.OrchestratorType != cns.Kubernetes {
-		return nil, types.UnsupportedOrchestratorType, "ReleaseIPConfig API supported only for kubernetes orchestrator"
+	if service.state.OrchestratorType != cns.KubernetesCRD && service.state.OrchestratorType != cns.Kubernetes && service.state.OrchestratorType != cns.ServiceFabric {
```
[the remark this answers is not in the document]
tamilmani1989: do we really need this validation as it supported for both k8s and standalone case

cns/hnsclient/hnsclient_windows.go
```diff
@@ -27,8 +27,9 @@ const (
 	ExtHnsNetworkGwAddress = "192.168.255.1"
 
 	// HNS network types
-	hnsL2Bridge = "l2bridge"
-	hnsL2Tunnel = "l2tunnel"
+	hnsL2Bridge    = "l2bridge"
+	hnsL2Tunnel    = "l2tunnel"
+	hnsTransparent = "transparent"
```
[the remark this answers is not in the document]
paulyufan2: we do not need to check this I think

cns/restserver/util.go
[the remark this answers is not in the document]
paulyufan2: Yes, we have to have this validation in SF

cns/restserver/ipam.go
```diff
@@ -91,6 +92,82 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
 	}, nil
 }
 
+// requestIPConfigHandlerHelperSF validates the request, assign IPs and return the IPConfigs
+func (service *HTTPRestService) requestIPConfigHandlerHelperSF(ctx context.Context, ipconfigsRequest cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
```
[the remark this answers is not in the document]
paulyufan2: changed it to  requestIPConfigHandlerHelperStandalone

cns/restserver/util.go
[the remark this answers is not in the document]
kmurudi: better to have all supported cases here

cns/service/main.go
kmurudi: i dont see if this is valid outside the "direct channel" block, K8s middleware is initialized inside CRD channel. currently this looks fine in this section, we can address this later when Jackie takes up the PR?

cns/restserver/util.go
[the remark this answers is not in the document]
tamilmani1989: i still don't get the reason why its required to have this validation check. what's the impact if we remove this validation?

cns/hnsclient/hnsclient_windows.go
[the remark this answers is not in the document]
tamilmani1989: why cns need to know about transparent network?.. hns network is created by CNI and not by CNS

cns/restserver/util.go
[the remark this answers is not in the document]
paulyufan2: if we remove this check, then when we test in the SF environment, it will generate this issue ""ReleaseIPConfig API supported only for kubernetes orchestrator""

cns/hnsclient/hnsclient_windows.go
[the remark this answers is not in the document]
paulyufan2: just double checked this...looks like we do not need this check

cns/restserver/util.go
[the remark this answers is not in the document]
tamilmani1989: no what happens if we remove condition all together as it supported both for k8s and standalone? whats the significance of this check? cc: @nddq

cns/hnsclient/hnsclient_windows.go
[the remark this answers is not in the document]
tamilmani1989: can you remove this change then?

cns/restserver/util.go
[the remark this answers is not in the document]
nddq: this predates my time here, but I'm guessing we used to not support other orchestrator type within this data path, and this is used to prevent them to invoking this API. But with the introduction of the middleware, other orchestrators can define their own data path to get and release IP configs, so you are right that we might be able get away with removing this all together? correct me if im wrong @rbtr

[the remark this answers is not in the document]
rbtr: > what's the impact if we remove this validation?

IMO, nothing, and we can remove it entirely. 
If you're running CNS in a scenario where you don't need to call Request/ReleaseIP, and you call it...that's mostly on you. I don't need extra validation in this handler to make sure you have started CNS in the right Mode. You're going to get _an error_ whether we check Mode here or not. And if we are gonna keep growing the list of acceptable modes in this check, we should just let it pass by default and let CNS fail the request because something else like the payload will be wrong.
The better way to accomplish this sort of Mode/API surface gating would be to only enable the Request/ReleaseIP routes when we are running in a supported Mode. Then the rest of the time, those APIs would just not be wired up and would 404. But our custom server mux is not that adjustable...

msvik (approved):

cns/restserver/util.go
```diff
@@ -756,12 +756,8 @@ func (service *HTTPRestService) SendNCSnapShotPeriodically(ctx context.Context,
 }
 
 func (service *HTTPRestService) validateIPConfigsRequest(ctx context.Context, ipConfigsRequest cns.IPConfigsRequest) (cns.PodInfo, types.ResponseCode, string) {
```
nddq, line 758: i think we can get rid of this function for good. All that it is actually validating now is whether the `ipConfigsRequest.OrchestratorContext` is nil and returning the podInfo, and even then, `cns.NewPodInfoFromIPConfigsRequest` will throw an error if the the orchestrator context is empty anyways. Let's move the getting podInfo part outside to where it is needed.

paulyufan2: Hi @nddq  I think for some funcs, they need to return errorCode and errorMessage; for example:
requestIPConfigHandlerHelper()

nddq (approved):

tamilmani1989 (approved):

cns/wireserver/net.go
rbtr: @kmurudi upstream now supports parsing Windows formatted MACs, can we unwind or simplify any change added here?

PR #2641, opened 2024-03-07
chore: switch to true enable exact pod name match

pjohnst5:
A while back, this `enableExactPodNameMatch` was made for ACI when they first onboarded to SWIFT
ACI no longer uses kubernetes, so this is not relevant anymore


pjohnst5:
Won't do this since it would affect current partners, but there is a repair item assigned to TM that is to fix the logic of `enableExactMatchPodName`

PR #2647, opened 2024-03-15
fix: Fixing Vnet Scale For Cillium to set IMDS NAT using HostPrimaryIP during Cluster Creation

nairashu:
fix: Fixing Vnet Scale For Cillium to set IMDS NAT using HostPrimaryIP during Cluster Creation

**Reason for Change**:
fix: Fixing Vnet Scale For Cillium to set IMDS NAT using HostPrimaryIP during Cluster Creation

**Issue Fixed**:
https://dev.azure.com/msazure/One/_workitems/edit/27315294

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ endif
 # Interrogate the git repo and set some variables
 REPO_ROOT				 = $(shell git rev-parse --show-toplevel)
 REVISION				?= $(shell git rev-parse --short HEAD)
-ACN_VERSION				?= $(shell git describe --exclude "azure-ipam*" --exclude "dropgz*" --exclude "zapai*" --tags --always)-asn
+ACN_VERSION				?= $(shell git describe --exclude "azure-ipam*" --exclude "dropgz*" --exclude "zapai*" --tags --always)
 AZURE_IPAM_VERSION		?= $(notdir $(shell git describe --match "azure-ipam*" --tags --always))
 CNI_VERSION				?= $(ACN_VERSION)
 CNI_DROPGZ_VERSION		?= $(notdir $(shell git describe --match "dropgz*" --tags --always))
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ endif
 # Interrogate the git repo and set some variables
 REPO_ROOT				 = $(shell git rev-parse --show-toplevel)
 REVISION				?= $(shell git rev-parse --short HEAD)
-ACN_VERSION				?= $(shell git describe --exclude "azure-ipam*" --exclude "dropgz*" --exclude "zapai*" --tags --always)
+ACN_VERSION				?= $(shell git describe --exclude "azure-ipam*" --exclude "dropgz*" --exclude "zapai*" --tags --always)-asn
 AZURE_IPAM_VERSION		?= $(notdir $(shell git describe --match "azure-ipam*" --tags --always))
 CNI_VERSION				?= $(ACN_VERSION)
 CNI_DROPGZ_VERSION		?= $(notdir $(shell git describe --match "dropgz*" --tags --always))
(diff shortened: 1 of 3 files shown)


cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -92,6 +92,15 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 		PrefixLength: uint8(subnetPrefix.Bits()),
 	}
 
+	if nc.Type == v1alpha.VNETBlock {
+		nodeIP, nerr := netip.ParsePrefix(nc.NodeIP)
+		if nerr != nil {
+			return nil, errors.Wrapf(err, "IP: %s", nc.NodeIP)
+		}
+		subnet.IPAddress = nodeIP.Addr().String()
```
thatmattlong, line 100: using node IP here kind of breaks the premise of podsubnet that you can have different policies for pods and nodes, I think. If you SNAT some traffic against the node IP, then rules you put on the podsubnet would not get applied to that traffic vs if you SNAT against an IP in the pod subnet. Do we want to make this change in behavior?

thatmattlong: For example if a customer has a NSG rule on the podsubnet to block traffic to some peered vnet, the podsubnet traffic would get SNAT'ed with Node IP here and the traffic would reach. So customer thinks they've blocked the traffic from pods to peered vnet, but the traffic would still work.

tamilmani1989: afaik in podsubnet, we snat only if traffic is destined to azure dns inside guest. so it won't be a problem right?

Makefile
```diff
@@ -34,7 +34,7 @@ endif
 # Interrogate the git repo and set some variables
 REPO_ROOT				 = $(shell git rev-parse --show-toplevel)
 REVISION				?= $(shell git rev-parse --short HEAD)
-ACN_VERSION				?= $(shell git describe --exclude "azure-ipam*" --exclude "dropgz*" --exclude "zapai*" --tags --always)
+ACN_VERSION				?= $(shell git describe --exclude "azure-ipam*" --exclude "dropgz*" --exclude "zapai*" --tags --always)-asn
```
nairashu, line 37: Will remove these edits after testing.

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -92,9 +92,23 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 		PrefixLength: uint8(subnetPrefix.Bits()),
 	}
 
+	// if nc.Type == v1alpha.VNETBlock {
```
nairashu, line 95: Will remove this commented code after testing

cns/kubecontroller/nodenetworkconfig/conversion.go
nairashu: @tamilmani1989 : So does the peering traffic flow directly through NMAgent route tables on the Vnet backbone without SNAT in case of peering?

Makefile
nairashu: Removed

cns/kubecontroller/nodenetworkconfig/conversion.go
nairashu: Removed

cns/kubecontroller/nodenetworkconfig/conversion.go
thatmattlong: Yes I was mixing up overlay and podsubnet snat scenarios, in podsubnet it is only snatted with primary IP of NC for azure dns traffic, I can't imagine that customers would want different DNS policies on node and pod subnets, but I think it's still worth calling out in AKS release notes as a breaking change

nairashu: Sounds good. Will add it to the release notes for CNS

neaggarwMS (changes requested):

nairashu:
Abandoning this PR as these changes have been combined with another in the PR here: https://github.com/Azure/azure-container-networking/pull/2660

PR #2649, opened 2024-03-18
ci: Refactor ipamAddResult to remove explicit default InterfaceInfo calls

jpayne3506:
**Reason for Change**:

Remove hardcoded interactions as part of an effort to move to a more dynamic and flexible code base to support future feature work.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/invoker_azure.go
+++ b/cni/network/invoker_azure.go
@@ -47,7 +47,7 @@ func NewAzureIpamInvoker(plugin *NetPlugin, nwInfo *network.NetworkInfo) *AzureI
 }
 
 func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, error) {
-	addResult := IPAMAddResult{}
+	addResult := IPAMAddResult{interfaceInfo: make(map[string]network.InterfaceInfo)}
 
 	if addConfig.nwCfg == nil {
 		return addResult, invoker.plugin.Errorf("nil nwCfg passed to CNI ADD, stack: %+v", string(debug.Stack()))
@@ -75,8 +75,8 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 
 	defer func() {
 		if err != nil {
-			if len(addResult.interfaceInfo[0].IPConfigs) > 0 {
-				if er := invoker.Delete(&addResult.interfaceInfo[0].IPConfigs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
+			if len(addResult.interfaceInfo) > 0 && len(addResult.interfaceInfo[string(cns.InfraNIC)].IPConfigs) > 0 {
+				if er := invoker.Delete(&addResult.interfaceInfo[string(cns.InfraNIC)].IPConfigs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
 					err = invoker.plugin.Errorf("Failed to clean up IP's during Delete with error %v, after Add failed with error %w", er, err)
 				}
 			} else {
@@ -116,7 +116,7 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 		routes[i] = network.RouteInfo{Dst: route.Dst, Gw: route.GW}
 	}
 
-	addResult.interfaceInfo = append(addResult.interfaceInfo, network.InterfaceInfo{
+	addResult.interfaceInfo[string(cns.InfraNIC)] = network.InterfaceInfo{
 		IPConfigs: ipconfigs,
 		Routes:    routes,
 		DNS: network.DNSInfo{
@@ -124,7 +124,7 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 			Servers: result.DNS.Nameservers,
 		},
 		NICType: cns.InfraNIC,
-	})
+	}
 	return addResult, err
 }
 
--- a/cni/network/invoker_azure.go
+++ b/cni/network/invoker_azure.go
@@ -75,8 +75,10 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 
 	defer func() {
 		if err != nil {
-			if len(addResult.defaultInterfaceInfo.IPConfigs) > 0 {
-				if er := invoker.Delete(&addResult.defaultInterfaceInfo.IPConfigs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
+			// defaultIndex := findDefaultInterface(addResult)
+			// Can hard code 0 here as it everything is new. Afriad of creating leaks if I complicate the logic
+			if len(addResult.interfaceInfo[0].IPConfigs) > 0 {
+				if er := invoker.Delete(&addResult.interfaceInfo[0].IPConfigs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
 					err = invoker.plugin.Errorf("Failed to clean up IP's during Delete with error %v, after Add failed with error %w", er, err)
 				}
 			} else {
@@ -116,8 +118,7 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 		routes[i] = network.RouteInfo{Dst: route.Dst, Gw: route.GW}
 	}
 
-	addResult.defaultInterfaceInfo = network.InterfaceInfo{IPConfigs: ipconfigs, Routes: routes, DNS: network.DNSInfo{Suffix: result.DNS.Domain, Servers: result.DNS.Nameservers}, NICType: cns.InfraNIC}
-
+	addResult.interfaceInfo = append(addResult.interfaceInfo, network.InterfaceInfo{IPConfigs: ipconfigs, Routes: routes, DNS: network.DNSInfo{Suffix: result.DNS.Domain, Servers: result.DNS.Nameservers}, NICType: cns.InfraNIC})
 	return addResult, err
 }
 
--- a/cni/network/invoker_azure.go
+++ b/cni/network/invoker_azure.go
@@ -75,8 +75,6 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 
 	defer func() {
 		if err != nil {
-			// defaultIndex := findDefaultInterface(addResult)
-			// Can hard code 0 here as it everything is new. Afriad of creating leaks if I complicate the logic
 			if len(addResult.interfaceInfo[0].IPConfigs) > 0 {
 				if er := invoker.Delete(&addResult.interfaceInfo[0].IPConfigs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
 					err = invoker.plugin.Errorf("Failed to clean up IP's during Delete with error %v, after Add failed with error %w", er, err)
@@ -118,7 +116,15 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 		routes[i] = network.RouteInfo{Dst: route.Dst, Gw: route.GW}
 	}
 
-	addResult.interfaceInfo = append(addResult.interfaceInfo, network.InterfaceInfo{IPConfigs: ipconfigs, Routes: routes, DNS: network.DNSInfo{Suffix: result.DNS.Domain, Servers: result.DNS.Nameservers}, NICType: cns.InfraNIC})
+	addResult.interfaceInfo = append(addResult.interfaceInfo, network.InterfaceInfo{
+		IPConfigs: ipconfigs,
+		Routes:    routes,
+		DNS: network.DNSInfo{
+			Suffix:  result.DNS.Domain,
+			Servers: result.DNS.Nameservers,
+		},
+		NICType: cns.InfraNIC,
+	})
 	return addResult, err
 }
 
--- a/cni/network/invoker_azure_test.go
+++ b/cni/network/invoker_azure_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/Azure/azure-container-networking/cni"
 	"github.com/Azure/azure-container-networking/cni/log"
+	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/ipam"
 	"github.com/Azure/azure-container-networking/network"
 	cniSkel "github.com/containernetworking/cni/pkg/skel"
@@ -239,15 +240,15 @@ func TestAzureIPAMInvoker_Add(t *testing.T) {
 			}
 
 			for _, ifInfo := range ipamAddResult.interfaceInfo {
-				switch string(ifInfo.NICType) {
-				case "DelegatedVMNIC":
-					// Secondary
-				case "BackendNIC":
-					// todo
-				default:
-					// Default | InfraNIC
+				switch ifInfo.NICType {
+				case cns.DelegatedVMNIC, cns.BackendNIC:
+					fmt.Print(errInvalidNIC)
+					require.FailNow("No coverage for this NICType")
+				case cns.InfraNIC:
 					fmt.Printf("want:%+v\nrest:%+v\n", tt.want, ifInfo.IPConfigs)
 					require.Exactly(tt.want, ifInfo.IPConfigs)
+				default:
+					require.FailNow("Unsupported NICType")
 				}
 			}
 		})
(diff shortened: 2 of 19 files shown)


cni/network/network.go
```diff
@@ -457,7 +457,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		if len(ipamAddResults) > 1 && !plugin.isDualNicFeatureSupported(args.Netns) {
 			errMsg := fmt.Sprintf("received multiple NC results %+v from CNS while dualnic feature is not supported", ipamAddResults)
 			logger.Error("received multiple NC results from CNS while dualnic feature is not supported",
-				zap.Any("results", ipamAddResult))
+				zap.Any("results", ipamAddResult)) // <- should this be Result(S)?
```
paulyufan2, line 460: I found zap logger won't print ipamAddResult or an object/list of objects; instead it can only print its field

cni/network/network.go
```diff
@@ -518,18 +519,20 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 
 		ipamAddConfig := IPAMAddConfig{nwCfg: nwCfg, args: args, options: options}
 		if !nwCfg.MultiTenancy {
+			// what does this .Add() do?
```
paulyufan2, line 522: This is to retrieve ipamAddResult from ipamInvoker which is swiftv2 path; my PR is to move this part before

cni/network/network.go
```diff
@@ -1404,3 +1417,15 @@ func convertCniResultToInterfaceInfo(result *cniTypesCurr.Result) network.Interf
 
 	return interfaceInfo
 }
+
+func findDefaultInterface(ipamAddResult IPAMAddResult) int {
```
paulyufan2, line 1421: let's return Bool here

network/manager.go
```diff
@@ -382,10 +382,10 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 	}
 
 	if nw.VlanId != 0 {
-		// the first entry in epInfo is InfraNIC type
-		if epInfo[0].Data[VlanIDKey] == nil {
+		// leverage passing default index, if failed, call function to find default
```
paulyufan2, line 385: where do we call function to find default index here

jpayne3506: This was a note for me to try a function call if my original design does not work.

cni/network/network.go
jpayne3506: if we don't return an int then we have to loop through for every `defaultInterface` code change.

cni/network/network.go
```diff
@@ -793,28 +808,37 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
...
+					IPAddresses:       addresses,
+					Routes:            secondaryCniResult.Routes,
+					MacAddress:        secondaryCniResult.MacAddress,
+					NICType:           secondaryCniResult.NICType,
+					SkipDefaultRoutes: secondaryCniResult.SkipDefaultRoutes,
+				})
+		case cns.BackendNIC:
+			// todo
+		default:
+			// InfraNic
```
jpayne3506, line 833: reassign `defaultIndex` here. the index for `interfaceInfo` and `epInfo` could potentially be different. 

Thanks @QxBytes

network/endpoint_linux.go
```diff
@@ -58,14 +58,15 @@ func (nw *network) newEndpointImpl(
...
+	_ int,
 ) (*endpoint, error) {
 	var (
 		err           error
 		hostIfName    string
 		contIfName    string
 		localIP       string
 		vlanid        = 0
-		defaultEpInfo = epInfo[0]
+		defaultEpInfo = epInfo[0] // this needs to be changed, out of scope...
```
jpayne3506, line 69: Out of scope for l1vh work, but can still use `defaultIndex`.

cni/network/invoker_azure_test.go
```diff
@@ -238,8 +238,18 @@ func TestAzureIPAMInvoker_Add(t *testing.T) {
 				require.Nil(err)
 			}
 
-			fmt.Printf("want:%+v\nrest:%+v\n", tt.want, ipamAddResult.defaultInterfaceInfo.IPConfigs)
-			require.Exactly(tt.want, ipamAddResult.defaultInterfaceInfo.IPConfigs)
+			for _, ifInfo := range ipamAddResult.interfaceInfo {
+				switch string(ifInfo.NICType) {
```
jpayne3506, line 242: Should I bring in the cns package?

jpayne3506:
test comment

cni/network/invoker_azure_test.go
vipul-21: Any reason why we didn't ? Given if CNS NIC type change or a new NIC type is added we need to update this test as well ?

cni/network/network.go
```diff
@@ -1404,3 +1427,14 @@ func convertCniResultToInterfaceInfo(result *cniTypesCurr.Result) network.Interf
...
 }
+
+func findDefaultInterface(ipamAddResult IPAMAddResult) int {
+	for i, ifInfo := range ipamAddResult.interfaceInfo {
+		if ifInfo.NICType == cns.InfraNIC {
+			return i
+		}
+	}
+
+	// error or nil here
```
vipul-21, line 1438: And a log as well either way.

cni/network/network.go
```diff
@@ -793,28 +807,39 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
 
 	epInfos := []*network.EndpointInfo{&epInfo}
 	// get secondary interface info
-	for _, secondaryCniResult := range opt.ipamAddResult.secondaryInterfacesInfo {
-		var addresses []net.IPNet
-		for _, ipconfig := range secondaryCniResult.IPConfigs {
-			addresses = append(addresses, ipconfig.Address)
-		}
+	for i, secondaryCniResult := range opt.ipamAddResult.interfaceInfo {
```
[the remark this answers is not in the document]
vipul-21: Maybe we should change those too then IMO(not in the PR though)

cni/network/network.go
```diff
@@ -1404,3 +1427,14 @@ func convertCniResultToInterfaceInfo(result *cniTypesCurr.Result) network.Interf
...
+
+func findDefaultInterface(ipamAddResult IPAMAddResult) int {
+	for i, ifInfo := range ipamAddResult.interfaceInfo {
+		if ifInfo.NICType == cns.InfraNIC {
+			return i
+		}
+	}
+
+	// error or nil here
+	return -1
```
vipul-21, line 1439: We will need an err i suppose, as we are using the `int` for indexing afterwards. It will be out of bounds for `-1`

cni/network/invoker_azure.go
```diff
@@ -75,8 +75,8 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 
 	defer func() {
 		if err != nil {
-			if len(addResult.defaultInterfaceInfo.IPConfigs) > 0 {
-				if er := invoker.Delete(&addResult.defaultInterfaceInfo.IPConfigs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
+			if len(addResult.interfaceInfo[0].IPConfigs) > 0 {
+				if er := invoker.Delete(&addResult.interfaceInfo[0].IPConfigs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
```
vipul-21, line 79: Are we sure the first added interface is the default one ? 
In case of multiple adds calls as well  (trying to understand this better.)

cni/network/invoker_cns.go
```diff
@@ -175,9 +175,16 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 				return IPAMAddResult{}, err
 			}
 		default:
+			// Assign default index
+			if defaultIndex == -1 {
+				// add a default Interface
+				// would put within configureDefaultAddResult() but we have a dependency on numInterfacesWithDefaultRoutes
+				addResult.interfaceInfo = append(addResult.interfaceInfo, network.InterfaceInfo{NICType: cns.InfraNIC})
+				defaultIndex = len(addResult.interfaceInfo) - 1
```
vipul-21, line 183: Correct me if I am wrong,
The `addresult.interfaceInfo` will be empty if `defaultindex` is -1, so we append the result. 
This `default` case will be called multiple times or once ? (Asking because of the condition `deafultIndex == -1`
 Why don't we use the `findDefaultIndex` function here ?

jpayne3506: > Why don't we use the findDefaultIndex function here ? + The addresult.interfaceInfo will be empty if defaultindex is -1, so we append the result.

We could, but considering that `addResult.interFaceInfo[]` has not been assigned an InfraNIC we can get away with not using it. Saving some cycles of iterating through the array. The `-1` here is being used as a short lived bool.

I can't really think of another way to have an InfraNIC interface created only once, but not erroneously with only 1 variable.

> This default case will be called multiple times or once ? (Asking because of the condition deafultIndex == -1

Multiple.

jpayne3506: > I can't really think of another way to have an InfraNIC interface created only once, but not erroneously with only 1 variable.

Actually, this might be possible with a map as we get an err if the key does not exist, but as it stands, I don't think we have anything that we can use as a standard unique key throughout the codebase.

cni/network/invoker_azure.go
jpayne3506: There is only 1 interface added to `IPAMAddResult` in this function.

cni/network/network.go
jpayne3506: added an error. Don't see a need to add a CNI log here as we capture the `InfraNIC` in multiple ways.

cni/network/network.go
jpayne3506: I am leaving in `-1` as I don't want to pass in an erroneous, but useable index.

cni/network/invoker_cns.go
```diff
@@ -175,9 +175,16 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 				return IPAMAddResult{}, err
 			}
 		default:
+			// Assign default index
+			if ifIndex == -1 {
+				// add a default Interface
+				// would put within configureDefaultAddResult() but we have a dependency on numInterfacesWithDefaultRoutes
+				addResult.interfaceInfo = append(addResult.interfaceInfo, network.InterfaceInfo{NICType: cns.InfraNIC})
```
jaer-tsun, line 182: shouldn't this come from CNS response?

cni/network/network.go
```diff
@@ -569,14 +590,14 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		}
 
 		var epInfo network.EndpointInfo
-		epInfo, err = plugin.createEndpointInternal(&createEndpointInternalOpt)
+		epInfo, err = plugin.createEndpointInternal(&createEndpointInternalOpt, ifIndex)
```
jaer-tsun, line 593: maybe we can pass interface info instead of indexing everywhere

cni/network/network.go
```diff
@@ -1404,3 +1438,13 @@ func convertCniResultToInterfaceInfo(result *cniTypesCurr.Result) network.Interf
 
 	return interfaceInfo
 }
+
+func findDefaultInterface(ipamAddResult IPAMAddResult) (int, error) {
```
jaer-tsun, line 1442: maybe return interface info instead of index

cni/network/invoker_azure_test.go
jpayne3506: Not sure. I could not find historical context through the PRs.

cni/network/invoker_cns.go
jpayne3506: Because we only want to create a single infra nic. Until we make another PR where I can use maps to check if the infra nic exists within addResult I am leveraging `ifIndex` to be both an index and a flag.

cni/network/network.go
jpayne3506: The only reason why I am doing it this way is for performance. We have to index for infra nic in multiple locations otherwise. 

As stated in earlier conversations, having to loop through the interfaceInfos and passing index could be removed if we change it from an array into a map. I am saving that for another PR as the scope for this one is just to remove `defaultInterfaceInfo` and `secondayInterfaceInfos`

cni/network/network.go
jpayne3506: If that is something that we want to do instead it can be done. Having the index is more lightweight than passing an entire struct through the flow. Would rather have the next PR change interfaceInfo into a map and then we no longer need to search for infra nic.

cni/network/invoker_cns.go
jaer-tsun: but CNS response will only have 1 infra nic too no?

cni/network/invoker_azure_test.go
tamilmani1989: @jpayne3506 yes we should do it instead of defining same thing in multiple places

cni/network/invoker_azure.go
```diff
@@ -75,8 +75,8 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
 
 	defer func() {
 		if err != nil {
-			if len(addResult.defaultInterfaceInfo.IPConfigs) > 0 {
-				if er := invoker.Delete(&addResult.defaultInterfaceInfo.IPConfigs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
+			if len(addResult.interfaceInfo[0].IPConfigs) > 0 {
```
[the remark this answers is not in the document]
jpayne3506: This `defer()` will not be ran until an interface has been appended. If any code after the defer() fails then we will have a panic that will skip the defer unless we want to leverage `recover()`.

cni/network/invoker_cns.go
jpayne3506: It will have 1 infra nic, but multiple podIPInfo associated when it is dualstack scenario.

tamilmani1989 (changes requested):

cni/network/invoker_cns.go
```diff
@@ -175,9 +175,16 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 				return IPAMAddResult{}, err
 			}
 		default:
+			// Assign default index
+			if ifIndex == -1 {
```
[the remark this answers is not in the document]
jpayne3506: Remove ifIndex, in favor of map. use macAddress.

cni/network/invoker_cns.go
```diff
@@ -384,15 +391,20 @@ func configureDefaultAddResult(info *IPResultInfo, addConfig *IPAMAddConfig, add
 		}
 	}
 
+	ifIndex, err := findDefaultInterface(*addResult)
```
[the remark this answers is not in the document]
jpayne3506: pass in interface.

cni/network/network_windows.go
```diff
@@ -157,8 +157,15 @@ func (plugin *NetPlugin) getNetworkName(netNs string, ipamAddResult *IPAMAddResu
...
+		ifIndex, err := findDefaultInterface(*ipamAddResult)
+		if err != nil {
+			logger.Error("Error finding InfraNIC interface",
+				zap.Error(err))
+			return "", errors.Wrap(err, "cns did not return an InfraNIC")
+		}
 		// networkName will look like ~ azure-vlan1-172-28-1-0_24
-		ipAddrNet := ipamAddResult.defaultInterfaceInfo.IPConfigs[0].Address
+		ipAddrNet := ipamAddResult.interfaceInfo[ifIndex].IPConfigs[0].Address
 		prefix, err := netip.ParsePrefix(ipAddrNet.String())
```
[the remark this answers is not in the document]
jpayne3506: consume interfaceInfo instead of searching for infraNic

network/endpoint.go
```diff
@@ -143,19 +143,20 @@ func (nw *network) newEndpoint(
 	nsc NamespaceClientInterface,
 	iptc ipTablesClient,
 	epInfo []*EndpointInfo,
```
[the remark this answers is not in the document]
jpayne3506: This is will be addressed with @QxBytes future ci

network/endpoint_linux.go
```diff
@@ -58,15 +58,14 @@ func (nw *network) newEndpointImpl(
...
-	epIndex int,
 ) (*endpoint, error) {
 	var (
 		err           error
 		hostIfName    string
 		contIfName    string
 		localIP       string
 		vlanid        = 0
-		defaultEpInfo = epInfo[epIndex]
+		defaultEpInfo = epInfo[0] // always first index of epinfo[]. Will be removed
```
jpayne3506, line 68: Will be removed with @QxBytes PR

jpayne3506:
Closing in favor of #2688

PR #2650, opened 2024-03-19
Start two CNS servers with different handlers

paulyufan2:
**Reason for Change**:

This PR is to start two CNS servers:
1.Localhost echo server => it will use default URL for CNI func handlers
2.HTTP legacy server => it will use private VM's ip or IP that customer provides for DNC/CNS func handlers

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/restserver/api.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/logger"
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/cns/wireserver"
+	acn "github.com/Azure/azure-container-networking/common"
```
nddq, line 23: why is this `acn` and can't be `common` instead?

cns/service/main.go
```diff
@@ -856,7 +857,15 @@ func main() {
...
 			return
 		}
+	}
 
+	logger.Printf("[Azure CNS] Start HTTP local echo server")
+	httpEchoRestService := v2.New(httpRestService)
+	if httpEchoRestService != nil {
+		err = httpEchoRestService.Start(z, cns.DefaultAPIServerURL)
+		if err != nil {
+			return
```
nddq, line 867: should log the error here first and then return

cns/restserver/v2/server.go
```diff
@@ -39,6 +40,9 @@ func (s Server) Start(log *zap.Logger, addr string) {
 	e.GET(cns.V2Prefix+cns.DeleteHostNCApipaEndpointPath, echo.WrapHandler(http.HandlerFunc(s.DeleteHostNCApipaEndpoint)))
 
 	if err := e.Start(addr); err != nil {
-		log.Error("failed to run server", zap.Error(err))
+		log.Error("failed to run echo server", zap.Error(err))
```
nddq, line 43: we should let whoever invoke `Start` to log the error returned instead, and if they already did, then this log would be redundant. Subsequently, we don't need to pass in `*zap.Logger` in the args since the only thing we are logging now is the err

cns/service.go
```diff
@@ -48,64 +48,65 @@ func NewService(name, version, channelMode string, store store.KeyValueStore) (*
...
 
-// GetAPIServerURL returns the API server URL.
-func (service *Service) getAPIServerURL() string {
-	urls, _ := service.GetOption(acn.OptCnsURL).(string)
-	if urls == "" {
-		urls = defaultAPIServerURL
+func (service *Service) AddListener(config *common.ServiceConfig) error {
+	// Fetch and parse the API server URL.
+	var nodeURL *url.URL
+	cnsURL, _ := service.GetOption(acn.OptCnsURL).(string)
```
nddq, line 54: is there a reason why we are not checking for err here?

cns/service.go
```diff
@@ -48,64 +48,65 @@ func NewService(name, version, channelMode string, store store.KeyValueStore) (*
...
+	if cnsURL == "" {
+		// get VM primary interface's private IP
+		nodeURL, _ = url.Parse(fmt.Sprintf("tcp://%s:%s", config.PrimaryInterfaceIP, defaultAPIServerPort))
+	} else {
+		// use the URL that customer provides
+		nodeURL, _ = url.Parse(cnsURL)
```
nddq, line 60: here as well

cns/service.go
```diff
@@ -48,64 +48,65 @@ func NewService(name, version, channelMode string, store store.KeyValueStore) (*
...
-	urls, _ := service.GetOption(acn.OptCnsURL).(string)
-	if urls == "" {
-		urls = defaultAPIServerURL
+func (service *Service) AddListener(config *common.ServiceConfig) error {
+	// Fetch and parse the API server URL.
+	var nodeURL *url.URL
+	cnsURL, _ := service.GetOption(acn.OptCnsURL).(string)
+	if cnsURL == "" {
+		// get VM primary interface's private IP
+		nodeURL, _ = url.Parse(fmt.Sprintf("tcp://%s:%s", config.PrimaryInterfaceIP, defaultAPIServerPort))
```
nddq, line 57: and here

cns/service/main.go
```diff
@@ -43,6 +43,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/multitenantcontroller"
 	"github.com/Azure/azure-container-networking/cns/multitenantcontroller/multitenantoperator"
 	"github.com/Azure/azure-container-networking/cns/restserver"
+	v2 "github.com/Azure/azure-container-networking/cns/restserver/v2"
```
nddq, line 46: i think we can rename this to `restserverv2`

nddq (changes requested):

cns/service.go
paulyufan2: GetOption does not return error

timraymond: @paulyufan2 The type assertion returns an `ok` value here. If that's `false` here, `cnsURL` will be the empty string which will lead to a silent misbehavior for a DRI to diagnose rather than an actionable error with immediate resolution.

cns/service/main.go
```diff
@@ -43,6 +43,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/multitenantcontroller"
 	"github.com/Azure/azure-container-networking/cns/multitenantcontroller/multitenantoperator"
 	"github.com/Azure/azure-container-networking/cns/restserver"
+	restserver2 "github.com/Azure/azure-container-networking/cns/restserver/v2"
```
nddq, line 46: should be restserverv2

cns/service/main.go
```diff
@@ -856,20 +858,32 @@ func main() {
 			logger.Errorf("Failed to start multiTenantController, err:%v.\n", err)
 			return
 		}
+	}
 
+	logger.Printf("[Azure CNS] Start HTTP local echo server")
+	httpEchoRestService := restserver2.New(httpRestService)
```
nddq, line 864: nit: instead of calling this `httpEchoRestService`, how about `localSrv`? since this is serving local requests, and then we can make the `httpRestService` to be `remoteSrv`

cns/service/main.go
```diff
@@ -856,20 +858,32 @@ func main() {
...
+	if httpEchoRestService != nil {
+		go func() {
```
nddq, line 866: im concerned about putting these servers' `Start` func within goroutines without passing in a context to prevent leaks.

cns/service/main.go
```diff
@@ -856,20 +858,32 @@ func main() {
...
 		if cnsconfig.EnablePprof {
 			httpRestService.RegisterPProfEndpoints()
 		}
 
-		err = httpRestService.Start(&config)
-		if err != nil {
-			logger.Errorf("Failed to start CNS, err:%v.\n", err)
-			return
-		}
+		go func() {
```
nddq, line 880: Same here, didn't `httpRestService.Start` starts its own goroutine within?

cns/service/main.go
paulyufan2: maybe name httpRestService as httpLegacyService?
httpEchoRestService as httpLocalRestService?

cns/service/main.go
timraymond: I would share those concerns, @nddq

paulyufan2: yes we should add context to control this async

cns/restserver/restserver.go
```diff
@@ -185,6 +185,9 @@ func NewHTTPRestService(config *common.ServiceConfig, wscli interfaceGetter, wsp
 		return nil, errors.Wrap(err, "failed to get primary interface from IMDS response")
 	}
 
+	// add primaryInterfaceIP to cns config
+	config.PrimaryInterfaceIP = primaryInterface.PrimaryIP
```
tamilmani1989, line 189: Is it required to add a new field in service config or can we refer from `serviceState.primaryInterface.PrimaryIP` ?

cns/service/main.go
tamilmani1989: i like @nddq suggestion. Lets call it `httpLocalService` and `httpRemoteService`.

cns/service.go
```diff
@@ -48,64 +48,77 @@ func NewService(name, version, channelMode string, store store.KeyValueStore) (*
...
+		// use the URL that customer provides
+		nodeURL, err = url.Parse(cnsURL)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "Failed to parse IP that customer provides")
 		}
+	}
 
-		if config.TlsSettings.TLSPort != "" {
+	logger.Debugf("legacy server url: %+v", nodeURL)
```
tamilmani1989, line 74: Lets not call it legacy instead "CNS rest server url" or something else?

tamilmani1989 (changes requested):

cns/restserver/restserver.go
paulyufan2: yes, we need to get primary interface IP from here and then use it in AddListener(config *common.ServiceConfig)

cns/service.go
```diff
@@ -24,11 +24,16 @@ import (
...
-	defaultAPIServerURL = "tcp://localhost:10090"
-	genericData         = "com.microsoft.azure.network.generic"
+	defaultAPIServerPort = "10090"
+	genericData          = "com.microsoft.azure.network.generic"
 )
 
+var invalidURLs = map[string]string{
+	"tcp://127.0.0.1:10090": "localhost",
+	"tcp://localhost:10090": "localhost",
+	"tcp://0.0.0.0:10090":   "default",
```
tamilmani1989, line 34: what if its 0.0.0.0:10091? imo, should skip this invalid url check.. if user passes cns url, then we should just use it

tamilmani1989 (changes requested):

cns/service/main.go
```diff
@@ -841,35 +843,48 @@ func main() {
...
-			httpRestService.RegisterPProfEndpoints()
+			httpRemoteRestService.RegisterPProfEndpoints()
 		}
 
-		err = httpRestService.Start(&config)
-		if err != nil {
-			logger.Errorf("Failed to start CNS, err:%v.\n", err)
-			return
-		}
+		go func() {
```
nddq, line 881: we should not start a goroutine here since `httpRemoteRestService.Start` will start its own goroutine within

cns/restserver/v2/server.go
```diff
@@ -17,28 +19,38 @@ func New(s *restserver.HTTPRestService) *Server {
...
+	e.POST(cns.V2Prefix+cns.CreateHostNCApipaEndpointPath, echo.WrapHandler(http.HandlerFunc(s.CreateHostNCApipaEndpoint)))
+	e.POST(cns.V2Prefix+cns.DeleteHostNCApipaEndpointPath, echo.WrapHandler(http.HandlerFunc(s.DeleteHostNCApipaEndpoint)))
 
 	if err := e.Start(addr); err != nil {
-		log.Error("failed to run server", zap.Error(err))
+		logger.Errorf("failed to run echo server due to %+v", err)
+		return errors.Wrap(err, "failed to start echo server")
 	}
+
+	// after context is done, shutdown echo server
```
tamilmani1989, line 48: nit: update comment 
```suggestion
	// after context is done, shutdown local server
```

cns/service/main.go
```diff
@@ -841,34 +852,53 @@ func main() {
...
+
+	// if user does not provide cns url by -c option, then start http local server
+	// TODO: we will deprecated -c option in next phase and start local server in any case
+	if config.EnableLocalServer {
+		logger.Printf("[Azure CNS] Start HTTP local server")
+
+		httpLocalRestService := restserverv2.New(httpRemoteRestService)
+		if httpLocalRestService != nil {
+			go func() {
+				err = httpLocalRestService.Start(rootCtx, defaultAPIServerURL)
```
tamilmani1989, line 896: aren't we honoring customer specified port via `-p`?

tamilmani1989 (changes requested):

cns/restserver/api_test.go
```diff
@@ -1720,12 +1720,17 @@ func startService() error {
...
+		// mock localhost as primary interface IP
+		config.PrimaryInterfaceIP = "localhost"
+
 		err = service.Init(&config)
 		if err != nil {
 			logger.Errorf("Failed to Init CNS, err:%v.\n", err)
 			return err
 		}
 
+		service := service.(*HTTPRestService)
```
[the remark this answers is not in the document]
paulyufan2: Removed, mixed with one of previous change

cns/restserver/v2/server.go
```diff
@@ -17,28 +19,38 @@ func New(s *restserver.HTTPRestService) *Server {
...
+	<-ctx.Done()
```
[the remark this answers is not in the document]
paulyufan2: yes

cns/service.go
```diff
@@ -48,64 +47,93 @@ func NewService(name, version, channelMode string, store store.KeyValueStore) (*
...
-
-// Initialize initializes the service and starts the listener.
-func (service *Service) Initialize(config *common.ServiceConfig) error {
-	log.Debugf("[Azure CNS] Going to initialize a service with config: %+v", config)
-
-	// Initialize the base service.
-	if err := service.Service.Initialize(config); err != nil {
-		return errors.Wrap(err, "failed to initialize")
-	}
+func (service *Service) AddListener(config *common.ServiceConfig) error {
```
tamilmani1989, line 50: can we add UT for this api and validate all possible options?

cns/service/main.go
```diff
@@ -841,34 +853,60 @@ func main() {
...
-			httpRestService.RegisterPProfEndpoints()
+			httpRemoteRestService.RegisterPProfEndpoints()
 		}
 
-		err = httpRestService.Start(&config)
-		if err != nil {
-			logger.Errorf("Failed to start CNS, err:%v.\n", err)
-			return
+		go func() {
+			err = httpRemoteRestService.Start(&config)
```
nddq, line 881: im pretty sure `Start` is not blocking, we can remove the go func wrapping around it here.

cns/restserver/v2/server_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func TestStartServerWithCNSPort(t *testing.T) {
+	var err error
+
+	logger.InitLogger("testlogs", 0, 0, "./")
+	cnsPort := "8000"
+
+	// Create the service with -p 8000
+	if err = startService(cnsPort, ""); err != nil {
+		t.Errorf("Failed to start CNS Service. Error: %v", err)
```
tamilmani1989, line 30:
```suggestion
		t.Errorf("Failed to connect to CNS Service on expected port:%s. Error: %v", cnsPort, err)
```

cns/restserver/v2/server_test.go
```diff
@@ -0,0 +1,100 @@
...
+func TestStartServerWithCNSURL(t *testing.T) {
+	var err error
+
+	logger.InitLogger("testlogs", 0, 0, "./")
+
+	// Create the service with -c "localhost:8000"
+	if err = startService("", "tcp://localhost:8500"); err != nil {
+		t.Errorf("Failed to start CNS Service. Error: %v", err)
+		os.Exit(1)
+	}
```
tamilmani1989, line 44:
```suggestion
       url := "tcp://localhost:8500"
	if err = startService("",  url); err != nil {
		t.Errorf("Failed to connect to CNS Service by this url:%s. Error: %v", url, err)
		os.Exit(1)
	}
```

cns/restserver/v2/server_test.go
```diff
@@ -0,0 +1,101 @@
...
+func TestStartServerWithCNSPort(t *testing.T) {
+	var err error
+
+	logger.InitLogger("testlogs", 0, 0, "./")
+	cnsPort := "8000"
+
+	// Create the service with -p 8000
+	if err = startService(cnsPort, ""); err != nil {
+		t.Errorf("Failed to connect to CNS Service on expected port:%s. Error: %v", cnsPort, err)
+		os.Exit(1)
```
timraymond, line 31: You should use Fatalf here instead and omit `os.Exit`. It shuts down the test less violently.

```suggestion
		t.Fatalf("Failed to connect to CNS Service on expected port:%s. Error: %v", cnsPort, err)
```

cns/restserver/v2/server_test.go
```diff
@@ -0,0 +1,101 @@
...
+func TestStartServerWithCNSURL(t *testing.T) {
+	var err error
+
+	logger.InitLogger("testlogs", 0, 0, "./")
+
+	// Create the service with -c "localhost:8000"
+	cnsURL := "tcp://localhost:8500"
+	if err = startService("", cnsURL); err != nil {
+		t.Errorf("Failed to connect to CNS Service by this cns url:%s. Error: %v", cnsURL, err)
+		os.Exit(1)
```
timraymond, line 44: Same here.

timraymond (changes requested):

nddq (approved):

cns/restserver/v2/server.go
```diff
@@ -17,28 +19,38 @@ func New(s *restserver.HTTPRestService) *Server {
...
+	e.POST(cns.RequestIPConfig, echo.WrapHandler(restserver.NewHandlerFuncWithHistogram(s.RequestIPConfigHandler, restserver.HTTPRequestLatency)))
+	e.POST(cns.RequestIPConfigs, echo.WrapHandler(restserver.NewHandlerFuncWithHistogram(s.RequestIPConfigsHandler, restserver.HTTPRequestLatency)))
+	e.POST(cns.ReleaseIPConfig, echo.WrapHandler(restserver.NewHandlerFuncWithHistogram(s.ReleaseIPConfigHandler, restserver.HTTPRequestLatency)))
+	e.POST(cns.ReleaseIPConfigs, echo.WrapHandler(restserver.NewHandlerFuncWithHistogram(s.ReleaseIPConfigsHandler, restserver.HTTPRequestLatency)))
+	e.POST(cns.PathDebugIPAddresses, echo.WrapHandler(http.HandlerFunc(s.HandleDebugIPAddresses)))
+	e.POST(cns.PathDebugPodContext, echo.WrapHandler(http.HandlerFunc(s.HandleDebugPodContext)))
+	e.POST(cns.PathDebugRestData, echo.WrapHandler(http.HandlerFunc(s.HandleDebugRestData)))
+	e.POST(cns.GetNetworkContainerByOrchestratorContext, echo.WrapHandler(http.HandlerFunc(s.GetNetworkContainerByOrchestratorContext)))
+	e.POST(cns.GetAllNetworkContainers, echo.WrapHandler(http.HandlerFunc(s.GetAllNetworkContainers)))
```
tamilmani1989, line 33: @nddq / @rbtr is all these api's POST call?  can you confirm. I see `GetAllNetworkContainers` use GET method from CNI

tamilmani1989: nvm i was checking 1.4.x version of CNS `GetAllNcsFromCNS` api.

cns/restserver/v2/server.go
```diff
@@ -17,28 +19,38 @@ func New(s *restserver.HTTPRestService) *Server {
...
-		log.Error("failed to run server", zap.Error(err))
+		logger.Errorf("failed to run echo server due to %+v", err)
+		return errors.Wrap(err, "failed to start echo server")
 	}
+
+	// after context is done, shutdown local server
+	<-ctx.Done()
+	if err := e.Shutdown(ctx); err != nil {
+		logger.Errorf("failed to shutdown echo server due to %+v", err)
+		return errors.Wrap(err, "failed to shutdown echo server")
```
rbtr, line 52: this probably doesn't need an if statement, just a return, since we either error and return err or shutdown and return err (except err is nil)

rbtr (approved):
seems reasonable to me

cns/service.go
```diff
@@ -48,64 +47,93 @@ func NewService(name, version, channelMode string, store store.KeyValueStore) (*
...
+	var (
+		err     error
+		nodeURL *url.URL
+	)
```
timraymond, line 54: It generally causes problems to predeclare variables like this. It's better to define them at the point of use using the `:=` operator.

cns/service.go
```diff
@@ -48,64 +47,93 @@ func NewService(name, version, channelMode string, store store.KeyValueStore) (*
...
+
+	// if cnsURL is empty the VM primary interface IP will be used
+	// if customer specifies -c option, then use this URL with warning message and it will be deprecated soon
+	cnsURL, _ := service.GetOption(acn.OptCnsURL).(string)
```
timraymond, line 58: Blocking since the unhandled type assertion is still not resolved.

cns/service.go
```diff
@@ -48,64 +47,93 @@ func NewService(name, version, channelMode string, store store.KeyValueStore) (*
...
+	// if customer provides port number by -p option, then use VM IP with this port and localhost server also uses this port
+	// otherwise it will use defaultAPIServerPort 10090
+	cnsPort, _ := service.GetOption(acn.OptCnsPort).(string)
```
timraymond, line 61: Same here. Please handle the type assertion failure.

cns/service.go
```diff
@@ -48,64 +47,93 @@ func NewService(name, version, channelMode string, store store.KeyValueStore) (*
...
+
+	if cnsURL == "" {
+		config.Server.EnableLocalServer = true
```
timraymond, line 64: I generally think it's bad practice to write back to a config received from the environment--that should be strictly immutable.

cns/service.go
timraymond, line 50: Much of this is massaging the config into derived configuration, right? It would probably make more sense to define methods on the config to handle that.

cns/service.go
paulyufan2: We will get rid of this flag once -c option is deprecated

cns/restserver/v2/server_test.go
```diff
@@ -0,0 +1,98 @@
...
+	var err error
+
+	logger.InitLogger("testlogs", 0, 0, "./")
+
+	// Create the service with -c "localhost:8000"
+	cnsURL := "tcp://localhost:8500"
+	if err = startService("", cnsURL); err != nil {
+		t.Fatalf("Failed to connect to CNS Service by this cns url:%s. Error: %v", cnsURL, err)
+	}
+}
```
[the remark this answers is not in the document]
paulyufan2: here we only mock this new server start; api_test.go includes all tests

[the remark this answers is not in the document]
timraymond: The point still stands though. These tests are testing `startService`, not the CNS server. `startService` internally contains far more degrees of freedom that are not expressed in the test cases here.

cns/restserver/v2/server.go
```diff
@@ -17,28 +19,38 @@ func New(s *restserver.HTTPRestService) *Server {
...
-	e.GET(cns.ReleaseIPConfig, echo.WrapHandler(restserver.NewHandlerFuncWithHistogram(s.ReleaseIPConfigHandler, restserver.HTTPRequestLatency)))
-	e.GET(cns.ReleaseIPConfigs, echo.WrapHandler(restserver.NewHandlerFuncWithHistogram(s.ReleaseIPConfigsHandler, restserver.HTTPRequestLatency)))
-	e.GET(cns.PathDebugIPAddresses, echo.WrapHandler(http.HandlerFunc(s.HandleDebugIPAddresses)))
-	e.GET(cns.PathDebugPodContext, echo.WrapHandler(http.HandlerFunc(s.HandleDebugPodContext)))
-	e.GET(cns.PathDebugRestData, echo.WrapHandler(http.HandlerFunc(s.HandleDebugRestData)))
-	e.GET(cns.GetNetworkContainerByOrchestratorContext, echo.WrapHandler(http.HandlerFunc(s.GetNetworkContainerByOrchestratorContext)))
-	e.GET(cns.GetAllNetworkContainers, echo.WrapHandler(http.HandlerFunc(s.GetAllNetworkContainers)))
-	e.GET(cns.CreateHostNCApipaEndpointPath, echo.WrapHandler(http.HandlerFunc(s.CreateHostNCApipaEndpoint)))
-	e.GET(cns.DeleteHostNCApipaEndpointPath, echo.WrapHandler(http.HandlerFunc(s.DeleteHostNCApipaEndpoint)))
+	e.POST(cns.RequestIPConfig, echo.WrapHandler(restserver.NewHandlerFuncWithHistogram(s.RequestIPConfigHandler, restserver.HTTPRequestLatency)))
```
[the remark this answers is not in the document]
paulyufan2: These functions are POST methods

cns/restserver/v2/server.go
paulyufan2: solved and return error directly

tamilmani1989 (dismissed):

timraymond (approved):
Removing PR block

tamilmani1989 (approved):

PR #2660, opened 2024-03-26
fix: Using Node IP as the primary IP allowing the use of all the IPs in the subnet for pods in Vnet Scale Mode and added the fix for Vnet Scale Cillium

nairashu:
**Reason for Change**:
fix: Using Node IP as the primary IP allowing the use of all the IPs in the subnet for pods in Vnet Scale Mode

**Issue Fixed**:
https://dev.azure.com/msazure/One/_workitems/edit/25432148

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/conversion.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion.go
@@ -88,13 +88,13 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 		return nil, errors.Wrapf(err, "invalid SubnetAddressSpace %s", nc.SubnetAddressSpace)
 	}
 
-	var subnet cns.IPSubnet
+	subnet := cns.IPSubnet{
+		PrefixLength: uint8(subnetPrefix.Bits()),
+	}
 	if nc.Type == v1alpha.VNETBlock {
 		subnet.IPAddress = nc.NodeIP
-		subnet.PrefixLength = uint8(subnetPrefix.Bits())
 	} else {
 		subnet.IPAddress = primaryPrefix.Addr().String()
-		subnet.PrefixLength = uint8(subnetPrefix.Bits())
 	}
 
 	req, err := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
--- a/cns/kubecontroller/nodenetworkconfig/conversion.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion.go
@@ -102,8 +102,5 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 		return nil, errors.Wrapf(err, "error while creating NC request from static NC")
 	}
 
-	if nc.Type == v1alpha.VNETBlock {
-		req.HostPrimaryIP = nc.NodeIP
-	}
 	return req, err
 }
--- a/cns/kubecontroller/nodenetworkconfig/conversion.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion.go
@@ -87,14 +87,23 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
 	if err != nil {
 		return nil, errors.Wrapf(err, "invalid SubnetAddressSpace %s", nc.SubnetAddressSpace)
 	}
-	subnet := cns.IPSubnet{
-		IPAddress:    primaryPrefix.Addr().String(),
-		PrefixLength: uint8(subnetPrefix.Bits()),
+
+	var subnet cns.IPSubnet
+	if nc.Type == v1alpha.VNETBlock {
+		subnet.IPAddress = nc.NodeIP
+		subnet.PrefixLength = uint8(subnetPrefix.Bits())
+	} else {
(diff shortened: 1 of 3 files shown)


cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,14 +87,23 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+		subnet.PrefixLength = uint8(subnetPrefix.Bits())
 	}
 
 	req, err := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error while creating NC request from static NC")
 	}
+
+	if nc.Type == v1alpha.VNETBlock {
+		req.HostPrimaryIP = nc.NodeIP
```
aggarwal0009, line 106: Curious, we don't need to set HostPrimaryIP for other modes?

aggarwal0009 (dismissed):
LGTM

cns/kubecontroller/nodenetworkconfig/conversion.go
nairashu: It is being set in the case of podsubnet in the DynamicNC Helper. For Overlay, it is not needed but noticed that for Azure Vnets using Cillium Dataplane the IMDS setup on the nodes requires this to be setup.

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,14 +87,23 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
 		return nil, errors.Wrapf(err, "invalid SubnetAddressSpace %s", nc.SubnetAddressSpace)
 	}
-	subnet := cns.IPSubnet{
-		IPAddress:    primaryPrefix.Addr().String(),
-		PrefixLength: uint8(subnetPrefix.Bits()),
+
+	var subnet cns.IPSubnet
+	if nc.Type == v1alpha.VNETBlock {
+		subnet.IPAddress = nc.NodeIP
+		subnet.PrefixLength = uint8(subnetPrefix.Bits())
```
jaer-tsun, line 94: nit: looks like subnet.PrefixLength is common for both cases

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -87,14 +87,23 @@ func CreateNCRequestFromStaticNC(nc v1alpha.NetworkContainer) (*cns.CreateNetwor
...
+		subnet.IPAddress = primaryPrefix.Addr().String()
+		subnet.PrefixLength = uint8(subnetPrefix.Bits())
 	}
 
 	req, err := createNCRequestFromStaticNCHelper(nc, primaryPrefix, subnet)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error while creating NC request from static NC")
 	}
+
+	if nc.Type == v1alpha.VNETBlock {
```
jaer-tsun, line 105: can this block be moved to createNCRequestFromStaticNCHelper?

cns/kubecontroller/nodenetworkconfig/conversion.go
nairashu: Fixed

cns/kubecontroller/nodenetworkconfig/conversion.go
nairashu: The Request is getting created in the helper on the return statement for common code between Overlay and VnetScale. So after that seemed the right fit to update which would be this line, so that it happens for both Windows and Linux.

jaer-tsun: could just create that req top of the func since secondaryIPConfigs is the only thing that gets updated, then update  with  nc.NodeIP in the VNETBlock

nairashu: Trying to remove this block and set the value in the createNCRequestFromStaticNCHelper method for both overlay and Vnet Scale

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
```diff
@@ -48,6 +46,7 @@ func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPre
 	}
 
 	return &cns.CreateNetworkContainerRequest{
+		HostPrimaryIP:        nc.NodeIP,
```
nairashu, line 49: Based on offline discussion with @thatmattlong : "anything not in pod cidr would get snatted with node IP in overlay so there is no need to do an explicit snat for IMDS "

jaer-tsun (approved):
lgtm

thatmattlong (dismissed):
Approved since they passed datapath E2Es 
```
• [SLOW TEST:1183.015 seconds]
CNI Overlay
pkg/e2ev2/runnerginkgo/ginkgo.go:89
  Azure CNI Overlay Datapath Tests
  pkg/e2ev2/runnerginkgo/runner.go:210
------------------------------
 
Ran 1 of 1 Specs in 1183.017 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
```

thatmattlong (approved):

PR #2665, opened 2024-04-01
feat: Adds an NC status check for more refined error messages.

abel2-code:
**Reason for Change**:
This adds an NC status check so that more actionable error messages can be given.

**Issue Fixed**:
N/A

**Requirements**:
- [ ] includes documentation
- [X] adds unit tests

**Notes**:
This is part of my intern project to improve error messaging.


cns/restserver/ipam.go
```diff
@@ -34,6 +35,19 @@ const (
 	InterfaceName     = "eth0"
 )
 
+func checkNCStatuses(ncStatuses map[string]containerstatus) (string, bool) {
```
timraymond, line 38: I feel like this should be a type... something like:

```go
type multicontainerstatus map[string]containerstatus
```

The `string` returned from this can be implemented by `(*multicontainerstatus).String()`, and the `bool` can be given meaning by implementing a `(*multicontainerstatus).AnyFailed()` method or something.

abel2-code: Updated

cns/restserver/ipam.go
```diff
@@ -34,6 +35,30 @@ const (
 	InterfaceName     = "eth0"
 )
 
+func (mcs *multicontainerstatus) String() string {
```
nairashu, line 38: What is the purpose of this method and why is the method named String()? It looks like you are appending unsuccessful error messages together for different NC IDs

cns/restserver/restserver.go
```diff
@@ -123,6 +123,8 @@ type containerstatus struct {
 	VfpUpdateComplete             bool // True when VFP programming is completed for the NC
 }
 
+type multicontainerstatus map[string]containerstatus
```
nairashu, line 126: Is this new code that you introduced?

nairashu: Camel Casing

cns/restserver/ipam.go
nairashu: Let's change this method name to GetUnsuccessfulStatuses

cns/restserver/ipam.go
```diff
@@ -34,6 +35,30 @@ const (
...
+
+func (mcs *multicontainerstatus) AnyFailed() bool {
+	for _, status := range *mcs {
+		if status.CreateNetworkContainerRequest.NCStatus != v1alpha.NCUpdateSuccess {
+			return true
+		}
+	}
+
+	return false
+}
```
nairashu, line 60: This method won't be needed as if you get 0 unsuccessfulErrorStatus Back you can infer that all NCs are in success mode.

cns/restserver/ipam.go
```diff
@@ -169,6 +210,23 @@ func (service *HTTPRestService) requestIPConfigsHandler(w http.ResponseWriter, r
 	if err != nil {
 		return
 	}
+
```
nairashu, line 213: Can we update the UT to verify these changes as well.

cns/restserver/ipam.go
```diff
@@ -110,6 +135,22 @@ func (service *HTTPRestService) requestIPConfigHandler(w http.ResponseWriter, r
 		return
 	}
 
+	// Check the status of the NC.
+	ncStatuses := multicontainerstatus(service.state.ContainerStatus)
```
nairashu, line 139: Update the UT to verify these changes as well

cns/restserver/ipam.go
abel2-code: Changed the name to better reflect the purpose of the method and moved it to the same file in which the struct is defined.

cns/restserver/ipam.go
abel2-code: Removed

cns/restserver/ipam.go
timraymond: @nairashu The method as it was follows best practices. It implements `fmt.Stringer` as documented [here](https://pkg.go.dev/fmt#Stringer). This is widely recognized by many parts of the standard library that handle string-like things. For example, you could have passed this type to `fmt.Println` like this: `fmt.Println("container status:", someMcs)` and it would identify and invoke `String()`. I would recommend reverting this rename.

cns/restserver/ipam.go
abel2-code: There aren't any existing tests that use requestIPConfigsHandler.

cns/restserver/ipam.go
abel2-code: There aren't any existing tests that use requestIPConfigHandler

cns/restserver/restserver.go
```diff
@@ -19,6 +21,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/types/bounded"
 	"github.com/Azure/azure-container-networking/cns/wireserver"
 	acn "github.com/Azure/azure-container-networking/common"
+	"github.com/Azure/azure-container-networking/crd/nodenetworkconfig/api/v1alpha"
```
rbtr, line 24: I don't love that this is being imported in to the `restserver` package. We have made it this far without having the restserver aware of the CRDs at all, let's try to keep it that way.

I think this import is only for the status codes? They should probably be defined in a higher level package (or with all our existing common types in the [CNS package](https://github.com/Azure/azure-container-networking/blob/master/cns/NetworkContainerContract.go)), then they can be used in both the restserver and the CRD packages.

rbtr (commented):
logically seems fine, my only comment is on organization

PR #2668, opened 2024-04-02
ETW logging implementation in CNI

sivakami-projects:
Feat: Features 🌈 Added support for logging to ETW using zap logger.

Reason for Change:
- Implemented logging to ETW using the zap logger to provide a unified logging experience in Azure Container Networking. 
- This feature introduces a zap writer syncer capable of writing to ETW, facilitating more efficient debugging and monitoring.

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

- The implementation introduces a zap core that directs logging output to ETW. 
- ETW logs are written for all windows machines.
- This configuration allows Azure Container Networking to log simultaneously to ETW and the file system, ensuring that logs are accessible through ETW while still being persisted to files. 
- This addition is designed to be transparent to existing logging mechanisms and does not interfere with the current logging functionality.


cni/log/logger.go
```diff
@@ -17,9 +19,10 @@ var (
 const (
 	maxLogFileSizeInMb = 5
 	maxLogFileCount    = 8
+	etwCNIEventName    = "Azure-CNI"
 )
 
-func initZapLog(logFile string) *zap.Logger {
+func initZapLog(logFile string, isEtwLoggingEnabled bool) *zap.Logger {
```
timraymond, line 25: I'm not crazy about bool options because they defy documentation. I think I'd rather this be a separate function with the common initialization factored out to a common func.

zapetw/write_syncer.go
```diff
@@ -0,0 +1,76 @@
+//go:build windows
+// +build windows
```
timraymond, line 2: If you name the file *_windows.go, this build tag should be implied by the toolchain.

zapetw/write_syncer.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+
+	"github.com/Microsoft/go-winio/pkg/etw"
+	"github.com/Microsoft/go-winio/pkg/guid"
+	"go.uber.org/zap/zapcore"
+)
+
+const providername = "Azure-Container-Networking-CCP"
+
+type EtwWriteSyncer struct {
```
timraymond, line 16: Nit: ETW should be capitalized since it's an acronym.

timraymond (commented):
Looks pretty good. Just had a few thoughts.

zapetw/unsupported_linux.go
```diff
@@ -0,0 +1,28 @@
+package zapetw
```
paulyufan2, line 1: Can you please explain what is this file used? I see we do not support Linux OS now and you have same functions but write noETWEvent support in this file. Do we need this file? Or the filename should be core_linux.go and maybe put some comments here?

zapetw/write_syncer_windows.go
```diff
@@ -0,0 +1,67 @@
...
+		e.eventName,
+		etw.WithEventOpts(
+			etw.WithLevel(e.etwLevel),
+		),
+		[]etw.FieldOpt{
+			etw.StringField("Message", string(p)),
+		},
+	)
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to write to ETW")
```
paulyufan2, line 41: should be errors.Wrap(err, "failed to write to ETW") since we do not need pass args here

zapetw/write_syncer_windows.go
```diff
@@ -0,0 +1,67 @@
...
+type ETWWriteSyncer struct {
+	provider  *etw.Provider
+	eventName string
+	etwLevel  etw.Level
+}
+
+func NewETWWriteSyncer(eventName string, zapLevel zapcore.Level) (*ETWWriteSyncer, error) {
+	provider, err := etw.NewProviderWithOptions(providername)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create ETW provider")
```
paulyufan2, line 20: same here

zapetw/core_windows.go
```diff
@@ -0,0 +1,39 @@
...
+		return nil, errors.Wrapf(err, "Failed to initialize ETW logger")
+	}
+	etwcore := zapcore.NewCore(getETWencoder(), zapcore.AddSync(etwSyncer), loggingLevel)
+	return etwcore, nil
+}
+
+func InitETWLogger(eventName string, loggingLevel zapcore.Level) (*zap.Logger, error) {
+	etwcore, err := getETWCore(eventName, loggingLevel)
+	if err != nil {
+		return nil, err
```
paulyufan2, line 27: can we wrap error and return errors.wrap() here as well like getETWCore()

zapetw/core_windows.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	return zap.New(etwcore, zap.AddCaller()), nil
+}
+
+func AttachETWLogger(baseLogger *zap.Logger, eventName string, loggingLevel zapcore.Level) (*zap.Logger, error) {
+	etwcore, err := getETWCore(eventName, loggingLevel)
+	if err != nil {
+		return baseLogger, err
```
paulyufan2, line 35: same here

zapetw/core_windows.go
sivakami-projects: The errors are already wrapped within the internal methods to provide specific context at their source. Initially I had chosen not to wrap the errors again.

zapetw/core_windows.go
sivakami-projects: The errors are already wrapped within the internal methods to provide specific context at their source. Initially I had chosen not to wrap the errors again.

zapetw/core_windows.go
```diff
@@ -0,0 +1,39 @@
...
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+func getETWencoder() zapcore.Encoder {
+	encoderConfig := zap.NewProductionEncoderConfig()
+	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	return zapcore.NewJSONEncoder(encoderConfig)
+}
+
```
rbtr, line 14: similar comment - this is standard zap encoder code, we don't need to do it and can let the caller do it

zapetw/core_windows.go
```diff
@@ -0,0 +1,39 @@
...
+func getETWCore(eventName string, loggingLevel zapcore.Level) (zapcore.Core, error) {
+	etwSyncer, err := NewETWWriteSyncer(eventName, loggingLevel)
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to initialize ETW logger core")
+	}
+	etwcore := zapcore.NewCore(getETWencoder(), zapcore.AddSync(etwSyncer), loggingLevel)
```
rbtr, line 20: what if I don't want to use the JSON encoder from `getETWencoder`?

zapetw/core_windows.go
rbtr, line 1: I think this whole file can probably go away

zapetw/write_syncer_windows.go
```diff
@@ -0,0 +1,67 @@
...
+		etwLevel:  mapZapLevelToETWLevel(zapLevel),
+	}, nil
+}
+
+func (e *ETWWriteSyncer) Write(p []byte) (int, error) {
+	err := e.provider.WriteEvent(
+		e.eventName,
+		etw.WithEventOpts(
+			etw.WithLevel(e.etwLevel),
+		),
```
rbtr, line 35:
```suggestion
		[]etw.EventOpts{etw.WithLevel(e.etwLevel)},
```

zapetw/write_syncer_windows.go
```diff
@@ -0,0 +1,67 @@
...
+		[]etw.FieldOpt{
+			etw.StringField("Message", string(p)),
```
rbtr, line 37: can we just 
```suggestion
			etw.StringField(string(p)),
```

rbtr (commented):
some initial thoughts

zapetw/core_windows.go
timraymond: Agreed. One of the nice things about structured logging is that the presentation of the structure is entirely incidental. You can swap it out for different formats if you like. We lose this by assuming JSON here.

paulyufan2:
lgtm, please get approval from Evan and TM.

paulyufan2 (approved):

cni/log/logger_linux.go
```diff
@@ -1,6 +1,18 @@
...
 const (
 	// LogPath is the path where log files are stored.
 	LogPath = "/var/log/"
 )
+
+var ErrETWNotSupported = errors.New("ETW is not supported for Linux")
+
+func GetETWCore(_ string, _ zapcore.Level) (zapcore.Core, error) {
+	return nil, ErrETWNotSupported
+}
```
rbtr, line 18: It is still a design issue that I can even call `GetETWCore` from Linux code, and that it always returns an error (by design!).
But if we think about this a little more generically, it becomes much cleaner and we can forget about this error.

Abstractly, what you're doing here is joining a platform-specific logging core in to the main core. This could still be config-driven, if it needed to be, but what's important is that from the perspective of the unplatformed code, it should happen for all platforms in exactly the same way. It shouldn't be an error on Linux if I call the same method on Linux as I do on Windows.

If we stop focusing on "ETW" and shift our perspective to "platform cores"...

```suggestion
func JoinPlatformCores(c zapcore.Core) (zapcore.Core, error) {
	return c, nil
}
```

pt. 1
... continued in suggestion in `logger_windows.go`

cni/log/logger_windows.go
```diff
@@ -1,6 +1,25 @@
...
+	encoderConfig := zap.NewProductionEncoderConfig()
+	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	jsonEncoder := zapcore.NewJSONEncoder(encoderConfig)
+
+	etwcore, err := zapetw.NewETWCore(eventName, jsonEncoder, loggingLevel)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create ETW core")
+	}
+	return etwcore, nil
+}
```
rbtr, line 25:
```suggestion

func JoinPlatformCores(c zapcore.Core, loggingLevel zapcore.Level) (zapcore.Core, error) {
	etwcore, err := etwCore(ETWCNIEventName, loggingLevel)
	if err != nil {
		return c, err
	}
	teecore := zapcore.NewTee(Logger.Core(), etwcore)
	return teecore, nil
}


func etwCore(loggingLevel zapcore.Level) (zapcore.Core, error) {
	encoderConfig := zap.NewProductionEncoderConfig()
	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
	jsonEncoder := zapcore.NewJSONEncoder(encoderConfig)
	c, err := zapetw.NewETWCore("Azure-CNI", jsonEncoder, loggingLevel)
	if err != nil {
		return nil, errors.Wrap(err, "failed to create ETW core")
	}
	return c, nil
}
```

pt. 2

cni/log/logger.go
```diff
@@ -30,13 +32,25 @@ func initZapLog(logFile string) *zap.Logger {
...
+func initZapLogWithETW(logFile string) *zap.Logger {
+	Logger := initZapLog(logFile)
+	etwcore, err := GetETWCore(ETWCNIEventName, loggingLevel)
+	if err != nil {
+		Logger.Error("Failed to attach ETW core to logger", zap.Error(err))
+		return Logger
+	}
+	teecore := zapcore.NewTee(Logger.Core(), etwcore)
+	Logger = zap.New(teecore, zap.AddCaller())
+	return Logger.With(zap.Int("pid", os.Getpid()))
```
rbtr, line 49: Finally, the whole `initZapLogWithETW` func goes away because it's not a concern of the root unplatformed logging code - which only needs to know that it is going to decorate the root logger core with some platform specific cores, and does _not_ need to know anything about the implementation of those platform specific things.

```suggestion
	textfilecore := zapcore.NewCore(jsonEncoder, logFileCNIWriter, loggingLevel)
	c, err := JoinPlatformCores(textfilecore, loggingLevel)
	if err != nil { 
		// should this be a fatal error? 
		// maybe we should simply fall back to the text core?
		c = textfilecore
	}
	return zap.New(c, zap.AddCaller()).With(zap.Int("pid", os.Getpid()))
}
```

pt. 3/3

rbtr (changes requested):

cni/log/logger.go
```diff
@@ -17,6 +17,8 @@ var (
 const (
 	maxLogFileSizeInMb = 5
 	maxLogFileCount    = 8
+	ETWCNIEventName    = "Azure-CNI"
```
rbtr, line 20: nit:
```suggestion
	etwCNIEventName    = "Azure-CNI"
```
capital letters mean `public` in Go, but this is not used outside of this package

cni/log/logger.go
```diff
@@ -30,13 +32,18 @@ func initZapLog(logFile string) *zap.Logger {
...
-	Logger := zap.New(core)
-	return Logger
+	textFileCore := zapcore.NewCore(jsonEncoder, logFileCNIWriter, loggingLevel)
+	core, err := JoinPlatformCores(textFileCore, loggingLevel)
+	if err != nil {
+		// If we fail to join the platform cores, fallback to the original core.
+		core = textFileCore
+	}
+	Logger := zap.New(core, zap.AddCaller())
+	return Logger.With(zap.Int("pid", os.Getpid()))
```
rbtr, line 42: nit:
```suggestion
	return zap.New(core, zap.AddCaller()).With(zap.Int("pid", os.Getpid()))
```
no need for the extra local reference

rbtr (dismissed):
couple nits, mostly LGTM

rbtr (dismissed):

paulyufan2 (dismissed):

cni/log/logger.go
```diff
@@ -30,13 +32,17 @@ func initZapLog(logFile string) *zap.Logger {
...
-	core := zapcore.NewCore(jsonEncoder, logFileCNIWriter, zapcore.DebugLevel)
-	Logger := zap.New(core)
-	return Logger
+	textFileCore := zapcore.NewCore(jsonEncoder, logFileCNIWriter, loggingLevel)
+	core, err := JoinPlatformCores(textFileCore, loggingLevel)
+	if err != nil {
+		// If we fail to join the platform cores, fallback to the original core.
+		core = textFileCore
+	}
+	return zap.New(core, zap.AddCaller()).With(zap.Int("pid", os.Getpid()))
```
tamilmani1989, line 41: is there any difference in logging format as we moved getting pid here?

zapetw/core_windows.go
```diff
@@ -0,0 +1,89 @@
+package zapetw
+
+import (
+	"github.com/Microsoft/go-winio/pkg/etw"
+	"github.com/pkg/errors"
+	"go.uber.org/zap/zapcore"
+)
+
+// <organization_name>-<product_name>-<component_name>
+const providername = "Azure-CNI-CNS-Monitoring"
```
tamilmani1989, line 10: can it be `AzureCNI-Monitoring`. This includes cns as well

zapetw/core_windows.go
```diff
@@ -0,0 +1,89 @@
...
+	return &ETWCore{
+		provider:     core.provider,
+		eventName:    core.eventName,
+		encoder:      core.encoder,
+		LevelEnabler: core.LevelEnabler,
+		fields:       append(core.fields, fields...),
+	}
+}
+
+func (core *ETWCore) Check(entry zapcore.Entry, checkedEntry *zapcore.CheckedEntry) *zapcore.CheckedEntry {
```
tamilmani1989, line 43: it would be nice if you add a brief comment on what this function does

zapetw/core_windows.go
```diff
@@ -0,0 +1,89 @@
...
+		provider:     provider,
+		eventName:    eventName,
+		encoder:      encoder,
+		LevelEnabler: levelEnabler,
+	}, nil
+}
+
+func (core *ETWCore) With(fields []zapcore.Field) zapcore.Core {
+	return &ETWCore{
+		provider:     core.provider,
```
tamilmani1989, line 35: in geneva/kusto, do we have fields to determine where this log is coming from like VM, subscription id, region,etc?

tamilmani1989: yes lets add that including resource group(rg). VM names are not unique. subid+rg+role_instance uniquely identifies the VM

tamilmani1989 (approved):

PR #2686, opened 2024-04-15
fix: Added -n option in iptables list cmd to prevent reverse dns lookup

tamilmani1989:
cherry pick from commit id b8f6db627ae7c80b933226bd8cf3bc3dd199eb86

**Reason for Change**:

**Issue Fixed**:

Cherry pick this PR https://github.com/Azure/azure-container-networking/pull/2682

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/iptables/iptables.go
+++ b/iptables/iptables.go
@@ -109,7 +109,7 @@ func RunCmd(version, params string) error {
 
 // check if iptable chain alreay exists
 func ChainExists(version, tableName, chainName string) bool {
-	params := fmt.Sprintf("-t %s -L %s", tableName, chainName)
+	params := fmt.Sprintf("-t %s -nL %s", tableName, chainName)
 	if err := RunCmd(version, params); err != nil {
 		return false
 	}


iptables/iptables.go
```diff
@@ -109,7 +109,7 @@ func RunCmd(version, params string) error {
 
 // check if iptable chain alreay exists
 func ChainExists(version, tableName, chainName string) bool {
-	params := fmt.Sprintf("-t %s -L %s", tableName, chainName)
+	params := fmt.Sprintf("-t %s -nL %s", tableName, chainName)
```
timraymond, line 112: Is it worth documenting these flags here? I know that you can look them up from the manpage, but it might be helpful to put it in front of readers' faces.

timraymond (approved):
One quick suggestion.

iptables/iptables.go
tamilmani1989: Thanks for suggestion. This is cherry-pick PR. will add it next time in the PR created against master

PR #2688, opened 2024-04-15
refactor: code changes for stateless cni and swift v2

QxBytes:
**Reason for Change**:

Modifies CNI Add and Delete Flow

- Integrates Stateless CNI Changes (ex: populating endpoint infos based on multiple entries returned from CNS).
- Adds Singularity/L1VH changes (i.e. network deletion) if stateless cni.
- Each add call now creates a single ipamAddResult, which has a map of interface infos. Each interface info now maps 1:1 with an endpoint info, and each endpoint info leads to the creation of an endpoint.
- Removes the network info struct, merging all fields into the endpoint info struct.
- Delete takes in a container id, finds all endpoints corresponding to that container id, and deletes endpoints based on what is returned.
- Removes handling of the consecutive add case.
- Creates endpoint ids as follows: Infra NIC endpoint ids are created as before, non-infra nic endpoint ids are created with an index appended to the end
- Creates stateless cni keys as follows: Infra NIC endpoint has the key set to the args interface info, otherwise it is set to the master interface name (interface name based on the mac address of that endpoint). The stateless cni has a map of container ids to another mapping of interface names to infos. The outer map uses containerID as the key.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/log/logger.go
vipul-21, line 1: Can remove this change as not adding value.

cni/network/invoker_azure.go
```diff
@@ -69,14 +69,11 @@ func (invoker *AzureIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, er
...
 	}
-	if len(result.IPs) > 0 {
-		addResult.hostSubnetPrefix = result.IPs[0].Address
-	}
 
 	defer func() {
 		if err != nil {
-			if len(addResult.defaultInterfaceInfo.IPConfigs) > 0 {
-				if er := invoker.Delete(&addResult.defaultInterfaceInfo.IPConfigs[0].Address, addConfig.nwCfg, nil, addConfig.options); er != nil {
+			if len(addResult.interfaceInfo) > 0 && len(addResult.interfaceInfo[string(cns.InfraNIC)].IPConfigs) > 0 {
```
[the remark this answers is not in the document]
vipul-21: Is it because of this `addResult.interfaceInfo[string(info.macAddress)] ` for secondaryInterfaces ?

cni/network/multitenancy.go
```diff
@@ -202,31 +203,39 @@ func (m *Multitenancy) GetAllNetworkContainers(
...
+			HostSubnetPrefix: hostSubnetPrefixes[i],
+		}
+
+		ipconfig, routes := convertToIPConfigAndRouteInfo(ifInfo.NCResponse)
+		ifInfo.IPConfigs = []*network.IPConfig{ipconfig}
+		ifInfo.Routes = routes
+		ifInfo.NICType = cns.InfraNIC
+
+		// assuming we only assign infra nics in this function
+		ipamResult.interfaceInfo[string(ifInfo.NICType)+strconv.Itoa(i)] = ifInfo
```
vipul-21, line 235: I think this might be the reason for having it `string` and not `NICType`. 
Might be useful to have a function that returns this `key` if we need to access this in multiple places. Otherwise, we will need to change the key everywhere if this changes in the future.

cni/network/network.go
```diff
@@ -205,8 +205,21 @@ func (plugin *NetPlugin) Stop() {
 	logger.Info("Plugin stopped")
 }
 
-// FindMasterInterface returns the name of the master interface.
-func (plugin *NetPlugin) findMasterInterface(nwCfg *cni.NetworkConfig, subnetPrefix *net.IPNet) string {
+// findInterfaceByMAC returns the name of the master interface
+func (plugin *NetPlugin) findInterfaceByMAC(macAddress string) string {
+	interfaces, _ := net.Interfaces()
```
vipul-21, line 210: Why are we not acknowledging the error here, if any ?

cni/network/network.go
```diff
@@ -313,14 +326,39 @@ func addNatIPV6SubnetInfo(nwCfg *cni.NetworkConfig,
...
+	networkID, err := plugin.getNetworkName(netNs, interfaceInfo, nwCfg)
+	if err != nil {
+		return "", err
+	}
+	return networkID, nil
+}
+
+// get network info for legacy
+func (plugin *NetPlugin) getNetworkInfo(netNs string, interfaceInfo *network.InterfaceInfo, nwCfg *cni.NetworkConfig) network.EndpointInfo {
+	networkID, _ := plugin.getNetworkID(netNs, interfaceInfo, nwCfg)
```
vipul-21, line 349: same here

cns/client/client.go
```diff
@@ -1023,11 +1023,11 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error)
...
 }
 
-// GetEndpoint calls the EndpointHandlerAPI in CNS to retrieve the state of a given EndpointID
-func (c *Client) GetEndpoint(ctx context.Context, endpointID string) (*restserver.GetEndpointResponse, error) {
+// GetEndpoint calls the EndpointHandlerAPI in CNS to retrieve the state of a given containerID
+func (c *Client) GetEndpoint(ctx context.Context, containerID string) (*restserver.GetEndpointResponse, error) {
 	// build the request
 	u := c.routes[cns.EndpointAPI]
-	uString := u.String() + endpointID
+	uString := u.String() + containerID
```
vipul-21, line 1030: Will this affect the existing endpoints ?

network/endpoint_windows.go
```diff
@@ -212,17 +215,24 @@ func (nw *network) configureHcnEndpoint(epInfo *EndpointInfo) (*hcn.HostComputeE
...
 			Minor: hcnSchemaVersionMinor,
 		},
-		MacAddress: epInfo.MacAddress.String(),
 	}
 
+	// macAddress type for InfraNIC is like "60:45:bd:12:45:65"
+	macAddress := epInfo.MacAddress.String()
+	if epInfo.NICType != cns.InfraNIC {
+		// convert the format of macAddress that HNS can accept, i.e, "60-45-bd-12-45-65" if NIC type is delegated NIC
+		macAddress = strings.Join(strings.Split(macAddress, ":"), "-")
```
vipul-21, line 232: Curious to know, is this something new ? When we read the HNS state or get endpoint info from HNS we will need it to covert it back to the `:` format

cni/network/network.go
```diff
@@ -434,7 +488,12 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		return fmt.Errorf("failed to create cns client with error: %w", err)
 	}
 
+	options := make(map[string]any)
+	ipamAddConfig := IPAMAddConfig{nwCfg: nwCfg, args: args, options: options}
+
 	if nwCfg.MultiTenancy {
```
QxBytes, line 494: only triggered in swiftv1 multitenancy ONLY add a comment

cni/network/network.go
```diff
@@ -779,47 +809,80 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
...
+	epInfo.Mode = nwInfo.Mode
+	epInfo.Subnets = nwInfo.Subnets
+	epInfo.PodSubnet = nwInfo.PodSubnet
+	epInfo.BridgeName = nwInfo.BridgeName
+	epInfo.NetNs = nwInfo.NetNs
+	epInfo.Options = nwInfo.Options
+	epInfo.DisableHairpinOnHostInterface = nwInfo.DisableHairpinOnHostInterface
+	epInfo.IPAMType = nwInfo.IPAMType
+	epInfo.IsIPv6Enabled = nwInfo.IsIPv6Enabled
+	epInfo.NetworkDNS = nwInfo.NetworkDNS
```
QxBytes, line 831: put everything in an epinfo (do not initialize twice)-- init once only-- rename addSubnetToNetworkInfo to addSubnetToEndpointInfo (etc. anything to NetworkInfo to EndpointInfo)

network/endpoint_windows.go
QxBytes: Paul mentioned this is required for delegated vmnics (which is a new nic type-- for infra nics things stay the same)

cns/client/client.go
QxBytes: I don't seem to see this change on my branch anymore-- this is just a variable name change so what actually changes is what we use as the "key"-- as long as it is consistent then what we set is what we get later

cni/network/invoker_azure.go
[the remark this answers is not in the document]
QxBytes: yes the key value varies (mac address, interface info + index, mac address etc.)

cni/network/invoker_azure.go
```diff
@@ -29,7 +29,7 @@ const (
 
 type AzureIPAMInvoker struct {
 	plugin delegatePlugin
-	nwInfo *network.NetworkInfo
+	nwInfo *network.EndpointInfo
```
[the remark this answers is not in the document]
QxBytes: When I merged the network info struct into the endpoint info struct, some fields had the same names, but I was not sure if they should be kept separate. For example, "DNS"-- previously we would have two separate fields: nwInfo.DNS and epInfo.DNS-- there may have been a difference between the DNS in nwinfo and epinfo. I kept the variable names as nwInfo or epInfo so that if there were another instance where I had accidentally merged two fields together when they should not have been, I would be able to tell if that field was accessed from a network info or endpoint info previously. Because of this, I want to keep those names as they are for now at least.

cni/network/network.go
QxBytes: added a log-- it seems like if we return an empty string, we return an error in create ep info

cni/network/multitenancy.go
QxBytes: added a getInterfaceInfoKey method to each of the 3 possible ways we populate ipam addresults/interface info:
- multitenancy
- cns invoker
- azure invoker
Having one unified function for all the different methods would be strange because, for example, in multitenancy, we input an index local to a loop. Once the map is populated in each of these 3 cases, I assume that we never access the map by key again (we only loop over its contents).

network/endpoint_windows.go
vipul-21: I see. When you say that it is delegated nics does that mean backend nic as well ? If it is only for delegated nic type, should the condition be for delegated nic and not on `all except infranic` ?

cni/network/invoker_azure.go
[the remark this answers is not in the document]
QxBytes: I want to keep the variable names as is because if we find out later that I accidentally merged something I should not have, it will be incredibly difficult to pin-point which endpointinfos were previously used in the context of networks, and which ones were used in the context of endpoints

cni/network/invoker_cns.go
```diff
@@ -164,29 +163,39 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
 				}
 			}
 
 			overlayMode := (invoker.ipamMode == util.V4Overlay) || (invoker.ipamMode == util.DualStackOverlay) || (invoker.ipamMode == util.Overlay)
-			if err := configureDefaultAddResult(&info, &addConfig, &addResult, overlayMode); err != nil {
+			if err := configureDefaultAddResult(&info, &addConfig, &addResult, overlayMode, key); err != nil {
 				return IPAMAddResult{}, err
 			}
+		default:
+			logger.Info("Unknown NIC type received from cns pod ip info", zap.String("nicType", string(info.nicType)))
```
vipul-21, line 198: Should this be a warning ?

cni/network/network.go
```diff
@@ -453,274 +527,221 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
-			logger.Info("Created network",
-				zap.String("networkId", networkID),
-				zap.String("subnet", ipamAddResult.hostSubnetPrefix.String()))
-			sendEvent(plugin, fmt.Sprintf("[cni-net] Created network %v with subnet %v.", networkID, ipamAddResult.hostSubnetPrefix.String()))
 		}
+	}()
+
+	epInfos := []*network.EndpointInfo{}
+	infraSeen := false
+	i := 0
```
paulyufan2, line 581: can we name it as endpointIndex?

cni/network/network.go
```diff
@@ -453,274 +527,221 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
 			policies:         policies,
-			endpointID:       endpointID,
 			k8sPodName:       k8sPodName,
 			k8sNamespace:     k8sNamespace,
 			enableInfraVnet:  enableInfraVnet,
 			enableSnatForDNS: enableSnatForDNS,
 			natInfo:          natInfo,
-		}
+			networkID:        networkID,
+
```
paulyufan2, line 601: can we remove these space?

cni/network/network.go
```diff
@@ -453,274 +527,221 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+			ifInfo:        &ifInfo,
+			ipamAddConfig: &ipamAddConfig,
+			ipv6Enabled:   ipamAddResult.ipv6Enabled,
 
-		var epInfo network.EndpointInfo
-		epInfo, err = plugin.createEndpointInternal(&createEndpointInternalOpt)
+			infraSeen: &infraSeen,
+			idx:       i,
+		}
+		var epInfo *network.EndpointInfo
```
paulyufan2, line 609: NIT: let's move this to line 376 as the global variable

cni/network/network.go
```diff
@@ -1008,83 +1062,88 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-			if network.IsNetworkNotFoundError(err) {
-				err = nil
-				return err
-			}
-
-			logger.Error("Failed to extract network name from network config", zap.Error(err))
-			err = plugin.Errorf("Failed to extract network name from network config. error: %v", err)
+	// Initialize values from network config.
+	if err != nil {
+		// If error is not found error, then we ignore it, to comply with CNI SPEC.
```
paulyufan2, line 1081: can you please explain this "if error is not found error...."

cni/network/network.go
```diff
@@ -1396,3 +1461,12 @@ func convertCniResultToInterfaceInfo(result *cniTypesCurr.Result) network.Interf
 
 	return interfaceInfo
 }
+
+// func findInfraNicInterfaceKey(ifInfos map[string]network.InterfaceInfo) network.InterfaceInfo {
```
paulyufan2, line 1465: do we need this function? If not, let's remove this comment?

cni/network/network_windows.go
```diff
@@ -358,15 +374,15 @@ func getLoopbackDSRPolicy(args PolicyArgs) ([]policy.Policy, error) {
 	return policies, nil
 }
 
-func getIPV6EndpointPolicy(nwInfo *network.NetworkInfo) (policy.Policy, error) {
+func getIPV6EndpointPolicy(subnetInfos []network.SubnetInfo) (policy.Policy, error) {
 	var eppolicy policy.Policy
 
-	if len(nwInfo.Subnets) < 2 {
+	if len(subnetInfos) < ipv6SubnetInfoCount {
```
paulyufan2, line 380: let's name ipv6SubnetInfoCount as dualStackCount

network/endpoint.go
```diff
@@ -53,21 +53,26 @@ type endpoint struct {
 	PODName                  string `json:",omitempty"`
 	PODNameSpace             string `json:",omitempty"`
 	InfraVnetAddressSpace    string `json:",omitempty"`
-	NetNs                    string `json:",omitempty"`
+	NetNs                    string `json:",omitempty"` // used in windows
 	// SecondaryInterfaces is a map of interface name to InterfaceInfo
 	SecondaryInterfaces map[string]*InterfaceInfo
+	// Store nic type since we no longer populate SecondaryInterfaces
+	NICType cns.NICType
```
paulyufan2, line 60: We still have secondaryInterfaces populated on line 58 and each interface object contains NICType field, why we store NICType and NICName here again?

network/endpoint.go
```diff
@@ -118,6 +138,8 @@ type InterfaceInfo struct {
 	DNS               DNSInfo
 	NICType           cns.NICType
 	SkipDefaultRoutes bool
+	HostSubnetPrefix  net.IPNet                        // Move this field from ipamAddResult
+	NCResponse        *cns.GetNetworkContainerResponse // TODO: create a new struct and keep fields that are being used
```
paulyufan2, line 142: my next PR will fix this -> remove unused fields from NCResponse, need to take time to understand this field

network/endpoint_windows.go
paulyufan2: Hi @vipul-21  this will be applied on delegatedNIC, IBNIC and Accelnet NIC together, I am not sure why they changed the macAddress format but I will follow up with HNS and containerd team.

network/endpoint_windows.go
```diff
@@ -381,6 +393,12 @@ func (nw *network) newEndpointImplHnsV2(cli apipaClient, epInfo *EndpointInfo) (
 		gateway = net.ParseIP(hnsResponse.Routes[0].NextHop)
 	}
 
+	nicName := epInfo.IfName
+	// infra nic nicname will look like eth0, but delegated/secondary nics will look like "vEthernet 2"
```
paulyufan2, line 397: NIT: infra NIC name on windows always look like eth0(or VEthernet 0); secondary interfaces name look like "vEthernet x"(x means 1-7, maximum up to 8 interfaces on one windows node)

network/endpoint_windows.go
```diff
@@ -399,6 +417,9 @@ func (nw *network) newEndpointImplHnsV2(cli apipaClient, epInfo *EndpointInfo) (
 		ContainerID:              epInfo.ContainerID,
 		PODName:                  epInfo.PODName,
 		PODNameSpace:             epInfo.PODNameSpace,
+		// SecondaryInterfaces:      make(map[string]*InterfaceInfo),
```
paulyufan2, line 420: we probably need this if we want to delete the hns network on stateful CNI, if we already convert the code to stateless, then let's remove this SecondaryInteraces field

network/manager.go
```diff
@@ -436,24 +431,28 @@ func (nm *networkManager) UpdateEndpointState(ep *endpoint) error {
...
-		if networkID == "" {
-			networkID = DefaultNetworkID
-		}
-		epInfo, err = epInfo.GetEndpointInfoByIPImpl(epInfo.IPAddresses, networkID)
-		if err != nil {
-			return nil, errors.Wrapf(err, "Get endpoint API returend with error")
+	epInfos := cnsEndpointInfotoCNIEpInfos(endpointResponse.EndpointInfo, containerID)
+
+	for i := 0; i < len(epInfos); i++ {
+		if epInfos[i].NICType == cns.InfraNIC {
```
behzad-mir, line 443: The NICType needs to be filled out manually in case we have incomplete state from Statefull CNI.

network/manager.go
```diff
@@ -39,6 +39,7 @@ const (
 	ContainerIDLength    = 8
 	EndpointIfIndex      = 0 // Azure CNI supports only one interface
 	DefaultNetworkID     = "azure"
+	dummyGUID            = "12345678-1234-1234-1234-123456789012"
```
paulyufan2, line 42: can we add comments here why we need this dummyGUID?

network/manager.go
```diff
@@ -475,39 +475,52 @@ func (nm *networkManager) DeleteEndpoint(networkID, endpointID string, epInfo *E
...
-		AllowInboundFromHostToNC: false,
+		AllowInboundFromHostToNC: false, // stateless currently does not support apipa
 		AllowInboundFromNCToHost: false,
 		EnableSnatOnHost:         false,
 		EnableMultitenancy:       false,
-		NetworkContainerID:       epInfo.Id,
+		NetworkContainerID:       epInfo.NetworkContainerID, // we don't use this as long as AllowInboundFromHostToNC and AllowInboundFromNCToHost are false
+		NetNs:                    dummyGUID,                 // to trigger hnsv2, windows
+		NICType:                  epInfo.NICType,
+		NICName:                  epInfo.IfName,
```
behzad-mir, line 508: ADD a comment in here that this filed will be used for deleting the delegated NIC in case of Swift V2

network/manager.go
```diff
@@ -692,46 +713,107 @@ func (nm *networkManager) GetEndpointID(containerID, ifName string) string {
...
-		// TODO: DelegatedNIC state will be added in a future PR
-		if ifName != InfraInterfaceName {
-			continue
-		}
+
+		// filling out the InfraNIC from the state
 		epInfo.IPAddresses = ipInfo.IPv4
 		epInfo.IPAddresses = append(epInfo.IPAddresses, ipInfo.IPv6...)
-		epInfo.IfName = ifName
+		epInfo.IfName = ifName // ifname (container veth peer) not used in linux (or even windows) deletion
```
behzad-mir, line 778: The comment needs to be changes since EpInfo . ifname is holding the ""eth0, et1,eth2" in the state unlike the ifname in Endpoint.

network/manager.go
```diff
@@ -692,46 +713,107 @@ func (nm *networkManager) GetEndpointID(containerID, ifName string) string {
...
+			IfIndex:            EndpointIfIndex, // Azure CNI supports only one interface
+			ContainerID:        endpointID,
+			PODName:            endpointInfo.PodName,
+			PODNameSpace:       endpointInfo.PodNamespace,
+			NetworkContainerID: endpointID,
+		}
+
+		// If we create an endpoint state with stateful cni and then swap to a stateless cni binary, ifname would not be populated
 		if ifName == "" {
 			ifName = InfraInterfaceName
```
behzad-mir, line 772: NICType should also be filled out in here if we need to use it to delete incomlete states.

cni/network/multitenancy.go
QxBytes: The three key formats are:
- Multitenancy: nic type + index in for loop
- CNS Invoker: Either mac address (if delegated nic) or infra nic (if infra nic)
- Azure Invoker: Always infra nic

cni/network/network.go
```diff
@@ -453,274 +527,221 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+			// Delete state in disk here
+			delErr := plugin.nm.DeleteState(epInfos)
+			if delErr != nil {
+				logger.Error("Could not delete state after detecting add failure", zap.Error(delErr))
+				return
+			}
+		}
+	}()
+	// CHECK: why does the cns client deal with apipa?-- "cns client" only used in windows to create apipa
+	err = plugin.nm.EndpointCreate(cnsclient, epInfos)
```
tamilmani1989, line 646: can we remove `cns.BackendNIC:` case since its noop anyway and can be matched with default

cni/network/network.go
```diff
@@ -730,96 +751,129 @@ func (plugin *NetPlugin) createEndpointInternal(opt *createEndpointInternalOpt)
...
 	}
-	epInfo.Policies = append(epInfo.Policies, epPolicies...)
 
-	// Populate addresses.
-	for _, ipconfig := range defaultInterfaceInfo.IPConfigs {
-		epInfo.IPAddresses = append(epInfo.IPAddresses, ipconfig.Address)
+	// for secondary (Populate addresses)
+	// initially only for infra nic but now applied to all nic types
+	addresses := make([]net.IPNet, len(opt.ifInfo.IPConfigs))
+	for i, ipconfig := range opt.ifInfo.IPConfigs {
```
tamilmani1989, line 760: should we initialize and assign endpointinfo here? and then call addsubnetoendpointinfo everything after this?

cni/network/network.go
```diff
@@ -999,6 +1052,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 			plugin.ipamInvoker = NewCNSInvoker(k8sPodName, k8sNamespace, cnsClient, util.ExecutionMode(nwCfg.ExecutionMode), util.IpamMode(nwCfg.IPAM.Mode))
 
 		default:
+			// nwInfo gets populated later in the function
 			plugin.ipamInvoker = NewAzureIpamInvoker(plugin, &nwInfo)
```
tamilmani1989, line 1056: how this work for dualnic scenario?

tamilmani1989: in dualnic , endpoints belong to 2 different networks

cni/network/network.go
```diff
@@ -1008,83 +1062,88 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-
-			logger.Error("Failed to extract network name from network config", zap.Error(err))
-			err = plugin.Errorf("Failed to extract network name from network config. error: %v", err)
+	// Initialize values from network config.
+	if err != nil {
+		// If error is not found error, then we ignore it, to comply with CNI SPEC.
+		if network.IsNetworkNotFoundError(err) {
+			err = nil
 			return err
 		}
```
tamilmani1989, line 1085: so networkid will be received from cns state right?

network/endpoint.go
```diff
@@ -53,21 +53,26 @@ type endpoint struct {
...
 	PODNameSpace             string `json:",omitempty"`
 	InfraVnetAddressSpace    string `json:",omitempty"`
-	NetNs                    string `json:",omitempty"`
+	NetNs                    string `json:",omitempty"` // used in windows
 	// SecondaryInterfaces is a map of interface name to InterfaceInfo
 	SecondaryInterfaces map[string]*InterfaceInfo
+	// Store nic type since we no longer populate SecondaryInterfaces
+	NICType cns.NICType
+	// Store the key used in stateless
+	NICName string
```
tamilmani1989, line 62: what is this nicname? why is it required and how is it different from Ifname?

network/endpoint.go
```diff
@@ -94,7 +99,22 @@ type EndpointInfo struct {
...
 	HNSNetworkID             string
-	HostIfName               string
+	HostIfName               string // unused in windows, and in linux
+	// Fields related to the network are below
+	MasterIfName                  string
+	AdapterName                   string
+	NetworkID                     string
+	Mode                          string
+	Subnets                       []SubnetInfo
+	PodSubnet                     SubnetInfo
```
tamilmani1989, line 109: Whats the diff between subnets and podsubnet?

network/endpoint.go
```diff
@@ -94,7 +99,22 @@ type EndpointInfo struct {
...
+	BridgeName                    string
+	NetNs                         string // used in windows
+	Options                       map[string]interface{}
+	DisableHairpinOnHostInterface bool
+	IPAMType                      string
```
tamilmani1989, line 114: why ipamtype is required?

network/endpoint_linux.go
```diff
@@ -94,18 +94,24 @@ func (nw *network) newEndpointImpl(
...
 		// Create a veth pair.
 		logger.Info("Generate veth name based on endpoint id")
-		hostIfName = fmt.Sprintf("%s%s", hostVEthInterfacePrefix, defaultEpInfo.Id[:7])
-		contIfName = fmt.Sprintf("%s%s-2", hostVEthInterfacePrefix, defaultEpInfo.Id[:7])
+		hostIfName = fmt.Sprintf("%s%s", hostVEthInterfacePrefix, defaultEpInfo.EndpointID[:7])
+		contIfName = fmt.Sprintf("%s%s-2", hostVEthInterfacePrefix, defaultEpInfo.EndpointID[:7])
+	}
+
+	nicName := epInfo.IfName
+	// infra nic nicname will look like eth0, but delegated/secondary nics will look like "eth1 eth2 eth3"
```
tamilmani1989, line 102:
```suggestion
	// infra nic nicname will look like eth0, but delegated/secondary nics will be project directly and its same as host delegated nic name"
```

network/endpoint_linux.go
```diff
@@ -94,18 +94,24 @@ func (nw *network) newEndpointImpl(
...
+	if epInfo.NICType != cns.InfraNIC {
+		nicName = epInfo.MasterIfName
 	}
```
tamilmani1989, line 105: cant we set epInfo.Ifname in cni/network when creating epInfo based on nic type? we can avoid creating another field NicName

network/endpoint.go
```diff
@@ -53,21 +53,26 @@ type endpoint struct {
...
+	EndpointID               string
 	ContainerID              string
 	NetNsPath                string
 	IfName                   string
 	SandboxKey               string
 	IfIndex                  int
 	MacAddress               net.HardwareAddr
-	DNS                      DNSInfo
+	EndpointDNS              DNSInfo
+	NetworkDNS               DNSInfo
```
tamilmani1989, line 75: for dns, we can skip network dns and just have endpoint dns.. its fine if we dont configure dns in network creation.. it should still work since we set it up during endpoint creation.. @behzad-mir already validated this

network/endpoint_windows.go
tamilmani1989: @paul hmm for IB nic, cni not going to invoke hns endpoint creation.. why its required for ib nic?

network/manager.go
```diff
@@ -39,6 +39,7 @@ const (
 	ContainerIDLength    = 8
 	EndpointIfIndex      = 0 // Azure CNI supports only one interface
 	DefaultNetworkID     = "azure"
+	dummyGUID            = "12345678-1234-1234-1234-123456789012"
 )
```
tamilmani1989, line 43: i thought we discussed to return hnsv2 if nm.Stateless and call hcn.V2ApiSupported();.. this is not required right?

network/manager.go
```diff
@@ -112,6 +114,10 @@ type NetworkManager interface {
 	GetNumberOfEndpoints(ifName string, networkID string) int
 	GetEndpointID(containerID, ifName string) string
 	IsStatelessCNIMode() bool
+	SaveState(eps []*endpoint) error
+	DeleteState(epInfos []*EndpointInfo) error
+	GetEndpointInfosFromContainerID(containerID string) []*EndpointInfo
+	GetEndpointState(networkID, containerID string) ([]*EndpointInfo, error)
```
tamilmani1989, line 120: why do we have 2 GET apis. where this is required?

network/manager.go
```diff
@@ -475,39 +475,52 @@ func (nm *networkManager) DeleteEndpoint(networkID, endpointID string, epInfo *E
...
+	if err != nil {
+		return err
+	}
+	if epInfo.NICType == cns.DelegatedVMNIC {
+		// we are currently assuming stateless is not running in linux
+		// CHECK: could this affect linux? (if it does, it could disconnect external interface, is that okay?)
+		// bad only when 1) stateless and 2) linux and 3) delegated vmnics exist
+		logger.Info("Deleting endpoint because delegated vmnic detected", zap.String("HNSNetworkID", nw.HnsId))
+		err = nm.deleteNetworkImpl(nw)
+		// no need to clean up state in stateless
```
tamilmani1989, line 521: this is windows specfic and windows specific delete implementation should handle this.. lets move this code to windows specific file

network/manager.go
```diff
@@ -517,9 +530,17 @@ func (nm *networkManager) GetEndpointInfo(networkID, endpointID string) (*Endpoi
...
 		logger.Info("calling cns getEndpoint API")
-		epInfo, err := nm.GetEndpointState(networkID, endpointID)
+		epInfos, err := nm.GetEndpointState(networkID, endpointID)
+		if err != nil {
+			return nil, err
+		}
+		for _, epInfo := range epInfos {
+			if epInfo.NICType == cns.InfraNIC {
+				return epInfo, nil
+			}
```
tamilmani1989, line 540: why return only for infranic?

tamilmani1989 (changes requested):
haven't reviewed completely.. reviewed part of it

cni/network/network.go
QxBytes: why? I don't want this variable to be visible/used prior to this line by accident

cni/network/network.go
QxBytes: this was already in the existing code but my understanding is that if the network is already not found, we shouldn't error out

network/endpoint.go
QxBytes: we shouldn't use secondary interfaces after this pr goes through (the pr should still support deletion etc. if secondary interfaces are populated though)

network/endpoint.go
QxBytes: okay I'll remove the comment since it seems like you will be taking care of this?

network/endpoint_windows.go
QxBytes: okay I'm removing this then since secondary interfaces is for the stateful case I believe

network/manager.go
QxBytes: This should be addressed in `cnsEndpointInfotoCNIEpInfo` if the`ifName == ""`, we populate the nictype with infra nic

network/manager.go
QxBytes: should be addressed by populating with infra nic if ifname is empty

network/manager.go
QxBytes: added a comment that describes the value of epInfo.IfName in this function-- please check if it addresses your concerns

network/manager.go
QxBytes: this is the comment we were discussing earlier regarding what changes we want to make if we need stateless cni to work with linux in the future, right?

network/manager.go
QxBytes: I'll add a comment.

I originally wanted to change UseHnsV2 to just return true if stateless is enabled. However, UseHnsV2 is called in both newNetwork and newEndpoint. In newNetwork, we have access to the network manager. In newEndpoint, we do not. Thus, I cannot add UseHnsV2 to network manager.

Two options remain:
1) Add dummy GUID during deletion. We want to always trigger UseHnsV2 to true in stateless deletion. UseHnsV2 is true when we have a guid in netns and the api is supported. So, we pass in a dummy guid to satisfy the first condition, and the second condition is always true in stateless.

2) Add netns to stateless. When we convert the saved cns state we can repopulate the netns in the epInfo field. However, I do not see us using the netns value anywhere (only whether or not it is a valid guid-- the actual guid value doesn't seem to matter).

network/manager.go
QxBytes: get endpoint info's method signature allows me to only return one endpoint info, and when we get the endpoint state (stateless) there can be many eps associated with a single container id. sidenote: when do we even use GetEndpointInfo-- it uses "Get" flow but when do we use that??

network/manager.go
QxBytes: do you mean just having nm.deleteNetworkImpl in DeleteEndpointState (no if conditional)? I wanted to do this but we don't want to delete any network in the linux case and the deleteNetworkImpl on the linux side seems to have some side effects (can you explain what this function does)? I would modify deleteNetworkImpl for linux to be a noop but it still seems to be used in manager.go/restore and cnm delete network

network/manager.go
QxBytes: use one if in stateless cni and other in stateful cni (during deletion)

network/manager.go
QxBytes: I originally wanted to change UseHnsV2 to just return true if stateless is enabled. However, UseHnsV2 is called in both newNetwork and newEndpoint. In newNetwork, we have access to the network manager. In newEndpoint, we do not. Thus, I cannot add UseHnsV2 to network manager.

Two options remain:
1) Add dummy GUID during deletion. We want to always trigger UseHnsV2 to true in stateless deletion. UseHnsV2 is true when we have a guid in netns and the api is supported. So, we pass in a dummy guid to satisfy the first condition, and the second condition is always true in stateless. (took this approach)

2) Add netns to stateless. When we convert the saved cns state we can repopulate the netns in the epInfo field. However, I do not see us using the netns value anywhere (only whether or not it is a valid guid-- the actual guid value doesn't seem to matter).

cni/network/network.go
QxBytes: network id won't be received from cns state, but hns network id will be. i believe this case (above) is only triggered if swiftv1 multitenancy and error from getting nwinfo

cni/network/network.go
QxBytes: assuming stateful, we call getEndpointInfosFromContainerID, which loops over all networks and finds the endpointInfos that match the container id. Then we call Delete Endpoint for each one.

cni/network/multitenancy_mock.go
```diff
@@ -154,15 +155,24 @@ func (m *MockMultitenancy) GetAllNetworkContainers(
...
-		ipamResults[i].defaultInterfaceInfo.Routes = routes
-		ipamResults[i].defaultInterfaceInfo.NICType = cns.InfraNIC
+		// one ncResponse gets you one interface info in the returned IPAMAddResult
+		ifInfo := network.InterfaceInfo{
+			NCResponse:       &cnsResponses[i],
+			HostSubnetPrefix: ipNets[i],
+		}
+
+		ipconfig, routes := convertToIPConfigAndRouteInfo(ifInfo.NCResponse)
+		ifInfo.IPConfigs = []*network.IPConfig{ipconfig}
```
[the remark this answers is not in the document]
QxBytes: on assigning ifInfo.IPConfigs?

cni/network/invoker_cns.go
```diff
@@ -164,20 +163,27 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
+			// Add secondary interface info from podIPInfo to ipamAddResult
+			info.hostSubnet = response.PodIPInfo[i].HostPrimaryIPInfo.Subnet
+			info.hostPrimaryIP = response.PodIPInfo[i].HostPrimaryIPInfo.PrimaryIP
+			info.hostGateway = response.PodIPInfo[i].HostPrimaryIPInfo.Gateway
+
 			if err := configureSecondaryAddResult(&info, &addResult, &response.PodIPInfo[i].PodIPConfig); err != nil {
 				return IPAMAddResult{}, err
 			}
-		default:
+		case cns.InfraNIC:
```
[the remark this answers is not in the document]
QxBytes: my code is expecting that once ipamAddResult and interfaceInfo inside it is populated and returned, there should always be a nic type-- if it is empty it should be populated with something, likely infra nic if nothing matches-- it will break if the nic type is left empty. To address your comment, if the nic type is empty we should perform the same actions as if it were "infra" (and populate the nic type as infra)-- will also check invoker_azure and multitenancy.go to check cases where nic type is empty

cni/network/invoker_azure_test.go
```diff
@@ -90,6 +94,15 @@ func getCIDRNotationForAddress(ipaddresswithcidr string) *net.IPNet {
...
 }
 
+// returns an ipnet, which contains the *masked* ip (zeroed out based on CIDR) and the mask itself
+func parseCIDR(ipaddresswithcidr string) *net.IPNet {
+	_, ipnet, err := net.ParseCIDR(ipaddresswithcidr)
+	if err != nil {
+		panic(fmt.Sprintf("failed to parse cidr with err: %v", err))
+	}
+	return ipnet
+}
```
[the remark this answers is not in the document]
QxBytes: used in uts; getCIDRNotationForAddress doesn't get the expected output in the right format so I created this

cni/network/invoker_cns.go
```diff
@@ -384,15 +392,27 @@ func configureDefaultAddResult(info *IPResultInfo, addConfig *IPAMAddConfig, add
 		}
 	}
 
+	if err != nil {
+		logger.Error("Error finding InfraNIC interface",
```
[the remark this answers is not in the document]
QxBytes: deleting this error checking block since errors are validated earlier

cni/network/invoker_cns.go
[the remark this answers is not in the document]
QxBytes: confirmed in multitenancy.go and invoker_azure.go that we always set nic type to infra when adding to the interface info map

cni/network/network.go
QxBytes: checking if the assignment can be done after

network/manager.go
QxBytes: adding comment with TODO to revisit later

network/endpoint.go
QxBytes: PodSubnet does not seem to be used in linux or windows-- deleting from endpoint info struct

network/manager.go
paulyufan2: thanks Alexander, as discussed with TM here, let's add a TODO here to fix this in next PR

network/endpoint_windows.go
paulyufan2: @tamilmani1989 sorry you are right, IB NIC does not interact with HNS endpoint creation. Accelnet NIC does.

cni/network/multitenancy_mock.go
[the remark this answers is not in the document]
QxBytes: 169 try using append

cni/network/network.go
QxBytes: fix linter issue

network/manager.go
QxBytes: address later when we create an endpoint delete function to match EndpointCreate

cni/network/network.go
```diff
@@ -1008,83 +1046,89 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+			// An error will not be returned if the endpoint is not found
 			// return a retriable error so the container runtime will retry this DEL later
 			// the implementation of this function returns nil if the endpoint doens't exist, so
 			// we don't have to check that here
 			return plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
 		}
+	}
+	logger.Info("Deleting the endpoints from the ipam")
+	// delete endpoint state in cns and in statefile
+	for _, epInfo := range epInfos {
```
tamilmani1989, line 1123: why we add a separate loop instead call ipam delete in same loop?

network/endpoint_linux.go
```diff
@@ -57,19 +57,19 @@ func (nw *network) newEndpointImpl(
...
+	epInfo *EndpointInfo,
 ) (*endpoint, error) {
 	var (
 		err           error
 		hostIfName    string
 		contIfName    string
 		localIP       string
 		vlanid        = 0
-		defaultEpInfo = epInfo[0]
+		defaultEpInfo = epInfo
```
tamilmani1989, line 68: why this assignment required? can we just keep epInfo?

network/endpoint_windows.go
```diff
@@ -212,21 +216,27 @@ func (nw *network) configureHcnEndpoint(epInfo *EndpointInfo) (*hcn.HostComputeE
...
-			hcnEndpoint.Policies = append(hcnEndpoint.Policies, epPolicy)
-		}
+	// macAddress type for InfraNIC is like "60:45:bd:12:45:65"
+	// if NICType is delegatedVMNIC, convert the macaddress format
+	macAddress := epInfo.MacAddress.String()
+	if epInfo.NICType == cns.DelegatedVMNIC {
+		// convert the format of macAddress that HNS can accept, i.e, "60-45-bd-12-45-65" if NIC type is delegated NIC
+		macAddress = strings.Join(strings.Split(macAddress, ":"), "-")
+	}
+	hcnEndpoint.MacAddress = macAddress
```
tamilmani1989, line 236: is macaddress needed for infra? I thought its required only for delegated nic transparent network

network/manager.go
```diff
@@ -517,9 +533,17 @@ func (nm *networkManager) GetEndpointInfo(networkID, endpointID string) (*Endpoi
 
 	if nm.IsStatelessCNIMode() {
 		logger.Info("calling cns getEndpoint API")
-		epInfo, err := nm.GetEndpointState(networkID, endpointID)
+		epInfos, err := nm.GetEndpointState(networkID, endpointID)
+		if err != nil {
+			return nil, err
+		}
+		for _, epInfo := range epInfos {
```
tamilmani1989, line 540: why return only for infra nic?

network/manager.go
```diff
@@ -692,46 +716,112 @@ func (nm *networkManager) GetEndpointID(containerID, ifName string) string {
...
+		err := nm.UpdateEndpointState(eps)
+		if err != nil {
+			return err
+		}
+	}
+
+	// we either use stateless cni and save via update endpoint state, or use the state file
+	if nm.IsStatelessCNIMode() {
+		return nil
+	}
```
tamilmani1989, line 736:
```suggestion
	if nm.IsStatelessCNIMode() {
		err := nm.UpdateEndpointState(eps)
        return err
    }
```

network/network.go
```diff
@@ -75,6 +75,7 @@ type NetworkInfo struct {
 	IPAMType                      string
 	ServiceCidrs                  string
 	IsIPv6Enabled                 bool
+	NICType                       string
```
tamilmani1989, line 78: are we still using NetworkInfo? i thought this structure is not used

QxBytes: removed NICType from NetworkInfo

network/manager.go
QxBytes: I believe we discussed this before? https://github.com/Azure/azure-container-networking/pull/2688#discussion_r1609180201 We can only return 1 endpoint info from this method. Is there a different intended behavior/a case where we would want to return an endpoint info of type delegated vmnic?

cni/network/network.go
QxBytes: I believe ipam invoker delete deletes the state in stateless mode so if we need to perform, for example, 3 deletes and fail during the second DeleteEndpoint call, we may end up in an inconsistent state. I believe you requested that all endpoints be deleted first before modifying the state (whether it be stored on the cns like in stateless or in the cni statefile).

network/endpoint_windows.go
paulyufan2: macAddress is not needed for infra. It only requires for delegated NIC now

tamilmani1989 (approved):

paulyufan2 (approved):

tamilmani1989 (approved):
🚢

PR #2693, opened 2024-04-18
feat: Adding backend interfaces support to CNS

shreyashastantram:
**Reason for Change**:

This feature adds backend interfaces support to CNS

**Issue Fixed**:

**Requirements**:
CNS should be able to return back with PnPId to CNI with ipconfigs response.
CNS should be able to determine the PnPID based on the macaddress.

- [ *] includes documentation
 https://microsoft.sharepoint.com/:w:/t/Aznet/EQI_uDcvkwRDkjBX4X2yZo4BTkGR5g55ttvupXYG_TfC6Q?e=o8iR2s
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -481,7 +481,7 @@ type PodIpInfo struct {
 	SkipDefaultRoutes bool
 	// Routes to configure on interface
 	Routes []Route
-	// PnpId is set for backend interfaces, Pnp Id identifies VF
+	// PnpId is set for backend interfaces, Pnp Id identifies VF. Plug and play id(pnp) is also called as PCI ID
 	PnPID string
 }
 
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -501,13 +501,14 @@ type IPConfigRequest struct {
 
 // Same as IPConfigRequest except that DesiredIPAddresses is passed in as a slice
 type IPConfigsRequest struct {
-	DesiredIPAddresses         []string        `json:"desiredIPAddresses"`
-	PodInterfaceID             string          `json:"podInterfaceID"`
-	InfraContainerID           string          `json:"infraContainerID"`
-	OrchestratorContext        json.RawMessage `json:"orchestratorContext"`
-	Ifname                     string          `json:"ifname"`                      // Used by delegated IPAM
-	SecondaryInterfacesExist   bool            `json:"secondaryInterfacesExist"`    // will be set by SWIFT v2 validator func
-	BackendInterfaceMacAddress string          `json:"BacknendInterfaceMacAddress"` // will be set by SWIFT v2 validator func
+	DesiredIPAddresses           []string        `json:"desiredIPAddresses"`
+	PodInterfaceID               string          `json:"podInterfaceID"`
+	InfraContainerID             string          `json:"infraContainerID"`
+	OrchestratorContext          json.RawMessage `json:"orchestratorContext"`
+	Ifname                       string          `json:"ifname"`                   // Used by delegated IPAM
+	SecondaryInterfacesExist     bool            `json:"secondaryInterfacesExist"` // will be set by SWIFT v2 validator func
+	BackendInterfaceExist        bool            `json:"BackendInterfaceExist"`    // will be set by SWIFT v2 validator func
+	BackendInterfaceMacAddresses []string        `json:"BacknendInterfaceMacAddress"`
 }
 
 // IPConfigResponse is used in CNS IPAM mode as a response to CNI ADD
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -481,6 +481,8 @@ type PodIpInfo struct {
 	SkipDefaultRoutes bool
 	// Routes to configure on interface
 	Routes []Route
+	// PnpId is set for backend interfaces, Pnp Id identifies VF
+	PnPID string
 }
 
 type HostIPInfo struct {
@@ -499,12 +501,13 @@ type IPConfigRequest struct {
 
 // Same as IPConfigRequest except that DesiredIPAddresses is passed in as a slice
 type IPConfigsRequest struct {
-	DesiredIPAddresses       []string        `json:"desiredIPAddresses"`
-	PodInterfaceID           string          `json:"podInterfaceID"`
-	InfraContainerID         string          `json:"infraContainerID"`
-	OrchestratorContext      json.RawMessage `json:"orchestratorContext"`
-	Ifname                   string          `json:"ifname"`                   // Used by delegated IPAM
-	SecondaryInterfacesExist bool            `json:"secondaryInterfacesExist"` // will be set by SWIFT v2 validator func
+	DesiredIPAddresses         []string        `json:"desiredIPAddresses"`
+	PodInterfaceID             string          `json:"podInterfaceID"`
+	InfraContainerID           string          `json:"infraContainerID"`
+	OrchestratorContext        json.RawMessage `json:"orchestratorContext"`
+	Ifname                     string          `json:"ifname"`                      // Used by delegated IPAM
+	SecondaryInterfacesExist   bool            `json:"secondaryInterfacesExist"`    // will be set by SWIFT v2 validator func
+	BackendInterfaceMacAddress string          `json:"BacknendInterfaceMacAddress"` // will be set by SWIFT v2 validator func
 }
 
 // IPConfigResponse is used in CNS IPAM mode as a response to CNI ADD
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -58,7 +58,11 @@ type CNSConfig struct {
 	UseMTLS                     bool
 	WatchPods                   bool `json:"-"`
 	WireserverIP                string
+<<<<<<< HEAD
 	GRPCSettings                GRPCSettings
+=======
+	BackendNICEnabled           bool
+>>>>>>> 827d2ff3 (Adding backend interfaces support for CNS)
 }
 
 type TelemetrySettings struct {
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -58,11 +58,8 @@ type CNSConfig struct {
 	UseMTLS                     bool
 	WatchPods                   bool `json:"-"`
 	WireserverIP                string
-<<<<<<< HEAD
 	GRPCSettings                GRPCSettings
-=======
 	BackendNICEnabled           bool
->>>>>>> 827d2ff3 (Adding backend interfaces support for CNS)
 }
 
 type TelemetrySettings struct {
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -59,7 +59,6 @@ type CNSConfig struct {
 	WatchPods                   bool `json:"-"`
 	WireserverIP                string
 	GRPCSettings                GRPCSettings
-	BackendNICEnabled           bool
 }
 
 type TelemetrySettings struct {
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -116,7 +116,7 @@ func (k *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req
 
 	// check the pod labels for Swift V2, set the request's SecondaryInterfaceSet flag to true and check if its MTPNC CRD is ready
 	if _, ok := pod.Labels[configuration.LabelPodSwiftV2]; ok {
-		req.SecondaryInterfacesExist = true
+
 		// Check if the MTPNC CRD exists for the pod, if not, return error
 		mtpnc := v1alpha1.MultitenantPodNetworkConfig{}
 		mtpncNamespacedName := k8stypes.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
@@ -126,6 +126,10 @@ func (k *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req
 		// Check if the MTPNC CRD is ready. If one of the fields is empty, return error
 		if !mtpnc.IsReady() {
 			return nil, types.UnexpectedError, errMTPNCNotReady.Error()
+		}
+		// If primary Ip is set in status field, it indicates the presence of secondary interfaces
+		if mtpnc.Status.PrimaryIP != "" {
+			req.SecondaryInterfacesExist = true
 		}
 		interfaceInfos := mtpnc.Status.InterfaceInfos
 		for index, interfaceInfo := range interfaceInfos {
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -40,7 +40,12 @@ var _ cns.IPConfigsHandlerMiddleware = (*K8sSWIFTv2Middleware)(nil)
 // and release IP configs handlers.
 func (k *K8sSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(defaultHandler, failureHandler cns.IPConfigsHandlerFunc) cns.IPConfigsHandlerFunc {
 	return func(ctx context.Context, req cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
+<<<<<<< HEAD
 		podInfo, respCode, message := k.validateIPConfigsRequest(ctx, &req)
+=======
+		podInfo, respCode, message := m.validateIPConfigsRequest(ctx, &req)
+		// Add a function to fetch the macaddresses
+>>>>>>> 496b4e3f (Handling multiple interface information)
 
 		if respCode != types.Success {
 			return &cns.IPConfigsResponse{
@@ -128,13 +133,18 @@ func (k *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req
 			return nil, types.UnexpectedError, errMTPNCNotReady.Error()
 		}
 		interfaceInfos := mtpnc.Status.InterfaceInfos
-		for _, interfaceInfo := range interfaceInfos {
+		for index, interfaceInfo := range interfaceInfos {
 			if interfaceInfo.NICType == string(cns.BackendNIC) {
-				req.BackendInterfaceMacAddress = interfaceInfo.MacAddress
+				if interfaceInfo.MacAddress == "" || interfaceInfo.NCID == "" || mtpnc.Status.NCID == "" {
+					return nil, types.UnexpectedError, errMTPNCNotReady.Error()
+				}
+				req.BackendInterfaceExist = true
+				req.BackendInterfaceMacAddresses[index] = interfaceInfo.MacAddress
 			}
 		}
 	}
 	logger.Printf("[SWIFTv2Middleware] pod %s has secondary interface : %v", podInfo.Name(), req.SecondaryInterfacesExist)
+	logger.Printf("[SWIFTv2Middleware] pod %s has backend interface : %v", podInfo.Name(), req.BackendInterfaceExist)
 	// retrieve podinfo from orchestrator context
 	return podInfo, types.Success, ""
 }
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -137,7 +137,7 @@ func (k *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req
 		interfaceInfos := mtpnc.Status.InterfaceInfos
 		for _, interfaceInfo := range interfaceInfos {
 			if interfaceInfo.DeviceType == v1alpha1.DeviceTypeInfiniBandNIC {
-				if interfaceInfo.MacAddress == "" || interfaceInfo.NCID == "" || mtpnc.Status.NCID == "" {
+				if interfaceInfo.MacAddress == "" || interfaceInfo.NCID == "" {
 					return nil, types.UnexpectedError, errMTPNCNotReady.Error()
 				}
 				req.BackendInterfaceExist = true
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -134,7 +134,7 @@ func (k *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req
 		}
 		interfaceInfos := mtpnc.Status.InterfaceInfos
 		for index, interfaceInfo := range interfaceInfos {
-			if interfaceInfo.NICType == string(cns.BackendNIC) {
+			if interfaceInfo.DeviceType == v1alpha1.DeviceTypeInfiniBandNIC {
 				if interfaceInfo.MacAddress == "" || interfaceInfo.NCID == "" || mtpnc.Status.NCID == "" {
 					return nil, types.UnexpectedError, errMTPNCNotReady.Error()
 				}
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -127,6 +127,12 @@ func (k *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req
 		if !mtpnc.IsReady() {
 			return nil, types.UnexpectedError, errMTPNCNotReady.Error()
 		}
+		interfaceInfos := mtpnc.Status.InterfaceInfos
+		for _, interfaceInfo := range interfaceInfos {
+			if interfaceInfo.NICType == string(cns.BackendNIC) {
+				req.BackendInterfaceMacAddress = interfaceInfo.MacAddress
+			}
+		}
 	}
 	logger.Printf("[SWIFTv2Middleware] pod %s has secondary interface : %v", podInfo.Name(), req.SecondaryInterfacesExist)
 	// retrieve podinfo from orchestrator context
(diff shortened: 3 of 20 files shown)


cns/restserver/ipam.go
```diff
@@ -47,9 +47,27 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
 		}, errors.New("failed to validate ip config request")
 	}
 
+	var podBackendInfo cns.PodIpInfo
+	if ipconfigsRequest.BackendInterfaceMacAddress != "" {
+		PnPID, err := service.getPNPIDFromMacAddress(ipconfigsRequest.BackendInterfaceMacAddress)
+		if err != nil {
+			return &cns.IPConfigsResponse{
+				Response: cns.Response{
+					ReturnCode: types.FailedToAllocateIPConfig,
```
prmathur-microsoft, line 56: Is this the right error message to return?

cns/restserver/restserver.go
```diff
@@ -72,6 +72,7 @@ type HTTPRestService struct {
 	cniConflistGenerator       CNIConflistGenerator
 	generateCNIConflistOnce    sync.Once
 	IPConfigsHandlerMiddleware cns.IPConfigsHandlerMiddleware
+	PnpIDByMacAddress          map[string]string
```
prmathur-microsoft, line 75: When will this get deleted?

prmathur-microsoft, line 75: What happens to this data on reboot?

cns/service/main.go
```diff
@@ -872,6 +872,11 @@ func main() {
 		}
 	}
 
+	if cnsconfig.BackendNICEnabled {
```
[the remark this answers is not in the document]
prmathur-microsoft: I think we should enable this by default as there is no way of know when to and when not to. And if MTPNC sends this data we should process it.

cns/restserver/ipam.go
shreyashastantram: I will add a new error code.

cns/restserver/restserver.go
shreyashastantram: it's an in memory cache, it gets deleted when cns terminates.

shreyashastantram: repopulate

cns/middlewares/k8sSwiftV2.go
```diff
@@ -128,6 +128,12 @@ func (m *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req
 		if mtpnc.Status.PrimaryIP == "" || mtpnc.Status.MacAddress == "" || mtpnc.Status.NCID == "" || mtpnc.Status.GatewayIP == "" {
 			return nil, types.UnexpectedError, errMTPNCNotReady.Error()
 		}
+		interfaceInfos := mtpnc.Status.InterfaceInfos
+		for _, interfaceInfo := range interfaceInfos {
+			if interfaceInfo.NICType == string(cns.BackendNIC) {
+				req.BackendInterfaceMacAddress = interfaceInfo.MacAddress
```
estebancams, line 134: Since at the moment we are only having one IB at most, this would work, but I would assert that we just find at most one of its kind or fail since later this could be and issue hard to itentify

cns/restserver/ipam.go
estebancams: Please also update the message below as well once you change the error code

cns/restserver/ipam.go
```diff
@@ -83,6 +101,7 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
 		}
 	}
 
+	podIPInfo = append(podIPInfo, podBackendInfo)
```
estebancams, line 104: should this go inside the if in line 51? Do we want to have an empty podBackendInfo in the slice in case no backendinterface info is passed in the request?

cns/restserver/util.go
```diff
@@ -40,6 +46,49 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
...
+	ctx, cancel := context.WithTimeout(context.Background(), Timeout)
+	defer cancel() // The cancel should be deferred so resources are cleaned up
+	cmd := exec.CommandContext(ctx, "powershell", GetMacAddressVFPPnpIDMapping)
+	output, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+	result := make(map[string]string)
+	if len(output) == 0 {
+		return nil, errors.New("Network adapter not found")
```
estebancams, line 79: Why do we error out here? Can't we have an empty mapping if IB is not available in the node?

cns/restserver/util.go
```diff
@@ -40,6 +46,49 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
...
+	}
+	lines := strings.Split(strings.TrimSpace((string(output))), "\n")
+	logger.Printf("received input:%s", lines)
+	for _, line := range lines {
+		parts := strings.Split(line, " ")
+		key := strings.ToUpper(strings.ReplaceAll(parts[0], "-", ":"))
```
estebancams, line 85: Any particular reason we make this reformatting?

cns/restserver/util.go
```diff
@@ -40,6 +46,49 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
...
+	output, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+	result := make(map[string]string)
+	if len(output) == 0 {
+		return nil, errors.New("Network adapter not found")
+	}
+	lines := strings.Split(strings.TrimSpace((string(output))), "\n")
+	logger.Printf("received input:%s", lines)
```
estebancams, line 82: Can we be more verbose here? And also add [Azure CNS] prefix to be consistent

cns/restserver/util.go
```diff
@@ -40,6 +46,49 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
...
+	if err != nil {
+		return err
+	}
+	service.PnpIDByMacAddress = VfMacAddressMapping
+	return nil
+}
+
+func (service *HTTPRestService) getPNPIDFromMacAddress(macAddress string) (string, error) {
+	if _, ok := service.PnpIDByMacAddress[macAddress]; !ok {
+		if err := service.SetPnpIDMacaddressMapping(); err != nil {
```
estebancams, line 62: What is the goal of refetching mappings here? and why error out if that fails?

cns/service/main.go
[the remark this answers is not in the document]
shreyashastantram: If we perform this during add call, it can lead to delay in ipconfigsresponse. Powershell commands sometimes takes longer than expected. Doing this over the boot time reduces this delay.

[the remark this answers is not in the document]
shreyashastantram: I have removed the need for config, I have moved this function to windows platform, it will be no op for linux. I will bucket this under cns swiftv2 mode hence it will get executed only for swift v2 and windows platform.

cns/NetworkContainerContract.go
```diff
@@ -474,6 +474,8 @@ type PodIpInfo struct {
 	SkipDefaultRoutes bool
 	// Routes to configure on interface
 	Routes []Route
+	// PnpId is set for backend interfaces, Pnp Id identifies VF
+	PnPID string
```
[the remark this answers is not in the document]
shreyashastantram: PnPID is being used by OS team, we want to be aligned on what os takes makes us of.

cns/NetworkContainerContract.go
```diff
@@ -492,12 +494,13 @@ type IPConfigRequest struct {
...
-	OrchestratorContext      json.RawMessage `json:"orchestratorContext"`
-	Ifname                   string          `json:"ifname"`                   // Used by delegated IPAM
-	SecondaryInterfacesExist bool            `json:"secondaryInterfacesExist"` // will be set by SWIFT v2 validator func
+	DesiredIPAddresses         []string        `json:"desiredIPAddresses"`
+	PodInterfaceID             string          `json:"podInterfaceID"`
+	InfraContainerID           string          `json:"infraContainerID"`
+	OrchestratorContext        json.RawMessage `json:"orchestratorContext"`
+	Ifname                     string          `json:"ifname"`                      // Used by delegated IPAM
+	SecondaryInterfacesExist   bool            `json:"secondaryInterfacesExist"`    // will be set by SWIFT v2 validator func
+	BackendInterfaceMacAddress string          `json:"BacknendInterfaceMacAddress"` // will be set by SWIFT v2 validator func
```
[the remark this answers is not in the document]
shreyashastantram: Yes, we will always have a macaddress available.

cns/restserver/ipam.go
shreyashastantram: fixed

cns/restserver/util.go
shreyashastantram: Addressed

cns/restserver/util.go
shreyashastantram: There can be case where ib's are not programmed during the start up. This is to address scenarios where it is not present at the boot up and available later.

cns/restserver/util.go
shreyashastantram: We have seen macaddress formatted with : instead of -

timraymond: This is why @rbtr and I maintain that MAC addresses should be parsed at the application border into an internal type that is purpose-built for dealing with it. See discussion here: https://github.com/Azure/azure-container-networking/pull/2634#discussion_r1530977847

rbtr: +1, stop kicking the can down the road and let's fix these stringly typed variables instead of adding more special handling for them

cns/restserver/util.go
```diff
@@ -40,6 +46,49 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
...
+	if err != nil {
+		return nil, err
+	}
+	result := make(map[string]string)
+	if len(output) == 0 {
+		return nil, errors.New("Network adapter not found")
+	}
+	lines := strings.Split(strings.TrimSpace((string(output))), "\n")
+	logger.Printf("received input:%s", lines)
+	for _, line := range lines {
```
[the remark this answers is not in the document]
shreyashastantram: Added

cns/restserver/util.go
shreyashastantram: Looked into it. I am using the internal type for parsing, but it is not supported in maps, we will need to use string for maps. If we need to change this in ipconfigs response for CNI, it requires changes in CNI contracts. are you suggesting to go for that?

cns/restserver/util.go
shreyashastantram: It was a debug line, I have removed it.

cns/configuration/configuration.go
```diff
@@ -57,6 +57,7 @@ type CNSConfig struct {
 	UseHTTPS                    bool
 	WatchPods                   bool `json:"-"`
 	WireserverIP                string
+	BackendNICEnabled           bool
```
jaer-tsun, line 60: where does this get used?

cns/middlewares/k8sSwiftV2.go
jaer-tsun, line 134: looks like this should be done outside of validate request func

platform/os_windows.go
```diff
@@ -343,3 +373,33 @@ func ReplaceFile(source, destination string) error {
...
 }
+
+/*
+	Sample output:  Get-NetAdapter | Select-Object MacAddress, PnpDeviceID| Format-Table -HideTableHeaders
+
+6C-A1-00-50-E4-2D PCI\VEN_8086&DEV_2723&SUBSYS_00808086&REV_1A\4&328243d9&0&00E0
+80-6D-97-1E-CF-4E USB\VID_17EF&PID_A359\3010019E3
+*/
+func FetchMacAddressPnpIDMapping(execClient ExecClient) (map[string]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
jaer-tsun, line 384: was the Timeout const in cns util supposed to be in this package?

cns/restserver/util.go
timraymond: You can write a type to mediate access to the map. This is just a thin wrapper on the map: https://go.dev/play/p/ix2n3cHLJM1

cns/restserver/ipam.go
nddq: we are technically failing to allocate IP Config due to `getPNPIDFromMacAddress` failed 🙂, so i think we can keep this and just change the error message below

cns/restserver/restserver.go
nddq: you might need to check if `saveState` has been called whenever we update this map, so it can be reconcile after CNS restarted.

cns/restserver/util.go
```diff
@@ -40,6 +44,32 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
 	service.state.Networks[networkName] = networkInfo
 }
 
+func (service *HTTPRestService) SetPnpIDMacaddressMapping() error {
+	service.Lock()
+	defer service.Unlock()
+	p := platform.NewExecClient(nil)
+	VfMacAddressMapping, err := platform.FetchMacAddressPnpIDMapping(p)
+	if err != nil {
+		return err
```
nddq, line 53:
```suggestion
		return errors.Wrap(err, "failed to fetch MACAddressPnpIDMapping")
```

platform/os_linux.go
```diff
@@ -136,6 +136,10 @@ func (p *execClient) ExecutePowershellCommand(_ string) (string, error) {
 	return "", nil
 }
 
+func (p *execClient) ExecutePowershellCommandWithContext(_ string, ctx context.Context) (string, error) {
```
nddq, line 139:
```suggestion
func (p *execClient) ExecutePowershellCommandWithContext(_ string, _ context.Context) (string, error) {
```

platform/os_linux.go
```diff
@@ -200,3 +204,7 @@ func HasMellanoxAdapter() bool {
 
 // Not needed for Linux
 func MonitorAndSetMellanoxRegKeyPriorityVLANTag(_ context.Context, _ int) {}
+
+func FetchMacAddressPnpIDMapping(execClient ExecClient) (map[string]string, error) {
```
nddq, line 208:
```suggestion
func FetchMacAddressPnpIDMapping(_ ExecClient) (map[string]string, error) {
```

platform/os_windows.go
```diff
@@ -343,3 +373,33 @@ func ReplaceFile(source, destination string) error {
...
+	}
+	result := make(map[string]string)
+	if len(output) != 0 {
+		// Split the output based on new line characters
+		lines := strings.Split(output, "\n")
+		for _, line := range lines {
+			// Split based on " " to fetch the macaddress and pci id
+			parts := strings.Split(line, " ")
+			// Changing the format of macaddress from xx-xx-xx-xx to xx:xx:xx:xx
+			formattedMacaddress := net.net.ParseMAC(parts[0])
```
nddq, line 398: `net.net.ParseMAC` doesn't look right, is it?

nddq (changes requested):

cns/middlewares/k8sSwiftV2.go
shreyashastantram: CNS read MTPNC value within that function, it's being used today to set SecondaryInterfacesExist flag, similar to that, instead of setting boolean value, I am setting macaddress.

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -61,6 +61,21 @@ type MultitenantPodNetworkConfigStatus struct {
 	MacAddress string `json:"macAddress,omitempty"`
 	// Gateway IP
 	GatewayIP string `json:"gatewayIP,omitempty"`
+	// InterfaceDetails
+	InterfaceInfos []InterfaceInfo `json:"interfaceInfo,omitempty"`
+}
+
+type InterfaceInfo struct {
```
timraymond, line 68: (non-blocking naming suggestion) This might be better named `NetworkInterface`. `Info` is superfluous in the name, since all structs carry information. `Interface` alone might be confusing because of the Go keyword `interface`.

platform/os_linux.go
shreyashastantram: Addressed

platform/os_linux.go
shreyashastantram: Added

cns/restserver/util.go
shreyashastantram: Added

cns/NetworkContainerContract.go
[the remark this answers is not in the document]
timraymond: Even if, in the code we want to align on internal terminology, it would be nice to have an "a.k.a." comment so it's possible to search for concepts in publicly-available documentation. From the outside, I will have no idea what a PnP is otherwise.

cns/restserver/util.go
```diff
@@ -40,6 +40,32 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
 	service.state.Networks[networkName] = networkInfo
 }
 
+func (service *HTTPRestService) SetPnpIDMacaddressMapping() error {
+	service.Lock()
+	defer service.Unlock()
```
timraymond, line 45: I would strongly recommend not using this. Embedding a `sync.RWMutex` is almost never a good idea unless you're implementing another kind of synchronization primitive (`sync.Cond` is a good example). Consequently, its embedding in `*HTTPRestService` is probably a mistake. Plus, it's not clear exactly what else is using this, so this runs the risk of deadlocking (maybe not now, but perhaps when future changes are made).

Instead, if you want a synchronized cache, I'd recommend creating a type to do exactly that with its own mutex to manage *only* its critical sections.

cns/restserver/util.go
shreyashastantram: I am unable to find a suitable place to add this map, It needs to be present in platform package but then we would not be able to use this in rest server package. Also, We are adding complexity for doing the string conversion, we are storing it as a string in map even using this mediate access. CNS also responds to CNI as string, I do not see any benefit in having this convertor in between. It might have been useful, if CNS had to perform some operations on it.

cns/middlewares/k8sSwiftV2.go
shreyashastantram: Addressed

cns/configuration/configuration.go
shreyashastantram: removed

platform/os_windows.go
shreyashastantram: Addressed

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
shreyashastantram: It's already committed in another PR .

nddq:
@shreyashastantram can you hit the `Resolve conversation` if the issue discussed in that conversation has been resolved, make the PR less clutter, thanks 🙂

cns/NetworkContainerContract.go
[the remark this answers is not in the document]
shreyashastantram: Addressed

cns/restserver/util.go
shreyashastantram: This gets called in two places(main during start up and utils), I added this to ensure that cache doesn't get overwritten. How does this result in deadlock?

shreyashastantram: Not necessarily, Overwriting is fine since values will remain constant. I will remove that section.

cns/restserver/util.go
```diff
@@ -40,6 +40,31 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
 	service.state.Networks[networkName] = networkInfo
 }
 
+func (service *HTTPRestService) SetPnpIDMacaddressMapping(ctx context.Context) error {
+	p := platform.NewExecClient(nil)
+	VfMacAddressMapping, err := platform.FetchMacAddressPnpIDMapping(ctx, p)
+	if err != nil {
+		return errors.Wrap(err, "failed to fetch MACAddressPnpIDMapping")
+	}
+	logger.Printf("[DEBUG] Macaddress mapping: %d", len(VfMacAddressMapping))
```
nddq, line 49: Do we need this log line? This will be print out for both windows and linux

cns/service/main.go
```diff
@@ -819,6 +819,15 @@ func main() {
 	// Initialze state in if CNS is running in CRD mode
 	// State must be initialized before we start HTTPRestService
 	if config.ChannelMode == cns.CRD {
+
+		if cnsconfig.EnableSwiftV2 && cnsconfig.SWIFTV2Mode == configuration.K8sSWIFTV2 {
+			// No-op for linux, mapping is set for windows in aks swiftv2 scenario
+			logger.Printf("Fetcing backend nics for debug")
+			if err = httpRemoteRestService.SetPnpIDMacaddressMapping(context.Background()); err != nil {
```
nddq, line 826: can use rootCtx here instead of context.Background()

cns/restserver/util.go
```diff
@@ -40,6 +40,31 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
...
+	service.PnpIDByMacAddress = VfMacAddressMapping
+	return nil
+}
+
+func (service *HTTPRestService) getPNPIDFromMacAddress(ctx context.Context, macAddress string) (string, error) {
+	if _, ok := service.PnpIDByMacAddress[macAddress]; !ok {
+		if err := service.SetPnpIDMacaddressMapping(ctx); err != nil {
```
nddq, line 56: I'm not a fan of having `SetPnpIDMacaddressMapping` as a restservice function. `restservice` as it has already been overloaded with so many unrelated funcs already, plus all this function does is setting the service's PnpIDMacaddress map, can you make this a separated helper function that return the map instead?

nddq (changes requested):

cns/restserver/util.go
shreyashastantram: This function gets called from main function as well, Also, it saves response in service cache which is in rest server package.

cns/restserver/util.go
shreyashastantram: No, I will remove this, just added now for debugging. we haven't tested it yet.

cns/restserver/util.go
nddq: yes, we can still do that if this is a helper function instead. Since the only time in `SetPnpIDMacaddressMapping` referenced `service` is `service.PnpIDByMacAddress = VfMacAddressMapping`, so I think it would be a good idea to turn `SetPnpIDMacaddressMapping` into a separated helper function instead .i.e. `GetPnpIDMacaddressMapping` for example, and then we can set `service.PnpIDByMacAddress` to be the map returned by that func

cns/restserver/util.go
```diff
@@ -40,6 +40,35 @@ func (service *HTTPRestService) setNetworkInfo(networkName string, networkInfo *
 	service.state.Networks[networkName] = networkInfo
 }
 
+func (service *HTTPRestService) SavePnpIDMacaddressMapping(ctx context.Context) error {
+	// If mapping is already set, skip setting it again.
+	if len(service.state.PnpIDByMacAddress) != 0 {
```
nddq, line 45: I think there will be an issue with subsequent calls to this function, here is the timeline of events:
- `getPNPIDFromMacAddress` is called for a new pod, in which its MAC address doesnt exist in `service.state.PnpIDByMacAddress`
- `getPNPIDFromMacAddress` invokes `SavePnpIDMacaddressMapping`, but since `service.state.PnpIDByMacAddress` might be populated from the initial invoke, its length is > 0, so we return immediately, therefore not able to get the underlying updated PnpIDMacaddressMapping

nddq (changes requested):

crd/multitenancy/api/v1alpha1/utils.go
```diff
@@ -1,28 +1,11 @@
...
-				interfaceInfo.MacAddress == "" ||
-				interfaceInfo.GatewayIP == "" ||
-				interfaceInfo.DeviceType == "" {
-				return false
-			}
-		}
-	}
-
-	return true
+	return !reflect.DeepEqual(m.Status, MultitenantPodNetworkConfigStatus{})
```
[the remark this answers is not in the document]
shreyashastantram: We are populating all the once, if status is not present. That's when we can say mtpnc not ready.

cns/middlewares/k8sSwiftV2.go
```diff
@@ -194,19 +212,27 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
...
-					IPAddress:    ip,
-					PrefixLength: uint8(prefixSize),
-				},
-				MacAddress:        interfaceInfo.MacAddress,
-				NICType:           nicType,
-				SkipDefaultRoutes: false,
-				HostPrimaryIPInfo: cns.HostIPInfo{
-					Gateway: interfaceInfo.GatewayIP,
-				},
+			if nicType != cns.BackendNIC {
```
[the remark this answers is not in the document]
shreyashastantram: We have added a new filed, not sure why it's done for the same value. Let me use the new one. I think you meant we should use cns.NodeNetworkInterfaceBackendNIC right? The other will still work since it's basically string comparison, but i will change it to avoid confusions.

cns/restserver/util.go
shreyashastantram: To give bit more context to this, Mapping doesn't change, it will remain constant. We are doing this in two places just to cover the case of failures during the start up.

nddq (approved):

PR #2700, opened 2024-04-21
Feat: Added ETW logging support for CNS.

sivakami-projects:
  
- [x] includes documentation
- [ ] adds unit tests

**Notes**:


cns/logger/cnslogger.go
```diff
@@ -17,6 +19,9 @@ type CNSLogger struct {
 	DisableMetricLogging bool
 	DisableEventLogging  bool
 
+	ETWLogger        *zap.Logger
```
timraymond, line 22: This feels like it would benefit from being a [Chain of Responsibility](https://refactoring.guru/design-patterns/chain-of-responsibility`). It doesn't do anything dramatically different than the `logger` field in this struct (I think the method set is the same). So if that logger was like a stdout logger and this one was an ETW logger, it shouldn't matter exactly _what_ they do. There should just be a chain of loggers that get dispatched into sequentially with each log message to be logged.

cns/logger/cnslogger_windows.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get ETW core")
+	}
+	return etwcore, nil
+}
+
+func GetETWCore(loggingLevel zapcore.Level) (zapcore.Core, error) {
+	encoderConfig := zap.NewProductionEncoderConfig()
+	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	jsonEncoder := zapcore.NewJSONEncoder(encoderConfig)
```
timraymond, line 27: I think we've discussed this elsewhere, but JSON shouldn't be assumed. The whole power of structured logging is that it's structured--not structured _in a particular way_. The particulars can be configured by the user (maybe they want XML).

cns/logger/cnslogger.go
timraymond: Okay, makes sense. I'm not sure why I didn't see that as a `*zap.Logger`, which would serve the same purpose.

cns/logger/cnslogger.go
```diff
@@ -17,6 +19,9 @@ type CNSLogger struct {
 	DisableMetricLogging bool
 	DisableEventLogging  bool
 
+	zapLogger        *zap.Logger
+	enableETWLogging bool
```
rbtr, line 23: I don't think the logger needs to carry this config around. Use the config to instantiate the logger, then we simply use the logger.

rbtr (changes requested):

cns/logger/cnslogger_linux.go
```diff
@@ -0,0 +1,14 @@
...
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	// LogPath is the path where log files are stored.
+	LogPath = "/var/log/"
+)
+
+func GetPlatformCores(_ zapcore.Level, _ zapcore.Encoder) (zapcore.Core, error) {
+	return nil, nil
```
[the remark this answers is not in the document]
sivakami-projects: Thanks for sharing about Nop core. I had the flags in the CNS logger to ensure that we don't call the nil zap logger. With the Nop core I would the flags in the logging methods.

cns/logger/cnslogger.go
sivakami-projects: I had this flag to ensure zap logger instance is not nil, as we don't have zap implementation for Linux. With the Nop core I wouldn't need this. I will remove them.

cns/logger/cnslogger_windows.go
```diff
@@ -0,0 +1,29 @@
...
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	// LogPath is the path where log files are stored.
+	LogPath         = ""
+	etwCNSEventName = "AzureCNS"
+)
+
+func GetPlatformCores(loggingLevel zapcore.Level, encoder zapcore.Encoder) (zapcore.Core, error) {
```
rbtr, line 15: is there a reason that this remains exported?

cns/logger/cnslogger_windows.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/Azure/azure-container-networking/zapetw"
+	"github.com/pkg/errors"
+	"go.uber.org/zap/zapcore"
+)
+
+const (
+	// LogPath is the path where log files are stored.
+	LogPath         = ""
```
rbtr, line 11: where is this used? why is in the platform-specific files?

cns/logger/zaplogger.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+func initZapLogger(loggingLevel zapcore.Level, encoder zapcore.Encoder) (*zap.Logger, error) {
+	platformCore, err := GetPlatformCores(loggingLevel, encoder)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get Platform cores")
+	}
+	// Create a new logger with the platform core.
+	return zap.New(platformCore, zap.AddCaller()).With(zap.Int("pid", os.Getpid())), nil
+}
```
rbtr, line 18: Seems like we could do this in `NewCNSLogger()` also and don't need it extracted in to its own func

cns/logger/cnslogger_windows.go
sivakami-projects: No. Making it lower camel case.

cns/logger/cnslogger_windows.go
sivakami-projects: It is not used anywhere. Removing it.

cns/logger/cnslogger.go
```diff
@@ -28,7 +32,19 @@ func NewCNSLogger(fileName string, logLevel, logTarget int, logDir string) (*CNS
...
 	}
 
-	return &CNSLogger{logger: l}, nil
+	encoderConfig := zap.NewProductionEncoderConfig()
+	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	jsonEncoder := zapcore.NewJSONEncoder(encoderConfig)
+
+	zapLogger, err := initZapLogger(zapcore.DebugLevel, jsonEncoder)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get ETW logger")
```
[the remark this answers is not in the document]
sivakami-projects: Agreed, no need to mention about ETW here. Changed it.

rbtr (approved):

PR #2705, opened 2024-04-24
ci: Remove AKS Engine storage account usage

jpayne3506:
**Reason for Change**:

We no longer utilize this storage account for AKS engine or Jenkins. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/pipeline.yaml
+++ b/.pipelines/pipeline.yaml
@@ -184,17 +184,6 @@ stages:
                 pathtoPublish: "$(Build.ArtifactStagingDirectory)"
               condition: succeeded()
 
-            - task: AzureCLI@1
-              inputs:
-                azureSubscription: $(BUILD_VALIDATIONS_SERVICE_CONNECTION)
-                scriptLocation: "inlineScript"
-                inlineScript: |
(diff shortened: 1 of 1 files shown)


.pipelines/pipeline.yaml
```diff
@@ -606,34 +595,3 @@ stages:
...
-          steps:
-            - checkout: none
-            - task: AzureCLI@1
-              inputs:
-                azureSubscription: $(BUILD_VALIDATIONS_SERVICE_CONNECTION)
-                scriptLocation: "inlineScript"
-                inlineScript: |
-                  BUILD_NUMBER=$(Build.BuildNumber)
-                  BUILD_NUMBER=${BUILD_NUMBER//./-}
-                  echo Deleting storage container with name acn-$BUILD_NUMBER and account name $(STORAGE_ACCOUNT_NAME)
```
estebancams, line 637: We use acn-$(STORAGE_ID) to name the storage during creation. Why are we using acn-$BUID_NUMBER here? Are they the same? Should we use one or the other instead?

jpayne3506: This is getting removed.

PR #2710, opened 2024-04-25
feat: eBPF ingress/egress TC program for cilium external LB 

camrynl:
**Reason for Change**:

This is a POC for fixing external load balancer services on Cilium dual stack clusters. This PR includes both ingress and egress tc programs that convert a link local address to global unicast and vice versa, respectively. 

**Issue Fixed**:

https://github.com/cilium/cilium/issues/31326

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr:
Is this susceptible to the same ordering issues being discussed [here](https://github.com/microsoft/retina/issues/252#issuecomment-2077121621)? I notice you're trying to attach to Cilium's qdisc

camrynl:
I've been testing these latest changes on a cluster with cilium's `--bpf-filter-priority=2`.

Previously, I did see that my filters would get removed if I tried using the same pref + handle as cilium and restarted pods. When my filters are set at prior 1 and cilium at prior 2, I haven't had any issues with the filters deleting after restarting cilium

```
filter protocol all pref 1 bpf chain 0
filter protocol all pref 1 bpf chain 0 handle 0x1 ingress_filter direct-action not_in_hw id 953 tag eedbf352a3397a97 jited
filter protocol all pref 2 bpf chain 0
filter protocol all pref 2 bpf chain 0 handle 0x1 cil_from_netdev-eth0 direct-action not_in_hw id 6373 tag 2929474d7184a654 jited

filter protocol all pref 1 bpf chain 0
filter protocol all pref 1 bpf chain 0 handle 0x1 egress_filter direct-action not_in_hw id 952 tag ae5bd94fad468f22 jited
filter protocol all pref 2 bpf chain 0
filter protocol all pref 2 bpf chain 0 handle 0x1 cil_to_netdev-eth0 direct-action not_in_hw id 6952 tag a0c933d81fd07f41 jited
```

Later I plan to install the program using an initcontainer from the cilium or cns daemonset, so I'm not sure if this will change the behavior.

bpf-tc/pkg/ingress/bpf/ingress.c
```diff
@@ -0,0 +1,90 @@
...
+    const struct in6_addr LINKLOCAL_ADDR = {{{0xfe, 0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x12, 0x34, 0x56, 0x78, 0x9a, 0xbc}}};
+
+    struct in6_addr src_addr;
+    struct ipv6hdr ipv6_hdr;
+
+    int ret = bpf_skb_load_bytes(skb, ETH_HLEN + offsetof(struct ipv6hdr, saddr), &src_addr, sizeof(src_addr));
+    if (ret != 0)
+    {
+        bpf_printk("bpf_skb_load_bytes failed with error code %d.\n", ret);
+        return TC_ACT_SHOT;
```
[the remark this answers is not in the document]
camrynl: should we return UNSPEC here instead?

bpf-tc/pkg/egress/egress_bpfeb.o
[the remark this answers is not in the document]
camrynl: I'll remove these

bpf-tc/linux.Dockerfile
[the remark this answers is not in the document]
camrynl: Yeah there is an effort to make installer images for each component instead of using one dropgz image, though I don't think this is the case in AKS yet so we might need add this to the existing dropgz as well. 

Might change this, not sure if we want to go with dropgz to install the image or install it another way cc:@tamilmani1989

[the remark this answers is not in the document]
rbtr: how soon are you trying to roll this out? it should be embedded in the azure-ipam installer image since it's for the cilium scenario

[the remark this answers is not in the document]
camrynl: We want to keep this for cilium dualstack only, so I've updated the dockerfile and removed the use of dropgz

This image will be separate, and installed directly from the cilium daemonset

bpf-tc/pkg/ingress/bpf/ingress.c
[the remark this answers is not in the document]
camrynl: in cases where the packet fails to load I replaced the return with UNSPEC

bpf-prog/bpf-tc/README
```diff
@@ -0,0 +1,59 @@
+# bpf-tc
+
+`bpf-tc` is a project that leverages eBPF (Extended Berkeley Packet Filter) technology for traffic control in Linux kernel. This is a POC to fix external load balancer services in cilium dualstack clusters.
+
+## Description
+
+The goal of this bpf program is to fix the issue described [here](https://github.com/cilium/cilium/issues/31326). It includes both egress and ingress TC programs. These programs are meant to replace the nftable rules since they don't work on cilium clusters. 
+The egress bpf code converts the destination IP of the packet from global unicast to link local, and ingress converts the source IP from link local to global unicast.
```
tamilmani1989, line 8:
```suggestion
The egress bpf code converts the destination IPv6 of the packet from global unicast to link local, and ingress converts the source IPv6 from link local to global unicast.
```

bpf-prog/bpf-tc/cmd/bpf-tc/main.go
```diff
@@ -0,0 +1,68 @@
...
+var logger *zap.Logger
+
+func main() {
+	// Set up logger
+	logger, _ = zap.NewProduction()
+	defer logger.Sync()
+
+	// Remove resource limits for kernels <5.11.
+	if err := rlimit.RemoveMemlock(); err != nil {
+		logger.Error("Removing memlock", zap.Error(err))
```
tamilmani1989, line 23: not exiting on error?

bpf-prog/bpf-tc/cmd/bpf-tc/main.go
```diff
@@ -0,0 +1,68 @@
...
+	fq := &netlink.GenericQdisc{
+		QdiscAttrs: netlink.QdiscAttrs{
+			LinkIndex: iface.Index,
+			Handle:    netlink.MakeHandle(0xffff, 0),
+			Parent:    netlink.HANDLE_CLSACT,
+		},
+		QdiscType: "clsact",
+	}
+	if err := netlink.QdiscReplace(fq); err != nil {
+		logger.Error("failed setting egress qdisc", zap.Error(err))
```
tamilmani1989, line 42: should return on error

bpf-prog/bpf-tc/include/helper.h
```diff
@@ -0,0 +1,18 @@
+#include <netinet/in.h>
+#include <stdbool.h>
+
+#define L4_HDR_OFF (ETH_HLEN + sizeof(struct ipv6hdr))
+#define BPF_F_PSEUDO_HDR (1ULL << 4)
+
+static __always_inline bool compare_ipv6_addr(const struct in6_addr *addr1, const struct in6_addr *addr2)
+{
+#pragma unroll
```
tamilmani1989, line 9: what does `#pragma unroll` mean?

bpf-prog/bpf-tc/linux.Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.21 AS builder
+ARG VERSION
+WORKDIR /bpf-prog/bpf-tc
+COPY ./bpf-prog/bpf-tc .
+COPY ./bpf-prog/bpf-tc/cmd/bpf-tc/*.go /bpf-prog/bpf-tc/
+RUN apt-get clean
+RUN apt-get update && apt-get install -y llvm clang linux-libc-dev linux-headers-generic libbpf-dev libc6-dev gcc-multilib nftables iproute2
+RUN for dir in /usr/include/x86_64-linux-gnu/*; do ln -s "$dir" /usr/include/$(basename "$dir"); done
```
tamilmani1989, line 8: why this is required?

bpf-prog/bpf-tc/pkg/egress/bpf/egress.c
```diff
@@ -0,0 +1,78 @@
...
+            return TC_ACT_SHOT;
+        }
+
+        // Update the checksum
+        __be32 sum = bpf_csum_diff((__be32 *)GLOBAL_UNICAST_ADDR.s6_addr32, sizeof(GLOBAL_UNICAST_ADDR),
+                                   (__be32 *)LINKLOCAL_ADDR.s6_addr32, sizeof(LINKLOCAL_ADDR), 0);
+
+        int offset = offsetof(struct tcphdr, check);
+
+        int retu = bpf_l4_csum_replace(skb, L4_HDR_OFF + offset, 0, sum, BPF_F_PSEUDO_HDR);
```
tamilmani1989, line 67: can we reuse ret?

bpf-prog/bpf-tc/pkg/egress/bpf/egress.c
```diff
@@ -0,0 +1,78 @@
...
+    struct ipv6hdr ipv6_hdr;
+
+    int ret = bpf_skb_load_bytes(skb, ETH_HLEN + offsetof(struct ipv6hdr, daddr), &dst_addr, sizeof(dst_addr));
+    if (ret != 0)
+    {
+        bpf_printk("bpf_skb_load_bytes failed with error code %d.\n", ret);
+        return TC_ACT_UNSPEC;
+    }
+
+    int ret_hdr = bpf_skb_load_bytes(skb, ETH_HLEN, &ipv6_hdr, sizeof(ipv6_hdr));
```
tamilmani1989, line 34: can we comment whats the diff between two load_bytes call?

bpf-prog/bpf-tc/pkg/egress/bpf/egress.c
```diff
@@ -0,0 +1,78 @@
...
+    if (ret_hdr != 0)
+    {
+        bpf_printk("bpf_skb_load_bytes failed with error code %d.\n", ret_hdr);
```
tamilmani1989, line 37: lets be specific and return different error message instead of same as above..

bpf-prog/bpf-tc/pkg/egress/bpf/egress.c
```diff
@@ -0,0 +1,78 @@
...
+    // Check if the packet is TCP
+    if (ipv6_hdr.nexthdr != IPPROTO_TCP)
+        return TC_ACT_UNSPEC;
+
+    // Check the destination address to determine if it is a global unicast address
+    if (compare_ipv6_addr(&dst_addr, &GLOBAL_UNICAST_ADDR))
+    {
+
+        bpf_printk("Destination address is a global unicast address. Setting new addr to link local.\n");
+        bpf_printk("Destination address is %pI6.\n", &dst_addr);
```
tamilmani1989, line 50: should we consider skipping this and print only for debug mode otherwise it will print for every probe packet

bpf-prog/bpf-tc/pkg/ingress/bpf/ingress.c
```diff
@@ -0,0 +1,75 @@
...
+    if (ret != 0)
+    {
+        bpf_printk("bpf_skb_load_bytes failed with error code %d.\n", ret);
+        return TC_ACT_UNSPEC;
+    }
+
+    int ret_hdr = bpf_skb_load_bytes(skb, ETH_HLEN, &ipv6_hdr, sizeof(ipv6_hdr));
+    if (ret_hdr != 0)
+    {
+        bpf_printk("bpf_skb_load_bytes failed with error code %d.\n", ret_hdr);
```
tamilmani1989, line 35: different error message than above

bpf-prog/bpf-tc/pkg/ingress/bpf/ingress.c
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    // Check if the packet is TCP
+    if (ipv6_hdr.nexthdr != IPPROTO_TCP)
+        return TC_ACT_UNSPEC;
+
+    // Check the source address to determine if it is Link Local
+    if (compare_ipv6_addr(&src_addr, &LINKLOCAL_ADDR))
+    {
+        bpf_printk("Source address is a link local address. Setting new addr to global unicast.\n");
```
tamilmani1989, line 46: skip printing this and enable only for debug.. we can add DEBUG macro and that should be define only if debug mode is enabled. it can be cmdline parameter of go program

bpf-prog/bpf-tc/pkg/ingress/bpf/ingress.c
```diff
@@ -0,0 +1,75 @@
...
+            return TC_ACT_SHOT;
+        }
+
+        // Update the checksum
+        __be32 sum = bpf_csum_diff((__be32 *)LINKLOCAL_ADDR.s6_addr32, sizeof(LINKLOCAL_ADDR.s6_addr32),
+                                   (__be32 *)GLOBAL_UNICAST_ADDR.s6_addr32, sizeof(GLOBAL_UNICAST_ADDR.s6_addr32), 0);
+
+        int offset = offsetof(struct tcphdr, check);
+
+        int retu = bpf_l4_csum_replace(skb, L4_HDR_OFF + offset, 0, sum, BPF_F_PSEUDO_HDR);
```
tamilmani1989, line 64: reuse ret

tamilmani1989 (changes requested):

bpf-prog/bpf-tc/include/helper.h
timraymond: @tamilmani1989 Likely a directive to force the compiler to unroll loops. I'm speculating that it's probably necessary (along with `__always_inline`) to convince the BPF verifier that loops terminate.

Makefile
```diff
@@ -35,6 +35,7 @@ endif
 REPO_ROOT				 = $(shell git rev-parse --show-toplevel)
 REVISION				?= $(shell git rev-parse --short HEAD)
 ACN_VERSION				?= $(shell git describe --exclude "azure-ipam*" --exclude "dropgz*" --exclude "zapai*" --tags --always)
+BPF_TC_VERSION			?= $(notdir $(shell git describe --match "bpf-tc*" --tags --always))
```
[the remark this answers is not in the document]
camrynl: ok, renamed with this

bpf-prog/bpf-tc/cmd/bpf-tc/main.go
camrynl: added a return here

bpf-prog/bpf-tc/cmd/bpf-tc/main.go
camrynl: added return here as well

bpf-prog/bpf-tc/linux.Dockerfile
camrynl: several of the header files are not available when compiling in the container, we need this to to create symlinks from /usr/include/x86_64-linux-gnu to directories in /usr/include so that everything can be compiled

bpf-prog/bpf-tc/pkg/ingress/bpf/ingress.c
camrynl: added DEBUG macro, it can be defined if we pass DEBUG=true as a build-arg when building the image. 

in order to set the macro from the go program cmdline we'd have to use cgo and import C

bpf-prog/ipv6-healthprobe-tc/cmd/ipv6-healthprobe-tc/main.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/cilium/ebpf/rlimit"
+	"go.uber.org/zap"
+)
+
+var logger *zap.Logger
+
+func main() {
+	// Set up logger
+	logger, _ = zap.NewProduction()
```
vipul-21, line 18: This won't write to a file right? Can we use something similar to what we have for cni: https://github.com/Azure/azure-container-networking/blob/4370e6e93c7623a70446580417df1ab0fc2b4ff9/cni/log/logger.go#L31

bpf-prog/ipv6-healthprobe-tc/pkg/ingress/bpf/ingress.c
```diff
@@ -0,0 +1,80 @@
...
+    if (compare_ipv6_addr(&src_addr, &LINKLOCAL_ADDR))
+    {
+
+#ifdef DEBUG
+        bpf_printk("Source address is a link local address. Setting new addr to global unicast.\n");
+        bpf_printk("Source address is %pI6.\n", &src_addr);
+#endif
+
+        // Store the new source address in the packet
+        int ret = bpf_skb_store_bytes(skb, ETH_HLEN + offsetof(struct ipv6hdr, saddr),
```
vipul-21, line 55: `BPF_F_RECOMPUTE_CSUM` is working for ingress ?

bpf-prog/ipv6-healthprobe-tc/cmd/ipv6-healthprobe-tc/main.go
camrynl: These logs will just add a few lines to the container on startup, and if any error occurs those will show as well. From my understanding NewProductionEncoderConfig is to allow us to customize the log format. NewProduction comes with a preset encoder config. Do we need to customize log format in this case?

bpf-prog/ipv6-healthprobe-tc/pkg/ingress/bpf/ingress.c
camrynl: oh good catch, need to change that field to 0

bpf-prog/ipv6-healthprobe-tc/cmd/ipv6-healthprobe-tc/main.go
vipul-21: My concern is where will this get logged in filesystem ? Are we specifying any file to which the logs will be written. If it is, then we can ignore the comment.

camrynl: ah ok, updated this to write to ipv6-healthprobe-tc.log

bpf-prog/bpf-tc/pkg/ingress/bpf/ingress.c
tamilmani1989: wanted to run with existing binary with debug mode and not another build.. may be we can take as separate PR but it would be useful feature to add to debug in customer cluster when we found ebpf program not working

Makefile
```diff
@@ -360,6 +374,34 @@ azure-ipam-image-pull: ## pull azure-ipam container image.
 		IMAGE=$(AZURE_IPAM_IMAGE) \
 		TAG=$(AZURE_IPAM_PLATFORM_TAG)
 
+# ipv6-healthprobe-tc
+
+ipv6-healthprobe-tc-image-name: # util target to print the ipv6-healthprobe-tc image name.
```
tamilmani1989, line 379: instead of healthprobe can we name as hp. something like `ipv6-hp-bpf`? and do it everywhere

bpf-prog/ipv6-healthprobe-tc/cmd/ipv6-healthprobe-tc/main.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/cilium/ebpf/rlimit"
+	"go.uber.org/zap"
+)
+
+var logger *zap.Logger
+
+func main() {
+	// Set up logger
+	config := zap.NewProductionConfig()
+	config.OutputPaths = []string{"stdout", "/var/log/ipv6-healthprobe-tc.log"}
```
tamilmani1989, line 21: can we log it as `/var/log/azure-ipv6-hp-bpf.log`?

bpf-prog/ipv6-healthprobe-tc/cmd/ipv6-healthprobe-tc/main.go
```diff
@@ -0,0 +1,92 @@
...
+	// Create a qdisc filter for traffic on the interface.
+	fq := &netlink.GenericQdisc{
+		QdiscAttrs: netlink.QdiscAttrs{
+			LinkIndex: iface.Index,
+			Handle:    netlink.MakeHandle(0xffff, 0),
+			Parent:    netlink.HANDLE_CLSACT,
+		},
+		QdiscType: "clsact",
+	}
+	if err := netlink.QdiscReplace(fq); err != nil {
```
tamilmani1989, line 64: does replace deletes and creates filter?

bpf-prog/ipv6-healthprobe-tc/pkg/egress/bpf/egress.c
```diff
@@ -0,0 +1,82 @@
...
+    const struct in6_addr LINKLOCAL_ADDR = {{{0xfe, 0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x12, 0x34, 0x56, 0x78, 0x9a, 0xbc}}};
+
+    // Define the global unicast address 2603:1062:0000:0001:fe80:1234:5678:9abc
+    const struct in6_addr GLOBAL_UNICAST_ADDR = {{{0x26, 0x03, 0x10, 0x62, 0x00, 0x00, 0x00, 0x01, 0xfe, 0x80, 0x12, 0x34, 0x56, 0x78, 0x9a, 0xbc}}};
+
+    struct in6_addr dst_addr;
+    struct ipv6hdr ipv6_hdr;
+
+    // Load the destination address from the packet
+    int ret = bpf_skb_load_bytes(skb, ETH_HLEN + offsetof(struct ipv6hdr, daddr), &dst_addr, sizeof(dst_addr));
```
tamilmani1989, line 28: can we move this after line 46?

bpf-prog/ipv6-healthprobe-tc/pkg/ingress/bpf/ingress.c
```diff
@@ -0,0 +1,80 @@
...
+    struct in6_addr src_addr;
+    struct ipv6hdr ipv6_hdr;
+
+    // Load the source address from the packet
+    int ret = bpf_skb_load_bytes(skb, ETH_HLEN + offsetof(struct ipv6hdr, saddr), &src_addr, sizeof(src_addr));
+    if (ret != 0)
+    {
+        bpf_printk("bpf_skb_load_bytes failed to load source address with error code %d.\n", ret);
+        return TC_ACT_UNSPEC;
+    }
```
tamilmani1989, line 31: same as above

bpf-prog/ipv6-healthprobe-tc/cmd/ipv6-healthprobe-tc/main.go
camrynl: yes as long as there is an existing qdisc with the same attributes as this new one, it will be replaced

Makefile
camrynl: ok, updated

bpf-prog/ipv6-hp-bpf/pkg/ingress/bpf/ingress.c
```diff
@@ -0,0 +1,80 @@
...
+
+    struct in6_addr src_addr;
+    struct ipv6hdr ipv6_hdr;
+
+    // Check if the packet is TCP
+    if (ipv6_hdr.nexthdr != IPPROTO_TCP)
+        return TC_ACT_UNSPEC;
+
+    // Load the source address from the packet
+    int ret = bpf_skb_load_bytes(skb, ETH_HLEN + offsetof(struct ipv6hdr, saddr), &src_addr, sizeof(src_addr));
```
tamilmani1989, line 30: move to a place where its being used. update in other program as well

bpf-prog/ipv6-hp-bpf/pkg/ingress/ingress.go
```diff
@@ -0,0 +1,57 @@
...
+	filters, err := netlink.FilterList(link, netlink.HANDLE_MIN_INGRESS)
+	if err != nil {
+		logger.Error("Failed to get filter list", zap.Error(err))
+		return err
+	}
+
+	// Check if egress filter exists and delete it. Filter is identified by priority=1 and handle=1.
+	// this is to avoid duplicate filters after restarting the daemonset
+	for _, filter := range filters {
+		if filter.Attrs().Priority == 1 && filter.Attrs().Handle == 1 {
```
tamilmani1989, line 28: what if kernel gives different handle id? can't we filter based on bpf prog name?

tamilmani1989 (approved):

vipul-21 (approved):

Issue #2703, opened 2024-04-24
Telemetry leaks sockets on error


QxBytes:
**What happened:** 
Starting telemetry gives error 
```
2024/04/20 00:17:26 [9756] StartServer: unmarshal error:invalid character 'm' looking for beginning of value
2024/04/24 16:39:41 [9756] StartServer: unmarshal error:invalid character ':' after top-level value
2024/04/24 16:39:42 [9756] StartServer: unmarshal error:invalid character ':' looking for beginning of value
```
and does not clean up/close connections on error.

**What you expected to happen:**  
The unmarshal error should not occur and the connections should be closed properly on error.

**How to reproduce it:**  
Start an AKS cluster and create a pod.

**Orchestrator and Version (e.g. Kubernetes, Docker):**  

**Operating System (Linux/Windows):**  

**Kernel (e.g. `uanme -a` for Linux or `$(Get-ItemProperty -Path "C:\windows\system32\hal.dll").VersionInfo.FileVersion` for Windows):**

**Anything else we need to know?:**
[Miscellaneous information that will assist in solving the issue.]

QxBytes:
Fix has rolled out to aks (cni versions 1.4.54 and 1.5.28)

PR #2711, opened 2024-04-25
fix: close connection on any return from azure-vnet-telemetry

QxBytes:
**Reason for Change**:

**Issue Fixed**:

Prevents leaking of sockets when azure-vnet-telemetry errors and fails to unmarshal mentioned in #2703 (this pr does not fix the unmarshaling error itself)

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
The unit test uses time.sleep to ensure the connection populates in the server, and we will be rewriting these tests later anyway with stateless cni.

### DIFF
--- a/telemetry/telemetrybuffer.go
+++ b/telemetry/telemetrybuffer.go
@@ -116,6 +116,27 @@ func (tb *TelemetryBuffer) StartServer() error {
 				tb.connections = append(tb.connections, conn)
 				tb.mutex.Unlock()
 				go func() {
+					defer func() {
+						var index int
+						var value net.Conn
+						var found bool
+
+						tb.mutex.Lock()
+						defer tb.mutex.Unlock()
+
+						for index, value = range tb.connections {
+							if value == conn {
+								conn.Close()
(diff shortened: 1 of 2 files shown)


telemetry/telemetrybuffer.go
```diff
@@ -116,6 +116,27 @@ func (tb *TelemetryBuffer) StartServer() error {
...
+						}
+
+						if found {
+							tb.connections = remove(tb.connections, index)
+						}
+					}()
+
 					for {
 						reportStr, err := read(conn)
 						if err == nil {
```
thatmattlong, line 142: I suggest inverting this now, we tend to prefer to handle errors first, so we should do
```go
if err != nil {
    return
}
```
then we can drop the else too, so we drop a level of indentation

tamilmani1989 (commented):
is it feasible to have UT for this?

tamilmani1989 (dismissed):

tamilmani1989 (approved):

PR #2716, opened 2024-04-29
Feat: adds a status check for NCs

abel2-code:
**Reason for Change**:
This is part of my intern project to help improve error messages.

**Issue Fixed**:
N/A

**Requirements**:
  
- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/ipam.go
+++ b/cns/restserver/ipam.go
@@ -116,6 +116,23 @@ func (service *HTTPRestService) RequestIPConfigHandler(w http.ResponseWriter, r 
 		return
 	}
 
+	// Check the status of the NC
+	ncStatuses := multiContainerStatus(service.state.ContainerStatus)
+	if unsuccessfulStatuses := ncStatuses.GetUnsuccessfulStatusErrors(); unsuccessfulStatuses != "" {
+		// If the status is anything other than success, we send a response back with the actual status and NC ID.
+		reserveResp := &cns.IPConfigResponse{
+			Response: cns.Response{
+				ReturnCode: types.FailedToAllocateIPConfig,
+				Message:    unsuccessfulStatuses,
+			},
+		}
+
+		w.Header().Set(cnsReturnCode, reserveResp.Response.ReturnCode.String())
+		err = service.Listener.Encode(w, &reserveResp)
+		logger.ResponseEx(service.Name+operationName, ipconfigRequest, reserveResp, reserveResp.Response.ReturnCode, err)
+		return
+	}
+
 	// doesn't fill in DesiredIPAddresses if it is empty in the original request
 	ipconfigsRequest := cns.IPConfigsRequest{
 		PodInterfaceID:      ipconfigRequest.PodInterfaceID,
@@ -175,6 +192,24 @@ func (service *HTTPRestService) RequestIPConfigsHandler(w http.ResponseWriter, r
 	if err != nil {
 		return
 	}
+
+	// Check the status of the NC
+	ncStatuses := multiContainerStatus(service.state.ContainerStatus)
+	if unsuccessfulStatuses := ncStatuses.GetUnsuccessfulStatusErrors(); unsuccessfulStatuses != "" {
+		// If any status is anything other than success, we send a response back with the actual status and NC ID.
+		reserveResp := &cns.IPConfigResponse{
+			Response: cns.Response{
+				ReturnCode: types.FailedToAllocateIPConfig,
+				Message:    unsuccessfulStatuses,
+			},
+		}
+
+		w.Header().Set(cnsReturnCode, reserveResp.Response.ReturnCode.String())
+		err = service.Listener.Encode(w, &reserveResp)
+		logger.ResponseEx(service.Name+operationName, ipconfigsRequest, reserveResp, reserveResp.Response.ReturnCode, err)
+		return
+	}
+
 	var ipConfigsResp *cns.IPConfigsResponse
 
 	// Check if IPConfigsHandlerMiddleware is set
--- a/cns/restserver/restserver.go
+++ b/cns/restserver/restserver.go
@@ -2,9 +2,11 @@ package restserver
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/http"
 	"net/http/pprof"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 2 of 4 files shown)


cns/restserver/restserver.go
```diff
@@ -135,6 +137,21 @@ type containerstatus struct {
 	VfpUpdateComplete             bool // True when VFP programming is completed for the NC
 }
 
+type multiContainerStatus map[string]containerstatus
+
+func (mcs *multiContainerStatus) GetUnsuccessfulStatusErrors() string {
```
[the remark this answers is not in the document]
nddq: (1/2) also not a fan of this function returning a string, maybe just return a list of NCs that doesn't have the `NCUpdateSuccess` status, and then we can do the processing in the caller?

cns/restserver/ipam.go
```diff
@@ -116,6 +116,23 @@ func (service *HTTPRestService) RequestIPConfigHandler(w http.ResponseWriter, r
 		return
 	}
 
+	// Check the status of the NC
+	ncStatuses := multiContainerStatus(service.state.ContainerStatus)
+	if unsuccessfulStatuses := ncStatuses.GetUnsuccessfulStatusErrors(); unsuccessfulStatuses != "" {
```
nddq, line 121: (2/2) and then we can just check for the length of the list returned.

cns/restserver/restserver.go
```diff
@@ -135,6 +137,21 @@ type containerstatus struct {
 	VfpUpdateComplete             bool // True when VFP programming is completed for the NC
 }
 
+type multiContainerStatus map[string]containerstatus
```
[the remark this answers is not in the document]
nairashu: +1. This does not need to be a type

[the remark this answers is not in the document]
nairashu: This variable is actually not needed as you are simply duplicating 'service.state.ContainerStatus'

cns/restserver/ipam.go
```diff
@@ -116,6 +116,23 @@ func (service *HTTPRestService) RequestIPConfigHandler(w http.ResponseWriter, r
...
+		// If the status is anything other than success, we send a response back with the actual status and NC ID.
+		reserveResp := &cns.IPConfigResponse{
+			Response: cns.Response{
+				ReturnCode: types.FailedToAllocateIPConfig,
+				Message:    unsuccessfulStatuses,
```
nairashu, line 126: All you are trying to do is get the NC Status and create an error message from it. So change the name of the method to 'GetNcStatusErrorMessage'

cns/restserver/restserver.go
```diff
@@ -135,6 +137,21 @@ type containerstatus struct {
...
+	for ncID := range *mcs {
+		ncStatus := (*mcs)[ncID].CreateNetworkContainerRequest.NCStatus
+		if ncStatus != cns.NCUpdateSuccess {
+			unsuccessfulStatus := fmt.Sprintf("Expected status for NC %s to be %s but got %s", ncID, string(cns.NCUpdateSuccess), string(ncStatus))
+			unsuccessfulStatuses = append(unsuccessfulStatuses, unsuccessfulStatus)
+		}
+	}
+
+	return strings.Join(unsuccessfulStatuses, "\n")
+}
```
nairashu, line 153:
```suggestion
func GetNcStatusErrorMessages(mcs map[string]containerstatus) string {
	var unsuccessfulStatuses []string
	for ncID, containerStatus := range mcs {
		ncStatus := containerStatus.CreateNetworkContainerRequest.NCStatus
		if ncStatus != cns.NCUpdateSuccess {
			unsuccessfulStatuses = append(unsuccessfulStatuses, fmt.Sprintf("Expected status for NC %s to be %s but got %s. ", ncID, string(cns.NCUpdateSuccess), string(ncStatus)))
		}
	}

	return strings.Join(unsuccessfulStatuses, "\n")
}
```

cns/restserver/restserver.go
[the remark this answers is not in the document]
nairashu: The purpose of the method is the create the CNI Error message so it is ok to return the string here directly.

cns/restserver/ipam.go
```diff
@@ -116,6 +116,23 @@ func (service *HTTPRestService) RequestIPConfigHandler(w http.ResponseWriter, r
...
+			},
+		}
+
+		w.Header().Set(cnsReturnCode, reserveResp.Response.ReturnCode.String())
+		err = service.Listener.Encode(w, &reserveResp)
+		logger.ResponseEx(service.Name+operationName, ipconfigRequest, reserveResp, reserveResp.Response.ReturnCode, err)
+		return
+	}
```
nairashu, line 134: By checking the NCStatus early you are blocking the call to the service. I don't think it is the right behavior here. This will block IP allocations from the IPs that are available on the node itself and are getting reclaimed while the NC status is SubnetFull or something else. We should not be determining the response state from the state of the NC. Basically it should be more of a helper method to add to the response message on line 153.

cns/restserver/restserver.go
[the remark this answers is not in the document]
timraymond: I somewhat disagree with the sentiment in this thread. This should be an `error` implementation if anything, but the problem as it currently stands is that `GetUnsuccessfulStatusErrors` can return `""` making it a poor `error` (since in its typical usage `err != nil`, but it is, in fact, nil).

So the idea of having some function determine whether or not there were errors is a good one, but it should actually produce an error or nil--not a string.

It should read something like this:

```go
err := anyFailed(service.state.ContainerStatus)
if err != nil {
       // create the response but use `err.Error()`
}
```

You could have an error type like this to better deal with this:

```go
type MultiContainerError map[string]v1alpha1.NCStatus

func (m *MultiContainerError) Error() string {
       out := bytes.NewBufferString("multiple NCs failed: ")
       for ncid, status := range m {
              fmt.Fprintf(out, "%s (%s) ")
       }
       return out.String()
}
```

cns/restserver/restserver.go
```diff
@@ -135,6 +137,21 @@ type containerstatus struct {
...
 }
 
+type multiContainerStatus map[string]containerstatus
+
+func (mcs *multiContainerStatus) GetUnsuccessfulStatusErrors() string {
+	var unsuccessfulStatuses []string
+	for ncID := range *mcs {
+		ncStatus := (*mcs)[ncID].CreateNetworkContainerRequest.NCStatus
+		if ncStatus != cns.NCUpdateSuccess {
+			unsuccessfulStatus := fmt.Sprintf("Expected status for NC %s to be %s but got %s", ncID, string(cns.NCUpdateSuccess), string(ncStatus))
```
timraymond, line 147: This is good output for a test, but not for an error. Generally, it should be concise and lower-case. As in my other comment, I think something like this would be more typical:

`multiple NCs failed: uuid1 (status1), uuid2 (status2), uuid3 (status3)`

It's implied that the expectation was success, so logging that just consumes log disk space. Though that is perfectly fine (and desirable!) to mention in `go test` output.

cns/restserver/restserver.go
```diff
@@ -135,6 +137,21 @@ type containerstatus struct {
...
+			unsuccessfulStatuses = append(unsuccessfulStatuses, unsuccessfulStatus)
+		}
+	}
+
+	return strings.Join(unsuccessfulStatuses, "\n")
```
timraymond, line 152: Also, it's unusual to join by newlines for errors since the log lines themselves are newline separated

PR #2717, opened 2024-04-29
feat: update CRDs to accomodate multiple pod nics

thatmattlong:
**Reason for Change**:
Updates the multitenancy CRD schemas to support multiple Pod NICs

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


crd/multitenancy/api/v1alpha1/podnetwork.go
```diff
@@ -36,15 +37,31 @@ type PodNetworkList struct {
...
 // PodNetworkSpec defines the desired state of PodNetwork
 type PodNetworkSpec struct {
+	// NetworkID is the identifier for the network, e.g. vnet guid or IB network ID
 	// +kubebuilder:validation:Optional
-	// customer vnet guid
-	VnetGUID string `json:"vnetGUID,omitempty"`
+	NetworkID string `json:"networkID,omitempty"`
+	// DeviceType is the device type that is required by this network
+	// +kubebuilder:validation:Optional
+	DeviceType DeviceType `json:"deviceType,omitempty"`
```
estebancams, line 55: neat: could we name this NetworkType? I see that the device determines the type of network it will live in, but why not just call it networkType?

thatmattlong: we want the values to match when we use device plugin, so orchestrator will pass, for example, `acn.azure.com/vnet-cni` to both `podNetwork.spec.deviceType` and the `pod.spec.resources.limits."acn.azure.com/vnet-cni"`. If we chose "networkType" instead, the orchestrators would then have to know about two separate magic values and maintain a mapping between them, which we don't want.

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -51,16 +47,34 @@ type MultitenantPodNetworkConfigSpec struct {
...
 
+type InterfaceInfo struct {
+	// NCID is the network container id
+	NCID string `json:"ncID,omitempty"`
+	// PrimaryIP is the ip allocated to the network container
+	PrimaryIP string `json:"primaryIP,omitempty"`
+	// MacAddress is the MAC Address of the VM's NIC which this network container was created for
+	MacAddress string `json:"macAddress,omitempty"`
+	// GatewayIP is the gateway ip of the injected subnet
+	GatewayIP string `json:"gatewayIP,omitempty"`
```
kmurudi, line 58: we planned on adding NICType as a field too withing InterfaceInfos so singularity can directly start using & writing to `InterfaceInfos[`] instead of adding NICType in `MultiTenantPodNetworkConfigStatus`

thatmattlong: done

crd/multitenancy/manifests/multitenancy.acn.azure.com_podnetworkinstances.yaml
```diff
@@ -71,6 +83,16 @@ spec:
...
                   type: string
                 type: array
+              podNetworkStatuses:
+                additionalProperties:
+                  description: PNIStatus indicates the status of PNI
+                  enum:
+                  - Ready
+                  - CreateReservationSetError
+                  - PodNetworkNotReady
+                  - InsufficientIPAddressesOnSubnet
```
prmathur-microsoft, line 93: Are you expecting us to add more for IB related here?

thatmattlong: We can if we need more but I don't think there is anything to add for IB as there is no real processing done for IB here, the only relevant ones would be Ready or PodNetworkNotReady, as IB doesn't create static reservations or use IP addresses

crd/multitenancy/api/v1alpha1/podnetworkinstance.go
```diff
@@ -36,19 +35,35 @@ type PodNetworkInstanceList struct {
...
+	// +kubebuilder:validation:Optional
+	PodNetwork string `json:"podnetwork,omitempty"`
+	// Deprecated - use PodNetworks
 	// +kubebuilder:default=0
-	PodIPReservationSize int `json:"podIPReservationSize"`
+	PodIPReservationSize int `json:"podIPReservationSize,omitempty"`
+	// PodNetworks describes each PodNetwork to attach to a single Pod
+	// optional for now in case orchestrator uses the deprecated fields
+	// +kubebuilder:validation:Optional
+	PodNetworks []PodNetworkInstanceItem `json:"podNetworks"`
```
estebancams, line 58: Suggestion: rename this to PodNetworkInstances[]. I think given that we are pushing down legacy fields from PNISpec to PNIItem, it would make sense to keep the same name for each of the items since the still represents the same idea of a 'reservation' instead of PodNetwork that represents a subnet registration

thatmattlong: I think it may be more confusing for a PodNetworkInstance to contain PodNetworkInstances

thatmattlong: I agree this is confusing as-is, going to rename to `PodNetworkConfigs`

estebancams (approved):

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -51,16 +47,38 @@ type MultitenantPodNetworkConfigSpec struct {
...
+	// Deprecated - use InterfaceInfos
+	// +kubebuilder:validation:Optional
 	MacAddress string `json:"macAddress,omitempty"`
-	// Gateway IP
+	// Deprecated - use InterfaceInfos
+	// +kubebuilder:validation:Optional
 	GatewayIP string `json:"gatewayIP,omitempty"`
+	// InterfaceInfos describes all of the network container goal state for this Pod
+	// +kubebuilder:validation:Optional
+	InterfaceInfos []InterfaceInfo `json:"interfaceInfos,omitempty"`
```
kmurudi, line 81: this will be populated then ignore higher fields in mtpncStatus

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -51,16 +47,38 @@ type MultitenantPodNetworkConfigSpec struct {
...
+	// DeviceType is the device type that this NC was created for
+	DeviceType DeviceType `json:"deviceType,omitempty"`
+}
+
 // MultitenantPodNetworkConfigStatus defines the observed state of PodNetworkConfig
 type MultitenantPodNetworkConfigStatus struct {
-	// network container id
+	// Deprecated - use InterfaceInfos
+	// +kubebuilder:validation:Optional
 	NCID string `json:"ncID,omitempty"`
```
kmurudi, line 69: these will be populated only for delegatedVMNIC (vnet nic & non-accelnet)

PR #2722, opened 2024-04-30
feat: Adding MACAddress to CNS endpoint State and refactoring stateless CNI code

behzad-mir:
Adding MACAddress to CNS endpoint State and refactoring stateless CNI code

- [ ] includes documentation
- [X] adds unit tests

**Notes**:


cns/restserver/ipam.go
```diff
@@ -1136,25 +1136,39 @@ func (service *HTTPRestService) UpdateEndpointHelper(endpointID string, req map[
...
 	logger.Printf("[updateEndpoint] Updating endpoint state for infra container %s", endpointID)
 	if endpointInfo, ok := service.EndpointState[endpointID]; ok {
-		for ifName, InterfaceInfo := range req {
-			logger.Printf("[updateEndpoint] Found existing endpoint state for infra container %s", endpointID)
-			if InterfaceInfo.HnsEndpointID != "" {
-				service.EndpointState[endpointID].IfnameToIPMap[ifName].HnsEndpointID = InterfaceInfo.HnsEndpointID
-				logger.Printf("[updateEndpoint] update the endpoint %s with HNSID  %s", endpointID, InterfaceInfo.HnsEndpointID)
+		for ifName, interfaceInfo := range req {
+			// This codition will create a map for SecodaryNIC and also also creates MAP entry for InfraNic in case that the initial goalState is using empty InterfaceName
+			if _, keyExist := service.EndpointState[endpointID].IfnameToIPMap[ifName]; !keyExist {
```
timraymond, line 1141: I think this needs some commenting. There's a lot of complex map manipulation (types in maps of maps) happening here that is not obvious to readers. Either that, or it might be more helpful to encapsulate what's happening here in a higher-level type whose operations are more intention-revealing.

network/manager.go
```diff
@@ -710,21 +712,21 @@ func cnsEndpointInfotoCNIEpInfo(endpointInfo restserver.EndpointInfo, endpointID
...
-		ifNametoIPInfoMap[ep.IfName].NICType = cns.InfraNIC
-		ifNametoIPInfoMap[ep.IfName].HnsEndpointID = ep.HnsId
-		ifNametoIPInfoMap[ep.IfName].HostVethName = ep.HostIfName
-	}
-	if ep.SecondaryInterfaces != nil {
-		for ifName, InterfaceInfo := range ep.SecondaryInterfaces {
-			ifNametoIPInfoMap[ifName].NICType = InterfaceInfo.NICType
-		}
+	ifNametoIPInfoMap[ep.IfName] = &restserver.IPInfo{
+		NICType:       cns.InfraNIC,
```
QxBytes, line 725: does the code break if we do `ep.NICType` ?

network/endpoint.go
```diff
@@ -30,6 +30,7 @@ type AzureHNSEndpoint struct{}
 type endpoint struct {
 	Id                       string
 	HnsId                    string `json:",omitempty"`
+	HNSNetworkID             string `json:",omitempty"`
```
QxBytes, line 33: can we add a comment that this is used only in windows?

behzad-mir: HNS is only for Windows so no need for comment

network/manager.go
behzad-mir: I do not see any NICType in ep. If it's created and filled out then it can be used

cns/restserver/ipam.go
behzad-mir: A comment has been made on top of the for block

network/manager.go
```diff
@@ -425,6 +425,7 @@ func (nm *networkManager) CreateEndpoint(cli apipaClient, networkID string, epIn
 func (nm *networkManager) UpdateEndpointState(ep *endpoint) error {
 	ifnameToIPInfoMap := generateCNSIPInfoMap(ep) // key : interface name, value : IPInfo
 	logger.Info("Calling cns updateEndpoint API with ", zap.String("containerID: ", ep.ContainerID), zap.String("HnsId: ", ep.HnsId), zap.String("HostIfName: ", ep.HostIfName))
+	logger.Info("ifnameToIPInfoMap:", zap.Any("ifnameToIPInfoMap", ifnameToIPInfoMap))
```
nddq, line 428: do we need to log out the entire map? also usage of `zap.Any` is not reccomended 🥲

cns/restserver/ipam.go
nddq: this is still very complicated to read. I agree with @timraymond , we should do some encapsulation here.

network/manager.go
```diff
@@ -710,21 +712,21 @@ func cnsEndpointInfotoCNIEpInfo(endpointInfo restserver.EndpointInfo, endpointID
...
 		epInfo.HostIfName = ipInfo.HostVethName
 		epInfo.HNSEndpointID = ipInfo.HnsEndpointID
+		epInfo.HNSNetworkID = ipInfo.HnsNetworkID
+		epInfo.MacAddress = net.HardwareAddr(ipInfo.MacAddress)
 	}
 	return epInfo
 }
 
+// generateCNSIPInfoMap generates a CNS ifNametoIPInfoMap structure based on CNI endpoint
 func generateCNSIPInfoMap(ep *endpoint) map[string]*restserver.IPInfo {
```
nddq, line 722: im curious about why we are no longer looping through all secondary interfaces and adding them to the map, what has changed?

behzad-mir: This will be added with another PR by @QxBytes since he is working on Swift 2.0 changes and couple of refactores related to epInfo

network/manager.go
behzad-mir: I was using it for debugging purposes on my tests. It's removed

cns/restserver/ipam.go
behzad-mir: @timraymond @nddq @tamilmani1989 
Added a new commit addressing the issue with a new func that encapsulates and simplified the block

tamilmani1989 (approved):

nddq (approved):

PR #2723, opened 2024-05-01
remove unnecessary log info

paulyufan2:
**Reason for Change**:

This PR is to remove some unnecessary logs from CNI

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/ipam/pool.go
+++ b/ipam/pool.go
@@ -293,7 +293,7 @@ func (as *addressSpace) requestPool(poolId string, subPoolId string, options map
 	var ap *addressPool
 	var err error
 
-	logger.Info("Requesting pool with", zap.String("poolId", poolId), zap.Any("options", options), zap.Any("v6", v6))
+	logger.Info("Requesting a new address pool")
 
 	if poolId != "" {
 		// Return the specific address pool requested.
@@ -377,9 +377,9 @@ func (as *addressSpace) releasePool(poolId string) error {
 	}
 
 	if addressesInUse = ap.IsAnyRecordInUse(); addressesInUse {
-		logger.Info("Skip releasing pool with due to address being in use", zap.String("poolId", poolId))
+		logger.Info("Skip releasing pool with due to address being in use")
 	} else {
-		logger.Info("Releasing pool as there are no allocations", zap.String("poolId", poolId))
+		logger.Info("Releasing pool as there are no allocations")
 		ap.RefCount = 0
 	}
 
@@ -459,7 +459,7 @@ func (ap *addressPool) requestAddress(address string, options map[string]string)
 	var addr *net.IPNet
 	id := options[OptAddressID]
 
-	logger.Info("Requesting address with", zap.String("address", address), zap.Any("options", options))
+	logger.Info("Requesting a new address from the address pool")
 
 	if address != "" {
 		// Return the specific address requested.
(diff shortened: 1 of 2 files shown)


jpayne3506 (approved):

ipam/pool.go
```diff
@@ -526,7 +526,7 @@ func (ap *addressPool) releaseAddress(address string, options map[string]string)
 	var err error
 
 	logger.Info("Releasing address with", zap.String("address", address), zap.Any("options", options))
-	defer func() { logger.Error("Address release completed with", zap.String("address", address), zap.Error(err)) }()
+	defer func() { logger.Error("Address release completed with", zap.Error(err)) }()
```
tamilmani1989, line 529: why are we removing address field.. this helps in debugging

network/endpoint.go
```diff
@@ -174,14 +174,14 @@ func (nw *network) deleteEndpoint(nl netlink.NetlinkInterface, plc platform.Exec
...
-			logger.Error("Failed to delete endpoint with", zap.String("endpointID", endpointID), zap.Error(err))
+			logger.Error("Failed to delete endpoint with", zap.Error(err))
 		}
 	}()
 
 	// Look up the endpoint.
 	ep, err := nw.getEndpoint(endpointID)
 	if err != nil {
-		logger.Error("Endpoint not found. Not Returning error", zap.String("endpointID", endpointID), zap.Error(err))
+		logger.Error("Endpoint not found. Not Returning error", zap.Error(err))
```
tamilmani1989, line 184: can we log endpoint id in different way instead of removing it?

network/endpoint.go
```diff
@@ -290,7 +290,7 @@ func (ep *endpoint) attach(sandboxKey string) error {
 
 	ep.SandboxKey = sandboxKey
 
-	logger.Info("Attached endpoint to sandbox", zap.String("id", ep.Id), zap.String("sandboxKey", sandboxKey))
+	logger.Info("Attached endpoint to sandbox", zap.String("id", ep.Id))
```
ZetaoZhuang, line 293: looks like zap.string("id", xxx) is reused in lots of places. maybe worth to add a common func for zap field for id going forward.  just an idea. 

func ID(id string) zap.Field {
	return zap.String("id", id)
}

ipam/pool.go
```diff
@@ -293,7 +293,7 @@ func (as *addressSpace) requestPool(poolId string, subPoolId string, options map
 	var ap *addressPool
 	var err error
 
-	logger.Info("Requesting pool with", zap.String("poolId", poolId), zap.Any("options", options), zap.Any("v6", v6))
+	logger.Info("Requesting a new address pool")
```
ramiro-gamarra, line 296: Since you are removing details about the request, is this line needed at all? I'm asking because without any info, the log line cannot be correlated to other things. It mostly indicates that the function was invoked, but you also have a log line below that accomplishes the same thing (I don't see a code path where the other line does not get called).

ipam/pool.go
```diff
@@ -377,9 +377,9 @@ func (as *addressSpace) releasePool(poolId string) error {
 	}
 
 	if addressesInUse = ap.IsAnyRecordInUse(); addressesInUse {
-		logger.Info("Skip releasing pool with due to address being in use", zap.String("poolId", poolId))
+		logger.Info("Skip releasing pool with due to address being in use")
 	} else {
-		logger.Info("Releasing pool as there are no allocations", zap.String("poolId", poolId))
+		logger.Info("Releasing pool as there are no allocations")
```
ramiro-gamarra, line 382: why remove pool id?

PR #2735, opened 2024-05-08
add azr metrics in heartbeats

ZetaoZhuang:
**Reason for Change**:

we want to add IsAZRSupported and HomeAZ as custom metrics in heartbeats, that way we know if a node has azr enabled or not.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


timraymond:
It looks like this suffers from some import cycles, so it won't compile.

cns/internal/heartbeat.go
```diff
@@ -0,0 +1,40 @@
...
+	"context"
+	"strconv"
+	"time"
+
+	"github.com/Azure/azure-container-networking/aitelemetry"
+	"github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/Azure/azure-container-networking/cns/restserver"
+	"github.com/Azure/azure-container-networking/cns/types"
+)
+
```
smittal22, line 16: Can we add doc comment explaining the utility of this function?

cns/service/main.go
```diff
@@ -932,7 +933,7 @@ func main() {
 	}
 
 	if !disableTelemetry {
-		go logger.SendHeartBeat(rootCtx, cnsconfig.TelemetrySettings.HeartBeatIntervalInMins)
+		go internal.SendHeartBeat(rootCtx, cnsconfig.TelemetrySettings.HeartBeatIntervalInMins, homeAzMonitor)
```
smittal22, line 936: Why don't we check the `cnsConfig.ChannelMode` here as well? If it's not direct, we don't even need to start SendHeartBeat go routine.

smittal22 (commented):
Left a few comments. Can sync offline as well.

cns/internal/heartbeat.go
```diff
@@ -0,0 +1,40 @@
...
+				Value:            1.0,
+				CustomDimensions: make(map[string]string),
+			}
+			getHomeAzResp := homeAzMonitor.GetHomeAz(ctx)
+			if getHomeAzResp.Response.ReturnCode == types.Success {
+				metric.CustomDimensions[logger.IsAZRSupportedStr] = strconv.FormatBool(getHomeAzResp.HomeAzResponse.IsSupported)
+				metric.CustomDimensions[logger.HomeAZStr] = strconv.FormatUint(uint64(getHomeAzResp.HomeAzResponse.HomeAz), 10)
+			}
+
+			logger.SendMetric(metric)
```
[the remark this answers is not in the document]
smittal22: Scratch that. We do need to send heartbeat to be able to know when AZR is unsupported. Would like to understand more as to how this metric is consumed though.

cns/internal/heartbeat.go
```diff
@@ -0,0 +1,40 @@
...
+	"strconv"
+	"time"
+
+	"github.com/Azure/azure-container-networking/aitelemetry"
+	"github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/Azure/azure-container-networking/cns/restserver"
+	"github.com/Azure/azure-container-networking/cns/types"
+)
+
+func SendHeartBeat(ctx context.Context, heartbeatIntervalInMins int, homeAzMonitor *restserver.HomeAzMonitor) {
```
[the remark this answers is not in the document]
ZetaoZhuang: it's not specific to AZR actually. It's a general function for collecting node id and orchestrator type also.

cns/service/main.go
ZetaoZhuang: heartbeat is not specific to Direct channel. we only add azr metric when its Direct channel.

cns/internal/heartbeat.go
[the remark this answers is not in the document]
ZetaoZhuang: when there is no cache value for azr, we still need to send metric which contains the node id and orchestrator type

cns/internal/heartbeat.go
timraymond, line 17: It would be a more flexible design if this accepted a `time.Duration`, and let the caller work out how to construct the right duration value. The fact that the configuration is defined in minutes shouldn't influence how this function operates. As it stands, this would be hard to unit test because we're forced to choose a minute value and stall the test suite for N minutes.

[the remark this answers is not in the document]
smittal22: This function was being used beforehand as well. Zetao is just moving it to a separate package to get rid of import cycles.

smittal22 (approved):
LGTM

timraymond (dismissed):

cns/internal/heartbeat.go
```diff
@@ -0,0 +1,41 @@
...
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			metric := aitelemetry.Metric{
+				Name: logger.HeartBeatMetricStr,
+				// This signifies 1 heartbeat is sent. Sum of this metric will give us number of heartbeats received
+				Value:            1.0,
+				CustomDimensions: make(map[string]string),
+			}
+			getHomeAzResp := homeAzMonitor.GetHomeAz(ctx)
```
ramiro-gamarra, line 32: if the intention is to only make this call if channel mode is direct, then the code should explicitly reflect that, rather than always making the call and assuming that `GetHomeAz` only works when channel mode is direct.

cns/internal/heartbeat.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2018 Microsoft. All rights reserved.
+// MIT License
+
+package internal
```
ramiro-gamarra, line 4: why `internal` specifically? package should reflect what the code's domain/intention is. `internal` is used for structuring packages and disallow importing, but is general in the path of a package, not the name of the package itself. please think about a more descriptive package name.

ramiro-gamarra (changes requested):
Please see comments for feedback that needs to be addressed.

cns/metric/heartbeat.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2018 Microsoft. All rights reserved.
+// MIT License
+
+package metric
```
ramiro-gamarra, line 4: go packages are usually in the plural

ramiro-gamarra (approved):

PR #2738, opened 2024-05-09
Add unblock condition for the swiftv2 tests.

csfmomo:
**Reason for Change**:
The underlying testing cluster for swiftv2 is not stable and need add condition to unblock the pipeline merge.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/multitenancy/swiftv2-e2e-job-template.yaml
+++ b/.pipelines/multitenancy/swiftv2-e2e-job-template.yaml
@@ -60,15 +60,3 @@ stages:
               name: ${{ parameters.name }}
               clusterName: ${{ parameters.clusterName }}-$(commitID)
               os: linux
-
-      # - template: ../cni/k8s-e2e/k8s-e2e-job-template.yaml
-      #   parameters:
-      #     sub: $(ACN_TEST_SERVICE_CONNECTION)
-      #     clusterName: ${{ parameters.clusterName }}-$(commitID)
-      #     os: ${{ parameters.os }}
-      #     dependsOn: ${{ parameters.name }}
-      #     datapath: true
-      #     dns: true
-      #     portforward: true
-      #     hostport: true
-      #     service: true
--- a/.pipelines/multitenancy/swiftv2-e2e-job-template.yaml
+++ b/.pipelines/multitenancy/swiftv2-e2e-job-template.yaml
@@ -10,7 +10,7 @@ parameters:
 
 stages:
   - stage: ${{ parameters.clusterName }}
-    condition: and( succeeded(), not(eq(dependencies.dualstackoverlaye2e.result,'SucceededWithIssues')) ) # Cant use parameters in dependencies
+    condition: and( succeeded(), not(eq(dependencies.setup.result,'SucceededWithIssues')) ) # Cant use parameters in dependencies
     displayName: Create Cluster - ${{ parameters.displayName }}
     dependsOn:
       - ${{ parameters.dependsOn }}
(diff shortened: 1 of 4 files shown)


.pipelines/multitenancy/swiftv2-e2e-job-template.yaml
```diff
@@ -31,7 +32,7 @@ stages:
           region: $(REGION_SWIFTV2_CLUSTER_TEST) # Swiftv2 has a specific region requirements
 
   - stage: ${{ parameters.name }}
-    # condition: and( succeeded(), not(eq(dependencies.dualstackoverlaye2e.result,'SucceededWithIssues')) ) # Cant use parameters in dependencies
+    condition: and( succeeded(), not(eq(dependencies.dualstackoverlaye2e.result,'SucceededWithIssues')) ) # Cant use parameters in dependencies
```
jpayne3506, line 35: `dependencies.dualstackoverlaye2e.result,` this should be the previous stage name.

.pipelines/multitenancy/swiftv2-e2e-job-template.yaml
```diff
@@ -10,6 +10,7 @@ parameters:
 
 stages:
   - stage: ${{ parameters.clusterName }}
+    condition: and( succeeded(), not(eq(dependencies.dualstackoverlaye2e.result,'SucceededWithIssues')) ) # Cant use parameters in dependencies
```
jpayne3506, line 13: This cluster create job needs a continue on error flag, not a conditional. See https://github.com/Azure/azure-container-networking/blob/4370e6e93c7623a70446580417df1ab0fc2b4ff9/.pipelines/templates/create-cluster.yaml#L46

jpayne3506 (changes requested):

.pipelines/multitenancy/swiftv2-e2e-job-template.yaml
```diff
@@ -28,10 +28,11 @@ stages:
           vmSize: ${{ parameters.vmSize }}
           k8sVersion: ${{ parameters.k8sVersion }}
           dependsOn: ${{ parameters.dependsOn }}
+          continueOnError: true
```
jpayne3506, line 31: This is a parameter, not a flag within `jobs.job`. If the intent is to pass flag then consume it with `${{ parameters.continueOnError }}` within the job template on the continueOnError field

test/integration/swiftv2/swiftv2_test.go
```diff
@@ -156,7 +156,7 @@ func TestSwiftv2PodToPod(t *testing.T) {
 			t.Logf("ping from pod %q to %q", pod.Name, ip)
```
jpayne3506, line 156: The point of adding the continue on error is to allow it to fail and not be blocking. Adding these changes we will not know when the E2E is functioning properly.

.pipelines/multitenancy/swiftv2-e2e-job-template.yaml
```diff
@@ -7,6 +7,7 @@ parameters:
   k8sVersion: ""
```
jpayne3506, line 7: Add continue on error to the E2E and remove the log template.

jpayne3506: If the E2E was not failing/broken, then we would not need this.

jpayne3506 (approved):

PR #2748, opened 2024-05-21
feat: moved overlay-byocni-up target to tagged ip

estebancams:
**Reason for Change**:
As part of the security sprint, we are updating all public IPs created to use IPs from Service Tags. Currently the Service Tag our team has requested is still pending, so this PR uses a temporary tag provided for testing: /TestTagging-O365.
This PR changes the standalone cluster ARM template to use a bicep file, mainly due to more readable syntax and implementation (especially when adding resource dependencies). The new bicep file will provision the AKS cluster as well as a public IP using a Service Tag, which will be used by the AKS cluster as an outbound IP. This prevents the default cluster creation from creating its own public outbound IP (which doesn't have a Service Tag)

**Issue Fixed**:
Pipeline CNI LSG calls Makefile to create cluster. Updating target

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


hack/aks/Makefile
```diff
@@ -88,21 +89,24 @@ overlay-net-up: ## Create vnet, nodenet subnets
 	$(AZCLI) network vnet create -g $(GROUP) -l $(REGION) --name $(VNET) --address-prefixes 10.0.0.0/8 -o none
 	$(AZCLI) network vnet subnet create -g $(GROUP) --vnet-name $(VNET) --name nodenet --address-prefix 10.10.0.0/16 -o none
 
+create-tagged-ip:
+	$(AZCLI) network public-ip create --name $(TAGGED_IP_PREFIX)-$(CLUSTER) --resource-group $(GROUP) --allocation-method static --ip-tags FirstPartyUsage=/TestTagging-O365 --version IPV4 --sku Standard
```
chengch801214, line 93: I think you should add `--location $(REGION)` and `--tier regional`

hack/aks/Makefile
```diff
@@ -88,21 +89,24 @@ overlay-net-up: ## Create vnet, nodenet subnets
...
-overlay-byocni-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster
+overlay-byocni-up: rg-up overlay-net-up create-tagged-ip ## Brings up an Overlay BYO CNI cluster
 	$(AZCLI) aks create -n $(CLUSTER) -g $(GROUP) -l $(REGION) \
 		--auto-upgrade-channel $(AUTOUPGRADE) \
 		--node-os-upgrade-channel $(NODEUPGRADE) \
 		--kubernetes-version $(K8S_VER) \
 		--node-count $(NODE_COUNT) \
 		--node-vm-size $(VM_SIZE) \
 		--load-balancer-sku standard \
+		--load-balancer-outbound-ip /subscriptions/$(SUB)/resourceGroups/$(GROUP)/providers/Microsoft.Network/publicIPAddresses/$(TAGGED_IP_PREFIX)-$(CLUSTER) \
```
chengch801214, line 109: From the docs I only see a parameter option of `--load-balancer-outbound-ips`, but if this works too that's fine.

chengch801214 (changes requested):

hack/aks/Makefile
estebancams: typo, thanks

estebancams:
LSG run: https://msazure.visualstudio.com/One/_build/results?buildId=95775347&view=results

PR #2751, opened 2024-05-24
feat: add mTLS to CNS

jackieluc:
**Reason for Change**:

Enable mTLS support for CNS

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


cns/service.go
```diff
@@ -190,6 +191,18 @@ func getTLSConfigFromFile(tlsSettings localtls.TlsSettings) (*tls.Config, error)
...
+		rootCAs, err := mtlsRootCAsFromCertificate(&tlsCert)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get root CAs for configuring mTLS")
+		}
+		tlsConfig.ClientAuth = tls.RequireAndVerifyClientCert
+		tlsConfig.ClientCAs = rootCAs
+		tlsConfig.RootCAs = rootCAs
+	}
+
+	log.Debugf("TLS configured successfully from file: %+v", tlsSettings)
```
ZetaoZhuang, line 204: just curious, do we have structure logging for this? what is the difference between logger.Debugf and log.Debugf there,

rbtr: CNS still mostly does not use structured logging. Here, `log` is the ACN logger, `logger` is the CNS logger...we should probably prefer `logger`?

paulyufan2:
can we add the description to the PR?

cns/service.go
paulyufan2: @rbtr  do we have plan to move logger to zap-logger for CNS?

cns/service.go
```diff
@@ -190,6 +191,18 @@ func getTLSConfigFromFile(tlsSettings localtls.TlsSettings) (*tls.Config, error)
 		},
 	}
 
+	if tlsSettings.UseMTLS {
```
paulyufan2, line 194: maybe we can have a function to check if useMTLs is enabled since line 240 is doing similar tlsConfig

cns/service.go
rbtr: there's no concrete plan for the work to happen, right now. CNS actually [already has a zap logger](https://github.com/Azure/azure-container-networking/blob/e491f20ae07df49c2962aee1d75a3170ec9dfc23/cns/service/main.go#L631-L636) that is used in a few places, but there is a much bigger migration that needs to happen where we port the existing `logger` to use a zap/zapai backend.

cns/service.go
```diff
@@ -224,9 +237,50 @@ func getTLSConfigFromKeyVault(tlsSettings localtls.TlsSettings, errChan chan<- e
...
+
 	return &tlsConfig, nil
 }
 
+// Given a TLS cert, return the root CAs
+func mtlsRootCAsFromCertificate(tlsCert *tls.Certificate) (*x509.CertPool, error) {
+	certPool := x509.NewCertPool()
+
+	// self signed, we assume this is a root CA
+	if len(tlsCert.Certificate) == 1 {
```
paulyufan2, line 261: if tlsCert length is not 1, what would be the case and should we continue to do certPool.AddCert on line 278?

cns/configuration/configuration.go
```diff
@@ -55,6 +55,7 @@ type CNSConfig struct {
 	TLSSubjectName              string
 	TelemetrySettings           TelemetrySettings
 	UseHTTPS                    bool
+	UseMTLS                     bool
```
msvik, line 58: This will need to be added in CNS wrapper in Aquarius repo also

cns/service/main.go
```diff
@@ -785,6 +785,7 @@ func main() {
 				KeyVaultCertificateName:            cnsconfig.KeyVaultSettings.CertificateName,
 				MSIResourceID:                      cnsconfig.MSISettings.ResourceID,
 				KeyVaultCertificateRefreshInterval: time.Duration(cnsconfig.KeyVaultSettings.RefreshIntervalInHrs) * time.Hour,
+				UseMTLS:                            cnsconfig.UseMTLS,
```
msvik, line 788: if tlssetting is set when UseHttps is true, then I am wondering how it is working for ACI. We need to follow up on this with ACI

jackieluc: i think you already followed up on this with ACI - resolving

cns/service.go
jackieluc: it'll break - i can add guards to make this more robust

cns/service.go
jackieluc: let me change this to use `logger`, thanks for the context around that here. do we have a way to also emit metrics? this would be useful for cases where we want to see a rollout of TLS/mTLS for specific partners/scenarios.

jackieluc: refactored to use `logger`

cns/service.go
jackieluc: i don't think this is needed as there are only two places this happens and is a niche scenario

cns/service.go
jackieluc: refactored to add guard clauses and added unit tests

cns/service.go
paulyufan2: okay thanks for letting me know

paulyufan2 (approved):

cns/configuration/configuration.go
jackieluc: created a PR for this in that repo

cns/service_test.go
```diff
@@ -0,0 +1,212 @@
...
+			svc.Uninitialize()
+		})
+		require.NoError(t, err)
+
+		err = svc.StartListener(config)
+		require.NoError(t, err)
+
+		client := &http.Client{}
+
+		req, err := http.NewRequestWithContext(context.TODO(), http.MethodGet, "http://localhost:10090", http.NoBody)
```
msvik, line 52: just curious what does GET on this URL return?

cns/service_test.go
```diff
@@ -0,0 +1,212 @@
...
+		require.NoError(t, err)
+
+		client := &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					MinVersion: tls.VersionTLS12,
+					MaxVersion: tls.VersionTLS13,
+					ServerName: config.TLSSettings.TLSSubjectName,
+					// #nosec G402 for test purposes only
+					InsecureSkipVerify: true,
```
msvik, line 91: Do we need this if we are specifying the expected ServerName above on the client side?

cns/service_test.go
```diff
@@ -0,0 +1,212 @@
...
+		req, err := http.NewRequestWithContext(context.TODO(), http.MethodGet, "http://localhost:10090", http.NoBody)
+		require.NoError(t, err)
+		resp, err := client.Do(req)
+		t.Cleanup(func() {
+			resp.Body.Close()
+		})
+		require.NoError(t, err)
+	})
+
+	t.Run("NewServiceWithTLS", func(t *testing.T) {
```
msvik, line 61: Since both TLS and mTLS configuration start both http and https listener, we should test both for completeness.

cns/service_test.go
jackieluc: it returns a `HTTP 404`. we only really care if a connection is established, ie. no errors in the request

cns/service_test.go
jackieluc: yes, we still need this because once the client is presented with the cert from the server, it also attempts to verify the cert chain. without this, we would get an error `x509: certificate signed by unknown authority` as it fails to verify the chain for the self-signed cert as it's CA isn't in the system's trusted CAs

cns/service_test.go
jackieluc: added HTTP listener tests for when TLS/mTLS are enabled

cns/service_test.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/Azure/azure-container-networking/cns/common"
+	"github.com/Azure/azure-container-networking/cns/logger"
+	acn "github.com/Azure/azure-container-networking/common"
+	serverTLS "github.com/Azure/azure-container-networking/server/tls"
+	"github.com/Azure/azure-container-networking/store"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewService(t *testing.T) {
```
msvik, line 31: It will be good to add a test case for TLSPort == "" also.

paulyufan2 (approved):

paulyufan2 (dismissed):

common/listener.go
```diff
@@ -55,6 +56,7 @@ func (l *Listener) StartTLS(errChan chan<- error, tlsConfig *tls.Config, address
 	}
 
 	l.tlsListener = list
+	l.tlsLocalAddress = address
```
rbtr, line 59: Is this only used for the log at L105? It looks redundant and I don't think we should add a struct field for it since we are logging it here at L60
> log.Printf("[Listener] Started listening on tls endpoint %s.", address)

cns/service_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+		for _, tc := range tt {
+			cert := tls.Certificate{
+				Certificate: tc.invalidCert,
+			}
+			r, err := mtlsRootCAsFromCertificate(&cert)
+			require.Error(t, err)
+			assert.Nil(t, r)
+		}
+	})
```
rbtr, line 248: These tests look good, but I think that it would be easier to read and add additional test cases later if they were table-driven instead of separate `test.Run`s

rbtr (changes requested):
Tests comment is non-blocking, but I think that redundant struct field should be removed

common/listener.go
jackieluc: yes, it is used for that. i'm thinking that if there is an absence of the log, it'd cause confusion in the logs if it were missing

jackieluc: without the struct field, perhaps this is better: `log.Printf("[Listener] Stopped listening on tls endpoint %s", l.tlsListener.Addr())` instead. i could also convert L113 to `l.listener.Addr()`.

you get log lines kinda like this when running locally:
```
2024/06/07 11:48:23 [11816] [Listener] Started listening on tls endpoint :10091.
2024/06/07 11:48:23 [11816] [Listener] Started listening on :10090.
2024/06/07 11:48:23 [11816] [Listener] Stopped listening on tls endpoint [::]:10091
2024/06/07 11:48:23 [11816] [Listener] Stopped listening on [::]:10090
```

any thoughts?

cns/service_test.go
jackieluc: added a check for `config.TLSSettings` being empty in the non-TLS case

cns/service_test.go
jackieluc: refactored

common/listener.go
rbtr: IMO pulling the data out of the listener like this is better than adding a field for it 👍

jackieluc: refactored this as well, thanks for suggestion

cns/service_test.go
rbtr: I really meant all of the new tests. But like I said, not a blocker

rbtr (approved):

cns/service_test.go
jackieluc: ah gotcha, i will keep the `Service` tests as is. if there are newer cases i'll consider refactoring them

msvik (approved):

PR #2757, opened 2024-05-30
[Do not merge] Bohui's PR: test for cns image: support aks windows (#2756)

kmurudi:
* Added NIC type and modified setRoutes function

* Modified NIC name

* Modified NIC name to align with DNC

* Update MultitenantPodNetworkConfig CRD to include NICType

* Update MultitenantPodNetworkConfig CRD to include NICType

* Added unit test

* Added unit test for NIC type

* Applied universal naming for NIC

* Updated CRD manifests

* Fix Unit Test

* Fix Lint issue: a line is over 238 characters

* Modified MTPNC

* Revert the name change for DelegatedVMNIC & BackendNIC

* Modified the getIPConfig function to adjust array of InterfaceInfos

* Set the first element from the first InterfaceInfos

* Fixed CRD

* Check for duplicate NCID

* Parse MTPNC primaryIP

* Fixed parse primaryIP and fixed duplicated ncid, added helper function

* Fix Lint issue: Wrap error

* Remove NIC Type in CRD

* Modified functions in the utils

* Fix lint by wrapping error

* Fixed unit test

* Check for device type and assign for NIC

* Modified setRoutes function and added UT

* Fixed ut in ipam

* Revert doc change in mtpnc

---------

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -3,8 +3,6 @@ package middlewares
 import (
 	"context"
 	"fmt"
-	"net/netip"
-
 	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/cns/configuration"
 	"github.com/Azure/azure-container-networking/cns/logger"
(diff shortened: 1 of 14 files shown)


cns/middlewares/k8sSwiftV2.go
```diff
@@ -135,150 +133,78 @@ func (m *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req
...
+			PodIPConfig: cns.IPSubnet{
+				IPAddress:    ip,
+				PrefixLength: uint8(prefixSize),
+			},
+			MacAddress:        mtpnc.Status.MacAddress,
+			NICType:           cns.DelegatedVMNIC,
+			SkipDefaultRoutes: false,
+			// InterfaceName is empty for DelegatedVMNIC
+		})
+	} else {
```
nddq, line 172: will we ever reach this else block for non-windows scenarios? If not, then this whole func can be implemented separately in the respective os-specific files (k8sSwiftV2_linux.go and k8sSwiftV2_windows.go)

nddq: right i got that part, but i was wondering about `getIPConfig` itself. When running in Linux, will we ever get the scenario where `len(mtpnc.Status.InterfaceInfos) != 0`?

kmurudi:
> @kmurudi why did you open this PR instead of @bohuini? Even the original #2756. And same question about #2759

this was opened to use the mainrepo branch to create a private image for cns using the ACN PR pipeline, #2759 was opened as a draft to check the changes jaeryn made to our pipeline to support the 2025 windows image. these 2 will be closed.

timraymond:
There should be a way to run these pipelines with a branch argument so we don't need throwaway do-not-review PRs, right @rbtr ?

cns/middlewares/k8sSwiftV2.go
timraymond: Can we write a test to answer these questions?

kmurudi:
> There should be a way to run these pipelines with a branch argument so we don't need throwaway do-not-review PRs, right @rbtr ?

this did not work for forked repo branches, only for the branchs off master repo

PR #2745, opened 2024-05-20
Added gRPC Server Boilerplate Code for CNS

Kevincxv:
**Reason for Change**:

Added boilerplate code for setting up a gRPC server in CNS during its startup. It includes the initial setup, configuration, and necessary gRPC server initialization code, enabling CNS to start a gRPC server for enhanced node health monitoring and operations.

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [x] includes documentation
- [ ] adds unit tests
- [x] relevant PR labels added

**Notes**:

PR #2758, opened 2024-05-30
gRPC Server Implementation

Kevincxv:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


timraymond:
@Kevincxv Is this different from #2745 ? If not, it's better to keep things in one PR to keep the discussion organized.

debecerra:
@timraymond This PR is going to cover a feature that builds on top of #2745, so #2745 isn't going anywhere. This draft PR is up for visibility on the new code changes and to avoid being blocked on the review process for the prerequisite PR

timraymond:
@debecerra Gotcha. Can we rebase it off of that branch instead of master so that way only the diffs from this work are present? That would resolve the confusion and at least make it easy to see what changed here alone.

rbtr:
> @debecerra Gotcha. Can we rebase it off of that branch instead of master so that way only the diffs from this work are present? That would resolve the confusion and at least make it easy to see what changed here alone.

+1 rebase it on the branch in that PR instead of master, and change the target branch on this PR to that branch also. Then the diff here will only be the changes added on top of that PR.

PR #2772, opened 2024-06-07
feat: New yaml file for updated CNI and CNS version

chengch801214:
**Reason for Change**:
Need an update to CNI version to resolve issue with newer kernels

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/mdnc/azure-cns-cni-1.5.28.yaml
+++ b/.pipelines/mdnc/azure-cns-cni-1.5.28.yaml
@@ -1,0 +1,207 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: azure-cns
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: kube-system
+  name: nodeNetConfigEditor
+rules:
+- apiGroups: ["acn.azure.com"]
+  resources: ["nodenetworkconfigs"]
+  verbs: ["get", "list", "watch", "patch", "update"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: pod-reader-all-namespaces
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: nodeNetConfigEditorRoleBinding
+  namespace: kube-system
+subjects:
+- kind: ServiceAccount
+  name: azure-cns
+  namespace: kube-system
+roleRef:
+  kind: Role
+  name: nodeNetConfigEditor
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: pod-reader-all-namespaces-binding
+subjects:
+- kind: ServiceAccount
+  name: azure-cns
+  namespace: kube-system
+roleRef:
+  kind: ClusterRole
+  name: pod-reader-all-namespaces
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: azure-cns
+  namespace: kube-system
+  labels:
+    app: azure-cns
+spec:
+  selector:
+    matchLabels:
+      k8s-app: azure-cns
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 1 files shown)


tamilmani1989 (dismissed):

paulyufan2 (dismissed):

.pipelines/mdnc/azure-cns-cni-1.5.28.yaml
```diff
@@ -0,0 +1,207 @@
...
+      tolerations:
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - operator: "Exists"
+          effect: NoExecute
+        - operator: "Exists"
+          effect: NoSchedule
+      initContainers:
+        - name: init-cni-dropgz
+          image: "mcr.microsoft.com/containernetworking/cni-dropgz:v0.1.4" # CNI 1.5.28
```
rbtr, line 99:
```suggestion
          image: "mcr.microsoft.com/containernetworking/azure-cni:v1.5.28"
```

.pipelines/mdnc/azure-cns-cni-1.5.28.yaml
```diff
@@ -0,0 +1,207 @@
...
+          command: ["/dropgz"]
+          args: ["deploy" , "azure-vnet", "-o", "/opt/cni/bin/azure-vnet", "azure-vnet-telemetry", "-o", "/opt/cni/bin/azure-vnet-telemetry", "azure-swift.conflist", "-o", "/etc/cni/net.d/10-azure.conflist"]
+          volumeMounts:
+            - name: cni-bin
+              mountPath: /opt/cni/bin
+            - name: cni-conflist
+              mountPath: /etc/cni/net.d
+      containers:
+        - name: cns-container
+          image: mcr.microsoft.com/containernetworking/azure-cns:v1.5.26
```
rbtr, line 110: why not bump both to 1.5.28?

rbtr (dismissed):

.pipelines/mdnc/azure-cns-cni-1.5.28.yaml
chengch801214: I believe the 1.5.28 CNS image is not available. Do the CNI and CNS versions have to match?

.pipelines/mdnc/azure-cns-cni-1.5.28.yaml
chengch801214: Is the dropgz approach no longer recommended?

chengch801214:
> Are the other manifests in this directory still in use? Is this strictly a version bump? Can we just write a kustomize template and let the consumer set these versions instead of hosting a manifest copy for every release?

Yes the others are still in use. Two of them are the ones we communicated to OpenAI to use, but our tests no longer work with them, so we created two other ones with updated CNI version for two different tests.

.pipelines/mdnc/azure-cns-cni-1.5.28.yaml
rbtr: CNI and CNS are always released together, so if there is a 1.5.28 of CNI, there is for CNS also.
They don't have to match, but if you don't have a specific reason that you are holding one back for, might as well keep them together.

.pipelines/mdnc/azure-cns-cni-1.5.28.yaml
rbtr: Correct. There is no functional difference though, this should be a drop-in replacement.

chengch801214: What about the /dropgz command? This init container has `command: ["/dropgz"]`, what would I replace that with if I use the azure-cni image directly?

chengch801214: Updated to use azure-cni:v1.5.28 directly.

.pipelines/mdnc/azure-cns-cni-1.5.28.yaml
chengch801214: Updated to use CNS 1.5.28

rohitams (approved):

thomasricci:
> Are the other manifests in this directory still in use? Is this strictly a version bump? Can we just write a kustomize template and let the consumer set these versions instead of hosting a manifest copy for every release?

yeah I think this makes sense. Otherwise we keep creating new files just to bump up versions

chengch801214:
> > Are the other manifests in this directory still in use? Is this strictly a version bump? Can we just write a kustomize template and let the consumer set these versions instead of hosting a manifest copy for every release?
> 
> yeah I think this makes sense. Otherwise we keep creating new files just to bump up versions. These files are purely consumer by our e2e pipeline and nobody else should be using them.

I'm going to merge this PR first to unblock our E2E test. Having a customizable way to input the version is useful and can be done next time we need an update.

PR #2789, opened 2024-06-14
fix: move bufio reader creation out of for loop to fix telemetry unmarshal errors

QxBytes:
If the bufio reader is created in the for loop that handles reads we get unmarshaling errors. Creating a single bufio reader outside of the for loop once the connection is accepted and then repeatedly looping over that single reader fixes this issue.

**Reason for Change**:

See above

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/telemetry/telemetrybuffer.go
+++ b/telemetry/telemetrybuffer.go
@@ -138,8 +138,8 @@ func (tb *TelemetryBuffer) StartServer() error {
 					}()
 					reader := bufio.NewReader(conn)
 					for {
-						reportStr, err := read(reader)
-						if err != nil {
+						reportStr, readErr := read(reader)
+						if readErr != nil {
 							return
 						}
 						var tmp map[string]interface{}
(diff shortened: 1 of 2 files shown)


telemetry/telemetrybuffer.go
```diff
@@ -229,8 +229,8 @@ func (tb *TelemetryBuffer) PushData(ctx context.Context) {
 }
 
 // read - read from the file descriptor
-func read(conn net.Conn) (b []byte, err error) {
-	b, err = bufio.NewReader(conn).ReadBytes(Delimiter)
+func read(reader *bufio.Reader) (b []byte, err error) {
```
rbtr, line 232: can you just inline this? i don't see a need for it to be a helper and separating it like this probably caused the issue

rbtr (approved):

PR #2798, opened 2024-06-19
fix: CNS should ignore NNCs that are being deleted

rbtr:
**Reason for Change**:

If an NNC is marked for deletion, CNS should not use it to reinitialize state or ingest any events for it.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -1201,6 +1201,9 @@ func reconcileInitialCNSState(ctx context.Context, cli nodeNetworkConfigGetter, 
 	}
 
 	logger.Printf("Retrieved NNC: %+v", nnc)
+	if !nnc.DeletionTimestamp.IsZero() {
+		return errors.New("failed to init CNS state: NNC is being deleted")
+	}
 
 	// If there are no NCs, we can't initialize our state and we should fail out.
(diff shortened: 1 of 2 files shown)


nddq (approved):
lgtm

cns/service/main.go
```diff
@@ -1201,6 +1201,9 @@ func reconcileInitialCNSState(ctx context.Context, cli nodeNetworkConfigGetter,
 	}
 
 	logger.Printf("Retrieved NNC: %+v", nnc)
+	if !nnc.DeletionTimestamp.IsZero() {
```
tyler-lloyd, line 1204: assuming there no weird kubernetes thing that sets a deletiontimestamp to `January 1, year 1, 00:00:00 UTC` just to make it non-nil to signal deletion?

rbtr: The timestamp is set on objects with finalizers to indicate that they are pending deletion to those things that own the finalizers. I don't think in practice that this is ever non-nil and zero

thatmattlong (dismissed):

nddq (approved):

PR #2815, opened 2024-06-25
ci: Change `release/v1.5` merge queue trigger

jpayne3506:
**Reason for Change**:

Enable merge queue for `release/v1.5`

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/pipeline.yaml
+++ b/.pipelines/pipeline.yaml
@@ -14,7 +14,7 @@ pr:
 trigger:
   branches:
     include:
-      - gh-readonly-queue/master/*
+      - gh-readonly-queue/release/v1.5/*
   tags:
     include:
       - "*"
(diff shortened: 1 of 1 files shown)


.pipelines/pipeline.yaml
```diff
@@ -14,7 +14,7 @@ pr:
 trigger:
   branches:
     include:
-      - gh-readonly-queue/master/*
+      - gh-readonly-queue/release/v1.5/*
```
rbtr, line 17: maybe we should preemptively go
```suggestion
      - gh-readonly-queue/release/*
```
so that we don't have to add to this every few months?

jpayne3506: Should this target master instead and then backport?

rbtr: We can take this and make that change an master also

rbtr (approved):

PR #2820, opened 2024-06-27
ci: align cilium charts with AKS

camrynl:
**Reason for Change**:
- systemd-netword-overrides init container was removed in charts for cilium 1.14 since this was fixed upstream
- adding new config value `dnsproxy-enable-transparent-mode: "false"` which is present in 1.13 and 1.14

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr (dismissed):
does it need to be backported to 1.5 CI also?

camrynl:
yes, this will need to be backported for 1.5 too

tamilmani1989 (approved):

PR #2821, opened 2024-07-01
Add public ips with service tags for LBs during cluster creation

shubham-pathak-03:
**Reason for Change**:

As per the new security requirements, all traffic for our services  must be under a service tag. Cluster creation under ACN repo creates load balancers and the associated ips use default service tags provided by Microsoft. This change creates new public ips and attaches it with LBs during cluster creation phase. 

**Requirements**:

**Notes**:


hack/aks/Makefile
```diff
@@ -88,21 +87,29 @@ overlay-net-up: ## Create vnet, nodenet subnets
...
 overlay-byocni-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster
+	$(AZCLI) network public-ip create --name "overlay-cni-public-ip" \
+		--resource-group $(GROUP) \
+		--allocation-method Static \
+		--ip-tags "FirstPartyUsage=/TestTagging-O365" \
+		--location $(REGION) \
+		--sku Standard \
+		--tier Regional \
+		--version IPv4
+	$(AZCLI) network public-ip show --resource-group $(GROUP) --name "overlay-cni-public-ip" --query "ipAddress" -o tsv
```
thomasricci, line 105: Do you need this call if you already compose the resourceId below?

hack/aks/Makefile
```diff
@@ -116,13 +123,22 @@ endif
 	@$(MAKE) set-kubeconf
 
 overlay-byocni-nokubeproxy-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster without kube-proxy
+	$(AZCLI) network public-ip create --name "overlay-cni-no-kube-public-ip" \
+		--resource-group $(GROUP) \
+		--allocation-method Static \
+		--ip-tags "FirstPartyUsage=/TestTagging-O365" \
```
thomasricci, line 129: Can you factorize this into a variable?
This gets replicated in many scenarios and we will have to change this IP at some point to use our own service tag and not this testing one.

hack/aks/Makefile
```diff
@@ -88,21 +87,29 @@ overlay-net-up: ## Create vnet, nodenet subnets
...
 
-
 ##@ AKS Clusters
 
 byocni-up: swift-byocni-up ## Alias to swift-byocni-up
 cilium-up: swift-cilium-up ## Alias to swift-cilium-up
 up: swift-up ## Alias to swift-up
 
 overlay-byocni-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster
+	$(AZCLI) network public-ip create --name "overlay-cni-public-ip" \
```
thomasricci, line 97: I would try to have variables for all these ip names (ex. "overlay-cni-public-ip") since they need to match in 2 places and they are getting repeated. Just a typo can cause regression. Ditto for all scenarios covered in this pipeline.

thomasricci (changes requested):

hack/aks/Makefile
shubham-pathak-03: This call is not needed; I've removed it. I was using it to explicitly print the ip for verification in the pipeline.

hack/aks/Makefile
shubham-pathak-03: Done

hack/aks/Makefile
shubham-pathak-03: Done

hack/aks/Makefile
```diff
@@ -116,13 +124,21 @@ endif
 	@$(MAKE) set-kubeconf
 
 overlay-byocni-nokubeproxy-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster without kube-proxy
+	$(AZCLI) network public-ip create --name $(IP_PREFIX)-$(CLUSTER) \
```
chengch801214, line 127: I would recommend making `$(IP_PREFIX)-$(CLUSTER)` a variable, so you can use it in the cluster create command as well.

hack/aks/Makefile
```diff
@@ -234,23 +299,41 @@ swift-up: rg-up swift-net-up ## Bring up a SWIFT AzCNI cluster
...
 
 swiftv2-multitenancy-cluster-up: rg-up
+	$(AZCLI) network public-ip create --name $(PUBLIC_IP) \
+		--resource-group $(GROUP) \
+		--allocation-method Static \
+		--ip-tags $(IP_TAG) \
+		--location $(REGION) \
+		--sku Standard \
+		--tier Regional \
+		--version IPv4
```
rbtr, line 309: instead of copy/pasting these 8 lines in to every cluster target, creating the public IP should be it's own target that can be reused like the `set-kubeconf` target
```make
public-ip: rg-up
	$(AZCLI) network public-ip create --name $(PUBLIC_IP) \
		--resource-group $(GROUP) \
		--allocation-method Static \
		--ip-tags $(IP_TAG) \
		--location $(REGION) \
		--sku Standard \
		--tier Regional \
		--version IPv4

...
...

swiftv2-multitenancy-cluster-up: rg-up
	@$(MAKE) public-ip // <- calls the new public ip target
		--network-plugin azure \
		--network-plugin-mode overlay \
		--kubernetes-version 1.28 \
		--nodepool-name "mtapool" \
		--node-vm-size $(VM_SIZE) \
		--node-count 2 \
		--load-balancer-outbound-ips /subscriptions/$(SUB)/resourceGroups/$(GROUP)/providers/Microsoft.Network/publicIPAddresses/$(PUBLIC_IP) \
		--nodepool-tags fastpathenabled=true \
		--no-ssh-key \
		--yes
	@$(MAKE) set-kubeconf

...
```

rbtr (changes requested):

k-routhu:
opening PR

hack/aks/Makefile
k-routhu: updated the PR to reflect this

k-routhu:
location: `westcentralus`
allowed ip range: `172.208.176.64 - 172.208.176.95`
ip address: `172.208.176.65`
[Cilium Nightly Pipeline run](https://msazure.visualstudio.com/One/_build/results?buildId=110283127&view=results)
image

k-routhu:
General Notes:
We had a couple of PRs out by Esteban and Shubham to update service tags. However, all of the PRs targeted hack/aks/Makefile, so I went ahead and made all of the updates in this branch.

Testing:
Tested changes in a couple of Cilium pipelines:
[CNI Nightly Pipeline](https://msazure.visualstudio.com/One/_build/results?buildId=110320455&view=results) - similar as above, but with all of the final changes
location: `westcentralus`
allowed ip range: `172.208.176.64 - 172.208.176.95`
ip address: `172.208.176.65`

[CNI Release Test](https://msazure.visualstudio.com/One/_build/results?buildId=110322488&view=results)
The cilium-overlay test fails at the scale test job, but this happens on the master branch for the same exact reason, so I believe that this is safe to merge, considering the other pipelines that have been run
All of the public ips created have service tags attached to them, and the ips fall within the allowed ranges for the locations they were created

[Cilium Private Test on v1.14](https://msazure.visualstudio.com/One/_build/results?buildId=110329919&view=results) as suggested by John Payne

[Cilium Private Test on v1.16](https://msazure.visualstudio.com/One/_build/results?buildId=110329944&view=results) as suggested by John Payne

[ACN PR Pipeline](https://msazure.visualstudio.com/One/_build/results?buildId=110334824&view=results)
The failing test also fails in master, so I believe this is safe to merge considering the other pipelines that have been run

rbtr:
@k-routhu This will not be merged since the author of the PR has not agreed to the CLA (and likely won't). You will need to fork off of this branch and open a new PR.

PR #2828, opened 2024-07-03
add prefix length from mtpnc to podipinfo & gatewayip

kmurudi:
**Reason for Change**:

AKS Middleware for Swiftv2 requires us to add the correct subnet-prefix length which was missing & set the gateway IP as well in CNS for pod connectivity to work.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/middlewares/k8sSwiftV2.go
```diff
@@ -182,6 +182,11 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
 			if prefixSize != prefixLength {
 				return nil, errors.Wrapf(errInvalidMTPNCPrefixLength, "mtpnc primaryIP prefix length is %d", prefixSize)
 			}
+			// Parse MTPNC SubnetAddressSpace to get the subnet prefix length
+			subnet, subnetPrefix, err := utils.ParseIPAndPrefix(interfaceInfo.PrimaryIP)
```
paulyufan2, line 186: lgtm, can we add an UT to make sure this parses the information correctly?

kmurudi: added subnetAddressSpace in mockclient to be covered by existing UTs

paulyufan2 (approved):

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -1,10 +1,114 @@
...
+				// Parse MTPNC primaryIP to get the IP address and prefix length
+				ip, prefixSize, err = utils.ParseIPAndPrefix(interfaceInfo.PrimaryIP)
+				if err != nil {
+					return nil, errors.Wrap(err, "failed to parse mtpnc primary IP and prefix")
+				}
+				if prefixSize != prefixLength {
+					return nil, errors.Wrapf(errInvalidMTPNCPrefixLength, "mtpnc primaryIP prefix length is %d", prefixSize)
+				}
+				// Parse MTPNC SubnetAddressSpace to get the subnet prefix length
+				subnet, subnetPrefix, err := utils.ParseIPAndPrefix(interfaceInfo.PrimaryIP)
```
thatmattlong, line 79: this is parsing the interfaceInfo.PrimaryIP not the SubnetAddressSpace

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -1,10 +1,114 @@
...
+				if err != nil {
```
thatmattlong, line 80: There is a lot of shared code between the linux and windows functions, with the only major difference that I see being the calculation of the subnet address space. Can we instead keep all of this in one function, and just have separate windows and linux versions of "calculateSubnetAddressSpace"? I am worried that if we change any portion of this existing function, the programmer will have to just know to change it in two places

thatmattlong (changes requested):
I think there is too much copy/pasted between the linux and windows versions. Let's keep one common function for the majority of the code, and then create windows/linux versions only for the part that is different.

cns/middlewares/k8sSwiftV2.go
```diff
@@ -156,85 +155,3 @@ func (k *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req
 	// retrieve podinfo from orchestrator context
 	return podInfo, types.Success, ""
 }
-
-// getIPConfig returns the pod's SWIFT V2 IP configuration.
-func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodInfo) ([]cns.PodIpInfo, error) {
```
shreyashastantram, line 161: Why is this section being removed from here? This does not change for windows or linux.

kmurudi: implementing this function in k8sswiftv2_linux & _windows, but will change that now & only move a small section of this function into other 2 files

cns/middlewares/k8sSwiftV2_windows.go
kmurudi: yes that's a better idea let me separate just the subnet prefix part

thatmattlong (approved):

PR #2830, opened 2024-07-03
feat: Fetch MacAddress of the primary NIC and then supply to HNS

bohuini:
**Reason for Change**:
Fetch MacAddress of the primary NIC and then supply to HNS

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


go.mod
```diff
@@ -7,8 +7,8 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.7.0
 	github.com/Azure/azure-sdk-for-go/sdk/keyvault/azsecrets v0.12.0
 	github.com/Masterminds/semver v1.5.0
-	github.com/Microsoft/go-winio v0.6.1
-	github.com/Microsoft/hcsshim v0.12.0
+	github.com/Microsoft/go-winio v0.6.2
+	github.com/Microsoft/hcsshim v0.12.1-0.20240627163216-200feabd854d
```
bohuini, line 11: The reason that `hcsshim.SetNnvManagementMacAddresses()` is not working is because the function is not in the latest release. I will have to import the latest commit to use this function. This can work in the code now but I don't think it can be merged into master before it's formal release right? @paulyufan2

cns/service/main.go
```diff
@@ -1607,3 +1623,15 @@ func PopulateCNSEndpointState(endpointStateStore store.KeyValueStore) error {
 	}
 	return nil
 }
+
+// GetPrimaryNICMacAddress fetches the MAC address of the primary NIC on the node.
+func GetPrimaryNICMacAddress(ctx context.Context) (string, error) {
```
bohuini, line 1628: @kmurudi 1. When you say "Primary NIC", does it refer to "infra NIC"? 2. Is this the correct approach to obtain the CNS's mac address?

kmurudi: 1. yes 
2. can we use any of the existing functions to get the primary nic (infra nic) from the wireserver client instead of using mtpnc? mtpnc will set the secondary nics' macaddress. 
refer getPrimaryHostInterface in https://github.com/Azure/azure-container-networking/blob/master/cns/restserver/util.go

not sure if `restserver` package can be imported here though, might create a cyclic package dependency if `main` is called anywhere in `restserver` package

cns/service/main.go
```diff
@@ -1002,6 +1005,19 @@ func main() {
 		}(privateEndpoint, infravnet, nodeID)
 	}
 
+	// Check if channel mode is SWIFT V2 and not Direct. And Supply the MAC address to the HNS
+	if cnsconfig.EnableSwiftV2 && cnsconfig.ChannelMode != cns.Direct {
```
kmurudi, line 1009: can we call this function/block somewhere where we already check for cnsconfig & channelmode = CRD/AKS?

go.mod
paulyufan2: Yes we do need this as official release instead of using private build...will check with team

paulyufan2:
please confirm skipDefaultRoutes flag from CNS; accelnet is alway being used in delegated transparent network; if we always have delegatedNIC + AccelnetNIC, we have to set skipDefaultRoutes to True for AccelnetNIC; not sure if one accelnetNIC can be appearning alone... @kmurudi

cns/service/main.go
```diff
@@ -873,6 +892,18 @@ func main() {
 			if err = httpRemoteRestService.SavePnpIDMacaddressMapping(rootCtx); err != nil {
 				logger.Errorf("Failed to fetch PnpIDMacaddress mapping: %v", err)
 			}
+			// Setting primary macaddress if VF is enabled on the nics for swiftv2
+			var wscli wscliInterface
```
thatmattlong, line 896: don't see this getting assigned any value, wouldn't this cause a nil pointer exception?

cns/service/main.go
```diff
@@ -873,6 +892,18 @@ func main() {
 			if err = httpRemoteRestService.SavePnpIDMacaddressMapping(rootCtx); err != nil {
 				logger.Errorf("Failed to fetch PnpIDMacaddress mapping: %v", err)
```
thatmattlong, line 893: are these errors fatal? Seems like we should return them instead of logging them

cns/service/main.go
```diff
@@ -873,6 +892,18 @@ func main() {
...
 			}
+			// Setting primary macaddress if VF is enabled on the nics for swiftv2
+			var wscli wscliInterface
+			// SWIFT V2 mode for accelnet, supply the MAC address to the HNS
+			macAddress, err := getPrimaryNICMACAddress(wscli)
+			if err != nil {
+				logger.Errorf("Failed to get primary NIC MAC address: %v", err)
```
thatmattlong, line 900: return error?

cns/service/main.go
```diff
@@ -873,6 +892,18 @@ func main() {
...
+			} else {
+				macAddresses := []string{macAddress}
+				if _, err := hcsshim.SetNnvManagementMacAddresses(macAddresses); err != nil {
+					logger.Errorf("Failed to set primary NIC MAC address: %v", err)
```
thatmattlong, line 904: return error?

cns/service/main.go
```diff
@@ -1612,3 +1643,28 @@ func PopulateCNSEndpointState(endpointStateStore store.KeyValueStore) error {
 	}
 	return nil
 }
+
+// getPrimaryNICMacAddress fetches the MAC address of the primary NIC on the node.
+func getPrimaryNICMACAddress(wscli wscliInterface) (string, error) {
+	res, err := wscli.GetInterfaces(context.TODO())
+	if err != nil {
+		return "", fmt.Errorf("failed to find primary interface info: %w", err)
```
thatmattlong, line 1651: let's use `errors.Wrap` instead of `fmt.Errorf`

cns/service/main_test.go
```diff
@@ -69,3 +69,64 @@ func TestSendRegisterNodeRequest_StatusAccepted(t *testing.T) {
 
 	assert.Error(t, sendRegisterNodeRequest(ctx, mockClient, httpServiceFake, nodeRegisterReq, url))
 }
+
+// // Mock for wscliInterface
```
thatmattlong, line 73: let's not check in commented-out code, let's remove all of this

dropgz/build/windows.Dockerfile
```diff
@@ -12,16 +12,16 @@ ARG OS
...
 
 FROM --platform=linux/${ARCH} mcr.microsoft.com/cbl-mariner/base/core:2.0 AS compressor
 ARG OS
 WORKDIR /dropgz
 COPY dropgz .
 COPY --from=azure-vnet /azure-container-networking/cni/azure-$OS-swift-overlay.conflist pkg/embed/fs/azure-swift-overlay.conflist
 COPY --from=azure-vnet /azure-container-networking/cni/azure-$OS-swift-overlay-dualstack.conflist pkg/embed/fs/azure-swift-overlay-dualstack.conflist
-COPY --from=azure-vnet /azure-container-networking/azure-vnet.exe pkg/embed/fs
-COPY --from=azure-vnet /azure-container-networking/azure-vnet-stateless.exe pkg/embed/fs
+# COPY --from=azure-vnet /azure-container-networking/azure-vnet.exe pkg/embed/fs <- DELETE
```
thatmattlong, line 23: reminder

kmurudi:
duplicate

PR #2842, opened 2024-07-11
feat: Add CNS API to retrieve VMUniqueID from IMDS

msvik:
**Reason for Change**:
Adding a new CNS API to retrieve vmuniqueID of the VM where CNS is running. This uses the existing imdsclient in CNS which already has a method to retrieve vmUniqueID.

This API will invoked by DNC for Swift v2 scenario for ACI/standalone to discover VMUniqueID for the node.

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -18,6 +18,7 @@ import (
 const (
 	SetOrchestratorType                      = "/network/setorchestratortype"
 	GetHomeAz                                = "/homeaz"
+	GetVMUniqueID                            = "/vmuniqueid"
 	CreateOrUpdateNetworkContainer           = "/network/createorupdatenetworkcontainer"
 	DeleteNetworkContainer                   = "/network/deletenetworkcontainer"
 	PublishNetworkContainer                  = "/network/publishnetworkcontainer"
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -18,7 +18,7 @@ import (
 const (
 	SetOrchestratorType                      = "/network/setorchestratortype"
 	GetHomeAz                                = "/homeaz"
-	GetVMUniqueID                            = "/vmuniqueid"
+	GetVMUniqueID                            = "/metadata/vmuniqueid"
 	CreateOrUpdateNetworkContainer           = "/network/createorupdatenetworkcontainer"
 	DeleteNetworkContainer                   = "/network/deletenetworkcontainer"
 	PublishNetworkContainer                  = "/network/publishnetworkcontainer"
--- a/cns/api.go
+++ b/cns/api.go
@@ -373,3 +373,8 @@ type EndpointRequest struct {
 	HostVethName  string `json:"hostVethName"`
 	InterfaceName string `json:"InterfaceName"`
 }
+
+type GetVMUniqueIDResponse struct {
+	Response   Response `json:"response"`
+	VMUniqueID string   `json:"VMUniqueID"`
+}
--- a/cns/api.go
+++ b/cns/api.go
@@ -306,8 +306,8 @@ type IpamPoolMonitorStateSnapshot struct {
 
 // Response describes generic response from CNS.
 type Response struct {
-	ReturnCode types.ResponseCode
-	Message    string
+	ReturnCode types.ResponseCode `json:"ReturnCode"`
+	Message    string             `json:"Message"`
 }
 
 // NumOfCPUCoresResponse describes num of cpu cores present on host.
--- a/cns/api.go
+++ b/cns/api.go
@@ -376,5 +376,5 @@ type EndpointRequest struct {
 
 type GetVMUniqueIDResponse struct {
 	Response   Response `json:"response"`
-	VMUniqueID string   `json:"VMUniqueID"`
+	VMUniqueID string   `json:"vmuniqueid"`
 }
--- a/cns/fakes/imdsclientfake.go
+++ b/cns/fakes/imdsclientfake.go
@@ -9,6 +9,7 @@ package fakes
 import (
 	"context"
 
+	"github.com/Azure/azure-container-networking/cns/imds"
 	"github.com/Azure/azure-container-networking/cns/wireserver"
 )
 
@@ -16,10 +17,13 @@ const (
 	// HostPrimaryIP 10.0.0.4
 	HostPrimaryIP = "10.0.0.4"
 	// HostSubnet 10.0.0.0/24
-	HostSubnet = "10.0.0.0/24"
+	HostSubnet    = "10.0.0.0/24"
+	SimulateError = "simulate-error"
 )
 
 type WireserverClientFake struct{}
+type MockIMDSCtxKey string
+type MockIMDSClient struct{}
 
 func (c *WireserverClientFake) GetInterfaces(ctx context.Context) (*wireserver.GetInterfacesResult, error) {
 	return &wireserver.GetInterfacesResult{
@@ -40,4 +44,17 @@ func (c *WireserverClientFake) GetInterfaces(ctx context.Context) (*wireserver.G
 			},
 		},
 	}, nil
+}
+
+func NewMockIMDSClient() *MockIMDSClient {
+	return &MockIMDSClient{}
+}
+
+func (m *MockIMDSClient) GetVMUniqueID(ctx context.Context) (string, error) {
+	key := MockIMDSCtxKey(SimulateError)
+	if ctx.Value(key) != nil {
+		return "", imds.ErrUnexpectedStatusCode
+	}
+
+	return "55b8499d-9b42-4f85-843f-24ff69f4a643", nil
 }
--- a/cns/fakes/imdsclientfake.go
+++ b/cns/fakes/imdsclientfake.go
@@ -17,8 +17,8 @@ const (
 	// HostPrimaryIP 10.0.0.4
 	HostPrimaryIP = "10.0.0.4"
 	// HostSubnet 10.0.0.0/24
-	HostSubnet    = "10.0.0.0/24"
-	SimulateError = "simulate-error"
+	HostSubnet                   = "10.0.0.0/24"
+	SimulateError MockIMDSCtxKey = "simulate-error"
 )
 
 type WireserverClientFake struct{}
@@ -51,8 +51,7 @@ func NewMockIMDSClient() *MockIMDSClient {
 }
 
 func (m *MockIMDSClient) GetVMUniqueID(ctx context.Context) (string, error) {
-	key := MockIMDSCtxKey(SimulateError)
-	if ctx.Value(key) != nil {
+	if ctx.Value(SimulateError) != nil {
 		return "", imds.ErrUnexpectedStatusCode
 	}
 
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -17,7 +17,6 @@ import (
 
 	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/cns/hnsclient"
-	"github.com/Azure/azure-container-networking/cns/imds"
 	"github.com/Azure/azure-container-networking/cns/logger"
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/cns/wireserver"
@@ -1532,9 +1531,9 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
 
 	switch r.Method {
 	case http.MethodGet:
-		imdsClient := getIMDSClient(r)
-		vmUniqueID, err := imdsClient.GetVMUniqueID(ctx)
+		vmUniqueID, err := service.imdsClient.GetVMUniqueID(ctx)
 		var resp cns.GetVMUniqueIDResponse
+		httpStatus := http.StatusOK
 		if err != nil {
 			resp = cns.GetVMUniqueIDResponse{
 				Response: cns.Response{
@@ -1542,6 +1541,7 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
 					Message:    errors.Wrap(err, "failed to get vmuniqueid").Error(),
 				},
 			}
+			httpStatus = http.StatusInternalServerError
 		} else {
 			resp = cns.GetVMUniqueIDResponse{
 				Response: cns.Response{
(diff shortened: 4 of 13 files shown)


cns/NetworkContainerContract.go
```diff
@@ -18,6 +18,7 @@ import (
 const (
 	SetOrchestratorType                      = "/network/setorchestratortype"
 	GetHomeAz                                = "/homeaz"
+	GetVMUniqueID                            = "/vmuniqueid"
```
thatmattlong, line 21: too late I suppose, but I wish we would have put the routes like home az under a `/metadata` path, then vm unique ID could have gone under there too

cns/restserver/api.go
```diff
@@ -1522,3 +1523,50 @@ func (service *HTTPRestService) nmAgentSupportedApisHandler(w http.ResponseWrite
...
+	default:
+		returnMessage := "[Azure CNS] Error. getVMUniqueID did not receive a GET."
+		returnCode := types.UnsupportedVerb
+		service.setResponse(w, returnCode, cns.GetHomeAzResponse{
+			Response: cns.Response{ReturnCode: returnCode, Message: returnMessage},
+		})
+	}
+}
+
+func getIMDSClient(r *http.Request) *imds.Client {
```
thatmattlong, line 1564: I think we can get rid of this function

cns/restserver/api.go
```diff
@@ -1522,3 +1523,50 @@ func (service *HTTPRestService) nmAgentSupportedApisHandler(w http.ResponseWrite
...
+
+// getVMUniqueID retrieves VMUniqueID from the IMDS
+func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] getVMUniqueID")
+	logger.Request(service.Name, "getVMUniqueID", nil)
+	ctx := r.Context()
+
+	switch r.Method {
+	case http.MethodGet:
+		imdsClient := getIMDSClient(r)
```
thatmattlong, line 1535: I think service should have a member `IMDSClient`, which is an interface defined as 

```golang
type IMDSClient interface {
    func GetVMUniqueID(ctx context.Context) (string, error)
}
````
Then, you would simply mock the IMDSClient in your unit tests by implementing a mock of your IMDSClient interface in your UT (say, `mockIMDSCli`) and setting `service.IMDSClient = &mockIMDSCli{}`. Then you don't need to pass along a port as an http header, and you treat the existing IMDSClient package like an external, third-party library (which is good for separation of concerns)

File: cns/api.go [hunk 4]
```diff
@@ -365,3 +365,8 @@ type EndpointRequest struct {
 	HostVethName  string `json:"hostVethName"`
 	InterfaceName string `json:"InterfaceName"`
 }
+
+type GetVMUniqueIDResponse struct {
+	Response   Response `json:"response"`
```
thatmattlong, line 370: Do we need a `Response` here? Can the http status code instead indicate success/failure?

thatmattlong (changes requested):

cns/restserver/restserver.go
```diff
@@ -280,6 +280,8 @@ func (service *HTTPRestService) Init(config *common.ServiceConfig) error {
 	listener.AddHandler(cns.NetworkContainersURLPath, service.getOrRefreshNetworkContainers)
 	listener.AddHandler(cns.GetHomeAz, service.getHomeAz)
 	listener.AddHandler(cns.EndpointPath, service.EndpointHandlerAPI)
+	listener.AddHandler(cns.GetVMUniqueID, service.getVMUniqueID)
```
rbtr, line 283: I don't see risk of this data, specifically, but I am generally concerned with CNS posing as an IMDS proxy. I think this is a good opportunity for us to evaluate which endpoints need to be always enabled, or should only be enabled based on config.

cns/NetworkContainerContract.go
msvik: I still made it as /metadata/vmuniqueid. At least any future metadata APIs can adhere to it.

cns/api.go
msvik: I need the error message in the response but don't really care about the application return code at least for this API. I looked at couple of other API handlers in CNS and there are two patterns for returning result
1. Always return http success and then Response in the result contains the real success/failure. 
2. Return the appropriate http code but still Response is returned and validated for success case.

I am leaning towards returning correct http code + error string. So either go with #2 or go with something like this. 
type GetVMUniqueIDResponse struct {	
	VMUniqueID string   `json:"VMUniqueID"`
    Error      string   `json:"error"`
}

cns/restserver/api.go
msvik: Thanks. I added up introducing mock IMDS client.

cns/restserver/restserver.go
msvik: This is needed for Swift v2 (for now). I don't think we have any explicit config in CNS to indicate Swift v2 mode

rbtr: Is this endpoint covered under the standalone DNC-CNS mTLS config?
My concern is that if this is for DNC, it is necessarily externally accessible. Exposing IMDS is not a good idea, and I don't want this going to AKS. 
So how can we restrict those 1P implicitly privileged endpoints?

msvik: This is not part of mTLS. For AKS, is CNS not local-host only?

msvik: Thinking little bit more about it, there are two options
1. Like Evans said, we add endpoints based on config, which means new endpoints will need a feature/scenario config in CNS and clients will be burdened with enabling it.
2. This is protected via API authentication from DNC to CNS. (which is where mTLS comes into picture). This is the ideal scenario in my mind.

I still don't understand the full nature of concern here and how it impacts AKS-SWIFT (as I am under the impression that we have a local host only endpoint for AKS-SWIFT scenario).

rbtr: > For AKS, is CNS not local-host only?

Not necessarily 🙁  That's why we have added a [separate server recently](https://github.com/Azure/azure-container-networking/pull/2650) - to move the AKS specific endpoints to a `localhost` only access. But afaik this server that you are adding paths to is still started in Swift. Maybe that's the change we should make?

rbtr: Wonder what @paulyufan2 thinks here

msvik: ok. My suggestion would be to solve this via 
1. local host only endpoints for APIs that don't need to be on the network
2. Authentication between DNC and CNS
3. Also block containers from reaching infra services on host namespace via ACLs or iptable/HNS rules etc.

That is better way to go than each individual developer worrying about adding a new API in CNS. #1 and #2 are already in progress. So this is good but it is long term and will take time.

For short term, only option I see if put it behind a config but we already have CNS APIs that reach out to NMA via wireserver. So not sure why would IMDS specifically be of concern.

msvik: Evans and I had offline conversation. After getting more information about the AKS security issue, this API is being exposed only for direct channel mode. It doesn't seem like that this API poses any security risk for AKS but I decided to expose it only for direct channel mode since that is the only scenario this API is needed in. In the fullness of time, AKS should move to local host only CNS for any risk from network APIs. Other scenarios should enable mTLS for CNS .

cns/api.go
```diff
@@ -298,8 +298,8 @@ type IpamPoolMonitorStateSnapshot struct {
 
 // Response describes generic response from CNS.
 type Response struct {
-	ReturnCode types.ResponseCode
-	Message    string
+	ReturnCode types.ResponseCode `json:"ReturnCode"`
```
thatmattlong, line 301: not blocking but just fyi just struct tags are superfluous when they are capitalized and match the field name

cns/fakes/imdsclientfake.go
```diff
@@ -9,17 +9,21 @@ package fakes
...
 	"github.com/Azure/azure-container-networking/cns/wireserver"
 )
 
 const (
 	// HostPrimaryIP 10.0.0.4
 	HostPrimaryIP = "10.0.0.4"
 	// HostSubnet 10.0.0.0/24
-	HostSubnet = "10.0.0.0/24"
+	HostSubnet    = "10.0.0.0/24"
+	SimulateError = "simulate-error"
```
thatmattlong, line 21: I would probably just type this const as
```golang
SimulateError MockIMDSCtxKey = "simulate-error"
```
and don't do the type assertion later as usage

cns/restserver/api.go
```diff
@@ -1522,3 +1522,43 @@ func (service *HTTPRestService) nmAgentSupportedApisHandler(w http.ResponseWrite
 
 	logger.Response(service.Name, nmAgentSupportedApisResponse, resp.ReturnCode, serviceErr)
 }
+
+// getVMUniqueID retrieves VMUniqueID from the IMDS
+func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Request) {
+	logger.Printf("[Azure CNS] getVMUniqueID")
```
thatmattlong, line 1528: Does this log line add any value? We are logging the request on the next line

cns/restserver/restserver.go
```diff
@@ -49,6 +49,10 @@ type wireserverProxy interface {
 	UnpublishNC(ctx context.Context, ncParams cns.NetworkContainerParameters, payload []byte) (*http.Response, error)
 }
 
+type IMDSClient interface {
```
thatmattlong, line 52: nit but this probably doesn't need to be publicly exported

cns/restserver/api.go
```diff
@@ -1522,3 +1522,43 @@ func (service *HTTPRestService) nmAgentSupportedApisHandler(w http.ResponseWrite
...
+			}
+			httpStatus = http.StatusInternalServerError
+		} else {
+			resp = cns.GetVMUniqueIDResponse{
+				Response: cns.Response{
+					ReturnCode: types.Success,
+				},
+				VMUniqueID: vmUniqueID,
+			}
+		}
```
thatmattlong, line 1552: I think style-wise, we'd prefer to handle errors right away and return early. I'd consider refactoring to something like
```golang
vmUniqueID, err := service.imdsClient.GetVMUniqueID(ctx)
if err != nil {
    resp := cns.GetVMUniqueIDResponse{
        Response: cns.Response{
            ReturnCode: types.UnexpectedError,
            Message: errors.Wrap(err, "failed to get vm unique ID").Error(),
        },
    }
    logger.Response(service.Name, resp, resp.Response.ReturnCode, err)
    respondJSON(w, http.StatusInternalServerError, resp)
    return
}

resp := cns.GetVMUniqueIDResponse{
    Response: cns.Response{
        ReturnCode: types.Success,
    },
    VMUniqueID: vmUniqueID,
}

....
```

cns/api.go
msvik: For some reason linter kept flagging it. So added this to satisfy the linter

cns/fakes/imdsclientfake.go
msvik: I have fixed this.

cns/restserver/api.go
msvik: removed it. seems redundant.

cns/restserver/restserver.go
msvik: correct. Fixed.

cns/restserver/api.go
```diff
@@ -1522,3 +1522,43 @@ func (service *HTTPRestService) nmAgentSupportedApisHandler(w http.ResponseWrite
...
+
+		respondJSON(w, httpStatus, resp)
+		logger.Response(service.Name, resp, resp.Response.ReturnCode, err)
+
+	default:
+		returnMessage := "[Azure CNS] Error. getVMUniqueID did not receive a GET."
```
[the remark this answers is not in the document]
msvik: Added the verb in the error message

cns/restserver/api.go
msvik: I have fixed it. Before idea was to remove some duplicate code.

cns/api.go
msvik: @thatmattlong  @rbtr  I think there is some inconsistency between running linter locally and then pipeline runs. Do you guys see this also?

rbtr: could be a version drift issue

cns/api.go
```diff
@@ -365,3 +365,8 @@ type EndpointRequest struct {
 	HostVethName  string `json:"hostVethName"`
 	InterfaceName string `json:"InterfaceName"`
 }
+
+type GetVMUniqueIDResponse struct {
+	Response   Response `json:"response"`
+	VMUniqueID string   `json:"VMUniqueID"`
```
thatmattlong, line 371: very nitpicky so feel free to ignore but my brain really wants consistent casing in the json field names, e.g.
```json
{
  "response": {...},
  "vmUniqueID": "..."
}
```

rbtr: we should at least be self-consistent. I feel like the convention when interacting with JSON in Go is to lowercase the fields

thatmattlong (approved):

cns/api.go
msvik: I kept it this way due to acronyms like VM or ID etc. Existing structs have it this way. But let me lowercase it all to follow the Go convention.

rbtr (approved):

cns/api.go
timraymond: @thatmattlong I still prefer to see JSON struct tags, even if they match the field name. It says "I expect to be turning this struct into JSON and I've thought about how clients will consume it."

thatmattlong (approved):

cns/api.go
thatmattlong: I suspect that matches the spirit of the linter that vipin mentioned

cns/fakes/imdsclientfake.go
```diff
@@ -41,3 +45,15 @@ func (c *WireserverClientFake) GetInterfaces(ctx context.Context) (*wireserver.G
...
 }
+
+func NewMockIMDSClient() *MockIMDSClient {
+	return &MockIMDSClient{}
+}
+
+func (m *MockIMDSClient) GetVMUniqueID(ctx context.Context) (string, error) {
+	if ctx.Value(SimulateError) != nil {
+		return "", imds.ErrUnexpectedStatusCode
+	}
```
timraymond, line 56: An easier way to do this is to have the struct be an empty husk of funcs that you can override at will in tests:

```go
type MockIMDSClient struct {
       GetVMUniqueIDF func(context.Context) (string, error)
}

func (m *MockIMDSClient) GetVMUniqueID(ctx context.Context) (string, error) {
       return m.GetVMUniqueIDF(ctx)
}

// ===
// in test:

func TestThing(t *testing.T) {
    // snip
    m := &MockIMDSClient{
              GetVMUniqueIDF: func(ctx context.Context) (string, error) {
                     return "", errors.New("boom")
              }
    }
}

msvik: I am trying to perform an end to end via API handlers. IMDSClient is mostly an internal and not really exposed to the test. But I like the approach and can use it in other places.

PR #2844, opened 2024-07-11
test: add ut for populating network and endpoint hns id in endpoint and endpoint info structs

QxBytes:
**Reason for Change**:

Modifies existing uts to confirm the following:
When we create a network, we add the hns network id to the network{} and EndpointInfo{} structs
When we create an endpoint, we add the hns endpoint id to the endpoint{} and EndpointInfo{} structs
When we create an endpoint, we add the hns network id to the endpoint{} struct from the passed in EndpointInfo{} struct

Also validates that `ContainerID` s for a particular ADD call are all the same and are not equal to the empty string and that the endpoint nic type is populated. An error **will now be returned** if the container ids of all endpoint{} structs are not the same, any of the container ids are the empty string, or the nic type field is not populated. This is only checked on add calls.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/network/endpoint.go
+++ b/network/endpoint.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Azure/azure-container-networking/netlink"
 	"github.com/Azure/azure-container-networking/network/policy"
 	"github.com/Azure/azure-container-networking/platform"
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
 )
 
@@ -385,4 +386,25 @@ func (epInfo *EndpointInfo) IsEndpointStateIncomplete() bool {
 		return true
 	}
 	return false
+}
+
+func (ep *endpoint) validateEndpoint() error {
+	if ep.ContainerID == "" || ep.NICType == "" {
+		return errors.New("endpoint struct must contain a container id and nic type")
+	}
+	return nil
+}
+
+func validateEndpoints(eps []*endpoint) error {
+	containerIDs := map[string]bool{}
+	for _, ep := range eps {
+		if err := ep.validateEndpoint(); err != nil {
+			return errors.Wrap(err, "failed to validate endpoint struct")
+		}
+		containerIDs[ep.ContainerID] = true
+	}
+	if len(containerIDs) != 1 {
+		return errors.New("multiple distinct container ids detected")
+	}
+	return nil
 }
--- a/network/endpoint.go
+++ b/network/endpoint.go
@@ -402,9 +402,10 @@ func validateEndpoints(eps []*endpoint) error {
 			return errors.Wrap(err, "failed to validate endpoint struct")
 		}
 		containerIDs[ep.ContainerID] = true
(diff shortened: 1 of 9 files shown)


network/endpoint.go
```diff
@@ -386,3 +387,25 @@ func (epInfo *EndpointInfo) IsEndpointStateIncomplete() bool {
 	}
 	return false
 }
+
+func (ep *endpoint) validateEndpoint() error {
+	if ep.ContainerID == "" || ep.NICType == "" {
+		return errors.New("endpoint struct must contain a container id and nic type")
```
paulyufan2, line 393: should be :endpoint struct must contain a container id **or** nic type

network/endpoint_test.go
```diff
@@ -374,4 +376,68 @@ var _ = Describe("Test Endpoint", func() {
...
+					},
+					{
+						ContainerID: "0ea7476f26d192f067abdc8b3df43ce3cdbe324386e1c010cb48de87eefef481",
+						NICType:     cns.DelegatedVMNIC,
+					},
+				}
+				Expect(validateEndpoints(eps)).ToNot(BeNil())
+			})
+		})
+		Context("When no container id", func() {
```
paulyufan2, line 412: let's also add a case when no container id for ib NIC because we found that issue before for both happy and unhappy path

paulyufan2 (commented):
lgtm

network/endpoint.go
```diff
@@ -386,3 +387,25 @@ func (epInfo *EndpointInfo) IsEndpointStateIncomplete() bool {
...
+	}
+	return nil
+}
+
+func validateEndpoints(eps []*endpoint) error {
```
paulyufan2, line 398: should we add some checks on updateEndpoint() side as well?

network/endpoint.go
QxBytes: I believe I'm checking that container id is _empty_ or the nic type is _empty_, and if either of those two fields are empty, the error message triggers. If there is just a container id, and nic type is empty, we will satisfy the second condition and error, so both fields must be set (that is, not the empty string) in the endpoint struct to avoid to error.

network/endpoint.go
paulyufan2: for IB, we are passing following fields:
        Id:          epInfo.MasterIfName,  (not a must)
        IfName:      epInfo.MasterIfName, (not a must)
        ContainerID: epInfo.ContainerID,
        MacAddress:  epInfo.MacAddress,
        NICType:     cns.BackendNIC,

network/endpoint.go
paulyufan2: the error message is "and"

paulyufan2 (approved):
lgtm

paulyufan2 (approved):

PR #2845, opened 2024-07-11
test: validate inputs to acn components

QxBytes:
**Reason for Change**:

Validates some fields in requests to the cns, and also validates some fields in the cni args and network config. 

**This PR modifies the behavior of CNS:**
- If the primary interface ip is not a valid ip or cidr, an error will now be raised
- Same if the gateway ip address is not an ip address

**This PR modifies the behavior of CNI:**
- The container id, ifname, and bridge name passed into cni must fully consists of zero or more alphanumeric, dots, dashes, parentheses, or underscores. Otherwise, an error will be raised

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
Please confirm there are no existing scenarios in cns where we have a primary interface name or gateway ip field that aren't ips.

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -36,6 +36,8 @@ import (
 	"go.uber.org/zap"
 )
 
+var validRegexp = regexp.MustCompile(`^[a-zA-Z0-9._\-\(\) ]*$`)
+
 const (
 	dockerNetworkOption = "com.docker.network.generic"
 	OpModeTransparent   = "transparent"
@@ -409,7 +411,8 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		return err
 	}
 
-	if err = plugin.validateArgs(args, nwCfg); err != nil {
+	if argErr := plugin.validateArgs(args, nwCfg); argErr != nil {
+		err = argErr
 		return err
 	}
 
@@ -938,7 +941,8 @@ func (plugin *NetPlugin) Get(args *cniSkel.CmdArgs) error {
 
 	logger.Info("Read network configuration", zap.Any("config", nwCfg))
 
-	if err = plugin.validateArgs(args, nwCfg); err != nil {
+	if argErr := plugin.validateArgs(args, nwCfg); argErr != nil {
+		err = argErr
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


cni/network/network.go
```diff
@@ -1451,3 +1468,21 @@ func convertCniResultToInterfaceInfo(result *cniTypesCurr.Result) network.Interf
...
+		return errors.New("invalid network config value")
+	}
+
+	return nil
+}
+
+// returns true if the string fully consists of zero or more alphanumeric, dots, dashes, parentheses, or underscores
+func isValidString(value string) bool {
+	pattern := `^[a-zA-Z0-9._\-\(\) ]*$`
+	re := regexp.MustCompile(pattern)
```
rbtr, line 1486: It's possible the Go compiler is smart enough to optimize this away, but I doubt it - you should move this to a package var so that it is not being recompiled on every request. The pattern could also be inlined instead of creating an extra intermediate var.

cns/NetworkContainerContract.go
```diff
@@ -131,9 +132,22 @@ func (req *CreateNetworkContainerRequest) Validate() error {
...
+		return errors.Wrapf(ErrInvalidIP, "GatewayIPAddress %s is not a valid ip address", req.IPConfiguration.GatewayIPAddress)
+	}
 	return nil
 }
 
+func isValidIP(ipStr string) bool {
+	if ip, _, err := net.ParseCIDR(ipStr); err == nil {
+		return ip != nil
+	}
+	return net.ParseIP(ipStr) != nil
```
rbtr, line 148: This block is a little hard to follow (because it's trying to be so compact). I think readability should be preferred here, where I need to be able to quickly grok what the valid/invalid inputs are and don't necessarily remember what `net.ParseIP` returns - is it an `error`? Is it a `net.IP` or a `*net.IP`? The answer significantly changes my interpretation of this function and its correctness.

rbtr (changes requested):

cni/network/network.go
```diff
@@ -1468,3 +1491,19 @@ func convertCniResultToInterfaceInfo(result *cniTypesCurr.Result) network.Interf
...
+		return errors.New("invalid network config value")
+	}
+
+	return nil
+}
+
+// returns true if the string fully consists of zero or more alphanumeric, dots, dashes, parentheses, or underscores
+func isValidString(value string) bool {
+	return validRegexp.MatchString(value)
+}
```
rbtr, line 1509: move the doc to the regexp declaration, rename it to something like `allowedInput` and drop this extra function def entirely

rbtr (approved):

paulyufan2 (approved):

QxBytes:
Wanted to see if I could add more cni input checks-- hold off on merging for now

QxBytes:
confirmed certain sections of code are unused and so ready to merge

PR #2850, opened 2024-07-12
fix: CNS should ignore NNCs that are being deleted (#2798)

rbtr:
**Reason for Change**:

Backport #2798
If an NNC is marked for deletion, CNS should not use it to reinitialize state or ingest any events for it.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rjdenney (approved):

nddq (approved):

rbtr:
[everything has passed except a VNET Scale test](https://msazure.visualstudio.com/One/_build/results?buildId=98380966&view=results), so I think I will bypass the queue and merge this if it fails out again...
@jpayne3506 @nairashu for viz

PR #2852, opened 2024-07-12
fix: add ipv6 port policy in addition to ipv4 policy if no host ip and ipv6 enabled

QxBytes:
**Reason for Change**:

Currently if a port mapping is specified in windows, if no host ip is specified, an ipv6 port mapping policy would not be created (only an ipv4 one). This pr makes it so that if no host ip is specified, and dualstack/ipv6 is enabled, we will create two policies, one for ipv6 and one for ipv4. If the host ip is specified, we will create only a single policy based on whether the host ip is ipv4 or ipv6.

Summary:
host port defined with host ip ipv4 (with or without ipv6 enabled) --> ipv4 port mapping policy only
host port defined with host ip ipv6 --> ipv6 port mapping policy only
host port defined without host ip (with ipv6 enabled) --> ipv4 and ipv6 port mapping policy created
host port defined without host ip (without ipv6 enabled) --> ipv4 port mapping policy created only

Updates tests to reflect the above behavior.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
Tested creating a port mapping on a windows dualstack cluster with no host ip and saw two port mapping policies (ipv4 and ipv6) created as expected. Also tested creating a pod with an ipv4 host ip and saw a single ipv4 policy created, even though the cluster was ipv6 enabled.

### DIFF
--- a/cni/network/network_windows.go
+++ b/cni/network/network_windows.go
@@ -201,7 +201,8 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) ([]
 		// To support hostport policy mapping for ipv6 in dualstack overlay mode
 		// uint32 NatFlagsIPv6 = 2
 
-		flag := hnsv2.NatFlagsLocalRoutedVip
+		// if host ip is specified, we create a policy to match that ip only (ipv4 or ipv6), or ipv4 if no host ip
+		flag := hnsv2.NatFlagsLocalRoutedVip // ipv4 flag
 		if mapping.HostIp != "" {
 			hostIP, err := netip.ParseAddr(mapping.HostIp)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


cni/network/network_windows.go
```diff
@@ -217,36 +218,55 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) ([]
...
+			logger.Info("Creating ipv6 port mapping policy", zap.Any("policy", ipv6HnsPortMappingPolicy))
+			policies = append(policies, *ipv6HnsPortMappingPolicy)
 		}
+	}
+
+	return policies, nil
+}
 
-		logger.Info("Creating port mapping policy", zap.Any("policy", hnsPolicy))
+func createPortMappingPolicy(hostPort, containerPort int, hostIP string, protocol uint32, flags hnsv2.NatFlags) (*policy.Policy, error) {
```
paulyufan2, line 244: did we figure out if two hostIPs are empty?

cni/network/network_windows.go
```diff
@@ -217,36 +218,55 @@ func getPoliciesFromRuntimeCfg(nwCfg *cni.NetworkConfig, isIPv6Enabled bool) ([]
 			}
 		}
```
tamilmani1989, line 220: can you add that tables as comment in code?

cni/network/network_windows_test.go
```diff
@@ -266,6 +286,71 @@ func TestSetPoliciesFromNwCfg(t *testing.T) {
...
+					ExternalPort: uint16(44000),
+					InternalPort: uint16(80),
+					VIP:          "2001:2002:2003::1",
+					Protocol:     policy.ProtocolTcp,
+					Flags:        hnsv2.NatFlagsIPv6,
+				},
+			},
+		},
+		{
+			name: "Runtime hostPort mapping polices with ipv4 hostIP on ipv6 enabled cluster",
```
tamilmani1989, line 300: did we cover all cases in table?

tamilmani1989 (approved):

paulyufan2 (approved):

PR #2857, opened 2024-07-17
fetch primary mac for hns accelnet

kmurudi:
**Reason for Change**:
swiftv2 for windows requires to send HNS the primary mac address on the vm where accelerated networking is enabled. we have chosen to do this on the startup of cns service as it is a one-time operation before serving requests for pod/container creation. 
This hns api will set VF (virtual functions) feature for all of the accelnet-nics on the VM (all except primary nic) & be a no-op on others if they are not accelnet-enabled nics. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/service/main.go
```diff
@@ -1612,3 +1630,28 @@ func PopulateCNSEndpointState(endpointStateStore store.KeyValueStore) error {
 	}
 	return nil
 }
+
+// getPrimaryNICMacAddress fetches the MAC address of the primary NIC on the node.
+func getPrimaryNICMACAddress(wscli wscliInterface) (string, error) {
```
paulyufan2, line 1635: let's put it to service.go

cns/service/main.go
```diff
@@ -873,6 +873,12 @@ func main() {
 			if err = httpRemoteRestService.SavePnpIDMacaddressMapping(rootCtx); err != nil {
 				logger.Errorf("Failed to fetch PnpIDMacaddress mapping: %v", err)
 			}
+
+			// Setting primary macaddress if VF is enabled on the nics for swiftv2
+			err = setVFForAccelnetNICs()
```
paulyufan2, line 878: we need to check if VF is enabled before calling this function?

kmurudi: nope it should be a no-op for non-accelnet nics

azure-ipam/go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/AzureAD/microsoft-authentication-library-for-go v1.2.2 // indirect
 	github.com/Masterminds/semver v1.5.0 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
-	github.com/Microsoft/hcsshim v0.12.3 // indirect
+	github.com/Microsoft/hcsshim v0.12.5 // indirect
```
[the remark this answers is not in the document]
kmurudi: need to test this locally if it gives no error

cns/service/main.go
```diff
@@ -875,6 +875,12 @@ func main() {
 			if err = httpRemoteRestService.SavePnpIDMacaddressMapping(rootCtx); err != nil {
 				logger.Errorf("Failed to fetch PnpIDMacaddress mapping: %v", err)
 			}
+
+			// No-op for linux, setting primary macaddress if VF is enabled on the nics for aks swiftv2
+			err = httpRemoteRestService.SetVFForAccelnetNICs()
+			if err != nil {
+				logger.Errorf("Failed to set VF for accelnet NICs: %v", err)
```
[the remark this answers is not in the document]
kmurudi: this did not block the cns running & pod creation last time when it failed, I think we can keep it for logging

cns/restserver/internalapi_windows.go
```diff
@@ -1,11 +1,55 @@
...
+	macAddresses := []string{macAddress}
+	if _, err := hcsshim.SetNnvManagementMacAddresses(macAddresses); err != nil {
+		return errors.Wrap(err, "Failed to set primary NIC MAC address")
+	}
+	return nil
+}
+
+// getPrimaryNICMacAddress fetches the MAC address of the primary NIC on the node.
+func (service *HTTPRestService) getPrimaryNICMACAddress() (string, error) {
+	res, err := service.wscli.GetInterfaces(context.TODO())
```
nddq, line 34: nit: instead of using `context.TODO()`, lets use a `context.WithTimeout()` here since we are exec-ing powershell command?

cns/service/main.go
```diff
@@ -875,6 +875,12 @@ func main() {
 			if err = httpRemoteRestService.SavePnpIDMacaddressMapping(rootCtx); err != nil {
 				logger.Errorf("Failed to fetch PnpIDMacaddress mapping: %v", err)
 			}
+
+			// No-op for linux, setting primary macaddress if VF is enabled on the nics for aks swiftv2 windows
+			logger.Printf("Setting VF for accelnet nics if enabled")
```
nddq, line 880: nit: even if this is no-op for Linux, this would still appear in the logs even though CNS is running in a Linux machine, which could be confusing for someone who doesn't have contexts. Maybe add some clarification?

nddq (approved):

PR #961, opened 2021-08-05
chore: update hcsshim for NPM

matmerr:
**Reason for Change**:

Needed for latest hcsshim update

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

PR #2864, opened 2024-07-19
ci: Remove winio from replace( )

jpayne3506:
**Reason for Change**:

hcsshim@v0.12.5 requires a winio version that has `winio.NewOpenFile`. So we need to update the winio version. Before we update hcsshim we should check that this winio update does not cause any regressions. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Loss of context for the original reasoning for replace( ) in #961 may introduce some bugs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -151,7 +151,6 @@ require (
 )
 
 replace (
-	github.com/Microsoft/go-winio => github.com/microsoft/go-winio v0.4.17
 	github.com/onsi/ginkgo => github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega => github.com/onsi/gomega v1.10.0
 )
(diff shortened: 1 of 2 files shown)


huntergregory (changes requested):
I don't think NPM tests will get triggered for this PR but if we could comment here when offline tests pass for this branch, then I could sign off for NPM

jpayne3506:
Even if the PR does not trigger the PR, we can still pair it by running the pipeline with the branch name.

huntergregory:
> /azp run NPM Conformance Tests

whoops wrong PR

paulyufan2 (approved):

huntergregory (approved):
Thanks for verifying with Windows NPM conformance

PR #2876, opened 2024-07-24
ci: [NPM] add bash directive to scale scripts

huntergregory:
**Reason for Change**:
Runs scheduled for master branch have failed since the `/bin/sh` interpreter does not support double bracket syntax:
```
./test-scale.sh: 62: [[: not found
```

Now we explicitly direct these scripts to use `/bin/bash`.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests


huntergregory:
First scale test failed since the cluster was no longer reachable

nddq (approved):

PR #2880, opened 2024-07-25
<DO NOT REVIEW> Fix condition to use default value in case parameter is not set

smittal22:
**Reason for Change**:
[PR](https://github.com/Azure/azure-container-networking/pull/1768) introduced a change where if the `MellanoxMonitorIntervalSecs` is not passed in we should use the default value, however the condition was faulty and in case of implicit default 0 value (when config value is not set), we are not defaulting to 30 secs explicit default value.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/platform/os_windows.go
+++ b/platform/os_windows.go
@@ -285,7 +285,7 @@ func hasNetworkAdapter(na adapter.NetworkAdapter) bool {
 func MonitorAndSetMellanoxRegKeyPriorityVLANTag(ctx context.Context, intervalSecs int) {
 	m := &mellanox.Mellanox{}
 	interval := defaultMellanoxMonitorInterval
-	if intervalSecs > 0 {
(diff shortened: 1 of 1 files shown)


ramiro-gamarra (changes requested):
if `inervalSecs` is 0, the default monitoring interval needs to be used

rbtr:
@smittal22 this PR shows no changes now, which doesn't seem right

smittal22:
> @smittal22 this PR shows no changes now, which doesn't seem right

@rbtr Yeah, I reverted it. Will work on it later. Might even close this out.

timraymond:
@smittal22 No need to push a revert commit on something that hasn't been merged. You can just leave it.

PR #2885, opened 2024-07-26
fix: add log for ipam data received

QxBytes:
**Reason for Change**:

Adds logs so we can tell if NMAgent response is empty/does not contain interfaces when using azure vnet ipam

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/ipam/azure.go
+++ b/ipam/azure.go
@@ -115,6 +115,8 @@ func (s *azureSource) refresh() error {
 		return err
 	}
 
+	log.Printf("[ipam] xml name received:%v interfaces:%d", doc.XMLName, len(doc.Interface))
+
 	// For each interface...
 	for _, i := range doc.Interface {
 		ifName := ""
@@ -141,6 +143,8 @@ func (s *azureSource) refresh() error {
 			log.Printf("[ipam] Failed to find interface with MAC address:%v.", i.MacAddress)
 			continue
 		}
+
+		log.Printf("[ipam] processing interface:%s IsPrimary:%t macAddress:%s", ifName, i.IsPrimary, i.MacAddress)
 
 		// For each subnet on the interface...
 		for _, s := range i.IPSubnet {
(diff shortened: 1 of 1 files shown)


rbtr (approved):

ipam/azure.go
```diff
@@ -115,6 +115,8 @@ func (s *azureSource) refresh() error {
 		return err
 	}
 
+	log.Printf("[ipam] xml name received:%v interfaces:%d", doc.XMLName, len(doc.Interface))
```
tamilmani1989, line 118: shouldn't we print whole xml output?

ipam/azure.go
```diff
@@ -142,6 +144,8 @@ func (s *azureSource) refresh() error {
 			continue
 		}
 
+		log.Printf("[ipam] processing interface:%s IsPrimary:%t macAddress:%s", ifName, i.IsPrimary, i.MacAddress)
```
tamilmani1989, line 147: log subnet too.. addresspool is created based on that

ipam/azure.go
tamilmani1989: ignore.. that will be lot if user has configured 200+ Ipconfigs..

ipam/azure.go
```diff
@@ -156,6 +168,7 @@ func (s *azureSource) refresh() error {
 				continue
 			}
 
+			log.Printf("[ipam] all ipaddresses found in ipsubnet struct: %d", len(s.IPAddress))
```
tamilmani1989, line 171:
```suggestion
			log.Printf("[ipam] Number of IPAddress found in ipsubnet %s: %d", len(s.IPAddress), s.Prefix)
```

ipam/azure.go
```diff
@@ -115,6 +118,13 @@ func (s *azureSource) refresh() error {
 		return err
 	}
 
+	log.Printf("[ipam] xml name received:%v interfaces:%d", doc.XMLName, len(doc.Interface))
+
+	if len(doc.Interface) == 0 {
```
tamilmani1989, line 123: can we log doc in this case?

QxBytes: if we log doc won't it log the entire struct with all the ip configs?

tamilmani1989 (approved):

PR #2886, opened 2024-07-29
feat: add device plugin support to CNS

thatmattlong:
**Reason for Change**:
Adds a device plugin implementation for tracking vnet and IB NICs to CNS. Does not enable execution of this plugin.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


cns/deviceplugin/plugin.go
```diff
@@ -0,0 +1,140 @@
...
+	deviceCount         int
+	kubeletSocket       string
+	deviceCheckInterval time.Duration
+}
+
+func NewPlugin(l *zap.Logger, resourceName string, socketWathcer *SocketWatcher, socket string,
+	initialDeviceCount int, kubeletSocket string, deviceCheckInterval time.Duration,
+) *Plugin {
+	return &Plugin{
+		Logger:              l.With(zap.String("resourceName", resourceName)),
```
timraymond, line 33: ❤️  making a child logger here

cns/deviceplugin/pluginmanager.go
```diff
@@ -0,0 +1,147 @@
...
+// Run runs the plugin manager until the context is cancelled or error encountered
+func (p *PluginManager) Run(ctx context.Context) error {
+	// clean up any leftover state from previous failed plugins
+	// this can happen if the process crashes before it is able to clean up after itself
+	if err := p.cleanOldState(); err != nil {
+		return errors.Wrap(err, "error cleaning state from previous plugin process")
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(defaultServerCountForDevicePlugin) // starting two goroutines to wait on
```
timraymond, line 92: I think it's easier to safely handle waitgroups if you always do `wg.Add(1)`. Here, it's no big deal, since you can audit it visually. It's as the code changes over time where it gets more dangerous.

cns/deviceplugin/pluginmanager.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error listing existing device plugin sockets")
+	}
+	for _, entry := range entries {
+		if strings.HasPrefix(entry.Name(), path.Base(getSocketPrefix(p.options.devicePluginDirectory, v1alpha1.DeviceTypeVnetNIC))) ||
+			strings.HasPrefix(entry.Name(), path.Base(getSocketPrefix(p.options.devicePluginDirectory, v1alpha1.DeviceTypeInfiniBandNIC))) {
+			// try to delete it
+			f := path.Join(p.options.devicePluginDirectory, entry.Name())
+			if err := os.Remove(f); err != nil {
+				return errors.Wrapf(err, "error removing old socket %s", f)
```
timraymond, line 131: I think implicitly quoting paths might be nicer in this circumstance:

```suggestion
				return errors.Wrapf(err, "error removing old socket %q", f)
```

cns/deviceplugin/pluginmanager_test.go
```diff
@@ -0,0 +1,283 @@
...
+	expectedIBNICs = 6
+	if err = manager.TrackDevices(v1alpha1.DeviceTypeVnetNIC, expectedVnetNICs); err != nil {
+		t.Fatalf("error tracking VNET devices: %v", err)
+	}
+
+	if err = manager.TrackDevices(v1alpha1.DeviceTypeInfiniBandNIC, expectedIBNICs); err != nil {
+		t.Fatalf("error tracking IB devices: %v", err)
+	}
+
+	<-time.After(3 * time.Second)
```
timraymond, line 76: Any way to remove this (ideally) ? Makes me feel like a source of future flakiness and I'm allergic to slow tests :)

cns/deviceplugin/pluginmanager_test.go
```diff
@@ -0,0 +1,283 @@
...
+	client := v1beta1.NewDevicePluginClient(conn)
+	resp, err := client.GetPreferredAllocation(context.Background(), &v1beta1.PreferredAllocationRequest{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "error calling GetPreferredAllocation")
+	}
+	return resp, nil
+}
+
+//nolint:staticcheck // TODO: Move to grpc.NewClient method
+func getPreStartContainerResponse(t *testing.T, pluginAddress string) (*v1beta1.PreStartContainerResponse, error) {
```
timraymond, line 266: Usually it gets weird if you're calling `t.Fatal` and returning errors as well. I would just return errors and leave it up to the caller to `t.Fatal`.

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+)
+
+const ErrorChanCapacity = 2
```
timraymond, line 16: I'd stick this next to the channel and mute the linter. I think the extra indirection makes it harder to read.

cns/deviceplugin/socketwatcher_test.go
```diff
@@ -0,0 +1,133 @@
...
+	select {
+	case <-done:
+	case <-time.After(5 * time.Second):
+		t.Fatal("socket watcher is still watching 5 seconds after file is deleted")
+	}
+}
+
+func TestWatchSocketTwice(t *testing.T) {
+	socket := "testdata/to-be-deleted.sock"
+	if _, err := os.Create(socket); err != nil {
```
timraymond, line 81: I think `testdata` is generally intended to be read-only fixtures. Nothing saying you _can't_ write in there, but usually I've seen `os.MkdirTemp` used for stuff like this so you can ensure writability.

out.txt
```diff
@@ -0,0 +1,1711 @@
+=== RUN   TestPluginManagerStartStop
```
timraymond, line 1: Should this have been committed?

aggarwal0009: removed

cns/deviceplugin/pluginmanager.go
aggarwal0009: It's the linter complaining. :(

cns/deviceplugin/pluginmanager_test.go
aggarwal0009: We want to wait enough time before the nic update is reflected. Making it smaller could actually make the test flaky.

cns/deviceplugin/server.go
aggarwal0009: updated

cns/deviceplugin/pluginmanager.go
aggarwal0009: updated

cns/deviceplugin/pluginmanager_test.go
aggarwal0009: updated

cns/deviceplugin/pluginmanager.go
timraymond: Sorry, I meant literally `wg.Add(1)` (i.e. the only acceptable argument to `wg.Add` is 1). That way you can be more sure that the goroutines that use the waitgroup are paired up with appropriate increments and decrements.

cns/deviceplugin/pluginmanager_test.go
thatmattlong: You can probably reduce the device update poller (see above `deviceplugin.PluginDeviceCheckInterval(time.Second)`) to something like 500ms and then wait on 1 second instead.

thatmattlong: Or, you could call `getDeviceCount` in a `retrier` until you get the expected result (or exceed some timeout threshold)

timraymond: This is usually because there are temporal dependencies that aren't adequately addressed. Usually this can be fixed by introducing a channel or some function to be used as a latch. Time is an illusion, and what might be an adequate sleep today might not be tomorrow. It is worth fighting to remove these.

cns/deviceplugin/pluginmanager_test.go
```diff
@@ -0,0 +1,303 @@
...
+
+	select {
+	case err := <-errChan:
+		return errors.Wrap(err, "error running fake kubelet grpc server")
+	case <-ctx.Done():
+	}
+	return nil
+}
+
+//nolint:staticcheck // TODO: Move to grpc.NewClient method
```
thatmattlong, line 179: could we move these nolint comments directly to the line we're trying to skip? having it here skips staticcheck for the whole function which we should try to avoid

cns/deviceplugin/pluginmanager_test.go
thatmattlong: I think in this case, the "black box" that we're testing is that a client can call TrackDevices, and _eventually_ `getDeviceCount` will reflect that new value. But we don't really know or make any guarantees on what eventually means. So I think retrying until some reasonable threshold is probably the most accurate way of testing the interface, I think injecting a function as a latch is kind of coupling the test and the implementation, in this case.

cns/deviceplugin/pluginmanager_test.go
aggarwal0009: moved

aggarwal0009 (dismissed):
LGTM cc @thatmattlong

cns/deviceplugin/pluginmanager_test.go
aggarwal0009: updated to 6 retry attempts with 500ms delay in between.

aggarwal0009 (dismissed):

cns/deviceplugin/pluginmanager.go
aggarwal0009: Done

cns/deviceplugin/socketwatcher_test.go
aggarwal0009: Addressed

aggarwal0009 (approved):

cns/deviceplugin/pluginmanager_test.go
timraymond: Yeah, I think that's more reasonable. At least there's a chance it could resolve sooner. Also, @thatmattlong , I must have written my comment concurrently, since I didn't see your input here when I posted.

cns/deviceplugin/testdata/.gitignore
```diff
@@ -0,0 +1 @@
+to-be-deleted.sock
```
timraymond, line 1: This can be removed now

timraymond (changes requested):
Just one quick cleanup

cns/deviceplugin/testdata/.gitignore
aggarwal0009: good catch. removed.

cns/deviceplugin/socketwatcher.go
```diff
@@ -0,0 +1,74 @@
...
+	// if a socket is already being watched, return its channel
+	if ch, ok := s.socketChans[socket]; ok {
+		return ch
+	}
+	// otherwise, start watching it and return a new channel
+	socketChan := make(chan struct{})
+	s.socketChans[socket] = socketChan
+	go func() {
+		defer close(socketChan)
+		ticker := time.NewTicker(s.options.statInterval)
```
timraymond, line 60: This leaks a goroutine. Need to `defer ticker.Stop()`.

timraymond (changes requested):
Sorry, there was one more (important) thing that I missed until now.

cns/deviceplugin/socketwatcher.go
aggarwal0009: fixed

miguelgoms (approved):

aggarwal0009 (approved):

timraymond (dismissed):

cns/deviceplugin/pluginmanager.go
```diff
@@ -0,0 +1,147 @@
...
+		o(&options)
+	}
+	return &PluginManager{
+		Logger: logger,
+		VnetNICPlugin: NewPlugin(logger, string(v1alpha1.DeviceTypeVnetNIC), socketWatcher,
+			getSocketName(options.devicePluginDirectory, v1alpha1.DeviceTypeVnetNIC), initialVnetNICCount,
+			options.kubeletSocket, options.deviceCheckInterval),
+		IBNICPlugin: NewPlugin(logger, string(v1alpha1.DeviceTypeInfiniBandNIC), socketWatcher,
+			getSocketName(options.devicePluginDirectory, v1alpha1.DeviceTypeInfiniBandNIC), initialIBNICCount,
+			options.kubeletSocket, options.deviceCheckInterval),
```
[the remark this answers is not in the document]
rbtr: it should either block on the pluginmanager being unstarted, or have a mutex on the plugins list to prevent races while allowing dynamic plugin addition

cns/deviceplugin/plugin.go
```diff
@@ -0,0 +1,143 @@
...
+	select {
+	case err := <-runErrChan:
+		return errors.Wrap(err, "error starting grpc server")
+	case err := <-readyErrChan:
+		if err != nil {
+			return errors.Wrap(err, "error waiting on grpc server to be ready")
+		}
+	case <-ctx.Done():
+		return nil
+	}
```
rbtr, line 86: as long as there is no ordering requirement, could simplify with errgroup
```suggestion
	g := errgroup.Group{}
	g.Go(func() error {
		return s.Run(childCtx)
	})
	g.Go(func() error {
		return s.Ready(childCtx)
	})
	if err := g.Wait(); err != nil {
		return errors.Wrap(err, "error running device plugin server")
	}
```

cns/deviceplugin/plugin.go
```diff
@@ -0,0 +1,143 @@
...
+	p.Logger.Info("cleaning up device plugin")
+	if err := os.Remove(p.Socket); err != nil && !os.IsNotExist(err) {
+		p.Logger.Panic("failed to remove socket", zap.Error(err))
+	}
+}
+
+func (p *Plugin) UpdateDeviceCount(count int) {
+	p.deviceCountMutex.Lock()
+	defer p.deviceCountMutex.Unlock()
+	p.deviceCount = count
```
rbtr, line 136: for a small write-only block like this, prefer to directly unlock. it's lighter and just as readable
```suggestion
	p.deviceCount = count
	p.deviceCountMutex.Unlock()
```

rbtr: Actually, do these need to be synchronized? it's primitive `int`, it has separate setter/getter methods, it is not directly accessed other than here...
I think the lock here is not necessary? WDYT @timraymond?

rbtr (changes requested):

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+	return &v1beta1.DevicePluginOptions{}, nil
+}
+
+func (s *Server) GetPreferredAllocation(context.Context, *v1beta1.PreferredAllocationRequest) (*v1beta1.PreferredAllocationResponse, error) {
+	return &v1beta1.PreferredAllocationResponse{}, nil
+}
+
+func (s *Server) PreStartContainer(context.Context, *v1beta1.PreStartContainerRequest) (*v1beta1.PreStartContainerResponse, error) {
+	return &v1beta1.PreStartContainerResponse{}, nil
+}
```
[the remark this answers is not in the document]
thatmattlong: indeed, it's the v1beta1.DevicePluginServer I believe (have to double check the type)

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+	}); err != nil {
+		return errors.Wrap(err, "error sending listAndWatch response")
+	}
+
+	// every interval, check if the current count has changed from what we've previously sent, and if so, send the new count
+	for {
+		select {
+		case <-s.listAndWatchDoneChan:
+			return nil
+		case <-time.After(s.deviceCheckInterval):
```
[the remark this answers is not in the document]
thatmattlong: choosing not to use ticker was intentional (but maybe erroneous?), my understanding of ticker was that if one operation was going slow, the next tickers will fire immediately, so suppose you had a ticker every second, but one iteration takes 30s for some reason, you'll get the next 30 ticks as fast as you can possibly process them. And it made me nervous potentially calling `stream.Send` in that kind of case, so I liked that `time.After` starts when it is called regardless of how long the previous iteration took

cns/deviceplugin/pluginmanager.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func PluginDeviceCheckInterval(i time.Duration) func(*pluginManagerOptions) {
+	return func(opts *pluginManagerOptions) {
+		opts.deviceCheckInterval = i
+	}
+}
+
+// PluginManager runs device plugins for vnet nics and ib nics
+type PluginManager struct {
```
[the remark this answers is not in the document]
thatmattlong: I weighed the pros and cons of having a more generic "plugin manager" that manages any kind of plugin (and even started coding it that way), but settled on this coupling because it keeps the management of the socket file all in one place. Not strongly tied to this approach but it felt simpler with no strong benefit to keeping it generic (probably are never going to add another device type to this)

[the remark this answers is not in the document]
thatmattlong: That said I'm ok with making this generic but there will be some refactoring involved, specifically around letting the plugins clean up any bad state that was left behind as part of `Run`

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+	listAndWatchDoneChan chan struct{}
+	deviceCheckInterval  time.Duration
+}
+
+func NewServer(address string, logger *zap.Logger, deviceCounter deviceCounter, deviceCheckInterval time.Duration) *Server {
+	return &Server{
+		address:              address,
+		logger:               logger,
+		deviceCounter:        deviceCounter,
+		listAndWatchDoneChan: make(chan struct{}),
```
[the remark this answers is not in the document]
thatmattlong: Indeed it is a shutdown channel, I've usually thought it was a code smell to stick a context in a struct, but maybe it's fine if well documented

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+func (s *Server) Run(ctx context.Context) error {
+	grpcServer := grpc.NewServer()
+	v1beta1.RegisterDevicePluginServer(grpcServer, s)
+	defer close(s.listAndWatchDoneChan)
+
+	l, err := net.Listen("unix", s.address)
+	if err != nil {
+		return errors.Wrap(err, "error listening on socket")
+	}
+	errChan := make(chan error, 2) //nolint:gomnd // disabled in favor of readability
```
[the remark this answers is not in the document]
thatmattlong: if unbuffered and context is cancelled, the `select` on L54 is broken by `ctx.Done()` before the deferred `Stop()` on L52 runs, which means there will be no listener for the `errChan` when `Serve` attempts to write to it on L50, which means that goroutine will never exit causing a mem leak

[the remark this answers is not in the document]
thatmattlong: could this be simplified to not be such a mental exercise? probably :) and looks like you proposed something below that makes this a lot easier to understand

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (s *Server) Allocate(_ context.Context, req *v1beta1.AllocateRequest) (*v1beta1.AllocateResponse, error) {
+	s.logger.Info("allocate request", zap.Any("req", *req))
+	resps := make([]*v1beta1.ContainerAllocateResponse, len(req.ContainerRequests))
+	for i, containerReq := range req.ContainerRequests {
+		resp := &v1beta1.ContainerAllocateResponse{
+			Envs: make(map[string]string),
+		}
+		for j := range containerReq.DevicesIDs {
+			resp.Envs[fmt.Sprintf("NIC_%d", j)] = containerReq.DevicesIDs[j]
```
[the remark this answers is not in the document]
thatmattlong: so we don't actually store any state here, the request comes in already asking for devices by ID, and we don't even really support that, we're just using this as a way for kubelet to count the devices, so we're returning a meaningless device ID of "nic-1", "nic-2" etc that means nothing

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+	return r, nil
+}
+
+func (s *Server) ListAndWatch(_ *v1beta1.Empty, stream v1beta1.DevicePlugin_ListAndWatchServer) error {
+	// send the initial count right away
+	advertisedCount := s.deviceCounter.getDeviceCount()
+	devices := make([]*v1beta1.Device, advertisedCount)
+	for i := range devices {
+		devices[i] = &v1beta1.Device{
+			ID:     fmt.Sprintf("NIC-%d", i),
```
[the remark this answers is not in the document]
thatmattlong: agreed for consistency and best practices but these actually don't matter and are totally made up, they are effectively meaningless but the device plugin contract requires that we return _something_

cns/deviceplugin/server.go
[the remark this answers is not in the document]
rbtr: Understand why you were worried about this, but the Go authors thought of it:
> 	// Give the channel a 1-element time buffer.
	// If the client falls behind while reading, we drop ticks
	// on the floor until the client catches up.

and 
https://cs.opensource.google/go/go/+/refs/tags/go1.22.6:src/time/tick.go;l=16-17

You won't get 30 ticks in your scenario, you will get 1 tick at the next 1 second interval from when the ticker started (which could be very soon! but will definitely be within 1 second from now.) VS the `time.After` behavior of: you will get another tick _after another one second_.
The problem with `time.After` is that if there is nothing listening, the backing timer/goroutine is leaked because the tick is never consumed. We can hit that if we enter the select and `listAndWatchDoneChan` closes and we return while we are sleeping one second for `time.After`.

cns/deviceplugin/plugin.go
thatmattlong: if removed here we probably want to move it to `TrackDevices` which is probably where it should have been all along

cns/deviceplugin/server.go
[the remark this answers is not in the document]
thatmattlong: makes sense, agreed let's use ticker

cns/deviceplugin/server.go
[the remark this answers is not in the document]
rbtr: but as soon as they exist somebody will try to use them for something 😢

cns/deviceplugin/plugin.go
timraymond: I think the lock is necessary. It's not the way it's accessed that's the issue... it's that the write could occur to a cache on one core that would not be visible to another thread on another core. It _could_ be okay in some execution histories, but not in arbitrary sequencing of events.

Also, for the defer... idk. For this trivial usage, yeah, you can drop the defer. I still kinda like it though, since it will run even in the presence of panics. Yes, this can't panic, but leaving the defer hardens it against future changes where the author was not so careful auditing for panics. I'd rather have a fatter critical section than "forget" to drop a lock because we panicked and it was recovered somewhere up the call stack (and we kept going). I feel similarly about Wait Groups and `Done()`. Really, what I'd like is something like:

```go
err := mu.WithLock(func() error {
       // some critical section
})
```

which some languages do... and we could Greenspun that feature, but it would be weird to introduce.

cns/deviceplugin/plugin.go
timraymond: errgroups can also take a context, which might help with handling of `cancel`.

cns/deviceplugin/pluginmanager.go
```diff
@@ -0,0 +1,147 @@
...
+
+	wg.Add(1) //nolint:gomnd // starting two goroutines to wait on
+	go func() {
+		defer wg.Done()
+		p.IBNICPlugin.Run(ctx)
+	}()
+
+	wg.Wait()
+	return nil
+}
```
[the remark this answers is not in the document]
timraymond: I like @rbtr 's proposal here

cns/deviceplugin/server.go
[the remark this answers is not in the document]
timraymond: Is there any way we can be actively hostile to someone trying to use them for something? Can we make them UUIDs?

cns/deviceplugin/plugin.go
rbtr: Good suggestion. I assumed `s.Run` and `s.Ready` would return when the context passed to them closes, but it would probably be more robust to pass it to the errgroup also
```go
	g, cancel := errgroup.WithContext(ctx)
	defer cancel()
	g.Go(func() error {
		return s.Run(ctx)
	})
	g.Go(func() error {
		return s.Ready(ctx)
	})
	if err := g.Wait(); err != nil {
		return errors.Wrap(err, "error running device plugin server")
	}
```

cns/deviceplugin/server.go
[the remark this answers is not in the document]
rbtr: I think it's most likely to be us, which is why I suggest generating them consistently 😅

cns/deviceplugin/server.go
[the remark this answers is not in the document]
aggarwal0009: Updated

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+		errChan <- grpcServer.Serve(l)
+	}()
+	defer grpcServer.Stop()
+
+	select {
+	case err := <-errChan:
+		return errors.Wrap(err, "error running grpc server")
+	case <-ctx.Done():
+		return nil
+	}
```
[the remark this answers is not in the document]
aggarwal0009: Updated code. In addition to what is suggested above,
1) Added defer l.close() to prevent resource leak
2) Used GracefulStop() instead of Stop()
3) Added error handing for  grpcServer.Serve(l). It is needed by linter as well

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+
+type Server struct {
+	address              string
+	logger               *zap.Logger
+	deviceCounter        deviceCounter
+	listAndWatchDoneChan chan struct{}
+	deviceCheckInterval  time.Duration
+}
+
+func NewServer(address string, logger *zap.Logger, deviceCounter deviceCounter, deviceCheckInterval time.Duration) *Server {
```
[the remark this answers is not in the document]
aggarwal0009: updated

cns/deviceplugin/socketwatcher.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func SocketWatcherStatInterval(d time.Duration) SocketWatcherOption {
+	return func(o *socketWatcherOptions) {
+		o.statInterval = d
+	}
+}
+
+type SocketWatcher struct {
+	socketChans map[string]chan struct{}
```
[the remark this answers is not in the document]
aggarwal0009: updated

cns/deviceplugin/plugin.go
```diff
@@ -0,0 +1,143 @@
...
+	deviceCheckInterval time.Duration
+}
+
+func NewPlugin(l *zap.Logger, resourceName string, socketWathcer *SocketWatcher, socket string,
+	initialDeviceCount int, kubeletSocket string, deviceCheckInterval time.Duration,
+) *Plugin {
+	return &Plugin{
+		Logger:              l.With(zap.String("resourceName", resourceName)),
+		ResourceName:        resourceName,
+		SocketWatcher:       socketWathcer,
```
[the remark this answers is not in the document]
aggarwal0009: updated

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+	case err := <-errChan:
+		return errors.Wrap(err, "error running grpc server")
+	case <-ctx.Done():
+		return nil
+	}
+}
+
+// Ready blocks until the server is ready
+func (s *Server) Ready(ctx context.Context) error {
+	c, err := grpc.DialContext(ctx, s.address, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock(), //nolint:staticcheck // TODO: Move to grpc.NewClient method
```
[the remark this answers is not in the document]
aggarwal0009: Yes and that is why I am using to nolint annotation to mute the linter with a TODO to move to non-deprecated methods .
I tried using the non-deprecated grpc.NewClient() method but could not make that to work for now.

In the interest of time, went with this. We plan to update this later.

cns/deviceplugin/server.go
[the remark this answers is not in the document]
aggarwal0009: Renamed to listAndWatchShutdownChan

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+		resps[i] = resp
+	}
+	r := &v1beta1.AllocateResponse{
+		ContainerResponses: resps,
+	}
+	return r, nil
+}
+
+func (s *Server) ListAndWatch(_ *v1beta1.Empty, stream v1beta1.DevicePlugin_ListAndWatchServer) error {
```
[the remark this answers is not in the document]
aggarwal0009: The first argument is *v1beta1.Empty to conform to the RPC method signature as defined in the gRPC service. But we are not using it in our implementation.

For the ListAndWatch RPC method, the request type is *v1beta1.Empty, which is a common pattern used when no data is needed from the client for the initial request.

cns/deviceplugin/server.go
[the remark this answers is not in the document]
aggarwal0009: Updated to use Ticker

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+		case <-time.After(s.deviceCheckInterval):
+			currentCount := s.deviceCounter.getDeviceCount()
+			if currentCount == advertisedCount {
+				continue
+			}
+			advertisedCount = currentCount
+			devices := make([]*v1beta1.Device, advertisedCount)
+			for i := range devices {
+				devices[i] = &v1beta1.Device{
+					ID:     fmt.Sprintf("vnet-nic-%d", i),
```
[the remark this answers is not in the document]
aggarwal0009: @thatmattlong Just to confirm, this does need to be vnet-nic-%d right? It can be NIC-%d?  If the device count was updated, then we can use the same device id nomenclature we used earlier, i.e. NIC-%d

cns/deviceplugin/server.go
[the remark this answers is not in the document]
aggarwal0009: Thats correct.
```
type DevicePluginServer interface {
	// GetDevicePluginOptions returns options to be communicated with Device
	// Manager
	GetDevicePluginOptions(context.Context, *Empty) (*DevicePluginOptions, error)
	// ListAndWatch returns a stream of List of Devices
	// Whenever a Device state change or a Device disappears, ListAndWatch
	// returns the new list
	ListAndWatch(*Empty, DevicePlugin_ListAndWatchServer) error
	// GetPreferredAllocation returns a preferred set of devices to allocate
	// from a list of available ones. The resulting preferred allocation is not
	// guaranteed to be the allocation ultimately performed by the
	// devicemanager. It is only designed to help the devicemanager make a more
	// informed allocation decision when possible.
	GetPreferredAllocation(context.Context, *PreferredAllocationRequest) (*PreferredAllocationResponse, error)
	// Allocate is called during container creation so that the Device
	// Plugin can run device specific operations and instruct Kubelet
	// of the steps to make the Device available in the container
	Allocate(context.Context, *AllocateRequest) (*AllocateResponse, error)
	// PreStartContainer is called, if indicated by Device Plugin during registeration phase,
	// before each container start. Device plugin can run device specific operations
	// such as resetting the device before making devices available to the container
	PreStartContainer(context.Context, *PreStartContainerRequest) (*PreStartContainerResponse, error)
}
```

cns/deviceplugin/server.go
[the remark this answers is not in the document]
aggarwal0009: Yes, we dont store anything locally. State management is done by K8s control plane. This is a dummy implementation for Allocate as we are not really using this functionality. 
We should make it consistent though, i.e. using NIC-%d at all places. 
Updated that.

cns/deviceplugin/server.go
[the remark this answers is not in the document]
aggarwal0009: I think there is some confusion here. Adding to what Matt was trying to explain earlier.
1) The device Ids are not stored locally in CNS. But rather they are managed / stored by K8s. Also they will not be used by anyone other than kubelet/k8s. 
2) We just need to return some unique device-Id for each nic resource. So here we are returning NIC-%d. Given the fact that we can only have a few nics per node, I don't think we really need to use UUID here. @timraymond 

Also, if we use UUIDs then, when/if the device count changes, we are sending the updated deviceIDs to the kubelet. Since we are not storing the device state / id locally, if we use UUID, we can't regenerate the deviceID / UUID. I think using the device count as deviceId suffix also helps us to generate them deterministically without storing them locally.  

3) Agreed that we should make them consistent, i.e. NIC-%d at all places. I can move the prefix to a constant.

cns/deviceplugin/server.go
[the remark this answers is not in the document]
thatmattlong: yep doesn't really matter, just need to have the `-%d` in there so we're consistently naming 3 nics as "nic-0, nic-1, nic-2" etc

cns/deviceplugin/server.go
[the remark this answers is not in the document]
rbtr: I don't think we should be adding new code that using deprecated features

cns/deviceplugin/server.go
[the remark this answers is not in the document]
rbtr: > I've usually thought it was a code smell to stick a context in a struct

Sticking the whole context in, kinda yes. But I think here we can capture its Done chan and use that without needing to create and close our own separate channel, since need to be able to shut down the GRPC stream (which has an embedded context) in the ListAndWatch loop later

cns/deviceplugin/pluginmanager.go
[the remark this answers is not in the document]
miguelgoms: this is still a significant refactoring of plugin-manager, for the sake of making this generic before we have a second plugin/scenario to validate that design. 

I don't think this needs to be blocking the PR to add device-plugin functionality at this point - we can keep this simple, and consider the refactoring if/when we have a concrete reason to extend this.

cns/deviceplugin/server.go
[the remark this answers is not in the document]
aggarwal0009: Updated to use const devicePrefix

cns/deviceplugin/plugin.go
aggarwal0009: > for a small write-only block like this, prefer to directly unlock. it's lighter and just as readable

Updated

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,157 @@
...
+		return errors.Wrap(err, "error sending listAndWatch response")
+	}
+
+	// every interval, check if the current count has changed from what we've previously sent, and if so, send the new count
+	ticker := time.NewTicker(s.deviceCheckInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-s.listAndWatchShutdownChan:
```
rbtr, line 123: v1beta1.DevicePlugin_ListAndWatchServer contains a `context`, so we should probably select on the stream closing also:
`case <-stream.Context().Done():`

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,157 @@
...
+
+type deviceCounter interface {
+	getDeviceCount() int
+}
+
+type Server struct {
+	address                  string
+	logger                   *zap.Logger
+	deviceCounter            deviceCounter
+	listAndWatchShutdownChan chan struct{}
```
rbtr, line 26: I think we should just call it `shutdownCh` since the point is to communicate to shut down the server

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,157 @@
...
+		listAndWatchShutdownChan: make(chan struct{}),
+		deviceCheckInterval:      deviceCheckInterval,
+	}
+}
+
+// Run starts the grpc server and blocks until an error or context is cancelled. Wait on Ready to know when the server is ready.
+func (s *Server) Run(ctx context.Context) error {
+	grpcServer := grpc.NewServer()
+	v1beta1.RegisterDevicePluginServer(grpcServer, s)
+	defer close(s.listAndWatchShutdownChan)
```
rbtr, line 44: instead of closing the shutdownChan ourselves, we could use the context that we get here to simplify further:
```suggestion
func (s *Server) Run(ctx context.Context) error {
	grpcServer := grpc.NewServer()
	v1beta1.RegisterDevicePluginServer(grpcServer, s)
	
	childCtx, cancel := context.WithCancel(ctx)
	defer cancel()
	s.shutdownCh = childCtx.Done() // assign the ctx Done signal to the shutdown ch so we don't have to close anything explicitly
```

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,157 @@
...
+
+	go func() {
+		<-ctx.Done()
+		grpcServer.GracefulStop()
+	}()
+
+	if err := grpcServer.Serve(l); err != nil && errors.Is(err, grpc.ErrServerStopped) {
+		return errors.Wrap(err, "error running grpc server")
+	}
+	return nil
```
rbtr, line 60: do we only care about returning the error if it is `ErrServerStopped`? Seems like we would want to return any error from `Serve`?

cns/deviceplugin/server.go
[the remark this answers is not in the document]
aggarwal0009: I did the POC with these methods. Would need to POC again with new methods. I tried making new method work but couldnt make it work. Is this blocking?

cns/deviceplugin/server.go
aggarwal0009: updated

cns/deviceplugin/server.go
aggarwal0009: It was a typo, I meant to do below. corrected the code now
```
	if err := grpcServer.Serve(l); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
		return errors.Wrap(err, "error running grpc server")
	}
```
 I believe grpc.ErrServerStopped indicates the server has stopped normally, typically in response to a call to GracefulStop(). This is a part of the normal operation and shutdown sequence for gRPC servers, so it's not considered an error.

cns/deviceplugin/server.go
aggarwal0009: Hmm. Updated

cns/deviceplugin/server.go
aggarwal0009: hmm. updated as below in select:
```
case <-stream.Context().Done():
            // Handle client disconnection or cancellation
            return stream.Context().Err()
```

cns/deviceplugin/plugin.go
aggarwal0009: Hmm. This does simply the code.  A correction to code above. This is the signature of errgroup.WithContext(ctx)
`func errgroup.WithContext(ctx context.Context) (*errgroup.Group, context.Context)
` 
Also seems like, we don’t explicitly call cancel or use defer cancel() because errgroup handles context cancellation internally. The errgroup will cancel its context if any of the goroutines return an error, which will propagate the cancellation to all other goroutines and the context used in the goroutines

Updated code accordingly.

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return errors.Wrap(err, "error dialing local grpc server")
+	}
+	if err := c.Close(); err != nil {
+		return errors.Wrap(err, "error closing connection to local grpc server")
+	}
+	return nil
+}
+
+func (s *Server) Allocate(_ context.Context, req *v1beta1.AllocateRequest) (*v1beta1.AllocateResponse, error) {
```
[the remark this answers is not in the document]
kmurudi: is the argument still there (with _) to conform to this method definition elsewhere?

cns/deviceplugin/plugin.go
thatmattlong: @aggarwal0009 is correct, `Run` we want to run "forever" (AKA until the context is cancelled), but `Ready` should return a nil error as soon as the grpc server is up and ready to listen to connections. This is a requirement of the k8s device plugin, you can only advertise yourself as a plugin once you're ready to accept grpc connections, otherwise kubelet will reject your registration. It's not very obvious and can be made more clear probably, but the original code returns `nil` on the `readyErrChan` when it is ready, breaking out of the `select`

cns/deviceplugin/server.go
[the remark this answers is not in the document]
thatmattlong: yes we have to implement `v1beta1.DevicePluginServer` to pass to the grpc register service call, so these method signatures are all defined in that interface

cns/deviceplugin/server.go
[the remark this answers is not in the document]
timraymond: @aggarwal0009 Can you comment to that effect for future readers of the code?

cns/deviceplugin/plugin.go
rbtr: hm, that wasn't clear at all in the code so that's still a problem

aggarwal0009: Nothing that some documentation won't fix. I can add appropriate documentation to make this clear.

cns/deviceplugin/pluginmanager.go
```diff
@@ -0,0 +1,147 @@
...
+	return func(opts *pluginManagerOptions) {
+		opts.deviceCheckInterval = i
+	}
+}
+
+// PluginManager runs device plugins for vnet nics and ib nics
+type PluginManager struct {
+	Logger        *zap.Logger
+	VnetNICPlugin *Plugin
+	IBNICPlugin   *Plugin
```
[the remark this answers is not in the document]
aggarwal0009: Updated to make pluginManager generic with list of plugins.

cns/deviceplugin/server.go
```diff
@@ -0,0 +1,162 @@
...
+	shutdownCh          <-chan struct{}
+	deviceCheckInterval time.Duration
+}
+
+func NewServer(logger *zap.Logger, address string, deviceCounter deviceCounter, deviceCheckInterval time.Duration) *Server {
+	return &Server{
+		address:             address,
+		logger:              logger,
+		deviceCounter:       deviceCounter,
+		shutdownCh:          make(chan struct{}),
```
rbtr, line 35: don't make and assign a chan here since we will reassign the shutdownCh with the Run ctx

rbtr: hm this may create an ordering dependency that `Run` **must** be called before `ListAndWatch` so that the channel is not nil 🤔
we might need an additional `started` flag that prevents other methods on the server from being called before it is `Ready`

cns/deviceplugin/pluginmanager.go
```diff
@@ -0,0 +1,114 @@
...
+	return nil
+}
+
+func (pm *PluginManager) TrackDevices(deviceType v1alpha1.DeviceType, count int) error {
+	pm.mu.Lock()
+	defer pm.mu.Unlock()
+	for _, plugin := range pm.plugins {
+		if plugin.deviceType == deviceType {
+			plugin.UpdateDeviceCount(count)
+			break
```
rbtr, line 110: i don't think this should break, otherwise it looks good

rbtr: method probably doesn't need to return `error`

cns/deviceplugin/server.go
aggarwal0009: Run will be called starting device plugin gRPC server
Once Ready, device plugin will register with kubelet.
Only once registration is complete with Kubelet, Kubelet will call ListAndWatch and Allocate() etc.
So, I think its fine

cns/deviceplugin/plugin.go
aggarwal0009: Added some documentation. Hopefully that makes it clear.

cns/deviceplugin/server.go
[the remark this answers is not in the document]
aggarwal0009: @timraymond Added comment/documentation

cns/deviceplugin/pluginmanager.go
```diff
@@ -0,0 +1,147 @@
...
+			strings.HasPrefix(entry.Name(), path.Base(getSocketPrefix(p.options.devicePluginDirectory, v1alpha1.DeviceTypeInfiniBandNIC))) {
+			// try to delete it
+			f := path.Join(p.options.devicePluginDirectory, entry.Name())
+			if err := os.Remove(f); err != nil {
+				return errors.Wrapf(err, "error removing old socket %q", f)
+			}
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
aggarwal0009: updated

cns/deviceplugin/pluginmanager.go
```diff
@@ -0,0 +1,147 @@
...
+	case v1alpha1.DeviceTypeInfiniBandNIC:
+		p.IBNICPlugin.UpdateDeviceCount(count)
+	case v1alpha1.DeviceTypeVnetNIC:
+		p.VnetNICPlugin.UpdateDeviceCount(count)
+	default:
+		return ErrUnknownDeviceType
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
aggarwal0009: updated

cns/deviceplugin/pluginmanager.go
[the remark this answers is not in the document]
aggarwal0009: updated

cns/deviceplugin/pluginmanager.go
rbtr: okay that's fine then

cns/deviceplugin/server.go
aggarwal0009: removed the initial assignment.

cns/deviceplugin/server.go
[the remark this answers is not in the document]
rbtr: > // DialContext calls NewClient and then exits idle mode.  If WithBlock(true) is
> // used, it calls Connect and WaitForStateChange until either the context
> // expires or the state of the ClientConn is Ready
> // Deprecated: use NewClient instead.  Will be supported throughout 1.x.

It seems like DialContext calls NewClient (which should be used instead here). It should be pretty trivial to make it work, since we would just do what it DialContext is doing internally.
But since they've documented it as supported for 1.x, this is not a hard blocker.

rbtr (dismissed):
overall this looks solid. the flow in plugin.go (esp. `run()`) still feels a little choppy, but that doesn't need to be a today problem. 
all of my concerns have been addressed 🚀

aggarwal0009 (approved):

rbtr (approved):

miguelgoms (approved):

miguelgoms (approved):

rbtr (approved):

rbtr:
@paulyufan2 this is blocked by #2931

miguelgoms (approved):

miguelgoms (approved):

paulyufan2 (approved):

Issue #2818, opened 2024-06-26
Race when creating azure-vnet lock directory with permissions


QxBytes:
**What happened:** 
On the first boot, no CNI binary is on the node, and so k8s creates the /var/run/azure-vnet directory with 0755 permissions automatically because it is a mount part of the azure-cns daemonset. Then the CNI is deployed.
The /var/run directory is not preserved between reboots.
Then, when the VM reboots, the CNI binary may run before k8s creates the /var/run/azure-vnet directory. When the CNI binary runs first, it creates the directory with 0644 permissions. This causes permission denied errors for the cns. Even if k8s creates/mounts the /var/run/azure-vnet directory later, it will see it already exists and won't recreate the directory with the 0755 permissions.

**What you expected to happen:**  
The CNI binary should create the directory with 0755 permissions.

**How to reproduce it:**  
Reboot the VM with the cns capabilities security context dropping all capabilities (so it doesn't bypass permission checks). There is a chance that the azure-cns pod will get stuck in crash loop backoff.

**Orchestrator and Version (e.g. Kubernetes, Docker):**  

**Operating System (Linux/Windows):**  

**Kernel (e.g. `uanme -a` for Linux or `$(Get-ItemProperty -Path "C:\windows\system32\hal.dll").VersionInfo.FileVersion` for Windows):**

**Anything else we need to know?:**
[Miscellaneous information that will assist in solving the issue.]

rbtr:
+ @jpayne3506 can you update the CNS specs that we use in CI with 
```yaml
securityContext:
  capabilities:
    drop:
      - ALL
    add:
      - NET_ADMIN # only necessary for delegated IPAM/Cilium
      - NET_RAW # only necessary for delegated IPAM/Cilium
```
and make sure that we have a test that's rebooting the Node and verifying CNS functionality afterwards

rbtr:
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods sounds like it would let the k8s chmod the directory when the CNS Pod mounts it as a Volume?

QxBytes:
Node Rebooting Test added in: #2901

jpayne3506:
Backport PR can be created for `release/v1.5`, but the validate package within `release/v1.4` needs to be updated to consume the work. Will leave this open until both have been merged

rbtr:
@QxBytes looks like this was fully backported, is it good to resolve?

[...]

PR #2902, opened 2024-08-07
ci: Update CNS daemonset capabilities within E2E

jpayne3506:
**Reason for Change**:

Updates CNS daemonset capabilities to ensure that a race condition does not occur when creating the `/var/run/azure-vnet` directory

**Issue Fixed**:

Contributes to CI/CD to cover #2818

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/integration/manifests/cns/daemonset-linux.yaml
+++ b/test/integration/manifests/cns/daemonset-linux.yaml
@@ -46,8 +46,11 @@ spec:
           args: [ "-c", "tcp://$(CNSIpAddress):$(CNSPort)", "-t", "$(CNSLogTarget)"]
           securityContext:
             capabilities:
+              drop:
+                - ALL
               add:
-                - NET_ADMIN
+                - NET_ADMIN # only necessary for delegated IPAM/Cilium
+                - NET_RAW # only necessary for delegated IPAM/Cilium
           volumeMounts:
             - name: log
               mountPath: /var/log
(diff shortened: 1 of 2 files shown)


timraymond (approved):
This looks fine to me, but I'm a little unclear as to how dropping capabilities prevents a race condition on creating a directory. Can you explain in a comment before you merge?

rbtr:
It is not _preventing_ a race, it is testing that we no longer have a race between the CNI and kubelet creating the dirs with incompatible permissions. CNI used to make it with 0644, but kubelet would make it with 0755 - so if CNI wins, CNS _should not_ have access to the dir.
...but it does, because it has [`CAP_DAC_OVERRIDE`](https://www.man7.org/linux/man-pages/man7/capabilities.7.html) by default. Dropping caps exposes the race and ensures we have resolved it by [changing the CNI Mkdir perms](https://github.com/Azure/azure-container-networking/pull/2819).
This was a fun one to debug 🥲

rbtr (approved):

rbtr:
fixes #2818

timraymond:
@rbtr Awesome, that makes sense to me :)

jpayne3506:
Thanks for the explanation @rbtr, was going to link the issue and ask for you to comment more if needed 😆 .

Issue #2648, opened 2024-03-15
Remove CNM Code


QxBytes:
**What happened:** 
CNM Code is deprecated and will be removed soon once migrations away from it complete.

**What you expected to happen:**  

**How to reproduce it:**  

**Orchestrator and Version (e.g. Kubernetes, Docker):**  

**Operating System (Linux/Windows):**  

**Kernel (e.g. `uanme -a` for Linux or `$(Get-ItemProperty -Path "C:\windows\system32\hal.dll").VersionInfo.FileVersion` for Windows):**

**Anything else we need to know?:**

rbtr:
@QxBytes please don't let issues like this be closed for stale

PR #2906, opened 2024-08-08
chore: remove cnm code

QxBytes:
**Reason for Change**:

Removes CNM code, related docs, makefile targets, and related components. CNI/CNS behavior should not change as we are not of aware of anyone using the cnm.

Removes ipamclient in the cns folder and related handlers:
- release ip
- reserve ip
- ip address utilization
- get unhealthy ips

Confirmed the cni's cns invoker queries uses functions release and reserve ip, but the actual underlying cns path queried is release/reserve ip **configs**. Reference searched for the release/reserve/ip address utilization/get unhealthy ips paths and cannot find anything

**Issue Fixed**:

#2648 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/api.go
```diff
@@ -22,11 +22,7 @@ const (
 	DeleteNetworkPath             = "/network/delete"
```
tamilmani1989, line 22: #should we remove `createnetworkpath` and `deletenetworkpath` too?

tamilmani1989 (commented):
lets make sure we remove all docs related to cnm

cns/api.go
QxBytes: If you know that they are not being used, we can delete them, but they are not used by CNM so they haven't been removed.

rbtr (approved):

tamilmani1989 (approved):

PR #2916, opened 2024-08-13
fix: add placeholder os version value for linux docker build

QxBytes:
**Reason for Change**:

If in the pipelines a linux agent cannot find buildah (for some reason), it falls back to docker. Docker expects non-empty build args. In linux, OS_VERSION is not used, so we don't populate it, but docker needs _some_ value for it, or there will be an error. So, if the value is not set, we put a placeholder value so there is no error in linux. In windows, we DOCKER_OS_VERSION just becomes the value of OS_VERSION like normal. The ARG OS_VERSION=none is not necessary but takes care of some of the dockerfile warnings.

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,7 +245,8 @@ BUILDX_ACTION  		?= --load
 CONTAINER_BUILDER   ?= buildah
 CONTAINER_RUNTIME   ?= podman
 CONTAINER_TRANSPORT ?= skopeo
-
+# docker wants a value even if it is not used
+DOCKER_OS_VERSION   ?= $(if $(OS_VERSION),$(OS_VERSION),"none")
 
 # prefer buildah, if available, but fall back to docker if that binary is not in the path or on Windows.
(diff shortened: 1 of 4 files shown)


rbtr:
I am concerned that _sometimes_ the pipelines use docker, and sometimes buildah. Pipeline runs should be reproducible and this inconsistency is not acceptable. I would rather explicitly set the CONTAINER_BUILDER=buildah and have the pipeline fail if it is not present, than have it silently use a different builder.

timraymond:
Agree w/ @rbtr. I would go so far as to argue for `alias docker="bash -c 'exit 1'"`

rbtr:
For some context, I set the Makefile up to fall back to Docker so that it would work _locally_ and wouldn't require everyone to install Buildah for development. The CI was always supposed to use the [Containers](https://github.com/containers/) tools.

PR #2923, opened 2024-08-15
Add UTs to configure secondary interfaces add Result for windows swiftv2 

paulyufan2:
**Reason for Change**:

This PR is to add UTs to configure secondary interfaces add Result for windows swiftv2 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/network/invoker_cns_test.go
QxBytes, line 1: Can we add a test for when mac address is invalid and/or pnp id is invalid?

cni/network/network.go
```diff
@@ -685,7 +685,10 @@ func (plugin *NetPlugin) findMasterInterface(opt *createEpInfoOpt) string {
 		return plugin.findMasterInterfaceBySubnet(opt.ipamAddConfig.nwCfg, &opt.ifInfo.HostSubnetPrefix)
 	case cns.NodeNetworkInterfaceFrontendNIC, cns.NodeNetworkInterfaceAccelnetFrontendNIC:
 		return plugin.findInterfaceByMAC(opt.ifInfo.MacAddress.String())
-	case cns.BackendNIC: // TODO: how to find interface with IB NIC by mac address
+	case cns.BackendNIC:
+		// if windows swiftv2 has right network drivers, there will be an NDIS interface while the VFs are mounted
+		// when the VF is dismounted, this interface will go away
+		// return an unique interface name to containerd
 		opt.ifInfo.Name = ibInterfacePrefix + strconv.Itoa(opt.endpointIndex)
```
QxBytes, line 692: I think I missed this change in another PR but I don't think we should be modifying if info in this function. It's only meant to _find_ the master interface name-- there shouldn't be any side effects (like modifying it opt.ifInfo.Name) or we might be left wondering how finding the master interface has modified some value.

cni/network/network_windows_test.go
```diff
@@ -732,6 +732,24 @@ func TestGetNetworkNameSwiftv2FromCNS(t *testing.T) {
...
+				tb:          &telemetry.TelemetryBuffer{},
+			},
+			netNs: "azure",
+			nwCfg: &cni.NetworkConfig{
+				CNIVersion:   "0.3.0",
+				MultiTenancy: false,
+			},
+			interfaceInfo: &network.InterfaceInfo{},
+			want:          parsedMacAddress,
+			wantErr:       true, // should return empty network name
```
QxBytes, line 751: After running this test it doesn't seem to pass on my computer locally? The network name is empty because I believe it goes through !nwCfg.Multitenancy path when nwCfg Name is an empty string but doesn't return an error.

cni/network/network_windows_test.go
```diff
@@ -729,8 +732,51 @@ func TestGetNetworkNameSwiftv2FromCNS(t *testing.T) {
...
+			},
+			netNs: "azure",
+			nwCfg: &cni.NetworkConfig{
+				CNIVersion:   "0.3.0",
+				MultiTenancy: false,
+			},
+			interfaceInfo: &network.InterfaceInfo{}, // return empty network name with empty interfaceInfo
+			want:          parsedMacAddress,
+			wantErr:       false,
+			wantCmpErr:    true,
```
QxBytes, line 756: Instead of asserting the two values are not equal, is it possible to just change want into what we do expect?

cni/network/network.go
QxBytes: Doesn't this: `opt.ifInfo.Name = ibInterfacePrefix + strconv.Itoa(opt.endpointIndex)` modify opt.ifInfo.Name? If you did not intend to modify `opt.ifInfo.Name` maybe in a future PR we could just return `ibInterfacePrefix + strconv.Itoa(opt.endpointIndex)` directly without storing it in an intermediate variable.

cni/network/invoker_cns_test.go
QxBytes: Okay unhappy path pnp/bad pnp cns will be coming in another pr

QxBytes (dismissed):

cni/network/network.go
```diff
@@ -685,9 +685,11 @@ func (plugin *NetPlugin) findMasterInterface(opt *createEpInfoOpt) string {
...
 	case cns.NodeNetworkInterfaceFrontendNIC, cns.NodeNetworkInterfaceAccelnetFrontendNIC:
 		return plugin.findInterfaceByMAC(opt.ifInfo.MacAddress.String())
-	case cns.BackendNIC: // TODO: how to find interface with IB NIC by mac address
-		opt.ifInfo.Name = ibInterfacePrefix + strconv.Itoa(opt.endpointIndex)
-		return opt.ifInfo.Name
+	case cns.BackendNIC:
+		// if windows swiftv2 has right network drivers, there will be an NDIS interface while the VFs are mounted
+		// when the VF is dismounted, this interface will go away
+		// return an unique interface name to containerd
+		return ibInterfacePrefix + strconv.Itoa(opt.endpointIndex)
```
tamilmani1989, line 692: will it break if we didnt set `opt.ifinfo.Name`

paulyufan2: it will not break because:
findMasterInterface() will return masterIfName and masterIfName is directly used in endpointInfo struct:
endpointInfo := network.EndpointInfo{
		NetworkID:                     opt.networkID,
		Mode:                          opt.ipamAddConfig.nwCfg.Mode,
		MasterIfName:                  **masterIfName,**  => this interface is returned from findMasterInterface
so we do not need to set opt.ifinfo.Name

tamilmani1989: no.. i think ifInfo is a pointer and passed to different places

tamilmani1989: `ifInfo.Name` not passed to endpointinfo. so should be good.

tamilmani1989 (approved):

cni/network/network.go
QxBytes: I believe opt.ifinfo.Name isn't used in _windows_ -- Currently, ibnic (where we perform this change) is only used in windows (Paul can confirm). After a reference search on windows for interface name, I only see usages for ifInfo name in endpoint.go's pretty string function in windows. In linux ifName is used in secondary endpoint client. Even on linux, it seems like the interface name field is just being set after getting it based on a system call, rather than from epInfo/opt.ifinfo (and then is never used).

QxBytes (approved):

tamilmani1989 (approved):

PR #2925, opened 2024-08-16
ci: run some cni windows uts in the pipeline

QxBytes:
**Reason for Change**:

Windows CNI tests are not currently run in the pipeline.

**Issue Fixed**:

- Moves unit tests to separate template file
- Modifies multitenancy mock to take in as a parameter the responses it would receive from CNS for greater control
- Moves the preset cnsResponse values in multitenancy_mock.go to the testing file and allows anyone to mix and match the preset cnsResponse values
- Removes usage of runtime.GOOS for the multitenancy test, instead moving the dualnic windows behavior to a windows-specific test file for both add and delete
- Existing tests shouldn't be changed-- just moved
- Removes second/consecutive add tests as they no longer apply

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


.pipelines/pipeline.yaml
```diff
@@ -49,98 +49,12 @@ stages:
...
-          steps:
-            - script: |
-                cd npm/
-                go test ./...
-              retryCountOnTaskFailure: 3
-              name: "TestWindows"
-              displayName: "Run Windows Tests"
+  - template: templates/run-unit-tests.yaml
+    parameters:
+      poolName: "$(BUILD_POOL_NAME_DEFAULT)"
```
jpayne3506, line 55: Any reason for passing this in?

network/ovs_networkclient_linux_test.go
```diff
@@ -28,27 +28,6 @@ func TestAddRoutes(t *testing.T) {
 	}
 }
 
-func TestCreateBridge(t *testing.T) {
```
jpayne3506, line 31: This change should be moved into a separate PR. Would like to backport this PR for additional windows coverage on other `release/*` branches.

jpayne3506 (changes requested):
Seeing that you made the UT's a template. Are you going to be putting it into other CI/CD?

.pipelines/pipeline.yaml
QxBytes: isn't the pool name a parameter used in the template when we specify the job? Previously the pool name parameter was passed into the "test" job and so wanted to keep similar functionality.

jpayne3506: It does not have to be. 

In the original code it was not passing a parameter but a pipeline variable. That same variable, as long as it is called within the same pipeline will match. The only reason we would want to add a parameter is if we want to make the UT file dynamic on where it can be used, on which pool, ect.. (which there is no need for right now).

jpayne3506 (approved):

PR #2928, opened 2024-08-16
fix: dualnic options to avoid overwritting vlanMaps

paulyufan2:
**Reason for Change**:

This PR is to fix dualnic options to avoid overwritting vlanMaps
In the windows dualnic multitenancy case, we receive two interface infos from cns and should create two corresponding networks and endpoints. The first and second endpoints created by hns should have vlan id 1 and vlan id 2, respectively. It is CNI's job to parse the cns input and send a request to hns to create the endpoints. 

However, during testing both endpoints have vlan id 2. This is because the ipamAddConfig's `options` field is unintentionally **shared** between each `EndpointInfo` struct. Modifying the vlan id of one options field in an endpoint info struct inadvertently modifies every other endpoint info struct's options. So, the vlan id reflected during endpoint creation is of the last endpoint info's vlan id only, as all endpoint info option fields point to the same underlying object in memory.

The proposed solution is to, for each endpoint info we create, we shallow copy the `ipamAddConfig`'s options field. So, every endpoint info gets its own copy of the options field which it can modify and store without modifying the other endpoint infos.

If the options field of type map[string]interface{} ever holds anything other than primitives before we perform the shallow copy (ex: if we store pointers to another object as a value in the map) the copied pointer will still be pointing to the same object and we will have the same issue, but as of now we have only identified a single instance where we write to the options field in `invoker_cns.go` (`ncPrimaryIP`) where the value is of type string, which is immutable.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
To be tested by partner team(s) before being merged.

### DIFF
--- a/cni/network/invoker.go
+++ b/cni/network/invoker.go
@@ -31,6 +31,15 @@ type IPAMAddResult struct {
 	ipv6Enabled bool
 }
 
+func (ipamAddConfig IPAMAddConfig) getOptions() *map[string]interface{} {
+	res := map[string]interface{}{}
+	for k, v := range ipamAddConfig.options {
+		// only support permittive type
+		res[k] = v
+	}
+	return &res
+}
+
 func (ipamAddResult IPAMAddResult) PrettyString() string {
 	pStr := "InterfaceInfo: "
 	for key := range ipamAddResult.interfaceInfo {
--- a/cni/network/invoker.go
+++ b/cni/network/invoker.go
@@ -31,15 +31,6 @@ type IPAMAddResult struct {
 	ipv6Enabled bool
 }
 
-func (ipamAddConfig IPAMAddConfig) getOptions() *map[string]interface{} {
-	res := map[string]interface{}{}
-	for k, v := range ipamAddConfig.options {
-		// only support permittive type
-		res[k] = v
-	}
-	return &res
-}
-
 func (ipamAddResult IPAMAddResult) PrettyString() string {
 	pStr := "InterfaceInfo: "
 	for key := range ipamAddResult.interfaceInfo {
@@ -47,4 +38,14 @@ func (ipamAddResult IPAMAddResult) PrettyString() string {
 		pStr += val.PrettyString()
 	}
 	return pStr
+}
+
+// shallow copy options from one map to a new options map
+func (ipamAddConfig IPAMAddConfig) shallowCopyIpamAddConfigOptions() map[string]interface{} {
+	res := map[string]interface{}{}
+	for k, v := range ipamAddConfig.options {
+		// only support primitive type
+		res[k] = v
+	}
+	return res
 }
--- a/cni/network/invoker_cns_test.go
+++ b/cni/network/invoker_cns_test.go
@@ -2112,19 +2112,30 @@ func TestShallowCopyIpamAddConfigOptions(t *testing.T) {
 	require.Equal(t, opts.options, res)
 
 	// modified copied res and make sure original opts is not changed
-	res[network.SNATIPKey] = "100"
-	res[dockerNetworkOption] = "200"
-	require.NotEqual(t, opts.options, res)
+	newSnapIPKeyValue := "100"
+	newDockerNetworkOptionValue := "200"
+
+	res[network.SNATIPKey] = newSnapIPKeyValue
+	res[dockerNetworkOption] = newDockerNetworkOptionValue
+
+	expectedOpts := map[string]interface{}{
+		network.SNATIPKey:   newSnapIPKeyValue,
+		dockerNetworkOption: newDockerNetworkOptionValue,
+		"intType":           10,
+		"floatType":         0.51,
+		"byteType":          byte('A'),
+	}
+	require.Equal(t, expectedOpts, res)
 
 	// make sure original object is equal to expected opts after copied res is changed
-	expectedOpts := map[string]interface{}{
+	expectedOriginalOpts := map[string]interface{}{
 		network.SNATIPKey:   "10",
 		dockerNetworkOption: "20",
 		"intType":           10,
 		"floatType":         0.51,
 		"byteType":          byte('A'),
 	}
-	require.Equal(t, opts.options, expectedOpts)
+	require.Equal(t, expectedOriginalOpts, opts.options)
 
 	// shallow copy empty opts and make sure it does not break anything
 	emptyOpts := IPAMAddConfig{
--- a/cni/network/invoker_cns_test.go
+++ b/cni/network/invoker_cns_test.go
@@ -2097,7 +2097,7 @@ func TestCNSIPAMInvoker_Add_SwiftV2(t *testing.T) {
 
 func TestShallowCopyIpamAddConfigOptions(t *testing.T) {
 	opts := IPAMAddConfig{
-		// mock all optios' fields
+		// mock different types of map value
 		options: map[string]interface{}{
 			network.SNATIPKey:   "10",
 			dockerNetworkOption: "20",
--- a/cni/network/invoker_cns_test.go
+++ b/cni/network/invoker_cns_test.go
@@ -2108,20 +2108,20 @@ func TestShallowCopyIpamAddConfigOptions(t *testing.T) {
 	}
 
 	// shallow copy all ipamAddConfig options
-	res1 := opts.shallowCopyIpamAddConfigOptions()
-	require.Equal(t, opts.options, res1)
+	res := opts.shallowCopyIpamAddConfigOptions()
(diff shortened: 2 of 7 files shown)


cni/network/invoker.go
```diff
@@ -31,6 +31,15 @@ type IPAMAddResult struct {
 	ipv6Enabled bool
 }
 
+func (ipamAddConfig IPAMAddConfig) getOptions() *map[string]interface{} {
```
QxBytes, line 34: since this is a function that is a part of IPAMAddConfig can we rename this to like `getShallowCopy` or `shallowCopy` or something like that

QxBytes: also I think map[string]interface{} is already a reference to the underlying map so you don't need the * indicating a pointer. The value of res itself can be thought of as an address (not the actual data).

cni/network/network.go
```diff
@@ -772,7 +773,7 @@ func (plugin *NetPlugin) createEpInfo(opt *createEpInfoOpt) (*network.EndpointIn
 		BridgeName:                    opt.ipamAddConfig.nwCfg.Bridge,
 		NetworkPolicies:               networkPolicies, // nw and ep policies separated to avoid possible conflicts
 		NetNs:                         opt.ipamAddConfig.args.Netns,
-		Options:                       opt.ipamAddConfig.options,
+		Options:                       *option,
```
QxBytes, line 776: I think you can make the getOptions call in-line

cni/network/network.go
```diff
@@ -435,6 +435,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		// previously we had a default interface info to select which interface info was the one to be returned from cni add
 		cniResult := &cniTypesCurr.Result{}
 		for key := range ipamAddResult.interfaceInfo {
+			logger.Info("Exiting add, interface info retrieved", zap.Any("ifInfo", ipamAddResult.interfaceInfo[key]))
```
QxBytes, line 438: Can we move this into its own dedicated loop-- currently if we find the infra nic we don't print out the rest of the interfaces

cni/network/network_linux.go
```diff
@@ -40,6 +41,8 @@ func setNetworkOptions(cnsNwConfig *cns.GetNetworkContainerResponse, nwInfo *net
 		vlanMap := make(map[string]interface{})
 		vlanMap[network.VlanIDKey] = strconv.Itoa(cnsNwConfig.MultiTenancyInfo.ID)
 		vlanMap[network.SnatBridgeIPKey] = cnsNwConfig.LocalIPConfiguration.GatewayIPAddress + "/" + strconv.Itoa(int(cnsNwConfig.LocalIPConfiguration.IPSubnet.PrefixLength))
+		logger.Info("Add %s and %s with %d to", zap.String("vlanIDKey", network.VlanIDKey), zap.String("SnatBridgeIPKey", network.SnatBridgeIPKey),
```
QxBytes, line 44: Since this is using zap, I don't think the %s and %d will be replaced by the values you specified. Zap will print the literal value `Add %s and %s with %d to` with the "%".

cni/network/network_windows.go
```diff
@@ -43,6 +43,7 @@ func setNetworkOptions(cnsNwConfig *cns.GetNetworkContainerResponse, nwInfo *net
 		logger.Info("Setting Network Options")
 		vlanMap := make(map[string]interface{})
 		vlanMap[network.VlanIDKey] = strconv.Itoa(cnsNwConfig.MultiTenancyInfo.ID)
+		logger.Info("Add %s with %d to", zap.String("vlanIDKey", network.VlanIDKey), zap.Int("vlanID", cnsNwConfig.MultiTenancyInfo.ID), zap.Any("vlanMap", vlanMap))
```
QxBytes, line 46: Similar to above

QxBytes (changes requested):
Is it possible to add the vlan id to ep info pretty print (endpoint.go) or similar pretty string (should not panic/err if no vlan id is present/map is nil)?

paulyufan2:
> Is it possible to add the vlan id to ep info pretty print (endpoint.go) or similar pretty string (should not panic/err if no vlan id is present/map is nil)?

we have already printed vlanID when setting options (key and value)

cni/network/invoker_cns_test.go
```diff
@@ -2108,6 +2108,20 @@ func TestShallowCopyIpamAddConfigOptions(t *testing.T) {
...
-	require.Equal(t, opts.options, res)
+	res1 := opts.shallowCopyIpamAddConfigOptions()
+	require.Equal(t, opts.options, res1)
+
+	// create a second opts for reference representing expected outputs
+	copyRes := res1
+	// modify res1 fields and make sure expectedOpts has not changed
+	res1[network.SNATIPKey] = "100"
+	res1[dockerNetworkOption] = "200"
+	require.Equal(t, copyRes, res1)
```
QxBytes: Based on the variable name, is copyRes intended to be a copy of res1's values? What is expectedOpts? If copyRes should not be modified, why are you checking that the two are equal? Shouldn't it be that copyRes should maintain the initial values if it were copied into a seprate map?

cni/network/invoker_cns_test.go
```diff
@@ -2108,6 +2108,20 @@ func TestShallowCopyIpamAddConfigOptions(t *testing.T) {
...
+
+	// shallow copy empty opts and make sure it does not break anything
+	emptyOpts := IPAMAddConfig{
+		options: map[string]interface{}{},
+	}
+	res3 := emptyOpts.shallowCopyIpamAddConfigOptions()
+	require.Equal(t, emptyOpts.options, res3)
```
QxBytes: could we also ensure nothing breaks if options is nil?

cni/network/invoker_cns_test.go
```diff
@@ -2094,3 +2094,34 @@ func TestCNSIPAMInvoker_Add_SwiftV2(t *testing.T) {
...
+		},
+	}
+
+	// shallow copy all ipamAddConfig options
+	res := opts.shallowCopyIpamAddConfigOptions()
+	require.Equal(t, opts.options, res)
+
+	// create a second opts for reference representing expected outputs
+	copyRes := res
+	// modify res1 fields and make sure copied res has changed
```
QxBytes, line 2116: This doesn't copy res. I believe it just copies the pointer to the underlying map. While this is valid, we want to test that after a shallow copy and modification of the copy, the original _has not_ changed.

cni/network/network.go
```diff
@@ -608,6 +608,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 	endpointIndex := 1
 	for key := range ipamAddResult.interfaceInfo {
 		ifInfo := ipamAddResult.interfaceInfo[key]
+		logger.Info("retrieving interfaceInfo:", zap.Any("ifInfo", ifInfo))
```
QxBytes, line 611: nit: could we change this to like "Processing interfaceInfo" or similar (we've already retrieved the interface info from the ipam)

QxBytes (changes requested):

cni/network/invoker_cns_test.go
QxBytes: we should test that after the shallow copy and modifying the shallow copy, the original has not changed

QxBytes (dismissed):
Ensure tested with partner teams if you haven't done so already

QxBytes (dismissed):

cni/network/network_linux.go
```diff
@@ -40,6 +41,8 @@ func setNetworkOptions(cnsNwConfig *cns.GetNetworkContainerResponse, nwInfo *net
 		vlanMap := make(map[string]interface{})
 		vlanMap[network.VlanIDKey] = strconv.Itoa(cnsNwConfig.MultiTenancyInfo.ID)
 		vlanMap[network.SnatBridgeIPKey] = cnsNwConfig.LocalIPConfiguration.GatewayIPAddress + "/" + strconv.Itoa(int(cnsNwConfig.LocalIPConfiguration.IPSubnet.PrefixLength))
+		logger.Info("Add vlanIDkey and SnatBridgeIPKey with vlanID to vlanMap", zap.String("vlanIDKey", network.VlanIDKey), zap.String("SnatBridgeIPKey", network.SnatBridgeIPKey),
```
tamilmani1989, line 44: log line is confusing.. also lets update `vlanMap` name to `optionsMap` and just log vlan key and snatbridge key

tamilmani1989 (changes requested):

tamilmani1989 (approved):

PR #2933, opened 2024-08-20
change apipa network gw address from .1 to .2

ZetaoZhuang:
**Reason for Change**:

currently cns using the same ip address "169.254.128.1" for both apipa gw and loopback adapter. This cause conflict issue when hns get restarted and not able to rehydrate the apipa endpoints. our long-term goal is to make this gw address configurable in dnc. But for a short-term fix, in cns, we want to overwrite the address to 169.254.128.2, so customer can recreate apipa endpoints by restarting the cns.
**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


rbtr:
what's the impact of this change? is it backwards compatible?

ZetaoZhuang:
here are some screenshots from testing the changes on my set up.
image

**vm ipconfig**
image

**container ipconfig**
image

rbtr:
thanks for adding this context. what happens if this change goes out to a node where containers are already running with the old .1 address?

ZetaoZhuang:
> thanks for adding this context. what happens if this change goes out to a node where containers are already running with the old .1 address?

This change only applies on new created apipa network/endpoints. It does not impact existed containers unless recreate them. @ashvindeodhar

tamilmani1989:
> thanks for adding this context. what happens if this change goes out to a node where containers are already running with the old .1 address?

@ZetaoZhuang did we test this with existing clusters with old cns and new cni, with all combinations?

cns/hnsclient/hnsclient_windows.go
```diff
@@ -600,6 +603,15 @@ func CreateHostNCApipaEndpoint(
 	return endpoint.Id, nil
 }
 
+// adhocAdjustIPConfig applies adhoc change on gw IP address
+func adhocAdjustIPConfig(localIPConfiguration *cns.IPConfiguration) {
```
QxBytes, line 607: Instead of adding this, is it possible to change the localIPConfiguration we pass into CreateHostNCApipaEndpoint? Is this a temporary solution? I feel like it might be a bit confusing if we pass in a local ip configuration and then the code changes what we directed cns to do to something else.

QxBytes (commented):
A couple clarifications:
- how do the apipa gw and loopback adapter come up the first time with no restart if they both use 169.254.128.1?
- is the long term fix that we get the gw address from elsewhere and cns consumes it?
- is the hns default gw address not changeable?

ZetaoZhuang:
> A couple clarifications:
> 
> * how do the apipa gw and loopback adapter come up the first time with no restart if they both use 169.254.128.1?
hns allows this the first time with no restart
> * is the long term fix that we get the gw address from elsewhere and cns consumes it?
yep. cns get it from dnc 
> * is the hns default gw address not changeable?
atm, right

cns/hnsclient/hnsclient_windows.go
ZetaoZhuang: the caller of CreateHostNCApipaEndpoint is https://github.com/Azure/azure-container-networking/blob/3ed0bcd0cae71c44396e518e887c5df3d8e81f75/cns/restserver/api.go#L1151. where is the place we handle the request. But I dont think that is the place to have this kind of logic to update the gw. what do you think? @QxBytes

QxBytes: Ideally I would like to get a different value from networkContainerDetails but I see that in order to do that we'd need to store the  .2 in the cns state, which leads us back to the long term fix of needing to get the gw address from elsewhere and cns consuming it.

QxBytes (approved):

cns/hnsclient/hnsclient_windows.go
```diff
@@ -604,6 +608,17 @@ func CreateHostNCApipaEndpoint(
 	return endpoint.Id, nil
 }
 
+// adjustIPConfig applies change on gw IP address.
```
behzad-mir, line 611: Please add a TODO to redesign the hnsclient on windows

behzad-mir (changes requested):
Please add a TODO to redesign the hnsclient on windows

cns/hnsclient/hnsclient_windows.go
```diff
@@ -53,6 +53,9 @@ const (
 	// Name of the loopback adapter needed to create Host NC apipa network
 	hostNCLoopbackAdapterName = "LoopbackAdapterHostNCConnectivity"
 
+	defaultHnsGwIPAddress       = "169.254.128.2"
+	hnsLoopbackAdapterIPAddress = "169.254.128.1"
+	apipaRemoteIPAddress        = "169.254.128.1"
```
ashvindeodhar, line 58: do we need these 2 different var names?
loopbackadapter is remote ip address

cns/hnsclient/hnsclient_windows.go
```diff
@@ -53,6 +53,9 @@ const (
 	// Name of the loopback adapter needed to create Host NC apipa network
 	hostNCLoopbackAdapterName = "LoopbackAdapterHostNCConnectivity"
 
+	defaultHnsGwIPAddress       = "169.254.128.2"
```
ashvindeodhar, line 56: can we please add a comment explaining why this GW is being set to .2? Please explain the HNS rehydration issue which needs this GW to be different than the loopback adapter ip

cns/hnsclient/hnsclient_windows.go
ashvindeodhar, line 611: pls add clarification on which GW IP address? NW or endpoint?

cns/hnsclient/hnsclient_windows.go
```diff
@@ -604,6 +608,17 @@ func CreateHostNCApipaEndpoint(
 	return endpoint.Id, nil
 }
 
+// adjustIPConfig applies change on gw IP address.
+// Currently, cns using the same ip address "169.254.128.1" for both apipa gw and loopback adapter. This cause conflict issue when hns get restarted and not able to rehydrate the apipa endpoints.
+// This func is to overwrite the address to 169.254.128.2 when the gateway address is 169.254.128.1
+func adjustIPConfig(localIPConfiguration *cns.IPConfiguration) {
```
ashvindeodhar, line 614: let's rename the func to indicate which IP config is getting updated? something along the lines of 
UpdateGwForLocalIPConfiguration

cns/hnsclient/hnsclient_windows.go
```diff
@@ -301,10 +305,10 @@ func createHostNCApipaNetwork(
 		if interfaceExists, _ := networkcontainers.InterfaceExists(hostNCLoopbackAdapterName); !interfaceExists {
 			ipconfig := cns.IPConfiguration{
 				IPSubnet: cns.IPSubnet{
-					IPAddress:    localIPConfiguration.GatewayIPAddress,
+					IPAddress:    hnsLoopbackAdapterIPAddress,
```
tamilmani1989, line 308: why changing here? anyway you are updating with `updateGwForLocalIPConfiguration`

cns/hnsclient/hnsclient_windows.go
```diff
@@ -301,10 +305,10 @@ func createHostNCApipaNetwork(
...
 					PrefixLength: localIPConfiguration.IPSubnet.PrefixLength,
 				},
-				GatewayIPAddress: localIPConfiguration.GatewayIPAddress,
+				GatewayIPAddress: hnsLoopbackAdapterIPAddress,
```
tamilmani1989, line 311: why changing here? anyway you are updating with `updateGwForLocalIPConfiguration`

ZetaoZhuang: we should always use .1 for this ip

ZetaoZhuang: loopback adapter gw should always be .1. we dont want to change it

cns/hnsclient/hnsclient_windows.go
```diff
@@ -604,6 +609,17 @@ func CreateHostNCApipaEndpoint(
 	return endpoint.Id, nil
 }
 
+// updateGwForLocalIPConfiguration applies change on APIPA NW gw IP address.
```
ashvindeodhar, line 612: isn't this change resulting into the GW value getting updated for both APIPA NW and APIPA endpoint?

ashvindeodhar (dismissed):

cns/hnsclient/hnsclient_windows.go
```diff
@@ -301,7 +305,7 @@ func createHostNCApipaNetwork(
 		if interfaceExists, _ := networkcontainers.InterfaceExists(hostNCLoopbackAdapterName); !interfaceExists {
 			ipconfig := cns.IPConfiguration{
 				IPSubnet: cns.IPSubnet{
-					IPAddress:    localIPConfiguration.GatewayIPAddress,
+					IPAddress:    hnsLoopbackAdapterIPAddress,
 					PrefixLength: localIPConfiguration.IPSubnet.PrefixLength,
 				},
 				GatewayIPAddress: localIPConfiguration.GatewayIPAddress,
```
tamilmani1989, line 311:
```suggestion
				GatewayIPAddress: defaultHnsGwIPAddress,
```

behzad-mir (approved):

tamilmani1989 (approved):

PR #2941, opened 2024-08-20
Add UTs to test findMasterInterface

paulyufan2:
**Reason for Change**:

This PR is to add UTs to test findMasterInterface()

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network_test.go
+++ b/cni/network/network_test.go
@@ -1253,13 +1253,11 @@ func (n *InterfaceGetterMock) GetNetworkInterfaceAddrs(iface *net.Interface) ([]
 	}
 
 	// actual net.Addr invokes syscall; here just create a mocked net.Addr{}
-	ip := &net.IPAddr{
-		IP:   []byte{'A'},
-		Zone: iface.Name,
-	}
-
 	res := []net.Addr{}
-	res = append(res, ip)
+	res = append(res, &net.IPNet{
+		IP:   net.IPv4(10, 255, 0, byte(iface.Index)),
+		Mask: net.IPv4Mask(255, 255, 255, 0),
+	})
 
 	return res, nil
 }
@@ -1787,7 +1785,8 @@ func TestFindMasterInterface(t *testing.T) {
 				netClient: &InterfaceGetterMock{
 					interfaces: []net.Interface{
 						{
-							Name: "eth0",
+							Index: 0,
+							Name:  "eth0",
 						},
 					},
 				},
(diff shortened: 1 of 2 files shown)


timraymond:
Why is the implementation changing for a PR that intends to only add tests?

cni/network/network_test.go
```diff
@@ -1246,6 +1247,21 @@ func (n *InterfaceGetterMock) GetNetworkInterfaces() ([]net.Interface, error) {
 	return n.interfaces, nil
 }
 
+func (n *InterfaceGetterMock) GetNetworkInterfaceAddrs(iface *net.Interface) ([]net.Addr, error) {
```
QxBytes, line 1250: We should be able to pass into the InterfaceGetterMock struct the potential []net.Addrs to return from this function and based on the interface, return a particular []net.Addr based on some field in the net.Interface struct. What you have here gives us a mocked response, but what if we have more complex tests that require us to return certain addresses for one particular interface, and another set of addresses for a different interface?

QxBytes (changes requested):

cni/network/network_test.go
paulyufan2: seems like we can only take advantage of index from net.Interface...

cni/network/network_test.go
```diff
@@ -1246,6 +1251,22 @@ func (n *InterfaceGetterMock) GetNetworkInterfaces() ([]net.Interface, error) {
...
+	if n.err != nil {
+		return nil, n.err
+	}
+
+	// actual net.Addr invokes syscall; here just create a mocked net.Addr{}
+	res := []net.Addr{}
+	for _, intf := range n.interfaces {
+		if iface.Name == intf.Name {
+			netAddr := n.intfAddr[iface.Name].ipNet
```
QxBytes, line 1263: Could we change the map value type to a []net.Addr (or if following your code's types more closely, []net.IPNet)-- is there a reason we set the type as interfaceAddr-- why can't we just use the single field's type ("ipnet net.IPNet") directly? 

Right now I think we can only return a single result in the slice/array that is returned. We can only map 1 net.Interface (the iface passed in) to 1 of the interfaces in our map anyway right? If so, we can probably just return the value from the if statement itself instead of going through the whole loop.

QxBytes (dismissed):

cni/network/network_test.go
```diff
@@ -1236,6 +1237,7 @@ func TestGetPodSubnetNatInfo(t *testing.T) {
 
 type InterfaceGetterMock struct {
 	interfaces []net.Interface
+	intfAddr   map[string][]net.Addr // key is interfaceName, value is one interface's CIDRs(IPs+Masks)
```
tamilmani1989, line 1240: can it be `interfaceAddrs`

QxBytes (approved):

tamilmani1989 (approved):

PR #2942, opened 2024-08-20
ci: remove ovs create bridge test

QxBytes:
**Reason for Change**:

TestCreateBridge is the only test that fails when running locally on dev machine (permission denied), making it difficult to isolate if a ut failure actually ocurred or not. This mode's behavior is unlikely to change so the test is removed/commented.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/ovs_networkclient_linux_test.go
+++ b/network/ovs_networkclient_linux_test.go
@@ -28,27 +28,6 @@ func TestAddRoutes(t *testing.T) {
 	}
 }
 
-func TestCreateBridge(t *testing.T) {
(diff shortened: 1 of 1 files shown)


timraymond (changes requested):
I'd rather see this gated behind a build tag or removed entirely (deleted code can be found in git and resurrected if needed).

tamilmani1989 (approved):

timraymond (approved):

PR #2961, opened 2024-08-27
Removed os.Exec'ing code

rejain456:
**Reason for Change**:

There was a CRI recently where it appears that when CNS is shelling out to PowerShell, spooling up the whole PowerShell env is quite heavy; thus, we are OOMing the CNS pod. There could be other contributing factors that haven't been identified yet as well.

**Issue Fixed**:

This pr gets rid of os.Exec'ing a shell (creating a whole new instance of PowerShell) each time and instead replacing the code with calling into windows registry. 

**Requirements**:

- [ ] includes documentation
- [ X] adds unit tests

**Notes**:

Work Item: https://msazure.visualstudio.com/One/_workitems/edit/29254102

### DIFF
--- a/platform/osRegistryInterface.go
+++ b/platform/osRegistryInterface.go
@@ -1,0 +1,46 @@
+//go:build windows
+// +build windows
+
+package platform
+
+import "golang.org/x/sys/windows/registry"
+
+// Registry interface for interacting with the Windows registry
+type Registry interface {
+	OpenKey(k registry.Key, path string, access uint32) (RegistryKey, error)
+}
+
+// RegistryKey interface to represent an open registry key
+type RegistryKey interface {
+	GetStringValue(name string) (string, uint32, error)
+	SetStringValue(name, value string) error
+	Close() error
+}
+
+type WindowsRegistry struct{}
+
+// WindowsRegistryKey implements the RegistryKey interface
+type WindowsRegistryKey struct {
+	key registry.Key
+}
+
+func (r *WindowsRegistry) OpenKey(k registry.Key, path string, access uint32) (RegistryKey, error) {
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, path, access)
+	if err != nil {
+		return nil, err
+	}
+	return &WindowsRegistryKey{key}, nil
+}
+
+func (k *WindowsRegistryKey) GetStringValue(name string) (val string, valtype uint32, err error) {
+	value, valType, err := k.key.GetStringValue(name)
+	return value, valType, err
+}
+
+func (k *WindowsRegistryKey) SetStringValue(name, value string) error {
+	return k.key.SetStringValue(name, value)
+}
+
+func (k *WindowsRegistryKey) Close() error {
+	return k.key.Close()
+}
--- a/platform/os_windows.go
+++ b/platform/os_windows.go
@@ -20,6 +20,9 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
 )
 
 const (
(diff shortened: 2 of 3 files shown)


platform/os_windows.go
```diff
@@ -294,6 +292,50 @@ func SetSdnRemoteArpMacAddress(execClient ExecClient) error {
 	return nil
 }
 
+// straight out of chat gpt
+func restartService(serviceName string) error {
+	// Connect to the service manager
+	m, err := mgr.Connect()
+	if err != nil {
+		return fmt.Errorf("could not connect to service manager: %v", err)
```
timraymond, line 300: Please use `errors.Wrap`

platform/os_windows.go
```diff
@@ -294,6 +292,50 @@ func SetSdnRemoteArpMacAddress(execClient ExecClient) error {
...
+	if err != nil {
+		return fmt.Errorf("could not query service status: %v", err)
+	}
+	for status.State != svc.Stopped {
+		time.Sleep(500 * time.Millisecond)
+		status, err = service.Query()
+		if err != nil {
+			return fmt.Errorf("could not query service status: %v", err)
+		}
+	}
```
timraymond, line 328: This should consult a `context.Context` for cancelation so that it will not run indefinitely.

platform/osRegistryInterface.go
```diff
@@ -0,0 +1,46 @@
+//go:build windows
+// +build windows
```
timraymond, line 2: This module requires at least Go 1.18 to build, so we no longer need to add `// +build windows`. The `//go:build` pragma is the only one needed.

platform/osRegistryInterface.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+type WindowsRegistry struct{}
+
+// WindowsRegistryKey implements the RegistryKey interface
+type WindowsRegistryKey struct {
+	key registry.Key
+}
+
+func (r *WindowsRegistry) OpenKey(k registry.Key, path string, access uint32) (RegistryKey, error) {
```
timraymond, line 27: `k` is unused here--I don't think that's intentional. Granted, we're probably always going to use HKLM, but who knows? If we're making a generic wrapper, I'd like to be able to, say, edit some keys in HKCU.

platform/osRegistryInterface.go
```diff
@@ -0,0 +1,46 @@
...
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, path, access)
+	if err != nil {
+		return nil, err
+	}
+	return &WindowsRegistryKey{key}, nil
+}
+
+func (k *WindowsRegistryKey) GetStringValue(name string) (val string, valtype uint32, err error) {
+	value, valType, err := k.key.GetStringValue(name)
+	return value, valType, err
```
timraymond, line 37: This simplifies to

```suggestion
	return k.key.GetStringValue(name)
```

platform/os_windows_test.go
```diff
@@ -4,18 +4,66 @@ import (
...
+}
+
+func (k *MockRegistryKey) SetStringValue(name, value string) error {
+	k.Values[name] = value
+	return nil
+}
+
+func (k *MockRegistryKey) Close() error {
+	return nil
+}
```
timraymond, line 52: Would it be incorrect to not close a key? It might be worth adding some bookkeeping in here to make sure it's getting called if so.

platform/os_windows_test.go
```diff
@@ -4,18 +4,66 @@ import (
...
 
+// MockRegistry is a mock implementation of the Registry interface
+type MockRegistry struct {
+	Keys map[string]*MockRegistryKey
+}
+
+// OpenKey opens a mock registry key.
+func (r *MockRegistry) OpenKey(k registry.Key, path string, access uint32) (RegistryKey, error) {
+	// Directly check if the key exists in the mock registry by its path
+	if key, exists := r.Keys[path]; exists {
```
timraymond, line 27: nit: Traditionally, the Go community writes the second parameter here as `ok`.
```suggestion
	if key, ok := r.Keys[path]; ok {
```

platform/os_windows_test.go
```diff
@@ -116,34 +164,36 @@ func TestExecuteCommandError(t *testing.T) {
...
+	//mockExecClient := NewMockExecClient(false)
+	mockRegistry := initMockRegistry()
 	// testing skip setting SdnRemoteArpMacAddress when hns not enabled
-	mockExecClient.SetPowershellCommandResponder(func(_ string) (string, error) {
-		return "False", nil
-	})
-	err := SetSdnRemoteArpMacAddress(mockExecClient)
+	// mockExecClient.SetPowershellCommandResponder(func(_ string) (string, error) {
+	// 	return "False", nil
+	// })
```
timraymond, line 172: Feel free to remove these. Dead code can be found and resurrected with git if necessary.

platform/osRegistryInterface.go
rbtr: as someone with no understanding of win reg, you could be making these up and I would have no idea 😂

rbtr:
while i'm thinking about it we'll probably want to backport this to 1.4/1.5 since those are still a substantial install base

platform/osRegistryInterface.go
timraymond: @rbtr this is cursed knowledge from working summers at my high school playing student sysadmin :-P The Norton Ghosts still haunt me.

PR #2965, opened 2024-08-28
Add UTs for CNI Windows Policy

paulyufan2:
**Reason for Change**:

This PR is to add UTs for CNI windows policy

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/policy/policy_windows_test.go
+++ b/network/policy/policy_windows_test.go
@@ -5,6 +5,7 @@ package policy
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 
 	"github.com/Microsoft/hcsshim/hcn"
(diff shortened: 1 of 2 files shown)


network/policy/policy_windows_test.go
```diff
@@ -40,11 +42,95 @@ var _ = Describe("Windows Policies", func() {
...
+			}
+			expectedPolicy := `{"Exceptions":["10.240.0.0/16","10.0.0.0/8"]}`
+
+			generatedPolicy, err := GetHcnOutBoundNATPolicy(policy, nil)
+			Expect(err).To(BeNil())
+			Expect(string(generatedPolicy.Settings)).To(Equal(expectedPolicy))
+		})
+	})
+
+	Describe("Test GetHcnRoutePolicy", func() {
```
QxBytes, line 109: could we add tests to ensure nothing panics/we get expected output for each of these tests if we pass in empty json into Data or any of the fields which accept json

network/policy/policy_windows_test.go
```diff
@@ -57,4 +143,80 @@ var _ = Describe("Windows Policies", func() {
...
+				Type:     hcn.PortMapping,
+				Settings: rawPortMappingPolicy,
+			})
+
+			hnsPolicy := Policy{
+				Type: PortMappingPolicy,
+				Data: portMappingPolicy,
+			}
+
+			testPolicies = append(testPolicies, hnsPolicy)
```
QxBytes, line 189: I know we've tested most of the policies from the witch statement in `GetHcnEndpointPolicies` individually, but can we have a test where we pass in more than 1 endpoint policy to confirm they are being looped over and added properly?

QxBytes (approved):
Follow up on GetHcnEndpointPolicies intended behavior as it only seems to select one type of policy based on the parameter ("EndpointPolicy" type).

PR #2967, opened 2024-08-28
fix: Handle async delete in stateless cni

behzad-mir:
**Issue Fixed**:

This is an attempt to fix the HNS endpoint leak by stateless CNI when the aynch delete is enabled.
CNS needs to delete the associate HNS endpoint ids upon releasing IPs in asynch delete mode.

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/network/network.go
```diff
@@ -1125,18 +1127,31 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 		// network ID is passed in and used only for migration
 		// otherwise, in stateless, we don't need the network id for deletion
 		epInfos, err = plugin.nm.GetEndpointState(networkID, args.ContainerID)
+		// if stateless CNI fail to get the endpoint from CNS for any reason other than  Endpoint Not found
+		if err != nil && !strings.Contains(err.Error(), "endpoint state could not be found in the statefile") {
```
tamilmani1989, line 1131: what if this error msg changed in cns tomorrow? don't we have error code?

tamilmani1989: probably use an error type for this .. common for cni and cns?

cni/network/network.go
```diff
@@ -1125,18 +1127,31 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+			if strings.Contains(err.Error(), "No connection could be made") {
+				addErr := fsnotify.AddFile("stateless_"+args.ContainerID, args.ContainerID, watcherPath)
```
tamilmani1989, line 1133: why adding stateless prefix? why can't we follow same pattern for release ip

cni/network/network.go
```diff
@@ -1125,18 +1127,31 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+				if addErr != nil {
+					logger.Error("Failed to add file to watcher", zap.String("containerID", args.ContainerID), zap.Error(addErr))
+					return errors.Wrap(addErr, fmt.Sprintf("failed to add file to watcher with containerID %s", args.ContainerID))
+				}
+
```
tamilmani1989, line 1138: shouldn't we return nil after adding file?

cni/network/network.go
```diff
@@ -1125,18 +1127,31 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+			return plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
+		}
 	} else {
 		epInfos = plugin.nm.GetEndpointInfosFromContainerID(args.ContainerID)
 	}
 
 	// for when the endpoint is not created, but the ips are already allocated (only works if single network, single infra)
-	// stateless cni won't have this issue
+	// stateless cni might get into this issue if endpoint is not present in the statefile
 	if len(epInfos) == 0 {
```
tamilmani1989, line 1148: this cannot happen for stateless cni right?

cns/fsnotify/fsnotify.go
```diff
@@ -68,7 +73,16 @@ func (w *watcher) releaseAll(ctx context.Context) {
 		}
 		file.Close()
 		podInterfaceID := string(data)
-
+		// in case of stateless CNI, CNS needs to remove HNS endpoitns first
+		if strings.Contains(podInterfaceID, "stateless") && runtime.GOOS == "windows" {
```
tamilmani1989, line 77: we have to rethink if need "stateless" check.. this can happen only if cx running 1.6 cns reads 1.5 cni async file.. this can happen only during upgrades. what's the harm in calling delete hns endpoint for those? it would be a no-op right?

tamilmani1989 (commented):
can we add any UTs for this?

cni/network/network.go
behzad-mir: Will do

cni/network/network.go
behzad-mir: There's no way for me to specify that this is a stateless case other than writing to the file. CNS only delete HNS in statelss case and there is no way for CNS to determine if the CNI is stateless or statefull

cni/network/network.go
behzad-mir: it can. When the contaierd issues 3 delete and the first one has gone successful

cni/network/network.go
rbtr: why not always try to delete the HNS state?

cns/hnsclient/hnsclient_linux.go
```diff
@@ -50,3 +50,9 @@ func DeleteHostNCApipaEndpoint(
 	networkContainerID string) error {
 	return nil
 }
+
+// DeleteHNSEndpointbyID deletes the HNS endpoint
```
rbtr, line 54: `hnsclient_linux` is nonsensical and should not exist, why is it like this 😢

tamilmani1989: the file should not exist at all or we should rename it

cni/network/network.go
behzad-mir: Because for stateful CNI we don't have the HNS ID in the state so we have to find the HNS via IP address first and then issue a delete. That's what we decided to do. We have to also make sure that the manage Endpoint state is indeed set to true. If in the future CIlium for Windows introduced then we need to revisit this again.

cns/hnsclient/hnsclient_linux.go
behzad-mir: I was struggling for a half day to see why the new functions is undefined and then saw this "linux" file. We have to rename this!

behzad-mir: I think it's an old code!

cni/network/network.go
tamilmani1989: in that case won't successive calls return endpoint not found error?

cns/client/client.go
```diff
@@ -1050,7 +1058,11 @@ func (c *Client) GetEndpoint(ctx context.Context, endpointID string) (*restserve
 		return nil, errors.Wrap(err, "failed to decode GetEndpointResponse")
 	}
 
+	if response.Response.ReturnCode == types.NotFound {
```
tamilmani1989, line 1061: did we make sure `types.NotFound` set by CNS when endpoint not found in its state?

cns/fsnotify/fsnotify.go
```diff
@@ -5,18 +5,25 @@ import (
...
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 )
 
+const DefaultNetworkID = "azure"
+
 type releaseIPsClient interface {
 	ReleaseIPs(ctx context.Context, ipconfig cns.IPConfigsRequest) error
+	GetEndpoint(ctx context.Context, endpointID string) (*restserver.GetEndpointResponse, error)
```
tamilmani1989, line 26: should we change interface name or add to new interface?

cns/fsnotify/fsnotify.go
```diff
@@ -68,7 +79,14 @@ func (w *watcher) releaseAll(ctx context.Context) {
 		}
 		file.Close()
 		podInterfaceID := string(data)
-
+		// in case of stateless CNI for Windows, CNS needs to remove HNS endpoitns first
+		if isStalessCNIMode(w.cnsconfig) {
+			// remove HNS endpoint
+			w.log.Info("deleting HNS Endpoint asynchronously")
+			if err := w.deleteHNSEndpoint(ctx, containerID); err != nil {
```
tamilmani1989, line 86: this is common file. should we name it as `deleteEndpoint` and move hns api to windows specific file?

tamilmani1989: i see you are doing it. probably rename `deleteEndpoint` in common file

cns/hnsclient/hnsclient_windows.go
```diff
@@ -685,3 +686,67 @@ func DeleteHostNCApipaEndpoint(
...
+		return fmt.Errorf("Failed to delete endpoint: %+v. Error: %v", hnsEndpointID, err)
+	}
+
+	logger.Errorf("[Azure CNS] Successfully deleted endpoint: %+v", hnsEndpointID)
+
+	return nil
+}
+
+// GetHNSEndpointbyIP returns an HNSEndpoint with the corrsponding HNS Endpoint ID that matches an specific IP Address.
+func GetHNSEndpointbyIP(ipv4, ipv6 []net.IPNet, networkID string) (string, error) {
```
tamilmani1989, line 725: why we pass ipv4, ipv6 separately.. can we detect keep it net.IPNet and based on v4/v6 we can search respective slices

.pipelines/pipeline.yaml
```diff
@@ -329,6 +329,17 @@ stages:
         vmSize: Standard_B2ms
         k8sVersion: ""
         dependsOn: "containerize"
+    # Azure Overlay E2E Stateless CNI tests
+    - template: singletenancy/azure-cni-overlay-stateless/azure-cni-overlay-stateless-e2e-job-template.yaml
```
tamilmani1989, line 333: can we also create one for dualstack?

cns/client/client.go
behzad-mir: yes, we were already doing it on API handler in CNS for that specific error

cns/fsnotify/fsnotify.go
behzad-mir: I can create a new interface for this.

cns/fsnotify/fsnotify.go
behzad-mir: Make sense

cni/network/network.go
behzad-mir: corrected in the new commit and we catch the error before this block.

cns/fsnotify/fsnotify.go
behzad-mir: changed it to IPAMCLient

cns/hnsclient/hnsclient_windows.go
behzad-mir: TH IPInfo struct has IPv4 and v6 separated in two different fields so we cannot pass a single parameter here.

cni/network/network.go
```diff
@@ -1125,18 +1126,37 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+		}
 	} else {
 		epInfos = plugin.nm.GetEndpointInfosFromContainerID(args.ContainerID)
 	}
 
 	// for when the endpoint is not created, but the ips are already allocated (only works if single network, single infra)
-	// stateless cni won't have this issue
-	if len(epInfos) == 0 {
+	// this block is not applied to stateless CNI
+	if len(epInfos) == 0 && !plugin.nm.IsStatelessCNIMode() {
```
tamilmani1989, line 1153: remove this check `!plugin.nm.IsStatelessCNIMode()`. its not necessary. anyway stateless cni would not get into this

cni/network/network.go
```diff
@@ -1125,18 +1126,37 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+				addErr := fsnotify.AddFile(args.ContainerID, args.ContainerID, watcherPath)
+				logger.Info("add containerid file for Asynch delete", zap.String("containerID", args.ContainerID), zap.Error(addErr))
+				if addErr != nil {
+					logger.Error("failed to add file to watcher", zap.String("containerID", args.ContainerID), zap.Error(addErr))
+					return errors.Wrap(addErr, fmt.Sprintf("failed to add file to watcher with containerID %s", args.ContainerID))
+				}
+				return nil
+			}
+			if errors.Is(err, network.ErrEndpointStateNotFound) {
+				logger.Info("Endpoint Not found for:", zap.String("containerID", args.ContainerID), zap.Error(err))
```
tamilmani1989, line 1141:
```suggestion
				logger.Info("Endpoint Not found", zap.String("containerID", args.ContainerID), zap.Error(err))
```

cni/network/network.go
```diff
@@ -1125,18 +1126,37 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+				return nil
+			}
+			logger.Error("Get Endpoint Stat API return with error for: ", zap.String("containerID", args.ContainerID), zap.Error(err))
```
tamilmani1989, line 1144:
```suggestion
			logger.Error("Get Endpoint Stat API returned error", zap.String("containerID", args.ContainerID), zap.Error(err))
```

cns/fsnotify/fsnotify.go
tamilmani1989: How about `endpointManager`?

cns/fsnotify/fsnotify.go
```diff
@@ -206,3 +224,34 @@ func (w *watcher) releaseIP(ctx context.Context, podInterfaceID, containerID str
...
+		w.log.Info("deleting HNS Endpoint with id ", zap.String("id", hnsEndpointID))
+		if err := hnsclient.DeleteHNSEndpointbyID(hnsEndpointID); err != nil {
+			return errors.Wrap(err, "failed to delete HNS endpoint with id "+ipInfo.HnsEndpointID)
+		}
+	}
+	return nil
+}
+
+// isStalessCNIMode verify if the CNI is running stateless mode
+func isStalessCNIMode(cnsconfig *configuration.CNSConfig) bool {
```
tamilmani1989, line 251:
```suggestion
func isStalessCNIWindows(cnsconfig *configuration.CNSConfig) bool {
```

cns/fsnotify/fsnotify.go
```diff
@@ -206,3 +224,34 @@ func (w *watcher) releaseIP(ctx context.Context, podInterfaceID, containerID str
...
+		if ipInfo.HnsEndpointID == "" {
+			if hnsEndpointID, err = hnsclient.GetHNSEndpointbyIP(ipInfo.IPv4, ipInfo.IPv6, DefaultNetworkID); err != nil {
+				return errors.Wrap(err, "failed to find HNS endpoint with id")
+			}
+		}
+		w.log.Info("deleting HNS Endpoint with id ", zap.String("id", hnsEndpointID))
+		if err := hnsclient.DeleteHNSEndpointbyID(hnsEndpointID); err != nil {
+			return errors.Wrap(err, "failed to delete HNS endpoint with id "+ipInfo.HnsEndpointID)
+		}
+	}
```
tamilmani1989, line 246: this should be in windows specific file. may be in another PR.. lets remove `hnsclient_linux.go` and `hnsclient_windows.go` and instead have `endpoint_linux.go` and `endpoint_windows.go` and abstract hns changes in `endpoint_windows.go`

cns/hnsclient/hnsclient_windows.go
tamilmani1989: hmm ok.. can we pass ipinfo reference atleast?

cni/network/network.go
behzad-mir: Will be removed.

cns/fsnotify/fsnotify.go
behzad-mir: Will make a TODO comment and do another PR to address this since it affects the multitenancy too and it's beyond this PR goal

cns/types/codes.go
```diff
@@ -45,6 +45,7 @@ const (
 	UnsupportedAPI                         ResponseCode = 43
 	FailedToAllocateBackendConfig          ResponseCode = 44
 	UnexpectedError                        ResponseCode = 99
+	ConnectionError                        ResponseCode = 404
```
tamilmani1989, line 48: instead of 404, can we do "45"

tamilmani1989 (changes requested):

cns/fsnotify/fsnotify.go
```diff
@@ -5,41 +5,47 @@ import (
...
+	"github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/Azure/azure-container-networking/cns/restserver"
 	"github.com/fsnotify/fsnotify"
 	"github.com/pkg/errors"
-	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 )
 
-type releaseIPsClient interface {
+const DefaultNetworkID = "azure"
```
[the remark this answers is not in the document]
behzad-mir: hcn.ListEndpointsOfNetwork() needs the HNS NEtworkID which we can get it by another API call hcn.GetNetworkByName(networkID)

cns/fsnotify/fsnotify.go
behzad-mir: missed this one. Mad the change in new commit.

cns/types/codes.go
behzad-mir: will do

cns/hnsclient/hnsclient_windows.go
```diff
@@ -75,6 +76,9 @@ const (
 
 	// signals a APIPA endpoint type
 	apipaEndpointType = "APIPA"
+
+	// default network id used by HNS
+	DefaultNetworkID = "azure"
```
tamilmani1989, line 81:
```suggestion
	defaultNetworkName= "azure"
```

behzad-mir: corrected

cni/network/network.go
behzad-mir: THe check has moved to a function named isStalessCNIWindows()

network/errors.go
```diff
@@ -3,6 +3,9 @@ package network
 import "errors"
 
 var (
-	errSubnetV6NotFound = errors.New("Couldn't find ipv6 subnet in network info")
-	errV6SnatRuleNotSet = errors.New("ipv6 snat rule not set. Might be VM ipv6 address missing")
+	errSubnetV6NotFound        = errors.New("Couldn't find ipv6 subnet in network info")                // nolint
+	errV6SnatRuleNotSet        = errors.New("ipv6 snat rule not set. Might be VM ipv6 address missing") // nolint
+	ErrEndpointStateNotFound   = errors.New("endpoint state could not be found in the statefile")
+	ErrConnectionFailure       = errors.New("couldn't connect to CNS daemonset")
```
rbtr, line 9: probably not daemonset?
```suggestion
	ErrConnectionFailure       = errors.New("couldn't connect to CNS")
```

cns/service/main.go
```diff
@@ -950,14 +950,17 @@ func main() {
...
+		logger.Printf("AsyncPodDelete is enabled")
 		cnsclient, err := cnsclient.New("", cnsReqTimeout) //nolint
 		if err != nil {
 			z.Error("failed to create cnsclient", zap.Error(err))
 		}
 		go func() {
 			_ = retry.Do(func() error {
 				z.Info("starting fsnotify watcher to process missed Pod deletes")
-				w, err := fsnotify.New(cnsclient, cnsconfig.AsyncPodDeletePath, z)
+				logger.Printf("starting fsnotify watcher to process missed Pod deletes")
```
rbtr, line 962: i don't think we need to dupe logs like this

cns/fsnotify/fsnotify.go
```diff
@@ -5,41 +5,45 @@ import (
...
-type releaseIPsClient interface {
+type endpointManager interface {
 	ReleaseIPs(ctx context.Context, ipconfig cns.IPConfigsRequest) error
+	GetEndpoint(ctx context.Context, endpointID string) (*restserver.GetEndpointResponse, error)
 }
 
 type watcher struct {
-	cli  releaseIPsClient
-	path string
-	log  *zap.Logger
```
rbtr, line 25: why did you change this from injecting a zap logger to the old bad global logger package?

cns/fsnotify/fsnotify.go
```diff
@@ -5,41 +5,45 @@ import (
...
-
+	cli           endpointManager
+	path          string
 	pendingDelete map[string]struct{}
 	lock          sync.Mutex
+	cnsconfig     *configuration.CNSConfig
```
rbtr, line 32: don't pass the cns config in here, this package does not need to know all of that 🙁

cns/fsnotify/fsnotify.go
```diff
@@ -55,29 +59,38 @@ func (w *watcher) releaseAll(ctx context.Context) {
...
 		if errReadingFile != nil {
-			w.log.Error("failed to read file content", zap.Error(errReadingFile))
+			logger.Errorf("failed to read file content %s", errReadingFile)
 		}
 		file.Close()
 		podInterfaceID := string(data)
-
-		w.log.Info("releasing IP for missed delete", zap.String("podInterfaceID", podInterfaceID), zap.String("containerID", containerID))
+		// in case of stateless CNI for Windows, CNS needs to remove HNS endpoitns first
+		if isStalessCNIWindows(w.cnsconfig) {
```
rbtr, line 77: The asyncdelete watcher should not need to know about the CNI mode. You need to compose it in such a way that it does the right thing with no additional knowledge instead of teaching it about all this extra CNS config and adding branches and complication.
One way you might accomplish this is by decorating the `ReleaseIPs` implementation that gets passed in during initialization. In `main` when we bulid this object, you already have all the context you need about if this is stateless CNI. _There_, wrap `ReleaseIPs` with this extra functionality and the call the underlying impl after you deleteEndpoint, like middleware.
I think you could implement that in a way that there are actually no changes to the fsnotify package at all, and only a small anonymous wrapper func added to this initialization in main.

cns/hnsclient/hnsclient_linux.go
rbtr: This file needs to go. Don't add anything to it, this is a bad pattern taht we shouldn't continue.

rbtr (changes requested):

cns/fsnotify/fsnotify.go
behzad-mir: Zap logger was not logging anything at all in CNS log file. I don't know if it works for Linuxor not but in case of Windows I was only able to see the logs in CNS log file after I used the old logger. Seems like we cannot use the two logger to write on the same file.

cns/fsnotify/fsnotify.go
behzad-mir: Correct. I'll only pass the IntiFromCNI and ManagedEndpointState.

cns/fsnotify/fsnotify.go
behzad-mir: That was my first thought implementing this then I got some issues with the pattern of going through Rest CNS client and APIs internally from CNS. The async delete is not a direct request and I don't think we should go through Rest APIs for its related operations. For the ReleaseIP and GetEndpointState, I think we don't have any other choice other than CNS client and APIs since they are written that way. However, for HNS we still can avoid this.
I had some concerns about having let's say 250 async delete ops and as a result it will create 250*3 rest APIs for (releaseIP + GetEndpointState + DeleteHNSEndpoint). 
I'm more inclined to a solution to avoid going through cnsClient for this. Not sure if we can achieve it without tying up asynchdelte watched to extra packages.

behzad-mir: @tamilmani1989 Let me know what you think.

cns/hnsclient/hnsclient_linux.go
behzad-mir: I tried to delete this file but the Make will fail. Seems like it fails whenever we run the Make in Linux. It seems we can only import hcn package in platfrom specifi file (windows). Do we have an example in our code base for similar situation?
In CNI I know we are adding *Iml to functions and create platfrom specific file like _windows or _linux

cns/client/client.go
```diff
@@ -1023,35 +1024,63 @@ func (c *Client) GetHomeAz(ctx context.Context) (*cns.GetHomeAzResponse, error)
 	return &getHomeAzResponse, nil
 }
 
+// GetEndpoint calls the EndpointHandlerAPI in CNS to retrieve the state of a given EndpointID
+func (c *Client) GetHNSEndpointID(ctx context.Context, endpointID string) ([]string, [][]net.IPNet, error) {
```
tamilmani1989, line 1028: can we add a brief comment on what this function returns

tamilmani1989: funciton name is GetHNSEndpointID and it returns IPs too. lets consider renaming it  and  should we return as [] struct

rbtr: or a map[string][]net.IPNet

behzad-mir: Ended up removing the function since we can cover the previous version (with GetEndpoint+deleteEndpoint) in EndpointManger package without import cycle

cns/endpointmanager/endpointmanager_linux.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"context"
+
+	"github.com/Azure/azure-container-networking/cns"
+)
+
+// ReleaseIPs implements an Interface in fsnotify for async delete of the HNS endpoint and IP addresses
+func (em *EndpointManager) ReleaseIPs(_ context.Context, _ cns.IPConfigsRequest) error {
+	return nil
+}
```
rbtr, line 12: This is not the correct behavior; it should work if we end up using it. Right now it is noop and won't release IPs. 
It should just call through:
```suggestion
func (em *EndpointManager) ReleaseIPs(ctx context.Context, req cns.IPConfigsRequest) error {
	return em.cli.ReleaseIPs(ctx, req)
}
```

rbtr (changes requested):
few minor things

cns/service/main.go
```diff
@@ -1536,3 +1547,11 @@ func PopulateCNSEndpointState(endpointStateStore store.KeyValueStore) error {
...
 	return nil
 }
+
+// isStalessCNIMode verify if the CNI is running stateless mode
+func isStalessCNIWindows(cnsconfig *configuration.CNSConfig) bool {
+	if !cnsconfig.InitializeFromCNI && cnsconfig.ManageEndpointState && runtime.GOOS == "windows" {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
behzad-mir: that's right. done

cns/configuration/configuration.go
```diff
@@ -231,3 +232,11 @@ func SetCNSConfigDefaults(config *CNSConfig) {
 	config.GRPCSettings.Enable = false
 	config.WatchPods = config.EnableIPAMv2 || config.EnableSwiftV2
 }
+
+// isStalessCNIMode verify if the CNI is running stateless mode
+func (cnsconfig *CNSConfig) IsStalessCNIWindows() bool {
+	if !cnsconfig.InitializeFromCNI && cnsconfig.ManageEndpointState && runtime.GOOS == "windows" {
+		return true
+	}
+	return false
```
rbtr, line 241: why so many ifs 😢 
```suggestion
	return !cnsconfig.InitializeFromCNI && cnsconfig.ManageEndpointState && runtime.GOOS == "windows"
```

rbtr (dismissed):

cns/hnsclient/hnsclient_windows.go
```diff
@@ -75,6 +76,9 @@ const (
 
 	// signals a APIPA endpoint type
 	apipaEndpointType = "APIPA"
+
+	// default network name used by HNS
+	DefaultNetworkName = "azure"
```
tamilmani1989, line 81:
```suggestion
	defaultNetworkName = "azure"
```

behzad-mir: done

tamilmani1989 (approved):

PR #2970, opened 2024-08-28
fix: add legacy IPAM metrics back to IPAMv2

rbtr:
**Reason for Change**:
IPAM metrics that were set by the old IPAM PoolMonitor were not set in the v2 monitor. This breaks the AMA workbooks for subnet usage monitoring on AKS 1.30+ where CNS 1.6 with v2 IPAM is deployed.

The metrics were missing because the v2 PoolMonitor does not use most of that source data anymore - it depends only on Pod count and Batch size, not the full IPAM utilization buckets and previous NNC state that the v1 implementation needed to scale up/down.

This fix creates a "metrics observer" which functionally performs what the IPAM v1 monitor used to do WRT the metrics in a portable closed scope. It is injected into the v2 monitor so that the v2 monitor does not need to be polluted with the data pulls that the v1 monitor needed and can still trigger the metrics updates.

TODO: As the v1 monitor is phased-out (hopefully in the 1.7 release of CNS), these metrics should be removed to more appropriate locations, such as moving the Allocated/Available/Assigned/etc IP state metrics in to the CNS core datastore IPConfig paths where those IP states are mutated.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/ipampool/metrics/observer.go
+++ b/cns/ipampool/metrics/observer.go
@@ -1,0 +1,157 @@
+package metrics
+
+import (
+	"context"
+	"fmt"
+	"net/netip"
+
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/types"
+	"github.com/Azure/azure-container-networking/crd/clustersubnetstate/api/v1alpha1"
+	"github.com/Azure/azure-container-networking/crd/nodenetworkconfig/api/v1alpha"
+	"github.com/pkg/errors"
+)
+
+// Subnet ARM ID /subscriptions/$(SUB)/resourceGroups/$(GROUP)/providers/Microsoft.Network/virtualNetworks/$(VNET)/subnets/$(SUBNET)
+const subnetARMIDTemplate = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/virtualNetworks/%s/subnets/%s"
+
+// ipPoolState is the current actual state of the CNS IP pool.
+type ipPoolState struct {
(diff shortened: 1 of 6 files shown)


cns/ipampool/metrics/observer.go
```diff
@@ -0,0 +1,157 @@
...
+	if subscription == "" || resourceGroup == "" || vnetID == "" || subnetID == "" {
+		return ""
+	}
+	return fmt.Sprintf(subnetARMIDTemplate, subscription, resourceGroup, vnetID, subnetID)
+}
+
+// observeMetrics observes the IP pool and updates the metrics. Blocking.
+//
+//nolint:lll // ignore line length
+func observeMetrics(ctx context.Context, ipcli func() map[string]cns.IPConfigurationStatus, nnccli func(context.Context) (*v1alpha.NodeNetworkConfig, error), csscli func(context.Context) ([]v1alpha1.ClusterSubnetState, error)) error {
```
nddq, line 76: naming things `*cli` is kinda misleading because these are all just getter funcs no? Also, maybe you can define the getter functions type elsewhere to make this cleaner

rbtr: that's all true, but I intentionally went with anonymous types to make the footprint for this as small as possible (and not actually reusable, which is why we would use interfaces). Nothing can escape and it is very easy to rip out later.

rbtr: and `cli` is `client` which a getter func arguably is

nddq (approved):

PR #2973, opened 2024-08-29
ci: add create cluster stage to cilium scale test pipeline

jshr-w:
**Reason for Change**:

Adds stage to create a cluster in Cilium scale test pipeline and allows to specify resource group name separately from cluster name.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/cni/cilium/cilium-scale-test.yaml
+++ b/.pipelines/cni/cilium/cilium-scale-test.yaml
@@ -2,6 +2,32 @@ pr: none
 trigger: none
 
 stages:
+  - stage: create_rg_cluster
+    displayName: "Create RG & Cluster"
+    jobs:
+      - job: create_rg_cluster
+        pool:
+          name: "$(BUILD_POOL_NAME_DEFAULT)"
+        steps:
+          - task: AzureCLI@2
+            inputs:
+              azureSubscription: $(TEST_SUB_SERVICE_CONNECTION)
+              scriptLocation: "inlineScript"
+              scriptType: "bash"
+              addSpnToEnvironment: true
+              inlineScript: |
+                set -ex
+                az extension add --name aks-preview
+
+                subscriptionId="$(az account list --query "[?isDefault].id" --output tsv)"
+                make -C ./hack/aks AZCLI=az ${CLUSTER_TYPE} CLUSTER=${CLUSTER} GROUP=${RESOURCE_GROUP} REGION=${LOCATION} NODE_COUNT=5 VM_SIZE=${VMSIZE} SUB=${subscriptionId} || {
+                  echo "Failed to create Cluster"
+                  exit 1
+                }
+                ls -lah
+                pwd
+                kubectl cluster-info
+                kubectl get po -owide -A
   - stage: update_daemonset_versions
     displayName: "Update Cilium + CNS Version and Restart Nodes"
     jobs:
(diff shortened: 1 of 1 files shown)


.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -2,6 +2,63 @@ pr: none
...
+              scriptType: "bash"
+              addSpnToEnvironment: true
+              inlineScript: |
+                set -ex
+                az extension add --name aks-preview
+                if [ ${CREATE_RG} ]; then
+                  az group create --name ${RESOURCE_GROUP} --location ${LOCATION}
+                fi
+                subscriptionId="$(az account list --query "[?isDefault].id" --output tsv)"
+                make -C ./hack/aks AZCLI=az ${CLUSTER_TYPE} CLUSTER=${CLUSTER} GROUP=${RESOURCE_GROUP} REGION=${LOCATION} NODE_COUNT=5 VM_SIZE=${VMSIZE} SUB=${subscriptionId}
```
jpayne3506, line 25: What functionality are you looking for?

The cluster creation in `hack/aks` creates a resource group every time.

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -2,6 +2,63 @@ pr: none
...
+                pwd
+                kubectl cluster-info
+                kubectl get po -owide -A
+
+                echo "install Cilium ${CILIUM_VERSION_TAG}"
+                export DIR=${CILIUM_VERSION_TAG%.*}
+                echo "installing files from ${DIR}"
+
+                echo "deploy Cilium ConfigMap"
+                if [ ${USE_DUALSTACK} ]; then
```
jpayne3506, line 36: This conditional is only false when empty. Even garbage input will result in true, including `false`.

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -2,6 +2,63 @@ pr: none
...
+                  envsubst '${CILIUM_VERSION_TAG},${CILIUM_IMAGE_REGISTRY},${IPV6_HP_BPF_VERSION}' < test/integration/manifests/cilium/v${DIR}/cilium-agent/templates/daemonset-dualstack.yaml | kubectl apply -f -
+                else
+                  envsubst '${CILIUM_VERSION_TAG},${CILIUM_IMAGE_REGISTRY}' < test/integration/manifests/cilium/v${DIR}/cilium-agent/templates/daemonset.yaml | kubectl apply -f -
+                fi
+
+                envsubst '${CILIUM_VERSION_TAG},${CILIUM_IMAGE_REGISTRY}' < test/integration/manifests/cilium/v${DIR}/cilium-operator/templates/deployment.yaml | kubectl apply -f -
+                kubectl get po -owide -A
+
+                echo "Deploy Azure-CNS"
+                sudo -E env "PATH=$PATH" make test-integration AZURE_IPAM_VERSION=v$(AZURE_IPAM_VERSION) CNS_VERSION=v$(CNS_VERSION) INSTALL_CNS=true INSTALL_OVERLAY=true CNS_IMAGE_REPO=$(CNS_IMAGE_REPO)
```
jpayne3506, line 59: Might want to consider moving this to its own job/step so it can be retried separately. Goldpinger is very flaky. Could also consider leveraging `make load-test` for a basic functionality test before scale testing.

jpayne3506: If you move it to its own job, you can retrigger it separately from the cluster creation steps. Which may be more useful than putting it in its own step.

camrynl, line 59: maybe instead of running integration tests we can apply the cns manifests and inject the env variables similar to how we do it for cilium? 

`test-integration` might cause flakes early in the pipeline

.pipelines/cni/cilium/cilium-scale-test.yaml
```diff
@@ -2,6 +2,63 @@ pr: none
...
+                fi
+                subscriptionId="$(az account list --query "[?isDefault].id" --output tsv)"
+                make -C ./hack/aks AZCLI=az ${CLUSTER_TYPE} CLUSTER=${CLUSTER} GROUP=${RESOURCE_GROUP} REGION=${LOCATION} NODE_COUNT=5 VM_SIZE=${VMSIZE} SUB=${subscriptionId}
+                ls -lah
+                pwd
+                kubectl cluster-info
+                kubectl get po -owide -A
+
+                echo "install Cilium ${CILIUM_VERSION_TAG}"
+                export DIR=${CILIUM_VERSION_TAG%.*}
```
vipul-21, line 32: I think this is similar to the next step where we update cilium+cns. I would recommend combining these two OR skip the next one if this step is executed

.pipelines/cni/cilium/cilium-scale-test.yaml
vipul-21: You can use `make test-load` and `CNS_ONLY=true` as we just want to deploy CNS and IPAM.
Example: ```
sudo -E env "PATH=$PATH" make test-load CNS_ONLY=true CNS_VERSION=<PUT CORRECT VERSION> AZURE_IPAM_VERSION=v0.2.0 INSTALL_CNS=true INSTALL_OVERLAY=true CNS_IMAGE_REPO=MCR
```

.pipelines/cni/cilium/cilium-scale-test.yaml
jshr-w: Ah! Good catch :)

camrynl (dismissed):

camrynl (approved):

PR #2979, opened 2024-08-29
Instantiate Device Plugin

aggarwal0009:
**Reason for Change**:
This PR enables device plugin

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/service/main.go
```diff
@@ -1037,6 +1078,91 @@ func main() {
...
+				for deviceType, count := range deviceCounts {
+					pluginManager.TrackDevices(deviceType, count)
+				}
+
+				// Exit polling loop once the CRD status is successfully processed
+				return nil
+			}
+
+			// Wait before polling again
+			time.Sleep(defaultNodeInfoCRDPollInterval)
```
timraymond, line 1161: `time.Ticker` (remembering to defer `Stop()`) is better here since it does not delay the context cancellation in the event that happens during the cooldown period.

cns/service/main.go
```diff
@@ -880,6 +884,43 @@ func main() {
...
+		pluginManager := deviceplugin.NewPluginManager(z)
+		pluginManager.AddPlugin(mtv1alpha1.DeviceTypeVnetNIC, initialVnetNICCount)
+		pluginManager.AddPlugin(mtv1alpha1.DeviceTypeInfiniBandNIC, initialIBNICCount)
+
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		// Start device plugin manager in a separate goroutine
+		go func() {
+			for {
```
timraymond, line 900: This `for` seems out of place if the intention is to restart `pluginManager` in the event of an error. It might also be wise to have a circuit breaker if it continually fails.

cns/service/main.go
```diff
@@ -880,6 +884,43 @@ func main() {
...
+				case <-ctx.Done():
+					z.Info("Context canceled, stopping plugin manager")
+					return
+				default:
+					if pluginErr := pluginManager.Run(ctx); pluginErr != nil {
+						z.Error("plugin manager exited with error", zap.Error(pluginErr))
+						time.Sleep(defaultDevicePluginRetryInterval)
+					} else {
+						return
+					}
```
timraymond, line 911: This construction is what makes me think the `for` should, at least, be more tightly placed around the `pluginManager`'s `Run()`.

cns/service/main.go
```diff
@@ -880,6 +884,43 @@ func main() {
 		}
 	}
 
+	if cnsconfig.EnableSwiftV2 && cnsconfig.EnableK8sDevicePlugin {
+		initialVnetNICCount := 0
+		initialIBNICCount := 0
```
timraymond, line 889: If we're not writing to these, and assigning them off to appease the magic number linter, they should probably be consts.

cns/service/main.go
```diff
@@ -1037,6 +1078,91 @@ func main() {
...
+		default:
+			// Fetch the CRD status
+			nodeInfo, err := nodeInfoCli.Get(ctx, node.Name)
+			if err != nil {
+				zlog.Error("Error fetching nodeinfo CRD", zap.Error(err))
+				return errors.Wrap(err, "failed to get nodeinfo crd")
+			}
+
+			// Check if the status is set
+			if !reflect.DeepEqual(nodeInfo.Status, mtv1alpha1.NodeInfoStatus{}) && len(nodeInfo.Status.DeviceInfos) > 0 {
```
timraymond, line 1134: `github.com/google/go-cmp/cmp.Equal` is better than `reflect.DeepEqual` for a lot of reasons. It's a drop-in replacement, and it does the right thing more consistently.

timraymond (changes requested):

cns/service/main.go
aggarwal0009: Updated

cns/service/main.go
aggarwal0009: makes sense

cns/service/main.go
aggarwal0009: added circuit breaker

cns/service/main.go
aggarwal0009: done

cns/service/main.go
aggarwal0009: updated

cns/service/main.go
aggarwal0009: We want to try running plugin only after a certain interval, hence the "for select case" structure around the pluginManager's Run()

cns/service/main.go
timraymond: I made this recommendation prior to learning that cmp was intended for tests only. Given that the intention here is to test whether this type is a zero value, I believe it should be sufficient to test whether `nodeInfo.Status.DeviceInfos != nil && len(nodeInfo.Status.DeviceInfos) > 0`. Second opinion might be nice though. @rbtr ?

miguelgoms (approved):

timraymond (approved):

timraymond:
@aggarwal0009 did not see you had updated this. Please hit the re-review button in the future so I see this.

PR #2983, opened 2024-08-30
Add join VNET call for every AZR NC unpublish call

smittal22:
**Reason for Change**:
For AZR scenarios, if NMAgent is restarted, it loses state and does not know what VNETs to subscribe to. As it no longer has VNET state, successive put nc or delete nc calls would fail. We, however, always call join VNET before put nc call. This PR adds such a fdlow for delete AZR nc calls as well.

PS: We don't need this flow for regular NCs, as NMAgent there uses NMA swift module to rehydrate VNETs if state is lost.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -1039,6 +1039,10 @@ func (service *HTTPRestService) unpublishNetworkContainer(w http.ResponseWriter,
 		}
 	}
 
+	/* For AZR scenarios, if NMAgent is restarted, it loses state and does not know what VNETs to subscribe to.
+	As it no longer has VNET state, delete nc calls would fail. We need to add join VNET call for all AZR
+	nc unpublish calls just like publish nc calls.
+	*/
 	if unpublishBody.AZREnabled || !service.isNetworkJoined(req.NetworkID) {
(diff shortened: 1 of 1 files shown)


cns/restserver/api.go
```diff
@@ -1029,7 +1030,16 @@ func (service *HTTPRestService) unpublishNetworkContainer(w http.ResponseWriter,
...
-	if !service.isNetworkJoined(req.NetworkID) {
+	var unpublishBody nmagent.DeleteContainerRequest
+	if req.DeleteNetworkContainerRequestBody != nil {
+		err = json.Unmarshal(req.DeleteNetworkContainerRequestBody, &unpublishBody)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("could not unmarshal delete network container body: %v", err), http.StatusBadRequest)
+			return
+		}
+	}
+
```
ashvindeodhar, line 1041: can you add more comments indicating why we need to join for AZR NCs?

smittal22: added a comment below

ashvindeodhar (approved):

nairashu (approved):

PR #2985, opened 2024-09-03
ci: ACN PR Pipeline Security Feature Branch

sheylatrudo:
**Reason for Change**:
Merges change for a PR branch for a OneBranch based pipeline build for Azure Container Networking PR pull request builds. This pipeline will live on the feature branch until the changes have been verified as in-parity, and secure.

**Issue Fixed**:
Fixes Component Governance alerts for migrating our github repos to use organization compliant builds.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/cni/singletenancy/cniv1-template.yaml
+++ b/.pipelines/cni/singletenancy/cniv1-template.yaml
@@ -39,14 +39,16 @@ stages:
           location: $(LOCATION_AMD64)
         ${{ if eq(parameters.arch, 'arm64') }}:
           location: $(LOCATION_ARM64)
-      pool:
-        name: $(BUILD_POOL_NAME_DEFAULT)
       dependsOn:
         - setup
         - build_images
       displayName: "Create Cluster - ${{ parameters.clusterName }}"
       jobs:
         - job: create_aks_cluster_with_${{ parameters.name }}
+          pool:
+            isCustom: true
+            type: linux
+            name: $(BUILD_POOL_NAME_DEFAULT)
           steps:
             - template: ../load-test-templates/create-cluster-template.yaml
               parameters:
@@ -72,14 +74,16 @@ stages:
           location: $(LOCATION_AMD64)
         ${{ if eq(parameters.arch, 'arm64') }}:
           location: $(LOCATION_ARM64)
-      pool:
-        name: $(BUILD_POOL_NAME_DEFAULT)
       dependsOn:
         - setup
         - build_images
       displayName: "Create Cluster - ${{ parameters.clusterName }}"
       jobs:
         - job: create_aks_cluster_with_${{ parameters.name }}
+          pool:
+            isCustom: true
+            type: linux
+            name: $(BUILD_POOL_NAME_DEFAULT)
           steps:
             - template: ../load-test-templates/create-cluster-template.yaml
               parameters:
@@ -106,8 +110,6 @@ stages:
         location: $(LOCATION_AMD64)
       ${{ if eq(parameters.arch, 'arm64') }}:
         location: $(LOCATION_ARM64)
-    pool:
-      name: $(BUILD_POOL_NAME_DEFAULT)
     dependsOn:
       - create_${{ parameters.name }}
       - publish
@@ -115,6 +117,10 @@ stages:
     displayName: "CNIv1 Test - ${{ parameters.clusterName }}"
     jobs:
       - job: update_cni
+        pool:
+          isCustom: true
+          type: linux
+          name: $(BUILD_POOL_NAME_DEFAULT)
         steps:
           - task: AzureCLI@2
             inputs:
@@ -158,6 +164,10 @@ stages:
         condition: and( and( not(canceled()), not(failed()) ), or( contains(variables.CONTROL_SCENARIO, 'scaleTest') , contains(variables.CONTROL_SCENARIO, 'all') ) )
         displayName: "Scale Test"
         dependsOn: npm_k8se2e
+        pool:
+          isCustom: true
+          type: linux
+          name: $(BUILD_POOL_NAME_DEFAULT)
         steps:
           - template: ../load-test-templates/pod-deployment-template.yaml
             parameters:
@@ -176,6 +186,10 @@ stages:
         condition: and( and( not(canceled()), not(failed()) ), or( contains(variables.CONTROL_SCENARIO, 'restartNode') , contains(variables.CONTROL_SCENARIO, 'all') ) )
         displayName: "Restart Test"
         dependsOn: deploy_pods
+        pool:
+          isCustom: true
+          type: linux
+          name: $(BUILD_POOL_NAME_DEFAULT)
         steps:
           - template: ../load-test-templates/restart-node-template.yaml
             parameters:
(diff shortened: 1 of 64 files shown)


.pipelines/containers/container-template.yaml
```diff
@@ -15,12 +15,19 @@ steps:
       az acr login -n $(ACR)
 
 - script: |
+    ls -la
```
sheylatrudo, line 18: Remove.

.pipelines/cni/singletenancy/cniv1-template.yaml
sheylatrudo, line 1: Do we need to update this file for ACN PR runs?

.pipelines/cni/singletenancy/cniv2-template.yaml
sheylatrudo, line 1: Do we need to update this file for ACN PR runs?

.pipelines/containers/manifest-template.yaml
```diff
@@ -51,8 +51,8 @@ steps:
     inputs:
       BuildDropPath: "$(Build.ArtifactStagingDirectory)"
 
-  - task: PublishBuildArtifacts@1
```
sheylatrudo, line 54: Can we verify that this is still being published in the new pipeline?

.pipelines/npm/npm-cni-integration-test.yaml
sheylatrudo, line 1: Do we need to update this file for ACN PR runs?

.pipelines/templates/log-template.yaml
```diff
@@ -265,11 +268,12 @@ steps:
             fi
           done
         fi
+      workingDirectory: $(ACN_DIR)
       displayName: Windows Logs
       condition: always()
 
-  - publish: $(System.DefaultWorkingDirectory)/${{ parameters.clusterName }}_${{ parameters.logType }}_Attempt_#$(System.StageAttempt)
```
sheylatrudo, line 271: Can we verify that the new pipeline creates these artifacts

chengch801214 (commented):
It looks like a lot of these files are renames, any reason behind that? And are the old files going to be deleted?

sheylatrudo:
@chengch801214 - Yes, this is to prevent collisions with internal pipeline changes to this repo while the feature is being testing. Currently there are multiple workstreams touching these files and we cannot block the higher priority work. Eventually the old files will be deleted after we have confirmed that we have parity across the two versions _and_ we have switched over to using the new pipeline for PR validation.

.pipelines/trigger.yaml
```diff
@@ -0,0 +1,51 @@
+pr:
```
jpayne3506, line 1: What is the plan to ensure that this hooks back into PRs?

.pipelines/trigger.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - ".hooks"
+      - ".vscode"
+      - ".github"
+      - docs
+
+
+trigger:
+  branches:
+    include:
+      - gh-readonly-queue/master/*
```
jpayne3506, line 18: Until we have hooks back into the PR we might want to turn off this trigger as all it will be doing is raising COGS.

.pipelines/trigger.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - "*"
+
+
+resources:
+  repositories:
+  - repository: azure-container-networking
+    type: github
+    name: Azure/azure-container-networking
+    endpoint: 'Azure-ACN RO Service Connection'
+    ref: refs/heads/feature/ob-onboard-0
```
jpayne3506, line 30: Does this need to be changed before we merge this to master?

.pipelines/trigger.yaml
```diff
@@ -0,0 +1,51 @@
...
+  vmImage: ubuntu-latest
+
+
+extends:
+  template: /.pipelines/template.trigger.jobs.yaml@azure-container-networking
+  parameters:
+    mainRepoRef: $(REPO_REF)
+    mainRepoCommit: $(REPO_COMMIT)
+    mainRepoName: $(REPO_NAME)
+    mainRepoType: $(REPO_TYPE)
```
jpayne3506, line 51: Doesn't referencing them this way change it from compile time to run time? I feel like that could have negative impacts later on in development of this feature.

.pipelines/singletenancy/dualstack-overlay/dualstackoverlay-e2e.jobs.yaml
```diff
@@ -0,0 +1,201 @@
...
+          clusterName: ${{ parameters.clusterName }}-$(commitID)
+          os: linux
+          dependsOn: ${{ parameters.name }}_windows
+          dualstack: true
+          dns: true
+          portforward: true
+          hostport: true
+          service: true
+
+      - template: ../../cni/k8s-e2e/k8s-e2e.jobs.yaml@ACNTools
```
jpayne3506, line 141: There is an ongoing bug with k8s 1.30 when using this testing suite. #3011 is addressing it for now within the current files on master. Expect this to fail similarly.

jpayne3506 (commented):
Seeing as we are creating new files entirely now, maybe we should move them to `.pipelines/devops/*` That way we know there is a clear separation between OB and standard yaml pipelines.

.pipelines/singletenancy/dualstack-overlay/dualstackoverlay-e2e.jobs.yaml
sheylatrudo: Sounds good. I'll keep an eye on it.

.pipelines/trigger.yaml
sheylatrudo: Fine for now until we are past the validation/master parity stage. Then we'll update this ref to master.

.pipelines/trigger.yaml
sheylatrudo: We just need to add a check like we have for ACN. We'll add that after we've merged the second round of changes.

.pipelines/trigger.yaml
sheylatrudo: We want this to run so we can validate parity. We can disable the pipeline if COGS becomes a problem.

sheylatrudo:
> Seeing as we are creating new files entirely now, maybe we should move them to `.pipelines/devops/*` That way we know there is a clear separation between OB and standard yaml pipelines.

@jpayne3506 The old files will eventually be deleted after the whole cycle has complete (merge this change -> validation/attain master parity -> merge changes -> add PR check to repo -> run checks side by side for a week -> deprecate old ACN PR Pipeline -> delete replaced files -> merge). It is much better if people have the visibility on these files.

.pipelines/trigger.yaml
sheylatrudo: Although...I can comment these out until we reach the next phase.

.pipelines/trigger.yaml
sheylatrudo: No. I don't think so. I think this is different because it is using a variable from a resource type, which can guarantee that a variable exists or not.

sheylatrudo: https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/resources-repositories-repository?view=azure-pipelines#variables

jpayne3506: > In each run, the metadata for a repository resource is available to all jobs in the form of runtime variables.

Ah so it is already a runtime variable.

jpayne3506 (approved):

PR #2986, opened 2024-09-03
[fix] ci: cluster deletion for swiftv2

jpayne3506:
**Reason for Change**:

Fixes the delete cluster stage by targeting the different subs. Additionally, amplifies the resource group deletion by targeting the correct region based on the env var.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

Needs to be backported to any release train that has multiple regions or subs in its E2E

### DIFF
--- a/.pipelines/pipeline.yaml
+++ b/.pipelines/pipeline.yaml
@@ -426,7 +426,7 @@ stages:
       variables:
         commitID: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.commitID'] ]
       jobs:
-        - job: delete
+        - job: delete_build
           displayName: Delete Cluster
           pool:
             name: "$(BUILD_POOL_NAME_DEFAULT)"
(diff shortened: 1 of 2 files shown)


.pipelines/pipeline.yaml
```diff
@@ -426,7 +426,7 @@ stages:
       variables:
         commitID: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.commitID'] ]
       jobs:
-        - job: delete
+        - job: delete_build
```
sheylatrudo, line 429: Does updating this name have any other side-effects? Is this stage references anywhere else?

jpayne3506: This is the last job in the last stage. It does not get referenced anywhere. The main purpose for adding the `_build` and `_test` is so that AZP knows that these are different jobs.

sheylatrudo (approved):
lgtm

PR #2989, opened 2024-09-04
fix: add dhcp discover packet to linux swiftv2 to ensure mappings exist for dns

QxBytes:
**Reason for Change**:

In linux swiftv2, nslookup using the wireserver may fail in certain scenarios. The theory is that this is because the host does not have mappings/state for the dns on the secondary/delegated nics (eth1). To ensure that state is created, we send a dhcp discover packet through the delegated nic (eth1) so the mappings are created. Any offers returned are ignored.

**Issue Fixed**:

See above.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
We assume that epInfo.IfName contains the actual interface name of the nic! The dhcp discover packet MUST get a successful offer or the add call will fail. If in swiftv2 we have multiple nics, we send the dhcp request for each one and all must succeed or the add call will fail. **Ensure the ifName is in epInfo.ifName**

Should affect Swiftv2 Linux only.
 
Reviewer(s):
- Code includes goroutines/synchronization (potential data races?)
- Verify indexing is correct/no possibility of panics/accessing out of bounds
- Code deals with sockets (verify all sockets always properly closed)

Tested:
- Swiftv2 linux create mt pod 1.6.5 cns (we do get a reply from the dhcp server), logs corroborate
- Attempt request with wrong mac address, we time out and error as expected
- Attempt request with wrong ifname, we time out and error as expected
- Context deadline error as expected if set context deadline to 0 sec
- Repro'd issue, then applied CNI fix, then recreated pod to issue dhcp request, and then I can contact dns via secondary nic as expected (previously I couldn't)


dhcp/dhcp.go
```diff
@@ -0,0 +1,365 @@
...
+	packetBytes, err := MakeRawUDPPacket(packet, *raddr, *laddr)
+	if err != nil {
+		return errors.Wrap(err, "error making raw udp packet")
+	}
+
+	// Create sockaddr_ll structure for sending the packet
+	remoteAddr := unix.SockaddrInet4{Port: laddr.Port, Addr: destination}
+
+	// Send the packet using the raw socket
+	return errors.Wrap(unix.Sendto(fd, packetBytes, 0, &remoteAddr), "failed unix send to")
```
timraymond, line 249: Passing around ints for file descriptors is a bit too C-like. The Go approach would be to create an `io.Writer` that implements `Write()` using `unix.Sendto`. That writer then gets passed around (as an `io.Writer`). That would also extract the linux-specific parts to the Writer implementation and leave the packet manufacturing platform-independent.

That might help with testing too, since `io.Writers` can be as simple as `bytes.Buffer{}`.

dhcp/dhcp.go
```diff
@@ -0,0 +1,365 @@
...
+		// a nil error means a reply was found for this txid
+		recvErrors <- nil
+	}(recvErrors)
+
+	select {
+	case err := <-recvErrors:
+		if err != nil {
+			return errors.Wrap(err, "error during receiving")
+		}
+	case <-time.After(DefaultTimeout):
```
timraymond, line 322: This leaks a goroutine if there's an error. Use `time.Ticker` and defer `time.(*Ticker).Stop()`,

dhcp/dhcp.go
```diff
@@ -0,0 +1,446 @@
...
+		return fd, errors.Wrap(err, "dhcp raw socket creation failure")
+	}
+	err = unix.SetsockoptInt(fd, unix.SOL_SOCKET, unix.SO_REUSEADDR, 1)
+	if err != nil {
+		return fd, errors.Wrap(err, "dhcp failed to set raw sockopt")
+	}
+	err = unix.SetsockoptInt(fd, unix.IPPROTO_IP, unix.IP_HDRINCL, 1)
+	if err != nil {
+		return fd, errors.Wrap(err, "dhcp failed to set second raw sockopt")
+	}
```
tamilmani1989, line 191: why it requires to set these options?

dhcp/dhcp.go
```diff
@@ -0,0 +1,446 @@
...
+	ret = append(ret, udp...)
+	ret = append(ret, payload...)
+	return ret, nil
+}
+
+// Receive DHCP response packet using reader
+func (c *DHCP) receiveDHCPResponse(ctx context.Context, reader io.ReadCloser, xid TransactionID) error {
+	recvErrors := make(chan error, 1)
+	go func(errs chan<- error) {
```
tamilmani1989, line 301: why spawning new go routine?

QxBytes: Recvfrom is a blocking call, so if something goes wrong with its timeout it won't return. Additionally, the timeout on the socket (on the Read(...)) call is how long until the socket times out and gives an error, but it **won't** error if we do get some data within the time out period. If we get some data (**even if it is not the packet we are looking for**, like wrong txid, wrong response opcode etc.) then we continue in the for loop. We then call recvfrom again which will _reset_ the timeout period-- without the secondary timeout at the bottom of the function, I believe we could stay stuck in the for loop as long as we receive packets.

dhcp/dhcp.go
QxBytes: REUSEADDR is to avoid "address already in use" errors.

IP_HDR_INCL is because I believe later on when we write to this broadcast socket, our packet _already_ contains the header (we create it with MakeRawUDPPacket). Omitting this I believe would mean another header would be added to the packet, which we don't want.

tamilmani1989: i dont think `REUSEADDR` is required. can we test without it?

dhcp/dhcp.go
tamilmani1989: can we also add comment stating that?

dhcp/dhcp.go
```diff
@@ -0,0 +1,461 @@
...
+	logger *zap.Logger
+}
+
+func New(logger *zap.Logger) *DHCP {
+	return &DHCP{
+		logger: logger,
+	}
+}
+
+type Socket struct {
```
timraymond, line 64: Looking at this Socket now, it's actually most of the way to a `net.Conn` (probably need `net.PacketConn` actually). You probably considered it, but I'd like to hear the reasoning: was there an implementation of that in the standard library that fit our needs here? If not, I'm fine with our own implementation... otherwise I'd rather use the standard library.

tamilmani1989: no we dont have dhcp client in standard library. thats why we building our own

tamilmani1989 (approved):

dhcp/dhcp.go
QxBytes: I think there might be trouble with this approach because we write using one type of socket with specific settings but then read using a different type of socket with different settings-- I needed control over the socket creation itself

tamilmani1989 (approved):

dhcp/dhcp.go
timraymond: Why didn't we use something like https://github.com/insomniacslk/dhcp ? It's actively maintained, used by many large projects? I get that "a little copying is better than a little dependency", but in this case where we're most of the way to implementing a full client, it feels like we should just use one. Otherwise it feels like NIH syndrome.

QxBytes: I initially went with this approach, but I think we didn't want to take on the dependencies/possible license issues. Feel free to follow up with Tamilmani on this.

PR #2992, opened 2024-09-05
Adding NpmLite Feature

rejain456:
**Reason for Change**:

This pr adds code changes for a new feature called NPM-Lite which is a subset of NPM.
NPM Lite provides Singularity team with a “light” NPM which addresses: 

- Heavy load on API Server: NPM Lite must support ~4K Nodes (NPM currently supports ~250 Nodes). 
- Windows NPM performance issues caused by HNS SetPolicies. 

**Issue Fixed**:

To make npm more scalable for the singularity team, npm lite consists of 2 main limitations:
1. Only accepts **CIDR-based** network policies
2. Enables the daemon set to watch for pods only under its node

This pr adds code to cover both the points above

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


npm/cmd/start.go
```diff
@@ -115,8 +117,20 @@ func start(config npmconfig.Config, flags npmconfig.Flags) error {
...
+	factory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
+	// npm-lite -> daemon set will listen to pods only in its own node
+	if config.Toggles.EnableNPMLite {
+		nodeName := os.Getenv("NODE_NAME")
+		factory = informers.NewSharedInformerFactoryWithOptions(
+			clientset,
+			resyncPeriod,
+			informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+				// Use field selector to filter pods based on their assigned node
+				options.FieldSelector = "spec.nodeName=" + nodeName
```
huntergregory, line 130: Is there a way to only modify the pod informer? The current change might not impact ns and netpol informers anyways, but perhaps it's safer to scope changes to only the pod informer?

npm/cmd/start.go
```diff
@@ -115,8 +117,20 @@ func start(config npmconfig.Config, flags npmconfig.Flags) error {
 	factor := rand.Float64() + 1 //nolint
 	resyncPeriod := time.Duration(float64(minResyncPeriod.Nanoseconds()) * factor)
 	klog.Infof("Resync period for NPM pod is set to %d.", int(resyncPeriod/time.Minute))
-	factory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
 
+	factory := informers.NewSharedInformerFactory(clientset, resyncPeriod)
+	// npm-lite -> daemon set will listen to pods only in its own node
+	if config.Toggles.EnableNPMLite {
+		nodeName := os.Getenv("NODE_NAME")
```
huntergregory, line 124: nit: I think we get the node name down below. Could probably share the same variable

huntergregory (commented):
CIDR check looks good. Wondering if we can/should scope the informer change to pod controller

npm/cmd/start.go
rejain456: yes changed it, thanks

npm/cmd/start.go
rejain456: I'll look more into it and check

rejain456: @huntergregory, 
I changed the code so that the informer change is only scoped to pod controller. Please let me know if that looks better

npm/npm.go
```diff
@@ -58,6 +58,7 @@ type NetworkPolicyManager struct {
 // NewNetworkPolicyManager creates a NetworkPolicyManager
 func NewNetworkPolicyManager(config npmconfig.Config,
 	informerFactory informers.SharedInformerFactory,
+	PodFactory informers.SharedInformerFactory,
```
huntergregory, line 61: nit: golang convention is to lowercase parameters. In general, lowercase means unexported/private to the package, and uppercase means the symbol is exported/accessible to other pkgs

npm/cmd/start.go
huntergregory: Looks good! Thanks

npm/npm.go
rejain456: Thanks for pointing that out! Fixed it

rejain456:
> CIDR check looks good. Wondering if we can/should scope the informer change to pod controller

sheylatrudo:
azp run NPM Conformance Tests

PR #2949, opened 2024-08-22
try and use powershell less (especially at startup) because 1) it eat…

paulgmiller:
Shellig out to powershell while sometimes convenient has a couple of issues 
1) large bump of memory inside your contaienr as the whole dotnetframework comes to ife
2) dependency on customer not muckign with host powrshell since we run in host namespace? (could use powershell core in our contaienr?)
3) parsing std out isnt always the best. 

**Reason for Change**:

Custoemr reporting ooms on azure-windows-cns. 

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] relevant PR labels added

**Notes**:

PR #2315, opened 2023-10-20
fix: skip setting SdnRemoteArpMacAddress when hns is not enabled

ZetaoZhuang:
**Reason for Change**:

fix: skip setting SdnRemoteArpMacAddress when hns is not enabled
**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [x] adds unit tests
- [ ] relevant PR labels added

**Notes**:

PR #1306, opened 2022-03-30
feat: Set remote ARP MAC address when CNS is running in CRD mode

jaer-tsun:
**Reason for Change**:

Set remote ARP MAC address when CNS is running in CRD mode

**Issue Fixed**:

Unblocks AKS-Swift on windows along with #1076 

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

PR #2993, opened 2024-09-05
fix: remove PowerShell from Windows registry interactions

rbtr:
**Reason for Change**:
PowerShell is heavy (+30mb of memory footprint per invocation). We don't need to use it to interact with the Windows Registry (or ServiceManager) - the Go stdlib has good native bindings that should use instead.

**Notes**:
This functionality was introduced in #1306. In #2315, it was changed to conditionally set the reg key value only if the HNS path exists. Why? It seems simpler and harmless to always try to set the reg key, so that's what I have done. Unclear if this is problematic - cc @ZetaoZhuang 

Supersedes and closes #2974 and #2961 (and #2949)

### DIFF
--- a/platform/os_windows.go
+++ b/platform/os_windows.go
@@ -20,6 +20,9 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
 )
 
 const (
@@ -61,23 +64,9 @@ const (
 	// for vlan tagged arp requests
 	SDNRemoteArpMacAddress = "12-34-56-78-9a-bc"
 
-	// Command to get SDNRemoteArpMacAddress registry key
-	GetSdnRemoteArpMacAddressCommand = "(Get-ItemProperty " +
-		"-Path HKLM:\\SYSTEM\\CurrentControlSet\\Services\\hns\\State -Name SDNRemoteArpMacAddress).SDNRemoteArpMacAddress"
-
-	// Command to set SDNRemoteArpMacAddress registry key
-	SetSdnRemoteArpMacAddressCommand = "Set-ItemProperty " +
-		"-Path HKLM:\\SYSTEM\\CurrentControlSet\\Services\\hns\\State -Name SDNRemoteArpMacAddress -Value \"12-34-56-78-9a-bc\""
-
-	// Command to check if system has hns state path or not
-	CheckIfHNSStatePathExistsCommand = "Test-Path " +
-		"-Path HKLM:\\SYSTEM\\CurrentControlSet\\Services\\hns\\State"
-
 	// Command to fetch netadapter and pnp id
+	// TODO: can we replace this (and things in endpoint_windows) with other utils from "golang.org/x/sys/windows"?
 	GetMacAddressVFPPnpIDMapping = "Get-NetAdapter | Select-Object MacAddress, PnpDeviceID| Format-Table -HideTableHeaders"
-
-	// Command to restart HNS service
-	RestartHnsServiceCommand = "Restart-Service -Name hns"
 
 	// Interval between successive checks for mellanox adapter's PriorityVLANTag value
 	defaultMellanoxMonitorInterval = 30 * time.Second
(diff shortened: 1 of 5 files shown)


ZetaoZhuang:
In https://github.com/Azure/azure-container-networking/pull/2315, it was changed to conditionally because there was a livesite that our customer got error during the cns start up to set SdnRemoteArpMacAddress. somehow hns was not enabled on their machines so we need to skip this for them.

rbtr:
@ZetaoZhuang is it okay to try to set it? Or can we not even try to set it if HNS is not enabled?

ZetaoZhuang:
> @ZetaoZhuang is it okay to try to set it? Or can we not even try to set it if HNS is not enabled?

from that livesite, if we set it when hns is not enabled, we will get error complaining about something like "hns state path does not exist" or so. that is the reason why we made it conditional. I believe we could set it if it does not complain anymore.

platform/os_windows.go
```diff
@@ -257,43 +248,73 @@ func (p *execClient) ExecutePowershellCommandWithContext(ctx context.Context, co
...
-func SetSdnRemoteArpMacAddress(execClient ExecClient) error {
-	exists, err := execClient.ExecutePowershellCommand(CheckIfHNSStatePathExistsCommand)
+func SetSdnRemoteArpMacAddress() error {
+	// open the registry key
+	k, err := registry.OpenKey(registry.LOCAL_MACHINE, "SYSTEM\\CurrentControlSet\\Services\\hns\\State", registry.READ|registry.SET_VALUE)
 	if err != nil {
-		errMsg := fmt.Sprintf("Failed to check the existent of hns state path due to error %s", err.Error())
-		log.Printf(errMsg)
-		return errors.Errorf(errMsg)
+		if !errors.Is(err, registry.ErrNotExist) {
```
ZetaoZhuang, line 255: should we return nil when err is registry.ErrNotExist?

rbtr: yes I should have flipped that conditional when I pushed the last update. fixed.

ZetaoZhuang (dismissed):
lgtm

platform/os_windows.go
```diff
@@ -257,43 +248,72 @@ func (p *execClient) ExecutePowershellCommandWithContext(ctx context.Context, co
...
+		return errors.Wrap(err, "could not connect to service manager")
+	}
+	defer m.Disconnect() //nolint:errcheck // ignore error
+	// open the HNS service
+	service, err := m.OpenService("hns")
+	if err != nil {
+		return errors.Wrap(err, "could not access service")
+	}
+	defer service.Close()
+	return errors.Wrap(restartService(service), "could not restart service")
```
ramiro-gamarra, line 277: nit: While I understand that this will only actually return an error if `restartService` returns an error itself, it makes this harder to read at first unless you are familiar with `github.com/pkg/errors` (plus more complicated when you want to debug and step through code). I've been asking people to avoid this in networking aquarius for same readability issues. Good ol' 
```go
if err := restartService(service); err != nil {
    return errors.Wrap(err, "could not restart service")
}

return nil
```

reads clearly IMO.

platform/os_windows.go
```diff
@@ -257,43 +248,72 @@ func (p *execClient) ExecutePowershellCommandWithContext(ctx context.Context, co
...
+	// was "Set-ItemProperty -Path HKLM:\\SYSTEM\\CurrentControlSet\\Services\\hns\\State -Name SDNRemoteArpMacAddress -Value \"12-34-56-78-9a-bc\""
+	if err := k.SetStringValue("SDNRemoteArpMacAddress", SDNRemoteArpMacAddress); err != nil {
+		return errors.Wrap(err, "could not set registry key")
 	}
-
+	log.Printf("[Azure CNS] SDNRemoteArpMacAddress regKey set successfully. Restarting hns service.")
 	return nil
 }
 
+type serv interface {
```
ramiro-gamarra, line 294: I didn't notice any other implementations of this, is the indirection needed at the moment?

platform/os_windows.go
```diff
@@ -257,43 +248,72 @@ func (p *execClient) ExecutePowershellCommandWithContext(ctx context.Context, co
...
+	if err != nil {
+		return errors.Wrap(err, "could not stop service")
+	}
+	// Wait for the service to stop
+	for status, err := s.Query(); status.State != svc.Stopped; status, err = s.Query() {
+		if err != nil {
+			return errors.Wrap(err, "could not query service status")
+		}
+		time.Sleep(500 * time.Millisecond) //nolint:gomnd // 500ms
+	}
```
ramiro-gamarra, line 312: Potentially infinite loop? Probably better to add a timeout via context and use tickers.

platform/os_windows.go
```diff
@@ -61,24 +64,12 @@ const (
...
-		"-Path HKLM:\\SYSTEM\\CurrentControlSet\\Services\\hns\\State -Name SDNRemoteArpMacAddress -Value \"12-34-56-78-9a-bc\""
-
-	// Command to check if system has hns state path or not
-	CheckIfHNSStatePathExistsCommand = "Test-Path " +
-		"-Path HKLM:\\SYSTEM\\CurrentControlSet\\Services\\hns\\State"
-
 	// Command to fetch netadapter and pnp id
+	// TODO can we replace this (and things in endpoint_windows) with "golang.org/x/sys/windows"
+	// var adapterInfo windows.IpAdapterInfo
+	// var bufferSize uint32 = uint32(unsafe.Sizeof(adapterInfo))
```
ramiro-gamarra, line 70: is this dead code or example code? either way seems like this todo could be a bit clearer

platform/os_windows.go
```diff
@@ -257,43 +248,72 @@ func (p *execClient) ExecutePowershellCommandWithContext(ctx context.Context, co
...
+	}
+	// Wait for the service to stop
+	for status, err := s.Query(); status.State != svc.Stopped; status, err = s.Query() {
+		if err != nil {
+			return errors.Wrap(err, "could not query service status")
+		}
+		time.Sleep(500 * time.Millisecond) //nolint:gomnd // 500ms
+	}
+	// Start the service again
+	return errors.Wrap(s.Start(), "could not start service")
```
ramiro-gamarra, line 314: same feedback as above

platform/os_windows.go
rbtr: fixed

platform/os_windows.go
rbtr: removed

platform/os_windows.go
rbtr: rewrote with ticker while keeping the do-while semantics

platform/os_windows.go
rbtr: clarified

platform/os_windows.go
rbtr: fixed

platform/os_windows.go
```diff
@@ -257,40 +246,79 @@ func (p *execClient) ExecutePowershellCommandWithContext(ctx context.Context, co
...
+}
 
-		sdnRemoteArpMacAddressSet = true
+func setSDNRemoteARPMACAddress(k key) error {
+	// Set the reg key
+	// was "Set-ItemProperty -Path HKLM:\\SYSTEM\\CurrentControlSet\\Services\\hns\\State -Name SDNRemoteArpMacAddress -Value \"12-34-56-78-9a-bc\""
+	if err := k.SetStringValue("SDNRemoteArpMacAddress", SDNRemoteArpMacAddress); err != nil {
+		return errors.Wrap(err, "could not set registry key")
 	}
+	log.Printf("[Azure CNS] SDNRemoteArpMacAddress regKey set successfully. Restarting hns service.")
```
nairashu, line 291: nit: Do we want to log this here as we are not really restarting the hns service here but in the caller of this method

platform/os_windows.go
```diff
@@ -257,40 +246,79 @@ func (p *execClient) ExecutePowershellCommandWithContext(ctx context.Context, co
...
+		}
+		if status.State == svc.Stopped {
+			break
+		}
+		select {
+		case <-ctx.Done():
+			return errors.New("context cancelled")
+		case <-ticker.C:
+		}
+	}
```
nairashu, line 317: You have not started the ticker. Should we add the ticker to the loop to start the ticker and make the loop finite
```suggestion
	for range ticker.C { // hacky cancellable do-while
		status, err := s.Query()
		if err != nil {
			return errors.Wrap(err, "could not query service status")
		}
		if status.State == svc.Stopped {
			break
		}
		select {
		case <-ctx.Done():
			return errors.New("context cancelled")
		}
	}
```

platform/os_windows.go
```diff
@@ -364,6 +392,7 @@ func GetProcessNameByID(pidstr string) (string, error) {
 	pidstr = strings.Trim(pidstr, "\r\n")
 	cmd := fmt.Sprintf("Get-Process -Id %s|Format-List", pidstr)
 	p := NewExecClient(nil)
+	// TODO not riemovign this because it seems to only be called in test?
```
nairashu, line 395: nit: removing spell

platform/os_windows_test.go
```diff
@@ -115,39 +127,11 @@ func TestExecuteCommandError(t *testing.T) {
 	require.ErrorIs(t, err, exec.ErrNotFound)
 }
 
-func TestSetSdnRemoteArpMacAddress_hnsNotEnabled(t *testing.T) {
```
nairashu, line 118: Should we add a test case for hns not enabled as well

platform/os_windows.go
rbtr: I don't know what you mean by "have not started the ticker". [Tickers](https://gobyexample.com/tickers) are immediately ready to use on initialization.

NACK on your suggestion. You have changed the semantics from `do-while`, to simply `while`. I want to test the condition first, then wait, then repeat. As a `while`, we wait first, then test the condition, then repeat.

nairashu: Good to NACK the suggestion the Copilot thread I read suggested suggested the following:
```
package main

import (
    "fmt"
    "time"
)

func main() {
    fmt.Println("Start Ticker Example")
    // Define an interval and create a new Ticker
    interval := time.Duration(2) * time.Second
    ticker := time.NewTicker(interval)
    defer ticker.Stop()

    // Start the ticker by constructing a loop
    i := 0
    for range ticker.C {
        i++
        countFuncCall(i)
        if i >= 5 {
            break
        }
    }
}

// Define the function to be called
func countFuncCall(i int) {
    fmt.Println("Function is called:", i, "times")
}
```
So I assumed you have to start the Ticker. Thanks for the clarification. Resolving this.

platform/os_windows_test.go
```diff
@@ -115,39 +127,11 @@ func TestExecuteCommandError(t *testing.T) {
...
-			return "True", nil
-		}
-		return "", nil
-	})
-	err := SetSdnRemoteArpMacAddress(mockExecClient)
-	assert.NoError(t, err)
-	assert.Equal(t, true, sdnRemoteArpMacAddressSet)
-	// reset sdnRemoteArpMacAddressSet
-	sdnRemoteArpMacAddressSet = false
+func TestSetSDNRemoteARPMACAddress(t *testing.T) {
```
rjdenney, line 130: Do we no longer need a test for SetSdnRemoteArpMacAddress or do we just need to test for setting the key value now?

nairashu (dismissed):

rjdenney (dismissed):

rjdenney (approved):
lgtm

PR #3005, opened 2024-09-10
feat: [NPM] Adding Npm Lite 

rejain456:
Reason for Change:

This pr adds code changes for a new feature called NPM-Lite which is a subset of NPM.
NPM Lite provides Singularity team with a “light” NPM which addresses:

- Heavy load on API Server: NPM Lite must support ~4K Nodes (NPM currently supports ~250 Nodes).
- Windows NPM performance issues caused by HNS SetPolicies.
- 
- Issue Fixed:

npm lite consists of 2 main limitations:

- Only accepts CIDR-based network policies
- Enables the daemon set to watch for pods only under its node

This pr adds code to cover both the points above

Testing
image

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


npm/pkg/controlplane/translation/translatePolicy_test.go
```diff
@@ -2919,3 +2919,114 @@ func TestEgressPolicy(t *testing.T) {
...
+	tests := []struct {
+		name           string
+		targetSelector *metav1.LabelSelector
+		peersFrom      []networkingv1.NetworkPolicyPeer
+		peersTo        []networkingv1.NetworkPolicyPeer
+		npmLiteEnabled bool
+		wantErr        bool
+	}{
+		{
+			name:           "peer nameSpaceSelector and ipblock in ingress rules",
```
huntergregory, line 2938: Thanks for writing tests. Could we add these cases as well?
- peer with both ipblock and selector
- no peers

npm/pkg/controlplane/translation/translatePolicy.go
```diff
@@ -34,6 +35,8 @@ var (
 	)
 	// ErrUnsupportedIPAddress is returned when an unsupported IP address, such as IPV6, is used
 	ErrUnsupportedIPAddress = errors.New("unsupported IP address")
+	// ErrUnsupportedNonCIDR is returned when non-CIDR blocks (i.e pod selectors or namespace selectors) are passed with a NPM Lite configuration
+	ErrUnsupportedNonCIDR = errors.New("unsupported Non-CIDR block for NPM Lite")
```
huntergregory, line 39: we're actually allowing default allow and default deny policies as well (in `TranslatePolicy()`). Could we modify this error description to express that?

huntergregory (changes requested):
Realizing an edge case, and we'll have to be careful with reading in the config.

Also for testing, could you add to the PR description steps or screenshots for how you verified that a non-CIDR policy throws an error, and how you verified that applying a CIDR policy correctly blocks / allows traffic?

npm/pkg/controlplane/translation/translatePolicy.go
```diff
@@ -403,11 +408,24 @@ func translateRule(npmNetPol *policies.NPMNetworkPolicy, netPolName string, dire
 					return err
 				}
 			}
+
+			// if npm lite is configured, check network policy only consists of CIDR blocks
+			err := NpmLiteValidPolicy(peer, NpmLiteEnabled)
```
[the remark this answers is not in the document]
huntergregory: "No peers" actually corresponds to "default deny all" or "default allow all" I believe?
https://github.com/Azure/azure-container-networking/pull/3005#discussion_r1755266123

npm/pkg/controlplane/translation/translatePolicy_test.go
huntergregory: This is 3 peers:
```
[]networkingv1.NetworkPolicyPeer{
				{
					NamespaceSelector: &metav1.LabelSelector{
						MatchLabels: map[string]string{
							"peer-nsselector-kay": "peer-nsselector-value",
						},
					},
				},
				{
					IPBlock: &networkingv1.IPBlock{
						CIDR:   "172.17.0.0/16",
						Except: []string{"172.17.1.0/24", "172.17.2.0/24"},
					},
				},
				{
					IPBlock: &networkingv1.IPBlock{
						CIDR: "172.17.0.0/16",
					},
				},
			},
```

Each peer corresponds to an element of the YAML's list (more intuitively, a "section" denoted by a `-`) in a NetworkPolicy spec, where each element is OR'd together. Within a peer, the rules are AND'd. For more details, see "namespaceSelector and podSelector" under https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors

> single peer with both ipblock and selector

It sounds like this is actually impossible from the docs though. It is not one of the 4 kinds of to/from selectors. Do you mind verifying that it's impossible to create such a NetPol peer with ipblock and pod selector?

npm/pkg/controlplane/translation/translatePolicy.go
```diff
@@ -387,6 +390,8 @@ func translateRule(npmNetPol *policies.NPMNetworkPolicy, netPolName string, dire
 		}
 	}
 
+	cfg := npmconfig.DefaultConfig
```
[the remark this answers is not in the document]
rejain456: changed it to reflect this, thank you! 


QQ: What is the benefit of reading the config in at runtime only?

npm/pkg/controlplane/translation/translatePolicy.go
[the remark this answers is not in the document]
rejain456: Re-factored so we call the validate function only once. @huntergregory , plase let me know if that looks better

npm/pkg/controlplane/translation/translatePolicy.go
```diff
@@ -403,11 +408,24 @@ func translateRule(npmNetPol *policies.NPMNetworkPolicy, netPolName string, dire
...
+				return err
+			}
+
 			// Do not need to run below code to translate PodSelector and NamespaceSelector
 			// since IPBlock field is exclusive in NetworkPolicyPeer (i.e., peer in this code).
 			continue
 		}
 
+		// check npm lite enabled and valid policy in case the cidr block is not the first peer
+		err := NpmLiteValidPolicy(peer, NpmLiteEnabled)
```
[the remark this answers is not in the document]
rejain456: Yup noted thanks

npm/pkg/controlplane/translation/translatePolicy_test.go
rejain456: Added test cases for:


- CIDR + label selector combo in same peer
- no peers 


for allow-all-internal case, added an exception check within the code for it

npm/pkg/controlplane/translation/translatePolicy.go
rejain456: Updated the description thanks

huntergregory: nit: could we modify the error description as well to help a user in case they try debugging this error in logs?

I believe this Lite mode will actually work for all Policies except those with an ingress/egress pod selector or namespace selector

huntergregory (changes requested):
minor suggestions

npm/pkg/controlplane/translation/translatePolicy.go
rejain456: updated the description

npm/pkg/controlplane/translation/translatePolicy_test.go
```diff
@@ -2925,6 +2925,8 @@ func TestNpmLiteCidrPolicy(t *testing.T) {
 	// Test 1) Npm lite enabled, CIDR + Namespace label Peers, returns error
 	// Test 2) NPM lite disabled, CIDR + Namespace label Peers, returns no error
 	// Test 3) Npm Lite enabled, CIDR Peers , returns no error
+	// Test 4) NPM Lite enabled, Combination of CIDR + Label in same peer, returns an error
+	// test 5) NPM Lite enabled, no peer, returns no error
```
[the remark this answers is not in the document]
rejain456: Yup added a test case with just port + protocol, it does not return an error @huntergregory

npm/pkg/controlplane/translation/translatePolicy.go
huntergregory: sorry, I mean inside the code 😅

huntergregory (changes requested):
Realizing that NPM Lite cannot support "named port" fields... We will need to check for that too

npm/pkg/controlplane/translation/translatePolicy.go
rejain456: haha got it! I updated it, @huntergregory let me know how that sounds

rejain456: Also great point out for named ports, added logic to output an error in case of named ports as well

huntergregory (approved):

npm/azure-npm.yaml
```diff
@@ -172,6 +172,7 @@ data:
             "EnableV2NPM":             true,
             "PlaceAzureChainFirst":    false,
             "ApplyInBackground":       true,
-            "NetPolInBackground":      true
+            "NetPolInBackground":      true,
+            "EnableNPMLite":           true
```
vakalapa, line 176: Dont change this file, instead create a new one

npm/npm.go
```diff
@@ -58,20 +60,24 @@ type NetworkPolicyManager struct {
...
+	k8sServerVersion *version.Info,
+	npmLiteToggle bool) *NetworkPolicyManager {
 	klog.Infof("API server version: %+v AI metadata %+v", k8sServerVersion, aiMetadata)
 
 	npMgr := &NetworkPolicyManager{
-		config:    config,
-		Dataplane: dp,
+		config:        config,
+		Dataplane:     dp,
+		NpmLiteToggle: npmLiteToggle,
```
vakalapa, line 74: set this with config.Toggles.NPMlite ? instead of a new bool

npm/config/config.go
```diff
@@ -51,6 +51,7 @@ var DefaultConfig = Config{
 		ApplyInBackground: true,
 		// NetPolInBackground is currently used in Linux to apply NetPol controller Add events in the background
 		NetPolInBackground: true,
+		EnableNPMLite:      true,
```
vakalapa, line 54: This default config should always be false, or else we will break existing deployments where the configmap is not explicitly set to false.

npm/pkg/dataplane/types_windows_test.go
```diff
@@ -425,7 +425,7 @@ func UpdatePolicy(policy *networkingv1.NetworkPolicy) *Action {
 
 // Do models policy updates in the NetworkPolicyController
 func (p *PolicyUpdateAction) Do(dp *DataPlane) error {
-	npmNetPol, err := translation.TranslatePolicy(p.Policy)
+	npmNetPol, err := translation.TranslatePolicy(p.Policy, false)
```
vakalapa, line 428: Do we need any test cases for this set to true ?

vakalapa (changes requested):

npm/npm.go
rejain456: yes changed it, thank you for pointing that out

npm/azure-npm.yaml
```diff
@@ -174,4 +174,4 @@ data:
             "ApplyInBackground":       true,
             "NetPolInBackground":      true
         }
-    }
+    }
```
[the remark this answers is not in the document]
rejain456: added a new line, the file is no longer shown! Thanks

npm/config/config.go
rejain456: Yup I had just enabled it for testing purposes with the intent of disabling it before merging the pr in. I can disable it in the pr and enable locally for testing purposes if that is the better way @vakalapa

npm/azure-npm.yaml
rejain456: created a new one!

npm/azure-npm-lite.yaml
huntergregory, line 1: can we move this to _npm/examples/npm-lite.yaml_?

npm/examples/windows/azure-npm-win.yaml
huntergregory, line 1: can we add windows DS with Linux DS at _npm/examples/npm-lite.yaml_?

huntergregory (changes requested):

npm/examples/windows/azure-npm-win.yaml
rejain456: Added both files (linux and windows azure-npm.yaml) under npm/examples/npm-lite.yaml. @huntergregory , does this look better?

npm/pkg/dataplane/types_windows_test.go
rejain456: I think it is ok if we don't as this function is used to test windows data plane code. I have already added unit tests to test the properties within the translate function policy itself using enabled/disabled npmlite. As the primary role of this is to test data plane code which I have not changed, I believe can go with default disabled npmlite. Please let me know what you think!

npm/examples/windows/azure-npm-win.yaml
huntergregory: I think it will just be a little easier to maintain one file, since they both use the same ServiceAccount, etc.

npm/pkg/controlplane/controllers/v2/networkPolicyController_test.go
```diff
@@ -48,7 +48,7 @@ func (f *netPolFixture) newNetPolController(_ chan struct{}, dp dataplane.Generi
 	kubeclient := k8sfake.NewSimpleClientset(f.kubeobjects...)
 	f.kubeInformer = kubeinformers.NewSharedInformerFactory(kubeclient, noResyncPeriodFunc())
 
-	f.netPolController = NewNetworkPolicyController(f.kubeInformer.Networking().V1().NetworkPolicies(), dp)
+	f.netPolController = NewNetworkPolicyController(f.kubeInformer.Networking().V1().NetworkPolicies(), dp, false)
```
huntergregory, line 51: could we add some NPM Lite tests in this file? Realized we are missing them from https://github.com/Azure/azure-container-networking/pull/3005#discussion_r1762124172

npm/pkg/dataplane/types_windows_test.go
huntergregory: Seems ok to me to handle all testing in control plane

huntergregory (commented):
nit: could we update the PR title to be like: `feat: [NPM] ...`

Could you also add NPM, linux, and windows GH labels to this PR?

npm/pkg/controlplane/controllers/v2/networkPolicyController_test.go
rejain456: Added a unit test here

npm/azure-npm-lite.yaml
rejain456: yes moved the file to this path and combined linux + windows in one file

npm/examples/azure-npm-lite.yaml
huntergregory, line 1: nit: ServiceAccount, etc. are duplicated in this file

npm/pkg/controlplane/controllers/v2/networkPolicyController_test.go
```diff
@@ -275,7 +276,31 @@ func TestAddNetworkPolicy(t *testing.T) {
...
+
+	addNetPol(f, netPolObj)
+	testCases := []expectedNetPolValues{
+		{1, 0, netPolPromVals{1, 1, 0, 0}},
+	}
+
+	checkNetPolTestResult("TestAddNetPol", f, testCases)
+}
+
+func TestAddNetworkPolicyWithNPMLite(t *testing.T) {
```
huntergregory, line 291: is it possible to add a test for the unhappy path? With translate failure

huntergregory (approved):
great job! 🚀

vakalapa (approved):

npm/examples/azure-npm-lite.yaml
rejain456: synced offline - can separate for now

PR #3016, opened 2024-09-13
Add join VNET call for every AZR NC unpublish call with fixed UTs

smittal22:
**Reason for Change**:
For AZR scenarios, if NMAgent is restarted, it loses state and does not know what VNETs to subscribe to. As it no longer has VNET state, successive put nc or delete nc calls would fail. We, however, always call join VNET before put nc call. This PR adds such a flow for delete AZR nc calls as well.

PS: We don't need this flow for regular NCs, as NMAgent there uses NMA swift module to rehydrate VNETs if state is lost.

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -20,6 +20,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/cns/wireserver"
 	"github.com/Azure/azure-container-networking/common"
+	"github.com/Azure/azure-container-networking/nmagent"
 	"github.com/pkg/errors"
 )
 
@@ -1029,7 +1030,20 @@ func (service *HTTPRestService) unpublishNetworkContainer(w http.ResponseWriter,
 
 	ctx := r.Context()
 
-	if !service.isNetworkJoined(req.NetworkID) {
+	var unpublishBody nmagent.DeleteContainerRequest
+	if req.DeleteNetworkContainerRequestBody != nil {
+		err = json.Unmarshal(req.DeleteNetworkContainerRequestBody, &unpublishBody)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("could not unmarshal delete network container body: %v", err), http.StatusBadRequest)
+			return
+		}
+	}
+
+	/* For AZR scenarios, if NMAgent is restarted, it loses state and does not know what VNETs to subscribe to.
+	As it no longer has VNET state, delete nc calls would fail. We need to add join VNET call for all AZR
+	nc unpublish calls just like publish nc calls.
+	*/
+	if unpublishBody.AZREnabled || !service.isNetworkJoined(req.NetworkID) {
 		joinResp, err := service.wsproxy.JoinNetwork(ctx, req.NetworkID) //nolint:govet // ok to shadow
 		if err != nil {
 			resp := cns.UnpublishNetworkContainerResponse{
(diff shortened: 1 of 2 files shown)


cns/restserver/api.go
```diff
@@ -1029,7 +1031,27 @@ func (service *HTTPRestService) unpublishNetworkContainer(w http.ResponseWriter,
...
+		// If the body contains only `""\n`, it is non-AZR NC
+		// In this case, we should not return an error
+		// However, if the body is not `""\n`, it is invalid and therefore, we must return an error
+		if !bytes.Equal(req.DeleteNetworkContainerRequestBody, []byte(`""`+"\n")) {
+			http.Error(w, fmt.Sprintf("could not unmarshal delete network container body: %v", err), http.StatusBadRequest)
+			return
+		}
+	} else {
+		// If unmarshalling was successful, it is an AZR NC
+		azrNC = true
```
timraymond, line 1047: This is too much knowledge about NMAgent quirks in a component that shouldn't care. I'd recommend making a `nmagent.NewDeleteContainerRequest` that returns a request and an error. Then all the API needs to care about is whether or not that function says it was able to create a request or not. Then, ideally... you can just ask the request if it's an AZR NC or not instead of needing the other bool for it.

smittal22: We are not really creating any request. Just passing along the body that CNS received to NMA as payload.

cns/restserver/api.go
```diff
@@ -1029,7 +1031,27 @@ func (service *HTTPRestService) unpublishNetworkContainer(w http.ResponseWriter,
...
 
-	if !service.isNetworkJoined(req.NetworkID) {
+	var unpublishBody nmagent.DeleteContainerRequest
+	var azrNC bool
+	err = json.Unmarshal(req.DeleteNetworkContainerRequestBody, &unpublishBody)
+	if err != nil {
+		// If the body contains only `""\n`, it is non-AZR NC
+		// In this case, we should not return an error
+		// However, if the body is not `""\n`, it is invalid and therefore, we must return an error
+		if !bytes.Equal(req.DeleteNetworkContainerRequestBody, []byte(`""`+"\n")) {
```
ramiro-gamarra, line 1041: the bytes are hard to parse due to the mix of raw and regular string literals. Easier to just write `[]byte{34, 34, 10}` and explain what it is in the comment?

cns/restserver/api_test.go
```diff
@@ -1068,20 +1068,71 @@ func TestUnpublishNCViaCNS(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			errPublish := publishNCViaCNS(vnet, tt.ncID, createNetworkContainerURL)
+			require.NoError(t, errPublish)
+			errUnpublish := unpublishNCViaCNS(vnet, tt.ncID, deleteNetworkContainerURL, tt.body)
+			if tt.requireError {
+				require.Error(t, errUnpublish)
```
ramiro-gamarra, line 1128: the error could be coming from multiple places, not necessarily from the bad input from the test. you need a more thorough check on the type of error you expect from this specific bad input.

cns/restserver/api.go
smittal22: Yeah, i had a tossup between the two and thought visually the bytes make more sense, but I can change that up.

smittal22: Changed

cns/restserver/api_test.go
smittal22: addressed

ramiro-gamarra (approved):

PR #3017, opened 2024-09-13
Add support in CNS NMAgent Client to fetch secondary IPs

santhoshmprabhu:
**Reason for Change**:
As part of the work to add support for NodeSubnet with Cilium, we need the CNS to be able to fetch IPs from NMAgent. This PR adds this ability to the NMAgent client in CNS. Specifically, this PR introduces:
1. A new request object for Secondary IP requests
2. A new method to interact with the NMAgent to get the IPs
3. A wrapper to invoke the new method in the NMAgent client. The wrapper implements basic throttling too.
4. Unit Tests

The fetch functionality mimics the behavior of azure-vnet-ipam, but has been re-implemented to match the NMAgent client architecture in CNS.

**Issue Fixed**:
NA

**Requirements**:

 - [ ] includes documentation
 - [x] adds unit tests


santhoshmprabhu:

nmagent/client.go
```diff
@@ -284,6 +288,70 @@ func (c *Client) GetHomeAz(ctx context.Context) (AzResponse, error) {
...
 
+func (c *Client) RefreshSecondaryIPsIfNeeded(ctx context.Context) (refreshNeeded bool, ips []string, err error) {
+	if time.Since(c.secondaryIPLastRefreshTime) < c.secondaryIPQueryInterval {
+		return false, nil, nil
+	}
+
+	c.secondaryIPLastRefreshTime = time.Now()
+	res, err := c.getSecondaryIPs(ctx)
+	return true, res, err
+}
```
timraymond, line 299: This needs to be done in a wrapper of the client because the ethos of this client is that it will always do what you want it to, not sometimes.

What is the design motivation here? There are better ways to respect rate limits if that's what we're going for.

nmagent/client.go
```diff
@@ -284,6 +288,70 @@ func (c *Client) GetHomeAz(ctx context.Context) (AzResponse, error) {
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, die(resp.StatusCode, resp.Header, resp.Body, req.URL.Path)
+	}
+
+	return parseSecondaryIPsFromWireServerResponse(resp)
+}
+
+func parseSecondaryIPsFromWireServerResponse(resp *http.Response) (res []string, err error) {
```
timraymond, line 320: This should generally parse the response and put it in a set of structs (in responses.go). Drilling into the response should be done by the user of the client, not the client itself. This allows it to be a general client for NMAgent, not a one-off data grabber (which was what this package was intended to replace).

nmagent/client.go
```diff
@@ -284,6 +288,70 @@ func (c *Client) GetHomeAz(ctx context.Context) (AzResponse, error) {
...
+				if a.IsPrimary {
+					continue
+				}
+
+				res = append(res, a.Address)
+				addressCount++
+			}
+			log.Printf("Got %d addresses from subnet %s", addressCount, s.Prefix)
+		}
+	}
```
timraymond, line 350: This is all business logic that belongs in the user of the client. The responsibility of the client ends at translating the XML response to a set of structs. Also, the client only returns wrapped errors: no logging (that's the caller's responsibility).

nmagent/client.go
```diff
@@ -284,6 +288,70 @@ func (c *Client) GetHomeAz(ctx context.Context) (AzResponse, error) {
...
+	if resp.StatusCode != http.StatusOK {
+		return nil, die(resp.StatusCode, resp.Header, resp.Body, req.URL.Path)
+	}
+
+	return parseSecondaryIPsFromWireServerResponse(resp)
+}
+
+func parseSecondaryIPsFromWireServerResponse(resp *http.Response) (res []string, err error) {
+	// Decode XML document.
+	var doc common.XmlDocument
```
timraymond, line 322: We can (and should) do this without taking the dependency on `common`. See `responses.go`

timraymond (changes requested):
Good start, I like the addition of the tests, but there's a few changes that need to be made to align it with the rest of the package.

santhoshmprabhu:
@timraymond Thanks for the comments, addressed them. Please take another look.

cns/nodesubnet/ip_fetcher.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+// Exposed for testing
+func (c *IPFetcher) SetSecondaryIPQueryInterval(interval time.Duration) {
+	c.secondaryIPQueryInterval = interval
+}
+
+// If secondaryIPQueryInterval has elapsed since the last fetch, fetch secondary IPs
+func (c *IPFetcher) RefreshSecondaryIPsIfNeeded(ctx context.Context) (refreshNeeded bool, ips []string, err error) {
```
timraymond, line 38: Generally, if it's not 100% success, it's an error. This includes the entire concept of "retriable" / "non-retriable" errors that we have scattered around (there's good ways of dealing with this if you run into it in the future--ask me). In this case, `refreshNeeded` should be a sentinal error (`ErrRefreshNotNeeded`). That way a caller can just:

```go
ips, err := RefreshSecondaryIPsBlah()
if err != nil && !errors.Is(err, ErrRefreshNotNeeded) {
       // log / alert / etc.
}
```

timraymond (changes requested):
Much better. I've got just a few more things which would be improvements given the history of this client, CNS in general, and this particular endpoint.

I think the debouncing / throttling implementation can be tackled separately, but as we discussed earlier I think that's worth doing. At the same time we can probably rip out the `cns/wireserver` package once these changes are done.

nmagent/responses.go
```diff
@@ -40,3 +40,26 @@ type NCVersionList struct {
...
+}
+
+type InterfaceSubnetInfo struct {
+	IPAddress []IPAddress `xml:"IPAddress"`
+	Prefix    string      `xml:"Prefix,attr"`
+}
+
+type Interface struct {
+	InterfaceSubnets []InterfaceSubnetInfo `xml:"IPSubnet"`
+	MacAddress       string                `xml:"MacAddress,attr"`
```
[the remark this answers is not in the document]
rbtr: link doesn't work for me (it takes me to the top of the diff view, not a thread), but I suspect it's to [this](https://github.com/Azure/azure-container-networking/pull/2634#discussion_r1575311420)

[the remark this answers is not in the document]
rbtr: the driving principle here is more thoroughly explained as [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

nmagent/client.go
```diff
@@ -284,6 +283,36 @@ func (c *Client) GetHomeAz(ctx context.Context) (AzResponse, error) {
 	return homeAzResponse, nil
 }
 
+func (c *Client) GetInterfaceIPInfo(ctx context.Context) (Interfaces, error) {
```
timraymond, line 286: It would be nice to put a doc comment here of the form: `// GetInterfaceIPInfo does x, y, z`

cns/nodesubnet/ip_fetcher.go
```diff
@@ -0,0 +1,77 @@
...
+	"net"
+	"time"
+
+	"github.com/Azure/azure-container-networking/nmagent"
+	"github.com/pkg/errors"
+)
+
+var ErrorRefreshSkipped = errors.New("refresh skipped due to throttling")
+
+// This interface is implemented by the NMAgent Client, and also a mock client for testing
```
timraymond, line 15: This will be interpreted as a doc comment. The typical style is a definition with the name of the type as the first word, so:

```suggestion
// InterfaceRetriever is an interface is implemented by the NMAgent Client, and also a mock client for testing.
```

cns/nodesubnet/ip_fetcher.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func NewIPFetcher(nmaClient InterfaceRetriever, queryInterval time.Duration) *IPFetcher {
+	return &IPFetcher{
+		ipFectcherClient:         nmaClient,
+		secondaryIPQueryInterval: queryInterval,
+	}
+}
+
+// If secondaryIPQueryInterval has elapsed since the last fetch, fetch secondary IPs
```
timraymond, line 35: This should also go in the body of the method so as not to accidentally appear in godoc

cns/nodesubnet/ip_fetcher.go
```diff
@@ -0,0 +1,77 @@
...
+	"context"
+	"log"
+	"net"
+	"time"
+
+	"github.com/Azure/azure-container-networking/nmagent"
+	"github.com/pkg/errors"
+)
+
+var ErrorRefreshSkipped = errors.New("refresh skipped due to throttling")
```
timraymond, line 13: This is okay as-is, but it's a bit more typical to see `ErrRefreshSkipped`.

timraymond (approved):
Mostly non-blocking feedback about doc comments but otherwise LGTM.

timraymond (approved):

rbtr (commented):
lgtm just one suggestion

rbtr (approved):

nmagent/client.go
```diff
@@ -284,6 +283,37 @@ func (c *Client) GetHomeAz(ctx context.Context) (AzResponse, error) {
...
+
+	if resp.StatusCode != http.StatusOK {
+		return out, die(resp.StatusCode, resp.Header, resp.Body, req.URL.Path)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return out, die(resp.StatusCode, resp.Header, resp.Body, req.URL.Path)
+	}
+
+	err = xml.NewDecoder(resp.Body).Decode(&out)
```
tamilmani1989, line 309: i know logging entire xml output would spam logs but we need to find a way to log if we get empty xml doc from nmagent

tamilmani1989 (approved):

nmagent/client.go
timraymond: @tamilmani1989 Under what conditions will that happen? We can (and should) detect that and create an error for it that says _why_ NMAgent is giving us empty xml.

PR #3023, opened 2024-09-19
Make throttling nmagent fetches for nodesubnet more dynamic

santhoshmprabhu:
**Reason for Change**:
We previously added fetch of secondary IPs to the NMAgent client in CNS. This PR makes the fetch asynchronous, with the ability to adjust the frequency dynamically. Specifically,
1. Min and max interval between fetches can be configured
2. Frequency is cut in half (interval doubled) if no diff is seen, up to the max interval.
3. Min interval is used as soon as a diff is seen

This is implemented using a time Ticker, but we wrap it with a wrapper so as to mock it for tests. This PR also includes some additional checks for empty responses from the nmagent (this was raised in [this comment](https://github.com/Azure/azure-container-networking/pull/3017#discussion_r1765862289)). Also have UTs.

**Issue Fixed**:
NA

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:


cns/nodesubnet/ip_fetcher_test.go
```diff
@@ -12,55 +13,117 @@ import (
...
 // Mock client that simply tracks if refresh has been called
 type TestClient struct {
-	fetchCalled bool
+	refreshCount int32
 }
 
 // Mock refresh
 func (c *TestClient) GetInterfaceIPInfo(_ context.Context) (nmagent.Interfaces, error) {
-	c.fetchCalled = true
+	atomic.AddInt32(&c.refreshCount, 1)
```
timraymond, line 21: Atomics are pretty rare unless there's a specific performance need. I forget the exact reasoning, but there's some sharp edges on them if I recall. Usually a mutex is fine.

timraymond (commented):
I think I'm going to recant some of what I said in the last PR now that I understand _why_ we're doing this debouncing. As I mentioned during the design review, rate limits and wireserver are a fact of life, so it now feels weird to have arbitrary logic treating the nmagent client carefully because it might blow our rate limit. The design intention of the NMAgent client was always that users can do `nmagent.DoSomethingForMe()` and it will block while it deals with all of the weirdness that happens with NMAgent and Wireserver. This includes async response handling... and I'm thinking it should also include respecting rate limits. If the request is going to blow the rate limit, the WireserverTransport in the NMAgent client needs to block that request until the request will succeed. Otherwise, we're just setting the user up for failure.

Doing this with tickers is definitely the right way (in particular the whole ticker resetting thing is how I've done respecting rate limits in the past). However, it should be done in https://github.com/Azure/azure-container-networking/blob/master/nmagent/internal/wireserver.go so that it's shared across the client.

cns/nodesubnet/ip_fetcher.go
```diff
@@ -10,46 +10,130 @@ import (
...
+	}
+
+	if len(response.Entries) == 0 {
+		return errors.New("no interfaces found in response from NMAgent")
 	}
 
-	c.secondaryIPLastRefreshTime = time.Now()
-	response, err := c.ipFectcherClient.GetInterfaceIPInfo(ctx)
+	primaryIP, secondaryIPs := flattenIPListFromResponse(&response)
+	err = c.consumer.UpdateIPsForNodeSubnet(primaryIP, secondaryIPs)
```
tamilmani1989, line 126: what primary IP is used for?

cns/nodesubnet/ip_fetcher.go
```diff
@@ -10,46 +10,130 @@ import (
...
+		intfFetcherClient:  client,
+		consumer:           consumer,
+		minRefreshInterval: minInterval,
+		maxRefreshInterval: maxInterval,
+		tickerInterval:     minInterval,
+	}
+}
+
+func (c *IPFetcher) UpdateFetchIntervalForNoObservedDiff() {
+	c.tickerInterval = min(c.tickerInterval*2, c.maxRefreshInterval) //nolint:gomnd // doubling interval
```
tamilmani1989, line 70: so interval increases exponentially for every refresh until it hits max and stays max? Lets add a comment describing that behavior

cns/nodesubnet/refreshticker.go
```diff
@@ -0,0 +1,37 @@
+package nodesubnet
```
tamilmani1989, line 1: can this ticker be own package so that any package can use this?

cns/nodesubnet/ip_fetcher.go
```diff
@@ -10,46 +10,130 @@ import (
...
+
+		for {
+			select {
+			case <-c.ticker.C():
+				err := c.RefreshSecondaryIPs(ctx)
+				if err != nil {
+					log.Printf("Error refreshing secondary IPs: %v", err)
+				}
+			case <-ctx.Done():
+				log.Println("IPFetcher stopped")
```
tamilmani1989, line 107: can use log.Printf.. i think log package adds new line by default

cns/nodesubnet/refreshticker.go
rbtr: we already use `avast/retry` elsewhere, do we need to implement our own exponential backoff retrier?

santhoshmprabhu: Took a look at it. It seems `avast/retry` is meant for synchronous retries, not for async polling.  If I understand correctly, the backoff there defines the time we wait to retry after a failure. In this case, we need periodic polling, irrespective of past polls succeeding or not.

cns/nodesubnet/ip_fetcher.go
santhoshmprabhu: You might recall that our plan is to use an artificial Nodesubnet NC to hold the IPs. NCs expect a primary IP to be present. Now, we could leave it empty in Nodesubnet, but it seemed better to populate it since we are fetching it anyway.

cns/nodesubnet/refreshticker.go
santhoshmprabhu: Moved the ticker to a separate package.

cns/nodesubnet/ip_fetcher_test.go
santhoshmprabhu: Switched to mutex.

cns/nodesubnet/ip_fetcher.go
tamilmani1989: ok got it. Lets add this as comment

tamilmani1989 (dismissed):
lgtm. left a minor comment

cns/nodesubnet/ip_fetcher.go
santhoshmprabhu: Ended up removing it, because it's ultimately unnecessary and was making endpoint state restoration complicated.

cns/nodesubnet/ip_fetcher.go
```diff
@@ -7,49 +7,144 @@ import (
...
 	}
 }
 
-func (c *IPFetcher) RefreshSecondaryIPsIfNeeded(ctx context.Context) (ips []netip.Addr, err error) {
-	// If secondaryIPQueryInterval has elapsed since the last fetch, fetch secondary IPs
-	if time.Since(c.secondaryIPLastRefreshTime) < c.secondaryIPQueryInterval {
-		return nil, ErrRefreshSkipped
+// UpdateFetchIntervalForNoObservedDiff informs IPFetcher that a diff was observed in the last fetch.
+// In the current design, this resets the fetch interval to the minimum.
+func (c *IPFetcher) UpdateFetchIntervalForObservedDiff() {
```
timraymond, line 87: I don't like that controlling the fetch interval is controllable outside of `*IPFetcher`. This feels like a leaky abstraction.

cns/nodesubnet/ip_fetcher.go
```diff
@@ -7,49 +7,144 @@ import (
...
+	}
+}
+
+// Start the IPFetcher.
+func (c *IPFetcher) Start(ctx context.Context) {
+	go func() {
+		// Do an initial fetch
+		err := c.RefreshSecondaryIPs(ctx)
+		if err != nil {
+			log.Printf("Error refreshing secondary IPs: %v", err)
```
timraymond, line 101: This shouldn't use a global logger

timraymond (commented):
I think one of my issues with this is that it feels like the job of "watch the results of this function / method / whatever and refresh it if it's changing" and "get IPs from NMAgent and process the response into a primary and secondary IPs" are entangled where they don't need to be. I think HomeAzMonitor is doing something similar, so it feels like there's some shared DNA where both could benefit from a generic Refresher construct.

cns/nodesubnet/ip_fetcher.go
```diff
@@ -6,50 +6,93 @@ import (
...
-		return nil, ErrRefreshSkipped
+// Fetch IPs from NMAgent and pass to the consumer
+func (c *IPFetcher) ProcessInterfaces(response nmagent.Interfaces) error {
+	if len(response.Entries) == 0 {
+		return errors.New("no interfaces found in response from NMAgent")
 	}
 
-	c.secondaryIPLastRefreshTime = time.Now()
-	response, err := c.ipFectcherClient.GetInterfaceIPInfo(ctx)
+	_, secondaryIPs := flattenIPListFromResponse(&response)
```
tamilmani1989, line 84: didnt you mention primary IP is required to stay in order with existing internal goalstate? is it not required now?

cns/refresh/fetcher.go
```diff
@@ -0,0 +1,108 @@
...
+
+				f.ticker.Reset(f.currentInterval)
+			}
+		}
+	}()
+}
+
+func (f *Fetcher[T]) updateFetchIntervalForNoObservedDiff() {
+	f.currentInterval = min(f.currentInterval*2, f.maxInterval) // nolint:gomnd // doubling logic
+}
```
tamilmani1989, line 104: this is going to be common logic for any fetcher? what if another fetcher want different logic

cns/nodesubnet/ip_fetcher.go
santhoshmprabhu: Primary IP is one field in the NC request object. I was previously passing the true primary IP, because we had it. But when restoring state from the statefile we don't have the primary IP, and that causes a diff in the NC request object when initializing from the state file. So leaving it empty to keep things consistent.

cns/refresh/fetcher.go
santhoshmprabhu: Right now this isn't vastly customizable. One could do binary exponential backoff, or static interval (by setting max = min). Other schemes can come as needed I guess.

cns/refresh/fetcher.go
```diff
@@ -0,0 +1,108 @@
...
+		consumeFunc:     consumeFunc,
+	}
+}
+
+func (f *Fetcher[T]) Start(ctx context.Context) {
+	go func() {
+		// do an initial fetch
+		res, err := f.fetchFunc(ctx)
+		if err != nil {
+			logger.Printf("Error refreshing secondary IPs: %v", err)
```
timraymond, line 56: Two things: don't use the global logger (inject it in `*Fetcher[T]` as an interface), then adjust the message to be accordingly generic.

cns/refresh/fetcher.go
```diff
@@ -0,0 +1,108 @@
...
+				logger.Printf("Fetcher stopped")
+				return
+			case <-f.ticker.C():
+				result, err := f.fetchFunc(ctx)
+				if err != nil {
+					logger.Errorf("Error fetching data: %v", err)
+				} else {
+					if reflect.DeepEqual(result, f.cache) {
+						f.updateFetchIntervalForNoObservedDiff()
+						logger.Printf("No diff observed in fetch, not invoking the consumer")
```
timraymond, line 84: Consider making this a debug log

cns/refresh/fetcher.go
```diff
@@ -0,0 +1,108 @@
...
+			select {
+			case <-ctx.Done():
+				logger.Printf("Fetcher stopped")
+				return
+			case <-f.ticker.C():
+				result, err := f.fetchFunc(ctx)
+				if err != nil {
+					logger.Errorf("Error fetching data: %v", err)
+				} else {
+					if reflect.DeepEqual(result, f.cache) {
```
timraymond, line 82: Never use `reflect.DeepEqual`, use `github.com/google/go-cmp/cmp.Equal` instead (it's a drop-in replacement that is safer).

timraymond (changes requested):
Looking better, there's a few changes needed around logging.

cns/refresh/fetcher.go
tamilmani1989: can we add a comment saying if static, then interval has to be set to max

santhoshmprabhu: Done.

cns/refresh/fetcher.go
rbtr: consider making the type implement `Comparable` instead of either of these

santhoshmprabhu: @rbtr As far as I could tell, comparable is just a characteristic of the struct, not an interface I could implement (https://go.dev/ref/spec#Comparison_operators). Could you please provide more info?

santhoshmprabhu: @rbtr I could add the constraint, but it seems struct types either satisfies the constraint or not based on their members. Specifically, the struct we are currently passing wouldn't satisfy the comparable constraint due to having slice members. Generally that would restrict potential structs to much I feel.

santhoshmprabhu: > @rbtr I could add the constraint, but it seems struct types either satisfy the constraint or not based on their members (we can't make a struct satisfy it, without editing the members). Specifically, the struct we are currently passing wouldn't satisfy the comparable constraint due to having slice members. Generally that would restrict potential structs to much I feel.

timraymond: So, @rbtr and I talked offline. I had tried the Comparable route and it was not fruitful for the reasons you mention @santhoshmprabhu . Also, my previous thought about using `cmp.Equal` was incorrect since that's entirely useful only for tests (usually people reach for `reflect.DeepEqual` in tests, where that advice usually applies). In this case, instead of parameterizing the type on `any`, parameterize it on a generic interface:

```go
type Equalable[T any] interface{
       Equal(other T) bool
}

type Fetcher[T Equalable[T]] struct {
       // snip
}
```

Then anyone using this is forced to implement `Equal(other T) bool`, which can provide equality guarantees through unit tests... then we can just call that method.

santhoshmprabhu: I had done this before I saw @timraymond 's comment, so the names are a little different. Please take a look.
cc: @rbtr

santhoshmprabhu: I am somewhat concerned about the amount of copying that the collection of Equal methods is adding, so if you folks think we need the function by value only for Interface and the rest can take pointers, that would be more efficient.

refresh/equaler.go
```diff
@@ -0,0 +1,5 @@
+package refresh
+
+type Equaler[T any] interface {
+	Equal(T) bool
```
rbtr, line 4: I think convention is `Equals`, and you may not need to export this interface?

santhoshmprabhu: Un-exported the interface. Regarding the function name, it seems go-cmp looks for `Equal`, so I was using that as a guide (though we aren't really using go-cmp itself of course) -
image

timraymond: The standard library convention is Equal, which is also probably why `cmp` uses it:

- `net.(IP).Equal`: https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/ip.go;l=359
- `time.(Time).Equal`: https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/time/time.go;l=301

nmagent/equality.go
```diff
@@ -0,0 +1,71 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// Equal compares two Interfaces objects for equality.
+func (i Interfaces) Equal(other Interfaces) bool {
+	return i.equalPtr(&other)
+}
```
timraymond, line 19: I don't think this double implementation should be necessary due to automatic referencing / dereferencing. @rbtr ?

santhoshmprabhu: I believe the automatic dereferencing works for i (in fact, you can see I didn't do `&i.equalPtr`, but he `other` parameter needs to be explicitly referenced I think.

timraymond: @santhoshmprabhu did you try with implementing only the pointer receiver on these and see if it breaks?

santhoshmprabhu: @timraymond Yup. The pointer version needs to be explicitly called with the `(&other)` param. We could of course have the client return pointers rather than the struct itself and then work entirely with pointers. But then that somewhat breaks from the existing pattern of structs and umarshaling in the client. Do you want me to do that?

timraymond: This passes your tests: https://github.com/Azure/azure-container-networking/commit/1fbfb6e7c97d8a0630b8f5ffe58b0165bab363bf

santhoshmprabhu: It does, but it copies objects back and forth several times. Each call to `Equal` creates copies of the two things you're trying to compare. You can see I had that first, explicitly switched to pointers in this commit:
https://github.com/Azure/azure-container-networking/pull/3023/commits/2331766ea80a04732880a9ebc9bd20f245e33dbe

timraymond: Fair, but unless we have a measurable performance impact, code readability is the priority over optimizations. If `Equal` shows up as a big blob on a pprof, then we know what optimization to take. Until it does though, we don't know if we're compromising readability for little gain.

santhoshmprabhu: Switched to the non-pointer versions only.

timraymond (dismissed):

tamilmani1989 (approved):

PR #3027, opened 2024-09-23
fix: add default route on customer vnets and provide default route from CNS to CNI

paulyufan2:
**Reason for Change**:

This PR is to fix one issue in swiftv2 Windows scenario:
When a pod is created, the default route is added on infra vnet:
root@swiftv2-pod-3:/# ip route
**default via 10.244.2.1 dev eth0 metric 1**
It leads to ping a VM IP in the same VNET that cannot work.

There are two issues::
1.CNS does not provide the default route to CNI;
2.CNI should only add the default route to secondary interface customer vnet; on Swiftv2 scenario, skipDefaultRoutes is set to true for infraNIC interface and false for a secondary interface; so if !info.skipDefaultRoutes, then add default route.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


tamilmani1989 (changes requested):
how it works for linux not windows? why we need this change specifically for windows

cni/network/invoker_cns.go
```diff
@@ -439,11 +439,15 @@ func configureDefaultAddResult(info *IPResultInfo, addConfig *IPAMAddConfig, add
...
-				Dst: defaultRouteDstPrefix,
-				Gw:  ncgw,
-			})
+			// on Swiftv2 windows scenario, skip to add default route on infra vnet;
+			if !info.skipDefaultRoutes {
+				resRoute = append(resRoute, network.RouteInfo{
+					Dst: defaultRouteDstPrefix,
+					Gw:  ncgw,
+				})
+			}
```
tamilmani1989, line 448: if routes are specified as part of cns goalstate why cni need to construct new one?

tamilmani1989: what scenario this else case solves?

network/endpoint_windows.go
```diff
@@ -428,7 +428,7 @@ func (nw *network) newEndpointImplHnsV2(cli apipaClient, epInfo *EndpointInfo) (
 	}
 
 	// Create the HCN endpoint.
-	logger.Info("Creating hcn endpoint", zap.Any("hcnEndpoint", hcnEndpoint), zap.String("computenetwork", hcnEndpoint.HostComputeNetwork))
+	logger.Info("Creating hcn endpoint", zap.Any("hcnEndpoint", hcnEndpoint), zap.String("computenetwork", hcnEndpoint.HostComputeNetwork), zap.Any("routes", hcnEndpoint.Routes))
```
tamilmani1989, line 431: can skip logging routes if hcnendpoint object already logs it.

tamilmani1989 (changes requested):

network/endpoint_windows.go
paulyufan2: zap log does not log Routes field from hcnEndpoint

cni/network/invoker_cns.go
paulyufan2: it's not correct...I built image based on this code and did not push the correct codes.

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -40,5 +40,12 @@ func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(podIPInfo *cns.Pod
 		},
 		GatewayIPAddress: interfaceInfo.GatewayIP,
 	}
+	// assign default route
+	route := cns.Route{
+		IPAddress:        "0.0.0.0/0",
```
prmathur-microsoft, line 45: Why 0.0.0.0/0

kmurudi: is this added for linux path though in current scenario?

paulyufan2: yes, it's added for Linux path, we should do the same thing for windows path

paulyufan2:
will use another PR for long-term solution

PR #3028, opened 2024-09-24
fix: missing logs in log file

jackieluc:
Fixes missing logs in the `azure-cns.log` file from `main()` because they're using `log` instead of `logger`. `logger` is [initialized in the beginning of `main()`](https://github.com/Azure/azure-container-networking/blob/bee72ee95fc6d2c5c2e7cb91c623f1cdc3331285/cns/service/main.go#L536).

Validated by checking these log lines aren't in Kusto for ago > (7d) when it should be as it's part of CNS startup.


msvik (approved):

nddq (approved):

paulyufan2 (approved):
lgtm

PR #3029, opened 2024-09-24
fix: make ipamv2 metrics resilient to missing custom resource definitions

rbtr:
**Reason for Change**:

The metrics observer added in #2970 fails fast if a source of metrics is unavailable. This prevents legacy metrics from being collected _at all_ in scenarios like (as occurred) the ClusterSubnetState custom resource definition is not installed in the cluster.

In this change, the observer is made resilient to missing data sources or errors in the data collection, and will make a best-effort attempt to record the metrics even in case of partial data availability.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


nddq:
lgtm, although this behavior (potentially missing data sources/ errors that affects the metrics) is unique to ipam v2 right?

rbtr:
yes

nddq (approved):

PR #3033, opened 2024-09-26
fix: [WIN-NPM] HNS Remote Endpoint Fix

rejain456:
**Reason for Change**:

As we were booting up npm agent on windows, we saw ~4K endpoints on a node with 16 pods, when there should be a 1:1 endpoint to pod ratio.

**Issue Fixed**:

Changed logic: When booting up the data plane, instead of getting all remote endpoints, we only get local endpoints which will be the endpoints for the pods in the node.
Testing:
Took a node with 11 non-system pods on it.
Ran windows NPM with new image and saw total of 12 endpoints getting reset vs ~4K that were getting reset originally. 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/pkg/dataplane/dataplane_windows.go
+++ b/npm/pkg/dataplane/dataplane_windows.go
@@ -98,7 +98,7 @@ func (dp *DataPlane) bootupDataPlane() error {
 	}
 
 	// for backwards compatibility, get remote allEndpoints to delete as well
-	allEndpoints, err := dp.getAllPodEndpoints()
+	allEndpoints, err := dp.getLocalPodEndpoints()
 	if err != nil {
 		return err
 	}
--- a/npm/pkg/dataplane/dataplane_windows.go
+++ b/npm/pkg/dataplane/dataplane_windows.go
@@ -97,7 +97,6 @@ func (dp *DataPlane) bootupDataPlane() error {
 		return npmerrors.SimpleErrorWrapper("failed to initialize dataplane", err)
 	}
 
-	// for backwards compatibility, get remote allEndpoints to delete as well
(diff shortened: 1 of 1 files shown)


npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -98,7 +98,7 @@ func (dp *DataPlane) bootupDataPlane() error {
 	}
 
 	// for backwards compatibility, get remote allEndpoints to delete as well
```
huntergregory, line 100: can remove this comment now

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -98,7 +98,7 @@ func (dp *DataPlane) bootupDataPlane() error {
...
-	allEndpoints, err := dp.getAllPodEndpoints()
+	allEndpoints, err := dp.getLocalPodEndpoints()
```
huntergregory, line 101: Can we delete the old function if it's unused now?

huntergregory (changes requested):
Can you modify the PR name to specify that it's about windows NPM? We've used this syntax in the past: `fix: [WIN-NPM] ...`

Could you also add NPM and windows GH labels to this PR?

npm/pkg/dataplane/dataplane_windows.go
rejain456: removed it, thanks

huntergregory (approved):

huntergregory:
no need to run azp NPM pipelines for this PR since it modifies a windows-only file, and we don't have windows tests currently

rayaisaiah (approved):

PR #3037, opened 2024-09-27
feat: add codeql exec wrapper sinks

QxBytes:
**Reason for Change**:

Creates separate alerts for each instance of unsafe exec wrappers. 

For example, a usage of
`ExecutePowershellCommand( ... )` would previously have shown bunched with other alerts that call `exec.Command` inside the `ExecutePowershellCommand` wrapper function. However, with this change, `ExecutePowershellCommand` itself is treated as a sink, so if there is a flow, each instance of writing `ExecutePowershellCommand` will create a _separate_ error on the PR, which would not have been detected before. This also applies to `ExecuteRawCommand`.

Also moves common specification of a sink to a .qll file so we can import it elsewhere.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/codeql/acn-addipamconfig.ql
+++ b/codeql/acn-addipamconfig.ql
@@ -11,18 +11,7 @@
 
 // Detect inputs from CNS add ipam result / CNS multitenancy ipam add result to command injection
 import go
-
-private class Sink extends DataFlow2::Node {
-  Sink() {
-    exists(DataFlow::CallNode c |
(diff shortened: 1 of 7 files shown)


rbtr (approved):

rbtr:
it seems silly since nothing relevant was touched, but
/azp run Azure Container Networking PR

PR #3040, opened 2024-10-01
Upgrade to go1.22.7

aggarwal0009:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-ARG VARIANT="1.22.7"
+ARG VARIANT="1.22"
 FROM mcr.microsoft.com/vscode/devcontainers/go:${VARIANT}
 RUN apt-get update && \
     export DEBIAN_FRONTEND=noninteractive && \
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-ARG VARIANT="1.21"
+ARG VARIANT="1.22.7"
 FROM mcr.microsoft.com/vscode/devcontainers/go:${VARIANT}
 RUN apt-get update && \
     export DEBIAN_FRONTEND=noninteractive && \
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,7 +6,7 @@
 	"build": {
 		"dockerfile": "Dockerfile",
 		"args": {
-			"VARIANT": "1.22-bullseye",
+			"VARIANT": "1.22.7-bullseye",
 			"NODE_VERSION": "none"
 		}
 	},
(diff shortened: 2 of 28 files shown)


isaac-dasan (approved):

paulyufan2 (approved):

vakalapa (approved):

.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG VARIANT="1.21"
+ARG VARIANT="1.22.7"
```
rbtr, line 1: we don't need to overspecify this to a patch version
```suggestion
ARG VARIANT="1.22"
```

.devcontainer/devcontainer.json
```diff
@@ -1,11 +1,12 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
 // https://github.com/microsoft/vscode-dev-containers/tree/v0.203.0/containers/go
+// comment
```
rbtr, line 3: ?

.devcontainer/devcontainer.json
```diff
@@ -1,11 +1,12 @@
...
 {
 	"name": "Go",
 	"build": {
 		"dockerfile": "Dockerfile",
 		"args": {
-			"VARIANT": "1.21-bullseye",
+			"VARIANT": "1.22.7-bullseye",
```
rbtr, line 9:
```suggestion
			"VARIANT": "1.22-bullseye",
```

rbtr (changes requested):
i'm not going to leave the same comment on every line but it needs to be fixed everywhere

huntergregory (changes requested):
Lgtm after a green NPM conformance test run

.devcontainer/devcontainer.json
aggarwal0009: removed

aggarwal0009:
> i'm not going to leave the same comment on every line but it needs to be fixed everywhere

I thought the CVE fixes are in go v1.22.7. cc @miguelgoms for clarification

rbtr:
> > i'm not going to leave the same comment on every line but it needs to be fixed everywhere
> 
> I thought the CVE fixes are in go v1.22.7. cc @miguelgoms for clarification

@aggarwal0009 that's correct. by specifying ONLY 1.22, you will get the _latest patch of 1.22.x_ which is currently 1.22.8. 
The only place we need to explicitly say "minimum of 1.22.7" (and should actually say 1.22.8, since that is the latest release), is in the `go.mod`

aggarwal0009:
> > > i'm not going to leave the same comment on every line but it needs to be fixed everywhere
> > 
> > 
> > I thought the CVE fixes are in go v1.22.7. cc @miguelgoms for clarification
> 
> @aggarwal0009 that's correct. by specifying ONLY 1.22, you will get the _latest patch of 1.22.x_ which is currently 1.22.8. The only place we need to explicitly say "minimum of 1.22.7" (and should actually say 1.22.8, since that is the latest release), is in the `go.mod`

Updated.

PR #3047, opened 2024-10-03
fix: issue dhcp request on windows secondary interfaces

QxBytes:
**Reason for Change**:

We need to issue a dhcp request to the host in order to create a mapping for the secondary interface. The mapping allows for us to query the azure dns server on the secondary interface. The dhcp discover request occurs before the interface has the intended ip or is moved into the container namespace, but it **does** have an apipa autoconfigured ip. The auto assigned ip allows us to set the source ip of the packet so that the dhcp packet goes out on the interface that we want. We assume the interface always gets an auto configured ip. Then, once cni completes, we expect that secondary interface to be moved into the container.

Adds validation for nwCfg.Master since we issue exec commands with the field, so ensure this validation does not break anything else.

Reorganizes code into dhcp linux, windows and common.

Reorganizes retry logic into separate package

Confirm that we only run this code if we are on swiftv2 windows **(is the nic type condition correct?)**. Confirmed yes.

Tested:
- standalone prototype binary
- swiftv2 windows happy path (saw packet on secondary)
- swiftv2 time out on waiting for response will correctly give context deadline exceeded
- swiftv2 multiple pods can be created
- swiftv2 ips are assigned according to mtpnc
- https://msazure.visualstudio.com/One/_build/results?buildId=105427745&view=results

Limitations:
- Assumes auto configuration ip will always be assigned to the interface

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Related to #2989


dhcp/dhcp.go
```diff
@@ -0,0 +1,260 @@
...
+type ExecClient interface {
+	ExecuteCommand(ctx context.Context, command string, args ...string) (string, error)
+}
+
+type DHCP struct {
+	logger     *zap.Logger
+	execClient ExecClient
+}
+
+func New(logger *zap.Logger, plc ExecClient) *DHCP {
```
estebancams, line 59: I suggest NewDHCP since from consumer code New is too general

QxBytes: I was trying to adhere to this guideline during naming since the package is named DHCP and any reference to new will be like `dhcp.New`-- https://go.dev/doc/effective_go#package-names

dhcp/dhcp.go
estebancams, line 1: This sounds like too low level for us to implement. Any reason we are not using some public package like this one? https://github.com/insomniacslk/dhcp

dhcp/dhcp_windows.go
```diff
@@ -3,20 +3,181 @@ package dhcp
...
+		// so we have context.Background()
+		ret, cleanupErr := c.execClient.ExecuteCommand(context.Background(), "netsh", "interface", "ipv4", "delete", "address", ifName, dummyIPAddressStr)
+		if cleanupErr != nil {
+			c.logger.Info("Failed to remove dummy ip on leaving function", zap.String("output", ret), zap.Error(err))
+		}
+		// wait for nic to retrieve autoconfiguration ip
+		time.Sleep(returnDelay)
+	}()
+	// it takes time for the address to be assigned
+	time.Sleep(addIPAddressDelay)
```
estebancams, line 125: Don't we have a way to test for the new ip to be assigned instead of guessing a delay?

dhcp/dhcp.go
QxBytes: I tried to use that package initially, but the package seems to use the unix package, and so doesn't work on windows. Similar conversation: https://github.com/Azure/azure-container-networking/pull/2989#discussion_r1759609746

dhcp/dhcp_windows.go
```diff
@@ -3,20 +3,181 @@ package dhcp
...
+	return nil
 }
 
-func (c *DHCP) DiscoverRequest(_ context.Context, _ net.HardwareAddr, _ string) error {
+// issues a dhcp discover request on an interface by assigning an ip to that interface
+// then, sends a packet with that interface's dummy ip, and then unassigns the dummy ip
+func (c *DHCP) DiscoverRequest(ctx context.Context, macAddress net.HardwareAddr, ifName string) error {
+	// validate interface name
+	if !allowedInput.MatchString(ifName) {
+		return errors.New("invalid dhcp discover request interface name")
```
timraymond, line 99: Might want to constantize this. I'm surprised linters didnt' flag this.

dhcp/dhcp_windows.go
```diff
@@ -3,20 +3,181 @@ package dhcp
...
+	"github.com/Azure/azure-container-networking/retry"
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
+	"golang.org/x/sys/windows"
 )
 
-type DHCP struct {
-	logger *zap.Logger
+const (
+	dummyIPAddressStr    = "169.254.128.10"
```
timraymond, line 17: The protocol ossification that makes this necessary is depressing

dhcp/dhcp_windows.go
timraymond: Yeah, +1 to @estebancams . I'd rather have a fast check loop here (500ms sleep) then time out.

network/network_windows.go
```diff
@@ -433,8 +435,31 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *EndpointInfo, extIf *exter
 	return nw, nil
 }
 
+func (nm *networkManager) sendDHCPDiscoverOnSecondary(client dhcpClient, mac net.HardwareAddr, ifName string) error {
+	// issue dhcp discover packet to ensure mapping created for dns via wireserver to work
+	// we do not use the response for anything
+	numSecs := 15 // we need to wait for the address to be assigned
+	timeout := time.Duration(numSecs) * time.Second
```
timraymond, line 442: Just combine these lines so it's more readable:

```go
timeout := 15 * time.Second // reads "15 seconds"
```

retry/retry.go
```diff
@@ -0,0 +1,15 @@
+package retry
+
+import "time"
+
+func Do(f func() error, maxRuns, sleepMs int) error {
```
timraymond, line 5: I already have a flexible implementation of this in `nmagent/internal` that is probably worth promoting to this package if you need it too. It can support constant backoff, exponental, linear, maximums retries, whatever you like. It also respects contexts and deals with errors.

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -673,6 +675,17 @@ func (client *TransparentVlanEndpointClient) DeleteEndpointsImpl(ep *endpoint, _
...
 }
 
+// Creates a new retrier with a fixed delay, and treats all errors as retriable
+func (client *TransparentVlanEndpointClient) Retry(f func() error) error {
+	retrier := retry.Retrier{
+		Cooldown: retry.Max(numRetries, retry.Fixed(sleepDelay)),
+	}
+	return errors.Wrap(retrier.Do(context.Background(), func() error {
+		// we always want to retry, so all errors are temporary errors
+		return retry.WrapTemporaryError(f()) // nolint
```
timraymond, line 685: There are some errors that you've defined that I don't think quite fit the idea of a TemporaryError (one that could be resolved by retrying). For example, `errInvalidIPv4Address` is unlikely to be fixed without direct intervention, no?

You certainly can wrap everything and force everything to retry, but it's a blunt instrument.

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -673,6 +675,17 @@ func (client *TransparentVlanEndpointClient) DeleteEndpointsImpl(ep *endpoint, _
 	return nil
 }
 
+// Creates a new retrier with a fixed delay, and treats all errors as retriable
+func (client *TransparentVlanEndpointClient) Retry(f func() error) error {
+	retrier := retry.Retrier{
+		Cooldown: retry.Max(numRetries, retry.Fixed(sleepDelay)),
+	}
+	return errors.Wrap(retrier.Do(context.Background(), func() error {
```
timraymond, line 683: Are there no timeouts and such defined on the client that could be bound to a context?

network/transparent_vlan_endpointclient_linux.go
QxBytes: For transparent vlan, I wanted to preserve the existing behavior, which is all errors are considered retriable. I don't think `errInvalidIPv4Address` is used in transparent vlan client, or in the retry method.

timraymond (approved):
LGTM on @estebancams 's approval

timraymond:
@QxBytes stalebot is targeting this PR again. Anything else needed here?

QxBytes:
there is no repro to test this fix on at the moment

PR #3050, opened 2024-10-04
feat: add IPConfig state logs/metrics to Request/Release handlers

rbtr:
**Reason for Change**:

Logs and records IPAM metrics on Request/Release and MarkPendingRelease calls.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/restserver/metrics.go
```diff
@@ -142,8 +144,51 @@ func stateTransitionMiddleware(i *cns.IPConfigurationStatus, s types.IPState) {
...
+			state.releasingIPs++
+		}
+	}
+
+	logger.Printf("Allocated IPs: %d, Assigned IPs: %d, Available IPs: %d, PendingProgramming IPs: %d, PendingRelease IPs: %d",
+		state.allocatedIPs,
+		state.assignedIPs,
+		state.availableIPs,
+		state.programmingIPs,
+		state.releasingIPs,
```
timraymond, line 188: Worth it to implement `fmt.Formatter` for `ipState`? I feel like you'd always want to see that output when dumping that struct into a string.

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/Azure/azure-container-networking
 
-go 1.22.0
-
-toolchain go1.22.7
+go 1.23
```
timraymond, line 3: ❤️

cns/restserver/metrics.go
```diff
@@ -142,8 +144,51 @@ func stateTransitionMiddleware(i *cns.IPConfigurationStatus, s types.IPState) {
...
+	// programmingIPs are the IPs in state "PendingProgramming".
+	programmingIPs int64
+	// releasingIPs are the IPs in state "PendingReleasr".
+	releasingIPs int64
+}
+
+// publishIPStateMetrics logs and publishes the IP Config state metrics to Prometheus.
+func (service *HTTPRestService) publishIPStateMetrics() {
+	// copy state
+	service.RLock()
```
timraymond, line 163: Could this cause a slow hangup from the server if say:

G1: Finishes handler
G1: defer service.Unlock() runs
G1: publishIPStateMetrics() starts
G2: Starts handler
G2: service.Lock()
G2: sleep for an eternity
G2: defer service.Unlock() runs
G2: publishIPStateMetrics() runs to completion
G1: service.RLock()
G1: finishes publishIPStateMetrics() and releases locks
G1: writes 200 and closes socket now that the handler func is finally done

cns/restserver/metrics.go
rbtr: yes, but it shouldn't be dumped to a string anywhere else ever... is it worth it for private throwaway containers?

cns/restserver/metrics.go
rbtr: valid point, need to reconsider this. i hate just shoving more things in to the server Locked scope but the mutexing on this god object is not very precise

cns/restserver/metrics.go
timraymond: Yeah, probably not in that case. Just thinking about ways things could be different.

cns/restserver/metrics.go
```diff
@@ -142,11 +143,88 @@ func stateTransitionMiddleware(i *cns.IPConfigurationStatus, s types.IPState) {
...
+
+// publishIPStateMetrics logs and publishes the IP Config state metrics to Prometheus.
+func (service *HTTPRestService) publishIPStateMetrics() {
+	if recorder == nil {
+		recorder = &asyncMetricsRecorder{
+			podIPConfigSrc: service.PodIPConfigStates,
+			sig:            make(chan struct{}),
+		}
+		go recorder.run()
+	}
```
timraymond, line 217: I think the race detector might complain about this. Two goroutines read `recorder`, decide it's nil, then instantiate and assign to `recorder` and start a goroutine. N-1 of those will leak.

Not a huge deal, but `sync.Once` might make sense here.

rbtr: I didn't want to add another field to the HTTPRestService for this, what do you think about the once living in the singelton?

timraymond: I think that's probably fine

timraymond (approved):

PR #3056, opened 2024-10-07
feat: make OverlayExtensionConfig `ExtensionIPRange` field immutable

robogatikov:
**Reason for Change**:
We want to make OverlayExtensionConfig `extensionIPRange` field immutable to prevent accidental changes to the field. Network Containers are getting programmed for every IP address in this field's CIDR range, and DNC/DNC-RC will not support updating this field.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


timraymond:
Can you update the description to include "why?" There's not enough context here to understand why this particular field should be immutable as opposed to others.

robogatikov:
> Can you update the description to include "why?" There's not enough context here to understand why this particular field should be immutable as opposed to others.

Yeah, it was not very clear from the description. I hope it is better now.

nairashu (approved):

nairashu (approved):

PR #3057, opened 2024-10-08
feat: create a dummy NC to store secondary IPs in nodesubnet deployments with Cilium

santhoshmprabhu:
**Reason for Change**:
This PR introduces a "fake NC" that we use to integrate IPs for NodeSubnet into the CNS code. The members in this "NC" except the secondary IPs are hard-coded. These NCs are used to populate the CNS state in two situations - 
1. A new set of IPs are retrieved from the NMAgent
2. At startup, from the on-disk statefile.

In this PR, we have some of the APIs that will allow the 2 scenarios above to be implemented. Specifically, we have an API to create an NC request using a give list of secondary IPs, which will then be used to update the CNS state after a fetch from NMAgent. We also have an StateReconciler interface that creates functions for populating the CNS state from the on-disk statefile.

**Issue Fixed**:
NA

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


cns/ipam/interfaces.go
```diff
@@ -0,0 +1,11 @@
+package ipam
+
+import (
+	"github.com/Azure/azure-container-networking/cns"
+	cnstypes "github.com/Azure/azure-container-networking/cns/types"
+	"github.com/Azure/azure-container-networking/crd/nodenetworkconfig/api/v1alpha"
+)
+
+type StateReconciler interface {
```
timraymond, line 9: It's a little odd to create a new package only for this one interface that you need in one place. It's likely better to move this to `./cns/nodesubnet` and unexport it.

cns/nodesubnet/initialization.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/ipam"
+	"github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/Azure/azure-container-networking/cns/types"
+	"github.com/pkg/errors"
+	"golang.org/x/exp/maps"
+)
+
+func ReconcileInitialCNSState(_ context.Context, ipamReconciler ipam.StateReconciler, podInfoByIPProvider cns.PodInfoByIPProvider) (int, error) {
```
timraymond, line 14: I appreciate trying to include context. Hopefully we can refactor that Reconcile function some day to support it.

cns/restserver/internalapi.go
```diff
@@ -378,9 +380,11 @@ func (service *HTTPRestService) ReconcileIPAMState(ncReqs []*cns.CreateNetworkCo
 		}
 	}
 
-	if err := service.MarkExistingIPsAsPendingRelease(nnc.Spec.IPsNotInUse); err != nil {
-		logger.Errorf("[Azure CNS] Error. Failed to mark IPs as pending %v", nnc.Spec.IPsNotInUse)
-		return types.UnexpectedError
+	if nnc != nil {
```
timraymond, line 383: Curious, how did this work before? If nnc == nil currently, L384 would have panicked.

santhoshmprabhu: It is currently never nil. But it will be in the future, in another PR that I plan to open after this one. In nodesubnet, we don't use NNCs.

cns/nodesubnet/initialization.go
santhoshmprabhu: I had actually tried to incorporate context, gave up in the end an made it `_`. :)

cns/ipam/interfaces.go
santhoshmprabhu: This isn't specific to nodesubnet though. Other deployment modes also have an IPAM state reconciliation step. I could put it in ./cns/ if you think that's not too broad.

cns/restserver/internalapi.go
rbtr: there is only one caller which doesn't allow a nil NNC to proceed
https://github.com/Azure/azure-container-networking/blob/79c2af56a536fff2205a1c90521f20616494708e/cns/service/main.go#L1134-L1143
I'm not opposed to re-checking in this func since we're passing a pointer around, but a nil NNC should be a fatal error and we should _immediately_ validate the inputs instead of getting mostly done with the work before we check it. 
Passing a nil ref is not a sustainable strategy for configuring behavior. It's hard to discover, except accidentally when we expect one behavior and are getting another because we _accidentally_ have a nil and should have errored earlier instead.

cns/ipam/interfaces.go
rbtr: Perhaps what Tim is trying to say is more that in Go interfaces are defined by the consumer, not the provider, to describe the behavior that an object must have to be _passed in to them_.  It usually doesn't make sense to have a public interface in a separate package like this. 

I don't see anything preventing this from being a private interface in the nodesubnet package, why does it matter that other modes have IPAM state reconciliation?

cns/restserver/internalapi.go
santhoshmprabhu: @rbtr , As you know, in Nodesubnet, we won't have any NNCs created, so this change is really to facilitate the use of this method to reconcile IPAM state for Nodesubnet, when no NNCs exist. Now, given we don't want nil as a parameter, I could pass an "empty" NNC, and then replace this check with a check on the `ncReq`, so that we don't end up trying to do NNC-specific  things in nodesubnet. Or I could just create a `ReconcileIPAMStateForNodesubnet` method that'd be substantially similar to this function, but without the NNC parameter. Shall I go with the empty NNC approach?

cns/ipam/interfaces.go
santhoshmprabhu: I can make it private, but that will leave two different private interfaces that are exactly the same - there's[this one](https://github.com/Azure/azure-container-networking/blob/master/cns/service/main.go#L1126) as well as what we are now discussing. If we keep this exposed, we can remove the other private interface in my next PR that will actually finish the implementation of nodesubnet.

rbtr: The point of [this interface](https://github.com/Azure/azure-container-networking/blob/master/cns/service/main.go#L1127C2-L1127C159) is to abstract the inputs in to [this function](https://github.com/Azure/azure-container-networking/blob/master/cns/service/main.go#L1132). _Anything_ that satisfies that interface signature is an acceptable input into that function. That's the only purpose that this interface definition exists for. Your code in a different package should define and accept a local interface which describes the behavior that it needs. If that happens to be the same right now, that's fine.
This has some context on the antipattern that we're trying to prevent from developing here: https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a

timraymond: Also relevant from the Go proverbs: ["A little copying is better than a little dependency"](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)

santhoshmprabhu: Done.

cns/restserver/internalapi.go
santhoshmprabhu: @rbtr, thoughts?

cns/nodesubnet/initialization.go
```diff
@@ -0,0 +1,36 @@
...
+	"golang.org/x/exp/maps"
+)
+
+type ipamReconciler interface {
+	ReconcileIPAMStateForNodeSubnet(ncRequests []*cns.CreateNetworkContainerRequest, podInfoByIP map[string]cns.PodInfo) cnstypes.ResponseCode
+}
+
+func ReconcileInitialCNSState(_ context.Context, ipamReconciler ipamReconciler, podInfoByIPProvider cns.PodInfoByIPProvider) (int, error) {
+	// Get previous PodInfo state from podInfoByIPProvider
+	podInfoByIP, err := podInfoByIPProvider.PodInfoByIP()
```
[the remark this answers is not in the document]
santhoshmprabhu: No. That is asynchronous. What this does is, it reads IPs from the statefile, assumes that they are valid, and creates entries for them in the pool. After this is done, we will start the async refresh, which will fetch the full list of IPs from NMAgent and add them to the pool. The conflist generation is blocked until the NMAgent fetch happens. So this is just step 1 in populating the IP pool.

cns/restserver/internalapi.go
santhoshmprabhu: The function is now broken into smaller functions, and we don't use the nil parameter anymore. Resolving this thread to allow merge to proceed. We have a todo in the code to address any further feedback.

cns/NetworkContainerContract.go
```diff
@@ -55,6 +55,7 @@ const (
 	JobObject              = "JobObject"
 	COW                    = "COW" // Container on Windows
 	BackendNICNC           = "BackendNICNC"
+	NodeSubnet             = "NodeSubnet" // Artificial Container Type for NodeSubnet
```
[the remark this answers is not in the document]
santhoshmprabhu: Done.

cns/NetworkContainerContract.go
```diff
@@ -66,6 +68,7 @@ const (
 	AzureFirstParty = "AzureFirstParty"
 	KubernetesCRD   = "KubernetesCRD"
 	// TODO: Add OrchastratorType as CRD: https://msazure.visualstudio.com/One/_workitems/edit/7711872
+	KubernetesNodeSubnet = "KubernetesNodeSubnet"
```
tamilmani1989, line 71: why do we need this. why can't we reuse `KubernetesCRD`

tamilmani1989: may be we can keep this and remove nodesubnet constant..

santhoshmprabhu: I had not used CRD because at the end of the day there is no CRD being used here. We can use this string instead of NodeSubnet, but they represent two different things - one is NC type, the other is Orchestrator type. It might be confusing to use the same string for both.

santhoshmprabhu: Discussed with @tamilmani1989, removing newly introduced constants.

tamilmani1989 (approved):

PR #3062, opened 2024-10-10
[DO NOT MERGE] Arm64 new sku testing Standard_D4ps_v6

pjohnst5:
This is just to test a new sku coming out in azure (arm64) : Standard_D4ps_v6

We're just making sure we understand what fails/succeeds, as this is going GA soon


jpayne3506:
@pjohnst5 we test arm64 skus on CNI Release Test also it is a little bit more through/intense. Perhaps test it on there as well? 😃

PR #3064, opened 2024-10-10
Changing CNS IPAM to allow for dualstack based on Family type

rjdenney:
**Reason for Change**:

This change allows us to have dualstack vnet block clusters using only one NC while still allowing for dualstack overlay clusters with two NCs.

Adds an IP family type that is added to an NC during reconciling. This IP family type is then check across all NCs in ipam to ensure that we hand out the appropriate number of IPs.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


timraymond:
Please mark this as draft if you don't intend to merge it yet

PR #3065, opened 2024-10-10
ci: add wait before adding windows nodepool if cluster is in updating state

QxBytes:
**Reason for Change**:

If a windows nodepool is added to a newly created aks cluster, the aks cluster may be in "Updating" state, causing the job to fail. This PR adds a check to see if the cluster is in updating state and sleeps until the cluster is no longer in the updating state. Also refactors adding windows nodepool into a common template and adds a check if the windows nodepool is already created (so we don't error if the nodepool is already added).

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
On detecting updating state, we wait, see line 54:
https://msazure.visualstudio.com/One/_build/results?buildId=105433502&view=logs&j=66af8358-44a8-5a80-90ff-77e03c31665a&t=8f8dc4ee-a8ef-5fc9-ab65-cc330785c488
When not detecting updating state, we proceed:
https://msazure.visualstudio.com/One/_build/results?buildId=105437613&view=logs&s=46b305b8-7ec3-51e0-83b8-e4b201974b98&j=66af8358-44a8-5a80-90ff-77e03c31665a

### DIFF
--- a/.pipelines/templates/add-windows-nodepool-job.yaml
+++ b/.pipelines/templates/add-windows-nodepool-job.yaml
@@ -13,6 +13,7 @@ jobs:
     demands:
     - agent.os -equals Linux
     - Role -equals $(CUSTOM_E2E_ROLE)
+  timeoutInMinutes: 30
   steps:
     - task: AzureCLI@2
       inputs:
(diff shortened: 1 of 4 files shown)


.pipelines/templates/add-windows-nodepool-job.yaml
```diff
@@ -0,0 +1,62 @@
...
+              --name "${{ parameters.clusterName }}" \
+              --resource-group "${{ parameters.clusterName }}" \
+              --query provisioningState)
+              
+              if echo "$cluster_state" | grep -q "Updating"; then
+                echo "Cluster is updating. Sleeping for 30 seconds..."
+                sleep 30
+              else
+                break
+              fi
```
jpayne3506, line 48: We should output `$cluster_state` once we are complete with the loop. The current way the logic is setup is only capturing if the state is of `Updating*`. In the future we may want to capture other states and it would be nice to know which one just by looking at the logs.

jpayne3506 (commented):
Thanks for templatizing this 😄

jpayne3506 (approved):

PR #3069, opened 2024-10-14
test: add uts for cni add function

QxBytes:
**Reason for Change**:

Adds unit tests to test that endpoint info struct creation is as expected based on args, cns/ipam invoker response, and network configuration for the following scenarios:
- swiftv2 linux ipam response (infra + delegated) -> endpoint info
- swiftv2 windows ipam response (infra + delegated) -> endpoint info
- swiftv1 windows multitenancy dual nic ipam response  (infra + infra) -> endpoint info
- swiftv1 linux multitenancy (infra) ipam response -> endpoint info

Each scenario tests the `Add` method from ipam response until we call an `EndpointCreate`, given args, ipam response, and network configuration and ensure the endpoint info struct generated _exactly matches_ what is expected (so all fields are checked) **except** for the interface name and endpoint id as they are nondeterministic identifiers generated during runtime.

Adds a check to ensure each endpoint info's Option and Data fields are separate.
Adds a check to ensure swiftv2 linux's endpoint info interface name is overwritten with the correct interface name.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
This does not add tests for swiftv2 ib. It also does not explicitly add tests for swiftv1 non multitenancy scenarios as they are mostly covered by the multitenancy ones (singletenancy is just a single ipam response). This does not test anything that happens inside `EndpointCreate`.

### DIFF
--- a/cni/network/network_linux_test.go
+++ b/cni/network/network_linux_test.go
@@ -5,7 +5,6 @@ package network
 
 import (
 	"fmt"
-	"log"
 	"net"
 	"regexp"
 	"testing"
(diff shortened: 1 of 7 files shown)


cni/network/network_linux_test.go
```diff
@@ -160,3 +168,387 @@ func TestAddSnatForDns(t *testing.T) {
...
+					}
+					// save the endpoint id before removing it
+					epID = endpointInfo.EndpointID
+					require.Regexp(t, regexp.MustCompile(wantedEndpointEntry.epIDRegex), epID)
+
+					// omit endpoint id and ifname fields as they are nondeterministic
+					endpointInfo.EndpointID = ""
+					endpointInfo.IfName = ""
+
+					require.Equal(t, wantedEndpointEntry.epInfo, endpointInfo)
```
jpayne3506, line 526: Why do we compare twice? Why do we omit these fields and then compare?

QxBytes: The tt.match is to identify which item in the map we are looking at, so that we compare the full endpoint info struct to the correct item in the map. The map ordering is nondeterministic.

We omit the endpoint id and interface name because they are based on which interface info is processed first in the map, and what interface info is processed first is nondeterministic since it is a map.

jpayne3506 (approved):

paulyufan2 (approved):

PR #3073, opened 2024-10-18
feat: support for cilium + nodesubnet

santhoshmprabhu:
**Reason for Change**:
This PR completes the implementation of CNS support for nodesubnet, so that we may use Cilium with nodesubnet. This PR includes the following changes:
1. "AzureHost" channel mode to configure fetch of IP addresses from NMAgent into CNS
2. Asynchronous IP fetcher in CNS to do the above IP fetch
3. Statefile reconciliation without NNCs

The changes have been validated with E2E runs, though the pipeline changes are to be merged in a separate PR (#3083).

**Issue Fixed**:

Cilium nodesubnet support

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


cns/restserver/nodesubnet.go
```diff
@@ -0,0 +1,59 @@
...
+	for i, ip := range secondaryIPs {
+		secondaryIPStrs[i] = ip.String()
+	}
+
+	networkContainerRequest := nodesubnet.CreateNodeSubnetNCRequest(secondaryIPStrs)
+
+	code, msg := service.saveNetworkContainerGoalState(*networkContainerRequest)
+	if code != types.Success {
+		logger.Debugf("Error in processing IP change")
+		return errors.Errorf("failed to save fetched ips. code: %d, message %s", code, msg)
```
timraymond, line 28: I am very much of the mindset that you should log or return errors--not both. Since we _can_ return an error here, I lean more towards just doing that. We can add more contextual information (as you've done), and it can be someone else's responsibility towards the top of the call stack to log all the errors (usually the HTTP handler).

cns/service/main.go
```diff
@@ -871,6 +872,31 @@ func main() {
...
+
+		// If cns manageendpointstate is true, then cns maintains its own state and reconciles from it.
+		// in this case, cns maintains state with containerid as key and so in-memory cache can lookup
+		// and update based on container id.
+		cns.GlobalPodInfoScheme = cns.InfraIDPodInfoScheme
+
+		var podInfoByIPProvider cns.PodInfoByIPProvider
+		podInfoByIPProvider, err = getPodInfoByIPProvider(rootCtx, cnsconfig, httpRemoteRestService, nil, "")
+		if err != nil {
+			logger.Errorf("[Azure CNS] Failed to get PodInfoByIPProvider: %v", err)
```
timraymond, line 889: If this is a structured log, tags like `[Azure CNS]` should be their own K/V pair. This might also be handled implicitly by extra information logged by kusto... so we might just be adding bytes for little gain.

cns/restserver/nodesubnet.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"context"
+	"net/netip"
+
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/logger"
+	nodesubnet "github.com/Azure/azure-container-networking/cns/nodesubnet"
+	"github.com/Azure/azure-container-networking/cns/types"
+	errors "github.com/pkg/errors"
```
timraymond, line 11: I don't think you need the alias here

```suggestion
	"github.com/pkg/errors"
```

cns/restserver/nodesubnet_test.go
```diff
@@ -0,0 +1,78 @@
...
+		t.Errorf("InitializeNodeSubnet returned an error: %v", err)
+	}
+
+	if service.GetNodesubnetIPFetcher() == nil {
+		t.Error("NodeSubnetIPFetcher is not initialized")
+	}
+
+	select {
+	case <-ctx.Done():
+		t.Error("Test context was canceled before conflist generation")
```
timraymond, line 60: Not necessarily because it was canceled... it's worth logging `t.Err()` to find out. Those canceling can specify a cause and that will be logged. It just makes it a little more flexible.

cns/restserver/nodesubnet_test.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil
+}
+
+func (m *MockCNIConflistGenerator) Close() error {
+	// Implement the Close method logic here if needed
+	return nil
+}
+
+func TestNodeSubnet(t *testing.T) {
+	mockPodInfoProvider := new(MockPodInfoByIPProvider)
```
timraymond, line 35: I know it's weird, but nobody uses `new`. The more common style is `&MockInfoByIPProvider{}`.

The only _caveat_ to that is related to generics (so as of Go 1.18). Since it's impossible to say "give me a zero value of type T", the only way to do it is `new(T)`.

cns/restserver/nodesubnet_test.go
```diff
@@ -0,0 +1,78 @@
...
+	return res, nil
+}
+
+// Mock implementation of CNIConflistGenerator
+type MockCNIConflistGenerator struct {
+	GenerateCalled chan bool
+}
+
+func (m *MockCNIConflistGenerator) Generate() error {
+	m.GenerateCalled <- true
```
timraymond, line 25: Usually for a latch like this, it's more common to do `chan struct{}`. Instead of sending a value on it, you just close it (`close(m.GenerateCalled)`). The difference is that sending a value will wake up one waiting goroutine (once this goroutine is able to perform the write), where closing will wake up all of them. This is usually what you want when writing tests like this.

cns/restserver/nodesubnet_test.go
```diff
@@ -0,0 +1,78 @@
...
+	defer service.Service.Uninitialize()
+
+	ctx, cancel := testContext(t)
+	defer cancel()
+
+	err := service.InitializeNodeSubnet(ctx, mockPodInfoProvider)
+	service.StartNodeSubnet(ctx)
+
+	if err != nil {
+		t.Errorf("InitializeNodeSubnet returned an error: %v", err)
```
timraymond, line 51: You might want to make this a `t.Fatalf`. When you have an error, it's worth asking "would making assertions on any values after this point be complete nonsense?" If the answer is "yeah, probably," then just Fatal the test. It reduces the test noise that you have to sift through to figure out what went wrong.

On the other hand, if you _can_ sensibly make multiple assertions and do want them logged, `Error` is great.

cns/restserver/nodesubnet_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+func TestNodeSubnet(t *testing.T) {
+	mockPodInfoProvider := new(MockPodInfoByIPProvider)
+
+	// Create a real HTTPRestService object
+	mockCNIConflistGenerator := &MockCNIConflistGenerator{
+		GenerateCalled: make(chan bool),
+	}
+	service := restserver.GetRestServiceObjectForNodeSubnetTest(mockCNIConflistGenerator)
+	defer service.Service.Uninitialize()
```
timraymond, line 42: You might consider using `testing.(*T).Cleanup` for these

cns/restserver/restserver.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/dockerclient"
 	"github.com/Azure/azure-container-networking/cns/logger"
 	"github.com/Azure/azure-container-networking/cns/networkcontainers"
+	nodesubnet "github.com/Azure/azure-container-networking/cns/nodesubnet"
```
timraymond, line 16: Same thing here. I don't think the alias is necessary since the package's name will be "nodesubnet" anyway.

```suggestion
	"github.com/Azure/azure-container-networking/cns/nodesubnet"
```

.pipelines/singletenancy/cilium-nodesubnet/cilium-nodesubnet-e2e-step-template.yaml
```diff
@@ -57,6 +57,7 @@ steps:
         pwd
         kubectl cluster-info
         kubectl get po -owide -A
+        CILIUM_VERSION_TAG=${CILIUM_NODE_SUBNET_VERSION}
```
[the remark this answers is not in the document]
santhoshmprabhu: `CILIUM_NODE_SUBNET_VERSION` is currently 1.14.10. Longer term the plan is to support Cilium 1.16. @tamilmani1989, is that still the plan or should we default to CILIUM_VERSION_TAG?

hack/aks/Makefile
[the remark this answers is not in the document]
santhoshmprabhu: Done

cns/restserver/nodesubnet.go
```diff
@@ -0,0 +1,59 @@
...
+	secondaryIPStrs := make([]string, len(secondaryIPs))
+	for i, ip := range secondaryIPs {
+		secondaryIPStrs[i] = ip.String()
+	}
+
+	networkContainerRequest := nodesubnet.CreateNodeSubnetNCRequest(secondaryIPStrs)
+
+	code, msg := service.saveNetworkContainerGoalState(*networkContainerRequest)
+	if code != types.Success {
+		logger.Debugf("Error in processing IP change")
```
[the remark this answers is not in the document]
santhoshmprabhu: Removed this in response to Tim's comment. The error will be eventually logged by the fetcher.

cns/restserver/nodesubnet_test.go
santhoshmprabhu: Done, please check again.

cns/restserver/nodesubnet_test.go
santhoshmprabhu: Moved the assertions that could be made independent of this to be checked before this, and made this Fatal.

cns/restserver/nodesubnet_test.go
santhoshmprabhu: Done.

cns/restserver/nodesubnet.go
santhoshmprabhu: Done.

cns/restserver/helper_for_nodesubnet_test.go
```diff
@@ -0,0 +1,80 @@
+package restserver
```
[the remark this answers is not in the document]
santhoshmprabhu: This is actually a trick to have extra methods in the `restserver` package that are part of testing only. We want to keep the test in a separate package, but want to have extra helper methods to create objects and validate results for the test.

cns/service/main.go
```diff
@@ -661,6 +661,8 @@ func main() {
 	}
 	if isManaged, ok := acn.GetArg(acn.OptManaged).(bool); ok && isManaged {
 		config.ChannelMode = cns.Managed
+	} else if cnsconfig.ChannelMode == cns.AzureHost {
```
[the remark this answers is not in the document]
santhoshmprabhu: Removed it.

test/integration/cilium-nodesubnet/ipconfigupdate.go
```diff
@@ -104,8 +104,12 @@ func main() {
 			for i := 2; i <= secondaryConfigCount+1; i++ {
 				ipConfig := make(map[string]interface{})
 				for k, v := range primaryIPConfig {
+					if k == "loadBalancerBackendAddressPools" {
```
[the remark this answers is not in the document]
santhoshmprabhu: Added a comment.

.pipelines/singletenancy/cilium-nodesubnet/cilium-nodesubnet-e2e-step-template.yaml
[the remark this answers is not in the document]
santhoshmprabhu: Checked offline with Tamilmani, we do want only 1.14 and higher.

.pipelines/pipeline.yaml
```diff
@@ -261,6 +261,17 @@ stages:
         k8sVersion: ""
         dependsOn: "containerize"
 
+    # Cilium Nodesubnet E2E tests
```
behzad-mir, line 264: I recommend breaking this to two separate PRs. One about CNS changes and one regarding the pipeline for two reasons:
1- This will make the roll back easier in case we encounter some issues.
2- It makes the reviews easier as well.

cns/restserver/nodesubnet.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	logger.Debugf("IP change processed successfully")
+
+	// saved NC successfully, generate conflist to indicate CNS is ready
+	go service.MustGenerateCNIConflistOnce()
+	return nil
+}
+
+func (service *HTTPRestService) InitializeNodeSubnet(ctx context.Context, podInfoByIPProvider cns.PodInfoByIPProvider) error {
```
behzad-mir, line 37: Missing comments for this func

cns/restserver/nodesubnet.go
```diff
@@ -0,0 +1,58 @@
...
+	// Set orchestrator type
```
behzad-mir, line 38: comment should start with lowercase letters

cns/service/main.go
```diff
@@ -871,6 +870,32 @@ func main() {
 		}
 	}
 
+	// AzureHost channelmode indeicates Nodesubnet. IPs are to be fetched from NMagent.
```
behzad-mir, line 873: indicates

test/integration/cilium-nodesubnet/ipconfigupdate.go
QxBytes, line 1: Also just noticed the `runCommand` function is using `bash` in the first parameter. While this is internal/non-exposed/testing, maybe we could make the function more specific (ex: `runAzCommand`) and pass in each parameter individually at some point.

cns/service/main.go
```diff
@@ -1505,6 +1506,50 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 	return nil
 }
 
+func getPodInfoByIPProvider(
```
behzad-mir, line 1509: Missing comments

behzad-mir (changes requested):

.pipelines/pipeline.yaml
santhoshmprabhu: Moved this and other pipeline-related changes out.

cns/restserver/nodesubnet.go
```diff
@@ -0,0 +1,62 @@
...
+
+	code, msg := service.saveNetworkContainerGoalState(*networkContainerRequest)
+	if code != types.Success {
+		return errors.Errorf("failed to save fetched ips. code: %d, message %s", code, msg)
+	}
+
+	logger.Debugf("IP change processed successfully")
+
+	// saved NC successfully, generate conflist to indicate CNS is ready
+	go service.MustGenerateCNIConflistOnce()
```
tamilmani1989, line 33: why cns starts go routine to generate conflist?

tamilmani1989, line 33: does cns generate conflist after it fetches secondary ip from nmagent?

cns/restserver/nodesubnet_test.go
```diff
@@ -0,0 +1,76 @@
...
+	service.StartNodeSubnet(ctx)
+
+	if service.GetNodesubnetIPFetcher() == nil {
+		t.Error("NodeSubnetIPFetcher is not initialized")
+	}
+
+	if err != nil {
+		t.Fatalf("InitializeNodeSubnet returned an error: %v", err)
+	}
+
```
tamilmani1989, line 55: inaddition to checking error, can it validate cache/memory state and check if expected?

cns/service/main.go
```diff
@@ -1019,6 +1045,10 @@ func main() {
 		}(privateEndpoint, infravnet, nodeID)
 	}
 
+	if config.ChannelMode == cns.AzureHost {
+		httpRemoteRestService.StartNodeSubnet(rootCtx)
```
tamilmani1989, line 1049: why dont we call startnodesubnet after initializenodesubnet?

cns/service/main.go
```diff
@@ -1249,40 +1279,11 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
-			if err != nil {
-				return nil, errors.Wrap(err, "failed to list Pods for PodInfoProvider")
-			}
-			podInfo, err := cns.KubePodsToPodInfoByIP(pods.Items)
-			if err != nil {
-				return nil, errors.Wrap(err, "failed to convert Pods to PodInfoByIP")
-			}
-			return podInfo, nil
-		})
+	podInfoByIPProvider, err := getPodInfoByIPProvider(ctx, cnsconfig, httpRestServiceImplementation, clientset, nodeName)
```
tamilmani1989, line 1282: node subnet uses `httpRemoteRestService` and this one use `httpRestServiceImplementation`. whats the difference between the two

cns/restserver/nodesubnet.go
santhoshmprabhu: This is mimicking how Swift handles conflist generation. As far as I can tell, it's safe to do this in a goroutine, but it could also be done in the calling thread rather than a separate goroutine. Moreover, in nodesubnet case, this call already happens in a goroutine that handles asynchronous IP fetches.

santhoshmprabhu: Yes. This method is invoked upon each new IP fetch from the NMAgent. The conflist generation is a no-op after the first invocation.

cns/service/main.go
santhoshmprabhu: We want to wait till the service http service is running, and we catch up on all async deletes.

cns/service/main.go
santhoshmprabhu: It is only a difference in variable names. In both cases, we are passing a variable of type `*restserver.HTTPRestService`. For nodesubnet, the invocation happens from `main`, and the variable `httpRemoteRestService` is defined there. In this case (Swift), the invocation is happening inside another function - `InitializeCRDState`. The parameters and local variables are named differently as compared to `main`.

cns/restserver/nodesubnet.go
```diff
@@ -0,0 +1,62 @@
...
+	return nil
+}
+
+// initialization steps for nodesubnet:
+// 1. Set orchestrator type to KubernetesCRD
+// 2. Reconcile initial CNS state from statefile
+// 3. Create the IP fetcher
```
timraymond, line 40: Minor, but as you have these, these will become doc comments. Generally the style for doc comments is a classic definition. Something like:

// NameOfTheThing is a <insert definition here>.

In this case:

```suggestion
// InitializeNodeSubnet prepares CNS for serving NodeSubnet requests.
// It sets the orchestrator type to KubernetesCRD, reconciles the initial
// CNS state from the statefile, then creates an IP fetcher.
```

cns/restserver/nodesubnet.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/logger"
+	nodesubnet "github.com/Azure/azure-container-networking/cns/nodesubnet"
+	"github.com/Azure/azure-container-networking/cns/types"
+	"github.com/pkg/errors"
+)
+
+var _ nodesubnet.IPConsumer = &HTTPRestService{}
+
+// Implement the UpdateIPsForNodeSubnet method for HTTPRestService
```
timraymond, line 16: Same thing here, this should be a definition of `UpdateIPsForNodeSubnet`.

cns/restserver/nodesubnet.go
```diff
@@ -0,0 +1,61 @@
...
+		return errors.Wrap(err, "reconcile initial CNS state")
+	}
+	// statefile (if any) is reconciled. Initialize the IP fetcher. Start the IP fetcher only after the service is started,
+	// and any pending async delete operations are completed.
+	service.nodesubnetIPFetcher = nodesubnet.NewIPFetcher(service.nma, service, 0, 0, logger.Log)
+
+	return nil
+}
+
+func (service *HTTPRestService) StartNodeSubnet(ctx context.Context) {
```
behzad-mir, line 59: Comment needs to be added

cns/restserver/nodesubnet_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"testing"
+
+	"github.com/Azure/azure-container-networking/cns/cnireconciler"
+	"github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/Azure/azure-container-networking/cns/restserver"
+	"github.com/Azure/azure-container-networking/cns/types"
+	"github.com/Azure/azure-container-networking/store"
+)
+
+func getMockStore() store.KeyValueStore {
```
behzad-mir, line 15: missing comment

cns/restserver/nodesubnet_test.go
```diff
@@ -0,0 +1,140 @@
...
+	close(m.GenerateCalled)
+	return nil
+}
+
+func (m *MockCNIConflistGenerator) Close() error {
+	// Implement the Close method logic here if needed
+	return nil
+}
+
+func TestNodeSubnet(t *testing.T) {
```
behzad-mir, line 70: Please add comments for all the new funcs

cns/restserver/nodesubnet_test.go
```diff
@@ -0,0 +1,140 @@
...
+	podInfoByIPProvider, err := cnireconciler.NewCNSPodInfoProvider(getMockStore())
+	if err != nil {
+		t.Fatalf("NewCNSPodInfoProvider returned an error: %v", err)
+	}
+
+	// Create a real HTTPRestService object
```
behzad-mir, line 76: comments within the func needs to start lowercase

cns/service/main.go
```diff
@@ -871,6 +870,32 @@ func main() {
...
 	}
 
+	// AzureHost channelmode indicates Nodesubnet. IPs are to be fetched from NMagent.
+	if config.ChannelMode == cns.AzureHost {
+		if !cnsconfig.ManageEndpointState {
+			logger.Errorf("ManageEndpointState must be set to true for AzureHost mode")
+			return
+		}
+
+		// If cns manageendpointstate is true, then cns maintains its own state and reconciles from it.
```
behzad-mir, line 880: Needs to start lowercase letters

cns/restserver/helper_for_nodesubnet_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/common"
+	"github.com/Azure/azure-container-networking/cns/fakes"
+	"github.com/Azure/azure-container-networking/cns/nodesubnet"
+	acn "github.com/Azure/azure-container-networking/common"
+	"github.com/Azure/azure-container-networking/nmagent"
+	"github.com/Azure/azure-container-networking/store"
+)
+
+func GetRestServiceObjectForNodeSubnetTest(t *testing.T, generator CNIConflistGenerator) *HTTPRestService {
```
behzad-mir, line 17: plz add comments

behzad-mir (changes requested):

cns/service/main.go
santhoshmprabhu: @behzad-mir this code is actually refactored, this comment was simply moved here as part of the refactoring. There is this and a few other comments that are already in our code that seem to start with an upper case letter. Just trying to be sure before editing pre-existing comments. Do we have some kind of a style guide about lower case in comments inside functions? I tried finding guidelines online, but couldn't find any for this case.

timraymond: @behzad-mir I'm not aware of any internal style restricting casing on comments within functions. Generally, yes, they're lowercase, but I'm also fine with anything that helps get the point across. We can use all of UTF-8 here.

The only style guide we _should_ follow is https://go.dev/doc/comment . I also very much appreciate when people leave doc comments, but we have so many undocumented ones that I understand when we don't.

cns/restserver/nodesubnet_test.go
timraymond: In the past decade, I don't think I've ever seen a doc comment for a test. I'm not even sure if `go doc` would do anything useful for them. If the test is doing something particularly obtuse, I try to either narrate as I go or drop a paragraph just below the function signature explaining what's going on in the test.

cns/service/main.go
tamilmani1989: then why dont we use  same variable?

cns/service/main.go
tamilmani1989: is async delete happen as part of http service start? i didnt understand clearly. can you add a comment why it should be started after http service start?

cns/restserver/nodesubnet.go
tamilmani1989: so its called after we get atleast one ip from nma? lets comment that

tamilmani1989: i didnt understand why this should be called in a separate go routine

cns/service/main.go
santhoshmprabhu: I've edited the comments that I added,  left the comments that were already present untouched.

cns/service/main.go
santhoshmprabhu: As I mentioned, these invocations are from two different places. So there is no same variable.  One invocation is from `main`, the other is from another function.

cns/restserver/nodesubnet.go
santhoshmprabhu: It doesn't have to be. I've switched it to a simple function call.

santhoshmprabhu: Done

cns/service/main.go
santhoshmprabhu: Async delete is done by having the HTTP service running, and then submitting the pending delete requests to the service. Added a comment.

tamilmani1989: couple of questions:
1. what happens if we start nodesubnet before async delete?
2. how do we make sure async delete is called before this call?

santhoshmprabhu: 1. I don't think there will be a major problem, I anticipate the pods pending deletion will stay in the in-memory state and then eventually deleted when async delete catche up
2. We do async delete a few lines before in this very function, before we call `StartNodeSubnet`

santhoshmprabhu: @tamilmani1989 If you think I should start nodesubnet early, I could try that, but I suspect we might run into some other issues if the HTTP service isn't running yet. In my implementation so far I have always finished async delete before starting nodesubnet.

santhoshmprabhu: @tamilmani1989 I tried starting nodesubnet early and it does work without errors. Then I recalled the other reason I did this - `StartNodeSubnet` will cause the conflist to be generated, which means CNS needs to be ready to handle IPAM queries. That doesn't happen until rest service is running. So conflist generation (and hence StartNodeSubnet) has to wait until then.

timraymond:
LGTM. Deferring approval to others with outstanding comments.

tamilmani1989 (approved):

behzad-mir (approved):

PR #3075, opened 2024-10-18
feat: why not proto

paulgmiller:
**Reason for Change**:
Proto reduces memory and bandwidth needed on large clusters with initial watch? Maybe haven't rtested this

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/cmd/start_server.go
+++ b/npm/cmd/start_server.go
@@ -81,6 +81,9 @@ func startControlplane(config npmconfig.Config, flags npmconfig.Flags) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


rbtr:
I don't see a reason we shouldn't do this everywhere we use client-go, given we have pretty narrow k8s version support windows. Probably also impactful for retina cc @vakalapa

vakalapa:
We can get this in, but there was less to minute differences in scale or performance when Hunter tested it.

PR #3083, opened 2024-10-23
test: pipeline changes for cilium nodesubnet

santhoshmprabhu:
**Reason for Change**:
This PR introduces the pipeline changes to test Cilium+Nodesubnet E2E. The code changes are in #3073. This PR includes:
1. Inclusion of Cilium+Nodesubnet E2E scenario in the PR pipeline
2. Updates to IP masq agent config to skip SNAT for cluster IPs
3. Better Go script to update IP configs in pipeline-createdCilium nodesubnet clusters
4. Minor changes to hack/Makefile to allow use of standard LB

**Issue Fixed**:
Pipeline changes for Cilium nodesubnet

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This PR should not be mreged before #3073


.pipelines/singletenancy/cilium-nodesubnet/cilium-nodesubnet-e2e-step-template.yaml
```diff
@@ -57,6 +57,7 @@ steps:
         pwd
         kubectl cluster-info
         kubectl get po -owide -A
+        CILIUM_VERSION_TAG=${CILIUM_NODE_SUBNET_VERSION}
```
vipul-21, line 60: Following the conversation from the other PR. The k8s version we are using is `1.30` for the cluster(byo cni), which means the cilium version should be 1.14 (Atleast). 
The CILIUM_VERSION_TAG is 1.13.13 which is only valid of k8s < 1.29 versions.(Which I think might be outdated) Why don't we update the  `CILIUM_VERSION_TAG` to be 1.14.10 rather than maintaining two cilium versions ?

santhoshmprabhu: @vipul-21 Are you suggesting we use 1.14.10 for all cilium tests including this one? That should work, but I am not sure if we're quite ready to stop all testing on 1.13.13. The purpose of CILIUM_NODE_SUBNET_VERSION is to have a different version just for nodesubnet.

.pipelines/singletenancy/cilium-nodesubnet/cilium-nodesubnet-e2e-step-template.yaml
```diff
@@ -58,7 +58,7 @@ steps:
         kubectl cluster-info
         kubectl get po -owide -A
         echo "install Cilium ${CILIUM_VERSION_TAG}"
-        export DIR=${CILIUM_VERSION_TAG%.*}
+        export DIR=$(echo ${CILIUM_VERSION_TAG#v} | cut -d. -f1,2)
```
vipul-21, line 61: Curious to know the behavior if the version does not have `v` in the version tag ? 
Context: Earlier OSS version did not have a `v` in the version, but our private build image have `v`. In the future, if we want to get rid of private build images so we might not have `v` in the tag.(not sure though)

vipul-21 (approved):

.pipelines/singletenancy/cilium-nodesubnet/cilium-nodesubnet-e2e-step-template.yaml
santhoshmprabhu: The first part of that operation is "remove `v` if present". So we are fine in both cases. In fact, when we merged #3089, we kept the pipeline variables unchanged (so no `v`). Then separately updated the variables.

jpayne3506 (approved):

PR #2454, opened 2023-12-19
chore: move cns from servercore to nanoserver

rbtr:
**Reason for Change**:

`servercore` has idle overhead of ~230 MB of memory and a base image size ~2 GB, vs `nanoserver`'s idle overhead of ~90 MB of memory and base image size around 250 MB. Given CNS image pull-time impacts node readiness, VHD space constraints, and recent issues and heightened awareness of memory usage on Windows, migrate to the lighter base image 🙂 

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] relevant PR labels added

**Notes**:

PR #3046, opened 2024-10-03
upgrade to go v1.22.7

aggarwal0009:
**Reason for Change**:
Upgrade to go v1.22.7 for CVE fixes

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] relevant PR labels added

**Notes**:

PR #3084, opened 2024-10-23
Revert "backport: chore: move cns from servercore to nanoserver (#2454)"

thatmattlong:
Reverts Azure/azure-container-networking#2935

CNS Daemonset on AKS does not have a volume mount for powershell, seemingly due to to https://github.com/Azure/AKS/issues/3885, so we need to keep using servercore on v1.4 branch until we add a volume mount to host powershell

### DIFF
--- a/cns/windows.Dockerfile
+++ b/cns/windows.Dockerfile
@@ -7,16 +7,13 @@ WORKDIR /usr/local/src
 COPY . .
 RUN GOOS=windows CGO_ENABLED=0 go build -a -o /usr/local/bin/azure-cns.exe -ldflags "-X main.version="$VERSION" -X "$CNS_AI_PATH"="$CNS_AI_ID"" -gcflags="-dwarflocationlists=true" cns/service/*.go
 
-# skopeo inspect --override-os windows docker://mcr.microsoft.com/windows/nanoserver:ltsc2019 --format "{{.Name}}@{{.Digest}}"
-FROM mcr.microsoft.com/windows/nanoserver@sha256:7f6649348a11655e3576463fd6d55c29248f97405f8e643cab2409009339f520 AS ltsc2019
+# intermediate for win-ltsc2019
+FROM mcr.microsoft.com/windows/servercore@sha256:6fdf140282a2f809dae9b13fe441635867f0a27c33a438771673b8da8f3348a4 as ltsc2019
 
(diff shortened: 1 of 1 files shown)


nairashu (approved):

nairashu (approved):

build/tools/go.mod
```diff
@@ -9,9 +9,9 @@ require (
 	github.com/golangci/golangci-lint v1.52.2
 	github.com/jstemmer/go-junit-report v1.0.0
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.3.0
-	google.golang.org/protobuf v1.30.0
+	google.golang.org/protobuf v1.34.2
 	mvdan.cc/gofumpt v0.5.0
-	sigs.k8s.io/controller-tools v0.12.0
+	sigs.k8s.io/controller-tools v0.16.5
```
rbtr, line 14: I don't think we can do this, [controller-runtime versions target specific k8s versions](https://github.com/kubernetes-sigs/controller-tools?tab=readme-ov-file#compatibility) and we need to freeze for 1.27 compatibility

thatmattlong: The problem is that controller-tools at 12.0 is not usable

thatmattlong: it panics: https://github.com/Azure/azure-container-networking/actions/runs/11486652516/job/32250025824

jpayne3506: Per `controller-tools` 
> Every minor version of controller-tools has been tested with a specific minor version of client-go. A controller-tools minor version may be compatible with other client-go minor versions, but this is by chance and neither supported nor tested. In general, we create one minor version of controller-tools for each minor version of client-go and other k8s.io/* dependencies.

This means that by chance it was working for 1.21 when they only meant for it to work for 1.20. We either rollback to 1.21 for our build tools, make a controller tools specific go.mod that will use go 1.20 or 1.21, or bump the controller-tools dep to v0.14.0+.

All have drawbacks...

jpayne3506: For more context, this only [broke for us once we updated](https://github.com/Azure/azure-container-networking/actions/runs/11153809821/job/31032529322?pr=3046) to 1.22.7 in #3046.

rbtr (approved):

rbtr:
I will bypass CRD Gen, I don't think we should change it in this revert PR

jpayne3506 (approved):

PR #3085, opened 2024-10-24
feat: adding stateless CNI support for ACI

behzad-mir:
feat: adding stateless CNI support for ACI>
New entry in CNS statefile will be added for a container upon updateEndpoint request if the entry does not already exist. This covers a special case for ACI.

In ACI the CNS path to allocate IP address and send goalstate is different than CNS IPAM for Cilium and azure-cni for AKS. As a result, the endpoint entry for statefile is not getting create upon IP allocation and this fix will create the whole entry when CNI send updateEndpoint API request to CNS.

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/ipam_test.go
+++ b/cns/restserver/ipam_test.go
@@ -2154,3 +2154,128 @@ func createAndSaveMockNCRequest(t *testing.T, svc *HTTPRestService, ncID string,
 	require.Equal(t, types.Success, returnCode)
 	require.Empty(t, returnMessage)
 }
+
+// Validate Statefile in Stateless CNI scenarios
+func TestStatelessCNIStateFile(t *testing.T) {
+	svc := getTestService(cns.KubernetesCRD)
+	svc.EndpointStateStore = store.NewMockStore("")
+	// test Case 1 - AKS SIngleTenancy
+	endpointInfo1ContainerID := "0a4917617e15d24dc495e407d8eb5c88e4406e58fa209e4eb75a2c2fb7045eea"
+	endpointInfo1 := &EndpointInfo{IfnameToIPMap: make(map[string]*IPInfo)}
+	endpointInfo1.IfnameToIPMap["eth0"] = &IPInfo{IPv4: []net.IPNet{{IP: net.IPv4(10, 0, 0, 1), Mask: net.IPv4Mask(255, 255, 255, 0)}}}
+	req1 := make(map[string]*IPInfo)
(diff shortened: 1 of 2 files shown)


QxBytes (commented):
could you add the details of the special case in the pr description (like the timeline of events and why there is no entry in the cns state by the time we call update endpoint helper)-- I thought the endpoint state is usually populated in updateEndpointState in cns, unless we don't request ips? Is InfraContainerID (in update endpoint state) typically equal to endpointID (in update endpoint helper)?

behzad-mir:
> could you add the details of the special case in the pr description (like the timeline of events and why there is no entry in the cns state by the time we call update endpoint helper)-- I thought the endpoint state is usually populated in updateEndpointState in cns, unless we don't request ips? Is InfraContainerID (in update endpoint state) typically equal to endpointID (in update endpoint helper)?

DOne

QxBytes (dismissed):

tamilmani1989 (commented):
can we write UT for this case

behzad-mir:
> can we write UT for this case

Done

cns/restserver/ipam_test.go
```diff
@@ -2154,3 +2154,128 @@ func createAndSaveMockNCRequest(t *testing.T, svc *HTTPRestService, ncID string,
...
+	tests := []struct {
+		name       string
+		endpointID string
+		req        map[string]*IPInfo
+		store      store.KeyValueStore
+		want       *EndpointInfo
+		wantErr    bool
+	}{
+		{
+			name:       "good",
```
tamilmani1989, line 2193: name should briefly describe test case.. couldn't interpret from what good is

cns/restserver/ipam_test.go
```diff
@@ -2154,3 +2154,128 @@ func createAndSaveMockNCRequest(t *testing.T, svc *HTTPRestService, ncID string,
...
+						IPv4:          []net.IPNet{{IP: net.IPv4(10, 0, 0, 1), Mask: net.IPv4Mask(255, 255, 255, 0)}},
+						HnsEndpointID: "5c15cccc-830a-4dff-81f3-4b1e55cb7dcb",
+						NICType:       cns.InfraNIC,
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name:       "good with ACI endpoint",
```
tamilmani1989, line 2209: same as above

tamilmani1989, line 2209: i hope goal is to test updateendpoint api given cns state doesnt have any entry for that container id

behzad-mir: THe ACI endpoint is absent in the statefile when the update is called

behzad-mir: done

cns/restserver/ipam_test.go
behzad-mir: updated the names

tamilmani1989 (approved):

PR #3086, opened 2024-10-24
[NPM Lite] Querying L1VH + Non-L1VH Endpoints

rejain456:
**Reason for Change**:

Currently, for non-l1vh window nodes, we query for hns endpoints with state 3 (attachedSharing). However, for L1VH nodes, the hns endpoints have state 2 (attached).  NPM Lite will use both non-l1vh window nodes as well as L1VH nodes.

**Issue Fixed**:

This pr aims to query for both states, state 2 for npm and npm lite and state 3 if npm lite is enabled.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/npm/pkg/dataplane/dataplane_windows.go
+++ b/npm/pkg/dataplane/dataplane_windows.go
@@ -50,19 +50,39 @@ func (dp *DataPlane) initializeDataPlane() error {
 		},
 		Flags: hcn.HostComputeQueryFlagsNone,
 	}
+	// Initialize Endpoint query used to filter healthy endpoints (vNIC) of Windows pods on L1VH Node
+	dp.endpointQueryL1VH.query = hcn.HostComputeQuery{
+		SchemaVersion: hcn.SchemaVersion{
+			Major: hcnSchemaMajorVersion,
+			Minor: hcnSchemaMinorVersion,
+		},
+		Flags: hcn.HostComputeQueryFlagsNone,
+	}
+
 	// Filter out any endpoints that are not in "AttachedShared" State. All running Windows pods with networking must be in this state.
 	filterMap := map[string]uint16{"State": hcnEndpointStateAttachedSharing}
-	filter, err := json.Marshal(filterMap)
-	if err != nil {
-		return npmerrors.SimpleErrorWrapper("failed to marshal endpoint filter map", err)
-	}
+	filterMapL1VH := map[string]uint16{"State": hcnEndpointStateAttached}
+
+	filter, err := marshalFilterMap(filterMap)
+
+	filterL1VH, err := marshalFilterMap(filterMapL1VH)
+
 	dp.endpointQuery.query.Filter = string(filter)
+	dp.endpointQueryL1VH.query.Filter = string(filterL1VH)
 
 	// reset endpoint cache so that netpol references are removed for all endpoints while refreshing pod endpoints
 	// no need to lock endpointCache at boot up
 	dp.endpointCache.cache = make(map[string]*npmEndpoint)
 
 	return nil
+}
+
+func marshalFilterMap(filtermap map[string]uint16) ([]byte, error) {
+	filter, err := json.Marshal(filtermap)
+	if err != nil {
+		return nil, npmerrors.SimpleErrorWrapper("failed to marshal endpoint filter map", err)
+	}
+	return filter, nil
 }
 
 func (dp *DataPlane) getNetworkInfo() error {
@@ -330,13 +350,15 @@ func (dp *DataPlane) getEndpointsToApplyPolicies(netPols []*policies.NPMNetworkP
 func (dp *DataPlane) getLocalPodEndpoints() ([]*hcn.HostComputeEndpoint, error) {
 	klog.Info("getting local endpoints")
 	timer := metrics.StartNewTimer()
-	endpoints, err := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQuery.query)
+	endpointsAttachedSharing, err := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQuery.query)
+	endpointsAttached, err := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQueryL1VH.query)
 	metrics.RecordListEndpointsLatency(timer)
 	if err != nil {
 		metrics.IncListEndpointsFailures()
 		return nil, npmerrors.SimpleErrorWrapper("failed to get local pod endpoints", err)
 	}
-
+	endpoints := append(endpointsAttachedSharing, endpointsAttached...)
+	klog.Infof("there are %+v endpoints in endpointsAttachedSharing and %+v endpoints in Attached", len(endpointsAttachedSharing), len(endpointsAttached))
 	epPointers := make([]*hcn.HostComputeEndpoint, 0, len(endpoints))
 	for k := range endpoints {
 		epPointers = append(epPointers, &endpoints[k])
--- a/npm/pkg/dataplane/dataplane_windows.go
+++ b/npm/pkg/dataplane/dataplane_windows.go
@@ -351,13 +351,12 @@ func (dp *DataPlane) getLocalPodEndpoints() ([]*hcn.HostComputeEndpoint, error) 
 	klog.Info("getting local endpoints")
 	timer := metrics.StartNewTimer()
 	endpoints, err := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQuery.query)
-	klog.Infof("There are %+v endpoints in AttachedSharing state", len(endpoints))
 	metrics.RecordListEndpointsLatency(timer)
 	if err != nil {
 		metrics.IncListEndpointsFailures()
 		return nil, npmerrors.SimpleErrorWrapper("failed to get local pod endpoints", err)
 	}
-	klog.Infof("NPM liste is enabled: %+v", dp.EnableNPMLite)
+
 	if dp.EnableNPMLite {
 		timer = metrics.StartNewTimer()
 		endpointsAttached, errL1vh := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQueryL1VH.query)
@@ -365,7 +364,6 @@ func (dp *DataPlane) getLocalPodEndpoints() ([]*hcn.HostComputeEndpoint, error) 
 			metrics.IncListEndpointsFailures()
 			return nil, npmerrors.SimpleErrorWrapper("failed to get local pod endpoints in L1VH", err)
 		}
-		klog.Infof("There are %+v endpoints in Attached state on l1vh", len(endpointsAttached))
 		endpoints = append(endpoints, endpointsAttached...)
 	}
 	epPointers := make([]*hcn.HostComputeEndpoint, 0, len(endpoints))
--- a/npm/pkg/dataplane/dataplane_windows.go
+++ b/npm/pkg/dataplane/dataplane_windows.go
@@ -350,15 +350,23 @@ func (dp *DataPlane) getEndpointsToApplyPolicies(netPols []*policies.NPMNetworkP
 func (dp *DataPlane) getLocalPodEndpoints() ([]*hcn.HostComputeEndpoint, error) {
 	klog.Info("getting local endpoints")
 	timer := metrics.StartNewTimer()
-	endpointsAttachedSharing, err := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQuery.query)
-	endpointsAttached, err := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQueryL1VH.query)
+	endpoints, err := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQuery.query)
+	klog.Infof("There are %+v endpoints in AttachedSharing state", len(endpoints))
 	metrics.RecordListEndpointsLatency(timer)
 	if err != nil {
 		metrics.IncListEndpointsFailures()
 		return nil, npmerrors.SimpleErrorWrapper("failed to get local pod endpoints", err)
 	}
-	endpoints := append(endpointsAttachedSharing, endpointsAttached...)
-	klog.Infof("there are %+v endpoints in endpointsAttachedSharing and %+v endpoints in Attached", len(endpointsAttachedSharing), len(endpointsAttached))
+	if dp.EnableNPMLite {
+		timer = metrics.StartNewTimer()
+		endpointsAttached, errL1vh := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQueryL1VH.query)
+		if errL1vh != nil {
+			metrics.IncListEndpointsFailures()
+			return nil, npmerrors.SimpleErrorWrapper("failed to get local pod endpoints in L1VH", err)
+		}
+		klog.Infof("There are %+v endpoints in Attached state on l1vh", len(endpointsAttached))
+		endpoints = append(endpoints, endpointsAttached...)
+	}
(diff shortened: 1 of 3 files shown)


npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -65,6 +77,14 @@ func (dp *DataPlane) initializeDataPlane() error {
 	return nil
 }
 
+func marshalFilterMap(filtermap map[string]uint16) ([]byte, error) {
```
timraymond, line 80: There isn't really anything interesting happening in this function (since it's just JSON marshaling), so it's probably best to just inline this.

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -65,6 +77,14 @@ func (dp *DataPlane) initializeDataPlane() error {
...
+	filter, err := json.Marshal(filtermap)
+	if err != nil {
+		return nil, npmerrors.SimpleErrorWrapper("failed to marshal endpoint filter map", err)
```
timraymond, line 83: `npmerrors.SimpleError` doesn't look like it should exist since it's just an `error`--it doesn't add anything. Use the following from `github.com/pkg/errors` (you will probably need to add that import--the suggestion is not enough)

```suggestion
		return nil, errors.Wrap(err, "failed to marshal endpoint filter map")
```

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -337,6 +357,15 @@ func (dp *DataPlane) getLocalPodEndpoints() ([]*hcn.HostComputeEndpoint, error)
 		return nil, npmerrors.SimpleErrorWrapper("failed to get local pod endpoints", err)
 	}
 
+	if dp.EnableNPMLite {
+		timer = metrics.StartNewTimer()
+		endpointsAttached, errL1vh := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQueryL1VH.query)
+		if errL1vh != nil {
+			metrics.IncListEndpointsFailures()
+			return nil, npmerrors.SimpleErrorWrapper("failed to get local pod endpoints in L1VH", err)
```
timraymond, line 365: The returned error is probably not the one you want (`err` != `errL1vh`)... this is why it's always best to just spell that `err` if you can. If the linter complains about shadowing the variable, it's wrong--feel free to drop a `//nolint` for it.

```suggestion
		endpointsAttached, err := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQueryL1VH.query)
		if err != nil {
			metrics.IncListEndpointsFailures()
			return nil, npmerrors.SimpleErrorWrapper("failed to get local pod endpoints in L1VH", err)
```

npm/pkg/dataplane/dataplane_windows.go
rejain456: made it inline again, @timraymond pls let me know if that looks better

npm/pkg/dataplane/dataplane_windows.go
rejain456: changed it to errors.Wrap, thank you for the suggestion

npm/pkg/dataplane/dataplane_windows.go
rejain456: @timraymond , here I was going for errl1vh as that error gets outputted if the call to list endpoints in lvh node fails. Is there a reason why it should be err and not errl1vh?

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -334,9 +352,19 @@ func (dp *DataPlane) getLocalPodEndpoints() ([]*hcn.HostComputeEndpoint, error)
...
 
+	if dp.EnableNPMLite {
+		timer = metrics.StartNewTimer()
+		endpointsAttached, errL1vh := dp.ioShim.Hns.ListEndpointsQuery(dp.endpointQueryL1VH.query)
+		metrics.RecordListEndpointsLatency(timer)
+		if errL1vh != nil {
+			metrics.IncListEndpointsFailures()
+			return nil, errors.Wrap(errL1vh, "failed to get local pod endpoints in L1VH")
+		}
+		endpoints = append(endpoints, endpointsAttached...)
```
huntergregory, line 366: Is there any risk if there are two of the same endpoint in this slice (two objects, same endpoint ID, different "state")? Not sure this would ever happen, but

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -50,13 +50,31 @@ func (dp *DataPlane) initializeDataPlane() error {
...
+	// Initialize Endpoint query used to filter healthy endpoints (vNIC) of Windows pods on L1VH Node
+	dp.endpointQueryL1VH.query = hcn.HostComputeQuery{
+		SchemaVersion: hcn.SchemaVersion{
+			Major: hcnSchemaMajorVersion,
+			Minor: hcnSchemaMinorVersion,
+		},
+		Flags: hcn.HostComputeQueryFlagsNone,
+	}
+
 	// Filter out any endpoints that are not in "AttachedShared" State. All running Windows pods with networking must be in this state.
```
huntergregory, line 62: nit: Could we add a comment for L1VH too?

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -50,13 +50,31 @@ func (dp *DataPlane) initializeDataPlane() error {
...
+
 	filter, err := json.Marshal(filterMap)
 	if err != nil {
-		return npmerrors.SimpleErrorWrapper("failed to marshal endpoint filter map", err)
+		return errors.Wrap(err, "failed to marshal endpoint filter map")
+	}
+
+	filterL1VH, errL1VH := json.Marshal(filterMapL1VH)
+	if errL1VH != nil {
+		return errors.Wrap(errL1VH, "failed to marshal endpoint filter map")
```
huntergregory: nit: Can we change the error message? Might make debugging easier one day

npm/pkg/dataplane/dataplane_windows.go
timraymond: @rejain456 in the existing logic, you're testing against `errL1vh`, but returning a wrapper around `err`... which will probably just output `failed to get local pod endpoints in L1VH: <nil>`, swallowing the context you want from `errL1vh`. Continually shadowing the definition of `err` implicitly scopes the lifetime of that value to the previous error handling block. That's why I was just saying, unless you _really_ need to name it something else, always naming it `err` is good practice to make sure the right thing gets output.

rejain456: That makes sense, thank you for the explanation. I changed errl1vh to err in both occurrences.

npm/pkg/dataplane/dataplane_windows.go
rejain456: yes added more detailed error messaged. Please let me know if that looks better

npm/pkg/dataplane/dataplane_windows.go
rejain456: yes added, thanks for the suggestion!

npm/pkg/dataplane/dataplane_windows.go
rejain456: Added a function which takes out common endpoints and returns a unique list of endpoints

npm/pkg/dataplane/dataplane_windows_test.go
```diff
@@ -86,6 +88,72 @@ func TestMultiJobApplyInBackground(t *testing.T) {
...
+			endpointsAttached: []hcn.HostComputeEndpoint{},
+			expected:          []hcn.HostComputeEndpoint{{Id: "456901"}},
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+
+		t.Run(tt.name, func(t *testing.T) {
+			result := removeCommonEndpoints(&tt.endpoints, &tt.endpointsAttached)
+			// Use reflect.DeepEqual as a backup if require.Equal doesn't work as expected
```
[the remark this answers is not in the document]
timraymond: Also, you should use `github.com/google/go-cmp/cmp` in tests instead of `reflect.DeepEqual`.

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -347,31 +346,67 @@ func (dp *DataPlane) getEndpointsToApplyPolicies(netPols []*policies.NPMNetworkP
...
+	endpoints = removeCommonEndpoints(&endpoints, &endpointsAttached)
+
 	epPointers := make([]*hcn.HostComputeEndpoint, 0, len(endpoints))
 	for k := range endpoints {
 		epPointers = append(epPointers, &endpoints[k])
 	}
 	return epPointers, nil
 }
 
+func removeCommonEndpoints(endpoints, endpointsAttached *[]hcn.HostComputeEndpoint) []hcn.HostComputeEndpoint {
```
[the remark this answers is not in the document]
timraymond: The exact operation it appears to do is the [symmetric difference](https://en.wikipedia.org/wiki/Symmetric_difference) of the two sets of endpoints.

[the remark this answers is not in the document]
rejain456: ah got it, thanks for the find. Removed the *

npm/pkg/dataplane/dataplane_windows_test.go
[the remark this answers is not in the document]
rejain456: Replaced deepEqual with cmp.Equal... thank you for the suggestion

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -347,31 +346,67 @@ func (dp *DataPlane) getEndpointsToApplyPolicies(netPols []*policies.NPMNetworkP
...
+	// Choose smaller and larger lists based on length for efficiency
```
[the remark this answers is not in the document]
rejain456: makes sense, removed the minor optimization as it won't make a noticeable difference

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -347,31 +346,67 @@ func (dp *DataPlane) getEndpointsToApplyPolicies(netPols []*policies.NPMNetworkP
...
+	}
+
+	// Store IDs of smaller list in a map for quick lookup
+	idMap := make(map[string]struct{}, len(*smallerEndpointsList))
+	for i := 0; i < len(*smallerEndpointsList); i++ {
+		ep := &(*smallerEndpointsList)[i]
+		idMap[ep.Id] = struct{}{}
+	}
+
+	// Append endpoints from larger list and remove common IDs from map
```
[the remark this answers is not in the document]
rejain456: Thanks for the tip. This is a more straightforward approach, changed the code to depict this

npm/pkg/dataplane/dataplane_windows.go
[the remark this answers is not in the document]
rejain456: Thanks Tim for the exact name. @huntergregory, changed the name to GetUniqueEndpoints. I chose this name as it may be easier to comprehend what the function is about. Please let me know if that works.

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -375,36 +375,22 @@ func (dp *DataPlane) getLocalPodEndpoints() ([]*hcn.HostComputeEndpoint, error)
...
-
-	// Store IDs of smaller list in a map for quick lookup
-	idMap := make(map[string]struct{}, len(*smallerEndpointsList))
-	for i := 0; i < len(*smallerEndpointsList); i++ {
-		ep := &(*smallerEndpointsList)[i]
+func GetUniqueEndpoints(endpoints, endpointsAttached []hcn.HostComputeEndpoint) []hcn.HostComputeEndpoint {
+	// Store IDs of endpoints list in a map for quick lookup
+	idMap := make(map[string]struct{}, len(endpoints))
+	for i := 0; i < len(endpoints); i++ {
+		ep := &(endpoints)[i]
```
huntergregory, line 382: what does the `&` do here?

npm/pkg/dataplane/dataplane_windows.go
```diff
@@ -375,36 +375,22 @@ func (dp *DataPlane) getLocalPodEndpoints() ([]*hcn.HostComputeEndpoint, error)
...
-	}
-
-	// Store IDs of smaller list in a map for quick lookup
-	idMap := make(map[string]struct{}, len(*smallerEndpointsList))
-	for i := 0; i < len(*smallerEndpointsList); i++ {
-		ep := &(*smallerEndpointsList)[i]
+func GetUniqueEndpoints(endpoints, endpointsAttached []hcn.HostComputeEndpoint) []hcn.HostComputeEndpoint {
+	// Store IDs of endpoints list in a map for quick lookup
+	idMap := make(map[string]struct{}, len(endpoints))
+	for i := 0; i < len(endpoints); i++ {
```
huntergregory, line 381: not required but golang usually loops through elements with `range` if index isn't needed
```suggestion
	for _, ep := range endpoints {
```

npm/pkg/dataplane/dataplane_windows_test.go
```diff
@@ -99,19 +99,19 @@ func TestRemoveCommonEndpoints(t *testing.T) {
 			name:              "1 value same",
 			endpoints:         []hcn.HostComputeEndpoint{{Id: "456901"}, {Id: "123456"}, {Id: "560971"}},
 			endpointsAttached: []hcn.HostComputeEndpoint{{Id: "567890"}, {Id: "123456"}, {Id: "789012"}},
-			expected:          []hcn.HostComputeEndpoint{{Id: "567890"}, {Id: "123456"}, {Id: "789012"}, {Id: "456901"}, {Id: "560971"}},
+			expected:          []hcn.HostComputeEndpoint{{Id: "456901"}, {Id: "123456"}, {Id: "560971"}, {Id: "567890"}, {Id: "789012"}},
```
huntergregory, line 102: just an FYI in case you need in the future. Sometimes it's easiest to sort both the output and the expected output for unit tests. That's a bit simpler for slices of strings versus structs but

npm/pkg/dataplane/dataplane_windows.go
rejain456: changed to use range, thanks

npm/pkg/dataplane/dataplane_windows.go
rejain456: removed the & -> had added it originally but realized no elements in endpoints directly need to be modified during iteration so don't need the &. Also removed this line completely in replacement of adding for range loop

npm/pkg/dataplane/dataplane_windows_test.go
rejain456: got it makes sense, thank you this is good to know for future use

npm/pkg/dataplane/dataplane_windows.go
rejain456: @huntergregory, when I updated it to use the range, I would see the following error:
image

I believe it is because the range loop always copies each struct automatically whereas in the for indexed loop, the struct does not get copies automatically. It only gets copies if code is written for it within the for block. 
Having each struct automatically copies results in the error:  rangeValCopy: each iteration copies 392 bytes (consider pointers or indexing) (gocritic)

huntergregory: sense

huntergregory (approved):

npm/pkg/dataplane/dataplane_windows.go
timraymond: That's actually a nice catch from the linter

PR #3092, opened 2024-10-29
Adding availability zone to nnc spec

ramiro-gamarra:
**Reason for Change**:

In order to communicate the availability zone for ncs from the virtual machine where they are placed, we first need to expand the nnc definition to include the az in its status. Population will be handled in a subsequent PR.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
+++ b/crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
@@ -48,6 +48,7 @@ type NodeNetworkConfigSpec struct {
 	RequestedIPCount int64    `json:"requestedIPCount"`
 	IPsNotInUse      []string `json:"ipsNotInUse,omitempty"`
 	// AvailabilityZone contains the Azure availability zone for the virtual machine where network containers are placed.
+	// +kubebuilder:validation:Optional
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
```diff
@@ -46,6 +47,8 @@ type NodeNetworkConfigSpec struct {
 	// +kubebuilder:validation:Optional
 	RequestedIPCount int64    `json:"requestedIPCount"`
 	IPsNotInUse      []string `json:"ipsNotInUse,omitempty"`
+	// AvailabilityZone contains the Azure availability zone for the virtual machine where network containers are placed.
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
```
thatmattlong, line 51: let's add `// +kubebuilder:validation:Optional` since we expect that some NNCs might get created/updated without this field for some time

thatmattlong (changes requested):
let's make the az property optional

crd/nodenetworkconfig/api/v1alpha/nodenetworkconfig.go
ramiro-gamarra: added

thatmattlong (approved):

PR #3095, opened 2024-10-30
fix: pass ai metadata to telemetry during cni image build

QxBytes:
Updates the cni dockerfile to take in arguments for the ai metadata, similar to the cns dockerfile.

**Reason for Change**:

The ai instrumentation key previously was not passed when building the azure vnet telemetry binary when making a *cni image*. The instrumentation key was passed when making the binary with the make file command, however.

Also fixed an error message not logging which hid the above error.

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Tested on podsubnet cluster. No logs were found to be uploaded, and noticed telemetry logs showed an ai handle creation error. However, after applying the cni image with the fix using the cni installer, we started to receive logs. The telemetry logs also no longer produced the error message.

### DIFF
--- a/cni/Dockerfile
+++ b/cni/Dockerfile
@@ -12,10 +12,12 @@ FROM --platform=linux/${ARCH} mcr.microsoft.com/cbl-mariner/base/core@sha256:a49
 FROM go AS azure-vnet
 ARG OS
 ARG VERSION
+ARG CNI_AI_PATH
+ARG CNI_AI_ID
 WORKDIR /azure-container-networking
 COPY . .
 RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet -trimpath -ldflags "-X main.version="$VERSION"" -gcflags="-dwarflocationlists=true" cni/network/plugin/main.go
-RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet-telemetry -trimpath -ldflags "-X main.version="$VERSION"" -gcflags="-dwarflocationlists=true" cni/telemetry/service/telemetrymain.go
+RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet-telemetry -trimpath -ldflags "-X main.version="$VERSION" -X "$CNI_AI_PATH"="$CNI_AI_ID"" -gcflags="-dwarflocationlists=true" cni/telemetry/service/telemetrymain.go
 RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet-ipam -trimpath -ldflags "-X main.version="$VERSION"" -gcflags="-dwarflocationlists=true" cni/ipam/plugin/main.go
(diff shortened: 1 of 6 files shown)


cni/telemetry/service/telemetrymain.go
```diff
@@ -167,8 +167,8 @@ func main() {
 		GetEnvRetryWaitTimeInSecs:    config.GetEnvRetryWaitTimeInSecs,
 	}
 
-	if tb.CreateAITelemetryHandle(aiConfig, config.DisableAll, config.DisableTrace, config.DisableMetric) != nil {
-		logger.Error("AI Handle creation error", zap.Error(err))
+	if aiCreationErr := tb.CreateAITelemetryHandle(aiConfig, config.DisableAll, config.DisableTrace, config.DisableMetric); aiCreationErr != nil {
+		logger.Error("AI Handle creation error:", zap.Error(aiCreationErr))
```
timraymond, line 171: `err` is fine for this, and is least surprising. Silence the linter if it complains about shadowing.

```suggestion
	if err := tb.CreateAITelemetryHandle(aiConfig, config.DisableAll, config.DisableTrace, config.DisableMetric); err != nil {
		logger.Error("AI Handle creation error:", zap.Error(err))
```

cni/Dockerfile
```diff
@@ -12,10 +12,12 @@ FROM --platform=linux/${ARCH} mcr.microsoft.com/cbl-mariner/base/core@sha256:a49
 FROM go AS azure-vnet
 ARG OS
 ARG VERSION
+ARG CNI_AI_PATH
```
timraymond, line 15: If we have to modify source to change this (it's a package variable, I assume)... then we can probably just hardcode it. Not sure that a build arg buys us anything.

QxBytes: this is mainly mirroring what is done in cns, but I can change it to just hardcode

QxBytes: actually it seems like the aiMetadata is variable is used for both cns and cni. It seems odd that we would hardcode the aiMetadata to CNI's id when CNS also uses that same path, unless we want to change that too.

timraymond: I didn't mean the value, I meant the key `aiMetadata` (which I presume is the value of `CNI_AI_PATH`) that is getting set here. If we decide to use something else, we'd have to make a code change to rename that package variable anyway, no?

QxBytes: Not sure if my later commits addressed your concerns-- I've set the aiMetadata equal to the cni ai id in the source code itself.

QxBytes: seeing logs being logged after addressing your suggestions after a pipeline run-- could you take another look?

rbtr (dismissed):
oops, thanks for fixing

cni/Dockerfile
rbtr: i think the reason not to hardcode was to just slightly obfuscate it from the public - not that it's a _secret_ but we're not yelling it everywhere

cns/service/main.go
```diff
@@ -672,7 +672,7 @@ func main() {
 	}
 
 	if telemetryDaemonEnabled {
-		logger.Printf("CNI Telemetry is enabled")
+		logger.Printf("CNI Telemtry is enabled")
```
timraymond, line 675: Looks like this may have gotten accidentally reverted.

rbtr (approved):

timraymond (approved):

PR #3103, opened 2024-11-04
expose MinTLSVersion config for TLS handshake

ZetaoZhuang:
**Reason for Change**:

A configuration change is required in CNS to be able to specify the optional config of minimum TLS version support desired for TLS handshake

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/service.go
+++ b/cns/service.go
@@ -28,7 +28,7 @@ const (
 	genericData          = "com.microsoft.azure.network.generic"
 )
 
-var errBootConfig = errors.New("configuration error")
+var errTLSConfig = errors.New("unsupported TLS version name from config")
 
 // Service defines Container Networking Service.
 type Service struct {
@@ -181,9 +181,9 @@ func getTLSConfigFromFile(tlsSettings localtls.TlsSettings) (*tls.Config, error)
 		PrivateKey:  privateKey,
 		Leaf:        leafCertificate,
 	}
-	minTLSVersionNumber, err := TLSVersionNumber(tlsSettings.MinTLSVersion)
+	minTLSVersionNumber, err := parseTLSVersionName(tlsSettings.MinTLSVersion)
 	if err != nil {
-		return nil, errors.Wrap(err, "MinTLSVersion is not valid")
+		return nil, errors.Wrap(err, "parsing MinTLSVersion from config")
 	}
 
 	tlsConfig := &tls.Config{
@@ -232,9 +232,9 @@ func getTLSConfigFromKeyVault(tlsSettings localtls.TlsSettings, errChan chan<- e
 		errChan <- cr.Refresh(ctx, tlsSettings.KeyVaultCertificateRefreshInterval)
 	}()
 
-	minTLSVersionNumber, err := TLSVersionNumber(tlsSettings.MinTLSVersion)
+	minTLSVersionNumber, err := parseTLSVersionName(tlsSettings.MinTLSVersion)
 	if err != nil {
-		return nil, errors.Wrap(err, "MinTLSVersion is not valid")
+		return nil, errors.Wrap(err, "parsing MinTLSVersion from config")
 	}
 
 	tlsConfig := tls.Config{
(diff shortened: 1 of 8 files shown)


cns/service.go
```diff
@@ -316,3 +327,20 @@ func (service *Service) SendErrorResponse(w http.ResponseWriter, errMsg error) {
 	err := acn.Encode(w, &resp)
 	logger.Errorf("[%s] %+v %s.", service.Name, &resp, err.Error())
 }
+
+// TLSVersionNumber returns the version number for the provided TLS version name
+// (e.g. 0x0301)
+func TLSVersionNumber(versionName string) (uint16, error) {
```
timraymond, line 333: I am surprised the standard library doesn't have a parse function for this. I would change this to `parseTLSVersionNumber` since that's the convention you would find there.

cns/service.go
```diff
@@ -316,3 +327,20 @@ func (service *Service) SendErrorResponse(w http.ResponseWriter, errMsg error) {
...
+	case "TLS 1.0":
+		return tls.VersionTLS10, nil
+	case "TLS 1.1":
+		return tls.VersionTLS11, nil
+	case "TLS 1.2":
+		return tls.VersionTLS12, nil
+	case "TLS 1.3":
+		return tls.VersionTLS13, nil
+	default:
+		return 0, errors.Wrap(errBootConfig, "unsupported TLS version name")
```
timraymond, line 344: "configuration error" isn't adding a whole lot of information here. I think you can just have `unsupported TLS version name` as the constantized error, then include what you tried to parse (the `versionName` arg) in the `Wrapf`.

cns/service.go
```diff
@@ -226,8 +232,13 @@ func getTLSConfigFromKeyVault(tlsSettings localtls.TlsSettings, errChan chan<- e
 		errChan <- cr.Refresh(ctx, tlsSettings.KeyVaultCertificateRefreshInterval)
 	}()
 
+	minTLSVersionNumber, err := TLSVersionNumber(tlsSettings.MinTLSVersion)
+	if err != nil {
+		return nil, errors.Wrap(err, "MinTLSVersion is not valid")
```
timraymond, line 237: Usually, for error wrapping, I put what I was trying to do when the error emerged. Here it would be:

```suggestion
		return nil, errors.Wrap(err, "parsing MinTLSVersion from config")
```

Doing this gives you a progressive chain of explanation.

cns/service.go
```diff
@@ -316,3 +327,20 @@ func (service *Service) SendErrorResponse(w http.ResponseWriter, errMsg error) {
...
 }
+
+// TLSVersionNumber returns the version number for the provided TLS version name
+// (e.g. 0x0301)
+func TLSVersionNumber(versionName string) (uint16, error) {
+	switch versionName {
+	case "TLS 1.0":
+		return tls.VersionTLS10, nil
+	case "TLS 1.1":
+		return tls.VersionTLS11, nil
```
jackieluc, line 338: since we already support a min TLS version of 1.2 as the hardcoded minimum, let's not support any TLS versions lesser than this. this will align with Azure's TLS version guidance as well for their services.

cns/service_test.go
```diff
@@ -322,3 +327,18 @@ func createTestCertificate(t *testing.T) string {
 
 	return testCertFilePath
 }
+
+func TestTLSVersionNumber(t *testing.T) {
```
jackieluc, line 331: add test cases for unsupported: TLS 1.0, TLS 1.1 as per TLS version support comment above

jackieluc (changes requested):

cns/service.go
timraymond: Yeah +1 to that. Easier to prevent a downgrade attack by just flat-out saying "no" to older TLS versions.

jackieluc (approved):

timraymond (approved):

PR #3105, opened 2024-11-05
feat: Windows HPC base image

rbtr:
Windows HPC are special.

They explicitly [do not run in a `chroot`](https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support#container-mounts:~:text=There%20will%20be%20no%20chroot%20equivalent) or similar execution sandbox, and their [filesystem is a merged view](https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support#container-mounts) of the hostfs and their shipped containerfs. This means that we don't need to ship a filesystem: we recently moved from `servercore` to `nanoserver` and observed that the lack of `pwsh.exe` in `nanoserver` is manageable (in containerd 1.7+, automatically, and in 1.6, by tweaking the container `PATH`). We can go further and remove the rest of the OS since, as HPC, we don't need it and can escape to the hostfs for dependencies like powershell.

Further, HPC do not have the [same Windows OS version compatibility requirements](https://github.com/containerd/containerd/issues/7431) (the "os-version") as regular Windows containers - any Windows HPC may run on any Windows Server Host (given the executables are compatible with that Host). 
In some of the testing with WS25 we have seen that this may happen - it was observed that containerd on WS25 inconsistently fetched the WS25 CNI image specified in our multiplatform manifest but was able to run other WS images without any issue.
Containerd >= 1.6 [should have compatibility for Windows HPC without an `os-version` set](https://github.com/containerd/containerd/pull/8101), so we don't need to build separate Windows images for all of our target WS variants and can simply build one without that extra metadata.

### DIFF
--- a/.pipelines/npm/npm-conformance-tests.yaml
+++ b/.pipelines/npm/npm-conformance-tests.yaml
@@ -74,18 +74,16 @@ stages:
             arch: amd64
             name: npm
             os: linux
-          npm_windows2022_amd64:
+          npm_windows_amd64:
             arch: amd64
             name: npm
             os: windows
-            os_version: ltsc2022
       steps:
         - template: ../containers/container-template.yaml
           parameters:
             arch: $(arch)
             name: $(name)
             os: $(os)
-            os_version: $(os_version)
 
   - stage: Create_Cluster_and_Run_Test
     displayName: NPM Conformance
@@ -204,7 +202,7 @@ stages:
 
                 az aks get-credentials -n $(AZURE_CLUSTER) -g $(RESOURCE_GROUP) --file ./kubeconfig
                 ./kubectl --kubeconfig=./kubeconfig apply -f $(Pipeline.Workspace)/s/npm/examples/windows/azure-npm.yaml
-                ./kubectl --kubeconfig=./kubeconfig set image daemonset/azure-npm-win -n kube-system azure-npm=$IMAGE_REGISTRY/azure-npm:windows-amd64-ltsc2022-$(TAG)
+                ./kubectl --kubeconfig=./kubeconfig set image daemonset/azure-npm-win -n kube-system azure-npm=$IMAGE_REGISTRY/azure-npm:windows-amd64-$(TAG)
 
               else
                 echo "Creating Linux Cluster";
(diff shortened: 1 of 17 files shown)


rbtr:
cc @tyler-lloyd and @paulgmiller

jpayne3506:
Checking if the makefile changes impact other E2E besides PR

npm/windows.Dockerfile
```diff
@@ -8,9 +8,7 @@ COPY . .
 RUN GOOS=windows CGO_ENABLED=0 go build -v -o /usr/local/bin/azure-npm.exe -ldflags "-X main.version="$VERSION" -X "$NPM_AI_PATH"="$NPM_AI_ID"" -gcflags="-dwarflocationlists=true" npm/cmd/*.go
 
 # intermediate for win-ltsc2022
-FROM mcr.microsoft.com/windows/servercore@sha256:45952938708fbde6ec0b5b94de68bcdec3f8c838be018536b1e9e5bd95e6b943 as ltsc2022
-
-FROM ${OS_VERSION} as windows
+FROM mcr.microsoft.com/windows/servercore@sha256:45952938708fbde6ec0b5b94de68bcdec3f8c838be018536b1e9e5bd95e6b943 as windows
```
jpayne3506, line 11: Are we not able to use `windows-host-process-containers-base-image` here?

jpayne3506 (approved):

npm/windows.Dockerfile
rbtr: I don't know and I'm not trying to make big changes to NPM, so I made the build work and left it

jpayne3506: Fair enough. We can look at this another time.

rbtr:
@huntergregory FYI since this needs an NPM reviewer - there should be no net change to NPM here, I only had to update the Dockerfile to match the pattern of everything else in the build matrix

npm/windows.Dockerfile
rbtr: @huntergregory FYI since this needs an NPM reviewer - there should be no net change to NPM here, I only had to update the Dockerfile to match the pattern of everything else in the build matrix

.pipelines/npm/npm-conformance-tests.yaml
```diff
@@ -74,18 +74,16 @@ stages:
...
             name: npm
             os: windows
-            os_version: ltsc2022
       steps:
         - template: ../containers/container-template.yaml
           parameters:
             arch: $(arch)
             name: $(name)
             os: $(os)
-            os_version: $(os_version)
```
huntergregory, line 88: NPM only supports 2022. Is this change a cause for concern in that regard?

rbtr: it should not be, 2022 is still able to pull this image

huntergregory (approved):

rbtr:
cc @tyler-lloyd @paulgmiller for viz

timraymond:
I'm going to say LGTM, but only in that there's nothing obviously wrong here to my eye. I don't know all the fiddly SKUs and IDs for windows, so I'll let others with more specific knowledge of those judge that aspect of this PR.

rbtr:
> I'm going to say LGTM, but only in that there's nothing obviously wrong here to my eye. I don't know all the fiddly SKUs and IDs for Windows (and I trust that @rbtr probably found the right ones), so I'll let others with more specific knowledge of those judge that aspect of this PR.

I figured out all the fiddly Windows SKUs way back when I originally set up multiplat image builds... this _removes_ them because it turns out we don't need them (or any OS in our Windows base images at all, really)

nddq (approved):

paulyufan2 (approved):

PR #3109, opened 2024-11-06
Fix: Add default route on delegatedNIC interface

paulyufan2:
**Reason for Change**:

This PR is to fix one issue in swiftv2 Windows scenario:
When a pod is created, the default route is added on infra vnet:
root@swiftv2-pod-3:/# ip route
default via 10.244.2.1 dev eth0 metric 1
It leads to ping a VM IP in the same VNET that cannot work.

There are two issues::
1.CNS does not provide the default route to CNI;
2.CNI should only add the default route to secondary interface customer vnet; on Swiftv2 scenario, skipDefaultRoutes is set to true for infraNIC interface and false for a secondary interface; so if !info.skipDefaultRoutes, then add default route.

Add dummy default route and pass to CNI as a workaround;

After this fix after a pod creation is done, the route table shows:
**root@swiftv2-pod-1:/# ip route
default via 10.241.0.1 dev eth1**
10.241.0.0/16 dev eth1 proto kernel scope link src 10.241.0.55
10.244.0.0/16 dev eth0 proto kernel scope link src 10.244.1.103

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/middlewares/k8sSwiftV2_linux.go
+++ b/cns/middlewares/k8sSwiftV2_linux.go
@@ -101,3 +101,5 @@ func addRoutes(cidrs []string, gatewayIP string) []cns.Route {
 func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(_ *cns.PodIpInfo, _ v1alpha1.InterfaceInfo, _ string) error {
 	return nil
 }
+
+func (k *K8sSWIFTv2Middleware) addDefaultRoute(_ *cns.PodIpInfo, _ string) {}
--- a/cns/middlewares/k8sSwiftV2_linux.go
+++ b/cns/middlewares/k8sSwiftV2_linux.go
@@ -102,4 +102,4 @@ func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(_ *cns.PodIpInfo, 
 	return nil
 }
 
-func (k *K8sSWIFTv2Middleware) addDefaultRoute(_ *cns.PodIpInfo, _ string) {}
+func (k *K8sSWIFTv2Middleware) addDefaultRoute(*cns.PodIpInfo, string) {}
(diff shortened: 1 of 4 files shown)


aggarwal0009 (dismissed):

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -40,5 +46,12 @@ func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(podIPInfo *cns.Pod
 		},
 		GatewayIPAddress: interfaceInfo.GatewayIP,
 	}
+	// assign the default route with gateway IP
```
tamilmani1989, line 49: i feel its not added in right function. either create new function to add this or add in function where routes are configured

kmurudi (approved):

isaac-dasan (dismissed):

cns/middlewares/k8sSwiftV2_linux.go
```diff
@@ -101,3 +101,5 @@ func addRoutes(cidrs []string, gatewayIP string) []cns.Route {
 func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(_ *cns.PodIpInfo, _ v1alpha1.InterfaceInfo, _ string) error {
 	return nil
 }
+
+func (k *K8sSWIFTv2Middleware) addDefaultRoute(_ *cns.PodIpInfo, _ string) {}
```
rbtr, line 105: nit:
```suggestion
func (k *K8sSWIFTv2Middleware) addDefaultRoute(*cns.PodIpInfo, string) {}
```

rbtr (approved):

rbtr (approved):

PR #3113, opened 2024-11-07
Remove l2bridge checker on windows and l2tunnel mode

paulyufan2:
**Reason for Change**:

This PR is to simplify windows l2bridge mode and remove l2tunnel mode from the system.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/azure-windows-swift-overlay-dualstack.conflist
+++ b/cni/azure-windows-swift-overlay-dualstack.conflist
@@ -5,7 +5,6 @@
     "plugins": [
         {
             "type": "azure-vnet",
-            "mode": "bridge",
             "bridge": "azure0",
             "capabilities": {
                 "portMappings": true,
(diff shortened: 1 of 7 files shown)


QxBytes (commented):
could we add a unit test where we pass in some value to nwInfo (type: EndpointInfo) and then call configure endpoint, ensuring that the resulting hcn network type is always l2 bridge?

QxBytes (dismissed):
Confirmed with paul that l2tunnel is not used anywhere. `azure-windows-multitenancy.conflist` is not modified as it could not be readily tested. Only time networking mode is not bridge in windows is if the nic type is a particular value. Linux side should be unaffected.

QxBytes (approved):

PR #3123, opened 2024-11-08
[WIP] ci: 1ES Template

sheylatrudo:
Adds the compliant 1ES Template for security compliance.


huntergregory (commented):
Can we get a green on this for win dockerfile change? At least for ws22 stage. Retry is broken, so a failure would require external ADO manual run.

/azp run NPM Conformance Tests

PR #3135, opened 2024-11-12
Remove bridgeName field from CNI windows conflist files

paulyufan2:
**Reason for Change**:

This PR is to cleanup windows cni config file to remove bridgeName fields

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/azure-windows-multitenancy.conflist
+++ b/cni/azure-windows-multitenancy.conflist
@@ -4,8 +4,6 @@
     "plugins": [
         {
             "type": "azure-vnet",
-            "mode": "bridge",
-            "bridge": "azure0",
             "multiTenancy":true,
             "enableSnatOnHost":true,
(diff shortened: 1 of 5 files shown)


QxBytes (commented):
So just confirming that "bridge" (the field) isn't used in windows, and mode will default to bridge if it is not specified? Could you also run the full pipeline with this change?

paulyufan2:
@QxBytes  pipeline passes

QxBytes (approved):

QxBytes (approved):
On windows, a reference search for the bridgeName field of endpoint info does not reveal any usages. it seems like you're only modifying windows conflists so there shouldn't be any effect.

QxBytes (approved):

PR #3140, opened 2024-11-12
[CNS] add config snapshot event metrics at an interval

jackieluc:
**Reason for Change**:
Inspired by @ramiro-gamarra's PR to emit config snapshots at an interval, we should do the same for CNS for a consistent view for both components.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/configuration/cns_config.json
+++ b/cns/configuration/cns_config.json
@@ -5,6 +5,7 @@
         "HeartBeatIntervalInMins": 30,
         "RefreshIntervalInSecs": 15,
         "SnapshotIntervalInMins": 60,
+        "ConfigSnapshotIntervalInMins": 60,
         "TelemetryBatchIntervalInSecs": 15,
         "TelemetryBatchSizeBytes": 16384
     },
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -79,6 +79,8 @@ type TelemetrySettings struct {
 	DebugMode bool
 	// Interval for sending snapshot events.
 	SnapshotIntervalInMins int
+	// Interval for sending config snapshot events.
+	ConfigSnapshotIntervalInMins int
 	// AppInsightsInstrumentationKey allows the user to override the default appinsights ikey
 	AppInsightsInstrumentationKey string
 }
@@ -175,6 +177,10 @@ func setTelemetrySettingDefaults(telemetrySettings *TelemetrySettings) {
 
 	if telemetrySettings.SnapshotIntervalInMins == 0 {
 		telemetrySettings.SnapshotIntervalInMins = 60
+	}
+
+	if telemetrySettings.ConfigSnapshotIntervalInMins == 0 {
+		telemetrySettings.ConfigSnapshotIntervalInMins = 60
 	}
 }
 
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -178,10 +178,6 @@ func setTelemetrySettingDefaults(telemetrySettings *TelemetrySettings) {
 	if telemetrySettings.SnapshotIntervalInMins == 0 {
 		telemetrySettings.SnapshotIntervalInMins = 60
 	}
-
-	if telemetrySettings.ConfigSnapshotIntervalInMins == 0 {
-		telemetrySettings.ConfigSnapshotIntervalInMins = 60
-	}
 }
 
 // set managed setting defaults
--- a/cns/metric/configsnapshot.go
+++ b/cns/metric/configsnapshot.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Azure/azure-container-networking/aitelemetry"
 	"github.com/Azure/azure-container-networking/cns/configuration"
 	"github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/pkg/errors"
 )
 
 // SendCNSConfigSnapshot emits CNS config periodically
@@ -21,7 +22,7 @@ func SendCNSConfigSnapshot(ctx context.Context, config *configuration.CNSConfig)
 
 	event, err := createCNSConfigSnapshotEvent(config)
 	if err != nil {
-		logger.Errorf("[Azure CNS] SendCNSConfigSnapshot failed to create event at an interval: %v", err)
+		logger.Errorf("[Azure CNS] SendCNSConfigSnapshot: %v", err)
 		return
 	}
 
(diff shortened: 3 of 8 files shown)


cns/configuration/cns_config.json
```diff
@@ -5,6 +5,7 @@
         "HeartBeatIntervalInMins": 30,
         "RefreshIntervalInSecs": 15,
         "SnapshotIntervalInMins": 60,
+        "ConfigSnapshotIntervalInMins": 60,
```
rbtr, line 8: CNS can't be reconfigured without being restarted. What do we gain by re-emitting the config at an interval?

ramiro-gamarra: @rbtr If we only emit this during start up, any CNS process that outlives the retention policy in Kusto (~90 days) won't be shown in the snapshots. Some constant logging of the config for multiple CNS nodes also gives us the ability to know what nodes in a cluster are active.

cns/metric/configsnapshot.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		logger.Errorf("[Azure CNS] SendCNSConfigSnapshot: %v", err)
+		return
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
```
ramiro-gamarra, line 33: this will only start emitting the first config snapshot one hour after cns starts. you may want to emit this right away as well.

cns/metric/configsnapshot.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return aitelemetry.Event{}, errors.Wrap(err, "failed to marshal config")
+	}
+
+	cs := md5.Sum(bb) //nolint:gosec // used for checksum
+	csStr := string(cs[:])
+
+	event := aitelemetry.Event{
+		EventName:  logger.ConfigSnapshotMetricsStr,
+		ResourceID: csStr, // not guaranteed unique, instead use VM ID and Subscription to correlate
```
ramiro-gamarra, line 50: some guid generated at process start up might be more useful than a hash that multiple processes will share

cns/configuration/cns_config.json
rbtr: I am convinced that we should re-emit the config, but I don't think that we should use that for liveness. And if we're not using it for liveness, and the retention is 90 days, 60 minutes seems a bit frequent.

cns/metric/configsnapshot.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func createCNSConfigSnapshotEvent(config *configuration.CNSConfig) (aitelemetry.Event, error) {
+	bb, err := json.Marshal(config) //nolint:musttag // no tag needed for config
+	if err != nil {
+		return aitelemetry.Event{}, errors.Wrap(err, "failed to marshal config")
+	}
+
+	cs := md5.Sum(bb) //nolint:gosec // used for checksum
+	csStr := string(cs[:])
```
rbtr, line 46: why do we need a checksum?

cns/metric/configsnapshot_test.go
```diff
@@ -0,0 +1,40 @@
...
+	expected := aitelemetry.Event{
+		EventName:  logger.ConfigSnapshotMetricsStr,
+		ResourceID: csStr,
+		Properties: map[string]string{
+			logger.CNSConfigPropertyStr:            string(bb),
+			logger.CNSConfigMD5CheckSumPropertyStr: csStr,
+		},
+	}
+
+	assert.Equal(t, expected, event)
```
ramiro-gamarra, line 39: this test is a bit odd since it duplicates internal logic of (read: tightly couples) `createCNSConfigSnapshotEvent`. if you change certain things in the function, the test will need to change as well. IMO a more robust test might do something like
```go
event, err := createCNSConfigSnapshotEvent(config)
require.NoError(t, err)

assert.Equal(t, logger.ConfigSnapshotMetricsStr, event.EventName)
assert.NotEmpty(t, event.ResourceID)
// some assertion that the test contains a config property with valid json
// some assertion that the config deserialized from those bytes matches your input config
```
let me know if this makes sense

cns/metric/configsnapshot.go
timraymond: Yeah, in theory you'd be able to filter lines where these get emitted and easily pinpoint when the config changed, which is probably of most interest.  Doing that precise filtering would also probably filter out all of the CNS startup fanfare that happens in the logs too, so I think it remains useful even if the config is immutable while CNS is running.

ramiro-gamarra: we have a separate table for this type of telemetry (snapshots) so no need to worry about logs at least

cns/configuration/cns_config.json
msvik: One of the usage scenarios for this is live site investigations where we want to determine the config with which CNS started. With that context, 60 minutes is not frequent.

cns/configuration/configuration.go
```diff
@@ -176,6 +178,10 @@ func setTelemetrySettingDefaults(telemetrySettings *TelemetrySettings) {
 	if telemetrySettings.SnapshotIntervalInMins == 0 {
 		telemetrySettings.SnapshotIntervalInMins = 60
 	}
+
+	if telemetrySettings.ConfigSnapshotIntervalInMins == 0 {
+		telemetrySettings.ConfigSnapshotIntervalInMins = 60
+	}
```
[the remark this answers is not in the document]
jackieluc: IMO it should be default because it'll provide us observability. what would be the cases where we don't want this?

cns/metric/configsnapshot.go
jackieluc: i'm not sure if this will add much benefit because the GUID still won't have much usage outside of this context

cns/metric/configsnapshot_test.go
jackieluc: that makes sense, i've refactored it

cns/metric/configsnapshot.go
jackieluc: added to emit right away as well. we log the config on CNS start up, but doesn't hurt to emit it right away either

cns/configuration/configuration.go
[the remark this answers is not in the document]
rbtr: I think it needs to be possible to turn it off, and using an interval of 0 to indicate that is intuitive

cns/metric/configsnapshot.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2018 Microsoft. All rights reserved.
+// MIT License
```
rbtr, line 2: nit remove this

jackieluc: removed

rbtr (approved):

PR #3143, opened 2024-11-13
[NPM] Update NPM base image to Distroless

rejain456:
**Reason for Change**:

Updated NPM base image to Distroless. 

**Issue Fixed**:

Updating npm base image eliminated the segmentation error we would see with ~20k chains added in iptables-nft with base image Ubuntu 20.04

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/npm/npm-conformance-tests.yaml
+++ b/.pipelines/npm/npm-conformance-tests.yaml
@@ -102,14 +102,17 @@ stages:
             AZURE_CLUSTER: "conformance-v2-background"
             PROFILE: "v2-background"
             IS_STRESS_TEST: "false"
+            OS_SKU: "Ubuntu"
           v2-ws22:
             AZURE_CLUSTER: "conformance-v2-ws22"
             PROFILE: "v2-default-ws22"
             IS_STRESS_TEST: "false"
+            OS_SKU: "Ubuntu"
           v2-linux-stress:
             AZURE_CLUSTER: "conformance-v2-linux-stress"
             PROFILE: "v2-background"
             IS_STRESS_TEST: "true"
+            OS_SKU: "Ubuntu"
       pool:
         name: $(BUILD_POOL_NAME_DEFAULT)
         demands:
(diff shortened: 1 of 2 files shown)


.pipelines/npm/npm-conformance-tests.yaml
```diff
@@ -210,6 +213,7 @@ stages:
                 --resource-group $(RESOURCE_GROUP) \
                 --name $(AZURE_CLUSTER) \
                 --network-plugin azure
```
huntergregory, line 215: going to need a `\`

.pipelines/npm/npm-conformance-tests.yaml
```diff
@@ -210,6 +213,7 @@ stages:
...
+                ----os-sku $(OS_SKU)
```
huntergregory: too many `-` :)

huntergregory:
@rbtr Do you have any advice for the docker file?

rbtr:
This looks good, it's very similar to [what I did for CNS](https://github.com/Azure/azure-container-networking/blob/master/cns/Dockerfile#L27-L28)

npm/linux.Dockerfile
```diff
@@ -2,12 +2,36 @@ FROM mcr.microsoft.com/oss/go/microsoft/golang:1.23 AS builder
 ARG VERSION
 ARG NPM_AI_PATH
 ARG NPM_AI_ID
+RUN apt-get update && apt-get install -y iptables ipset ca-certificates conntrack grep && apt-get autoremove -y && apt-get clean
```
rbtr, line 5: make this a separate build stage to keep the concerns distinct and so it can parallelize - let the Go builder only build Go and we'll make an iptables target to get those bits.

move to something like 
```Dockerfile
FROM mariner-core AS iptables
RUN tdnf install -y iptables
```

then change your copies to `--from iptables`

PR #3149, opened 2024-11-14
feat: refactor cni telemetry

QxBytes:
**Reason for Change**:

Currently the telemetry CNI is sending is insufficient to debug CNI issues. This PR refactors the cni telemetry to send more and better quality logs.
- Moves telemetry into a package level variable so it is made accessible everywhere
- Removes sending certain metrics as they are not used
- Sets the subcontext to the container id. The container id is kept consistent throughout CNI calls for the same pod, meaning an ADD and DEL call (and all related logs) for the same pod will have the same subcontext/container id. The container id is also what is stored in stateless mode as one of the keys.
- Sets the operation id before any telemetry events are sent. The operation id is used for sampling should we end up enabling it.

Examples of Logged information (Will be added in a separate PR-- this PR is focused on refactoring)
- CNI add network configuration, arguments
- CNI add completion with endpoint info struct information (contains hns endpoint id and hns network id), interface results from the ipam invoker, and any error that occurred
- CNI del network configuration, arguments
- CNI del completion with error that occurred
- HNS Endpoint struct **before** creation / HNS Endpoint Id during deletion
- HNS Network struct **before** creation / HNS Network Id during deletion
- Deletion/Release of each IP (even if does not exist)
- Mapping sent to CNS during stateless CNI mode during Update Endpoint State
- Exact CNS response from CNS ipam invoker
- Exact CNS response from multitenancy ipam invoker
- Transparent vlan creating/deleting vlan veth interface

Potential additions:
- endpoint and network structs saved to azure-vnet.json statefile

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
Pipeline run to prove logs sent to kusto: https://msazure.visualstudio.com/One/_build/results?buildId=108208651&view=results
Passing run: https://msazure.visualstudio.com/One/_build/results?buildId=108563465&view=results

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -15,7 +15,6 @@ import (
 	"github.com/Azure/azure-container-networking/cni"
 	"github.com/Azure/azure-container-networking/cni/api"
 	"github.com/Azure/azure-container-networking/cni/log"
-	telemetryclient "github.com/Azure/azure-container-networking/cni/telemetry/client"
 	"github.com/Azure/azure-container-networking/cni/util"
 	"github.com/Azure/azure-container-networking/cns"
 	cnscli "github.com/Azure/azure-container-networking/cns/client"
@@ -30,6 +29,7 @@ import (
 	"github.com/Azure/azure-container-networking/platform"
 	nnscontracts "github.com/Azure/azure-container-networking/proto/nodenetworkservice/3.302.0.744"
 	"github.com/Azure/azure-container-networking/store"
+	"github.com/Azure/azure-container-networking/telemetry"
 	cniSkel "github.com/containernetworking/cni/pkg/skel"
 	cniTypes "github.com/containernetworking/cni/pkg/types"
 	cniTypesCurr "github.com/containernetworking/cni/pkg/types/100"
@@ -39,8 +39,8 @@ import (
 
 // matches if the string fully consists of zero or more alphanumeric, dots, dashes, parentheses, spaces, or underscores
 var (
-	allowedInput = regexp.MustCompile(`^[a-zA-Z0-9._\-\(\) ]*$`)
-	telemetry    = telemetryclient.Telemetry
+	allowedInput    = regexp.MustCompile(`^[a-zA-Z0-9._\-\(\) ]*$`)
+	telemetryClient = telemetry.Client
 )
 
 const (
@@ -295,10 +295,10 @@ func (plugin *NetPlugin) getPodInfo(args string) (name, ns string, err error) {
 }
 
 func (plugin *NetPlugin) setCNIReportDetails(containerID, opType, msg string) {
-	telemetry.Settings().OperationType = opType
-	telemetry.Settings().SubContext = containerID
-	telemetry.Settings().EventMessage = msg
-	telemetry.Settings().Version = plugin.Version
+	telemetryClient.Settings().OperationType = opType
+	telemetryClient.Settings().SubContext = containerID
+	telemetryClient.Settings().EventMessage = msg
+	telemetryClient.Settings().Version = plugin.Version
 }
 
 func addNatIPV6SubnetInfo(nwCfg *cni.NetworkConfig,
@@ -383,10 +383,10 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 	if err != nil {
 		return err
 	}
-	telemetry.Settings().ContainerName = k8sPodName + ":" + k8sNamespace
+	telemetryClient.Settings().ContainerName = k8sPodName + ":" + k8sNamespace
 
 	plugin.setCNIReportDetails(args.ContainerID, CNI_ADD, "")
-	telemetry.SendEvent(fmt.Sprintf("[cni-net] Processing ADD command with args {ContainerID:%v Netns:%v IfName:%v Args:%v Path:%v StdinData:%s}.",
+	telemetryClient.SendEvent(fmt.Sprintf("[cni-net] Processing ADD command with args {ContainerID:%v Netns:%v IfName:%v Args:%v Path:%v StdinData:%s}.",
 		args.ContainerID, args.Netns, args.IfName, args.Args, args.Path, args.StdinData))
 
 	iptables.DisableIPTableLock = nwCfg.DisableIPTableLock
@@ -437,7 +437,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 			zap.Any("IPs", cniResult.IPs),
 			zap.Error(log.NewErrorWithoutStackTrace(err)))
 
-		telemetry.SendEvent(fmt.Sprintf("ADD command completed with [ipamAddResult]: %s [epInfos]: %s [error]: %v ", ipamAddResult.PrettyString(), network.FormatStructPointers(epInfos), err))
+		telemetryClient.SendEvent(fmt.Sprintf("ADD command completed with [ipamAddResult]: %s [epInfos]: %s [error]: %v ", ipamAddResult.PrettyString(), network.FormatStructPointers(epInfos), err))
 	}()
 
 	ipamAddResult = IPAMAddResult{interfaceInfo: make(map[string]network.InterfaceInfo)}
@@ -493,7 +493,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		// triggered only in swift v1 multitenancy
 		// dual nic multitenancy -> two interface infos
 		// multitenancy (swift v1) -> one interface info
-		telemetry.Settings().Context = "AzureCNIMultitenancy"
+		telemetryClient.Settings().Context = "AzureCNIMultitenancy"
 		plugin.multitenancyClient.Init(cnsClient, AzureNetIOShim{})
 
 		// Temporary if block to determining whether we disable SNAT on host (for multi-tenant scenario only)
@@ -973,7 +973,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 		logger.Info("DEL command completed",
 			zap.String("pod", k8sPodName),
 			zap.Error(log.NewErrorWithoutStackTrace(err)))
-		telemetry.SendEvent(fmt.Sprintf("DEL command completed: [released ip]: %+v [podname]: %s [namespace]: %s [error]: %v", nwCfg.IPAM.Address, k8sPodName, k8sNamespace, err))
+		telemetryClient.SendEvent(fmt.Sprintf("DEL command completed: [released ip]: %+v [podname]: %s [namespace]: %s [error]: %v", nwCfg.IPAM.Address, k8sPodName, k8sNamespace, err))
 	}()
 
 	// Parse network configuration from stdin.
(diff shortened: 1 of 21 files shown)


cni/telemetry/client/telemetry_client.go
```diff
@@ -0,0 +1,72 @@
...
+func (c *TelemetryClient) sendLog(msg string) {
+	if c.logger == nil {
+		return
+	}
+	c.logger.Info("Telemetry Event", zap.String("message", msg))
+}
+
+func (c *TelemetryClient) SendEvent(msg string) {
+	c.sendLog(msg)
+	c.sendTelemetry(msg)
```
ramiro-gamarra, line 71: what is the difference between log and telemetry here?

cni/telemetry/client/telemetry_client.go
```diff
@@ -0,0 +1,72 @@
...
+
+var Telemetry = NewTelemetryClient(&telemetry.CNIReport{})
+
+func NewTelemetryClient(report *telemetry.CNIReport) *TelemetryClient {
+	return &TelemetryClient{
+		CNIReportSettings: report,
+	}
+}
+
+func (c *TelemetryClient) ConnectTelemetry(logger *zap.Logger) {
```
ramiro-gamarra, line 32: why can't this be part of the constructor? wouldn't that remove the need for nil checks?

QxBytes: If we `ConnectTelemetry` in `NewTelemetryClient` I believe we would try to connect to the telemetry service during tests, which I do not want to occur. `Connect / StartAndConnect` are also called by different main methods (stateless vs. non-stateless cni) whereas `NewTelemetryClient` is called for both. Adding attempts to _only_ connect to `NewTelemetryClient` (via `ConnectTelemetry`) would be incorrect behavior in non-stateless cni because we want to start the telemetry service before any connection is attempted.

cni/telemetry/client/telemetry_client.go
QxBytes: logs are sent to `azure-vnet.log`. Telemetry is sent to the telemetry service on the node which then sends it to application insights.

cni/network/plugin/main.go
```diff
@@ -52,6 +54,16 @@ func printVersion() {
 	fmt.Printf("Azure CNI Version %v\n", version)
 }
 
+// send error report to hostnetagent if CNI encounters any error.
+func reportPluginError(reportManager *telemetry.ReportManager, tb *telemetry.TelemetryBuffer, err error) {
+	logger.Error("Report plugin error")
+	reflect.ValueOf(reportManager.Report).Elem().FieldByName("ErrorMessage").SetString(err.Error())
```
timraymond, line 60: Why do we have to use reflection here?

cni/network/stateless/main.go
```diff
@@ -50,6 +52,16 @@ func printVersion() {
 	fmt.Printf("Azure CNI Version %v\n", version)
 }
 
+// send error report to hostnetagent if CNI encounters any error.
+func reportPluginError(reportManager *telemetry.ReportManager, tb *telemetry.TelemetryBuffer, err error) {
+	logger.Error("Report plugin error")
+	reflect.ValueOf(reportManager.Report).Elem().FieldByName("ErrorMessage").SetString(err.Error())
```
timraymond, line 58: Same reflection question

cni/network/plugin/main.go
QxBytes: refactored to remove reflection and use the single telemetry package level variable instead

cni/network/stateless/main.go
QxBytes: see above

timraymond:
LGTM on @ramiro-gamarra 's approval

cni/network/network.go
```diff
@@ -479,18 +436,12 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 			zap.String("pod", k8sPodName),
 			zap.Any("IPs", cniResult.IPs),
 			zap.Error(log.NewErrorWithoutStackTrace(err)))
+
+		telemetry.SendEvent(fmt.Sprintf("ADD command completed with [ipamAddResult]: %s [epInfos]: %s [error]: %v ", ipamAddResult.PrettyString(), network.FormatStructPointers(epInfos), err))
```
ramiro-gamarra, line 440: Per one of your previous comments, `telemetry.SendEvent` will both write to `azure-vnet.log` as well as send to telemetry service. Aren't we then writing to the log file twice here if `logger` also writes to the file? Can we keep this PR to **only** a pure refactor without introducing new side effects?

cni/network/network.go
```diff
@@ -479,18 +436,12 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
 
 	ipamAddResult = IPAMAddResult{interfaceInfo: make(map[string]network.InterfaceInfo)}
 
-	// Parse Pod arguments.
-	k8sPodName, k8sNamespace, err := plugin.getPodInfo(args.Args)
-	if err != nil {
-		return err
-	}
-
-	plugin.report.ContainerName = k8sPodName + ":" + k8sNamespace
```
ramiro-gamarra, line 492: what is the impact of removing this from `report`?

cni/network/network.go
```diff
@@ -1011,25 +955,21 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
 		zap.String("path", args.Path),
 		zap.ByteString("stdinData", args.StdinData))
-	sendEvent(plugin, fmt.Sprintf("[cni-net] Processing DEL command with args {ContainerID:%v Netns:%v IfName:%v Args:%v Path:%v, StdinData:%s}.",
-		args.ContainerID, args.Netns, args.IfName, args.Args, args.Path, args.StdinData))
 
 	defer func() {
 		logger.Info("DEL command completed",
 			zap.String("pod", k8sPodName),
 			zap.Error(log.NewErrorWithoutStackTrace(err)))
+		telemetry.SendEvent(fmt.Sprintf("DEL command completed: [released ip]: %+v [podname]: %s [namespace]: %s [error]: %v", nwCfg.IPAM.Address, k8sPodName, k8sNamespace, err))
```
ramiro-gamarra, line 972: same as above. if `logger` already writes to the file, then we are just duplicating data, as well as going away from structured logs.

cni/telemetry/client/telemetry_client.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/Azure/azure-container-networking/telemetry"
+	"go.uber.org/zap"
+)
+
+const (
+	telemetryNumberRetries          = 5
+	telemetryWaitTimeInMilliseconds = 200
+)
+
+type TelemetryClient struct {
```
ramiro-gamarra, line 17: There's a lot of stutter in the naming. Can't there just be a `telemetry` package with a `Client` type?

cni/telemetry/client/telemetry_client.go
```diff
@@ -0,0 +1,118 @@
...
+	cniReportSettings *telemetry.CNIReport
+	tb                *telemetry.TelemetryBuffer
+	logger            *zap.Logger
+	lock              sync.Mutex
+}
+
+type TelemetryInterface interface {
```
ramiro-gamarra, line 24: There's only a single implementation of this interface. This is not needed.

cni/telemetry/client/telemetry_client.go
```diff
@@ -0,0 +1,118 @@
...
+const (
+	telemetryNumberRetries          = 5
+	telemetryWaitTimeInMilliseconds = 200
+)
+
+type TelemetryClient struct {
+	cniReportSettings *telemetry.CNIReport
+	tb                *telemetry.TelemetryBuffer
+	logger            *zap.Logger
+	lock              sync.Mutex
```
ramiro-gamarra, line 21: is CNI spinning goroutines? What is this lock serializing?

ramiro-gamarra (dismissed):
I may still be missing some details about the purpose of this refactor, but seems to me that logs are getting duplicated and the abstractions introduced are not cleaning up the code much yet.

cni/telemetry/client/telemetry_client.go
QxBytes: moved the telemetry client to the existing telemetry package

cni/telemetry/client/telemetry_client.go
QxBytes: currently we don't send telemetry from multiple goroutines as far as I can tell, but if we want to send telemetry from anywhere in the codebase in the future, I would prefer to serialize this and not worry about races etc. later, but I can remove the lock for now if you would like.

cni/network/network.go
QxBytes: yes you are correct, I've removed the duplicate logger statement in the telemetry client

cni/network/network.go
QxBytes: the code to populate the `ContainerName` field has been moved up in the function to line 386 (it should still be in the report).

cni/network/network.go
QxBytes: see above-- I was trying to replicate what I was seeing in the cnslogger but since we log before anyway this isn't helpful

behzad-mir (approved):
lgtm

network/endpoint.go
```diff
@@ -150,15 +150,31 @@ type apipaClient interface {
 	CreateHostNCApipaEndpoint(ctx context.Context, networkContainerID string) (string, error)
 }
 
+func FormatStructPointers[T any](slice []*T) string {
```
tamilmani1989, line 153: `FormatPointerArrayToString` or `TransformPointerArrayToString`?

tamilmani1989: what if array is not a pointer?

tamilmani1989, line 153: add function description and its purpose

network/manager.go
```diff
@@ -422,13 +422,13 @@ func (nm *networkManager) UpdateEndpointState(eps []*endpoint) error {
 	}
 
 	ifnameToIPInfoMap := generateCNSIPInfoMap(eps) // key : interface name, value : IPInfo
-	for _, ipinfo := range ifnameToIPInfoMap {
-		logger.Info("Update endpoint state", zap.String("hnsEndpointID", ipinfo.HnsEndpointID), zap.String("hnsNetworkID", ipinfo.HnsNetworkID),
+	for key, ipinfo := range ifnameToIPInfoMap {
+		logger.Info("Update endpoint state", zap.String("key", key), zap.String("hnsEndpointID", ipinfo.HnsEndpointID), zap.String("hnsNetworkID", ipinfo.HnsNetworkID),
```
tamilmani1989, line 426:
```suggestion
		logger.Info("Update endpoint state", zap.String("ifname", key), zap.String("hnsEndpointID", ipinfo.HnsEndpointID), zap.String("hnsNetworkID", ipinfo.HnsNetworkID),
```

network/endpoint.go
QxBytes: if we try to pass in an array of structs instead of pointers, there will be a compilation error

cni/network/network.go
```diff
@@ -421,20 +379,20 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
+
+	plugin.setCNIReportDetails(args.ContainerID, CNI_ADD, "")
+	telemetryClient.SendEvent(fmt.Sprintf("[cni-net] Processing ADD command with args {ContainerID:%v Netns:%v IfName:%v Args:%v Path:%v StdinData:%s}.",
+		args.ContainerID, args.Netns, args.IfName, args.Args, args.Path, args.StdinData))
+
 	iptables.DisableIPTableLock = nwCfg.DisableIPTableLock
-	plugin.setCNIReportDetails(nwCfg, CNI_ADD, "")
 
 	defer func() {
-		operationTimeMs := time.Since(startTime).Milliseconds()
```
vipul-21, line 428: Did we check if this metric is being pushed somewhere ?

cni/network/network.go
```diff
@@ -421,20 +379,20 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 		return err
 	}
 
+	// Parse Pod arguments.
+	k8sPodName, k8sNamespace, err := plugin.getPodInfo(args.Args)
+	if err != nil {
+		return err
+	}
+	telemetryClient.Settings().ContainerName = k8sPodName + ":" + k8sNamespace
```
vipul-21, line 387: Can `Settings()` be `nil` ?

vipul-21:
The changes lgtm to me apart from couple of comments. 
@tamilmani1989 Can you please also take a look as well as.

cni/network/network.go
QxBytes: yes the time to run the add/del call gets pushed to kusto

cni/network/network.go
QxBytes: Settings() comes from c.cniReportSettings which is a pointer, but that is always initialized to at least an empty Client struct assuming the telemetry client is created with NewClient()-- which it always should be as it is created in telemetry_client.go as a package level variable

cni/network/network.go
vipul-21: So why are we removing these metrics ?
I know we aren't using it, but @behzad-mir did end up using these metrics. So would be nice to have it and not depend on custom scripts.

vipul-21: Didn't see the `operationTimeMs := time.Since(startTime).Milliseconds()` still be calculated. Should be good !

vipul-21 (approved):

PR #3151, opened 2024-11-14
update for new home az contract from nma

ZetaoZhuang:
**Reason for Change**:

NMA introduces a new field apiVersion in home az endpoint for the purpose of fixing the ipv6 issue in swift azr, we need to make changes on cns to correspond to this contract change.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


nmagent/responses.go
```diff
@@ -38,7 +38,8 @@ type NCVersionList struct {
 }
 
 type AzResponse struct {
-	HomeAz uint `json:"homeAz"`
+	HomeAz     uint `json:"homeAz"`
+	APIVersion uint `json:"apiVersion"`
```
timraymond, line 42: How is this used? Nothing else seems to change, so why does anything care what APIVersion is present?

ZetaoZhuang: This is going to be consumed in dnc and used for indicating a Dualstack fix from NMA

timraymond: Can you describe the ipv6 fix needed in a little more detail? My concern with this is that nothing about the API is actually changing. We're just adding a cryptic field to it that clients are just supposed to understand. What does it mean when `APIVersion` becomes 3, 4, so on? A more obvious field would be something like `NeedsIPv6Fix = true` or something, but I don't have enough context on the entire problem to recommend a good API (hence the request for more detail).

ZetaoZhuang: Currently NMA is rollout a new fix for the azr swift regarding dualstacks secnarios. they are adding a new field in home az   api to indicate the fix already there, so dnc will check this apiVersion coming from NMA response to decide enable azr or not. This APIVersion will always be 2. we should raise alert if we saw apiVersion other than 2.

ZetaoZhuang: but NMA mentioned this value might change in the future for different purpose

timraymond: If NMAgent is trying to indicate whether or not AZR should be enabled based on some programming they do, then the field should be `EnableAZR = true / false`. We shouldn't care about some fix or whatever that they have to do. Can we push back on this? The fact of NMAgent's APIs being generally not great isn't justification for allowing them to continue to make them worse.

If the answer is "no, the decision is firm," then we should fix this at the system border so we don't have to spread the knowledge about what a `2` means through the entire ACN stack. The NMAgent client (whose designed responsibility is a respository of all of the weird things NMAgent does) should translate this to something meaningful like `EnableAZR` in the response.

smittal22: While I agree with your comment, Tim, NMA wants us to really use APIVersion here where they can probably change version to reuse this in case, they need to provide such a fix for another issue later (just having a boolean for one case might warrant more changes later from their side). That being said, @ZetaoZhuang we should check with NMA if 0 and 2 are the only values we should expect from them.

I agree this is not clean at all, but this is what they're pushing us to use from their side. Currently, this is a hotfix like change for us and ideally, we shouldn't need to use this later.

ZetaoZhuang: I checked with Vishwesh, he confirmed that 2 is the only valid value for now. we should also allow 0 as well. that is the default value of the field when NMA is still running the old version, and apiVersion does not exist

timraymond: Okay, thanks for clarifying that we don't have any control over what they do. I'll make suggestions to isolate their bad design, though, so we don't have to deal with it forever.

cns/restserver/homeazmonitor.go
```diff
@@ -154,7 +154,15 @@ func (h *HomeAzMonitor) Populate(ctx context.Context) {
...
 	}
-	h.update(types.Success, "Get Home Az succeeded", cns.HomeAzResponse{IsSupported: true, HomeAz: azResponse.HomeAz})
+	// validate APIVersion, APIVersion is a uint, so its value >=0
+	// 0 should be valid when NMA version is old and does not have the apiVersion value in home az response
+	if azResponse.APIVersion > 0 && azResponse.APIVersion != 2 {
+		returnMessage := fmt.Sprintf("[HomeAzMonitor] invalid APIVersion value from nmagent: %d", azResponse.APIVersion)
+		returnCode := types.UnexpectedError
+		h.update(returnCode, returnMessage, cns.HomeAzResponse{IsSupported: true})
+		return
+	}
```
timraymond, line 164: All of this knowledge about NMAgent should move to a method of `nmagent.AzResponse`. Something like:

```go
func (a AzResponse) Valid() bool {
       return a.APIVersion >= 0 && a.APIVersion < 3
}

func (a AzResponse) ShouldEnableAZR() bool {
       return a.Valid() && a.APIVersion == 2
}
```

Further, in the `HomeAz` method of the client, you can test for validity of the response and return an error instead stating that there is an invalid APIVersion from NMAgent.

cns/api.go
```diff
@@ -356,6 +356,7 @@ type NmAgentSupportedApisResponse struct {
 type HomeAzResponse struct {
 	IsSupported bool `json:"isSupported"`
 	HomeAz      uint `json:"homeAz"`
+	APIVersion  uint `json:"apiVersion"`
```
timraymond, line 359: This should indicate whether or not we should Enable AZR (per earlier discussion). We can set this from doing:

```go
resp, err := nmaClient.HomeAz(ctx)
if err != nil {
       return errors.Wrap(err, "getting home az")
}
return cns.HomeAzResponse{
       EnableAZR: resp.ShouldEnableAZR(),
}
```

Then this can be directly consumed in clients without further interpretation about the meaning of codes.

timraymond (changes requested):

cns/restserver/homeazmonitor.go
timraymond: Hard disagree. `APIVersion = 2` has a specific meaning, designed for this specific purpose. There is no room for interpretation on behalf of clients. I'm flexible on the name (`AZREnablementRequested`, `NMAgentAppliedTheIPv6Fix`, or something), but the name must convey the true meaning so that it is self-documenting for future readers and for other users of the `nmagent` package.

nmagent/responses.go
```diff
@@ -38,7 +38,19 @@ type NCVersionList struct {
...
+
+func (az AzResponse) Valid() bool {
+	// 0 should be valid when NMA version is old and does not have the apiVersion value in home az response
+	//nolint:gomnd // these magic numbers are made by nma design
+	return az.APIVersion == 0 || az.APIVersion == 2
+}
+
+func (az AzResponse) NmaAppliedTheIPV6Fix() bool {
+	//nolint:gomnd // this magic number is made by nma design
+	return az.APIVersion == 2
```
msvik, line 53: Shouldn't this be az.APIVersion >= 2 instead of az.APIVersion == 2 since it can change in future?

cns/restserver/homeazmonitor.go
```diff
@@ -154,7 +154,14 @@ func (h *HomeAzMonitor) Populate(ctx context.Context) {
 		h.update(returnCode, returnMessage, cns.HomeAzResponse{IsSupported: true})
 		return
 	}
-	h.update(types.Success, "Get Home Az succeeded", cns.HomeAzResponse{IsSupported: true, HomeAz: azResponse.HomeAz})
+	// validate APIVersion value
+	if !azResponse.Valid() {
```
msvik, line 158: Is it ok to IsValid check after checking AZ? Should we do Valid check before it?

nmagent/responses.go
ZetaoZhuang: Vishewesh mentioned that for now we only want to allow APIVersion as 2, and he want us to raise alerts if we get APIVersion other than 2.  from what he means, when we get APIVersion as "3", we should mark it as an invalid version number and then raise alert on it. That way, we may not want it to indicate the fix is applied.

cns/restserver/homeazmonitor.go
ZetaoZhuang: imo, its ok. the order does not matter here. the APIversion is for indicating the ipv6 fix which is independent of the value of home az. so it could be home az is 0, but APIversion is either valid or invalid.

nmagent/responses.go
msvik: Since this API version field is being introduced as general purpose field which can be used for future bugs also, we should account for what would be AZ behavior if it becomes 3 some day.

cns/api.go
ZetaoZhuang: addressed

timraymond (dismissed):

nmagent/responses.go
```diff
@@ -37,8 +43,74 @@ type NCVersionList struct {
...
+func (h HomeAZFix) String() string {
+	switch h {
+	case HomeAZFixInvalid:
+		return "HomeAZFixInvalid"
+	case HomeAZFixIPv6:
+		return "HomeAZFixIPv6"
+	default:
+		return "Unknown HomeAZ Fix"
+	}
+}
```
smittal22, line 59: @timraymond / @ZetaoZhuang Do we need this function? Don't think we're using this anywhere.

timraymond: Yes, this is called by the `fmt` package using reflection. Otherwise, it's an int with no meaning in logs.

nmagent/responses.go
```diff
@@ -37,8 +43,74 @@ type NCVersionList struct {
...
+			if requested == present {
+				found = true
+			}
+		}
+
+		if !found {
+			return false
+		}
+	}
+	return true
```
smittal22, line 113: Can get rid of all this code if we just change it to uint and take decisions based on 0 or 2 value. Anything other than that is invalidated already.

nmagent/responses_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+			got := test.resp.ContainsFixes(test.fixes...)
+
+			exp := test.exp
+			if got != exp {
+				t.Error("unexpected response from ContainsFixes: exp:", exp, "got:", got)
```
smittal22, line 62: Just for my understanding. Is there a strong reason to use Testify package's _**.Error**_ and _**.Fatal**_ over require package? IMO require is more succinct and readable.

smittal22 (commented):
Let me know once you address these.

nmagent/responses.go
smittal22: I see, thanks for the explanation. I didn't know about reflection, so I read up on it.

nmagent/responses.go
```diff
@@ -37,8 +43,74 @@ type NCVersionList struct {
...
+
+const (
+	HomeAZFixInvalid HomeAZFix = iota
+	HomeAZFixIPv6
+)
+
 type AzResponse struct {
-	HomeAz uint `json:"homeAz"`
+	HomeAz       uint
+	AppliedFixes []HomeAZFix
```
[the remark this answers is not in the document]
timraymond: Hard disagree. From everything I've seen, the meaning of this uint is to describe fixes that have been applied that would be of interest to a client for whatever reason. It may be *easier* to have a uint here, but it's also a horrible user experience that spreads esoteric knowledge about magic numbers throughout multiple codebases. Yes, today `apiVersion` can be `0` or `2`, but it is better engineering to assume that there will be future values for this to convey other fixes that have been applied. This is a slice because, in my experience in this industry, sooner or later there will be more than one fix of interest to clients. This is here to accommodate that eventuality.

This design is more robust, because however this contract changes in the future will require changes in **exactly one place** without breaking clients.

nmagent/responses.go
timraymond: I think I've answered this in the other comment, but here's another example:

```go
// do something if the apiVersion indicates that the ipv6 fix was applied
//nolint:gomnd // see nmagent documentation for these values
switch response.apiVersion {
       case 2:
              // do something here
       case 0:
              // do something else
       default:
              return fmt.Errorf("unexpected apiVersion value %d", response.apiVersion)
}
```
vs.

```go
if response.ContainsFixes(nmagent.HomeAZFixIPv6) {
       // do something here
}
```

The latter is self-documenting code, doesn't require dealing with invalid values (easy to forget), and doesn't require disabling a common linter all the time. It's also more flexible, as it supports detecting N fixes.

nmagent/responses_test.go
smittal22: We don't need to wait for this. Resolving.

nmagent/responses.go
smittal22: Sounds reasonable.

nmagent/responses.go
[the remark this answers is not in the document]
smittal22: Okay. Let's keep it this way then.  It is more future proof, whether we need it right now or not.

nmagent/responses_test.go
timraymond: I find `require` and `assert` to be obtuse, and they obscure simple things like "which is the expected value and the one that you received." It's easy to accidentally swap them and get confusing output in test failures. Using `testing` lets me have better control over the failure output so that I can make sure that it's useful to whoever is reading it (usually me, but not always). Also, `testing.(*T).Error` and `testing.(*T).Fatal` are part of the standard library and have a small API surface area that you can keep in your head. `require` and `assert` have an enormous number of functions to assert everything under the sun... that I usually never need. If I do routinely need to perform some complex assertion, I just create a helper with `testing.(*T).Helper` and use that. This is good enough to test the entire standard library, and it's good enough for me.

If I choose to use an assertion package on a project, I much prefer the lightweight https://github.com/matryer/is than testify.

Anyway, I realize others like `testify` and don't share my opinion of it. Given that, I only ask people to pick one and stick with it (don't mix styles in the same test) and leave the choice to them.

smittal22 (approved):
LGTM

nmagent/responses_test.go
smittal22: Thanks for the explanation, Tim. Sounds good.

smittal22 (approved):

ZetaoZhuang:
Hey @timraymond , can you approve again? Thanks

timraymond (approved):

PR #3156, opened 2024-11-18
ci: combine go tests such that any test failure will report a failure for the entire task

QxBytes:
**Reason for Change**:

When the go tests are split up, only the status of the last command run determines the status of the entire step/task. Any previous steps, even if they exited with a failure, won't cause the task to visibly fail. By combining all tests together in one command, if this single command fails (due to any of the tests failing), the step/task will visibly fail on the pipeline.

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
**Will fail in the pipeline until npm windows tests are fixed (this is expected, as previously those tests were silently failing).**
When those npm changes are merged in, I will rebase this PR against master to resolve the errors.
Will be backported to `release/v1.5`

### DIFF
--- a/.pipelines/templates/run-unit-tests.yaml
+++ b/.pipelines/templates/run-unit-tests.yaml
@@ -39,9 +39,8 @@ stages:
         steps:
           # Only run one go test per script
           - script: |
-              set CGO_ENABLED=1
               cd azure-container-networking/
-              go test -race -timeout 30m ./npm/... ./cni/... ./platform/...
+              go test -timeout 30m ./npm/... ./cni/... ./platform/...
             retryCountOnTaskFailure: 3
             name: "TestWindows"
             displayName: "Run Windows Tests"
--- a/.pipelines/templates/run-unit-tests.yaml
+++ b/.pipelines/templates/run-unit-tests.yaml
@@ -37,10 +37,11 @@ stages:
         pool:
           name: "$(BUILD_POOL_NAME_DEFAULT_WINDOWS_ALT)"
         steps:
-          # Test changes under review; only run one go test per script
+          # Only run one go test per script
           - script: |
+              set CGO_ENABLED=1
               cd azure-container-networking/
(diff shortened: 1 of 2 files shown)


.pipelines/templates/run-unit-tests.yaml
```diff
@@ -37,13 +37,10 @@ stages:
...
+          # Test changes under review; only run one go test per script
           - script: |
-              cd npm/
-              go test ./...
-              cd ../cni/
-              go test ./...
-              cd ../platform/
-              go test ./...
+              cd azure-container-networking/
+              go test ./npm/... ./cni/... ./platform/...
```
timraymond, line 43: I would add `-race` and `-timeout` as well.

QxBytes: hmm I've tried adding -race but it seems to require CGO_ENABLED=1 (enabled), but on this particular node agent running the script (which is windows), it gives an error saying gcc is not found, making me believe this isn't possible at the moment, unless you have any other ideas of how to run -race? https://msazure.visualstudio.com/One/_build/results?buildId=108557696&view=logs&j=1e95c5bf-f785-5c1d-a4c1-33b038e882fb&t=4058cacc-5989-52ae-5341-5beff346fc28

QxBytes: so if I'm understanding this correctly, the windows node at the moment doesn't have c compiler, and so we cannot have cgo enabled. In the future, we will need to somehow update the node agent to include gcc? In the meantime a `-timeout` has been added.

timraymond: No C compiler is required on the node for CGO_ENABLED=1. CGo means that some dependencies are dynamically linked (e.g. DNS resolution). As long as the DLLs for performing those functions are present on the system, it will work.

timraymond (approved):

rejain456 (approved):

PR #3157, opened 2024-11-19
feat: Creating NNC with HomeAz info in AKS-Swift Workflows when CNS starts up behind a configuration flag

nairashu:
**Reason for Change**:
feat: Creating NNC with HomeAz info in AKS-Swift Workflows when CNS starts up behind a configuration flag

**Issue Fixed**:
N/A

**Requirements**:

- [X] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,37 +3,37 @@ issues:
   max-issues-per-linter: 0
   new-from-rev: origin/master
 linters:
-  presets: 
-  - bugs
-  - error
-  - format
-  - performance
-  - unused
+  presets:
+    - bugs
+    - error
+    - format
+    - performance
+    - unused
   disable:
-  - maligned
-  - scopelint
+    - maligned
+    - scopelint
+    - gomnd
   enable:
-  - exportloopref
-  - goconst
-  - gocritic
-  - gocyclo
-  - gofmt
-  - gomnd
-  - goprintffuncname
-  - gosimple
-  - lll
-  - misspell
-  - nakedret
-  - promlinter
-  - revive
+    - exportloopref
+    - goconst
+    - gocritic
+    - gocyclo
+    - gofmt
+    - goprintffuncname
+    - gosimple
+    - lll
+    - misspell
+    - nakedret
+    - promlinter
+    - revive
 linters-settings:
   gocritic:
     enabled-tags:
-    - "diagnostic"
-    - "style"
-    - "performance"
+      - "diagnostic"
+      - "style"
+      - "performance"
     disabled-checks:
-    - "hugeParam"
+      - "hugeParam"
   govet:
     check-shadowing: true
   lll:
--- a/.pipelines/mdnc/azure-cns-cni-1.4.39.1.yaml
+++ b/.pipelines/mdnc/azure-cns-cni-1.4.39.1.yaml
@@ -12,7 +12,7 @@ metadata:
 rules:
 - apiGroups: ["acn.azure.com"]
   resources: ["nodenetworkconfigs"]
-  verbs: ["get", "list", "watch", "patch", "update"]
+  verbs: ["create", "delete", "get", "list", "watch", "patch", "update"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -35,6 +35,7 @@ type CNSConfig struct {
 	EnableSubnetScarcity        bool
 	EnableSwiftV2               bool
 	InitializeFromCNI           bool
+	EnableHomeAz                bool
 	KeyVaultSettings            KeyVaultSettings
 	MSISettings                 MSISettings
 	ManageEndpointState         bool
--- a/cns/configuration/configuration.go
+++ b/cns/configuration/configuration.go
@@ -35,7 +35,7 @@ type CNSConfig struct {
 	EnableSubnetScarcity        bool
 	EnableSwiftV2               bool
 	InitializeFromCNI           bool
-	EnableHomeAz                bool
+	EnableHomeAZ                bool
 	KeyVaultSettings            KeyVaultSettings
 	MSISettings                 MSISettings
 	ManageEndpointState         bool
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -633,3 +633,11 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
 func (service *HTTPRestService) SetVFForAccelnetNICs() error {
 	return service.setVFForAccelnetNICs()
 }
+
+// GetHomeAz - Get the Home Az for the Node where CNS is running.
+func (service *HTTPRestService) GetHomeAz() (homeAzResponse cns.GetHomeAzResponse) {
+	service.RLock()
+	homeAzResponse = service.homeAzMonitor.GetHomeAz(context.TODO())
+	service.RUnlock()
+	return
+}
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -635,9 +635,13 @@ func (service *HTTPRestService) SetVFForAccelnetNICs() error {
 }
 
 // GetHomeAz - Get the Home Az for the Node where CNS is running.
-func (service *HTTPRestService) GetHomeAz(ctx context.Context) (homeAzResponse cns.GetHomeAzResponse) {
+func (service *HTTPRestService) GetHomeAz(ctx context.Context) (cns.GetHomeAzResponse, error) {
 	service.RLock()
-	homeAzResponse = service.homeAzMonitor.GetHomeAz(ctx)
-	service.RUnlock()
-	return
+	defer service.RUnlock()
+	homeAzResponse := service.homeAzMonitor.GetHomeAz(ctx)
+	if homeAzResponse.Response.ReturnCode == types.NotFound {
+		return homeAzResponse, errors.New(homeAzResponse.Response.Message)
+	}
+
+	return homeAzResponse, nil
 }
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -635,9 +635,9 @@ func (service *HTTPRestService) SetVFForAccelnetNICs() error {
 }
 
 // GetHomeAz - Get the Home Az for the Node where CNS is running.
-func (service *HTTPRestService) GetHomeAz() (homeAzResponse cns.GetHomeAzResponse) {
+func (service *HTTPRestService) GetHomeAz(ctx context.Context) (homeAzResponse cns.GetHomeAzResponse) {
 	service.RLock()
-	homeAzResponse = service.homeAzMonitor.GetHomeAz(context.TODO())
+	homeAzResponse = service.homeAzMonitor.GetHomeAz(ctx)
 	service.RUnlock()
 	return
 }
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -1308,7 +1308,8 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 		// Create Node Network Config CRD and update the Home Az field with the cache value from the HomeAz Monitor
 		nnc = createBaseNNC(node)
 	}
-	if err := directcli.Create(ctx, &nnc); err != nil {
+
+	if err = directcli.Create(ctx, &nnc); err != nil {
 		return errors.Wrap(err, "failed to create base NNC")
 	}
 
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -1306,36 +1306,46 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 	// Create the base NNC CRD if HomeAz is enabled
 	if cnsconfig.EnableHomeAz {
 		homeAzResponse := httpRestServiceImplementation.GetHomeAz(ctx)
-		availabilityZone := strconv.FormatUint(uint64(homeAzResponse.HomeAzResponse.HomeAz), 10)
-		logger.Printf("[Azure CNS] HomeAz: %s", availabilityZone)
+		logger.Printf("[Azure CNS] HomeAz: %s", strconv.FormatUint(uint64(homeAzResponse.HomeAzResponse.HomeAz), 10))
 		// Create Node Network Config CRD and update the Home Az field with the cache value from the HomeAz Monitor
 		var nnc *v1alpha.NodeNetworkConfig
 		if nnc, err = directnnccli.Get(ctx, types.NamespacedName{Namespace: "kube-system", Name: nodeName}); err != nil {
 			logger.Errorf("[Azure CNS] failed to get existing NNC: %v", err)
 		}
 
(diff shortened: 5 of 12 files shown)


timraymond:
@nairashu If this is still WIP, can you mark this as draft?

cns/configuration/configuration.go
```diff
@@ -35,6 +35,7 @@ type CNSConfig struct {
 	EnableSubnetScarcity        bool
 	EnableSwiftV2               bool
 	InitializeFromCNI           bool
+	EnableHomeAz                bool
```
ramiro-gamarra, line 38: Maybe this should be `EnableNNCCreation`. It will be clearer when AKS reviews toggles to CNS config.

cns/service/main.go
```diff
@@ -1303,6 +1303,52 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 	// TODO(rbtr): nodename and namespace should be in the cns config
 	directscopedcli := nncctrl.NewScopedClient(directnnccli, types.NamespacedName{Namespace: "kube-system", Name: nodeName})
 
+	// Create the base NNC CRD if HomeAz is enabled
+	if cnsconfig.EnableHomeAz {
+		homeAzResponse := httpRestServiceImplementation.GetHomeAz(ctx)
+		az := homeAzResponse.HomeAzResponse.HomeAz
```
ramiro-gamarra, line 1309: This cannot be used as is. When there is a cache miss, the response for GetHomeAz will contain some error values which should be checked. I don't know why the original function doesn't return an error to make this clearer.

cns/service/main.go
```diff
@@ -1303,6 +1303,52 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+		logger.Printf("[Azure CNS] HomeAz: %d", az)
+		// Create Node Network Config CRD and update the Home Az field with the cache value from the HomeAz Monitor
+		var nnc *v1alpha.NodeNetworkConfig
+		if nnc, err = directnnccli.Get(ctx, types.NamespacedName{Namespace: "kube-system", Name: nodeName}); err != nil {
+			logger.Errorf("[Azure CNS] failed to get existing NNC: %v", err)
+		}
+
+		newNNC := createBaseNNC(node)
+		if nnc == nil {
+			logger.Printf("[Azure CNS] Creating new base NNC")
```
ramiro-gamarra, line 1319: Error conditions are not clear here. If there is a transient network error, the get should be retried no? Otherwise, is there a way to detect that we reached the apiserver and it returned "not found"? That should be the only time when we attempt to create the nnc.

cns/service/main.go
```diff
@@ -1303,6 +1303,52 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+			newNNC.Spec.AvailabilityZone = az
+			if err = directcli.Create(ctx, newNNC); err != nil {
+				return errors.Wrap(err, "failed to create base NNC")
+			}
+		} else {
+			logger.Printf("[Azure CNS] Patching existing NNC with new Spec with HomeAz %d", az)
```
ramiro-gamarra, line 1325: If this is just patching an existing nnc, shouldn't we just use the fetched `nnc` variable and set the availability zone?

cns/service/main.go
```diff
@@ -1303,6 +1303,52 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+			newNNC.DeletionTimestamp = nnc.DeletionTimestamp
+
+			// Delete existing NNC and create new one with updated HomeAz
+			if err = directcli.Delete(ctx, nnc); err != nil {
+				return errors.Wrap(err, "[Azure CNS]: failed to delete existing NNC")
+			}
+
+			if err = directcli.Create(ctx, newNNC); err != nil {
+				return errors.Wrap(err, "[Azure CNS]: failed to create new NNC")
+			}
```
ramiro-gamarra, line 1347: Is there no update method? Deleting and recreating seems very prone to issues.

ramiro-gamarra (changes requested):
A few questions/thoughts

cns/service/main.go
nairashu: I was running into issue trying to simply Patch the NNC. So after testing went this route. Let me retry again. Tried a few combinations of testing. Also attempted to simply update the az field on the existing NNC but that is possible only once the new NNC changes in DNC-RC are rolled out first before these updates.

cns/service/main.go
nairashu: I attempted that and ran into the error as my testing was with a scenario where DNC-RC created an NNC without this field. Probably should update this to the suggestion.

cns/configuration/configuration.go
nairashu: We typically control the component configurations closer to our service feature naming so keeping it related to HomeAz makes more sense as the flag is to use the HomeAzMonitor overall to retrieve the Az and then create the NNC from that.

.golangci.yml
```diff
@@ -3,37 +3,37 @@ issues:
...
+    - error
+    - format
+    - performance
+    - unused
   disable:
-  - maligned
-  - scopelint
+    - maligned
+    - scopelint
+    - gomnd
```
timraymond, line 15: I fully support removing `gomnd` since I think it does more harm than good, but I also think any linter changes should be done in a separate PR so we can have a discussion about just doing that.

cns/configuration/configuration.go
timraymond: I know that "Az" is everywhere, so it's probably impractical to change at this point, but "AZ" is an abbreviation, so this should be `EnableHomeAZ` by typical Go naming conventions.

cns/service/main.go
```diff
@@ -1303,6 +1303,52 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
 
+	// Create the base NNC CRD if HomeAz is enabled
+	if cnsconfig.EnableHomeAz {
+		homeAzResponse := httpRestServiceImplementation.GetHomeAz(ctx)
+		az := homeAzResponse.HomeAzResponse.HomeAz
+		logger.Printf("[Azure CNS] HomeAz: %d", az)
+		// Create Node Network Config CRD and update the Home Az field with the cache value from the HomeAz Monitor
+		var nnc *v1alpha.NodeNetworkConfig
+		if nnc, err = directnnccli.Get(ctx, types.NamespacedName{Namespace: "kube-system", Name: nodeName}); err != nil {
+			logger.Errorf("[Azure CNS] failed to get existing NNC: %v", err)
```
timraymond, line 1314: There's no return here, so this will continue. Is that intentional?

cns/service/main.go
```diff
@@ -1303,6 +1303,52 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
...
+		}
+
+		newNNC := createBaseNNC(node)
+		if nnc == nil {
+			logger.Printf("[Azure CNS] Creating new base NNC")
+			newNNC.Spec.AvailabilityZone = az
+			if err = directcli.Create(ctx, newNNC); err != nil {
+				return errors.Wrap(err, "failed to create base NNC")
+			}
+		} else {
```
timraymond, line 1324: Usually `else` is not preferred in Go if it can be eliminated.

.golangci.yml
nairashu: We can discuss that in a seperate PR to remove gomnd

cns/configuration/configuration.go
nairashu: Changed this to EnableHomeAZ

cns/restserver/internalapi.go
```diff
@@ -633,3 +633,11 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
 func (service *HTTPRestService) SetVFForAccelnetNICs() error {
 	return service.setVFForAccelnetNICs()
 }
+
+// GetHomeAz - Get the Home Az for the Node where CNS is running.
+func (service *HTTPRestService) GetHomeAz(ctx context.Context) (homeAzResponse cns.GetHomeAzResponse) {
+	service.RLock()
+	homeAzResponse = service.homeAzMonitor.GetHomeAz(ctx)
+	service.RUnlock()
+	return
```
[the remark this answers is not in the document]
nairashu: Actually missed adding the homeAzResponse field to the return value here.

cns/service/main.go
nairashu: Yes, that is intentional as we could have a scenario where the NNC does not exist while we are trying to move the NNC creation logic gradually over to CNS.

cns/restserver/internalapi.go
```diff
@@ -633,3 +633,11 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
 func (service *HTTPRestService) SetVFForAccelnetNICs() error {
 	return service.setVFForAccelnetNICs()
 }
+
+// GetHomeAz - Get the Home Az for the Node where CNS is running.
+func (service *HTTPRestService) GetHomeAz(ctx context.Context) (homeAzResponse cns.GetHomeAzResponse) {
+	service.RLock()
+	homeAzResponse = service.homeAzMonitor.GetHomeAz(ctx)
+	service.RUnlock()
```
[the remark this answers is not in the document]
nairashu: Added the defer but also noticed this file is missing defer in multiple other places

cns/restserver/internalapi.go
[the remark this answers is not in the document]
nairashu: I have updated the API logic here a bit to validate for cache miss scenario as well.

cns/service/main.go
nairashu: Updated the client API to handle the cache miss logic and error out and checking for errors in here before updating homeAz value.

cns/service/main.go
timraymond: Might be worth leaving a comment to that effect.

cns/restserver/internalapi.go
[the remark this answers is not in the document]
timraymond: "Leave the campsite cleaner than you found it" applies :)

cns/service/main.go
nairashu: Might take up Ramiro's suggestion and add a retrier incase of network error in reaching the API server or atleast checking if the error is clearly a NNC Not Found error before proceeding

cns/service/main.go
nairashu: Adding a retrier in here.

cns/service/main.go
nairashu: Removing the else condition.

cns/restserver/internalapi.go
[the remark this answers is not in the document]
nairashu: Totally on the same page with that philosophy so did add in the defer above.

cns/service/main.go
nairashu: Direct Patching works in the case where the NNC metadata has the AvailabilityZone metadata but if it is not present. Then direct patching causes and a failure during writing it to the API server.

cns/service/main.go
nairashu: Logic here has changed.

cns/service/main.go
nairashu: I was able to use the Update method effectively by using the resourceID from the existing NNC so updated to use that API.

cns/configuration/configuration.go
```diff
@@ -35,6 +35,7 @@ type CNSConfig struct {
 	EnableSubnetScarcity        bool
 	EnableSwiftV2               bool
 	InitializeFromCNI           bool
+	EnableHomeAZ                bool
```
ramiro-gamarra, line 38: I think the previous comment got lost but I think we need further discussion on the name of this field. I don't think `EnableHomeAZ` is the correct name because:
- Home AZ is already enabled when mode is direct, which makes this flag confusing.
- The major feature is the NNC creation, which should be independent of home az being available or not.
I still think `EnableNNCCreation` is a better name. Would like to get feedback from @thatmattlong and @rbtr on this.

cns/service/main.go
```diff
@@ -1303,6 +1303,68 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 	// TODO(rbtr): nodename and namespace should be in the cns config
 	directscopedcli := nncctrl.NewScopedClient(directnnccli, types.NamespacedName{Namespace: "kube-system", Name: nodeName})
 
+	// Create the base NNC CRD if HomeAz is enabled
+	if cnsconfig.EnableHomeAZ {
```
ramiro-gamarra, line 1307: better to extract the scope opened by this flag to a different function. it will help clean up some code paths.

cns/configuration/configuration.go
rbtr: agree, flag should tightly describe what function it does, not the abstract name of the scenario.
`CreateNNC` would be even better

.golangci.yml
rbtr: will you reset the changes in this file?

.pipelines/mdnc/azure-cns-cni-1.4.39.1.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
 rules:
 - apiGroups: ["acn.azure.com"]
   resources: ["nodenetworkconfigs"]
-  verbs: ["get", "list", "watch", "patch", "update"]
+  verbs: ["create", "delete", "get", "list", "watch", "patch", "update"]
```
rbtr, line 15: i don't think we'll need delete

cns/restserver/internalapi.go
```diff
@@ -633,3 +633,15 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
+func (service *HTTPRestService) GetHomeAz(ctx context.Context) (cns.GetHomeAzResponse, error) {
+	service.RLock()
+	defer service.RUnlock()
+	homeAzResponse := service.homeAzMonitor.GetHomeAz(ctx)
+	if homeAzResponse.Response.ReturnCode == types.NotFound {
+		return homeAzResponse, errors.New(homeAzResponse.Response.Message)
+	}
+
+	return homeAzResponse, nil
+}
```
rbtr, line 647: what value is this adding over calling `service.homeAzMonitor.GetHomeAz` directly?

cns/service/main.go
rbtr: i would like to see NNC creation decoupled completely from HomeAz

.pipelines/mdnc/azure-cns-cni-1.4.39.1.yaml
nairashu: The design is to move the node reconcile logic over to CNS. 
It makes more sense to create and cleanup the NNC based on when the node is created and deleted. As a result, I added the operational capability to delete the NNC as well to CNS as eventually we can clean up the nodes faster in this case. DNC, DNC-RC and DNCCleanup can work async as it does today to cleanup the NC and other associations. 

Why do we want to avoid CNS from doing this?

.golangci.yml
timraymond: I opened https://github.com/Azure/azure-container-networking/pull/3330

PR #3166, opened 2024-11-22
chore: add release branches to dependabot gomod updates

rbtr:
**Reason for Change**:

Adds release branches to dependabot config for gomod updates.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -62,3 +62,43 @@ updates:
   commit-message:
     prefix: "deps"
   open-pull-requests-limit: 10
+- package-ecosystem: "gomod"
+  directory: "/"
+  schedule:
+    interval: "daily"
+  reviewers:
+    - "azure/azure-sdn-members"
+  commit-message:
+    prefix: "deps"
+  open-pull-requests-limit: 10
+  target-branch: "release/v1.5"
+- package-ecosystem: "gomod"
+  directory: "/azure-ipam"
+  schedule:
+    interval: "daily"
+  reviewers:
+    - "azure/azure-sdn-members"
+  commit-message:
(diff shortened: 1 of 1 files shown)


jpayne3506 (approved):
As it stands this is good, but because of us pinning release trains to other deps I would recommend looking at [disabling updates for certain versions + deps](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#example-disabling-version-updates-for-some-dependencies)

thatmattlong (approved):

.github/dependabot.yaml
```diff
@@ -62,3 +62,43 @@ updates:
   commit-message:
     prefix: "deps"
   open-pull-requests-limit: 10
+- package-ecosystem: "gomod"
+  directory: "/"
```
huntergregory, line 66: curious the logic for having the same line for v1.5 for both `/` and `/azure-ipam`? Assuming the former will cover all directories?

rbtr:
> As it stands this is good, but because of us pinning release trains to other deps I would recommend looking at [disabling updates for certain versions + deps](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#example-disabling-version-updates-for-some-dependencies)

once we interact with dependabot on a PR it opens, it tracks these things automatically

.github/dependabot.yaml
rbtr: no, this line tells dependabot which directory to look in for the `go.mod`. We need it to update both `/go.mod` and `/azure-ipam/go.mod`

PR #3205, opened 2024-11-26
fix: route issues on Swiftv2 Windows

paulyufan2:
**Reason for Change**:

This is the long term solution to fix Swiftv2 Windows Routes issues to make sure windows cns gets all required cidrs from AKS.
Fixes include:
1.Windows CNS gets infravnet/pod/node cidrs from configs
2.Add these routes and send them to the CNI

**Issue Fixed**:

**Background of the issue:**
In Linux Swiftv2, CNS fetches infravnet/pod/node cidrs from the node's envs set by AKS; we should do same thing for Windows;
 

**Requirements**:

Require to validate this PR after AKS-RP PR to be merged

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/middlewares/k8sSwiftV2.go
```diff
@@ -249,3 +249,52 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
...
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to get infraVNETCIDRs from env")
+	}
+	infraVNETCIDRsv4, infraVNETCIDRsv6, err := utils.ParseCIDRs(infraVNETCIDRs)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "failed to parse infraVNETCIDRs")
+	}
+	fmt.Printf("Received infraVNETCIDRsv4: %v, infraVNETCIDRsv6: %v", infraVNETCIDRs, infraVNETCIDRsv6)
+
+	// Get and parse podCIDRs from env
```
kmurudi, line 269: is this commented code old or kept for testing some alternative?

paulyufan2: just temp fix, reverted back

QxBytes (commented):
Could you add a bit more about why these new cidrs need to be added and what problem this PR addresses? Like what couldn't be done before that can be done now with these routes being added? How did things work (or fail to work) beforehand?

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -50,11 +60,58 @@ func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(podIPInfo *cns.Pod
...
+	if err != nil {
+		return "", errors.Wrap(err, "failed to parse cidr")
+	}
+	ip = ip.To4()
+	ip[3] = 1
+
+	return ip.String(), nil
+}
+
+// Windows uses .1 as the gateway IP for each CIDR
```
shreyashastantram, line 84: is this comment in the right place?

cns/middlewares/k8sSwiftV2_linux.go
```diff
@@ -86,7 +46,14 @@ func (k *K8sSWIFTv2Middleware) setRoutes(podIPInfo *cns.PodIpInfo) error {
 	return nil
 }
 
-func addRoutes(cidrs []string, gatewayIP string) []cns.Route {
+// assignSubnetPrefixLengthFields is a no-op for linux swiftv2 as the default prefix-length is sufficient
+func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(_ *cns.PodIpInfo, _ v1alpha1.InterfaceInfo, _ string) error {
+	return nil
+}
+
```
shreyashastantram, line 53: Why default route is a no-op in linux?

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -50,11 +60,58 @@ func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(podIPInfo *cns.Pod
 	return nil
 }
 
-// add default route with gateway IP to podIPInfo
-func (k *K8sSWIFTv2Middleware) addDefaultRoute(podIPInfo *cns.PodIpInfo, gwIP string) {
+// add default route with gateway IP to podIPInfo for delegated interface
+func (k *K8sSWIFTv2Middleware) addDefaultRoute(podIPInfo *cns.PodIpInfo, gatewayIP string) {
```
shreyashastantram, line 64: Do you need a different function for this? can't you add a default route in addroutes?

cns/middlewares/k8sSwiftV2_linux.go
paulyufan2: because of HNS problem and we must send default route to HNS on windows

cns/middlewares/k8sSwiftV2_linux_test.go
```diff
@@ -342,10 +343,10 @@ func TestSetRoutesSuccess(t *testing.T) {
 		} else {
 			assert.Equal(t, ipInfo.SkipDefaultRoutes, false)
 		}
-
 	}
+
 	for i := range podIPInfo {
-		assert.DeepEqual(t, podIPInfo[i].Routes, desiredPodIPInfo[i].Routes)
+		reflect.DeepEqual(podIPInfo[i].Routes, desiredPodIPInfo[i].Routes)
```
timraymond, line 349: Preference towards `github.com/google/go-cmp/cmp.Equal` over `reflect.DeepEqual` (it's a drop-in replacement)

cns/middlewares/k8sSwiftV2.go
```diff
@@ -249,3 +249,59 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
...
+	v4IPs = append(v4IPs, podCIDRsV4...)
+	v6IPs = append(v6IPs, podCIDRv6...)
+
+	return v4IPs, v6IPs, nil
+}
+
+// CNS gets node and service CIDRs from configuration env and parse them to get the v4 and v6 IPs
+func (k *K8sSWIFTv2Middleware) GetCidrs() (v4IPs, v6IPs []string, err error) {
+	v4IPs = []string{}
+	v6IPs = []string{}
```
timraymond, line 278: Don't use named returns please, they have surprising semantics. For example, these two lines are unnecessary since they are auto-initialized.

cns/middlewares/k8sSwiftV2.go
```diff
@@ -249,3 +249,59 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
 func (k *K8sSWIFTv2Middleware) Type() cns.SWIFTV2Mode {
 	return cns.K8sSWIFTV2
 }
+
+// CNS gets pod CIDRs from configuration env and parse them to get the v4 and v6 IPs
+// Containerd reassigns the IP to the adapter and kernel configures the pod cidr route by default, so windows swiftv2 does not require pod cidr
```
QxBytes, line 254: I feel like the comment here would be better above the OS-specific implementations of getInfraRoutes. GetPodCidrs (the function) by _itself_ doesn't do anything os-specific.

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -50,11 +63,47 @@ func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(podIPInfo *cns.Pod
...
+	for i, cidr := range cidrs {
+		routes[i] = cns.Route{
+			IPAddress:        cidr,
+			GatewayIPAddress: defaultGateway,
+		}
+	}
+	return routes
+}
+
+func (k *K8sSWIFTv2Middleware) getInfraRoutes(podIPInfo *cns.PodIpInfo) ([]cns.Route, error) {
```
QxBytes, line 89: have we asked containerd/hns if the adding pod cidrs can be idempotent? or if there is some setting to make it not error if we try to add that same route twice? It would allow us to avoid all these os specific implementations

paulyufan2: Containerd team confirmed that they prefer us to avoid sending pod cidr as a duplicate route to them because if they change their side, many other scenarios would have to be retested since containerd gcs behavior will get changed completely

cns/middlewares/k8sSwiftV2_linux.go
```diff
@@ -97,9 +64,32 @@ func addRoutes(cidrs []string, gatewayIP string) []cns.Route {
 	return routes
 }
 
-// assignSubnetPrefixLengthFields is a no-op for linux swiftv2 as the default prefix-length is sufficient
-func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(_ *cns.PodIpInfo, _ v1alpha1.InterfaceInfo, _ string) error {
-	return nil
-}
+func (k *K8sSWIFTv2Middleware) getInfraRoutes(podIPInfo *cns.PodIpInfo) ([]cns.Route, error) {
```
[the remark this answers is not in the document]
paulyufan2: yes Riya's PR has refactored the middlewire and I am waiting for her PR to be checked in and I will make changes on top of her PR

cns/middlewares/k8sSwiftV2_linux.go
```diff
@@ -97,9 +64,32 @@ func addRoutes(cidrs []string, gatewayIP string) []cns.Route {
...
+	v4PodIPs, v6PodIPs, err := k.GetPodCidrs()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get pod CIDRs")
+	}
+
+	v4IPs = append(v4IPs, v4PodIPs...)
+	v6IPs = append(v6IPs, v6PodIPs...)
+
+	if ip.Is4() {
+		routes = append(routes, k.addRoutes(v4IPs, overlayGatewayv4)...)
```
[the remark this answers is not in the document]
QxBytes: thanks for clarifying, could we also add that note (regarding the gateway ip) above this function if it is still present after the refactor?

PR #3248, opened 2024-12-06
fix: let Windows CNS use the InClusterConfig

rbtr:
**Reason for Change**:
As of [AKS 1.30](https://github.com/Azure/AKS/releases/tag/2024-06-09), service account tokens refresh every ~1 hour when OIDC is enabled. They were previously valid for a year.
[This `setkubeconfigpath.ps1` script](https://github.com/Azure/azure-container-networking/blob/47b243c42fd16119a96ab6d06eb602ac2ce40e7d/npm/examples/windows/setkubeconfigpath.ps1) was at some point necessary on Windows to create a valid kubeconfig for CNS (and NPM and others), but based on my testing we don't need it today. It copies the token from the token file to create a custom kubeconfig from a template at startup and then we pass CNS _that_ file via `--kubeconfig`. 

The script runs at startup and never re-runs, so the token that exists at Pod start is the token CNS will try to use forever. Yearly token lifespans were long enough that no CNS Pod was ever up long enough to hit token expiration.

This becomes an issue with hourly token lifespans. An hour after Pod start, the token becomes invalid and CNS can no longer auth to the API server. For PodSubnet clusters, this permanently prevents CNS from being able to scale the IPAM pool and provide more Pod IPs.

**Issue Fixed**:

https://github.com/Azure/AKS/issues/4679

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/Dockerfile
+++ b/cns/Dockerfile
@@ -32,8 +32,10 @@ COPY --from=builder /go/bin/azure-cns /usr/local/bin/azure-cns
 ENTRYPOINT [ "/usr/local/bin/azure-cns" ]
 EXPOSE 10090
 
-# skopeo inspect docker://mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0 --format "{{.Name}}@{{.Digest}}"
-FROM mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image@sha256:b4c9637e032f667c52d1eccfa31ad8c63f1b035e8639f3f48a510536bf34032b as windows
+# mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0
+FROM --platform=windows/${ARCH} mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image@sha256:b4c9637e032f667c52d1eccfa31ad8c63f1b035e8639f3f48a510536bf34032b as hpc
+
+FROM hpc as windows
 COPY --from=builder /go/bin/azure-cns /azure-cns.exe
 ENTRYPOINT ["azure-cns.exe"]
 EXPOSE 10090
--- a/cns/Dockerfile
+++ b/cns/Dockerfile
@@ -32,12 +32,8 @@ COPY --from=builder /go/bin/azure-cns /usr/local/bin/azure-cns
 ENTRYPOINT [ "/usr/local/bin/azure-cns" ]
 EXPOSE 10090
 
-# mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0
-FROM --platform=windows/${ARCH} mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image@sha256:b4c9637e032f667c52d1eccfa31ad8c63f1b035e8639f3f48a510536bf34032b as hpc
-
-FROM hpc as windows
-COPY --from=builder /azure-container-networking/cns/kubeconfigtemplate.yaml kubeconfigtemplate.yaml
-COPY --from=builder /azure-container-networking/npm/examples/windows/setkubeconfigpath.ps1 setkubeconfigpath.ps1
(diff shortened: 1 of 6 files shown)


cns/azure-cns-windows.yaml
```diff
@@ -38,14 +38,12 @@ spec:
           command: ["powershell.exe"]
           args: 
             [ 
-              '.\setkubeconfigpath.ps1', ";",
               'powershell.exe', '.\azure-cns.exe',
```
thatmattlong, line 41: drop this one too since `powershell.exe` is the command?

test/integration/manifests/cns/daemonset-windows.yaml
```diff
@@ -52,8 +52,6 @@ spec:
           command: ["powershell.exe"]
           args:
             [
-              '.\setkubeconfigpath.ps1',
-              ";",
               '.\azure-cns.exe',
```
thatmattlong, line 55: maybe just make this the command and drop `powershell.exe`?

cns/azure-cns-windows.yaml
thatmattlong: nvm, suggest doing the same that I mentioned in the comment on the integration test

thatmattlong (changes requested):
no powershell.exe needed I think

cns/Dockerfile
```diff
@@ -32,8 +32,6 @@ EXPOSE 10090
 
 # skopeo inspect docker://mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0 --format "{{.Name}}@{{.Digest}}"
 FROM mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image@sha256:b4c9637e032f667c52d1eccfa31ad8c63f1b035e8639f3f48a510536bf34032b as windows
-COPY --from=builder /azure-container-networking/cns/kubeconfigtemplate.yaml kubeconfigtemplate.yaml
-COPY --from=builder /azure-container-networking/npm/examples/windows/setkubeconfigpath.ps1 setkubeconfigpath.ps1
```
thatmattlong, line 36: we may want to leave these around in the dockerfile for a bit in case the release of cns which has this removed gets ahead of the overlaymgr release that will remove them from the manifest (or overlaymgr rolls back)

rbtr: I think we won't be releasing CNS urgently, the production impact can be mitigated purely through a daemonset change in AKS

jpayne3506: Is there a need to keep `cns/kubeconfigtemplate.yaml` and `npm/examples/windows/setkubeconfigpath.ps1` in the repo any longer?

rbtr: likely no but they are referenced in the NPM build, and we need to make sure nothing we're building out of this branch can land on k8s <1.28

jpayne3506 (approved):

cns/Dockerfile
rbtr: the AKS fix has been released for long enough that this is not a concern

paulyufan2 (dismissed):

paulyufan2 (dismissed):

cns/Dockerfile.tmpl
```diff
@@ -36,8 +36,6 @@ EXPOSE 10090
 FROM --platform=windows/${ARCH} {{.WIN_HPC_PIN}} as hpc
 
 FROM hpc as windows
-COPY --from=builder /azure-container-networking/cns/kubeconfigtemplate.yaml kubeconfigtemplate.yaml
-COPY --from=builder /azure-container-networking/npm/examples/windows/setkubeconfigpath.ps1 setkubeconfigpath.ps1
```
jpayne3506, line 40: https://github.com/Azure/azure-container-networking/blob/188f0bc8705eec9ba12ceea13ca3baaa874e6b49/.pipelines/build/dockerfiles/cns.Dockerfile.tmpl#L9-L10 This needs to be reflected over here as well.

jpayne3506 (changes requested):

jpayne3506 (approved):

PR #3269, opened 2024-12-13
feat: CNS checks apiserver in healthz

tyler-lloyd:
**Reason for Change**:

CNS should regularly check if it is able to reach the apiserver to mitigate the risk of losing access
and then failing silently. The issue linked below is a scenario where CNS was not reloading the
refreshed SA token from disk resulting in controller-runtime's failure to update its cache and watch
pods and NNCs. This unauthorized failure went unnoticed until workloads were scaled up on the node,
at which point CNS was unable to update the NNC to request for more IPs so the pods were stuck in
Pending.

For now, this will just add a checker when the `ChannelMode` is `CRD`. Any other modes will get the
default `Ping` checker built-in to controller-runtime.

**Issue Fixed**:

Follow up for https://github.com/azure/aks/issues/4679

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/healthserver/healthz.go
+++ b/cns/healthserver/healthz.go
@@ -1,7 +1,6 @@
 package healthserver
 
 import (
-	"fmt"
 	"net/http"
 
 	"github.com/Azure/azure-container-networking/cns"
@@ -22,16 +21,20 @@ func init() {
 	utilruntime.Must(v1alpha.AddToScheme(scheme))
 }
 
-func NewHealthzHandlerWithChecks(cnsConfig *configuration.CNSConfig) http.Handler {
+// NewHealthzHandlerWithChecks will return a [http.Handler] for CNS's /healthz endpoint.
+// Depending on what we expect CNS to be able to read (based on the [configuration.CNSConfig])
+// then the checks registered to the handler will test for those expectations. For example, in
+// ChannelMode: CRD, the health check will ensure that CNS is able to list NNCs successfully.
+func NewHealthzHandlerWithChecks(cnsConfig *configuration.CNSConfig) (http.Handler, error) {
 	cfg, err := ctrl.GetConfig()
 	if err != nil {
-		panic(fmt.Errorf("unable to get config: %w", err))
+		return nil, errors.Wrap(err, "unable to get a kubeconfig")
 	}
 	cli, err := client.New(cfg, client.Options{
 		Scheme: scheme,
 	})
 	if err != nil {
-		panic(fmt.Errorf("unable to create client: %w", err))
+		return nil, errors.Wrap(err, "unable to create a client")
 	}
 
 	checks := make(map[string]healthz.Checker)
@@ -54,5 +57,5 @@ func NewHealthzHandlerWithChecks(cnsConfig *configuration.CNSConfig) http.Handle
 	// otherwise it will look for a check named "healthz" and return a 404 if not there.
 	return http.StripPrefix("/healthz", &healthz.Handler{
 		Checks: checks,
-	})
+	}), nil
 }
--- a/cns/healthserver/healthz.go
+++ b/cns/healthserver/healthz.go
@@ -1,8 +1,11 @@
 package healthserver
 
 import (
+	"fmt"
 	"net/http"
 
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/configuration"
 	"github.com/Azure/azure-container-networking/crd/nodenetworkconfig/api/v1alpha"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


cns/healthserver/healthz.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	checks := map[string]healthz.Checker{
+		"nnc": func(req *http.Request) error {
+			ctx := req.Context()
+			// we just care that we're allowed to List NNCs so set limit to 1 to minimize
+			// additional load on apiserver
+			if err := cli.List(ctx, &v1alpha.NodeNetworkConfigList{}, &client.ListOptions{
```
rbtr, line 39: I wonder if listing NNCs is the best we can do - this may be immediately problematic for cilium nodesubnet where we now run CNS but do not expect NNCs (or even to talk to the API server? @santhoshmprabhu)...

tyler-lloyd: anything in the cnsconfig that's reliable to tell us if we need to read NNCs? `CNIConflistScenario         string`? or maybe try and list pods (keeping limit at 1) if ipamv2 is enabled?

santhoshmprabhu: `ChannelMode==AzureHost` is how we configure nodesubnet. @rbtr, maybe `ChannelMode==CRD || ChannelMode==MultiTenantCRD`  is the right check for NNCs?

tyler-lloyd: is there a tl;dr for what these channel modes map to or what they mean?

```
// ChannelMode :- CNS channel modes
const (
	Direct         = "Direct"
	Managed        = "Managed"
	CRD            = "CRD"
	MultiTenantCRD = "MultiTenantCRD"
	AzureHost      = "AzureHost"
)
```

cns/healthserver/healthz.go
```diff
@@ -0,0 +1,54 @@
...
+func NewHealthzHandlerWithChecks() http.Handler {
+	cfg, err := ctrl.GetConfig()
+	if err != nil {
+		panic(err)
+	}
+	cli, err := client.New(cfg, client.Options{
+		Scheme: schema,
+	})
+	if err != nil {
+		panic(err)
```
timraymond, line 31: Same thing: please remove the panic.

timraymond (changes requested):

cns/healthserver/healthz.go
```diff
@@ -0,0 +1,54 @@
...
+var schema = runtime.NewScheme()
+
+func init() {
+	utilruntime.Must(v1alpha.AddToScheme(schema))
+}
+
+func NewHealthzHandlerWithChecks() http.Handler {
+	cfg, err := ctrl.GetConfig()
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
tyler-lloyd: done

cns/healthserver/healthz.go
tyler-lloyd: done

cns/service/main.go
```diff
@@ -642,7 +642,13 @@ func main() {
 			return nil
 		}),
 	}
-	go healthserver.Start(z, cnsconfig.MetricsBindAddress, &healthz.Handler{}, readyChecker)
+
+	healthzHandler, err := healthserver.NewHealthzHandlerWithChecks(cnsconfig)
```
ramiro-gamarra, line 646: If I'm reading the code correctly, this code is not behind any flags and will fail if there's no available kubeconfig/can't create an apiserver client. CNS does not always run in k8s nodes and this would be a breaking change for non-k8s scenarios. Please put this behind some flag.

ramiro-gamarra (dismissed):
This change should be behind a flag since it is otherwise a breaking change for non-k8s scenarios.

cns/service/main.go
tyler-lloyd: it is only going to try and read from kubeconfig when the `ChannelMode` is `CRD`. Is that ever set for non k8s scenarios?

ramiro-gamarra: Usually set to `direct` for when DNC is in the same vnet as CNS and can ping it directly. This constructor (`NewHealthzHandlerWithChecks`) errors if it can't build a controller-runtime client though, no?

tyler-lloyd: yes, you're right my bad. forgot I put the client initialization outside the CRD block. updated in this commit https://github.com/Azure/azure-container-networking/pull/3269/commits/f4c50b565c27e19cde378a47e1df113bfd2332c8 ... I still feel like using ChannelMode CRD is safe here since it's already being used to fall into this section (https://github.com/Azure/azure-container-networking/blob/281770bf9714eba4e6b1fe413fe0fb253226d450/cns/service/main.go#L819) which calls `InitializeCRDState` which unconditionally tries to load a kubeconfig the same way we're doing here.

cns/healthserver/healthz.go
tyler-lloyd: @rbtr @santhoshmprabhu lmk what you think now. The nnc checker is only initialized for channelMode `CRD`. I guess we could also include MTCRD but I wanted to keep changes minimal for now (although maybe we should _start_ with MTCRD to keep blast radius smaller).

I'm also open to only returning unhealthy if we get a 401 this is mostly to address the "expired token" issue.

rbtr (approved):
stalebot reminded me of this 😟 looks good today, i'm happy to take it

cns/healthserver/healthz_test.go
```diff
@@ -0,0 +1,290 @@
...
+	require.NoError(t, err)
+
+	// set env var to kubeconfig
+	os.Setenv("KUBECONFIG", kubeConfigFile)
+
+	t.Cleanup(func() {
+		server.Close()
+		os.Remove(kubeConfigFile)
+		os.Unsetenv("KUBECONFIG")
+	})
```
timraymond, line 225: Nice use of `testing.(*T).Cleanup`

timraymond (approved):

PR #3270, opened 2024-12-13
backport: Windows HPC base image (#3105)

rbtr:
Windows HPC are special.

They explicitly [do not run in a chroot](https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support#container-mounts:~:text=There%20will%20be%20no%20chroot%20equivalent) or similar execution sandbox, and their [filesystem is a merged view](https://github.com/kubernetes/enhancements/tree/master/keps/sig-windows/1981-windows-privileged-container-support#container-mounts) of the hostfs and their shipped containerfs. This means that we don't need to ship a filesystem: we recently moved from servercore to nanoserver and observed that the lack of pwsh.exe in nanoserver is manageable (in containerd 1.7+, automatically, and in 1.6, by tweaking the container PATH). We can go further and remove the rest of the OS since, as HPC, we don't need it and can escape to the hostfs for dependencies like powershell.

Further, HPC do not have the https://github.com/containerd/containerd/issues/7431 (the "os-version") as regular Windows containers - any Windows HPC may run on any Windows Server Host (given the executables are compatible with that Host).
In some of the testing with WS25 we have seen that this may happen - it was observed that containerd on WS25 inconsistently fetched the WS25 CNI image specified in our multiplatform manifest but was able to run other WS images without any issue.
Containerd >= 1.6 https://github.com/containerd/containerd/pull/8101, so we don't need to build separate Windows images for all of our target WS variants and can simply build one without that extra metadata.


thatmattlong (approved):

jpayne3506 (approved):

jpayne3506 (dismissed):

huntergregory (dismissed):
NPM changes lgtm

PR #3277, opened 2024-12-18
ci: [Service Tags] add public ips with service tags for LBs during cluster creation

k-routhu:
**Reason for Change**:

Adding service tags to all public ips. This change adds public ips and attaches these ips as load balancer outbound ips during cluster creation. Branch `spathak/add-service-tags` is where all the changes were made and tested -- created this branch based off of that one.

**Testing**:
- All testing was done on branch `spathak/add-service-tags` 
- We had a couple of PRs out by Esteban and Shubham to update service tags. However, all of the PRs targeted hack/aks/Makefile, so I went ahead and made all of the updates in this branch.
- Also added ipv6 public ips to load balancer outbound ips for dualstack clusters to resolve errors

Tested changes in a couple of Cilium pipelines:

[Cilium Nightly Pipeline](https://msazure.visualstudio.com/One/_build/results?buildId=120250380&view=results)
The public ips created have service tags attached to them, and the ips fall within the allowed ranges for the locations they were created. E2E tests are failing in this run and in master too, so failures aren't bc of these changes.

[CNI Release Test](https://msazure.visualstudio.com/One/_build/results?buildId=120221298&view=results)
All of the public ips created have service tags attached to them, and the ips fall within the allowed ranges for the locations they were created

[CNI - LSG Integration Test](https://msazure.visualstudio.com/One/_build/results?buildId=120088946&view=results)
The clusters get created here successfully with the correct public ips. The integration tests do fail, but I checked with John Payne -- he mentioned that the integration tests fail frequently because they're heavily dependent on a partner team (Canonical). He suggested running Cilium Private Test pipelines on v1.14 and v1.16 (the next two on the list) and the remaining pipelines with the service tags. The remaining pipelines look good and create public ips as expected, so I believe we should be good to go on that front.

[ACN PR Pipeline](https://msazure.visualstudio.com/One/_build/results?buildId=120225260&view=results)
All of the public ips created have service tags attached to them, and the ips fall within the allowed ranges for the locations they were created


hack/aks/Makefile
```diff
@@ -8,25 +8,28 @@ AZIMG   = mcr.microsoft.com/azure-cli
...
+OS				?= linux # Used to signify if you want to bring up a windows nodePool on byocni clusters
+OS_SKU          ?= Ubuntu
+OS_SKU_WIN      ?= Windows2022
+REGION          ?= westus2
+VM_SIZE	        ?= Standard_B2s
+VM_SIZE_WIN     ?= Standard_B2s
+IP_TAG          ?= FirstPartyUsage=/DelegatedNetworkControllerTest
+IP_PREFIX       ?= serviceTaggedIp
+PUBLIC_IPv4       ?= $(IP_PREFIX)-$(CLUSTER)-v4
+PUBLIC_IPv6       ?= $(IP_PREFIX)-$(CLUSTER)-v6
```
jpayne3506, line 25: Formatting

hack/aks/Makefile
```diff
@@ -43,6 +46,26 @@ azcfg: ## Set the $AZCLI to use aks-preview
...
+
+public-ipv6: rg-up
+	$(AZCLI) network public-ip create --name $(PUBLIC_IPv6) \
+		--resource-group $(GROUP) \
+		--allocation-method Static \
+		--ip-tags $(IP_TAG) \
+		--location $(REGION) \
+		--sku Standard \
+		--tier Regional \
+		--version IPv6
```
jpayne3506, line 67: Duplicate `rg-up` calls when creating cluster. I would remove the `rg-up` from the make calls `public-ipv4` and `public-ipv6`

hack/aks/Makefile
```diff
@@ -185,13 +213,14 @@ overlay-up: rg-up overlay-net-up ## Brings up an Overlay AzCNI cluster
...
 
 swift-byocni-up: rg-up swift-net-up ## Bring up a SWIFT BYO CNI cluster
+	@$(MAKE) public-ipv4
 	$(AZCLI) aks create -n $(CLUSTER) -g $(GROUP) -l $(REGION) \
 		--auto-upgrade-channel $(AUTOUPGRADE) \
 		--node-os-upgrade-channel $(NODEUPGRADE) \
 		--kubernetes-version $(K8S_VER) \
 		--node-count $(NODE_COUNT) \
 		--node-vm-size $(VM_SIZE) \
-		--load-balancer-sku standard \
```
jpayne3506, line 194: Inconsistent removal of `--load-balancer-sku`. Does it need to be `standard` or can we keep it as `basic`?

jpayne3506 (changes requested):

hack/aks/Makefile
```diff
@@ -98,6 +120,7 @@ up: swift-up ## Alias to swift-up
 
 
 nodesubnet-byocni-nokubeproxy-up: rg-up overlay-net-up ## Brings up an NodeSubnet BYO CNI cluster without kube-proxy
+	@$(MAKE) public-ipv4
```
rohitams, line 123: will this rerun the public ip creation? if so, should we have the ip creation as a global step so it provisions just once?

hack/aks/Makefile
```diff
@@ -98,6 +120,7 @@ up: swift-up ## Alias to swift-up
 
 
 nodesubnet-byocni-nokubeproxy-up: rg-up overlay-net-up ## Brings up an NodeSubnet BYO CNI cluster without kube-proxy
```
thomasricci, line 122: Are these steps run in parallel or in a sequential order?
Public IPs from ST are limited so we should try to either reuse or not create too many public ips at the same time

hack/aks/Makefile
thomasricci, line 67: Did we request IPV6 ips for our test ST ?

hack/aks/Makefile
k-routhu: can be basic, I meant to remove `--load-balancer-sku standard` for `nodesubnet-byocni-nokubeproxy-up` since I'm adding `--load-balancer-outbound-ips <publicip v4>,<publicip v6>`

hack/aks/Makefile
k-routhu: yes, this reruns public ip creation, but since the Cilium pipelines do cluster creation in parallel, we need to provision different public ips

hack/aks/Makefile
k-routhu: these steps are run in parallel, but I do believe we have the capacity to create the public ips. The majority of the clusters are created in westus and westus2, and they have 254 and 512 available vips respectively, and we create about 10-12 clusters (and public ips) in each location, and delete them all at the end

hack/aks/Makefile
k-routhu: updated this

k-routhu: discussed offline, should be good to go on ipv6

hack/aks/Makefile
```diff
@@ -8,25 +8,28 @@ AZIMG   = mcr.microsoft.com/azure-cli
...
+PUBLIC_IPv4     ?= $(IP_PREFIX)-$(CLUSTER)-v4
+PUBLIC_IPv6     ?= $(IP_PREFIX)-$(CLUSTER)-v6
 KUBE_PROXY_JSON_PATH ?= ./kube-proxy.json
 
 # overrideable variables
 SUB        ?= $(AZURE_SUBSCRIPTION)
 CLUSTER    ?= $(USER)-$(REGION)
 GROUP      ?= $(CLUSTER)
 VNET       ?= $(CLUSTER)
-
```
jpayne3506, line 29: nit: Lets keep this line in.

hack/aks/Makefile
jpayne3506: All `?=` should be lined up.

hack/aks/Makefile
jpayne3506: Does this implicitly set every cluster to standard for the load balancer sku?

hack/aks/Makefile
```diff
@@ -43,6 +46,26 @@ azcfg: ## Set the $AZCLI to use aks-preview
...
+
+public-ipv6:
+	$(AZCLI) network public-ip create --name $(PUBLIC_IPv6) \
+		--resource-group $(GROUP) \
+		--allocation-method Static \
+		--ip-tags $(IP_TAG) \
+		--location $(REGION) \
+		--sku Standard \
+		--tier Regional \
+		--version IPv6
```
rbtr, line 67: are these identical except for "v4" and "v6"?
```suggestion
ip:
	$(AZCLI) network public-ip create --name $(IP_PREFIX)-$(CLUSTER)-$(IPVERSION) \
		--resource-group $(GROUP) \
		--allocation-method Static \
		--ip-tags $(IP_TAG) \
		--location $(REGION) \
		--sku Standard \
		--tier Regional \
		--version IP$(IPVERSION)

ipv4:
	@$(MAKE) ip IPVERSION=v4

ipv6:
	@$(MAKE) ip IPVERSION=v6

```

hack/aks/Makefile
```diff
@@ -8,25 +8,28 @@ AZIMG   = mcr.microsoft.com/azure-cli
...
+NODE_COUNT_WIN	?= $(NODE_COUNT)
+NODEUPGRADE     ?= NodeImage
+OS				?= linux # Used to signify if you want to bring up a windows nodePool on byocni clusters
+OS_SKU          ?= Ubuntu
+OS_SKU_WIN      ?= Windows2022
+REGION          ?= westus2
+VM_SIZE	        ?= Standard_B2s
+VM_SIZE_WIN     ?= Standard_B2s
+IP_TAG          ?= FirstPartyUsage=/DelegatedNetworkControllerTest
+IP_PREFIX       ?= serviceTaggedIp
```
rbtr, line 23: are these strings pre-determined?

hack/aks/Makefile
rbtr: why are you reinvoking Make for this? you could make the IP target a dependency here, like `rg-up` etc.

rbtr (changes requested):

timraymond:
Why are we using `make` for this? Can't most of this be done more predictably by Bicep/ARM?

rbtr:
> Why are we using `make` for this? Can't most of this be done more predictably by Bicep/ARM?

@timraymond yes and no. 
AzCLI is basically a wrapper to ARM when used like this and that's what it actually sends, so in a literal sense, yes. Bicep would be even better for declarative infrastructure.

I originally wrote this Makefile because I needed to build infra that required certain subscription flags enabled (which ARM may be able to do, but was documented as AzCLI commands), and a certain preview CLI installed locally and/or a preview AZ AKS plugin installed locally.
And, more importantly to me at the time, AzCLI (including ARM, Bicep, and anything that depends on it) was and is still completely busted on Fedora. Platforming everything in the azure-cli container makes this totally portable.

I also find it nice to have a central list of capabilities. We could write Bicep for everything here, but I think I would still want the Makefile to exist as the single entrypoint to all of the `bicep apply <scenario>` so that it's all easy to find (and to wrap in the azcli image so that IT WORKS AT ALL 😤)

hack/aks/Makefile
k-routhu: `IP_TAG` is predetermined, but `IP_PREFIX` is not -- I believe that's just a variable Shubham had created for naming the public ips we create and tag for this workflow

k-routhu:
re-opening PR

hack/aks/Makefile
```diff
@@ -8,17 +8,22 @@ AZIMG   = mcr.microsoft.com/azure-cli
...
+PUBLIC_IP_ID	?= /subscriptions/$(SUB)/resourceGroups/$(GROUP)/providers/Microsoft.Network/publicIPAddresses
+PUBLIC_IPv4     ?= $(PUBLIC_IP_ID)/$(IP_PREFIX)-$(CLUSTER)-v4
+PUBLIC_IPv6     ?= $(PUBLIC_IP_ID)/$(IP_PREFIX)-$(CLUSTER)-v6
 KUBE_PROXY_JSON_PATH ?= ./kube-proxy.json
```
jpayne3506, line 27: Line up `?=`

hack/aks/Makefile
```diff
@@ -72,6 +94,22 @@ vars: ## Show the input vars configured for the cluster commands
...
+		--sku Standard \
+		--tier Regional \
+		--version IP$(IPVERSION)
+
+ipv4:
+	@$(MAKE) ip IPVERSION=v4
+
+ipv6:
+	@$(MAKE) ip IPVERSION=v6
+
```
jpayne3506, line 112: duplicate

hack/aks/Makefile
```diff
@@ -43,6 +48,23 @@ azcfg: ## Set the $AZCLI to use aks-preview
...
+		--location $(REGION) \
+		--sku Standard \
+		--tier Regional \
+		--version IP$(IPVERSION)
+
+ipv4:
+	@$(MAKE) ip IPVERSION=v4
+
+ipv6:
+	@$(MAKE) ip IPVERSION=v6
```
jpayne3506, line 65: nit: this looks excessive. We could possibly leverage https://www.gnu.org/software/make/manual/make.html#Target_002dspecific 
```suggestion
ipv4: IPVERSION = v4
ipv4: ip

ipv6: IPVERSION = v6
ipv6: ip
```

jpayne3506: @rbtr, opinions?

hack/aks/Makefile
jpayne3506: It is a cogs bump.

jpayne3506 (changes requested):

jpayne3506 (approved):
Lgtm.

rbtr (approved):

k-routhu:
reopening PR

jpayne3506 (approved):

rbtr (approved):

PR #3279, opened 2024-12-18
fix: let NNCResonciler process update events with different generations when IPAMv2 is enabled

### LINKED ISSUE #3278: CNS can miss NNC updates when its watch is down
**What happened:** 

CNS filters out NNC update events if a spec and status change are made while a watch is down.

CNS should process added IPs to an NNC when it re-establishes its watch even if the spec has changed since the last time it saw it.

**How to reproduce it:**  

Difficult to repro with CNS and NNCs. Better repro is to show that if you kill a watch on a controller and update a resource's spec and status, the controller will see 1 Update after it re-establishes the watch and the old and new object generations will be different.

tyler-lloyd:
**Reason for Change**:

IPAMv2 can handle update events even when the update is made from CNS itself, like patching
the spec for more IPs. This solves the problem of missed updates when a watch is re-established
and an NNC has been updated in both its spec and status. IPAMv2 is idempotent and will not
fall into a feedback loop where after CNS patches the NNC for more IPs its NNC reconciler still
sees IPs needed so it requests more IPs again, which triggers the reconciler again to request
more IPs, and so on ...

**Issue Fixed**:

Fixes #3278 

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

The first commit is the basic implementation. The 2nd commit does some more refactoring. I am fine to take just the first commit if you all don't like the refactoring attempt.

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/reconciler.go
+++ b/cns/kubecontroller/nodenetworkconfig/reconciler.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 
 	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/configuration"
 	"github.com/Azure/azure-container-networking/cns/logger"
 	"github.com/Azure/azure-container-networking/cns/restserver"
 	cnstypes "github.com/Azure/azure-container-networking/cns/types"
@@ -157,7 +158,7 @@ func (r *Reconciler) Started(ctx context.Context) (bool, error) {
 }
 
 // SetupWithManager Sets up the reconciler with a new manager, filtering using NodeNetworkConfigFilter on nodeName.
-func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, node *v1.Node) error {
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, node *v1.Node, cnsconfig *configuration.CNSConfig) error {
 	r.nnccli = nodenetworkconfig.NewClient(mgr.GetClient())
 	err := ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha.NodeNetworkConfig{}).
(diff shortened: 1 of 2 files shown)


cns/service/main.go
```diff
@@ -1482,7 +1482,7 @@ func InitializeCRDState(ctx context.Context, httpRestService cns.HTTPService, cn
 		// wait for the Reconciler to run once on a NNC that was made for this Node.
 		// the nncReadyCtx has a timeout of 15 minutes, after which we will consider
 		// this false and the NNC Reconciler stuck/failed, log and retry.
-		nncReadyCtx, cancel := context.WithTimeout(ctx, 15*time.Minute) //nolint // it will time out and not leak
+		nncReadyCtx, cancel := context.WithTimeout(ctx, 15*time.Minute) // nolint // it will time out and not leak
```
timraymond, line 1485: I believe the nolint pragmas are intended to be directly next to the `//`

```suggestion
		nncReadyCtx, cancel := context.WithTimeout(ctx, 15*time.Minute) //nolint // it will time out and not leak
```

timraymond (approved):
This seems reasonable to me

cns/service/main.go
tyler-lloyd: here's an arbitrary webpage to say that you're right. wonder why my auto formatter is changing these lines. https://golangci-lint.run/usage/false-positives/#:~:text=tests%20for%20it.-,Use,because,-machine%2Dreadable%20comments

will revert these.

tyler-lloyd: except I wonder if `//nolint` by itself does anything and you really need to do `//nolint:all` so my formatter thinks these are just normal comments ... ?

timraymond: `//nolint` by itself probably does nothing... I have always seen it needing a specific linter. Also, there's another linter that complains if you don't have a reason attached.

cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -157,29 +158,32 @@ func (r *Reconciler) Started(ctx context.Context) (bool, error) {
 }
 
 // SetupWithManager Sets up the reconciler with a new manager, filtering using NodeNetworkConfigFilter on nodeName.
-func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, node *v1.Node) error {
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager, node *v1.Node, cnsconfig *configuration.CNSConfig) error {
```
rbtr, line 161: instead of the whole cnsconfig, can we pass in a `checkGeneration bool` or something like that? don't want the nnc reconciler depending on the cns/config

rbtr (changes requested):

rbtr (approved):
lgtm let's see if it blows up 🧪

PR #3286, opened 2024-12-20
[NPM Lite] Default Deny CNS Changes

rejain456:
**Reason for Change**:

As part of adding default deny so pods can't communicate with one another when network policies are not present, this pr is part 2 which updates the cns code, creating default deny acl and sending it to cni.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/middlewares/k8sSwiftV2.go
```diff
@@ -102,19 +117,21 @@ func (k *K8sSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(defaultHandler, fa
 
 // validateIPConfigsRequest validates if pod is multitenant by checking the pod labels, used in SWIFT V2 AKS scenario.
 // nolint
-func (k *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req *cns.IPConfigsRequest) (podInfo cns.PodInfo, respCode types.ResponseCode, message string) {
+func (k *K8sSWIFTv2Middleware) validateIPConfigsRequest(ctx context.Context, req *cns.IPConfigsRequest) (podInfo cns.PodInfo, respCode types.ResponseCode, message string, defaultDenyACL bool) {
```
timraymond, line 120: I'm not excited about adding a fourth return value here, and I generally don't like to return bools (since they carry no documentation about their meaning).

I would almost be of a mind to do the `k.Cli.Get(ctx, mtpncNamespacedName, &mtpnc)` to find out the status of the defaultDenyACL one level up the call stack where we want to use that value, but I understand that incurs a penalty for the second network call to the api server.

The only alternate proposal I have is caching these values on `*K8sSWIFTv2Middleware`, but all the usual hazards of caching apply here. For something applying network policy, I'm leery of doing this.

I suppose this is okay for now unless we can think of an alternative that avoids it.

cns/middlewares/k8sSwiftV2.go
```diff
@@ -55,6 +57,19 @@ func (k *K8sSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(defaultHandler, fa
 		if !req.SecondaryInterfacesExist {
 			return ipConfigsResp, err
 		}
+
+		// ipConfigsResp has infra IP configs -> if defaultDenyACLbool is enabled, add the default deny acl's pn the infra IP configs
+		for i := range ipConfigsResp.PodIPInfo {
+			ipInfo := &ipConfigsResp.PodIPInfo[i]
```
timraymond, line 63: Any reason not to do:

```suggestion
		for _, ipInfo := range ipConfigsResp.PodIPInfo {

```

cns/middlewares/k8sSwiftV2.go
```diff
@@ -55,6 +57,19 @@ func (k *K8sSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(defaultHandler, fa
...
+			// there will be no pod connectivity to and from those pods
+			if defaultDenyACLbool && ipInfo.NICType == cns.InfraNIC {
+				err = addDefaultDenyACL(ipInfo)
+				if err != nil {
+					logger.Errorf("failed to add default deny acl's for pod %v with err %v", podInfo.Name(), err)
```
timraymond, line 68: This will continue here... is that intentional?

cns/middlewares/k8sSwiftV2.go
rejain456: Hi, there are 2 reasons why I did not go that route

-  I see the following error: rangeValCopy: each iteration copies 256 bytes (consider pointers or indexing) (gocritic)

 image

- This line of code:  **_, ipInfo := range ipConfigsResp.PodIPInfo** creates a copy of ipInfo and &ipInfo creates a pointer to the copy; however, does not change the slice itself. 

Please feel free to correct me on this, but that was my thinking.

timraymond: It feels like premature optimization to me. I'd prefer the more readable suggestion unless we can point at some benchmarks / profiling that proves this large copy is a performance bottleneck.

cns/middlewares/k8sSwiftV2.go
rbtr: I think this method has become too overloaded. `validate-` means tell me if this input is valid and should _only_ return a bool.

cns/middlewares/k8sSwiftV2_linux.go
```diff
@@ -103,3 +103,7 @@ func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(_ *cns.PodIpInfo,
 }
 
 func (k *K8sSWIFTv2Middleware) addDefaultRoute(*cns.PodIpInfo, string) {}
+
+func addDefaultDenyACL(_ *cns.PodIpInfo) error {
```
rbtr, line 107: we shouldn't call this from common code if it requires a stub in an _os file. we should be able to abstract in such a way that we don't do this.

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -58,3 +62,58 @@ func (k *K8sSWIFTv2Middleware) addDefaultRoute(podIPInfo *cns.PodIpInfo, gwIP st
...
+
+	podIPInfo.DefaultDenyACL = append(podIPInfo.DefaultDenyACL, additionalArgs...)
+
+	return nil
+}
+
+// create the default deny acl's that need to be added to the list defaultDenyACL field in podIpInfo
+func getDefaultDenyACLPolicy(direction hcn.DirectionType) ([]byte, error) {
+	const DefaultDenyPriority = 10000
+	const policyType = "ACL"
```
rbtr, line 97: these function-scoped const definitions don't make sense to me. if this is only used in one place, hardcode with a comment.

rbtr (changes requested):

cns/middlewares/k8sSwiftV2_windows.go
timraymond: @rejain456 is this from the magic number detector linter?

rejain456: @timraymond , what exactly is the magic number detector linter? 10000 priority was used as its lower priority than npm block default priority of 3000, so if an npm policy was applied than it would overwrite the default deny among the pods

timraymond: Sorry, I was speculating that you might have needed to constantize `10000` to satisfy the linter. That linter (gomnd) complains if it sees a number used as an argument without any explanation about what it is. Usually, the answer is to make it a constant so it then has a name associated with it.

But yeah, I agree with @rbtr , I'd just inline it and drop a comment about what it is.

Also, you can do `10_000` to make it a little more visually obvious.

rejain456: got it thank you for the comment, I will inline it and add a comment on what it means. Regarding the priority, the value should be a number so it would be great if you can explain a little more on where the 10_000 would go?

cns/middlewares/k8sSwiftV2.go
rejain456: @timraymond, can you please elaborate on which readable suggestion you are referring to?

timraymond: Sorry @rejain456 , when I refer to "the suggestion" here, I mean the code suggestion in the top comment here: https://github.com/Azure/azure-container-networking/pull/3286#discussion_r1896095353

rejain456: @timraymond , got it thanks for clarifying. Regarding that I believe this line of code: `_, ipInfo := range ipConfigsResp.PodIPInfo` creates a copy of ipInfo and &ipInfo creates a pointer to the copy; however, does not change the slice itself. In that case, the property PodIPInfo won't get changed in  ipConfigsResp.PodIPInfo. Whereas the line: 
`for i := range ipConfigsResp.PodIPInfo { ipInfo := &ipConfigsResp.PodIPInfo[i]` creates a pointer to the original PodIPInfo list and will change the properties in it. 

Please let me know if any part of that is incorrect

cns/NetworkContainerContract.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/Azure/azure-container-networking/cni"
```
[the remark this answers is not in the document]
rejain456: @rbtr, thanks for pointing it out. I instead add the KVPair stuct in a common folder so that both cns and cni can use the struct and eliminated the circular dependency. Please let me know if this new approach works! Thanks

cns/middlewares/k8sSwiftV2_windows.go
rejain456: Thanks @timraymond  and @rbtr  for suggesting this. I updated the pr to make the const definitions, added a comment, and converted 10000 to 10_000 to make it more visually obvious.

cns/middlewares/k8sSwiftV2_linux.go
thatmattlong: I'm actually ok with this, I feel it's better to have an os stub than to branch in the common code based on `runtime.GOOS`

rbtr: I don't want us to check `runtime.GOOS` either. But we can do a better job using the `_<os>.go` pattern...I feel like I have done this several times recently, but I can't find a good example immediately.

At k8sSwiftV2_linux.go L105 right above here, there's a stub for `addDefaultRoute` which noops. If we only need to set a default route on Windows, putting `addDefaultRoute` in the OS abstraction is the wrong place - what are we doing when we need to addDefaultRoute? That's the thing that actually has different behavior per-OS, and so that's the thing that should be in the OS specific files. 

If we need these stubs, the abstraction is wrong...or actually doesn't exist. We're not abstracting away the OSes, we're actually coding them all together and branching _implicitly_ based on the GOOS. The `if` just moves out of the code and in to my head, because now _I_ need to know that if GOOS=linux then this method call is noop when I'm reading the common calling code.

thatmattlong: > That's the thing that actually has different behavior per-OS, and so that's the thing that should be in the OS specific files.

While this is probably true, I don't want this PR to become too much about refactoring that. There's *a lot* of stuff in the calling code which we'd have to duplicate and/or refactor which I would rather not do now, and I feel this doesn't put us in much worse a situation for when (if) we do take up that work.

timraymond: idk, if there's only one OS that needs special handling, and the rest are all NOPs, I find `runtime.GOOS` more readable. It works better with static analysis too, since you don't need to set GOOS for the toolchain to realize there is an entire separate implementation.

Now if we had wholly different implementations, each with their own considerations, I'd consider build tags more necessary.

rbtr: That works if the special handling is entirely your code and doesn't work if it's platform specific things like using registry package which is compiled for and can only be referenced from code compiled for Windows. We do a lot of this - Linux iptables, netns, etc vs Windows HCS, registry, etc

timraymond: Fair enough--if there's compile-time reasons, that's certainly a good reason. I guess lately I've seen a few instances where it's just `exec.Cmd`s for either `powershell.exe` or their Linux equivalent. In those circumstances, the specific string can be decided at runtime.

cns/middlewares/k8sSwiftV2.go
rejain456: @timraymond  @rbtr, changed the name of the function to accurately represent what it is doing

cns/middlewares/k8sSwiftV2_linux.go
rejain456: resolving this after talked to Matthew about it

cns/middlewares/k8sSwiftV2.go
```diff
@@ -55,6 +57,20 @@ func (k *K8sSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(defaultHandler, fa
 		if !req.SecondaryInterfacesExist {
 			return ipConfigsResp, err
 		}
+
+		// ipConfigsResp has infra IP configs -> if defaultDenyACLbool is enabled, add the default deny acl's pn the infra IP configs
+		for i := range ipConfigsResp.PodIPInfo {
+			ipInfo := &ipConfigsResp.PodIPInfo[i]
+			// there will be no pod connectivity to and from those pods
+			if defaultDenyACLbool && ipInfo.NICType == cns.InfraNIC {
```
tamilmani1989, line 65: why cns need to check infranic type? dnc-rc creates mtpnc and it will make sure it set only for infranic right? cns doesn't throw error here either if nic type is not infranic and defaultdeny is set.

rejain456: currently the defaultDenyACL bool field is under MTPNC spec field. I don't think DNC-RC sets it only for infra nic, it currently takes the value from pni and copies it to the spec field in MTPNC. CNS then checks if the bool is true and if it is, then it adds the default deny endpoint policies for the infra nic. @tamilmani1989 please let me know if you think we can go about a different way

cns/middlewares/k8sSwiftV2.go
```diff
@@ -100,21 +116,23 @@ func (k *K8sSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(defaultHandler, fa
...
-		return nil, types.UnexpectedError, errBuf.Error()
+		return nil, types.UnexpectedError, errBuf.Error(), defaultDenyACLbool
 	}
 	logger.Printf("[SWIFTv2Middleware] validate ipconfigs request for pod %s", podInfo.Name())
 	podNamespacedName := k8stypes.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
 	pod := v1.Pod{}
 	if err := k.Cli.Get(ctx, podNamespacedName, &pod); err != nil {
 		errBuf := errors.Wrapf(err, "failed to get pod %+v", podNamespacedName)
-		return nil, types.UnexpectedError, errBuf.Error()
+		return nil, types.UnexpectedError, errBuf.Error(), defaultDenyACLbool
```
rbtr, line 135: any error must be the last return. also, 4 returns seems too many to me

cns/middlewares/k8sSwiftV2_linux.go
rbtr: I don't consider this resolved. 
Aside from this stub function, this change has added a 4th return to _everywhere_ that GetPodInfoForIPConfigsRequest is called but which is only set, used, or tested in Windows. This is just unnecessary churn, if the OS specific features were encapped properly this would be a much smaller and simpler change.

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -58,3 +62,55 @@ func (k *K8sSWIFTv2Middleware) addDefaultRoute(podIPInfo *cns.PodIpInfo, gwIP st
 	}
 	podIPInfo.Routes = append(podIPInfo.Routes, route)
 }
+
+// append the default deny acl's to the list defaultDenyACL field in podIpInfo
+func addDefaultDenyACL(podIPInfo *cns.PodIpInfo) error {
+	blockEgressACL, err := getDefaultDenyACLPolicy(hcn.DirectionTypeOut)
+	if err != nil {
+		return errors.Wrap(err, "Failed to create default deny ACL policy egress")
```
rbtr, line 70: error strings are not sentences, they're just fragments that can be embedded and strung together - don't capitalize 
```suggestion
		return errors.Wrap(err, "failed to create default deny ACL policy egress")
```

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -58,3 +62,55 @@ func (k *K8sSWIFTv2Middleware) addDefaultRoute(podIPInfo *cns.PodIpInfo, gwIP st
 	}
 	podIPInfo.Routes = append(podIPInfo.Routes, route)
 }
+
+// append the default deny acl's to the list defaultDenyACL field in podIpInfo
+func addDefaultDenyACL(podIPInfo *cns.PodIpInfo) error {
```
rbtr, line 67: I find this - passing a reference and mutating it - a bad pattern. It's impossible to tell from the calling code that the `podIPInfo` is modified. Even if I suspect that, it's impossible to tell how it changes. 

Sometimes this may be okay if this is func does something very complex which depends on the internal state of the passed object. But when I look at the contents of this function...it doesn't even need the podIPInfo. This could be a more flexible and testable stateless pure function if it was refactored to return the `[]policy.Policy` instead of accepting this struct and appending the policy internally.

As it is, this has side-effects and side-effects are bad™️.

Also it's worth mentioning that Go is not C, and passing refs is not always more performant than passing values, because we have to track them for GC.

Also also, this as-written has a nil-pointer exception because there is no check that the podIPInfo input is valid before trying to assign to its `EndpointPolicies`. 

@timraymond may have additional things to say here. This is directly related to the `for range` iteration that he thought seemed off...but maybe didn't identify this as the source of the smell.

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -58,3 +62,55 @@ func (k *K8sSWIFTv2Middleware) addDefaultRoute(podIPInfo *cns.PodIpInfo, gwIP st
...
+		Priority:  10_000, // default deny priority will be 10_000
+	}
+
+	denyACLJSON, err := json.Marshal(denyACL)
+	if err != nil {
+		return nil, errors.Wrap(err, "error marshalling default deny policy to json")
+	}
+
+	return denyACLJSON, nil
+}
```
rbtr, line 116: These read as static to me. Do we need to dynamically generate them for every request? They have no variable inputs.

rbtr (changes requested):

cns/middlewares/k8sSwiftV2_windows.go
rejain456: Hey Evan, we do not dynamically generate them for every request. For every request that calls this function, the same block in and block out acl's are provided; thus, the static behavior. Please let me know if you have any other questions regarding that

cns/middlewares/k8sSwiftV2.go
rejain456: Moved errors to the last spot, thanks

cns/middlewares/k8sSwiftV2_windows.go
rejain456: lowercased it thanks

cns/middlewares/k8sSwiftV2_windows.go
rejain456: I modified the code so the addDefaultDenyACL function returns policy.Policy and outside of the function, this policy.Policy is added to the endpoints field in PodInfo. @rbtr  please let me know if that looks better

cns/middlewares/k8sSwiftV2_windows.go
rbtr: >For every request that calls this function


Yes, this is what I'm saying. Why do we need to recreate these ACLs by calling this function on every request? They don't change.

rejain456: hey @rbtr, after looking more into it, I realized instead of making this function static, I can make it more extensible so in the future we other types of policies need to be added, it can be added using this same function. I made the function more dynamic + changed the name of it. Please let me know if that looks better!

cns/middlewares/k8sSwiftV2_linux.go
timraymond: I agree with @rbtr , and after staring at this for an hour, I think the problem is that `GetPodInfoForIPConfigsRequest` is trying to do way too much. We're not helping it by threading an additional responsibility of producing this boolean. That's indicative that there's some feature envy of this method though... it has some data that the callers want that can't be obtained any other way. You can also see this with manufacturing a `cns.PodInfo` with a bunch of facts, throwing errors if certain labels aren't set, etc. We might as well rename the method `DoK8SThings` for how specific it is (read: this is all pre-existing, so please don't take this as something you did @rejain456 ). Also, this method flattens errors to strings, which is no good either.

I think a better structure for this is roughly in two methods:
1. Get the data we need from k8s
2. Validate that it's in the right form

This should make it easier to test as well. 1) can be done with an integration test. 2) can be exhaustively unit tested. Once we have the data, and verified that everything we expect is there, then we can just directly access the `MultitenantPodNetworkConfig`'s `DefaultDenyACL` (since we trust that 2) did its job).

However, I recognize that time constraints exist. I think a workable solution for now could be to pack `defaultDenyACLbool` into `cns.PodInfo` (since it seems to be a grab-bag of facts anyway), which would eliminate the extra return.

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -58,3 +61,42 @@ func (k *K8sSWIFTv2Middleware) addDefaultRoute(podIPInfo *cns.PodIpInfo, gwIP st
...
+		Direction string `json:"Direction"`
+		Priority  int    `json:"Priority"`
+	}
+
+	endpointPolicy := EndpointPolicy{
+		Type:      policyType,
+		Action:    action,
+		Direction: direction,
+		Priority:  priority,
+	}
```
timraymond, line 94: This is totally fine for throwaway structs (definitely for JSON, and especially if you plan to use `EndpointPolicy` more than once), but there's an even more disposable form you can use if you like:

```go
endpointPolicy := struct{
    Type string
    Action string
    Direction string
    Priority int
}{
    Type: policyType,
    Action: action,
    Direction: direction,
    Priority: priority
}
```

I omitted the struct tags for `json` in this one since I don't have the convenience of an editor to add them, but I like that you included them here even though they will be the same as the defaults. I think it's nice to add them to say "I intend for this struct to be JSON at some point".

cns/middlewares/k8sSwiftV2_linux.go
rejain456: @timraymond , the cns.PodInfo passed back is an interface not the podIpInfo Struct. In this case, is your apporach to add DefaultDenyBool() as a function in the interface?

cns/middlewares/k8sSwiftV2_windows.go
rejain456: Thank you for this suggestion, I implemented it this way

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -58,3 +65,155 @@ func (k *K8sSWIFTv2Middleware) addDefaultRoute(podIPInfo *cns.PodIpInfo, gwIP st
...
+			return &cns.IPConfigsResponse{
+				Response: cns.Response{
+					ReturnCode: respCode,
+					Message:    message,
+				},
+			}, errors.New("failed to validate IP configs request")
+		}
+
+		// checks for default deny bool variable
+		defaultDenyACLbool, err := GetDefaultDenyBool(mtpnc)
```
QxBytes, line 139: could we change the comment for this function to better document what it does? Like what it takes in, what it modifies, what it returns, what it calls into.

cns/middlewares/k8sSwiftV2_linux.go
rejain456: @timraymond , @rbtr , thank you for the suggestions. I efactored the GetPodInfoForIPConfigsRequest() function to make it clearer and seperated the IPConfigsRequestHandlerWrapper function into os specific files which got rid of the stub in the linux file. Please let me know if this is the best approach to take?

cns/middlewares/k8sSwiftV2_windows.go
rejain456: Updated the comments, thanks!

rbtr (approved):
latest changes address all my comments, lgtm, let's make sure it passes all swift v2 tests

timraymond (approved):
LGTM 👍

rejain456:
Confirmed with @chandanAggarwal, to merge this pr in

PR #3287, opened 2024-12-20
[NPM Lite] Support Network Policies Through CNS

rejain456:
**Reason for Change**:

As part of adding default deny so pods can't communicate with one another when network policies are not present, this pr is part 3  which updates the cni code, retrieving the default deny acl from cns and creating HNS endpoints of the ACL's for the pods

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cni/network/network.go
```diff
@@ -590,6 +590,11 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 	}
 
 	policies := cni.GetPoliciesFromNwCfg(nwCfg.AdditionalArgs)
+	for key := range ipamAddResult.interfaceInfo {
+		if key == string(cns.InfraNIC) {
```
tamilmani1989, line 594: why do we need infranic check? policy will be empty for other nic right?

cni/network/invoker_cns_test.go
```diff
@@ -773,12 +806,24 @@ func TestCNSIPAMInvoker_Add(t *testing.T) {
...
+								{
+									EndpointPolicies: expectedEndpointPolicies,
+								},
+							},
+							Response: cns.Response{
+								ReturnCode: 0,
+								Message:    "",
+							},
+						},
+						err: errors.New("failed error from CNS"), //nolint "error for ut"
```
paulyufan2, line 820: can we add details on what's error from CNS?

cni/network/network.go
```diff
@@ -590,6 +590,8 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 	}
 
 	policies := cni.GetPoliciesFromNwCfg(nwCfg.AdditionalArgs)
+	policies = append(policies, ipamAddResult.interfaceInfo[string(cns.InfraNIC)].EndpointPolicies...)
```
paulyufan2: can we add UTs in network_windows_test.go to make sure cni network gets correct policies

paulyufan2:
can we add comment to the description that this change is only applied on windows

cni/network/network.go
```diff
@@ -590,6 +590,9 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 	}
 
 	policies := cni.GetPoliciesFromNwCfg(nwCfg.AdditionalArgs)
+	for key := range ipamAddResult.interfaceInfo {
```
paulyufan2, line 593: if the change is only applied to windows, can you please add policies to os specific file: network_windows.go

behzad-mir: second this. If this is windows specific then it's better to be done in endpoint_windows

tamilmani1989 (dismissed):

cni/network/network.go
tamilmani1989: good point but there is nothing windows specific in this code block. i would rather prefer to keep here and can extend to linux if required.

tamilmani1989: its just setting up interfaceinfo

paulyufan2: make sense

paulyufan2 (approved):

cni/network/network.go
tamilmani1989: @rejain456 we can move this codeblock where cni already iterate `ipamAddResult.interfaceInfo` here: https://github.com/Azure/azure-container-networking/blob/1296fd78d96a4a81f3836d0fb5dd61ea4cd21a85/cni/network/network.go#L613
so we can avoid one more loop

rejain456: That makes sense Tamilmani, thanks for the suggestion Pual and Behzad

cni/network/network.go
rejain456: removed the check! Thank you

QxBytes (commented):
Could you also add a test in cni/network/network_windows_test.go > TestPluginWindowsAdd (modify either test) such that the returned cns response includes a different endpoint policy to be returned for each response and then confirm that it propagates to the network and endpoint policies fields in the respective wanted EndpointInfo structs properly?

cni/network/network.go
```diff
@@ -616,6 +616,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 
 		natInfo := getNATInfo(nwCfg, options[network.SNATIPKey], enableSnatForDNS)
 		networkID, _ := plugin.getNetworkID(args.Netns, &ifInfo, nwCfg)
+		policies = append(policies, ipamAddResult.interfaceInfo[key].EndpointPolicies...)
```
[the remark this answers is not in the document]
rejain456: Thanks a lot for the suggestion, updated the code

tamilmani1989 (approved):

QxBytes (dismissed):

QxBytes (approved):

rejain456:
Confirmed with @chandanAggarwal, to merge this pr in

thatmattlong (approved):

PR #3297, opened 2025-01-02
fix: fixing HNS network already exist error in Stateless CNI

behzad-mir:
**Reason for Change**:
This PR fix an issue in stateless CNI where the "HNS network already exists" error was not being handled correctly.
Since Stateless CNI add calls work in parallel, it is possible that two add process make API calls to create HNS netwrok at the same time and one of them fails since the other one successfully created the network (HNS process calls sequentially). In that case the second process should not exit in error and instead fetch the HNS network that was created by the first process and continue with the add call.

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


network/network_windows.go
```diff
@@ -360,10 +360,20 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *EndpointInfo, extIf *exter
 		if errors.As(err, &hcn.NetworkNotFoundError{}) {
 			logger.Info("Creating hcn network", zap.Any("hcnNetwork", hcnNetwork))
 			hnsResponse, err = Hnsv2.CreateNetwork(hcnNetwork)
-			if err != nil {
-				return nil, fmt.Errorf("Failed to create hcn network: %s due to error: %v", hcnNetwork.Name, err)
+			if err == nil {
+				logger.Info("Successfully created hcn network with response", zap.Any("hnsResponse", hnsResponse))
+			} else {
+				if strings.Contains(err.Error(), "already exists") {
+					// fetch the network name again since the paralel CNI Add call has created the HNS network
```
MikeZappa87, line 367: spelling

network/network_windows.go
```diff
@@ -360,10 +360,20 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *EndpointInfo, extIf *exter
...
+					hnsResponse, err = Hnsv2.GetNetworkByName(hcnNetwork.Name)
```
tamilmani1989, line 368: no need to fetch again if cni gets already exists error.

behzad-mir: We need to fetch again to fill out hnsResponse since it is being used later on in the code

tamilmani1989: ok

network/network_windows.go
```diff
@@ -360,10 +360,20 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *EndpointInfo, extIf *exter
 		if errors.As(err, &hcn.NetworkNotFoundError{}) {
 			logger.Info("Creating hcn network", zap.Any("hcnNetwork", hcnNetwork))
 			hnsResponse, err = Hnsv2.CreateNetwork(hcnNetwork)
-			if err != nil {
-				return nil, fmt.Errorf("Failed to create hcn network: %s due to error: %v", hcnNetwork.Name, err)
+			if err == nil {
+				logger.Info("Successfully created hcn network with response", zap.Any("hnsResponse", hnsResponse))
+			} else {
+				if strings.Contains(err.Error(), "already exists") {
+					// fetch the network name again since the parallel CNI Add call has created the HNS network
```
tamilmani1989, line 367: lets add a log printing err.Error() before calling get hns network

network/network_windows.go
```diff
@@ -360,10 +360,20 @@ func (nm *networkManager) newNetworkImplHnsV2(nwInfo *EndpointInfo, extIf *exter
...
+					hnsResponse, err = Hnsv2.GetNetworkByName(hcnNetwork.Name)
+					if err != nil {
+						return nil, fmt.Errorf("Failed to get hcn network: %s due to error: %v", hcnNetwork.Name, err)
+					}
+					logger.Info("Successfully fetched hcn network with response", zap.Any("hnsResponse", hnsResponse))
```
tamilmani1989, line 372: since you are logging whole struct, lets make sure its not printing any byte array

network/network_windows.go
behzad-mir: done

network/network_windows.go
behzad-mir: Printed hnsreponse.ID instead in new commit

PR #3332, opened 2025-01-07
ci: skip check-log-errors in cilium connectivity

camrynl:
**Reason for Change**:
check-log-errors sometimes catches on transient errors and fails the entire connectivity test suite
this test case is skipped in other pipelines, adding skip to ACN until issue is resolved
opened a GH issue in cilium - https://github.com/cilium/cilium/issues/36992

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
+++ b/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
@@ -218,7 +218,7 @@ stages:
             displayName: "Get AKS Cluster"
           - script: |
               kubectl delete ns load-test
-              cilium connectivity test --connect-timeout 4s --request-timeout 30s --test '!pod-to-pod-encryption,!node-to-node-encryption'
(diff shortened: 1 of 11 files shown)


vipul-21:
Couple of things on the PR:
- Can we please update the description with the cilium issue link you opened.
- Is there a way we can retry the test in case of log failures rather than skipping it? The concern here is what if we have an issue which is not covered by the test and is reported in logs only ? Just curious, how frequently does this check fail ?

vipul-21:
> Is there a way we can retry the test in case of log failures rather than skipping it? 

We are already skipping in our cilium release pipeline.

vipul-21 (approved):

camrynl:
Will continue tracking github issue to see if we will eventually be able to reallow this case

joestringer:
Any chance you'd be open to submitting a change to Cilium to avoid the transient error log in the first place if Cilium can recover from it?

Issue #3299, opened 2025-01-03
log: [NPM] on crash, logs not sent to application insights


huntergregory:
If NPM crashes, it queues error logs before the crash, but these do not reach app insights since NPM does not clean up the telemetry handle.

huntergregory:
related to #3145

huntergregory:
Non-issue. Sending telemetry is a blocking call

huntergregory:
reopened after further investigation

PR #3337, opened 2025-01-08
fix(log): time waiting for appinsights to close was unbounded

huntergregory:
**Reason for Change**:

When CNS (and NPM after #3333) crash, they block on a call to close the app insights client so that queued logs are sent. This relies on a channel which might not close until much after the timeout which NPM/CNS specify. The ApplicationInsights-Go repo recommends having a timer of our own for the maximum time we want to wait for app insights to close.

This PR changes CNS and NPM so that they wait at most 30 seconds.

**Issue Fixed**:

Related to #3299.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/aitelemetry/telemetrywrapper.go
+++ b/aitelemetry/telemetrywrapper.go
@@ -27,6 +27,7 @@ const (
 	azurePublicCloudStr              = "AzurePublicCloud"
 	hostNameKey                      = "hostname"
 	defaultTimeout                   = 10
+	maxCloseTimeoutInSeconds         = 30
 	defaultBatchIntervalInSecs       = 15
 	defaultBatchSizeInBytes          = 32768
 	defaultGetEnvRetryCount          = 5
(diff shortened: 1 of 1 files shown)


aitelemetry/telemetrywrapper.go
```diff
@@ -27,6 +27,7 @@ const (
 	azurePublicCloudStr              = "AzurePublicCloud"
 	hostNameKey                      = "hostname"
 	defaultTimeout                   = 10
+	maxCloseTimeoutInSeconds         = 30
```
rbtr, line 30: I'm divided on whether this should start out as a `time.Duration` as best practice, or should remain an int to match the existing vars. It would be harder to accidentally set a very very small or very very large timeout if we created them all as durations from day zero...

aitelemetry/telemetrywrapper.go
```diff
@@ -330,8 +331,32 @@ func (th *telemetryHandle) Close(timeout int) {
...
 	// wait for items to be sent otherwise timeout
-	<-th.client.Channel().Close(time.Duration(timeout) * time.Second)
+	// similar to the example in the appinsights-go repo: https://github.com/microsoft/ApplicationInsights-Go#shutdown
+	select {
+	case <-th.client.Channel().Close(time.Duration(timeout) * time.Second):
+		// timeout specified for retries.
+
+		// If we got here, then all telemetry was submitted
+		// successfully, and we can proceed to exiting.
+	case <-time.After(time.Duration(maxWaitTimeInSeconds) * time.Second):
```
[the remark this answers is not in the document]
huntergregory: thanks, even if it wasn't necessary, I made sure to stop the timer now

aitelemetry/telemetrywrapper.go
huntergregory: I agree. I wasn't sure what is best, but I felt like continuing with the seconds-integer convention to lessen the blast radius for the existing code

timraymond: I am definitely team `time.Duration`, but I agree with keeping it consistent here. I think we need a public awareness campaign that `time.Duration` is the right first choice in many circumstances (it's silly to make the user do time math to come up with 300, when they could just have "5m" in a config, and it would Just Work).

rbtr (approved):

PR #3338, opened 2025-01-09
[NPM Lite] Default Deny - Updating Network Container Contract

rejain456:
**Reason for Change**:

as part of default deny work, updated network container contract to add DefaultDenyACL field. For background, cns will send cni the Default Deny acl's, if the default deny bool is set to true, which cni will use to create hns policies

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -505,7 +505,7 @@ type PodIpInfo struct {
 	// PnpId is set for backend interfaces, Pnp Id identifies VF. Plug and play id(pnp) is also called as PCI ID
 	PnPID string
 	// Default Deny ACL's to configure on HNS endpoints for Swiftv2 window nodes
-	EdpointPolicies []policy.Policy
+	EndpointPolicies []policy.Policy
 }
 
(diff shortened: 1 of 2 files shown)


cns/NetworkContainerContract.go
```diff
@@ -503,6 +504,8 @@ type PodIpInfo struct {
 	Routes []Route
 	// PnpId is set for backend interfaces, Pnp Id identifies VF. Plug and play id(pnp) is also called as PCI ID
 	PnPID string
+	// Default Deny ACL's to configure on HNS endpoints for Swiftv2 window nodes
+	DefaultDenyACL []acn.KVPair
```
thatmattlong, line 508: These are just generic ACLs that you intend to pass deny ACLs to, correct? If yes, let's call this EndpointACLs instead.

rejain456: Thanks for the suggestion, changed the name + type

thatmattlong (approved):

PR #3372, opened 2025-01-28
tool: validator script for Azure NPM to Cilium migration

rayaisaiah:
**Reason for Change**:
Adds a tool to validate cx network polices are safe to migrate from Azure NPM to Cilium.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,397 @@
...
+		fmt.Println("For more details please see \033[32maka.ms/azurenpmtocilium\033[0m.")
+	}
+}
+
+func checkEndportNetworkPolicies(policiesByNamespace map[string][]networkingv1.NetworkPolicy) bool {
+	networkPolicyWithEndport := false
+	for namespace, policies := range policiesByNamespace {
+		for _, policy := range policies {
+			foundEndPort := false
+			for _, egress := range policy.Spec.Egress {
```
huntergregory, line 108: ingress as well

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,397 @@
...
+					if port.EndPort != nil {
+						foundEndPort = true
+						if !networkPolicyWithEndport {
+							fmt.Printf("%-30s | %-30s \n", "NetworkPolicy with endPort", "❌")
+							fmt.Println("Policies affected:")
+							networkPolicyWithEndport = true
+						}
+						fmt.Printf("❌ Found NetworkPolicy: \033[31m%s\033[0m with endPort field in namespace: \033[31m%s\033[0m\n", policy.Name, namespace)
+						// Exit egress.port loop
+						break
```
huntergregory, line 119: enhancement: Instead of only printing the first flagrant policy, it would be nice to see a list of all flagrant policies. Same for other checks

tools/azure-npm-to-cilium-validator.go
huntergregory: I would make a function that you can reuse for Ingress and egress ports

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,397 @@
...
+	if err != nil {
+		log.Fatalf("Error getting namespaces: %v\n", err)
+	}
+
+	// Store network policies and services in maps
+	policiesByNamespace := make(map[string][]networkingv1.NetworkPolicy)
+	servicesByNamespace := make(map[string][]corev1.Service)
+
+	// Iterate over namespaces and store policies/services
+	for _, ns := range namespaces.Items {
```
huntergregory, line 49: this lint is suggesting to index `namespaces.Items[k]`

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,397 @@
...
+func checkCIDRNetworkPolicies(policiesByNamespace map[string][]networkingv1.NetworkPolicy) bool {
+	networkPolicyWithCIDR := false
+	for namespace, policies := range policiesByNamespace {
+		for _, policy := range policies {
+			foundCIDRIngress := false
+			foundCIDREgress := false
+			// Check the ingress field for cidr
+			for _, ingress := range policy.Spec.Ingress {
+				for _, from := range ingress.From {
+					if from.IPBlock != nil {
```
huntergregory, line 146: Let's make a common function between ingress/egress

tools/azure-npm-to-cilium-validator.go
huntergregory: Alternatively, copy `.Items` into a slice of pointers using indexing

huntergregory (changes requested):
Later can we unit test each check? For services, one UT for each case?

tools/azure-npm-to-cilium-validator.go
rayaisaiah: this does print all the flagrant polices it just print them as it finds them instead of storing them in a list and printing them all at the end. i can add them to a list and print at the end if preferred.

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,402 @@
...
+	for _, s := range slice {
+		if s == item {
+			return true
+		}
+	}
+	return false
+}
+
+func difference(slice1, slice2, slice3 []string) []string {
+	m := make(map[string]bool)
```
huntergregory, line 388: nit: don't worry about changing but in golang we usually use `map[string]struct{}` as a set. The `struct{}` is an empty struct which takes up no memory. You can verify if something is in the set with `ok`: `element, ok := m[s]`

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,402 @@
...
+		return true
+	}
+
+}
+
+func hasIngressPolicies(policies []networkingv1.NetworkPolicy) bool {
+	// Check if any policy is ingress
+	for _, policy := range policies {
+		for _, ingress := range policy.Spec.Ingress {
+			if len(ingress.From) > 0 {
```
huntergregory, line 299: or ingress.Ports

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,402 @@
...
+	return false
+}
+
+func checkServiceRisk(service v1.Service, namespace string, policiesListAtNamespace []networkingv1.NetworkPolicy, safeServices []string) []string {
+	for _, policy := range policiesListAtNamespace {
+		for _, ingress := range policy.Spec.Ingress {
+			// Check if there is an allow all ingress policy that matches labels the service is safe
+			if len(ingress.From) == 0 && len(ingress.Ports) == 0 {
+				// Check if there is an allow all ingress policy with empty selectors return true as the policy allows all services in the namespace
+				if len(policy.Spec.PodSelector.MatchLabels) == 0 {
```
huntergregory, line 313: edge case: there could be matchExpressions

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,402 @@
...
+		for _, ingress := range policy.Spec.Ingress {
+			if len(ingress.From) > 0 {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func checkServiceRisk(service v1.Service, namespace string, policiesListAtNamespace []networkingv1.NetworkPolicy, safeServices []string) []string {
```
huntergregory, line 307: nit: this function would be a little simpler if it returned a bool instead

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,402 @@
...
+			if len(ingress.From) == 0 && len(ingress.Ports) == 0 {
+				// Check if there is an allow all ingress policy with empty selectors return true as the policy allows all services in the namespace
+				if len(policy.Spec.PodSelector.MatchLabels) == 0 {
+					fmt.Printf("found an allow all ingress policy: %s with empty selectors so service %s in the namespace %s is safe\n", policy.Name, service.Name, namespace)
+					safeServices = append(safeServices, fmt.Sprintf("%s/%s", namespace, service.Name))
+					return safeServices
+				}
+				// Check if there is an allow all ingress policy that matches the service labels
+				if checkPolicyMatchServiceLabels(service.Spec.Selector, policy.Spec.PodSelector.MatchLabels) {
+					fmt.Printf("found an allow all ingress policy: %s with matching selectors so service %s in the namespace %s is safe\n", policy.Name, service.Name, namespace)
```
huntergregory, line 320: nit: let's only print verified issues (maybe we can keep these other logs as "debug" messages when enabled?)

huntergregory: Can hold off on this optimization until everything is finalized though

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,402 @@
...
+					safeServices = append(safeServices, fmt.Sprintf("%s/%s", namespace, service.Name))
+					return safeServices
+				}
+			}
+			// If there are no ingress from but there are ports in the policy; check if the service is safe
+			if len(ingress.From) == 0 && len(ingress.Ports) > 0 {
+				// If the policy targets all pods (allow all) or only pods that are in the service selector, check if traffic is allowed to all the service's target ports
+				if len(policy.Spec.PodSelector.MatchLabels) == 0 || checkPolicyMatchServiceLabels(service.Spec.Selector, policy.Spec.PodSelector.MatchLabels) {
```
huntergregory, line 328: nit: move all pod selector logic into a single function to reuse in both places

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,402 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func checkServiceTargetPortMatchPolicyPorts(servicePorts []v1.ServicePort, policyPorts []networkingv1.NetworkPolicyPort) bool {
+	ingressPorts := []string{}
+	for _, port := range policyPorts {
+		ingressPorts = append(ingressPorts, fmt.Sprintf("%d/%s", port.Port.IntVal, string(*port.Protocol)))
```
huntergregory, line 365: 1. protocol might be empty
2. Let's check string values (for named ports) as well

tools/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+
+	// Check if all the services target ports are in the policies ingress ports
+	for _, port := range servicePorts {
+		servicePort := fmt.Sprintf("%d/%s", port.TargetPort.IntValue(), port.Protocol)
+		if !contains(ingressPorts, servicePort) {
```
huntergregory, line 371: Might as well do a double for a loop here without the string concatenation

huntergregory (commented):
comments on the service check

tools/azure-npm-to-cilium-validator.go
rayaisaiah: wouldn't it also save storage and be optimal by avoiding storing flagrant policies to a list if possible?

tools/azure-npm-to-cilium-validator.go
rayaisaiah: good point updated

tools/azure-npm-to-cilium-validator.go
rayaisaiah: updated to use slice of pointers

tools/azure-npm-to-cilium-validator.go
huntergregory: Agh I misread. This is good at least for now

rayaisaiah: updated to print all flagrant policies. i chose not to add them to a list but instead print as i find them to be optimal in storage but if its preferred i can add them to a list

tools/azure-npm-to-cilium-validator.go
rayaisaiah: Added ingress/egress helpers for both endport and cidr checks

tools/azure-npm-to-cilium-validator.go
rayaisaiah: ohh thats useful! i wasnt sure how that worked

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,408 @@
...
+		return true
+	}
+
+}
+
+func hasIngressPolicies(policies []networkingv1.NetworkPolicy) bool {
+	// Check if any policy is ingress
+	for _, policy := range policies {
+		for _, ingress := range policy.Spec.Ingress {
+			if len(ingress.From) > 0 || len(ingress.Ports) > 0 {
```
rayaisaiah, line 298: Sanity check this is correct logic

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,408 @@
...
+				if foundEndPort {
+					foundNetworkPolicyWithEndport = true
+					break
+				}
+			}
+		}
+	}
+	// Print no impact if no network policy has endport
+	if !foundNetworkPolicyWithEndport {
+		log.Printf("%-30s | %-30s \n", "NetworkPolicy with endPort", "✅")
```
rayaisaiah, line 133: using `log.Print` instead of `fmt.Print` to verify prints in unit tests. is this okay?

tools/azure-npm-to-cilium-validator.go
rayaisaiah: added

tools/azure-npm-to-cilium-validator.go
rayaisaiah: updated

tools/azure-npm-to-cilium-validator.go
rayaisaiah: good point i can update the list outside the function

rayaisaiah: updated

tools/azure-npm-to-cilium-validator.go
rayaisaiah: i can update if you feel its more optimal

tools/azure-npm-to-cilium-validator.go
rayaisaiah: this was only for testing purposes and will be removed

tools/azure-npm-to-cilium-validator.go
rayaisaiah: updated and added a todo to further optimize after printing logs [here](https://github.com/Azure/azure-container-networking/pull/3372#discussion_r1942094961)

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,411 @@
...
+			// Check if there is an allow all ingress policy that matches labels the service is safe
+			if len(ingress.From) == 0 && len(ingress.Ports) == 0 {
+				// Check if there is an allow all ingress policy with empty selectors return true as the policy allows all services in the namespace
+				if checkPolicySelectorsAreEmpty(policy.Spec.PodSelector) {
+					fmt.Printf("found an allow all ingress policy: %s with empty selectors so service %s in the namespace %s is safe\n", policy.Name, service.Name, namespace)
+					return true
+				}
+				// Check if there is an allow all ingress policy that matches the service labels
+				if checkPolicyMatchServiceLabels(service.Spec.Selector, policy.Spec.PodSelector.MatchLabels) {
+					// TODO add this to above logic and check in one if statement after i am done printing the logs
```
rayaisaiah, line 320: reminder to do this

tools/azure-npm-to-cilium-validator.go
rayaisaiah: 1. should we just compare port values better policy and service? could a service be allowed if it only has a port `8080` and in a namespace with a policy with port `8080` and protocol `TCP`?
2. Add a check if the target port is a string then it is a named port and service is at risk and to return `false`

tools/azure-npm-to-cilium-validator.go
rayaisaiah: updated

rayaisaiah:
Build tables utilizing: https://statics.teams.cdn.office.net/evergreen-assets/safelinks/1/atp-safelinks.html

tools/azure-npm-to-cilium-validator.go
rayaisaiah: Update: looks like policies defaults to `TCP` if protocol is left blank

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+	return egressPolicies
+}
+
+func checkExternalTrafficPolicyServices(namespaces *corev1.NamespaceList, servicesByNamespace *map[string][]corev1.Service, policiesByNamespace *map[string][]networkingv1.NetworkPolicy) ([]string, []string) {
+	var servicesAtRisk, noSelectorServices, safeServices []string
+
+	for _, namespace := range namespaces.Items {
+		// Check if are there ingress policies in the namespace if not skip
+		policyListAtNamespace := (*policiesByNamespace)[namespace.Name]
```
rayaisaiah, line 168: is it okay to deference the pointers like this?

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+
+	// Get namespaces
+	namespaces, err := clientset.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		log.Fatalf("Error getting namespaces: %v\n", err)
+	}
+
+	// Store network policies and services in maps
+	policiesByNamespace := make(map[string][]networkingv1.NetworkPolicy)
```
huntergregory, line 45:
```suggestion
	policiesByNamespace := make(map[string][]*networkingv1.NetworkPolicy)
```

Reference a pointer to the _struct_ prevents the need to copy the full struct around

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,446 @@
...
+			continue
+		}
+		servicesByNamespace[ns.Name] = services.Items
+	}
+
+	// Print the migration summary
+	printMigrationSummary(namespaces, &policiesByNamespace, &servicesByNamespace)
+}
+
+func checkEndportNetworkPolicies(policiesByNamespace *map[string][]networkingv1.NetworkPolicy) ([]string, []string) {
```
huntergregory, line 77:
```suggestion
func checkEndportNetworkPolicies(policiesByNamespace map[string][]*networkingv1.NetworkPolicy) ([]string, []string) {
```

A map is already a pointer under the hood. We rarely if ever need a pointer to a pointer

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
huntergregory: see:
- https://github.com/Azure/azure-container-networking/pull/3372/files#r1943853384
- https://github.com/Azure/azure-container-networking/pull/3372/files#r1943852320

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: this makes so much more sense thanks for explaining it. i forgot how pointers worked its been a minute haha

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: this makes a lot more sense than what i was doing when i think about it

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,452 @@
...
+	return false
+}
+
+func getEgressPolicies(policiesByNamespace map[string][]*networkingv1.NetworkPolicy) []string {
+	var egressPolicies []string
+	for namespace, policies := range policiesByNamespace {
+		for _, policy := range policies {
+			for _, egress := range policy.Spec.Egress {
+				// If the policy has a egress field thats not an egress allow all flag it
+				if len(egress.To) > 0 || len(egress.Ports) > 0 {
```
rayaisaiah, line 159: this logic will also not catch egress deny all is that okay or should we flag that as well and only allow egress allow all?

huntergregory: I believe deny all == Egress is nil? Could double check the logic in translation.go

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,448 @@
...
+		}
+	}
+	return false
+}
+
+func checkPolicySelectorsAreEmpty(podSelector *metav1.LabelSelector) bool {
+	return len(podSelector.MatchLabels) == 0 && len(podSelector.MatchExpressions) == 0
+}
+
+func checkPolicyMatchServiceLabels(serviceLabels, policyLabels *map[string]string) bool {
```
rayaisaiah, line 253: is this pointer okay since i cant make them `map[string]*string`?

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: egress is nil in deny all but then it wont get flagged when trying to find egress network policies for the N`PM does not enforce policies on egress to local node IP. Cilium does.` check as i thought only egress allow all polices are safe

tools/azure-npm-to-cilium-validator.go
rayaisaiah: updated

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,448 @@
...
+		}
+	}
+	return false
+}
+
+func checkPolicySelectorsAreEmpty(podSelector *metav1.LabelSelector) bool {
+	return len(podSelector.MatchLabels) == 0 && len(podSelector.MatchExpressions) == 0
+}
+
+func checkPolicyMatchServiceLabels(serviceLabels, policyLabels map[string]string) bool {
```
rayaisaiah, line 253: okay these are not pointers as they are `map[string]string`?

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator_test.go
```diff
@@ -0,0 +1,589 @@
...
+								{},
+							},
+						},
+					},
+				},
+			},
+			expectedEgressPolicies: []string{},
+		},
+		{
+			name: "Deny all egress policy",
```
rayaisaiah, line 442: update this test if it should add it to list

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,448 @@
...
+			}
+		}
+		if !serviceTargetPortPolicyPort {
+			return false
+		}
+	}
+	return true
+}
+
+func difference(slice1, slice2, slice3 *[]string) []string {
```
rayaisaiah, line 302: return a pointer?

timraymond: Slices are (wide) pointers :)

rayaisaiah:
> Check how netpols with no ports work in npm with services with protocols and ports (services are required to have ports) image

image
it looks like if a network policy on NPM just targets a protocol itll let all traffic with that protocol (no matter the ports) in

tools/azure-npm-to-cilium-validator.go
rayaisaiah: removed prints

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+				// If the service has externalTrafficPolicy is set to "Cluster" add it to the riskServices list (ExternalTrafficPolicy: "" defaults to Cluster)
+				if externalTrafficPolicy != corev1.ServiceExternalTrafficPolicyTypeLocal {
+					// Any service with externalTrafficPolicy=Cluster is at risk so need to elimate any services that are incorrectly flagged
+					riskServices = append(riskServices, fmt.Sprintf("%s/%s", namespace.Name, service.Name))
+					// If the service has no selector add it to the noSelectorServices list
+					if service.Spec.Selector == nil {
+						noSelectorServices = append(noSelectorServices, fmt.Sprintf("%s/%s", namespace.Name, service.Name))
+					}
+					// Check if are there services with selector that match the network policy
+					if checkServiceRisk(service, policyListAtNamespace) {
```
huntergregory, line 194: use if-else here. Services without selectors don't need to be checked

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+						safeServices = append(safeServices, fmt.Sprintf("%s/%s", namespace.Name, service.Name))
+					}
+				}
+			}
+		}
+	}
+
+	// Remove all the safe services from the services at risk
+	unsafeRiskServices = difference(&riskServices, &safeServices)
+	// Remove all the safe services from the no selector services
```
huntergregory, line 204: all services without selectors are actually unsafe

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+	// Get services that have externalTrafficPolicy!=Local
+	unsafeRiskServices, unsafeNoSelectorServices := getExternalTrafficPolicyClusterServices(namespaces, servicesByNamespace, policiesByNamespace)
+
+	// Print the services that are at risk
+	printUnsafeServices(&unsafeRiskServices, &unsafeNoSelectorServices)
+
+	fmt.Println("+------------------------------+-------------------------------+")
+	if len(ingressEndportNetworkPolicy) > 0 || len(egressEndportNetworkPolicy) > 0 ||
+		len(ingressPoliciesWithCIDR) > 0 || len(egressPoliciesWithCIDR) > 0 ||
+		len(egressPolicies) > 0 || len(unsafeRiskServices) > 0 {
```
huntergregory, line 356:
```suggestion
		len(egressPolicies) > 0 || len(unsafeRiskServices) > 0 ||  len(unsafeNoSelectorService) > 0 {
```

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
huntergregory: right, no need for pointers for basic data types

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+		}
+	}
+	return false
+}
+
+func checkPolicySelectorsAreEmpty(podSelector *metav1.LabelSelector) bool {
+	return len(podSelector.MatchLabels) == 0 && len(podSelector.MatchExpressions) == 0
+}
+
+func checkPolicyMatchServiceLabels(serviceLabels, policyLabels map[string]string) bool {
```
huntergregory, line 250: If the policy has matched expressions, return false

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+		// If the target port is a string then it is a named port and service is at risk
+		if servicePort.TargetPort.Type == intstr.String {
+			return false
+		}
+
+		// Check if all the services target ports are in the policies ingress ports
+		matchedserviceTargetPortToPolicyPort := false
+		for _, policyPort := range *policyPorts {
+			// Check if the policys port exists
+			if policyPort.Port == nil {
```
huntergregory, line 284: as discussed offline: empty policy port matches all ports

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+				return false
+			}
+			// If the port is a string then it is a named port and service is at risk
+			if policyPort.Port.Type == intstr.String {
+				return false
```
huntergregory, line 289: I would just ignore this port and continue on instead

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+			}
+			if servicePort.TargetPort.IntValue() == int(policyPort.Port.IntVal) && string(servicePort.Protocol) == string(*policyPort.Protocol) {
```
huntergregory, line 291: nit: handle empty protocols (default to TCP?)

One might be empty and the other might not

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator_test.go
rayaisaiah: updated as it is denying egress traffic

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated to catch egress deny all

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+		if !matchedPolicyLabelToServiceLabel {
+			return false
+		}
+	}
+	return true
+}
+
+func checkServiceTargetPortMatchPolicyPorts(servicePorts *[]corev1.ServicePort, policyPorts *[]networkingv1.NetworkPolicyPort) bool {
+	for _, servicePort := range *servicePorts {
+		// If the target port is a string then it is a named port and service is at risk
```
huntergregory, line 275: just to be safe, can we use port instead if targetport has a zero value?

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+			}
+		}
+		if !matchedPolicyLabelToServiceLabel {
+			return false
+		}
+	}
+	return true
+}
+
+func checkServiceTargetPortMatchPolicyPorts(servicePorts *[]corev1.ServicePort, policyPorts *[]networkingv1.NetworkPolicyPort) bool {
```
rayaisaiah, line 273: > Check how netpols with no ports work in npm with services with protocols and ports (services are required to have ports) image

image
it looks like if a network policy on NPM just targets a protocol itll let all traffic with that protocol (no matter the ports) in 

image

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+	var diff []string
+	for _, s := range *slice1 {
+		if _, ok := m[s]; !ok {
+			diff = append(diff, s)
+		}
+	}
+	return diff
+}
+
+func printMigrationSummary(namespaces *corev1.NamespaceList, policiesByNamespace map[string][]*networkingv1.NetworkPolicy, servicesByNamespace map[string][]*corev1.Service) {
```
rayaisaiah, line 317: Build tables utilizing: https://statics.teams.cdn.office.net/evergreen-assets/safelinks/1/atp-safelinks.html

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+		}
+		if !matchedPolicyLabelToServiceLabel {
+			return false
+		}
+	}
+	return true
+}
+
+func checkServiceTargetPortMatchPolicyPorts(servicePorts *[]corev1.ServicePort, policyPorts *[]networkingv1.NetworkPolicyPort) bool {
+	for _, servicePort := range *servicePorts {
```
huntergregory, line 274: I noticed that Cilium treats port 0 in a special way. Can we return false if either port is 0 or port is non-zero and targetport is 0?

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,450 @@
...
+					egressPoliciesWithEndport = append(egressPoliciesWithEndport, fmt.Sprintf("%s/%s", namespace, policy.Name))
+					break
+				}
+			}
+		}
+	}
+	return ingressPoliciesWithEndport, egressPoliciesWithEndport
+}
+
+func checkEndportInPolicyRules(ports *[]networkingv1.NetworkPolicyPort) bool {
```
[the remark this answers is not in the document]
huntergregory: Saw this in several places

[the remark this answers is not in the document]
rayaisaiah: updated. what about the difference function since it is taking in two string arrays?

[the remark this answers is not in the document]
rayaisaiah: or the print comments taking in the arrays? should i be return pointers to the array for functions like `getEndportNetworkPolicies`?

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: what about if there is a service with no selector and an allow all policy?

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: see https://github.com/Azure/azure-container-networking/pull/3372#discussion_r1947096277

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
huntergregory: true! And in your code, there's no chance of a policy's selector matching an empty service selector, right? Unless policy selector is empty (allow all)

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated but might not be necessary depending on logic with overlap between `unsafeRiskServices `and `unsafeNoSelectorServices` currently

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: good catch updated

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: empty protocols default to TCP for both policies and services

image

image

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: wont targetport default to port if left blank?
image
and a service has to have a port to be valid?

rayaisaiah: wont targetport default to port if left blank?
image
and a service has to have a port to be valid?
image

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated to return false if policy port is 0 or service target port is 0

rayaisaiah:
TODO: add a go mod file and a read me to this folder as well with instructions on how to run the script (including running go mod tidy/vendor in this subfolder)

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: no because in `checkPolicyMatchServiceLabels` it checks

```
if len(policyLabels) > len(serviceLabels) {
	return false
}
```

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
huntergregory: In terms of "does this policy match the service", this function returns true if there are no matchLabels even if there are matchExpressions causing the policy _not_ to match the service

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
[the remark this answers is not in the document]
huntergregory: there's no need to point to an array or a map. It makes sense for structs _within_ maps/arrays to be pointers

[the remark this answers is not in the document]
rayaisaiah: i see that makes a lot of thense. ill keep that in mind thank you!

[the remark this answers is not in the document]
rayaisaiah: updated

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated and added a ut `LoadBalancer service with externalTrafficPolicy=Cluster with no selector and an allow all ingress policy with a matchExpressions selector`

rayaisaiah:
TODO: Additionally, if a LoadBalancer "allows all" with ports, it must allow 168.63.129.16 without ports.

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -218,44 +225,43 @@ func hasIngressPolicies(policies []*networkingv1.NetworkPolicy) bool {
...
+	// Check if there is an allow all ingress policy with empty selectors if so the service is safe
+	if len(podSelector.MatchLabels) == 0 && len(podSelector.MatchExpressions) == 0 {
+		return true
+	}
 
-func checkPolicyMatchServiceLabels(serviceLabels, policyLabels map[string]string) bool {
-	// Return false if the policy has more labels than the service
-	if len(policyLabels) > len(serviceLabels) {
+	// Return false if the policy has no labels or more labels than the service
+	if len(podSelector.MatchLabels) == 0 || len(podSelector.MatchLabels) > len(serviceLabels) {
```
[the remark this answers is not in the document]
rayaisaiah: isnt it more direct to check the value explicitly? cause technically matchlabels and matchexpressions could both be greater than 0. maybe update to check if 

```
len(podSelector.MatchLabels) == 0 || len(podSelector.MatchExpressions) > 0 || 
len(podSelector.MatchLabels) > len(serviceLabels)
```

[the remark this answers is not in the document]
rayaisaiah: if a policy has a matchlabels and matchexpressions and the matchlabels correctly matches to the service labels that should still return false?

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -270,24 +276,51 @@ func checkPolicyMatchServiceLabels(serviceLabels, policyLabels map[string]string
...
+				}
+				continue
+			}
+			// If the port is a string then it is a named port and it cant be evaluated
 			if policyPort.Port.Type == intstr.String {
+				continue
+			}
+			// If the target port is 0 then it is at risk as Cilium treats port 0 in a special way
+			if int(policyPort.Port.IntVal) == 0 {
 				return false
```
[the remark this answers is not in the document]
rayaisaiah: good catch updated!

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -270,24 +276,51 @@ func checkPolicyMatchServiceLabels(serviceLabels, policyLabels map[string]string
...
+
 		// Check if all the services target ports are in the policies ingress ports
 		matchedserviceTargetPortToPolicyPort := false
-		for _, policyPort := range *policyPorts {
-			// Check if the policys port exists
-			if policyPort.Port == nil {
+		for _, policyPort := range policyPorts {
+			// Check if the policys port and protocol exists
+			if policyPort.Port == nil && policyPort.Protocol == nil {
 				return false
```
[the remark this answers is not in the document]
rayaisaiah: this might be able to be safely removed as we check `len(ingress.Ports) > 0 ` before calling the function so that there would have to be at least one port value.

if not you are correct and we can consider it a match as the policy would match all ports/protocols in that namespace if its `ports.port` and `ports.protocol` are nil

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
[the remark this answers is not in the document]
rayaisaiah: discussed offline and updated

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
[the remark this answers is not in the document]
rayaisaiah: discussed outside and updated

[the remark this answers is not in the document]
rayaisaiah: added note: Note: ingress.Ports.protocol will never be nil if len(ingress.Ports) is greater than 0. It defaults to "TCP" if not set

rayaisaiah:
Squash commits before merging?

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,507 @@
...
+		if !matchedserviceTargetPortToPolicyPort {
+			return false
+		}
+	}
+	return true
+}
+
+func checkAllowsLoadBalancerIngressCIDR(from []networkingv1.NetworkPolicyPeer) bool {
+	for _, peer := range from {
+		if peer.IPBlock != nil && peer.IPBlock.CIDR == "168.63.129.16/32" {
```
rayaisaiah, line 352: use the net package or something to see if our IP is contained in the CIDR

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,493 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	// Remove all the safe services from the services at risk
+	unsafeRiskServices = difference(riskServices, safeServices)
+	// Remove all the safe services from the no selector services
+	unsafeNoSelectorServices = difference(noSelectorServices, safeServices)
```
rayaisaiah, line 214: do we need to even return no selector services?

unsafeRiskServices = riskServices - safeServices
unsafeNoSelectorServices = noSelectorServices - safeServices

all unsafeNoSelectorServices will be either safeServices or unsafeRiskServices

rayaisaiah: updated

rayaisaiah: added the service selectors to the other arrays instead so still prints out whether the service has a selector or not

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -282,6 +296,12 @@ func checkServiceTargetPortMatchPolicyPorts(servicePorts []corev1.ServicePort, p
 		return false
 	}
 
+	// If the policy is allowing no traffic from ports then the service is at risk
+	// Note: ingress.Ports.protocol will never be nil if len(ingress.Ports) is greater than 0. It defaults to "TCP" if not set
+	if len(policyPorts) == 0 {
```
huntergregory, line 301: if there are no ports, it allows all ports. If there is a port, it matters what is inside the port

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -191,26 +194,25 @@ func getExternalTrafficPolicyClusterServices(
...
-					riskServices = append(riskServices, fmt.Sprintf("%s/%s", namespace.Name, service.Name))
-					// If the service has no selector add it to the noSelectorServices list
+					// Check if the service has a selector
+					serviceSelector := "selectors"
 					if service.Spec.Selector == nil {
-						noSelectorServices = append(noSelectorServices, fmt.Sprintf("%s/%s", namespace.Name, service.Name))
+						serviceSelector = "no selectors"
 					}
-					// Check if are there services with selector that match the network policy
+					riskServices = append(riskServices, fmt.Sprintf("%s/%s/%s", namespace.Name, service.Name, serviceSelector))
```
huntergregory, line 202: can we remove the selector info? It actually won't be relevant for printing

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: your right this explicit check is not needed as the logic has already checked port > 0 in `checkNoServiceRisk`

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -236,8 +238,14 @@ func checkNoServiceRisk(service *corev1.Service, policiesListAtNamespace []*netw
 					return true
 				}
 			}
+			// Check if service is a loadbalancer and policy allows 168.63.129.16 and has no ports
+			if service.Spec.Type == corev1.ServiceTypeLoadBalancer && len(ingress.Ports) == 0 {
+				if checkPolicyMatchServiceLabels(service.Spec.Selector, policy.Spec.PodSelector) && checkAllowsLoadBalancerIP(ingress.From) {
```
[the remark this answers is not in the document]
rayaisaiah: good optimization added

rayaisaiah:
TODO: Add a check for policies with ingress or egress named ports

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,500 @@
...
+					}
+				}
+			}
+		}
+	}
+	return false
+}
+
+func checkPolicyMatchServiceLabels(serviceLabels map[string]string, podSelector metav1.LabelSelector) bool {
+	// Check if there is an allow all ingress policy with empty selectors if so the service is safe
```
huntergregory, line 258: nit: this means "target all"

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,500 @@
...
+					if from.IPBlock != nil && from.IPBlock.CIDR != "" {
+						// Check if the health probe IP is blocked in the CIDR except
+						if from.IPBlock.Except != nil {
+							for _, except := range from.IPBlock.Except {
+								_, excecptCidr, err := net.ParseCIDR(except)
+								if err != nil {
+									continue
+								}
+								if excecptCidr.Contains(healthProbeIP) {
+									return false
```
huntergregory, line 361: should break instead. This rule doesn't allow the IP but another rule might

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,500 @@
...
+				if len(ingress.From) == 0 && len(ingress.Ports) == 0 {
+					return true
+				}
+				// If the policy targets all pods (allow all) or only pods that are in the service selector, check if traffic is allowed to all the service's target ports
+				// Note: ingress.Ports.protocol will never be nil if len(ingress.Ports) is greater than 0. It defaults to "TCP" if not set
+				if len(ingress.Ports) > 0 && checkServiceTargetPortMatchPolicyPorts(service.Spec.Ports, ingress.Ports) {
+					switch service.Spec.Type {
+					// If service the service is a node port check the policy does not have from rules that could disrupt traffic
+					case corev1.ServiceTypeNodePort:
+							if len(ingress.From) == 0 {
```
huntergregory, line 241: Should unconditionally make this check for both node port and load balancer

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: removed health probe ip check

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated now that there is no health probe ip check for ingress.from.ipBlock.cidr

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated

vakalapa:
Can we also add telemetry from this script back to our kusto ? We can reuse the NPM telemetry ? Get back the details of which fqdn (clusters') customer has run the script on and the result of the run.

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,505 @@
...
+	// Get services that have externalTrafficPolicy!=Local that are unsafe (might have traffic disruption)
+	unsafeServices := getUnsafeExternalTrafficPolicyClusterServices(namespaces, servicesByNamespace, policiesByNamespace)
+	// Add the services that are at risk
+	addUnsafeServicesToTable(table, unsafeServices)
+
+	table.Render()
+
+	// Print the total number of policies
+	printTotalPolicies(policiesByNamespace)
+
```
rayaisaiah, line 418: print number of pods per namespace too?

rayaisaiah:
limit the output of each check to 25 policies/clusters to reduce output length?

tools/azure-npm-to-cilium-validator/README.md
```diff
@@ -0,0 +1,50 @@
...
+```bash
+go mod tidy && go mod vendor
+```
+
+## Running the Tool
+
+Run the following command with the path to your kube config file with the cluster you want to validate.
+
+```bash
+go run azure-npm-to-cilium-validator.go --kubeconfig ~/.kube/config
```
rayaisaiah, line 37: update with flag commands (after i get the right name)

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,644 @@
...
+	policiesByNamespace map[string][]*networkingv1.NetworkPolicy,
+	servicesByNamespace map[string][]*corev1.Service,
+	podsByNamespace map[string][]*corev1.Pod,
+) {
+	// Get the network policies with endports
+	ingressEndportNetworkPolicy, egressEndportNetworkPolicy := getEndportNetworkPolicies(policiesByNamespace)
+
+	// Send endPort telemetry
+	// Note: The operationID is set to a high number so it doesn't conflict with other telemetry
+	metrics.SendLog(10000, fmt.Sprintf("Found %d network policies with endPort", len(ingressEndportNetworkPolicy)+len(egressEndportNetworkPolicy)), metrics.DonotPrint)
```
huntergregory, line 118: nit: make 10000 a const

huntergregory, line 118: add prefix like "[migration script]..."

huntergregory (approved):

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
```diff
@@ -0,0 +1,646 @@
...
+		}
+		podsByNamespace[ns.Name] = make([]*corev1.Pod, len(pods.Items))
+		for i := range pods.Items {
+			podsByNamespace[ns.Name][i] = &pods.Items[i]
+		}
+	}
+
+	// Create telemetry handle
+	// Note: npmVersionNum and imageVersion telemetry is not needed for this tool so they are set to abitrary values
+	err = metrics.CreateTelemetryHandle(0, "", "014c22bd-4107-459e-8475-67909e96edcb")
```
vakalapa, line 99: Even if NPM version is not needed, can we give some specific version or delimiter here ? so the search on these logs becomes very easy. 

Like say NPM-script-v0.0.1 something ?, so the kusto query can just filter on this ?

vakalapa: this version based filter would be better than log based "has" filter

vakalapa (dismissed):

tools/azure-npm-to-cilium-validator/azure-npm-to-cilium-validator.go
rayaisaiah: updated

vakalapa (approved):

huntergregory (approved):

PR #3375, opened 2025-01-28
fix: route issues on Swiftv2 Windows

paulyufan2:
**Reason for Change**:

This is the long term solution to fix Swiftv2 Windows Routes issues to make sure windows cns gets all required cidrs from AKS.
Fixes include:
1.Windows CNS gets infravnet/pod/node cidrs from configs
2.Add these routes and send them to the CNI

**Issue Fixed**:

In Linux Swiftv2, CNS fetches infravnet/pod/node cidrs from the node's envs set by AKS; we should do same thing for Windows;

**Requirements**:

This PR is based on npm-lite PR with cns middleware refactoring and linked with this PR:
https://github.com/Azure/azure-container-networking/pull/3205

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/middlewares/k8sSwiftV2.go
```diff
@@ -249,3 +158,118 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
...
+		routes[i] = cns.Route{
+			IPAddress:        cidr,
+			GatewayIPAddress: gatewayIP,
+		}
+	}
+	return routes
+}
+
+// CNS gets node and service CIDRs from configuration env and parse them to get the v4 and v6 IPs
+func (k *K8sSWIFTv2Middleware) GetCidrs() (v4IPs, v6IPs []string, err error) {
```
QxBytes, line 242: I believe there was a mention we shouldn't have named returns

cns/middlewares/k8sSwiftV2_linux_test.go
```diff
@@ -3,6 +3,7 @@ package middlewares
 import (
 	"context"
 	"fmt"
+	"reflect"
```
QxBytes, line 6: I believe tim also mentioned an alternative to reflect here

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -58,3 +80,166 @@ func (k *K8sSWIFTv2Middleware) addDefaultRoute(podIPInfo *cns.PodIpInfo, gwIP st
...
+		return ipConfigsResp, nil
+	}
+}
+
+func GetDefaultDenyBool(mtpnc v1alpha1.MultitenantPodNetworkConfig) bool {
+	// returns the value of DefaultDenyACL from mtpnc
+	return mtpnc.Status.DefaultDenyACL
+}
+
+func (k *K8sSWIFTv2Middleware) getInfraRoutes(podIPInfo *cns.PodIpInfo) ([]cns.Route, error) {
```
QxBytes, line 223: Please add comments to both getInfraRouteImpls regarding what they do and what the gateway ip for example is/doing

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -16,10 +32,16 @@ func (k *K8sSWIFTv2Middleware) setRoutes(podIPInfo *cns.PodIpInfo) error {
 		// TODO: Remove this once HNS fix is ready
 		route := cns.Route{
 			IPAddress:        "0.0.0.0/0",
-			GatewayIPAddress: "0.0.0.0",
+			GatewayIPAddress: defaultGateway,
```
QxBytes, line 35: I believe your other PR had some addDefaultRoute-- what happened to it here-- is it not used?

cns/middlewares/k8sSwiftV2_windows_test.go
```diff
@@ -1,22 +1,30 @@
...
 	"github.com/Azure/azure-container-networking/crd/multitenancy/api/v1alpha1"
+	"github.com/Azure/azure-container-networking/network/policy"
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
 	"gotest.tools/v3/assert"
 )
 
 func TestSetRoutesSuccess(t *testing.T) {
 	middleware := K8sSWIFTv2Middleware{Cli: mock.NewClient()}
+	t.Setenv(configuration.EnvServiceCIDRs, "10.0.0.0/16")
```
QxBytes, line 21: could we also set the Pod cidr variable here like in the linux case to ensure that it does not get consumed?

cns/middlewares/k8sSwiftV2_linux.go
```diff
@@ -30,50 +32,12 @@ func (k *K8sSWIFTv2Middleware) setRoutes(podIPInfo *cns.PodIpInfo) error {
 		routes = append(routes, virtualGWRoute, route)
 
 	case cns.InfraNIC:
-		// Get and parse infraVNETCIDRs from env
-		infraVNETCIDRs, err := configuration.InfraVNETCIDRs()
+		// get service and infravnet routes
```
QxBytes, line 35: Could you modify this comment-- https://github.com/Azure/azure-container-networking/compare/jainriya/npmliteCNSchanges...coderefactorcnsswiftv2#diff-70deeff65e796118967cdd314e7812a08e0fba22d64f116eaf098a50a250642bR35 and this one https://github.com/Azure/azure-container-networking/compare/jainriya/npmliteCNSchanges...coderefactorcnsswiftv2#diff-bc799fe974a6b7f59d815d1e4123ea023fec16051cba9edb16815e8d36973823R39 to be more explicit with what they are doing since the current comment makes it seem like they are doing the same thing, but in reality the linux one is doing more (adds in pod cidrs too)

cns/middlewares/k8sSwiftV2_linux.go
```diff
@@ -86,15 +50,57 @@ func (k *K8sSWIFTv2Middleware) setRoutes(podIPInfo *cns.PodIpInfo) error {
...
-func addRoutes(cidrs []string, gatewayIP string) []cns.Route {
-	routes := make([]cns.Route, len(cidrs))
-	for i, cidr := range cidrs {
-		routes[i] = cns.Route{
-			IPAddress:        cidr,
-			GatewayIPAddress: gatewayIP,
-		}
+// CNS gets pod CIDRs from configuration env and parse them to get the v4 and v6 IPs
+// Containerd reassigns the IP to the adapter and kernel configures the pod cidr route by default, so windows swiftv2 does not require pod cidr
+func (k *K8sSWIFTv2Middleware) GetPodCidrs() (v4IPs, v6IPs []string, err error) {
```
QxBytes, line 55: I think you moved this into the linux impl from the previous pr but the code to get the pod cidrs themselves isn't linux-specific, right? Like: would this code still work on windows (even if we don't use it on windows)? If the above is yes, it can remain in the common golang file (and if so move that comment about windows on line 54 to getInfraRoutes or an os-specific function), otherwise you can ignore this

QxBytes (commented):
Could you ensure that all functions you add have detailed comments about them? If they are os-specific, it would be helpful to explain what that function does such that when we read the other os-specific implementation, we can determine what the difference is easily.

QxBytes (approved):

rbtr (approved):
diff against #3286 LGTM, thanks for integrating the OS changes here

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -50,11 +72,174 @@ func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(podIPInfo *cns.Pod
...
+func (k *K8sSWIFTv2Middleware) getInfraRoutes(podIPInfo *cns.PodIpInfo) ([]cns.Route, error) {
+	var routes []cns.Route
+
+	ip, err := netip.ParseAddr(podIPInfo.PodIPConfig.IPAddress)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse podIPConfig IP address %s", podIPInfo.PodIPConfig.IPAddress)
+	}
+
+	// TODO: add ipv6 when supported
+	v4IPs, _, err := k.GetCidrs()
```
tamilmani1989, line 235: what k.GetCidrs() returns? can we comment?

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -50,11 +72,174 @@ func (k *K8sSWIFTv2Middleware) assignSubnetPrefixLengthFields(podIPInfo *cns.Pod
 	return nil
 }
 
-// add default route with gateway IP to podIPInfo
+// add default route with gateway IP to podIPInfo for delegatedNIC interface
```
tamilmani1989, line 75: this api doesn't care delegated nic or infra nic. move this change to place where you call this api

cns/middlewares/k8sSwiftV2_windows.go
tamilmani1989: probably consider renaming to GetInfravnetAndServiceCidrs()

cns/middlewares/k8sSwiftV2_windows.go
paulyufan2: it's os-specific codes which was checked in short-term solution; yes it does not care about nic type, just modified the comment here

tamilmani1989 (approved):

kmurudi (approved):
lgtm, pending testing/validation on cluster

QxBytes (approved):

PR #3376, opened 2025-01-29
ci: increase goldpinger replicas from 4 to 8 and add topology constraints

QxBytes:
**Reason for Change**:

Attempts to fix this issue in our pipelines: https://msazure.visualstudio.com/One/_build/results?buildId=113599589&view=logs&j=b1429627-982b-5d7c-f874-2297f1590463&t=f4ed6e9d-3ea8-54bd-90c0-ce2d41a370dc

It looks like there is a possibility that one node becomes unbalanced (ex: both metrics pods, gateway keepers etc. all land on the same node), and so new pods are scheduled all on the other node to balance resources. This PR increases the number of replicas such that it is more likely that at least two pods land on each node (which is a hard requirement in the test). It also adds a topology constraint to encourage pods to distribute themselves on the nodes.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Successful runs:
https://msazure.visualstudio.com/One/_build/results?buildId=113216481
https://msazure.visualstudio.com/One/_build/results?buildId=113216663
https://msazure.visualstudio.com/One/_build/results?buildId=113217222
https://msazure.visualstudio.com/One/_build/results?buildId=113258554
https://msazure.visualstudio.com/One/_build/results?buildId=113266372

### DIFF
--- a/test/integration/manifests/datapath/linux-deployment-ipv6.yaml
+++ b/test/integration/manifests/datapath/linux-deployment-ipv6.yaml
@@ -86,3 +86,10 @@ spec:
               periodSeconds: 5
         nodeSelector:
           kubernetes.io/os: linux
+        topologySpreadConstraints:
+        - maxSkew: 1
+          topologyKey: kubernetes.io/hostname
+          whenUnsatisfiable: ScheduleAnyway
+          labelSelector:
+            matchLabels:
+              app: "goldpinger"
--- a/test/integration/manifests/datapath/linux-deployment-ipv6.yaml
+++ b/test/integration/manifests/datapath/linux-deployment-ipv6.yaml
@@ -4,7 +4,7 @@ metadata:
   name: goldpinger-deploy
   namespace: linux-datapath-test
 spec:
-  replicas: 4
+  replicas: 8
   selector:
     matchLabels:
       app: goldpinger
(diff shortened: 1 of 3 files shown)


jpayne3506 (approved):
👍 Very few scenarios still use goldpinger. Do you have a schedule for a set of automated runs to see if this was effective or not?

QxBytes:
> 👍 Very few scenarios still use goldpinger. Do you have a schedule for a set of automated runs to see if this was effective or not?

I ran several runs on the pipeline page (might need to scroll down on https://msazure.visualstudio.com/One/_build?definitionId=95007&_a=summary) (see the ones with commit message "try topology constraints to balance" and none failed with the error this intends to fix-- the only failures were for other reasons. Also looks like this issue may be present in the v1.5.x release train as well https://msazure.visualstudio.com/One/_build/results?buildId=113892380&view=logs&j=b1429627-982b-5d7c-f874-2297f1590463&t=f4ed6e9d-3ea8-54bd-90c0-ce2d41a370dc

jpayne3506:
Most dualstack runs are failing now because goldpinger pods are landing on the same node. Let's get this merged

paulyufan2 (approved):

PR #3389, opened 2025-02-04
feat: add metrics for count of known NNC and NCs

rbtr:
**Reason for Change**:

Exposes a metric for known NNC and NCs in the NNC reconciler and sync host NC loop.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/metrics.go
+++ b/cns/kubecontroller/nodenetworkconfig/metrics.go
@@ -24,6 +24,18 @@ var (
 			Help: "Unused IP count.",
 		},
 	)
+	hasNNC = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "nnc_has_nodenetworkconfig",
+			Help: "Has received a NodeNetworkConfig",
+		},
+	)
+	ncs = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "nnc_ncs",
+			Help: "Network Container count in the NodeNetworkConfig",
+		},
+	)
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


cns/kubecontroller/nodenetworkconfig/metrics.go
```diff
@@ -24,12 +24,26 @@ var (
...
 	)
+	hasNNC = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "nnc_has_nodenetworkconfig",
+			Help: "Has received a NodeNetworkConfig",
+		},
+	)
+	ncs = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "nnc_ncs",
```
timraymond, line 35: Typically, don't prometheus metrics that are counts end with  `_total`?

rbtr: Yes, but these are not counts in the way that Prometheus uses that term.

They say add `_total`
> for a unit-less accumulating count

This is an instantaneous count and may go up or down, ie not accumulating. I don't think total is appropriate. You wouldn't dx/dt (nnc_ncs) to get a rate, for example.

Their general guidance is to suffix the metric with the plural units - NC is the unit here, as is IP in the ones above.

timraymond: ah, okay. I understand now

timraymond (approved):

paulyufan2 (approved):

PR #3397, opened 2025-02-05
feat: templatize dockerfile base images and auto update

rbtr:
**Reason for Change**:

Templatize Dockerfiles so that base image updates may be automated

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


nddq:
i did not have any luck with `peter-evans/create-pull-request@v5` [here](https://github.com/microsoft/retina/pull/1161/) since making pull request from GH is disabled org-wide 
image

rbtr:
> i did not have any luck with `peter-evans/create-pull-request@v5`

well, that's disappointing. so this is not automatable.

cni/Dockerfile
```diff
@@ -1,13 +1,15 @@
+# !! AUTOGENERATED - DO NOT EDIT !!
```
timraymond, line 1: Is there any way we can reject commits that touch this by hand? I suppose this applies to auto-generated things in general, but habits are so ingrained w/r/t Dockerfile edits that it might be more of a hazard here.

rbtr: that will happen implicitly because the pipeline added here generates and then checks for changes. so if you edit the Dockerfiles directly the pipeline will clobber your changes and the tree will be dirty

timraymond (approved):

thatmattlong (approved):

paulyufan2 (approved):

PR #3400, opened 2025-02-08
fix: add iptables rules for dns in vnet scale cilium case

QxBytes:
**Reason for Change**:

In the vnet scale scenario, we use the node's ip as the primary ip instead of an ip from the pod's subnet (what happens in pod subnet mode). In cilium, we add snat rules from the cns, and base our snat on the primary ip + subnet. However, the primary ip is now the node ip, meaning we only snat packets from the node's primary ip w/ the pod subnet cidr. It should be that we also snat packets from the pod subnet's network/cidr. This change modifies the cns to grab any secondary ip from the pod subnet, combine it with the pod subnet cidr, and create an iptable rule for the pod subnet. The cidr is for the whole subnet space (applies to all static ip blocks allocated).

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
Manually tested cilium w/ updated cns and iptable rule(s) are created with pod subnet cidr. DNS traffic succeeds afterwards.

### DIFF
--- a/cns/restserver/internalapi_linux.go
+++ b/cns/restserver/internalapi_linux.go
@@ -15,6 +15,14 @@ import (
 
 const SWIFT = "SWIFT-POSTROUTING"
 
+type IPtablesProvider struct {
+	iptc iptablesClient
+}
+
+func (c *IPtablesProvider) GetIPTables() (iptablesClient, error) {
+	return goiptables.New()
+}
+
 // nolint
 func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainerRequest) (types.ResponseCode, string) {
 	service.Lock()
@@ -24,7 +32,7 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 	// in podsubnet case, ncPrimaryIP is the pod subnet's primary ip
 	// in vnet scale case, ncPrimaryIP is the node's ip
 	ncPrimaryIP, _, _ := net.ParseCIDR(req.IPConfiguration.IPSubnet.IPAddress + "/" + fmt.Sprintf("%d", req.IPConfiguration.IPSubnet.PrefixLength))
-	ipt, err := goiptables.New()
+	ipt, err := service.iptables.GetIPTables()
 	if err != nil {
 		return types.UnexpectedError, fmt.Sprintf("[Azure CNS] Error. Failed to create iptables interface : %v", err)
 	}
--- a/cns/restserver/internalapi_linux.go
+++ b/cns/restserver/internalapi_linux.go
@@ -91,5 +91,40 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 			return types.FailedToRunIPTableCmd, "[Azure CNS] failed to insert SNAT IMDS rule : " + err.Error()
 		}
 	}
+
+	// use any secondary ip + the nnc prefix length to get an iptables rule to allow dns traffic
+	// this should be idempotent if req.IPConfiguration.IPSubnet.IPAddress is an ip in the nc's subnet
+	for _, v := range req.SecondaryIPConfigs {
+		// put the ip address in standard cidr form (where we zero out the parts that change the rule)
+		_, podSubnet, _ := net.ParseCIDR(v.IPAddress + "/" + fmt.Sprintf("%d", req.IPConfiguration.IPSubnet.PrefixLength))
+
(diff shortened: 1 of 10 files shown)


nairashu (approved):

cns/restserver/internalapi_linux.go
```diff
@@ -91,5 +91,40 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 			return types.FailedToRunIPTableCmd, "[Azure CNS] failed to insert SNAT IMDS rule : " + err.Error()
 		}
 	}
+
+	// use any secondary ip + the nnc prefix length to get an iptables rule to allow dns traffic
+	// this should be idempotent if req.IPConfiguration.IPSubnet.IPAddress is an ip in the nc's subnet
+	for _, v := range req.SecondaryIPConfigs {
```
nairashu, line 97: Will this override podsubnet clusters as well. Would recommend testing these changes with a podsubnet and Overlay cluster as well such that we do not break existing SNAT rules for those.

nairashu (changes requested):
Let's replace existing SNAT rules and verify IMDS works if not we might have to update the IMDS rule as well for Vnet Scale in Cillium. Source should be coming from the secondary IPs to use the podsubnet and destination should be ncPrimaryIP for SNAT. For IMDS same source but destination is always HostPrimaryIP

cns/restserver/internalapi_linux.go
QxBytes: tested on cilium, also testing in pipelines. I believe this should be okay since the primary ip is always an ip in the pod subnet space. For the iptable rule, whether we use the primary ip or any ip in the subnet space, we should get the same network. Like 10.10.10.1/24 is the same as 10.10.10.2/24 to iptables.

cns/restserver/internalapi_linux.go
```diff
@@ -56,40 +58,50 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
...
 
-	snatUDPRuleexist, err := ipt.Exists(iptables.Nat, SWIFT, "-m", "addrtype", "!", "--dst-type", "local", "-s", ncIPNet.String(), "-d", networkutils.AzureDNS, "-p", iptables.UDP, "--dport", strconv.Itoa(iptables.DNSPort), "-j", iptables.Snat, "--to", ncPrimaryIP.String())
-	if err != nil {
-		return types.UnexpectedError, fmt.Sprintf("[Azure CNS] Error. Failed to check for existence of SNAT UDP rule : %v", err)
-	}
-	if !snatUDPRuleexist {
-		logger.Printf("[Azure CNS] Inserting SNAT UDP rule ...")
-		err = ipt.Insert(iptables.Nat, SWIFT, 1, "-m", "addrtype", "!", "--dst-type", "local", "-s", ncIPNet.String(), "-d", networkutils.AzureDNS, "-p", iptables.UDP, "--dport", strconv.Itoa(iptables.DNSPort), "-j", iptables.Snat, "--to", ncPrimaryIP.String())
+	// use any secondary ip + the nnc prefix length to get an iptables rule to allow dns and imds traffic from the pods
+	for _, v := range req.SecondaryIPConfigs {
```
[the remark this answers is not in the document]
nairashu: Something like this:

if len(req.SecondaryIPConfigs) == 0 { // Overlay NNC scenario
    // current DNS and IMDS rules
} else { //podsubnet and vnetscale scenario
    // New DNS and IMDS rules getting the subnet range from the secondary IP and using the Node IP as the destination
}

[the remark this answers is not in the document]
QxBytes: the loop only runs once, taking any of the secondary ip configs (there is a break at the end). In overlay it was mentioned that we don't run this code path. Tested and noticed the cns config for overlay also has adding in the snat as set to "false".

cns/service/main.go
```diff
@@ -747,7 +747,7 @@ func main() {
 		Logger:     logger.Log,
 	}
 
-	httpRemoteRestService, err := restserver.NewHTTPRestService(&config, wsclient, &wsProxy, nmaClient,
+	httpRemoteRestService, err := restserver.NewHTTPRestService(&config, wsclient, &wsProxy, &restserver.IPtablesProvider{}, nmaClient,
```
tamilmani1989, line 750: replacing nmaClient with iptables? what nmaClient intialized with?

tamilmani1989: nvm

tamilmani1989 (approved):

nairashu (approved):

PR #3431, opened 2025-02-19
ci: split dualstack and overlay e2e tests by operating system

QxBytes:
**Reason for Change**:

Splits dualstack and overlay E2E testing stages into stages dedicated to each operating system.
Dualstack E2E --> Dualstack Windows E2E and Dualstack Linux E2E
Overlay E2E --> Overlay Windows E2E and Overlay Linux E2E
Increases load test timeout from 30m to 40m due to timeouts (even though it seemed like the test was progressing)

**Issue Fixed**:

Dualstack and Overlay E2E stages take the longest to run in the pipeline and are the most likely to fail. These stages will fail faster if run in parallel.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Green run: https://msazure.visualstudio.com/One/_build/results?buildId=115530896&view=results

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -772,7 +772,7 @@ test-load: ## run all load tests
 	AZURE_IPAM_VERSION=$(AZURE_IPAM_VERSION) \
 		CNI_VERSION=$(CNI_VERSION)
 		CNS_VERSION=$(CNS_VERSION) \
-		go test -timeout 30m -race -tags=load ./test/integration/load... -v
+		go test -timeout 40m -race -tags=load ./test/integration/load... -v
 
 test-validate-state:
 	cd test/integration/load && go test -mod=readonly -count=1 -timeout 30m -tags load --skip 'TestE2E*' -run ^TestValidateState
--- a/hack/aks/Makefile
+++ b/hack/aks/Makefile
@@ -389,7 +389,7 @@ dualstack-overlay-up: rg-up overlay-net-up ## Brings up an dualstack Overlay clu
 		--yes
 	@$(MAKE) set-kubeconf
 
-dualstack-overlay-byocni-up: rg-up overlay-net-up ## Brings up an dualstack Overlay BYO CNI cluster
+linux-dualstack-overlay-byocni-up: rg-up overlay-net-up ## Brings up an dualstack Overlay BYO CNI cluster
 	$(AZCLI) aks create -n $(CLUSTER) -g $(GROUP) -l $(REGION) \
 		--auto-upgrade-channel $(AUTOUPGRADE) \
 		--node-os-upgrade-channel $(NODEUPGRADE) \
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -772,7 +772,7 @@ test-load: ## run all load tests
 	AZURE_IPAM_VERSION=$(AZURE_IPAM_VERSION) \
 		CNI_VERSION=$(CNI_VERSION)
 		CNS_VERSION=$(CNS_VERSION) \
-		go test -timeout 30m -race -tags=load ./test/integration/load... -v
+		go test -timeout 40m -race -tags=load ./test/integration/load... -v
```
jpayne3506, line 775: Were we seeing timeouts in Windows case?

hack/aks/Makefile
```diff
@@ -114,7 +114,10 @@ nodesubnet-byocni-nokubeproxy-up: rg-up overlay-net-up ## Brings up an NodeSubne
 		--yes
 	@$(MAKE) set-kubeconf
 
-overlay-byocni-up: rg-up overlay-net-up ## Brings up an Overlay BYO CNI cluster
+# alias for linux-overlay-byocni-up
+overlay-byocni-up: linux-overlay-byocni-up
```
jpayne3506, line 118: I am in favor of leaving the makefile commands as they were and then use them an alias for windows. So, overlay-byocni-up is alias'd for windows-overlay-byocni-up without the need for the linux-overlay-byocni-up

Also, this conflicts with using make conditionals to bring up windows nodepools. Let's make it uniform throughout. Maybe this effort should be a seperate PR..?

jpayne3506 (changes requested):
Core .yaml changes lgtm. I like the makefile changes, but let's keep our usage consistent throughout.

hack/scripts/wait-cluster-update.sh
```diff
@@ -0,0 +1,16 @@
+# wait for cluster to update
```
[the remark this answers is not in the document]
QxBytes: this was to fix that issue a while back where if you tried to update the cluster with adding a windows nodepool while the cluster was updating, it would fail (logic moved from pipeline file to this script)-- this script just waits for the cluster to finish updating if it is updating before we add the windows pool

hack/aks/Makefile
QxBytes: do you mean changing line 118 to like: `overlay-byocni-up: windows-overlay-byocni-up` -- I've only aliased it to the above to maintain the existing behavior (which does not include windows nodes) but this can be changed.

for the conditionals being used to bring up windows nodepools, we already have make targets which create an os-specific cluster: https://github.com/Azure/azure-container-networking/pull/3431/files#diff-57046f2686ea559e17b63e0715ed093f51f3f557b4fcb783512962a45f24bc7aL345 (linux-cniv1-up...)

jpayne3506 (approved):
Plan to clean up the makefile in a follow-on PR?

behzad-mir (approved):

PR #3437, opened 2025-02-20
feat: cns logger v2 [1/2]

rbtr:
Adds v2 logger package for CNS. [1/2]

The v2 logger is built with zap and offers structured logging, derived contexts, multiple output targets and formats including stdout, files, ETW, and AppInsights.
The logger constructor is opinionated and will initialize targets and formats fully based on what is supported per-platform. The logger is intended to be used directly.

This pt. 1 change is strictly the addition of the v2 logger package machinery.

### DIFF
--- a/cns/logger/v2/config.go
+++ b/cns/logger/v2/config.go
@@ -2,9 +2,9 @@ package logger
 
 import (
 	"encoding/json"
-	"time"
 
 	loggerv1 "github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/Azure/azure-container-networking/internal/time"
 	"github.com/pkg/errors"
 	"go.uber.org/zap/zapcore"
 )
@@ -40,4 +40,35 @@ func (c *Config) UnmarshalJSON(data []byte) error {
 	}
 	c.level = lvl
 	return nil
+}
+
+// Normalize checks the Config for missing or illegal values and sets them
+// to defaults if appropriate.
+func (c *Config) Normalize() {
+	if c.File != nil {
+		if c.File.Filepath == "" {
+			c.File.Filepath = defaultFilePath
+		}
+		if c.File.MaxBackups == 0 {
+			c.File.MaxBackups = defaultMaxBackups
+		}
+		if c.File.MaxSize == 0 {
+			c.File.MaxSize = defaultMaxSize
+		}
+	}
+	if c.AppInsights != nil {
+		if c.AppInsights.IKey == "" {
+			c.AppInsights.IKey = defaultIKey
+		}
+		if c.AppInsights.GracePeriod.Duration == 0 {
+			c.AppInsights.GracePeriod.Duration = defaultGracePeriod
+		}
+		if c.AppInsights.MaxBatchInterval.Duration == 0 {
+			c.AppInsights.MaxBatchInterval.Duration = defaultMaxBatchInterval
+		}
+		if c.AppInsights.MaxBatchSize == 0 {
+			c.AppInsights.MaxBatchSize = defaultMaxBatchSize
+		}
+	}
+	c.normalize()
 }
--- a/cns/logger/v2/config.go
+++ b/cns/logger/v2/config.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	loggerv1 "github.com/Azure/azure-container-networking/cns/logger"
-	cores "github.com/Azure/azure-container-networking/cns/logger/v2/cores"
 	"github.com/pkg/errors"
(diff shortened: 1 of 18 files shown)


rbtr:
The ETW changes will drop out as #3433 merges.

cns/logger/v2/logger.go
```diff
@@ -0,0 +1,44 @@
...
+	closer = append(closer, aiCloser)
+	if err != nil {
+		return nil, closer.Close, err //nolint:wrapcheck // it's an internal pkg
+	}
+	platformCore, platformCloser, err := platformCore(cfg)
+	closer = append(closer, platformCloser)
+	if err != nil {
+		return nil, closer.Close, err
+	}
+	core := zapcore.NewTee(stdoutCore, fileCore, aiCore, platformCore)
```
timraymond, line 42: It might be nice to have this be flexible instead of requiring all three.

cns/logger/v2/config.go
```diff
@@ -0,0 +1,50 @@
...
+	Level       string                  `json:"level"`
+	level       zapcore.Level           `json:"-"`
+	AppInsights cores.AppInsightsConfig `json:"appInsights"`
+	File        cores.FileConfig        `json:"file"`
+}
+
+// UnmarshalJSON implements json.Unmarshaler for the Config.
+// It only differs from the default by parsing the
+// Level string into a zapcore.Level and setting the level field.
+func (c *Config) UnmarshalJSON(data []byte) error {
```
rbtr, line 36: needs UTs

cns/logger/v2/cores/file.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+type FileConfig struct {
+	Filepath   string        `json:"filepath"`
+	Level      string        `json:"level"`
+	level      zapcore.Level `json:"-"`
```
rbtr, line 15: these won't be _Marshalled_ correctly since I didn't write a custom Marshaller, but would we ever be going from struct -> text for Config?

cns/logger/v2/logger.go
rbtr: I'm trying to balance between it being customizable, and it being opinionated, since it is the core CNS logger and I want to to generally behave in a consistent way. Are there scenarios where this should be variable?

rbtr: Tried this. I think I even managed to get it to where knowledge of the platform specific bits doesn't escape this package (even to the global Config struct). The tradeoff on that is that there is no defaulting on the config - it will run only exactly what exactly how tell it in the config file, and zero values are legal.

timraymond: @rbtr awesome :) To answer your earlier question, Jackie and I are currently working on a proposal to rip AI out, so the sink config may change. The flexibility is appreciated.

cns/logger/v2/config.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		if c.AppInsights.GracePeriod.Duration == 0 {
+			c.AppInsights.GracePeriod.Duration = defaultGracePeriod
+		}
+		if c.AppInsights.MaxBatchInterval.Duration == 0 {
+			c.AppInsights.MaxBatchInterval.Duration = defaultMaxBatchInterval
+		}
+		if c.AppInsights.MaxBatchSize == 0 {
+			c.AppInsights.MaxBatchSize = defaultMaxBatchSize
+		}
```
jpayne3506, line 71: Check for negative values as well?

cns/logger/v2/cores/file_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestFileConfig_UnmarshalJSON(t *testing.T) {
+	tests := []struct {
+		name    string
+		have    []byte
+		want    *FileConfig
+		wantErr bool
+	}{
+		{
+			name: "valid",
```
jpayne3506, line 19: Could add UT that checks for negative value going to default here as well

jpayne3506 (approved):

cns/logger/v2/config.go
rbtr: what I'm really doing here is checking for _default_ values (meaning the user omitted them from the config) and setting a reasonable default. If they intentionally put a negative value and get unreasonable durations, that's fine with me.

paulyufan2 (approved):

QxBytes (approved):

PR #3438, opened 2025-02-21
feat: cns logger v2 [2/2]

rbtr:
Adds v2 logger package for CNS. [2/2]

The v2 logger is built with zap and offers structured logging, derived contexts, multiple output targets and formats including stdout, files, ETW, and AppInsights.
The logger constructor is opinionated and will initialize targets and formats fully based on what is supported per-platform. The logger is intended to be used directly.

This pt. 2 change adds a migration path to the v2 logger via CNS config. `EnableLoggerV2=true` replaces the global logger during CNS init with a shimmed v2 logger.

### DIFF
--- a/cns/logger/v2/cores/ai.go
+++ b/cns/logger/v2/cores/ai.go
@@ -62,6 +62,6 @@ func ApplicationInsightsCore(cfg *AppInsightsConfig) (zapcore.Core, func(), erro
 	core := zapai.NewCore(cfg.level, sink)
 	core = core.WithFieldMappers(zapai.DefaultMappers)
 	// add normalized fields for the built-in AI Tags
-	// TODO(rbtr): move to the caller
+
 	return core.With(cfg.Fields), aiclose, nil
 }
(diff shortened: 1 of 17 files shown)


rbtr:
Changes in #3433 and #3437 will drop out as those merge

cns/service/main.go
```diff
@@ -629,13 +628,18 @@ func main() {
 		}
 	}
 
-	// configure zap logger
-	zconfig := zap.NewProductionConfig()
-	zconfig.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
-	if z, err = zconfig.Build(); err != nil {
+	// build the zap logger
+	z, c, err := loggerv2.New(&loggerv2.Config{})
```
rbtr, line 632: need to populate this config

cns/logger/v2/cores/ai.go
```diff
@@ -0,0 +1,58 @@
...
+		zap.String("version", "v0.0.0"),
+		zap.String("account", "SubscriptionID"),
+		zap.String("anonymous_user_id", "VMName"),
+		zap.String("session_id", "VMID"),
+		zap.String("AppName", "name"),
+		zap.String("Region", "Location"),
+		zap.String("ResourceGroup", "ResourceGroupName"),
+		zap.String("VMSize", "VMSize"),
+		zap.String("OSVersion", "OSVersion"),
+		zap.String("VMID", "VMID"),
```
rbtr, line 56: need to push these in from caller?

jpayne3506 (approved):

PR #3449, opened 2025-02-24
Expose GET NC list API in CNS client

smittal22:
**Reason for Change**:
Expose GET NC list API in CNS client. DNC will call this endpoint to fetch the list of NCs programmed on the node.
In the backend, CNS calls NMAgent's _http://%s/machine/plugins/?comp=nmagent&type=NetworkManagement/interfaces/api-version/%s_ endpoint

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -1317,6 +1317,7 @@ func (service *HTTPRestService) nmAgentNCListHandler(w http.ResponseWriter, r *h
 		if ncVersionerr != nil {
 			returnCode = types.NmAgentNCVersionListError
 			returnMessage = "[Azure-CNS] " + ncVersionerr.Error()
+			break
 		}
 
 		for _, container := range ncVersionList.Containers {
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -1306,10 +1306,10 @@ func (service *HTTPRestService) nmAgentNCListHandler(w http.ResponseWriter, r *h
 	logger.Request(service.Name, "nmAgentNCListHandler", nil)
 	var (
 		returnCode           types.ResponseCode
-		returnMessage        string
 		networkContainerList []string
 	)
 
+	returnMessage := "Successfully fetched NC list from NMAgent"
 	ctx := r.Context()
 	switch r.Method {
 	case http.MethodGet:
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -1300,3 +1300,39 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
 		})
 	}
 }
+
+// This function is used to query all NCs on a node from NMAgent
+func (service *HTTPRestService) nmAgentNCListHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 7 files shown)


cns/restserver/api.go
```diff
@@ -1300,3 +1300,39 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
...
+
+	returnMessage := "Successfully fetched NC list from NMAgent"
+	ctx := r.Context()
+	switch r.Method {
+	case http.MethodGet:
+		ncVersionList, ncVersionerr := service.nma.GetNCVersionList(ctx)
+		if ncVersionerr != nil {
+			returnCode = types.NmAgentNCVersionListError
+			returnMessage = "[Azure-CNS] " + ncVersionerr.Error()
+		}
```
msvik, line 1320: We should not iterate the loop below if we see this error.

cns/restserver/api_test.go
```diff
@@ -1302,6 +1302,31 @@ func TestNmAgentSupportedApisHandler(t *testing.T) {
 	fmt.Printf("nmAgentSupportedApisHandler Responded with %+v\n", nmAgentSupportedApisResponse)
 }
 
+func TestNMAgentNCListHandler(t *testing.T) {
```
msvik, line 1305: I think we should enhance this test by passing in the JSON body with NC IDs, then decoding the JSON body successfully and matching the NC ID.

smittal22: Added. I created an NCID with a random guid and mocked NMA's response. After I call CNS' endpoint of the new API, I receive the same NC ID in the NC list.

cns/restserver/restserver.go
```diff
@@ -292,6 +292,7 @@ func (service *HTTPRestService) Init(config *common.ServiceConfig) error {
 	listener.AddHandler(cns.NetworkContainersURLPath, service.getOrRefreshNetworkContainers)
 	listener.AddHandler(cns.GetHomeAz, service.getHomeAz)
 	listener.AddHandler(cns.EndpointPath, service.EndpointHandlerAPI)
+	listener.AddHandler(cns.GetNCList, service.nmAgentNCListHandler)
```
msvik, line 295: Evan had some sort of concern about exposing APIs to the AKS side (if it is not needed) and that's why I had added the below statement where we add API only for direct channel mode. Should we do the same here?

smittal22: I made the requested change after going through your conversation with Evan. Thanks

smittal22:
CNS log when DNC calls this endpoint:
```
2025/02/26 20:08:26 [6596] [azure-cns] Received string nmAgentNCListHandler.
2025/02/26 20:08:26 [6596] [azure-cns] Sent *cns.NCListResponse &{Response:{ReturnCode:Success Message:Successfully fetched NC list from NMAgent} NCList:[86fd8b75-94b7-23d8-6ca4-6821b082111f]}.
```

paulyufan2 (approved):

cns/restserver/restserver.go
rbtr: thanks for remembering this

PR #3457, opened 2025-02-26
ci: add vnet scale cilium pipeline stage

QxBytes:
**Reason for Change**:

Adds cilium podsubnet vnet scale to the pipeline for e2e testing of cns and related components. Also updates the makefile to explicitly include pod allocation mode static block for targets that are intended to use vnet scale mode. Currently there seem to be makefile targets that are meant to create vnet scale clusters but upon creation and examination of their nncs, the mode only mentions 
```
kubernetes.azure.com/podnetwork-subnet=podnet
kubernetes.azure.com/podnetwork-type=vnet
```
We expect 
```
kubernetes.azure.com/podnetwork-subnet=podnet
kubernetes.azure.com/podnetwork-type=vnetblock
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Will fail until relevant backend changes go in for creating vnet scale clusters with network plugin none
Done: https://msazure.visualstudio.com/One/_build/results?buildId=120192520&view=results

### DIFF
--- a/hack/aks/Makefile
+++ b/hack/aks/Makefile
@@ -184,7 +184,7 @@ overlay-up: rg-up overlay-net-up ## Brings up an Overlay AzCNI cluster
 		--yes
 	@$(MAKE) set-kubeconf
 
-swift-byocni-up: rg-up swift-net-up ## Bring up a SWIFT BYO CNI cluster
+swift-byocni-up: rg-up swift-net-up ## Bring up a SWIFT (Podsubnet) BYO CNI cluster
 	$(AZCLI) aks create -n $(CLUSTER) -g $(GROUP) -l $(REGION) \
 		--auto-upgrade-channel $(AUTOUPGRADE) \
 		--node-os-upgrade-channel $(NODEUPGRADE) \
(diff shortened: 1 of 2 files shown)


hack/aks/Makefile
```diff
@@ -307,6 +308,7 @@ vnetscale-swift-byocni-nokubeproxy-up: rg-up vnetscale-swift-net-up ## Bring up
 		--no-ssh-key \
 		--os-sku $(OS_SKU) \
 		--kube-proxy-config $(KUBE_PROXY_JSON_PATH) \
+		--pod-ip-allocation-mode StaticBlock \
```
nairashu, line 311: Does this work? Have you attempted to create a Static Block cluster with --network-plugin being none. IIRC this would throw an AKS-RP exception previously.

nairashu (approved):

hack/aks/Makefile
QxBytes: yes, pipeline run is also successful: https://msazure.visualstudio.com/One/_build/results?buildId=120192520&view=results

PR #3465, opened 2025-03-01
chore: start breaking up CNS main

rbtr:
no code changes, only rearranging.
starting to break up the huge spaghetti that is `main`


nddq (dismissed):
afaik there is no code changes, right? also I was wondering if moving main out of cns/service would be an option down the line 🤔

rbtr:
> afaik there is no code changes, right? 

No code changes, only copy/pasting.

> also I was wondering if moving main out of cns/service would be an option down the line 🤔

eventually - I'm aiming for cobra. But it's very far down the line.

cns/service/main.go
```diff
@@ -1,6 +1,4 @@
-// Copyright 2017 Microsoft. All rights reserved.
-// MIT License
-
+// nolint // it's not worth it
```
pjohnst5: 😆+1 haha

cns/service/version.go
```diff
@@ -0,0 +1,12 @@
+package main
```
pjohnst5, line 4: Yes this makes a lot more sense

cns/service/main.go
```diff
@@ -380,12 +340,6 @@ func init() {
 	}()
 }
 
-// Prints description and version information.
-func printVersion() {
-	fmt.Printf("Azure Container Network Service\n")
```
pjohnst5, line 384: I'm just wondering if up above, it makes sense also to tear out `var args = acn.ArgumentList{..}` and also put it into a file like `args.go` or something

cns/service/main.go
```diff
@@ -380,12 +340,6 @@ func init() {
...
-	fmt.Printf("Version %v\n", version)
```
pjohnst5, line 385: I think we should also take out `type NodeInterrogator` ,`func registerNode()` and `func sendRegisterNodeRequest()` and put them into a file called `managed.go`, or something, since they are only used if cns is in "managed" mode it seems (not sure really what that is but it could also make this file smaller)

cns/service/main.go
```diff
@@ -380,12 +340,6 @@ func init() {
...
-}
```
pjohnst5, line 386: Similarly, we could take out `func getPodInfoByIPProvider()` and put it in `cns/restserver/nodesubnet.go` because there's already a function in there `func InitializeNodeSubnet()` that also gets called if cns is in `AzureHost` mode

cns/service/main.go
```diff
@@ -866,7 +820,7 @@ func main() {
 
 		logger.Printf("Set GlobalPodInfoScheme %v (InitializeFromCNI=%t)", cns.GlobalPodInfoScheme, cnsconfig.InitializeFromCNI)
 
-		err = InitializeCRDState(rootCtx, httpRemoteRestService, cnsconfig)
+		err = initializeCRDState(rootCtx, httpRemoteRestService, cnsconfig)
```
pjohnst5: Should we also move `func InitializeMultiTenantController()` into `aks.go`?

cns/service/main.go
```diff
@@ -117,25 +90,12 @@ const (
 	initialIBNICCount                = 0
 )
 
-type cniConflistScenario string
```
pjohnst5, line 120: Alright, just a few more things we may want to move out of here and into aks.go
the const:
`defaultNodeInfoCRDPollInterval` (maybe that doesn't really warrant a move)

The function:
`func pollNodeInfoCRDAndUpdatePlugin` (just the definition of the function not the function call)

pjohnst5 (commented):
Yeah I think once these are also moved then we are good to go:
`aks.go`
- `func InitializeMultiTenantController()`
- `func pollNodeInfoCRDAndUpdatePlugin()`

`managed.go` ?
- `type NodeInterrogator`
- `func registerNode()`
- `func sendRegisterNodeRequest()`
- These are only used if cns is `Managed` channel mode

~~`cns/restserver/nodesubnet.go`~~
~~`func getPodInfoByIPProvider()` could go here, because there's another function `func InitalizeNodeSubnet()` that is already there, and is only called when channel mode is `AzureHost`~~
This actually results in a cycle

And then this one is probably a stretch, should we make an `args.go` for the `var args = acn.ArgumentList{..}`?

Super nice to be doing this! Thanks Evan!

cns/service/main.go
pjohnst5: This actually results in a cycle, nvm
image

pjohnst5:
Hi @rbtr , I took a stab at this in #4038 (since it needed a rebase) and added a few more moves
The changes you made are first, and then I added 4 more changes after that
rbtr
pjohnst5

PR #3467, opened 2025-03-03
chore: update to go1.24

rbtr:
notably, this kills `hack/tools` in favor of the [new tools directive in the go.mod](https://go.dev/blog/go1.24#tool-improvements).
Go 1.24 now has FIPS certification of the 1P crypto, but we are currently still required to use MSFT Go...


azure-ipam/Dockerfile
```diff
@@ -3,8 +3,8 @@ ARG DROPGZ_VERSION=v0.0.12
 ARG OS_VERSION
 ARG OS
 
-# skopeo inspect docker://mcr.microsoft.com/oss/go/microsoft/golang:1.23.2-cbl-mariner2.0 --format "{{.Name}}@{{.Digest}}"
-FROM --platform=linux/${ARCH} mcr.microsoft.com/oss/go/microsoft/golang@sha256:f8613198423d5cb702961f1547f9cb061f8da1c6ca9ce8da4824eb47db663cd7 AS go
+# skopeo inspect docker://mcr.microsoft.com/oss/go/microsoft/golang:1.24-cbl-mariner2.0 --format "{{.Name}}@{{.Digest}}"
+FROM --platform=linux/${ARCH} mcr.microsoft.com/oss/go/microsoft/golang@sha256:15c9b9b8449f55446243ce20c5d3808cc18625d0b358d70aaad402fb73c0766f AS go
```
jpayne3506, line 7: This image tag will be constantly overwritten. Using the sha here prevents us from pulling the newer versions which is counter to what I believe this is trying to do. New sha is `605d0a6f05734845927f450a62e081bd19c03dbe1fdbf993cfedc8506c9192b4`


Also, we want reproducible and expected builds. Why not specify `1.24.X` image tag?

go.mod
```diff
@@ -139,10 +137,162 @@ require (
 )
 
 require (
+	4d63.com/gocheckcompilerdirectives v1.2.1 // indirect
+	4d63.com/gochecknoglobals v0.2.2 // indirect
```
jpayne3506, line 141: What is bringing in `4d63.com/*` packages..?

azure-ipam/Dockerfile
nddq: > Using the sha here prevents us from pulling the newer versions

Yes, this is the goal with sha pinning. Pulling images via sha will allow us to control the exact image that we want to use across builds, hence reproducible and expected 🙂

go.mod
nddq: might have come with all the `go get -tool`

.github/workflows/golangci.yaml
```diff
@@ -15,19 +15,18 @@ jobs:
...
       fail-fast: false
       matrix:
-        go-version: ['1.22.x', '1.23.x']
         os: [ubuntu-latest, windows-latest]
     name: Lint
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/setup-go@v5
       with:
-        go-version: ${{ matrix.go-version }}
```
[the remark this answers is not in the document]
nddq: i think we can do `go-version-file: go.mod` here?

azure-ipam/Dockerfile
jpayne3506: I want to be able to answer this question. What is the expected patch version? 

I want to be able to update this later and know when I pull the sha for 'golang:1.24-cbl-mariner2.0' it is a certain version. Stdlib CVE updates have taken up residence in my mind.

go.mod
jpayne3506: I need to read up more on 1.24. Curious if this change makes it where tools are added to our binaries...

huntergregory (dismissed):
can we run npm conformance?

go.mod
rbtr: it does not

.github/workflows/crdgen.yaml
```diff
@@ -12,19 +12,15 @@ on:
       - checks_requested
 jobs:
   crdgen:
-    strategy:
-      matrix:
-        go-version: ['1.22', '1.23']
-        os: [ubuntu-latest]
     name: CRDs are Generated
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
rbtr: that's what I want. the required status checks will be updated to match the new workflow names

.github/workflows/golangci.yaml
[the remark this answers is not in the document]
rbtr: > This will break all PRs every go minor version bump

I'm not sure that's true, but I'm amenable to the `go-version-file` change anyway

azure-ipam/Dockerfile
rbtr: stdlib CVE updates come in new patch versions. Do you want to have to update the hint with that new patch version, or just rerun rerun it and copy/paste the SHA?
See also #3397

jpayne3506: Thought about this more and if I really wanted detailed information on the patch version being used, I would only care after trivy has informed me I need to update.

PR #3492, opened 2025-03-10
fix: add vnetscale staticblock changes to aks makefile

rbtr:
The `vnetscale-*` targets in the AKS Makefile did not have the flag which indicates `StaticBlock` allocation mode and therefore build normal dynamic podsubnet clusters instead.

Incidentally this means that the VnetScale E2E was not using a VnetScale cluster and this will fix that.

### DIFF
--- a/hack/aks/Makefile
+++ b/hack/aks/Makefile
@@ -274,9 +274,6 @@ swiftv2-dummy-cluster-up: rg-up swift-net-up ## Bring up a SWIFT AzCNI cluster
 		--yes
 	@$(MAKE) set-kubeconf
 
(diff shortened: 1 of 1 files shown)


nairashu (approved):

QxBytes:
The vnet scale cluster in the e2e won't create until byocni + vnet scale is enabled in aks rp [link to failure](https://msazure.visualstudio.com/One/_build/results?buildId=117773881&view=logs&j=bcdfe2cb-2619-502e-a393-045f96c9c863&t=a972fadd-e986-5293-75f8-e51011b79840&l=121)

jpayne3506:
Blocking CI/CD. Wait for next AKS-RP release before adding to merge queue.

PR #3498, opened 2025-03-12
fix: don't restart HNS if the ARP regkey is not changed

rbtr:
don't restart HNS when setting SDNRemoteArpMacAddress (or at all, it's fragile).
reverts behavior changed in https://github.com/Azure/azure-container-networking/pull/3343 (timeout and retry if HNS is not responding) and https://github.com/Azure/azure-container-networking/pull/2993 (restart HNS after the regkey is set unconditionally)

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -799,14 +799,11 @@ func main() {
 	}
 
 	// Setting the remote ARP MAC address to 12-34-56-78-9a-bc on windows for external traffic if HNS is enabled
-	arpCtx, arpCtxCancel := context.WithTimeout(rootCtx, 30*time.Second)
-	err = platform.SetSdnRemoteArpMacAddress(arpCtx)
+	err = platform.SetSdnRemoteArpMacAddress(rootCtx)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cns/service/main.go
```diff
@@ -799,14 +799,11 @@ func main() {
 	}
 
 	// Setting the remote ARP MAC address to 12-34-56-78-9a-bc on windows for external traffic if HNS is enabled
-	arpCtx, arpCtxCancel := context.WithTimeout(rootCtx, 30*time.Second)
-	err = platform.SetSdnRemoteArpMacAddress(arpCtx)
+	err = platform.SetSdnRemoteArpMacAddress(rootCtx)
```
thatmattlong, line 802: just curious why drop the timeout?

thatmattlong (approved):

cns/service/main.go
rbtr: we didn't previously timeout and this is just fully reverting any behavioral change

ashvindeodhar (approved):

PR #3529, opened 2025-03-25
fix: carefully retry restarting HNS if it hangs

rbtr:
#3498 reverted two functional changes to the CNS init HNS interaction:
1) always restart HNS after attempting to set the regkey
2) retry restarting HNS if it hangs

It turns out that HNS hangs are common enough that they notably impact CNS on Windows becoming ready and are preventing that revert from passing tests and rolling out. Thus, it is not safe for CNS to block on HNS restarting - if it hangs, CNS never becomes ready, meaning the Node never becomes Ready.

This is a retake on (2) while keeping (1) reverted - we will _only_ try to restart HNS if the regkey is **not already set**. But when we do try to restart it, we will retry this instead of hanging forever.

### DIFF
--- a/platform/os_windows.go
+++ b/platform/os_windows.go
@@ -17,6 +17,7 @@ import (
 	"github.com/Azure/azure-container-networking/log"
 	"github.com/Azure/azure-container-networking/platform/windows/adapter"
 	"github.com/Azure/azure-container-networking/platform/windows/adapter/mellanox"
+	"github.com/avast/retry-go/v4"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"golang.org/x/sys/windows"
@@ -302,32 +303,61 @@ func restartHNS(ctx context.Context) error {
 	}
 	defer service.Close()
 	// Stop the service
-	_, err = service.Control(svc.Stop)
-	if err != nil {
-		return errors.Wrap(err, "could not stop service")
+	log.Printf("Stopping HNS service")
+	_ = retry.Do(
+		tryStopServiceFn(ctx, service),
+		retry.UntilSucceeded(),
+		retry.Context(ctx),
+	)
+	// Start the service again
+	log.Printf("Starting HNS service")
+	if err := service.Start(); err != nil {
+		return errors.Wrap(err, "could not start service")
 	}
-	// Wait for the service to stop
-	ticker := time.NewTicker(500 * time.Millisecond) //nolint:gomnd // 500ms
-	defer ticker.Stop()
-	for { // hacky cancellable do-while
+	log.Printf("HNS service started")
+	return nil
+}
+
+type managedService interface {
+	Control(control svc.Cmd) (svc.Status, error)
+	Query() (svc.Status, error)
+}
+
+func tryStopServiceFn(ctx context.Context, service managedService) func() error {
+	return func() error {
 		status, err := service.Query()
 		if err != nil {
 			return errors.Wrap(err, "could not query service status")
 		}
+		// If the service is already stopped, no need to stop it again
 		if status.State == svc.Stopped {
-			break
+			return nil
 		}
-		select {
-		case <-ctx.Done():
-			return errors.New("context cancelled")
-		case <-ticker.C:
+		_, err = service.Control(svc.Stop)
+		if err != nil {
+			return errors.Wrap(err, "could not stop service")
 		}
+		// Wait for the service to stop
+		ticker := time.NewTicker(500 * time.Millisecond) //nolint:gomnd // 500ms
+		defer ticker.Stop()
+		for {
+			log.Printf("Waiting for HNS service to stop")
+			status, err := service.Query()
+			if err != nil {
+				return errors.Wrap(err, "could not query service status")
+			}
+			if status.State == svc.Stopped {
+				log.Printf("HNS service stopped")
+				break
+			}
+			select {
+			case <-ctx.Done():
+				return errors.New("context cancelled")
+			case <-ticker.C:
+			}
+		}
+		return nil
 	}
-	// Start the service again
-	if err := service.Start(); err != nil {
-		return errors.Wrap(err, "could not start service")
-	}
-	return nil
 }
 
 func HasMellanoxAdapter() bool {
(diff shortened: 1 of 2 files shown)


platform/os_windows.go
```diff
@@ -302,32 +303,61 @@ func restartHNS(ctx context.Context) error {
...
+}
+
+func tryStopServiceFn(ctx context.Context, service managedService) func() error {
+	return func() error {
 		status, err := service.Query()
 		if err != nil {
 			return errors.Wrap(err, "could not query service status")
 		}
+		// If the service is already stopped, no need to stop it again
 		if status.State == svc.Stopped {
```
paulgmiller, line 333: theres no stopping state right?

platform/os_windows.go
```diff
@@ -302,32 +303,61 @@ func restartHNS(ctx context.Context) error {
...
+	// Start the service again
+	log.Printf("Starting HNS service")
+	if err := service.Start(); err != nil {
+		return errors.Wrap(err, "could not start service")
 	}
-	// Wait for the service to stop
-	ticker := time.NewTicker(500 * time.Millisecond) //nolint:gomnd // 500ms
-	defer ticker.Stop()
-	for { // hacky cancellable do-while
+	log.Printf("HNS service started")
```
paulgmiller, line 317: do we know this? Should we query the service or anything else?

platform/os_windows.go
```diff
@@ -302,32 +303,61 @@ func restartHNS(ctx context.Context) error {
...
+		// If the service is already stopped, no need to stop it again
 		if status.State == svc.Stopped {
-			break
+			return nil
 		}
-		select {
-		case <-ctx.Done():
-			return errors.New("context cancelled")
-		case <-ticker.C:
+		_, err = service.Control(svc.Stop)
```
ashvindeodhar, line 336: is first return value the new status? Is this async call where you would need to keep querying or does the first return value give us the new status?

platform/os_windows.go
rbtr: there is stop pending, added

platform/os_windows.go
rbtr: It's not clear if that is guaranteed, so I did not assume so, and instead query afterward to be sure.

>// Control sends state change request c to the service s. It returns the most
// recent status the service reported to the service control manager, and an
// error if the state change request was not accepted.
// Note that the returned service status is only set if the status change
// request succeeded, or if it failed with error ERROR_INVALID_SERVICE_CONTROL,
// ERROR_SERVICE_CANNOT_ACCEPT_CTRL, or ERROR_SERVICE_NOT_ACTIVE.

platform/os_windows.go
rbtr: update to retry start in the same way we retry stop

platform/os_windows.go
jackieluc: could it potentially get stuck in the stop pending state too?

ashvindeodhar (dismissed):

platform/os_windows.go
rbtr: I think it's possible, but the only thing we can safely do about that is wait

platform/os_windows.go
```diff
@@ -302,32 +303,99 @@ func restartHNS(ctx context.Context) error {
...
+		ticker := time.NewTicker(500 * time.Millisecond) //nolint:gomnd // 500ms
+		defer ticker.Stop()
+		for {
+			log.Printf("Waiting for service to stop")
+			status, err := service.Query()
+			if err != nil {
+				return errors.Wrap(err, "could not query service status")
+			}
+			if status.State == svc.Stopped {
+				log.Printf("service stopped")
```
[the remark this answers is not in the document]
rbtr: yeah, this is not stdlib `log`, it's acn log. Printf is the Println

platform/os_windows.go
```diff
@@ -302,32 +303,99 @@ func restartHNS(ctx context.Context) error {
...
+			_, err = service.Control(svc.Stop)
+			if err != nil {
+				return errors.Wrap(err, "could not stop service")
+			}
+		}
+		// Wait for the service to stop
+		ticker := time.NewTicker(500 * time.Millisecond) //nolint:gomnd // 500ms
+		defer ticker.Stop()
+		for {
+			log.Printf("Waiting for service to stop")
```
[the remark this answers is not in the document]
rbtr: Likely not valuable and removed. I can draw the same conclusion from a lack of further logs that I can from logging "waiting" over and over, unless I doubt that my ticker works.

platform/os_windows.go
```diff
@@ -302,32 +303,101 @@ func restartHNS(ctx context.Context) error {
 	}
 	defer service.Close()
 	// Stop the service
-	_, err = service.Control(svc.Stop)
-	if err != nil {
-		return errors.Wrap(err, "could not stop service")
+	log.Printf("Stopping HNS service")
+	_ = retry.Do(
```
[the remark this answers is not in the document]
rbtr: `retry.UntilSucceeded` means that no error can ever be returned here

platform/os_windows.go
```diff
@@ -302,32 +303,101 @@ func restartHNS(ctx context.Context) error {
...
+		tryStopServiceFn(ctx, service),
+		retry.UntilSucceeded(),
+		retry.Context(ctx),
+	)
+	// Start the service again
+	log.Printf("Starting HNS service")
+	_ = retry.Do(
```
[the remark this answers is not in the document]
rbtr: `retry.UntilSucceeded` means that no error can ever be returned here

ashvindeodhar (approved):

timraymond (approved):

platform/os_windows.go
paulgmiller, line 307: would exponential back off make sense here?  if its not comign up should we wait longer.

platform/os_windows.go
```diff
@@ -302,32 +303,101 @@ func restartHNS(ctx context.Context) error {
 	}
 	defer service.Close()
```
paulgmiller, line 304: Mostly I just want commetns about how

1) HNS should hot reload this. Restart is not an api

2) working on getting this into agentbaker

3) there  is a race if we set the reg ke then crash before restart.

paulgmiller (approved):

PR #3540, opened 2025-03-27
fix: backoff retry and timeouts in HNS restart 

rbtr:
#3529 refactored the HNS restart such that the stop/start could be retried - this adds a deadline to fail and cause that retry, and changes the retry strategy to exponential backoff instead of constant

### DIFF
--- a/platform/os_windows.go
+++ b/platform/os_windows.go
@@ -308,6 +308,7 @@ func restartHNS(ctx context.Context) error {
 		tryStopServiceFn(ctx, service),
 		retry.UntilSucceeded(),
 		retry.Context(ctx),
+		retry.DelayType(retry.BackOffDelay),
 	)
 	// Start the service again
 	log.Printf("Starting HNS service")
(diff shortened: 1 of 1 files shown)


platform/os_windows.go
```diff
@@ -391,8 +397,8 @@ func tryStopServiceFn(ctx context.Context, service managedService) func() error
 				break
 			}
 			select {
-			case <-ctx.Done():
-				return errors.Wrap(ctx.Err(), "context cancelled")
+			case <-deadline.Done():
+				return errors.Wrap(deadline.Err(), "context cancelled")
```
ashvindeodhar, line 401: nit: deadline exceeded

platform/os_windows.go
```diff
@@ -391,8 +397,8 @@ func tryStopServiceFn(ctx context.Context, service managedService) func() error
 				break
 			}
 			select {
-			case <-ctx.Done():
-				return errors.Wrap(ctx.Err(), "context cancelled")
+			case <-deadline.Done():
```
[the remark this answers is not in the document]
timraymond: The implementation of `context.DeadlineExceeded`'s `Error()` already says "context deadline exceeded", so all of this seems unnecessary. I'd say just return the error without wrapping and mute everything else that complains about it.

platform/os_windows.go
rbtr: unwapped, deadline error has suffifient context in the message already

timraymond (approved):

ashvindeodhar (approved):

PR #3543, opened 2025-03-31
ci: add fqdn with cilium local redirect policy test

QxBytes:
**Reason for Change**:

Adds a test to validate functionality of a cilium local redirect policy when a cilium network policy is applied-- for example, confirms whether certain dns requests are redirected and increment a metric when they are allowed in the network policy (and confirms that they don't increment the counter when not allowed). Updates necessary utility methods and creates a function that issues an exec on pod command but does _not_ retry on failure.

**Issue Fixed**:

N/A

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
Example run: https://msazure.visualstudio.com/One/_build/results?buildId=119685764&view=logs&j=a725b465-af6c-5b16-2e20-4e0bf1d0563e&t=d9595ca1-827d-5105-2385-c679bcc42718
Nightly: https://msazure.visualstudio.com/One/_build/results?buildId=119489437&view=results
ACN PR run (no regression): https://msazure.visualstudio.com/One/_build/results?buildId=119699714&view=results

### DIFF
--- a/test/integration/lrp/lrp_fqdn_test.go
+++ b/test/integration/lrp/lrp_fqdn_test.go
@@ -1,0 +1,96 @@
+//go:build lrp
+
+package lrp
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-container-networking/test/internal/kubernetes"
+	ciliumClientset "github.com/cilium/cilium/pkg/k8s/client/clientset/versioned"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	fqdnCNPPath    = ciliumManifestsDir + "fqdn-cnp.yaml"
+	enableFQDNFlag = "enable-l7-proxy"
+)
+
+// TestLRPFQDN tests if the local redirect policy in a cilium cluster is functioning with a
+// FQDN Cilium Network Policy. As such, enable-l7-proxy should be enabled in the config
+// The test assumes the current kubeconfig points to a cluster with cilium, cns,
+// and kube-dns already installed. The lrp feature flag should also be enabled in the cilium config
+// Resources created are automatically cleaned up
+// From the lrp folder, run: go test ./ -v -tags "lrp" -run ^TestLRPFQDN$
+func TestLRPFQDN(t *testing.T) {
+	ctx := context.Background()
+
+	selectedPod, cleanupFn := setupLRP(t, ctx)
+	defer cleanupFn()
+	require.NotNil(t, selectedPod)
+
+	cs := kubernetes.MustGetClientset()
+	config := kubernetes.MustGetRestConfig()
+	ciliumCS, err := ciliumClientset.NewForConfig(config)
+	require.NoError(t, err)
+
+	// ensure enable l7 proxy flag is enabled
+	ciliumCM, err := kubernetes.GetConfigmap(ctx, cs, kubeSystemNamespace, ciliumConfigmapName)
+	require.NoError(t, err)
+	require.Equal(t, "true", ciliumCM.Data[enableFQDNFlag], "enable-l7-proxy not set to true in cilium-config")
+
+	_, cleanupCNP := kubernetes.MustSetupCNP(ctx, ciliumCS, fqdnCNPPath)
+	defer cleanupCNP()
+
+	tests := []struct {
+		name                   string
+		command                []string
+		expectedMsgContains    string
+		expectedErrMsgContains string
+		countIncreases         bool
+	}{
+		{
+			name:                "nslookup google succeeds",
+			command:             []string{"nslookup", "www.google.com", "10.0.0.10"},
+			expectedMsgContains: "Server:",
+			countIncreases:      true,
+		},
+		{
+			name:                   "wget google succeeds",
+			command:                []string{"wget", "-O", "index.html", "www.google.com", "--timeout=5"},
+			expectedErrMsgContains: "saved",
+			countIncreases:         true,
+		},
+		{
+			name:                "nslookup bing succeeds",
+			command:             []string{"nslookup", "www.bing.com", "10.0.0.10"},
+			expectedMsgContains: "Server:",
+			countIncreases:      true,
+		},
+		{
+			name:                   "wget bing fails but dns succeeds",
+			command:                []string{"wget", "-O", "index.html", "www.bing.com", "--timeout=5"},
+			expectedErrMsgContains: "timed out",
+			countIncreases:         true,
+		},
+		{
+			name:                "nslookup example fails",
+			command:             []string{"nslookup", "www.example.com", "10.0.0.10"},
+			expectedMsgContains: "REFUSED",
+			countIncreases:      false,
+		},
+		{
+			// won't be able to nslookup, let alone query the website
+			name:                   "wget example fails",
+			command:                []string{"wget", "-O", "index.html", "www.example.com", "--timeout=5"},
+			expectedErrMsgContains: "bad address",
+			countIncreases:         false,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			testLRPCase(t, ctx, *selectedPod, tt.command, tt.expectedMsgContains, tt.expectedErrMsgContains, tt.countIncreases)
+		})
+	}
+}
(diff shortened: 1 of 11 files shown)


test/internal/kubernetes/utils.go
```diff
@@ -258,6 +258,15 @@ func MustSetupLRP(ctx context.Context, clientset *cilium.Clientset, lrpPath stri
 	}
 }
 
+func MustSetupCNP(ctx context.Context, clientset *cilium.Clientset, cnpPath string) (ciliumv2.CiliumNetworkPolicy, func()) { // nolint
+	cnp := mustParseCNP(cnpPath)
+	cnps := clientset.CiliumV2().CiliumNetworkPolicies(cnp.Namespace)
```
vipul-21, line 263: Can we rename this to `cnpClient` ? Seems like multiple cnp are being setup.

test/integration/lrp/lrp_fqdn_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	cs := kubernetes.MustGetClientset()
+	config := kubernetes.MustGetRestConfig()
+	ciliumCS, err := ciliumClientset.NewForConfig(config)
+	require.NoError(t, err)
+
+	// ensure enable l7 proxy flag is enabled
+	ciliumCM, err := kubernetes.GetConfigmap(ctx, cs, kubeSystemNamespace, ciliumConfigmapName)
+	require.NoError(t, err)
+	require.Equal(t, "true", ciliumCM.Data[enableFQDNFlag], "enable-l7-proxy not set to true in cilium-config")
```
vipul-21, line 41: nit: You can use `requre.True` here

test/integration/lrp/lrp_fqdn_test.go
```diff
@@ -0,0 +1,96 @@
...
+	tests := []struct {
+		name                   string
+		command                []string
+		expectedMsgContains    string
+		expectedErrMsgContains string
+		countIncreases         bool
+	}{
+		{
+			name:                "nslookup google succeeds",
+			command:             []string{"nslookup", "www.google.com", "10.0.0.10"},
```
vipul-21, line 55: Why do we have to specify the svc ip of kubedns ?

test/integration/lrp/lrp_test.go
```diff
@@ -153,24 +169,43 @@ func TestLRP(t *testing.T) {
...
 
-	t.Log("calling nslookup from client")
-	// nslookup to 10.0.0.10 (coredns)
-	val, err := kubernetes.ExecCmdOnPod(ctx, cs, clientDS.Namespace, selectedClientPod, clientContainer, []string{
-		"nslookup", "google.com", "10.0.0.10",
-	}, config)
-	require.NoError(t, err, string(val))
-	// can connect
-	require.Contains(t, string(val), "Server:")
+	t.Log("calling command from client")
```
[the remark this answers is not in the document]
QxBytes: yes if the command times out or otherwise fails (even if the test expects it to fail) we print out the command-- if it succeeds we don't print the command (this is the current behavior of the exec on pod command)
Ex:
```
=== RUN   TestLRPFQDN/wget_example_fails
    lrp_test.go:172: calling command from client
2025/04/01 12:23:52 Error: lrp-test-fcrtd had error command terminated with exit code 1 from command - [wget -O index.html www.example.com --timeout=5]
```

test/internal/kubernetes/utils.go
```diff
@@ -482,47 +491,58 @@ func writeToFile(dir, fileName, str string) error {
...
+		result = output
+		return err
 	}
 	retrier := retry.Retrier{Attempts: ShortRetryAttempts, Delay: RetryDelay}
 	err := retrier.Do(ctx, execCmdOnPod)
 	return result, errors.Wrapf(err, "could not execute the cmd %s on %s", cmd, podName)
 }
 
+// ExecCmdOnPodOnce runs a command on the specified pod and returns its standard output, standard error output, and error in that order
+// The command does not retry when the command fails (ex: due to timeout), unlike ExecCmdOnPod
```
vipul-21, line 504: Can't we reuse the same `ExecCmdOnPod` maybe pass in the retry variable ?

test/integration/lrp/lrp_fqdn_test.go
```diff
@@ -0,0 +1,97 @@
...
+			name:                "nslookup example fails",
+			command:             []string{"nslookup", "www.example.com", "10.0.0.10"},
+			expectedMsgContains: "REFUSED",
+			countIncreases:      false,
+		},
+		{
+			// won't be able to nslookup, let alone query the website
+			name:                   "wget example fails",
+			command:                []string{"wget", "-O", "index.html", "www.example.com", "--timeout=5"},
+			expectedErrMsgContains: "bad address",
```
vipul-21, line 87: Why bad address ? Shouldn't the dns be refused and we should see `refused` here too ?

test/integration/lrp/lrp_fqdn_test.go
```diff
@@ -0,0 +1,97 @@
...
+		name                   string
+		command                []string
+		expectedMsgContains    string
+		expectedErrMsgContains string
+		countIncreases         bool
+	}{
+		{
+			name:                "nslookup google succeeds",
+			command:             []string{"nslookup", "www.google.com", "10.0.0.10"},
+			expectedMsgContains: "Server:",
```
vipul-21, line 56: Even if there is a refused in nslookup we would still see `Server` and `Address` pointing to dns svc ip. Let's check for the cmd exit code rather than just the msg string.

test/integration/lrp/lrp_fqdn_test.go
QxBytes: Data is a string[string] so when I try to change it to require.True it doesn't compile

test/integration/lrp/lrp_fqdn_test.go
QxBytes: It shouldn't be required but I wanted to be explicit since that is the svc ip I want it to be sent to

test/integration/lrp/lrp_fqdn_test.go
QxBytes: wget gives a different error message when I tested it

test/internal/kubernetes/utils_create.go
```diff
@@ -189,6 +189,18 @@ func mustCreateCiliumLocalRedirectPolicy(ctx context.Context, lrpClient typedcil
 	}
 }
 
+func mustCreateCiliumNetworkPolicy(ctx context.Context, cnpClient typedciliumv2.CiliumNetworkPolicyInterface, cnp ciliumv2.CiliumNetworkPolicy) {
+	if err := cnpClient.Delete(ctx, cnp.Name, metav1.DeleteOptions{}); err != nil {
```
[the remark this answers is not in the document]
QxBytes: Might re-examine the other mustCreate functions at a later date since they still use .Delete( ) instead of their `mustDelete` counterparts

test/integration/lrp/lrp_test.go
```diff
@@ -153,24 +169,43 @@ func TestLRP(t *testing.T) {
...
+
+	val, errMsg, err := kubernetes.ExecCmdOnPodOnce(ctx, cs, clientPod.Namespace, clientPod.Name, clientContainer, clientCmd, config)
```
vipul-21: I think we are not checking this error ?

test/integration/lrp/lrp_fqdn_test.go
vipul-21:
```
k exec -it container2-59f64d8c8b-d9jtd -- wget www.example.com
--2025-04-01 23:12:33--  http://www.example.com/
Resolving www.example.com (www.example.com)... failed: Temporary failure in name resolution.
wget: unable to resolve host address 'www.example.com'
command terminated with exit code 4
```
Tried on my test cluster. What image are we using for the pod ?

QxBytes: busybox, I get `"wget: bad address 'www.example.com'\n"` on the error channel and nothing from stdout

test/integration/lrp/lrp_test.go
QxBytes: should be here-- I'll reorder it so its right after the exec: https://github.com/Azure/azure-container-networking/pull/3543/files#diff-c3075c56974b82a218bb780da1ac3570cf9421e3e545a1a68afb72eb76083634R180

vipul-21 (dismissed):

jpayne3506 (approved):

vipul-21 (approved):

PR #3549, opened 2025-04-01
feat: stateless cni compatibility with nodesubnet

santhoshmprabhu:
**Reason for Change**:
Stateless CNI is currently not compatible with nodesubnet, even with the changes that were recently incorporated into CNS for Cilium+Nodesubnet support. Specifically, IP configs returned from CNS in Nodesubnet mode do not populate NC-specific fields - ncGatewayIPAddress, ncSubnetPrefix, ncPrimarIP etc. The CNI today expects these values, since CNS is invoked only for Podsubnet/Overlay cases. With this PR, we have the following changes:
1. Have a configuration option to configure the CNI for nodesubnet
2. In nodesubnet mode, the CNI uses hostGateway and subnet to configure the pods
3. The conflist for Azure CNI Nodesubnet on both windows and linux additionally specifies the mode as "nodesubnet". Today, the mode is specified only for overlay. Nodesubnet uses azure-vnet-ipam as the IPAM type, and podsubnet uses azure-cns as IPAM type with mode unspecified. With this change, cniv1 clusters can continue using azure-vnet-ipam (mode is ignored), whenever we deploy stateless CNI with nodesubnet, we'll use azure-cns as the type, and nodesubnet as the mode. Making this change here allows us to leave AgentBaker CSE scripts untouched.
4. Doesn't block pod deletion if HNS network is not found. Rather, we log the event and proceed with pod deletion. Effectively, this allows deleting pods that never got an IP. This is an enhancement to help debug issues more easily.

The PR touches code that appears to have TODO items to clean up the Overlay configuration parameter. I'm planning to do that separately once I validate that the new parameter is fully rolled out.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -205,8 +205,7 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 				}
 			}
 
-			overlayMode := (invoker.ipamMode == util.V4Overlay) || (invoker.ipamMode == util.DualStackOverlay) || (invoker.ipamMode == util.Overlay)
-			if err := configureDefaultAddResult(&info, &addConfig, &addResult, overlayMode, key); err != nil {
+			if err := configureDefaultAddResult(&info, &addConfig, &addResult, key); err != nil {
 				return IPAMAddResult{}, err
 			}
 		default:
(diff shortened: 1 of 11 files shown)


behzad-mir (approved):
lgtm

network/endpoint_windows.go
```diff
@@ -531,7 +531,7 @@ func (nw *network) deleteEndpointImpl(_ netlink.NetlinkInterface, _ platform.Exe
 
 	if ep.HnsId == "" {
 		logger.Error("No HNS id found. Skip endpoint deletion", zap.Any("nicType", ep.NICType), zap.String("containerId", ep.ContainerID))
-		return fmt.Errorf("No HNS id found. Skip endpoint deletion for nicType %v, containerID %s", ep.NICType, ep.ContainerID) //nolint
+		return nil
```
behzad-mir, line 534: I think adding a comment about why we are returning nil here would be helpful

rbtr: maybe it could return an error of some known "NotFound" type, and the caller could tolerate that, instead

santhoshmprabhu: It seems we want to allow this behavior, at least according to this UT:
https://github.com/Azure/azure-container-networking/blob/20d09ab789046047e6cecb4cd1a40e99af936c2f/network/endpoint_windows_test.go#L118

@paulyufan mentioned that this UT is run manually locally, not as part of the PR pipeline. The UT fails on master, and needs this change to pass actually. Possibly `deleteEndpointImpl` drifted from intended behavior by mistake?

rbtr: hm why this isn't run in the PR pipeline? because it's in a _windows test?

santhoshmprabhu: Not sure about that actually. We do have a Windows Tests stage in the PR pipeline of course. Maybe @paulyufan could weigh in?

paulyufan2: Hi @rbtr , TestDeleteEndpointImplHnsV2WithEmptyHNSID() is part of _windows UT test cases and PR pipeline does not run them. We tried to run this test case locally and it's failed.

jpayne3506: Below contains UT(s) we run for windows in the pipeline. 
https://github.com/Azure/azure-container-networking/blob/a2a2ab89b94e68245f29118ce9c450517b8cee01/.pipelines/templates/run-unit-tests.yaml#L43

santhoshmprabhu: @rbtr, we can expand the scope for our Windows test to include this (and other) test(s), but given this info, what are your thoughts on the diff above?

rbtr: unchanged from my first comment

santhoshmprabhu: @rbtr addressed, please take a look.

cni/network/invoker_cns.go
```diff
@@ -379,32 +378,49 @@ func getRoutes(cnsRoutes []cns.Route, skipDefaultRoutes bool) ([]network.RouteIn
...
 		return errors.Wrap(err, "Unable to parse IP from response: "+info.podIPAddress+" with err %w")
 	}
 
-	ncgw := net.ParseIP(info.ncGatewayIPAddress)
-	if ncgw == nil {
-		// TODO: Remove v4overlay and dualstackoverlay options, after 'overlay' rolls out in AKS-RP
-		if !overlayMode {
+	podGateway := net.ParseIP(info.ncGatewayIPAddress)
+	// TODO: Remove v4overlay and dualstackoverlay options, after 'overlay' rolls out in AKS-RP
+	isPodSubnetMode := (ipamMode != util.V4Overlay) && (ipamMode != util.DualStackOverlay) && (ipamMode != util.Overlay) && (ipamMode != util.Nodesubnet)
```
MikeZappa87, line 410: could a function like: 

`func isPodSubnetMode(ipamMode string) bool {
    switch ipamMode {
    case util.V4Overlay, util.DualStackOverlay, util.Overlay, util.Nodesubnet:
        return false
    default:
        return true
    }
}`

Make this easier to read/test?

santhoshmprabhu: It would, but I'm planning to finish up that TODO and get rid of two of these. Subsequently, I'm thinking of making podsubnet a separate config option, at which point this check would just be `ipamMode == util.Podsubnet`

MikeZappa87: Right on. The (....) && (...) && (...) was a little odd, I didn't understand the need for the ()

santhoshmprabhu: I'm a little paranoid that way about operator precedence :D

MikeZappa87: `ipamMode != util.V4Overlay && 
ipamMode != util.DualStackOverlay && 
ipamMode != util.Overlay && 
ipamMode != util.Nodesubnet
`
or

`(ipamMode != util.V4Overlay && 
ipamMode != util.DualStackOverlay && 
ipamMode != util.Overlay && 
ipamMode != util.Nodesubnet)
`

are functionality equivalent :-) I will admit the current does stand out a little

santhoshmprabhu: removed the parentheses

cni/network/invoker_cns.go
```diff
@@ -379,32 +378,49 @@ func getRoutes(cnsRoutes []cns.Route, skipDefaultRoutes bool) ([]network.RouteIn
...
 	if ip == nil || err != nil {
 		return errors.Wrap(err, "Unable to parse IP from response: "+info.podIPAddress+" with err %w")
 	}
 
-	ncgw := net.ParseIP(info.ncGatewayIPAddress)
-	if ncgw == nil {
-		// TODO: Remove v4overlay and dualstackoverlay options, after 'overlay' rolls out in AKS-RP
-		if !overlayMode {
+	podGateway := net.ParseIP(info.ncGatewayIPAddress)
+	// TODO: Remove v4overlay and dualstackoverlay options, after 'overlay' rolls out in AKS-RP
```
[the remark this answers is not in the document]
santhoshmprabhu: image

Brand new cluster - it seems this isn't rolled out.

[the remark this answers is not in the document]
santhoshmprabhu: This isn't yet rolled out because in CNI 1.4 we don't have the unified overlay parameter. We could possibly update the CNS config map in a way that preserved v4overlay only in 1.4 and unifies the option in 1.5 and 1.6, but that has not been done likely to avoid complicating the CNS config logic. There's a separate ongoing conversation about how this needs to be handled in AgentBaker for Windows nodes.

santhoshmprabhu:
Had a chat with @tamilmani1989. He suggested to keep CNI behavior independent of IPAM modes, which means moving the handling of Nodesubnet case into CNS. Closing this PR, will open a separate one for CNS.

PR #3551, opened 2025-04-01
fix: delete cni statefile when unable to be parsed

QxBytes:
**Reason for Change**:

Sometimes in certain scenarios (usually windows), if there is a crash of the OS, null bytes may be written to the state and log file. When the CNI tries to restore the state, it is unable to read the statefile and fails. All subsequent retries will fail as the state is irrecoverable. This PR changes this behavior to delete the entire cni statefile if there is a syntax error (ex: if there are a bunch of null bytes in the file), as manual intervention would be needed to recover anyway. The null statefile issue only seems to appear on the pipelines on windows nodes.

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
This issue appears sporadically

### DIFF
--- a/network/manager.go
+++ b/network/manager.go
@@ -5,6 +5,7 @@ package network
 
 import (
 	"context"
+	"encoding/json"
 	"net"
 	"sync"
 	"time"
@@ -192,6 +193,7 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
 	// Read any persisted state.
 	err := nm.store.Read(storeKey, nm)
 	if err != nil {
+		var syntaxErr *json.SyntaxError
 		if err == store.ErrKeyNotFound {
 			logger.Info("network store key not found")
 			// Considered successful.
@@ -199,6 +201,11 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
 		} else if err == store.ErrStoreEmpty {
 			logger.Info("network store empty")
 			return nil
+		} else if errors.As(err, &syntaxErr) {
+			// if null chars detected or failed to parse, state is unrecoverable; delete it
+			logger.Error("Failed to parse corrupted state, deleting", zap.Error(err))
+			nm.store.Remove()
+			return errors.Wrap(err, "failed to parse corrupted state")
 		} else {
 			logger.Error("Failed to restore state", zap.Error(err))
 			return err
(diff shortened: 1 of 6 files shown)


network/manager.go
```diff
@@ -192,13 +193,19 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
 	// Read any persisted state.
 	err := nm.store.Read(storeKey, nm)
 	if err != nil {
```
MikeZappa87, line 195: Does it make sense to log the contents invalid and all to help assist with troubleshooting?

QxBytes: I was looking into this and the store's file name and contents are not accessible-- currently leaning towards just adding a method to the interface to dump the contents of the store in string format.

store/json_test.go
```diff
@@ -129,6 +129,14 @@ func TestKeyValuePairsAreWrittenAndReadCorrectly(t *testing.T) {
 		t.Fatalf("Failed to create KeyValueStore %v\n", err)
 	}
 
+	defer os.Remove(testFileName)
+
+	// Dump empty store.
+	_, err = kvs.Dump()
+	if err == nil {
+		t.Fatal("Expected store to not exist")
```
vipul-21, line 137: We expect the store to exist right ? why the `nil` check and a `fatal` error ?

network/manager.go
```diff
@@ -192,13 +193,25 @@ func (nm *networkManager) restore(isRehydrationRequired bool) error {
...
 			logger.Info("network store empty")
 			return nil
+		} else if errors.As(err, &syntaxErr) {
+			// if null chars detected or failed to parse, state is unrecoverable; delete it
+			logger.Error("Failed to parse corrupted state, deleting", zap.Error(err))
+			contents, readErr := nm.store.Dump()
+			if readErr != nil {
+				logger.Error("Could not read corrupted state", zap.Error(readErr))
+			} else {
+				logger.Info("Logging state", zap.String("stateFile", contents))
```
vipul-21, line 211: nit: logging corrupted state.

store/json_test.go
QxBytes: The store struct does exist, but the actual json representation on disk doesn't exist yet.

network/manager.go
QxBytes: We log the corrupted state for debugging as per the suggestion by Michael above. Is there some drawback of doing this?

PR #3592, opened 2025-04-14
ci: Add check for cilium error logs to cilium nightly pipeline

jpayne3506:
**Reason for Change**:

Adds an additional level of alerting for breaking changes occurring in upstream cilium.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


vipul-21:
If we want to test the cilium logs for error, we can use the connectivity test for that. They have check log error and specify the `log-check-levels` as `error` so that it does not fail on warning.

vipul-21 (approved):

PR #3595, opened 2025-04-16
create APIPA with DelegatedNICs

akanso:
Enable testing APIPA with swiftV2.

This is not a long term solution. This is just a branch to enable a build that can be used for testing.

The longer-term fix includes adding a contract that allows the specification of APIPA as an additional interface.

**Issue Fixed**:

Ability to create APIPA even without having the mechanism to specify 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


tamilmani1989 (changes requested):
this is for testing. should not merge this PR

PR #3602, opened 2025-04-22
fix: validate iptable rule exists after calling insert or append iptable rule

QxBytes:
**Reason for Change**:

It was speculated that even though the iptables command returns with no errors, it was possible that the iptables rule was never created. This change aims to validate the rule was created and will error out the cni plugin if the rule specified in append or insert was not found (using a -C check).

Validated against transparent vlan and pipeline with no error: https://msazure.visualstudio.com/One/_build/results?buildId=121810862&view=results

**Issue Fixed**:

See above.

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/iptables/iptables.go
+++ b/iptables/iptables.go
@@ -3,6 +3,7 @@ package iptables
 // This package contains wrapper functions to program iptables rules
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/Azure/azure-container-networking/cni/log"
@@ -11,6 +12,7 @@ import (
 )
 
 var logger = log.CNILogger.With(zap.String("component", "cni-iptables"))
+var errCouldNotValidateRuleExists = errors.New("could not validate iptable rule exists after insertion")
 
 // cni iptable chains
 const (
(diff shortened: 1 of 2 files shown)


iptables/iptables.go
```diff
@@ -189,7 +198,14 @@ func (c *Client) AppendIptableRule(version, tableName, chainName, match, target
 	}
 
 	cmd := c.GetAppendIptableRuleCmd(version, tableName, chainName, match, target)
-	return c.RunCmd(version, cmd.Params)
+	err := c.RunCmd(version, cmd.Params)
```
MikeZappa87, line 201: RunCmd is calling ExecuteRawCommand which is deprecated. We should prioritize not using ExecuteRawCommand anymore.

QxBytes: The RunCmd here is from the existing code. Moving away from RunCmd can be done but would require refactoring  how we program iptables rules (and ensuring there is no regression).

vipul-21 (approved):

jpayne3506 (approved):

PR #3648, opened 2025-05-09
feat: explicit optional compression in dropgz

rbtr:
Previously, dropgz has expected and required that embedded blobs be gzip compressed.
This change makes decompressing the stream optional and defaults to no decompression, by adding a `--compression` flag and simply streaming the embed out if compression is unset.
`dropgz deploy` now defaults to assuming uncompressed embeds. Only gzip compression is supported and may be set.


dropgz/README.md
QxBytes, line 1: Would it be beneficial to add a new README with updated instructions for running dropgz?

dropgz/cmd/payload.go
```diff
@@ -32,7 +38,7 @@ func checksum(srcs, dests []string) error {
 	if len(srcs) != len(dests) {
 		return errors.Wrapf(embed.ErrArgsMismatched, "%d and %d", len(srcs), len(dests))
 	}
-	rc, err := embed.Extract("sum.txt")
+	rc, err := embed.Extract("sum.txt", compression)
```
QxBytes, line 41: Were there issues with compression previously? Why are we adding an option to disable compression (since previously my understanding is that it was always enabled?)?

QxBytes: Is it possible to test that checksum etc. is working when we don't have compression? Has it been tested manually once maybe?

QxBytes (commented):
If dropgz's default behavior now assumes uncompressed files, and previously we placed compressed files into the dropgz image (like inside azure-cni), do we also need to update these other dockerfiles (to place uncompressed files), or make it so that dropgz runs with gzip compression (your new flag) in pipelines etc.? Or does this compression on pertain to how the files are transferred out?

dropgz/cmd/payload.go
rbtr: I have verified that everything still works as expected.
As for why - when I created dropgz it was to ship a bunch of blobs in a single executable that knew how to self-extract those blobs. I made it expect and require that the blobs be pre-gzipped because my use case was to put that dropgz executable in a container and use it to get CNI bins on to Nodes, and I expected that making the executable as small as possible would mean making the image as small as possible and would result in the minimal image pull time and fastest CNI ready. 
This may not actually be right...
Containers are stored and transmitted compressed (as gzip by default) and the double compression was not actually reducing the final transmitted image size. Instead, it was adding compute time to decompress twice.

rbtr:
>do we also need to update these other dockerfiles (to place uncompressed files)

Yes, but dropgz is a separate go mod and is used as a library in those dockerfiles. It must be released before they can be updated, and if we make no change to them they just keep using the older version with compression.

dropgz/cmd/payload.go
rbtr: Also, we try to pre-cache images in the Node VHD, so transmit time on our happy path is 0. It's a waste of time to have additional decompression when the image is already on-disk.

QxBytes:
> > do we also need to update these other dockerfiles (to place uncompressed files)
> 
> Yes, but dropgz is a separate go mod and is used as a library in those dockerfiles. It must be released before they can be updated, and if we make no change to them they just keep using the older version with compression.

So let's say this PR gets merged and then we build a new cni/cns image right after-- will this image be broken because we place compressed images inside when we build dropgz, and then try to extract with the default compression as none?

rbtr:
nothing will break or change until we release a dropgz 0.0.13, and update the CNS/CNI dockerfiles to use it

QxBytes (approved):

PR #3649, opened 2025-05-10
Smtrudo/binary signing

sheylatrudo:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/build/binaries.jobs.yaml
+++ b/.pipelines/build/binaries.jobs.yaml
@@ -49,10 +49,11 @@ jobs:
     - job: sign_${{ job_data.job }}
       displayName: "Sign Binary - ${{ job_data.displayName }} -"
       strategy: ${{ job_data.strategy }}
-      dependsOn:
-      - binaries_${{ job_data.job }}
       pool:
-        type: linux
+        ${{ if eq(job_data.job, 'windows_amd64') }}:
+          type: windows
+        ${{ else }}:
+          type: linux
       variables:
         ob_outputDirectory: $(Build.SourcesDirectory)
         ob_artifactSuffix: _$(artifact)
--- a/.pipelines/build/binaries.jobs.yaml
+++ b/.pipelines/build/binaries.jobs.yaml
@@ -43,3 +43,29 @@ jobs:
           target: $(name)
           os: $(OS)
           arch: $(ARCH)
+
+
+  - ${{ elseif and(eq(job_data.templateContext.action, 'sign'), eq(parameters.official, 'true')) }}:
+    - job: sign_${{ job_data.job }}
+      displayName: "Sign Binary - ${{ job_data.displayName }} -"
+      strategy: ${{ job_data.strategy }}
+      dependsOn:
+      - binaries_${{ job_data.job }}
+      pool:
+        type: linux
+      variables:
+        ob_outputDirectory: $(Build.SourcesDirectory)
+        ob_artifactSuffix: _$(artifact)
+        ob_git_checkout: false
+      steps:
+      - task: DownloadPipelineArtifact@2
+        inputs:
+          targetPath: $(Build.SourcesDirectory)/${{ job_data.templateContext.repositoryArtifact }}
+          artifact: '${{ job_data.templateContext.repositoryArtifact }}'
+
+      - task: onebranch.pipeline.signing@1
+        inputs:
+          command: 'sign'
+          signing_profile: 'external_distribution'
+          files_to_sign: '**/*'
+          search_root: $(Build.SourcesDirectory)/${{ job_data.templateContext.repositoryArtifact }}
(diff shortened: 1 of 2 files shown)


.pipelines/build/images.jobs.yaml
```diff
@@ -0,0 +1,132 @@
+parameters:
+- name: images
+  type: jobList
+
+
+jobs:
+- ${{ each job_data in parameters.images }}:
```
jpayne3506, line 7: What unique information do we need to have to call this template correctly? Technically every single line here is a parameter https://github.com/Azure/azure-container-networking/blob/0efbbc79c2383ac48af23ee98823f59c5e5aeb73/.pipelines/run-pipeline.yaml#L53-L174


As a nit, I feel like this could have been pulled up one level https://github.com/Azure/azure-container-networking/blob/0efbbc79c2383ac48af23ee98823f59c5e5aeb73/.pipelines/run-pipeline.yaml#L49-L52

and iterated.

Instead we are passing a normal job.strategy.matrix into a jobList and parsing it.

.pipelines/build/ob-prepare.steps.yaml
```diff
@@ -51,7 +51,7 @@ steps:
     echo "##vso[task.setvariable variable=Tag;isOutput=true]$TAG"
     echo "Tag: $TAG"
 
-    IMAGEREPOPATH="artifact/dd590928-4e04-48cb-9d3d-ee06c5f0e17f/buddy"
+    IMAGEREPOPATH="artifact/dd590928-4e04-48cb-9d3d-ee06c5f0e17f/$BUILD_TYPE"
```
jpayne3506, line 54: TODO: ACR cleanup task for both repo paths

.pipelines/build/scripts/cni.sh
```diff
@@ -0,0 +1,65 @@
...
+
+
+CNI_NET_DIR="$REPO_ROOT"/cni/network/plugin
+pushd "$CNI_NET_DIR"
+  go build -v -a -trimpath \
+    -o "$OUT_DIR"/bin/azure-vnet"$FILE_EXT" \
+    -ldflags "-X main.version="$CNI_VERSION"" \
+    -gcflags="-dwarflocationlists=true" \
+    ./main.go
+popd
```
jpayne3506, line 19: Why are we continuously using pushd and popd? Can we just use it once for ACN root and then call the respective make commands or just `azure-cni-plugin`.

If not, why not just use relative pathing as we do in the Dockerfile?
https://github.com/Azure/azure-container-networking/blob/0efbbc79c2383ac48af23ee98823f59c5e5aeb73/cni/Dockerfile#L19-L24

jpayne3506: This logic goes to all `build/scripts/*`

jpayne3506 (commented):
Are the Dockerfiles in `/build/dockerfiles/*` there to give us reproducibility? I thought we had a script that was generating these per run.

.pipelines/build/binaries.jobs.yaml
```diff
@@ -43,3 +43,37 @@ jobs:
           target: $(name)
           os: $(OS)
           arch: $(ARCH)
+
+
+  - ${{ elseif and(eq(job_data.templateContext.action, 'sign'), job_data.templateContext.isOfficial) }}:
```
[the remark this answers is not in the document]
sheylatrudo: Signing is really only necessary for official builds. Have 2 different actions allows me to separate the two jobs, but also ensure that the dependencies are populated as needed.

.pipelines/build/scripts/cni.sh
sheylatrudo: Relative pathing can be flaky and some utilities ask for non-relative paths. Absolute paths are much easier to debug in logs. They are also less subject to issues when repo paths change on a higher level.

.pipelines/build/dockerfiles/npm.Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+COPY ${ARTIFACT_DIR}/files/kubeconfigtemplate.yaml kubeconfigtemplate.yaml
+COPY ${ARTIFACT_DIR}/scripts/setkubeconfigpath.ps1 setkubeconfigpath.ps1
+COPY ${ARTIFACT_DIR}/scripts/setkubeconfigpath-capz.ps1 setkubeconfigpath-capz.ps1
+COPY ${ARTIFACT_DIR}/bin/azure-npm.exe npm.exe
+
+CMD ["npm.exe", "start" "--kubeconfig=.\\kubeconfig"]
+
+
+FROM --platform=linux/${ARCH} mcr.microsoft.com/mirror/docker/library/ubuntu:24.04 as linux
```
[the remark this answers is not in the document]
jpayne3506: Will address this when we backport. NPM is not released on master

rohitams (approved):

jpayne3506 (approved):

PR #3653, opened 2025-05-12
change apipa endpoint gateway back to .1

ZetaoZhuang:
**Reason for Change**:

revert the apipa endpoint gw change  to .1  to resolve the nc to host connection issue in Linux contianer. This does not impact the initial change for changing the apipa network gw and loop adapter defualt gw from .1 to .2 on host.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This fix already been verified by updating the legacy runner cns with this fix for both windows and linux containers.

### DIFF
--- a/cns/hnsclient/hnsclient_windows.go
+++ b/cns/hnsclient/hnsclient_windows.go
@@ -525,6 +525,7 @@ func configureHostNCApipaEndpoint(
 		endpoint.Policies = append(endpoint.Policies, endpointPolicy)
 	}
 
+	// keep Apipa Endpoint gw as 169.254.128.1 to make sure NC to host connectivity work for both Linux and Windows containers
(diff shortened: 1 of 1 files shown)


cns/hnsclient/hnsclient_windows.go
```diff
@@ -527,7 +526,7 @@ func configureHostNCApipaEndpoint(
 	}
 
 	hcnRoute := hcn.Route{
```
msvik, line 528: We should leave a comment here for the change at line 528.

msvik (approved):

QxBytes (approved):
confirmed w/ zetao that new endpoints have "endpoint gateway addr" as .1 and existing endpoints can retain their "endpoint gateway addr" as .2 without issue

PR #3658, opened 2025-05-13
Prefix on nicv6 support

NihaNallappagari:
**Reason for Change**:
Current Problem: 
VNET currently limits scaling to 64K IP addresses per VNET due to per-IP route mappings. To support larger AKS clusters, the platform is introducing per-prefix (CIDR) route mapping, allowing a single mapping to represent multiple IPs (e.g., /24 enables 256 IPs per mapping). This enables scaling up to 16 million IPs per VNET without impacting the underlying platform.

Change needed:
Prefix on NIC v4 is supported in Swiftv1, but Swiftv1 does not support IPv6. This change upgrades Prefix on NIC v4 functionality to Swiftv2 and introduces IPv6 support in Swiftv2.

Changes
This PR has changes specific to Prefix on NICv6
CNS
--Consuming PrimaryIPv6, GatewayIPv6, MacAddress (DelegatedNIC) from NNC CRD because of dualstack NC
--IP allocation: Assign IPs of each IPFamily as part of RequestIPs api request (Currently when a pod is created, IPAM RequestsIPs from CNS where CNS picks one IP from each NC and hands it over to IPAM. 

IPAM
--Change RequestIPs response parsing to read GatewayIPv6 and MacAddress
--Populates Interfaces with MacAddress which is used by CNI to plumb routes to send traffic

[Design doc for pocv6 with sample NNC](https://microsoft.sharepoint.com/:w:/t/Aznet/EZy1F1lEUZpNiPwk6JsUtvcBHapTeO4yNEI6EMUxjElXnA?e=hNT4PN)
**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -146,8 +146,7 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 	for _, podIPInfo := range resp.PodIPInfo {
 		// Skip if interface already seen
 		// This is to avoid duplicate interfaces in the result
-		// which can happen if multiple IPs are assigned to the same interface
-		// or if multiple interfaces are assigned to the same pod
+		// Deduplication is necessary because there is one podIPInfo entry for each IP family(IPv4 and IPv6), and both may point to the same interface or if multiple interfaces are assigned to the same pod
 		if podIPInfo.MacAddress == "" || seenInterfaces[podIPInfo.MacAddress] {
 			continue
 		}
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -115,7 +115,7 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 	p.logger.Debug("Received CNS IP config response", zap.Any("response", resp))
 
 	// Get Pod IP and gateway IP from ip config response
-	podIPNet, err := ipconfig.ProcessIPConfigsResp(resp)
+	podIPNet, gatewayIP, err := ipconfig.ProcessIPConfigsResp(resp)
 	if err != nil {
 		p.logger.Error("Failed to interpret CNS IPConfigResponse", zap.Error(err), zap.Any("response", resp))
 		return cniTypes.NewError(ErrProcessIPConfigResponse, err.Error(), "failed to interpret CNS IPConfigResponse")
@@ -136,7 +136,32 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 				Mask: net.CIDRMask(ipNet.Bits(), 128), // nolint
 			}
 		}
+		ipConfig.Gateway = (*gatewayIP)[i]
 		cniResult.IPs[i] = ipConfig
+	}
+
+	cniResult.Interfaces = []*types100.Interface{}
+	seenInterfaces := map[string]bool{}
+
+	for _, podIPInfo := range resp.PodIPInfo {
+		// Skip if interface already seen
+		// This is to avoid duplicate interfaces in the result
+		// which can happen if multiple IPs are assigned to the same interface
+		// or if multiple interfaces are assigned to the same pod
+		if podIPInfo.MacAddress == "" || seenInterfaces[podIPInfo.MacAddress] {
+			continue
+		}
+
+		infMac, err := net.ParseMAC(podIPInfo.MacAddress)
+		if err != nil {
+			p.logger.Error("Failed to parse interface MAC address", zap.Error(err), zap.String("macAddress", podIPInfo.MacAddress))
+			return cniTypes.NewError(cniTypes.ErrUnsupportedField, err.Error(), "failed to parse interface MAC address")
+		}
+
+		cniResult.Interfaces = append(cniResult.Interfaces, &types100.Interface{
+			Mac: infMac.String(),
+		})
+		seenInterfaces[podIPInfo.MacAddress] = true
 	}
 
 	// Get versioned result
--- a/azure-ipam/ipam.go
+++ b/azure-ipam/ipam.go
@@ -146,7 +146,7 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 	for _, podIPInfo := range resp.PodIPInfo {
 		// Skip if interface already seen
 		// This is to avoid duplicate interfaces in the result
-		// Deduplication is necessary because there is one podIPInfo entry for each IP family(IPv4 and IPv6), and both may point to the same interface or if multiple interfaces are assigned to the same pod
+		// Deduplication is necessary because there is one podIPInfo entry for each IP family(IPv4 and IPv6), and both may point to the same interface
 		if podIPInfo.MacAddress == "" || seenInterfaces[podIPInfo.MacAddress] {
 			continue
 		}
--- a/azure-ipam/ipam_test.go
+++ b/azure-ipam/ipam_test.go
@@ -92,6 +92,7 @@ func (c *MockCNSClient) RequestIPAddress(ctx context.Context, ipconfig cns.IPCon
 					IPAddress:    "10.0.1.10",
 					PrefixLength: 24,
 				},
+				MacAddress: "00:11:22:33:44:55",
 				NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
 					IPSubnet: cns.IPSubnet{
 						IPAddress:    "10.0.1.0",
@@ -124,6 +125,56 @@ func (c *MockCNSClient) RequestIPs(ctx context.Context, ipconfig cns.IPConfigsRe
 		e.Code = types.UnsupportedAPI
 		e.Err = errUnsupportedAPI
 		return nil, e
+	case "happyArgsDual":
+		result := &cns.IPConfigsResponse{
+			PodIPInfo: []cns.PodIpInfo{
+				{
+					PodIPConfig: cns.IPSubnet{
+						IPAddress:    "10.0.1.10",
+						PrefixLength: 24,
+					},
+					MacAddress: "00:11:22:33:44:55",
+					NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
+						IPSubnet: cns.IPSubnet{
+							IPAddress:    "10.0.1.0",
+							PrefixLength: 24,
+						},
+						DNSServers:       nil,
+						GatewayIPAddress: "10.0.0.1",
+					},
+					HostPrimaryIPInfo: cns.HostIPInfo{
+						Gateway:   "10.0.0.1",
+						PrimaryIP: "10.0.0.1",
+						Subnet:    "10.0.0.0/24",
+					},
+				},
+				{
+					PodIPConfig: cns.IPSubnet{
+						IPAddress:    "fd11:1234::1",
+						PrefixLength: 120,
+					},
+					MacAddress: "00:11:22:33:44:55", // Same MAC for dual-stack scenario
+					NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
+						IPSubnet: cns.IPSubnet{
+							IPAddress:    "fd11:1234::",
+							PrefixLength: 120,
+						},
+						DNSServers:         nil,
+						GatewayIPv6Address: "fe80::1234:5678:9abc",
+					},
+					HostPrimaryIPInfo: cns.HostIPInfo{
+						Gateway:   "fe80::1234:5678:9abc",
+						PrimaryIP: "fe80::1234:5678:9abc",
+						Subnet:    "fd11:1234::/120",
+					},
+				},
+			},
+			Response: cns.Response{
+				ReturnCode: 0,
+				Message:    "",
+			},
+		}
+		return result, nil
 	case "failProcessCNSResp":
 		result := &cns.IPConfigsResponse{
 			PodIPInfo: []cns.PodIpInfo{
@@ -180,6 +231,7 @@ func (c *MockCNSClient) RequestIPs(ctx context.Context, ipconfig cns.IPConfigsRe
 						IPAddress:    "10.0.1.10",
 						PrefixLength: 24,
 					},
+					MacAddress: "00:11:22:33:44:55",
 					NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
 						IPSubnet: cns.IPSubnet{
 							IPAddress:    "10.0.1.0",
@@ -199,6 +251,7 @@ func (c *MockCNSClient) RequestIPs(ctx context.Context, ipconfig cns.IPConfigsRe
 						IPAddress:    "fd11:1234::1",
 						PrefixLength: 120,
 					},
+					MacAddress: "00:11:22:33:44:55", // Same MAC for dual-stack scenario
 					NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
 						IPSubnet: cns.IPSubnet{
 							IPAddress:    "fd11:1234::",
@@ -211,6 +264,46 @@ func (c *MockCNSClient) RequestIPs(ctx context.Context, ipconfig cns.IPConfigsRe
 						Gateway:   "fe80::1234:5678:9abc",
 						PrimaryIP: "fe80::1234:5678:9abc",
 						Subnet:    "fd11:1234::/120",
+					},
+				},
+				{
+					PodIPConfig: cns.IPSubnet{
+						IPAddress:    "192.168.1.10",
+						PrefixLength: 24,
+					},
+					MacAddress: "aa:bb:cc:dd:ee:ff",
+					NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
+						IPSubnet: cns.IPSubnet{
+							IPAddress:    "192.168.1.0",
+							PrefixLength: 24,
+						},
+						DNSServers:       nil,
+						GatewayIPAddress: "192.168.1.1",
+					},
+					HostPrimaryIPInfo: cns.HostIPInfo{
+						Gateway:   "192.168.1.1",
+						PrimaryIP: "192.168.1.1",
+						Subnet:    "192.168.1.0/24",
+					},
+				},
+				{
+					PodIPConfig: cns.IPSubnet{
+						IPAddress:    "172.16.1.10",
+						PrefixLength: 24,
+					},
+					MacAddress: "",
+					NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
+						IPSubnet: cns.IPSubnet{
+							IPAddress:    "172.16.1.0",
+							PrefixLength: 24,
+						},
+						DNSServers:       nil,
+						GatewayIPAddress: "172.16.1.1",
+					},
+					HostPrimaryIPInfo: cns.HostIPInfo{
+						Gateway:   "172.16.1.1",
+						PrimaryIP: "172.16.1.1",
+						Subnet:    "172.16.1.0/24",
 					},
 				},
 			},
@@ -308,6 +401,11 @@ func TestCmdAdd(t *testing.T) {
 			args: buildArgs("happyArgsSingle", happyPodArgs, happyNetConfByteArr),
 			want: &types100.Result{
 				CNIVersion: "1.0.0",
+				Interfaces: []*types100.Interface{
+					{
+						Mac: "00:11:22:33:44:55",
+					},
+				},
 				IPs: []*types100.IPConfig{
 					{
 						Address: net.IPNet{
@@ -326,6 +424,11 @@ func TestCmdAdd(t *testing.T) {
 			args: buildArgs("happyArgsDual", happyPodArgs, happyNetConfByteArr),
 			want: &types100.Result{
 				CNIVersion: "1.0.0",
+				Interfaces: []*types100.Interface{
+					{
+						Mac: "00:11:22:33:44:55", // Single interface for dual-stack
+					},
+				},
 				IPs: []*types100.IPConfig{
 					{
 						Address: net.IPNet{
@@ -340,6 +443,53 @@ func TestCmdAdd(t *testing.T) {
 							Mask: net.CIDRMask(120, 128),
 						},
 						Gateway: net.ParseIP("fe80::1234:5678:9abc"),
+					},
+				},
+				DNS: cniTypes.DNS{},
+			},
+			wantErr: false,
+		},
+		{
+			name: "Test MAC address deduplication and multi-interface",
+			args: buildArgs("happyArgsMultiInterface", happyPodArgs, happyNetConfByteArr),
+			want: &types100.Result{
+				CNIVersion: "1.0.0",
+				Interfaces: []*types100.Interface{
+					{
+						Mac: "00:11:22:33:44:55", // First unique MAC (dual-stack)
+					},
+					{
+						Mac: "aa:bb:cc:dd:ee:ff", // Second unique MAC
+					},
+				},
+				IPs: []*types100.IPConfig{
+					{
+						Address: net.IPNet{
+							IP:   net.IPv4(10, 0, 1, 10),
+							Mask: net.CIDRMask(24, 32),
+						},
+						Gateway: net.IPv4(10, 0, 0, 1),
+					},
+					{
+						Address: net.IPNet{
+							IP:   net.ParseIP("fd11:1234::1"),
+							Mask: net.CIDRMask(120, 128),
+						},
+						Gateway: net.ParseIP("fe80::1234:5678:9abc"),
+					},
+					{
+						Address: net.IPNet{
+							IP:   net.IPv4(192, 168, 1, 10),
+							Mask: net.CIDRMask(24, 32),
+						},
+						Gateway: net.IPv4(192, 168, 1, 1),
+					},
+					{
+						Address: net.IPNet{
+							IP:   net.IPv4(172, 16, 1, 10),
+							Mask: net.CIDRMask(24, 32),
+						},
+						Gateway: net.IPv4(172, 16, 1, 1),
 					},
 				},
 				DNS: cniTypes.DNS{},
--- a/azure-ipam/ipam_test.go
+++ b/azure-ipam/ipam_test.go
@@ -58,6 +58,33 @@ func (c *MockCNSClient) RequestIPAddress(ctx context.Context, ipconfig cns.IPCon
 			},
 		}
 		return result, nil
+	case "nilGateway":
+		result := &cns.IPConfigResponse{
+			PodIpInfo: cns.PodIpInfo{
+				PodIPConfig: cns.IPSubnet{
+					IPAddress:    "10.0.1.10",
+					PrefixLength: 24,
+				},
+				NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
+					IPSubnet: cns.IPSubnet{
+						IPAddress:    "10.0.1.0",
+						PrefixLength: 24,
+					},
+					DNSServers:       nil,
+					GatewayIPAddress: "", // nil/empty gateway
+				},
+				HostPrimaryIPInfo: cns.HostIPInfo{
+					Gateway:   "",
+					PrimaryIP: "10.0.0.1",
+					Subnet:    "10.0.0.0/24",
+				},
+			},
+			Response: cns.Response{
+				ReturnCode: 0,
+				Message:    "",
+			},
+		}
+		return result, nil
 	default:
 		result := &cns.IPConfigResponse{
 			PodIpInfo: cns.PodIpInfo{
@@ -92,7 +119,7 @@ func (c *MockCNSClient) RequestIPs(ctx context.Context, ipconfig cns.IPConfigsRe
 	switch ipconfig.InfraContainerID {
 	case "failRequestCNSArgs":
 		return nil, errFoo
-	case "happyArgsSingle", "failProcessCNSRespSingleIP", "failRequestCNSArgsSingleIP":
+	case "happyArgsSingle", "failProcessCNSRespSingleIP", "failRequestCNSArgsSingleIP", "nilGateway":
 		e := &client.CNSClientError{}
 		e.Code = types.UnsupportedAPI
 		e.Err = errUnsupportedAPI
@@ -129,8 +156,8 @@ func (c *MockCNSClient) RequestIPs(ctx context.Context, ipconfig cns.IPConfigsRe
 							IPAddress:    "fd11:1234::",
 							PrefixLength: 112,
 						},
-						DNSServers:       nil,
-						GatewayIPAddress: "fe80::1234:5678:9abc",
+						DNSServers:         nil,
+						GatewayIPv6Address: "fe80::1234:5678:9abc",
 					},
 					HostPrimaryIPInfo: cns.HostIPInfo{
 						Gateway:   "fe80::1234:5678:9abc",
@@ -177,8 +204,8 @@ func (c *MockCNSClient) RequestIPs(ctx context.Context, ipconfig cns.IPConfigsRe
 							IPAddress:    "fd11:1234::",
 							PrefixLength: 120,
 						},
-						DNSServers:       nil,
-						GatewayIPAddress: "fe80::1234:5678:9abc",
+						DNSServers:         nil,
+						GatewayIPv6Address: "fe80::1234:5678:9abc",
 					},
 					HostPrimaryIPInfo: cns.HostIPInfo{
 						Gateway:   "fe80::1234:5678:9abc",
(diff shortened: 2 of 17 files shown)


NihaNallappagari:

azure-ipam/ipam.go
```diff
@@ -136,9 +136,33 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
...
 
+	cniResult.Interfaces = []*types100.Interface{}
+	seenInterfaces := map[string]bool{}
+
+	for _, podIPInfo := range resp.PodIPInfo {
+		// Skip if interface already seen
+		// This is to avoid duplicate interfaces in the result
+		// Deduplication is necessary because there is one podIPInfo entry for each IP family(IPv4 and IPv6), and both may point to the same interface or if multiple interfaces are assigned to the same pod
+		if podIPInfo.MacAddress == "" || seenInterfaces[podIPInfo.MacAddress] {
+			continue
```
tamilmani1989, line 151: skipping s fine but should we add other ip address in cni result structure?

azure-ipam/ipconfig/ipconfig.go
```diff
@@ -63,23 +65,38 @@ func CreateIPConfigsReq(args *cniSkel.CmdArgs) (cns.IPConfigsRequest, error) {
...
+		if podIPNet.Addr().Is4() {
+			gatewayIP = net.ParseIP(resp.PodIPInfo[i].NetworkContainerPrimaryIPConfig.GatewayIPAddress)
+		} else if podIPNet.Addr().Is6() {
+			gatewayIP = net.ParseIP(resp.PodIPInfo[i].NetworkContainerPrimaryIPConfig.GatewayIPv6Address)
+		}
+
+		if gatewayIP != nil {
+			gatewaysIPs[i] = gatewayIP
+		} else {
+			logger.Printf("gatewayIP is nil or gateway ip parse failed for pod ip %s", resp.PodIPInfo[i].PodIPConfig.IPAddress)
```
tamilmani1989: should we return error if gwip is nil?

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
```diff
@@ -16,12 +16,15 @@ import (
 func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPrefix netip.Prefix, subnet cns.IPSubnet) (*cns.CreateNetworkContainerRequest, error) {
 	secondaryIPConfigs := map[string]cns.SecondaryIPConfig{}
 
-	// iterate through all IP addresses in the subnet described by primaryPrefix and
-	// add them to the request as secondary IPConfigs.
-	for addr := primaryIPPrefix.Masked().Addr(); primaryIPPrefix.Contains(addr); addr = addr.Next() {
-		secondaryIPConfigs[addr.String()] = cns.SecondaryIPConfig{
-			IPAddress: addr.String(),
-			NCVersion: int(nc.Version),
+	// in the case of vnet prefix on swift v2 the primary IP is a /32 and should not be added to secondary IP configs
```
tamilmani1989: @kadevu is primary ip /32 only on swiftv2 and not on swiftv1? I knew we had this discussion, sorry i forgot the reason behind this check

cns/restserver/internalapi.go
```diff
@@ -191,6 +191,13 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
 	outdatedNCs := map[string]struct{}{}
 	programmedNCs := map[string]struct{}{}
 	for idx := range service.state.ContainerStatus {
+
+		// For prefix-on-NIC SwiftV2 NIC scenario, the NC details is published to different path in NMAgent and call to NMA here creates error. So Skipping nma check, synchost version logic.
+		ncStatus := service.state.ContainerStatus[idx]
+		if mac := ncStatus.CreateNetworkContainerRequest.NetworkInterfaceInfo.MACAddress; mac != "" {
+			logger.Printf("NC %s has MAC address for prefix on nic swiftv2 scenario, skipping syncHostNCVersion", ncStatus.ID)
+			continue
```
tamilmani1989: so we detect swiftv2 based on mac? why not add a config field in cns to specify swiftv2. i guess there is one for singularity which we can reuse

cns/restserver/internalapi_linux.go
```diff
@@ -68,6 +68,12 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 
 	// use any secondary ip + the nnc prefix length to get an iptables rule to allow dns and imds traffic from the pods
 	for _, v := range req.SecondaryIPConfigs {
+		// check if the ip address is IPv4
+		if net.ParseIP(v.IPAddress).To4() == nil {
```
tamilmani1989, line 125: why we skip if not ipv4?

tamilmani1989:
@rbtr can you review this as well since it touches cns changes for aks swift too

cns/restserver/internalapi.go
NihaNallappagari: Sure. This is specific to swiftv2 NIC, we can make use of config fields. Is this the file you are talking about [cns config](https://github.com/Azure/azure-container-networking/blob/master/cns/configuration/cns_config.json). If so, I don't find a field specific to singularity. Could you share the field you are talking about? Thanks

azure-ipam/ipam.go
NihaNallappagari: The skip is only for interfaces, not for ip's. All ip's are added to [cniResults](https://github.com/Azure/azure-container-networking/pull/3658/files/e00a5b71118fbf1b6d7e39ba2aded03238e351f3#diff-4016ee971c9dc67ecde0d75e8dff42b3c188fa03e21ea2e005157b958a4f2c3eR124)

azure-ipam/ipconfig/ipconfig.go
NihaNallappagari: Updated code to return error message

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
NihaNallappagari: @kadevu is having issues to comment on the PR. His response "yes this is only for Swiftv2. This is basically the primary ip to assign the to the delegated nic and not assigned to any pod which is in line with the InterfaceCompartment published to pubsub. In case of Swiftv1, this field has /28 instead and all those ips are assigned to the pods"

cns/restserver/internalapi_linux.go
NihaNallappagari: pocv6 SecondaryIPConfigs have ipv6 and ipv4 addresses. This programSNATRules sets ipv4 iptables SNAT rules for DNS and IMDS, and can't handle ipv6 address. Without this lines it does throw error while creating a pod

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
tamilmani1989: since we are not assigning any ip to delegated nic, why it has to be /32 and not /28 as like swiftv1?

kadevu: the interfacecompartment schema enforces customeripaddress to be /32 and customeripprefix to be /28. For the customeripadderss we carve out a /28 and then assign a /32 from it. The rest of the IPs in the /28 are not programmed by nmagent which is why we cannot use these for the pods. To make sure there is no wastage of ips we use the other ips from the /28 from which we got the /32 to assign customeripaddress for other nodes in the nodepool.

azure-ipam/ipconfig/ipconfig.go
```diff
@@ -3,9 +3,11 @@ package ipconfig
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/netip"
 
 	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/logger"
```
rbtr: azure-ipam should not import the cns logger

cns/restserver/internalapi.go
rbtr: what about this one?
https://github.com/Azure/azure-container-networking/blob/f617f1fe0da3e7540bd2e2a6b61b751dcee2fcb3/cns/configuration/configuration.go#L37

cns/restserver/internalapi_linux.go
rbtr: shouldn't `programSNATRules` be upgraded to support v6 instead of only calling it for v4?

cns/restserver/internalapi_test.go
```diff
@@ -312,6 +312,40 @@ func TestPendingIPsGotUpdatedWhenSyncHostNCVersion(t *testing.T) {
 	}
 }
 
+func TestSyncHostNCVersion_SkipsNCVersionForPrefixOnNICSwiftV2(t *testing.T) {
```
rbtr: why a new test and not a new scenario added to the existing `TestSyncHostNCVersion`?

cns/restserver/ipam.go
```diff
@@ -995,38 +995,85 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
...
+			ip := net.ParseIP(secIPConfig.IPAddress)
+			if ip == nil {
+				continue
+			}
+
+			if ip.To4() != nil {
+				ncIPFamilies[cns.IPv4] = struct{}{}
+			} else {
+				ncIPFamilies[cns.IPv6] = struct{}{}
+			}
```
rbtr: use `netip.ParseAddr` to get a `netip.Addr`, then you can call `Addr#Is4` or `Addr#Is6` instead. prefer netip to net package in new code

rbtr (changes requested):
are there AzCNI changes I'm missing here or is this only for Cilium?

cns/restserver/internalapi_linux.go
NihaNallappagari: The programSNATRules function may or may not be required for IPv6. I'm currently investigating how the behavior affects our scenario and what changes might be necessary to support IPv6. If needed, I plan to add a separate function specifically for IPv6 handling. However, to avoid expanding the scope of this PR, I’d prefer to keep it focused on the basic scenarios for now.

cns/restserver/internalapi_test.go
NihaNallappagari: Existing TestSyncHostNCVersion is different from this scenario. This scenario take macAddress into consideration and skips the synchost update. I update the test to reuse most of TestSyncHostNCVersion code.

azure-ipam/ipconfig/ipconfig.go
```diff
@@ -63,23 +64,39 @@ func CreateIPConfigsReq(args *cniSkel.CmdArgs) (cns.IPConfigsRequest, error) {
...
+		var gatewayStr string
+		if podIPNet.Addr().Is4() {
+			gatewayStr = podInfo.NetworkContainerPrimaryIPConfig.GatewayIPAddress
+		} else if podIPNet.Addr().Is6() {
+			gatewayStr = podInfo.NetworkContainerPrimaryIPConfig.GatewayIPv6Address
+		}
+
+		gatewayIP := net.ParseIP(gatewayStr)
+		if gatewayIP == nil {
+			return nil, nil, errors.Errorf("failed to parse gateway IP %q for pod ip %s", gatewayStr, podInfo.PodIPConfig.IPAddress)
```
tamilmani1989: if nil, why we log gatewayStr. will it cause any panic? did we add UT for this?

cns/restserver/internalapi_linux.go
tamilmani1989: we have to confirm if azure dns/azure imds has ipv6 address to reach out to

azure-ipam/ipconfig/ipconfig.go
NihaNallappagari: fixed

azure-ipam/ipconfig/ipconfig.go
NihaNallappagari: Gateway string is either nil or could not be parsed. Logging may help diagnose the issue. As we returning error, ipam will handle the error case and ignores the return values of pods and gatway ip https://github.com/Azure/azure-container-networking/blob/8806489bf31dbfb2d9a169da7f44799e568a31d0/azure-ipam/ipam.go#L118. Thanks for pointing out the missing test. I've added the appropriate test scenario to cover this case.

cns/restserver/internalapi_linux.go
NihaNallappagari: Added to comment that dns and imds doesn't support ipv6. I will followup with AKS SME's and update their documentation

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
NihaNallappagari: I removed the if check for single ip. We have some discussions going with Control plane SMEs. I will do a follow-up to change this

cns/restserver/internalapi.go
NihaNallappagari: I removed the mac address check. Once the NMAgent exposes NC details, I will make the changes in a new PR

cns/restserver/internalapi_test.go
NihaNallappagari: test is no longer needed. Removed it

cns/restserver/ipam.go
NihaNallappagari: Made the changes mentioned. Thanks

azure-ipam/ipam.go
```diff
@@ -136,9 +136,33 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
...
+			p.logger.Error("Failed to parse interface MAC address", zap.Error(err), zap.String("macAddress", podIPInfo.MacAddress))
+			return cniTypes.NewError(cniTypes.ErrUnsupportedField, err.Error(), "failed to parse interface MAC address")
+		}
+
+		cniResult.Interfaces = append(cniResult.Interfaces, &types100.Interface{
+			Mac: infMac.String(),
+		})
+		seenInterfaces[podIPInfo.MacAddress] = true
+	}
+
```
tamilmani1989, line 165: can we add UT for this case with dualstack ips?

cns/restserver/ipam.go
```diff
@@ -995,38 +996,85 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
...
+			if addr.Is4() {
+				ncIPFamilies[cns.IPv4] = struct{}{}
+			} else if addr.Is6() {
+				ncIPFamilies[cns.IPv6] = struct{}{}
+			}
+		}
+	}
+	// Makes sure we have at least one IPFamily across all NCs
+	numOfIPFamilies := len(ncIPFamilies)
+
```
tamilmani1989: can we move this to separate function say `getNumberofIPFamily`

tamilmani1989: add UT for this too

cns/restserver/internalapi.go
```diff
@@ -620,7 +620,6 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
 	} else {
 		logger.Errorf(returnMessage)
 	}
-
```
tamilmani1989, line 652: can you revert this file if no changes are made?

azure-ipam/ipam.go
NihaNallappagari: Added UT for [dual stack scenarios](https://github.com/Azure/azure-container-networking/pull/3658/files#diff-a3876d4605652e54eb4192d86603b8ceb4f443e9abcb7b59296d2641e45b9d61R453)

cns/restserver/internalapi.go
NihaNallappagari: reverted

cns/restserver/ipam.go
NihaNallappagari: Added [test](https://github.com/Azure/azure-container-networking/pull/3658/files#diff-2ef2adc803f49c2758c4c22d6f4f786a91bbc04999747af6e41292d68d808325R2313) and moved chunk of code to new function

NihaNallappagari:
> are there AzCNI changes I'm missing here or is this only for Cilium?

The changes are related to only Cilium, no azure cni changes made

azure-ipam/ipam_test.go
```diff
@@ -298,24 +424,83 @@ func TestCmdAdd(t *testing.T) {
...
+			want: &types100.Result{
+				CNIVersion: "1.0.0",
+				Interfaces: []*types100.Interface{
+					{
+						Mac: "00:11:22:33:44:55", // First unique MAC (dual-stack)
+					},
+					{
+						Mac: "aa:bb:cc:dd:ee:ff", // Second unique MAC
+					},
+				},
```
tamilmani1989: why two macs? wont there be only interface in pod?

cns/restserver/ipam.go
```diff
@@ -995,38 +996,64 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
...
+	if len(ipsToAssign) != numberOfIPs {
 		for ncID := range service.state.ContainerStatus {
-			if _, found := ipsToAssign[ncID]; found {
-				continue
+			for ipFamily := range ncIPFamilies {
+				if _, found := ipsToAssign[generateAssignedIPKey(ncID, ipFamily)]; found {
+					continue
+				}
+				return podIPInfo, errors.Errorf("not enough IPs available of type %s for %s, waiting on Azure CNS to allocate more with NC Status: %s",
+					ipFamily, ncID, string(service.state.ContainerStatus[ncID].CreateNetworkContainerRequest.NCStatus))
```
tamilmani1989, line 1048: do we have UT for catching this error case? also why we iterate all ncs. dont we have reference to NC or find NC based on podipconfig state?

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
tamilmani1989: lets add a TODO

cns/restserver/internalapi.go
tamilmani1989: lets add a TODO

cns/restserver/ipam.go
NihaNallappagari: Yes there is a [unit test](https://github.com/Azure/azure-container-networking/pull/3658/files#diff-2ef2adc803f49c2758c4c22d6f4f786a91bbc04999747af6e41292d68d808325R2387) to cover this error scenario.  podipconfig  has nc id can only determines the nc's that has available ip's. This block of code checks for if there is any missing family for specific ip. For instance, if ipv6 provision fails for one NC, this chunk of code will check for those missing cases and throws an error

azure-ipam/ipam_test.go
NihaNallappagari: yes, there won't be a scenario of 2 interfaces to a pod. No code changes are made in specific to this scenario, removed the test case of pod with 2 interfaces

cns/restserver/internalapi.go
NihaNallappagari: Added todo

tamilmani1989 (approved):

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
```diff
@@ -13,15 +14,18 @@ import (
 // by adding all IPs in the the block to the secondary IP configs list. It does not skip any IPs.
 //
 //nolint:gocritic //ignore hugeparam
-func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPrefix netip.Prefix, subnet cns.IPSubnet) (*cns.CreateNetworkContainerRequest, error) {
+func createNCRequestFromStaticNCHelper(nc v1alpha.NetworkContainer, primaryIPPrefix netip.Prefix, subnet cns.IPSubnet, config *configuration.CNSConfig) (*cns.CreateNetworkContainerRequest, error) {
```
rbtr: Instead of passing in the _whole CNS config_, what's the least amount of information we can give this function?

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
```diff
@@ -13,15 +14,18 @@ import (
...
 	secondaryIPConfigs := map[string]cns.SecondaryIPConfig{}
 
 	// iterate through all IP addresses in the subnet described by primaryPrefix and
 	// add them to the request as secondary IPConfigs.
-	for addr := primaryIPPrefix.Masked().Addr(); primaryIPPrefix.Contains(addr); addr = addr.Next() {
-		secondaryIPConfigs[addr.String()] = cns.SecondaryIPConfig{
-			IPAddress: addr.String(),
-			NCVersion: int(nc.Version),
+	// Process primary prefix IPs in all scenarios except when nc.Type is v1alpha.VNETBlock AND SwiftV2 is enabled
+	if !(config.EnableSwiftV2 && nc.Type == v1alpha.VNETBlock) {
```
rbtr: Don't change behavior based on some unconnected scenario config. At this low level of implementation, behavior should change based on behavioral config. The scenario shouldn't make it down the stack like this. 
`createNCRequestFromStaticNCHelper` should not know or care about SwiftV2. It only needs to know "process primary prefix IP or not".
This seems minor but is actually a huge architectural tenet. As implemented, this behavior is not reusable if we add a new scenario which wants it - we have to thread a whole new scenario all the way down in to this func. Or, the caller, which already knows the scenario, can pass the bool - with no code change here.

cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -43,18 +44,20 @@ type Reconciler struct {
...
 	nodeIP             string
+	config             *configuration.CNSConfig
 }
 
 // NewReconciler creates a NodeNetworkConfig Reconciler which will get updates from the Kubernetes
 // apiserver for NNC events.
 // Provided nncListeners are passed the NNC after the Reconcile preprocesses it. Note: order matters! The
 // passed Listeners are notified in the order provided.
-func NewReconciler(cnscli cnsClient, ipampoolmonitorcli nodeNetworkConfigListener, nodeIP string) *Reconciler {
+func NewReconciler(cnscli cnsClient, ipampoolmonitorcli nodeNetworkConfigListener, nodeIP string, config *configuration.CNSConfig) *Reconciler {
```
rbtr: Not going to repeat myself everywhere but comments apply globally.

cns/restserver/ipam.go
```diff
@@ -1061,10 +1088,18 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
 		return podIPInfo, fmt.Errorf("not enough IPs available, waiting on Azure CNS to allocate more")
 	}
 
-	logger.Printf("[AssignDesiredIPConfigs] Successfully assigned IPs for pod %+v", podInfo)
+	//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
+	logger.Printf(
+		"[AssignAvailableIPConfigs] Successfully assigned IPs for pod %+v",
+		podInfo,
+	)
```
rbtr, line 1088: please don't just throw newlines around everywhere unnecessarily

cns/service/main.go
```diff
@@ -1311,7 +1311,13 @@ type ipamStateReconciler interface {
...
 // TODO(rbtr) where should this live??
 // reconcileInitialCNSState initializes cns by passing pods and a CreateNetworkContainerRequest
-func reconcileInitialCNSState(ctx context.Context, cli nodeNetworkConfigGetter, ipamReconciler ipamStateReconciler, podInfoByIPProvider cns.PodInfoByIPProvider) error {
+func reconcileInitialCNSState(
+	ctx context.Context,
+	cli nodeNetworkConfigGetter,
+	ipamReconciler ipamStateReconciler,
+	podInfoByIPProvider cns.PodInfoByIPProvider,
+	config *configuration.CNSConfig,
+) error {
```
rbtr: Why? My monitor is wider than it is tall 🙁

cns/NetworkContainerContract.go
```diff
@@ -755,3 +756,11 @@ type NodeRegisterRequest struct {
...
 	NmAgentSupportedApis []string
 }
+
+// IPFamily - Enum for determining IPFamily when retrieving IPs from network containers
+type IPFamily string
+
+const (
+	IPv4 IPFamily = "ipv4"
+	IPv6 IPFamily = "ipv6"
+)
```
rbtr, line 766: Is this actually necessary?

cns/restserver/ipam.go
```diff
@@ -995,38 +996,64 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
...
+	numOfIPFamilies := service.GetIPFamilyCount()
+
+	// Get the actual IP families map for validation
+	ncIPFamilies := service.getIPFamiliesMap()
+
+	// Determine the number of IPs to assign based on IP families found
+	numberOfIPs := numOfNCs
+	if numOfIPFamilies != 0 {
+		numberOfIPs = numOfIPFamilies
+	}
```
rbtr: This is a substantial behavior change to the very core of CNS. Why is the "1 IP from each NC" strategy insufficient?
If it is necessary, I'm not sure if we should make the behavior change in-place like this. It would probably be better to implement it from scratch and take a branching approach, to not bring additional legacy behavior/side effects along.

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
NihaNallappagari: done

cns/kubecontroller/nodenetworkconfig/conversion_linux.go
NihaNallappagari: As suggested, instead of sending the full config, I'm passing the swiftv2 flag. I could pass processPrimaryPrefixIP, but the stack is called from different places and NC details aren't available at high level to check NC type. Since the NC type is known here and instead of making changes to couple of places in hierarchy, I'm checking for VNETBlock here and using the swiftv2 flag

cns/kubecontroller/nodenetworkconfig/reconciler.go
NihaNallappagari: done

cns/restserver/ipam.go
NihaNallappagari: I added it to aid debugging; it doesn't affect functionality, removing it now

cns/NetworkContainerContract.go
NihaNallappagari: This is used in multiple places to check for IPv4 or IPv6. Defining as const improves type safety and consistency across

cns/restserver/ipam.go
NihaNallappagari: For dual stack, assigning only one IP per NC isn't sufficient. This change ensures that each NC gets two IPs.
I didn’t quite follow what you meant by 'implement from scratch and separate the path from existing legacy code.' Could you share any suggestions?

cns/service/main.go
NihaNallappagari: This is a formatting change. I reverted to original

cns/restserver/ipam.go
NihaNallappagari: We discussed this offline, and everyone agreed on the change. Marking the conversation as resolved

cns/restserver/ipam.go
```diff
@@ -995,38 +996,64 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
...
+
+	// Get the number of distinct IP families (IPv4/IPv6) across all NC's
+	numOfIPFamilies := service.GetIPFamilyCount()
+
+	// Get the actual IP families map for validation
+	ncIPFamilies := service.getIPFamiliesMap()
+
+	// Determine the number of IPs to assign based on IP families found
+	numberOfIPs := numOfNCs
+	if numOfIPFamilies != 0 {
```
rbtr: this can't be 0 (except if there are no NCs)? I think all these re/assignments are unnecessary and you could simply use the GetIPFamilyCount() result

rbtr (commented):
lgtm with one question/comment

cns/restserver/ipam.go
NihaNallappagari: @rbtr  I agree, assignments have been removed as they aren’t needed. The no NC scenario is already handled earlier in the flow, so no error handling is needed.

rbtr (approved):

azure-ipam/ipam.go
```diff
@@ -136,7 +136,14 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
 				Mask: net.CIDRMask(ipNet.Bits(), 128), // nolint
 			}
 		}
-		ipConfig.Gateway = (*gatewayIP)[i]
+
+		if gatewayIP != nil && len(*gatewayIP) > i {
```
tamilmani1989: can you add comment what you are validating by this condition `len(*gatewayIP) > i`

NihaNallappagari: @tamilmani1989 I had originally added a warning log to flag when the gateway was missing for debugging purposes. Since the "ADD success" log that follows already includes the gateway, the earlier log was redundant. I’ve removed and relevant nil checks. Thanks

azure-ipam/ipam.go
```diff
@@ -136,9 +136,33 @@ func (p *IPAMPlugin) CmdAdd(args *cniSkel.CmdArgs) error {
...
 	}
 
+	cniResult.Interfaces = []*types100.Interface{}
+	seenInterfaces := map[string]bool{}
+
+	for _, podIPInfo := range resp.PodIPInfo {
+		// Skip if interface already seen
+		// This is to avoid duplicate interfaces in the result
+		// Deduplication is necessary because there is one podIPInfo entry for each IP family(IPv4 and IPv6), and both may point to the same interface
+		if podIPInfo.MacAddress == "" || seenInterfaces[podIPInfo.MacAddress] {
```
tamilmani1989, line 150: why we need this check `podIPInfo.MacAddress == ""`

cns/NetworkContainerContract.go
```diff
@@ -398,9 +398,10 @@ type NetworkInterfaceInfo struct {
 
 // IPConfiguration contains details about ip config to provision in the VM.
 type IPConfiguration struct {
-	IPSubnet         IPSubnet
-	DNSServers       []string
-	GatewayIPAddress string
+	IPSubnet           IPSubnet
+	DNSServers         []string
+	GatewayIPAddress   string
+	GatewayIPv6Address string
```
tamilmani1989, line 404: can you point where this field is used? is it used by cni

cns/restserver/ipam.go
```diff
@@ -995,38 +996,58 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
...
 	// This map is used to store whether or not we have found an available IP from an NC when looping through the pool
 	ipsToAssign := make(map[string]cns.IPConfigurationStatus)
 
 	// Searches for available IPs in the pool
 	for _, ipState := range service.PodIPConfigState {
-		// check if an IP from this NC is already set side for assignment.
-		if _, ncAlreadyMarkedForAssignment := ipsToAssign[ipState.NCID]; ncAlreadyMarkedForAssignment {
+
+		// get the IPFamily of the current ipState
+		var ipStateFamily cns.IPFamily
```
tamilmani1989: we can assign default to ipv4 and remove else

cns/restserver/ipam.go
```diff
@@ -995,38 +996,58 @@ func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([
...
+		if net.ParseIP(ipState.IPAddress).To4() != nil {
```
tamilmani1989: I see you used Is4() in other place. can we use same here instead of To4()

cns/NetworkContainerContract.go
NihaNallappagari: Yes gatewayip's are used in cilium to setup eth1 routes

azure-ipam/ipam.go
NihaNallappagari: This ensures that scenarios with an empty MAC address is properly handled. Without the check, an empty address fails to parse and throws an error during the parsing step. In fact, the invalid gateway test I added covers this case: when the MAC address is empty, it fails with the expected error "invalid MAC address; failed to parse interface MAC address", if the check is removed.

cns/restserver/ipam.go
NihaNallappagari: Ya I can do that.

cns/restserver/ipam.go
NihaNallappagari: done

PR #3667, opened 2025-05-15
ci: [1/?] Automated version update pipeline

jpayne3506:
**Reason for Change**:

Maintenace for multiple release train versions are slipping. This PR is the first of multiple to leverage variable groups to allow E2E pipelines to use latest managed version where applicable. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


.pipelines/cni/ado-automation/var-pipeline.yaml
```diff
@@ -0,0 +1,211 @@
...
+      #         AZURE_DEVOPS_EXT_PAT: $(System.AccessToken)
+
+
+  - stage: cluster
+    displayName: "Cluster Create"
+    dependsOn:
+      - setup
+    variables:
+      commitID: $[ stagedependencies.setup.env.outputs['EnvironmentalVariables.commitID'] ]
+    jobs:
```
camrynl, line 61: let's add one for 1.32 as well since cilium has a different tag here

camrynl: would it be redundant to have testing for every k8s? or we can go based off of where variable groups actually differ in tags

ex: 1.30 and 1.31 have the same tags so we only setup tests for one OR setup all tags for 1.30 and 1.31

jpayne3506: Talked offiline. Doing all k8s versions as this will allow us to parameterize respective pipelines based off k8s version the cluster was created with.

camrynl (approved):

.pipelines/templates/create-cluster-steps.yaml
```diff
@@ -0,0 +1,55 @@
...
+        make -C ./hack/aks ${{ parameters.clusterType }} \
+        AZCLI=az REGION=${{ parameters.region }} SUB=$(SUB_AZURE_NETWORK_AGENT_BUILD_VALIDATIONS) \
+        CLUSTER=${{ parameters.clusterName }} \
+        VM_SIZE=${{ parameters.vmSize }} VM_SIZE_WIN=${{ parameters.vmSizeWin }} \
+        OS_SKU_WIN=${{ parameters.osSkuWin }} OS=${{ parameters.os }} \
+        LTS=${{ lower(parameters.LTS) }}
+
+        echo "Cluster successfully created"
+    displayName: Cluster - ${{ parameters.clusterType }}
+    continueOnError: ${{ contains(parameters.clusterType, 'dualstack') }}
```
vipul-21, line 55: Why continue on error if we have dualstack ?

jpayne3506: Need to remove this.

PR #3676, opened 2025-05-20
chore: updating cilium nightly charts

camrynl:
**Reason for Change**:
nightly pipeline is experiencing some failures because the charts are too old
updating from 1.14 > 1.17 charts

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/cni/cilium/nightly-release-test.yml
+++ b/.pipelines/cni/cilium/nightly-release-test.yml
@@ -159,7 +159,7 @@ stages:
               scriptType: "bash"
               addSpnToEnvironment: true
               inlineScript: |
-                make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=$(clusterName)-$(commitID)
+                make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=ciliumnightly-$(commitID)
 
                 set -e
                 echo "Run Cilium Connectivity Tests"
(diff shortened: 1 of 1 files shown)


QxBytes (approved):

jpayne3506 (approved):

QxBytes (commented):
Summarizing the conversation with camryn:
- Changes to the overlay pipeline can break the nightly pipeline since the PR check does not include running the cilium nightly pipeline and the nightly eventually calls into the overlay step template for example
- As long as we create a new cilium-config-hubble.yaml with each new version with updated charts, the v${DIR} should pull the correct version
- Eventually all nightly pipeline-specific files should be replaced with v${DIR} syntax where DIR is just the latest available version we have-- right now this PR just mostly copies the files in the cilium/v1.17 folder into the nightly pipeline files to update them
- The nightly pipeline builds against the latest cilium master so as of now v1.18, but as a best effort we use the charts/yamls from v1.17 as the v1.18 charts don't exist yet
- The only manual difference not generated from cilium helm is this change
```
- apiGroups:
  - ""
  resourceNames:
  - cilium-config
  resources:
  - configmaps
  verbs:
  - list
  - watch
  ```
  which is a change required for cilium v1.18 to work.

PR #3678, opened 2025-05-22
ci: azure-ipam patch upgrade testing 

camrynl:
**Reason for Change**:
First pr for patch upgrade testing in CNI release pipeline. The CNI release patch upgrade tests will follow the same flow that we use for cilium patch upgrade tests. 
Create cluster > deploy current images > e2e > deploy upgrade images > e2e > delete cluster

Adding TEST_AZURE_IPAM_VERSION and TEST_CNS_VERSION and new conditions to check these vars in the upgrade scenario. 
(CNS patch upgrade work still needs to be completed)

further documentation/instruction for use will be added when pipeline work is complete

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
+++ b/.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
@@ -11,37 +11,40 @@ parameters:
   hubbleEnabled: false
   dualstackVersion: ""
   cni: "cilium"
+  upgradeScenario: ""
+  upgradeAzureIpam: ""
 
 # Condition confirms that:
 # Previous job has reported Succeeded. Previous job is currently setup which controls variable assignment and we are dependent on its success.
 # CONTROL_CNI either contains 'cniv1' or 'all'. It is not case sensitive
 stages:
-  - stage: create_${{ parameters.name }}
-    condition: and( succeeded(), and( or( contains(variables.CONTROL_CNI, 'cilium') , contains(variables.CONTROL_CNI, 'all') ), or( contains(variables.CONTROL_OS, 'linux'), contains(variables.CONTROL_OS, 'all') ) ) )
-    variables:
-      ${{ if contains(parameters.clusterName, 'rdma') }}:
-        location: $(LOCATION_RDMA)
-      ${{ elseif eq(parameters.arch, 'arm64') }}:
-        location: $(LOCATION_ARM64)
-      ${{ else }}:
-        location: $(LOCATION_AMD64)
-      commitID: $[ stagedependencies.setup.env.outputs['SetEnvVars.commitID'] ]
-    dependsOn:
-      - setup
-      - build_images
-    displayName: "Create Cluster - ${{ parameters.clusterName }}"
-    jobs:
-      - job: create_aks_cluster_with_${{ parameters.name }}
-        pool:
-          name: "$(BUILD_POOL_NAME_DEFAULT)"
-        steps:
-          - template: ../load-test-templates/create-cluster-template.yaml
-            parameters:
-              clusterType: ${{ parameters.clusterType }}
-              clusterName: ${{ parameters.clusterName }}-$(commitID)
-              nodeCount: ${{ parameters.nodeCount }}
-              vmSize: ${{ parameters.vmSize }}
-              region: $(location)
+  - ${{if eq(parameters.upgradeAzureIpam, '')}}:
+    - stage: create_${{ parameters.name }}
+      condition: and( succeeded(), and( or( contains(variables.CONTROL_CNI, 'cilium') , contains(variables.CONTROL_CNI, 'all') ), or( contains(variables.CONTROL_OS, 'linux'), contains(variables.CONTROL_OS, 'all') ) ) )
+      variables:
+        ${{ if contains(parameters.clusterName, 'rdma') }}:
+          location: $(LOCATION_RDMA)
+        ${{ elseif eq(parameters.arch, 'arm64') }}:
+          location: $(LOCATION_ARM64)
+        ${{ else }}:
+          location: $(LOCATION_AMD64)
+        commitID: $[ stagedependencies.setup.env.outputs['SetEnvVars.commitID'] ]
+      dependsOn:
+        - setup
+        - build_images
+      displayName: "Create Cluster - ${{ parameters.clusterName }}"
+      jobs:
+        - job: create_aks_cluster_with_${{ parameters.name }}
+          pool:
+            name: "$(BUILD_POOL_NAME_DEFAULT)"
+          steps:
+            - template: ../load-test-templates/create-cluster-template.yaml
+              parameters:
+                clusterType: ${{ parameters.clusterType }}
+                clusterName: ${{ parameters.clusterName }}-$(commitID)
+                nodeCount: ${{ parameters.nodeCount }}
+                vmSize: ${{ parameters.vmSize }}
+                region: $(location)
 
 # Conditions for below E2E test scenarios confirm that:
 # Pipeline has not been canceled and that the previous job has reports anything other than failure(Succeeded, SuccededWithIssues, Skipped). Previous job is declared by dependsOn:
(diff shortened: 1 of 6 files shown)


.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -60,7 +63,10 @@ stages:
     pool:
       name: "$(BUILD_POOL_NAME_DEFAULT)"
     dependsOn:
-      - create_${{ parameters.name }}
+      - ${{ if and(eq(parameters.upgradeAzureIpam, ''), eq(parameters.upgradeScenario, false)) }}:
```
vipul-21: Is this the only condition where we depend on cluster creation ? 
what if `parameters.upgradeScenario` is "" ? Can we add a table of the combination/expectation of these variables in the description ?

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -118,22 +124,33 @@ stages:
...
-                    IPAM=$(TEST_AZURE_IPAM_VERSION)
-                    echo "TEST_AZURE_IPAM_VERSION is set to $IPAM"
-                  fi
-                  if [ -z "$TEST_CNS_VERSION" ]
+                  if [ ${{parameters.upgradeScenario}} = "true" ]
                   then
-                    echo "TEST_CNS_VERSION is not set, using default value"
-                    CNS=$(make cns-version)
+                    echo "Upgrade scenario is true, using upgrade azure ipam and cns version from pipeline variables"
+                    IPAM=$(UPGRADE_AZURE_IPAM_VERSION)
```
vipul-21: Even if `upgradeAzureIpam` is false, will we use the `UPGRADE_AZURE_IPAM_VERSION` ?

.pipelines/cni/load-test-templates/pod-deployment-template.yaml
```diff
@@ -26,7 +27,10 @@ steps:
   - template: ../../templates/log-template.yaml
     parameters:
       clusterName: ${{ parameters.clusterName }}
-      logType: scaleTest
       os: ${{ parameters.os }}
       cni: ${{ parameters.cni }}
       jobName: ${{ parameters.jobName }}
+      ${{ if eq(parameters.upgradeScenario, 'true') }}:
+        logType: upgradeScaleTest
```
vipul-21: Maybe we pass the `jobname` as `upgrade_scale_test` (Similarly at other places), will that avoid the parameter of `logType` ?

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
camrynl: for this PR, yes. eventually when we have cns or cni patch upgrades, we will need to skip the cluster create there as well.

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
camrynl: oh, will need to add in another condition to check whether or not `UPGRADE_AZURE_IPAM_VERSION` is set because even if `upgradeScenario` is true we may only be testing the upgrade for cns and not azure-ipam

.pipelines/cni/load-test-templates/pod-deployment-template.yaml
camrynl: it will also require us to make some conditional check to alter the `jobname` with `upgrade` 
the `logType` parameter already exists, but we need to give it a new name so the artifacts don't conflict

vipul-21 (approved):

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -169,22 +200,48 @@ stages:
...
-                  else
-                    IPAM=$(TEST_AZURE_IPAM_VERSION)
-                    echo "TEST_AZURE_IPAM_VERSION is set to $IPAM"
-                  fi
-                  if [ -z "$TEST_CNS_VERSION" ]
+                  if [ ${{parameters.upgradeScenario}} = "true" ]
                   then
-                    echo "TEST_CNS_VERSION is not set, using default value"
-                    CNS=$(make cns-version)
+                    echo "Upgrade scenario is true, using upgrade azure ipam and cns version from pipeline variables"
```
QxBytes, line 178: is this basically a copy of the above for the non-hubble case? Is there a way we can extract this logic to a template or similar? If we make a modification to one it could get desynced with the other case?

.pipelines/cni/pipeline.yaml
```diff
@@ -391,6 +403,125 @@ stages:
       arch: amd64
       dualstackVersion: ${CILIUM_DUALSTACK_VERSION}
       cni: "cilium_dualstack"
+      upgradeScenario: false
+
+
+  ## If upgradeScenario is set, redeploy new IPAM version to existing clusters and run tests
+  - ${{if eq(parameters.upgradeScenario, true)}}:
```
QxBytes, line 410: if this is performing the same e2e test as above, would it be possible to make a template that performs the upgrade test where we pass all these parameters and the template automatically creates the before and after upgrade stages? The template could potentially also take in the upgrade scenario as true or false or a pipeline variable/expression for enable/disable. If we add more scenarios in the future I feel like it would be easy to forget to add the upgrade stage

.pipelines/cni/pipeline.yaml
```diff
@@ -418,6 +549,16 @@ stages:
       - windows_overlay_HNS
       - windows19_overlay_HNS
       - setup
+      - ${{if eq(parameters.upgradeScenario, true)}}:
```
QxBytes, line 552: do we create clusters for the upgrade case? It seems like we skip the create stage (in cilium overlay load test template) if upgrade is true?

camrynl: yes, if we are doing the upgrade scenario then the cluster create will be skipped and the old clusters will be reused

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -11,37 +11,39 @@ parameters:
...
-          name: "$(BUILD_POOL_NAME_DEFAULT)"
-        steps:
-          - template: ../load-test-templates/create-cluster-template.yaml
-            parameters:
-              clusterType: ${{ parameters.clusterType }}
-              clusterName: ${{ parameters.clusterName }}-$(commitID)
-              nodeCount: ${{ parameters.nodeCount }}
-              vmSize: ${{ parameters.vmSize }}
-              region: $(location)
+  - ${{if eq(parameters.upgradeScenario, false)}}:
```
vipul-21, line 20: If `upgradeScenario` is `false` we are not creating cluster ? the description says we need to create the cluster for upgrade scenario ?

camrynl: if `upgradeScenario `is `false `we create the cluster, if `upgradeScenario` is `true `we skip this since the cluster is already created

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
camrynl: going to pull the cns/ipam install out into it's own job so we don't need to have a repeated template call

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
camrynl: in the pipeline.yaml I am calling this template and passing the parameter upgradeScenario.  
` - template: cilium/cilium-overlay-load-test-template.yaml`
for each scenario there are two calls to the template, the first contains `upgradeScenario: false`, and the next contains `upgradeScenario: true`

.pipelines/cni/pipeline.yaml
camrynl: will keep this in mind to address in a follow up

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -235,6 +269,7 @@ stages:
               scaleup: ${SCALEUP_CILIUM}
               nodeCount: ${{ parameters.nodeCount }}
               cni: ${{ parameters.cni }}
+              upgradeScenario: ${{ parameters.upgradeScenario }}
```
QxBytes: should we also pass the logType parameter here instead of upgradeScenario, similar to how you did it above?

QxBytes (commented):
I noticed some failures on the CNI Release test pipeline-- are those unrelated to these changes? Also is it possible to run I believe the lsg pipeline since I think it also uses the load-test-templates? Not sure if other pipelines should be run as well but trying to minimize us breaking pipelines that use these templates (fairly confident they shouldn't break but wouldn't hurt to just have it running in the background)

vipul-21 (approved):

PR #3692, opened 2025-05-29
Fix CNS and CNI ETW log interference issue

### LINKED ISSUE #3691: ETW logging conflict between CNS and CNI
**What happened:** 
When enabling ETW logging via Geneva for both CNS and CNI components, only one of the services emits logs successfully. The other service’s logs are either missing or not routed correctly. This appears to be a conflict in the Geneva logging pipeline.

Both CNS and CNI should be able to emit logs concurrently to Geneva using ETW without interfering with each other’s telemetry.

**How to reproduce it:**  
1. Enable Geneva ETW logging for both CNS and CNI on a Windows node.
2. Use the same Geneva profile name (e.g., ACN-Monitoring) for both services.
3. Observe that only one of the services (typically CNS) emits logs to Geneva, while the other (CNI) does not.

Kubernetes (AKS), version details can vary — issue observed in multiple clusters.

Windows

Applicable to Windows kernel versions used in AKS Windows node pools.

- Both CNS and CNI are configured to use the same Geneva profile name: ACN-Monitoring.
- This may cause a conflict in the Geneva Monitoring Agent or ETW session registration.
- A workaround might involve assigning unique profile names per component or isolating their ETW providers.

byte-msft:
**Reason for Change**:

This PR addresses a logging conflict between CNS and CNI when both components attempt to emit ETW logs using the same Geneva profile name. Previously, both services were configured to use ACN-Monitoring, which caused one of the services (typically CNI) to fail to emit logs due to ETW session conflicts.

**Changes**
Introduced separate Geneva profile names:
- ACN-Monitoring-CNS for CNS
- ACN-Monitoring-CNI for CNI

**Issue Fixed**:

Fixes #3691

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/log/logger.go
+++ b/cni/log/logger.go
@@ -17,7 +17,7 @@ var (
 const (
 	maxLogFileSizeInMb = 5
 	maxLogFileCount    = 8
-	etwCNIEventName    = "Azure-CNI"
+	etwCNIEventName    = "AzureCNI"
 	loggingLevel       = zapcore.DebugLevel
 )
 
--- a/cni/Dockerfile
+++ b/cni/Dockerfile
@@ -6,7 +6,7 @@ ARG OS_VERSION
 ARG OS
 
 # mcr.microsoft.com/oss/go/microsoft/golang:1.23-cbl-mariner2.0
-FROM --platform=linux/${ARCH} mcr.microsoft.com/oss/go/microsoft/golang@sha256:bc9d564dc7279c749a9bb1abfa03f872d24f60bf7e37e61f35dda98c2ccec0c4 AS go
+FROM --platform=linux/${ARCH} mcr.microsoft.com/oss/go/microsoft/golang@sha256:f1cb092e89f5f3448b6db87729039c61541cb85747c690f760b3017218d449bb AS go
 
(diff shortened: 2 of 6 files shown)


cni/log/logger.go
```diff
@@ -17,7 +17,7 @@ var (
 const (
 	maxLogFileSizeInMb = 5
 	maxLogFileCount    = 8
-	etwCNIEventName    = "Azure-CNI"
+	etwCNIEventName    = "AzureCNI"
```
rbtr, line 20: does this require downstream changes? are there consumers (like any alerting) which expect the previous event name today?

byte-msft: Hi @rbtr 

Currently we are not using ETW streaming at all, and this dash in the name causes an error while streaming logs to Geneva. 

This is the first step of turning on the logging into a Geneva using etw.

jackieluc (approved):

cni/log/logger.go
msvik, line 20: Curious to know what is this change for?

PR #3693, opened 2025-05-29
fix: use "vEthernet (LoopbackAdapterHostNCConnectivity)" when CNS is …

behzad-mir:
The CNS code currently will call to the  net.InterfaceByName(LoopbackAdapterHostNCConnectivity) to check for the existing Loopback interface when the Loopback adapter has not been created for Apipa HNSNetwork. This change will make CNS to also check for existence of the interface: "vEthernet (LoopbackAdapterHostNCConnectivity)" in case that HNSNetwork Has not been rehydrated, and the loopback interface is leaked.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/hnsclient/hnsclient_windows.go
+++ b/cns/hnsclient/hnsclient_windows.go
@@ -53,6 +53,9 @@ const (
 	// Name of the loopback adapter needed to create Host NC apipa network
 	hostNCLoopbackAdapterName = "LoopbackAdapterHostNCConnectivity"
 
+	// Name of the loopback adapter created by HNS for Host NC apipa network
+	vEthernethostNCLoopbackAdapterName = "vEthernet (" + hostNCLoopbackAdapterName + ")"
+
 	// HNS rehydration issue requires this GW to be different than the loopback adapter ip, so we set it to .2
 	defaultHnsGwIPAddress       = "169.254.128.2"
 	hnsLoopbackAdapterIPAddress = "169.254.128.1"
@@ -301,7 +304,16 @@ func createHostNCApipaNetwork(
 		}
 
 		// Create loopback adapter needed for this HNS network
-		if interfaceExists, _ := networkcontainers.InterfaceExists(hostNCLoopbackAdapterName); !interfaceExists {
+		// We need to fitst check the existence of either "LoopbackAdapterHostNCConnectivity" or the vEthernet(LoopbackAdapterHostNCConnectivity) interfaces
+		// If neither exists, we create the loopback adapter with the specified IP configuration.
+		loopbackInterfaceExists, _ := networkcontainers.InterfaceExists(hostNCLoopbackAdapterName)
+		vethernetLoopbackInterfaceExists, _ := networkcontainers.InterfaceExists(vEthernethostNCLoopbackAdapterName)
+		if loopbackInterfaceExists {
+			logger.Printf("%s already created, skipping loopback interface creation", hostNCLoopbackAdapterName)
+		}
+		if vethernetLoopbackInterfaceExists {
+			logger.Printf("%s already created, skipping loopback interface creation", vEthernethostNCLoopbackAdapterName)
+		} else if !loopbackInterfaceExists && !vethernetLoopbackInterfaceExists {
 			ipconfig := cns.IPConfiguration{
 				IPSubnet: cns.IPSubnet{
 					IPAddress:    hnsLoopbackAdapterIPAddress,


cns/hnsclient/hnsclient_windows.go
```diff
@@ -53,6 +53,9 @@ const (
 	// Name of the loopback adapter needed to create Host NC apipa network
 	hostNCLoopbackAdapterName = "LoopbackAdapterHostNCConnectivity"
 
+	// Name of the loopback adapter needed to search for Host NC apipa network
+	vEthernethostNCLoopbackAdapterName = "vEthernet (LoopbackAdapterHostNCConnectivity)"
```
rbtr, line 57: this has the `hostNCLoopbackAdapterName` substring, should the matcher be `contains` instead of strict equals?

behzad-mir: @rbtr The API that it will end up calling to is net.InterfaceByName(iFaceName) and that will do eaxct match so that's the reason we got the Apipa loopback interface at first place.

cns/hnsclient/hnsclient_windows.go
```diff
@@ -301,7 +304,7 @@ func createHostNCApipaNetwork(
 		}
 
 		// Create loopback adapter needed for this HNS network
-		if interfaceExists, _ := networkcontainers.InterfaceExists(hostNCLoopbackAdapterName); !interfaceExists {
+		if interfaceExists, _ := networkcontainers.InterfaceExists(vEthernethostNCLoopbackAdapterName); !interfaceExists {
```
msvik, line 307: @behzad-mir  @tamilmani1989  Say a loopback adapter has been created but hns network is still pending to be created. If in a case like if AzureContainerNetworking.exe gets invoked again, what would be the behavior? I am thinking it will fail but retries will succeed with this fix. Do we need to worry about such cases where calls can get interleaved?

msvik (dismissed):

cns/hnsclient/hnsclient_windows.go
tamilmani1989: can `vEthernethostNCLoopbackAdapterName` be "vEthernet (" + hostNCLoopbackAdapterName + ")"?

cns/hnsclient/hnsclient_windows.go
tamilmani1989: we have to test that scenario. @behzad-mir we can purposefully return error from cns after loopadapater creation before hns network creation and then retry should succeed

behzad-mir: The second ADD call only happens if HNSNetowrk and Loppback are created since the ADD calls are sequential.
The case that tHe loopback adapter exist but the HNS APipa network get deleted (which is the root cause of this issue) the fix was able to resolve the issue.

santhoshmprabhu (commented):
Can we please add a UT?

santhoshmprabhu (approved):
Approving to expedite fix for Sev2, tests to be added in follow up.

cns/hnsclient/hnsclient_windows.go
```diff
@@ -301,7 +304,16 @@ func createHostNCApipaNetwork(
 		}
 
 		// Create loopback adapter needed for this HNS network
-		if interfaceExists, _ := networkcontainers.InterfaceExists(hostNCLoopbackAdapterName); !interfaceExists {
+		// We need to fitst check the existence of either "LoopbackAdapterHostNCConnectivity" or the vEthernet(LoopbackAdapterHostNCConnectivity) interfaces
```
santhoshmprabhu, line 307: Typo in comment

cns/hnsclient/hnsclient_windows.go
```diff
@@ -301,7 +304,16 @@ func createHostNCApipaNetwork(
...
+		// If neither exists, we create the loopback adapter with the specified IP configuration.
+		loopbackInterfaceExists, _ := networkcontainers.InterfaceExists(hostNCLoopbackAdapterName)
+		vethernetLoopbackInterfaceExists, _ := networkcontainers.InterfaceExists(vEthernethostNCLoopbackAdapterName)
```
santhoshmprabhu, line 310: Let's check this only if the first check returns false.

santhoshmprabhu (approved):

cns/hnsclient/hnsclient_windows.go
behzad-mir: The Veth can only exist if the HNS add the interface and remove the original loopback.

cns/hnsclient/hnsclient_windows.go
behzad-mir: Will be addressed in follow up

rbtr (approved):

PR #3756, opened 2025-06-25
ci: add azure ipam and azure ip masq merger uts to pipeline

QxBytes:
Updates codeowners and adds uts for these go modules to the pr pipeline, combining code coverage as necessary. Previously, azure ipam and azure ip masq merger uts weren't included in make test-all because they were in different modules from the root module. Running make workspace before running the coverage allows us to include these numbers in the overall code coverage.

**Reason for Change**:

See above.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -827,6 +827,7 @@ test-all: test-azure-ipam test-azure-ip-masq-merger test-main ## run all unit te
 
 test-main: 
 	go test -mod=readonly -buildvcs=false -tags "unit" --skip 'TestE2E*' -race -covermode atomic -coverprofile=coverage-main.out $(COVER_PKG)/...
+	go tool cover -func=coverage-main.out
 
 test-integration: ## run all integration tests.
 	AZURE_IPAM_VERSION=$(AZURE_IPAM_VERSION) \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -855,10 +858,10 @@ test-extended-cyclonus: ## run the cyclonus test for npm.
 	cd ..
 
 test-azure-ipam: ## run the unit test for azure-ipam
-	cd $(AZURE_IPAM_DIR) && go test
+	cd $(AZURE_IPAM_DIR) && go test -race -covermode atomic -coverprofile=../coverage-azure-ipam.out && go tool cover -func=../coverage-azure-ipam.out
```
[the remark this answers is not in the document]
QxBytes: this form of the command is only used twice at the moment

Makefile
```diff
@@ -822,9 +823,11 @@ RESTART_CASE ?= false
 # CNI type is a key to direct the types of state validation done on a cluster.
 CNI_TYPE ?= cilium
 
-test-all: ## run all unit tests.
-	go test -mod=readonly -buildvcs=false -tags "unit" --skip 'TestE2E*' -race -covermode atomic -coverprofile=coverage-all.out $(COVER_PKG)/...
-	go tool cover -func=coverage-all.out
+test-all: test-azure-ipam test-azure-ip-masq-merger test-main ## run all unit tests.
+
+test-main: 
+	go test -mod=readonly -buildvcs=false -tags "unit" --skip 'TestE2E*' -race -covermode atomic -coverprofile=coverage-main.out $(COVER_PKG)/...
```
jpayne3506, line 829: Doesn't this test all directories as it is executed from root?

QxBytes: it tests all directories relevant to the root/current module-- it skips over directories like azure-ipam since they have their own go.mod files-- this is essentially the same command as previous and based on the code coverage report it doesn't look like those modules are included: https://msazure.visualstudio.com/One/_build/results?buildId=128333086&view=codecoverage-tab

PR #3768, opened 2025-07-01
fix: update network error msg to match kubelet expectations

isaac-dasan:
Signed-off-by: GitHub <noreply@github.com>

**Reason for Change**:
This is a perf fix. The particular error message is retries with faster backOffInterval than default error messages by the kubelet.

**Issue Fixed**:

**Requirements**:
- [x] uses [conventional commit messages]
- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -17,7 +17,9 @@ import (
 )
 
 var (
-	errMTPNCNotReady            = errors.New("mtpnc is not ready")
+	// In AKS, for kubelet to retry faster it is particularly looking for "network is not ready" error string.
+	// https://github.com/kubernetes/kubernetes/blob/efd2a0d1f514be96a2f012fc3cb40f7c872b4e67/pkg/kubelet/pod_workers.go#L1495C2-L1501C3
+	errMTPNCNotReady            = errors.New("network is not ready - mtpnc is not ready")
 	errInvalidSWIFTv2NICType    = errors.New("invalid NIC type for SWIFT v2 scenario")
 	errInvalidMTPNCPrefixLength = errors.New("invalid prefix length for MTPNC primaryIP, must be 32")
 )


isaac-dasan:
> it looks like from upstream code this will reduce the retry interval from ~10 seconds to ~1 second-- wondering if there are concerns if the cns never comes up or if this fast retry will exhaust resources (ex: x10 more fail logs if cns never comes up)? Was there a particular incident or reasoning for needing the 10 sec improvement?

To clarify this doesn't happen when "CNS doesn't come up" instead this error message is thrown when MTPNC is not found by CNS. Yes this will make 10x more error logs in CNS and CNI, but this is the only way to address perf bottle neck related to swiftv2 podStartupLatency. It is not acceptable for customers to be waiting for a pod for > 30 secs , but today we are at > 50 secs. This fix will shave off 10 to 15 secs and we must take it.

This is in swiftv2 multitenant code path only which is exceedingly small number of clusters in AKS today, so not all CNI error logs will be noisy.

QxBytes (approved):

paulyufan2 (approved):

cns/middlewares/k8sSwiftV2.go
```diff
@@ -17,7 +17,9 @@ import (
 )
 
 var (
-	errMTPNCNotReady            = errors.New("mtpnc is not ready")
+	// In AKS, for kubelet to retry faster it is particularly looking for "network is not ready" error string.
+	// https://github.com/kubernetes/kubernetes/blob/efd2a0d1f514be96a2f012fc3cb40f7c872b4e67/pkg/kubelet/pod_workers.go#L1495C2-L1501C3
+	errMTPNCNotReady            = errors.New("network is not ready - mtpnc is not ready")
```
rbtr, line 22: @isaac-dasan @QxBytes we should have imported the error string and wrapped it here instead of copying, that would protect us (or at least help detect) from drift (here OR in upstream):

```suggestion
	errMTPNCNotReady            = errors.Errorf("%s: mtpnc is not ready", kerrors.NetworkNotReadyErrorMsg)
```

rbtr: ideally upstream would have made this a public typed error that we could wrap and rethrow, and they would use error unwrapping to detect it...but this would be the next best thing since it's just strings

isaac-dasan: @rbtr - the error string is part of a internal package 'kubelet' and not exported - https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/errors.go

rbtr: @isaac-dasan it is certainly exported.

https://github.com/kubernetes/kubernetes/blob/4142fb530ceb25d9965a3e6925be79a03d918cba/pkg/kubelet/errors.go#L23

```go
package kubelet

import "errors"

const (
	// NetworkNotReadyErrorMsg is used to describe the error that network is not ready
	NetworkNotReadyErrorMsg = "network is not ready"
)

var (
	// ErrNetworkUnknown indicates the network state is unknown
	ErrNetworkUnknown = errors.New("network state unknown")
)
```

`NetworkNotReadyErrorMsg` is a public (exported) const. And the `kubelet` package is not marked `internal` to that module, so it is importable.

isaac-dasan: @rbtr - It is certainly importable, but not ideal. Importing k8s.io/kubernetes/pkg/kubelet just for a constant is like:
Bringing an entire cruise ship into your garage because you want the microwave from the kitchen.

We won't be just importing the kubelet package but the entire Kubernetes repo and it's dependecies and it will bloat up the go sum and build and might also break things since the Kubernetes repo has often breaking changes

thatmattlong: I agree with @rbtr here and disagree with the above analogy. It's important that we guard against drift. Importing the module would be small, there isn't much in that module. In future let's import well-known consts.

rbtr: I am not worried about the size of the go.sum, and nor should you be. That's not for humans.
We already depend on significant chunks of the Kubernetes project and have to manage version compatibility for those dependencies. We know how to do that. (But we would only be _importing_ the kubelet package here.)

I don't understand why we're bike-shedding this. It is importable. The dependency costs you nothing. Referencing the upstream type improves understandability and maintainability and reduces potential future breakages. Why not?

isaac-dasan: I updated the other PR #3776 
@rbtr - you are correct about golang being clever about the packaging of the CNS binaries and not adding unused libraries. The binary size remained at 1.1M before and after.
But here is the interesting thing - `du -sh $(go env GOPATH)/pkg/mod` gave me **491M** for master and **2.4G** for my branch. For developers this will cost more local memory footprint.

isaac-dasan: @rbtr , @thatmattlong 
**Does this truly prevent string drift?** 
No. Here is why:
- The current Kubernetes package version (1.30.14) that is referenced in our go.mod file has "network is not ready" string.
- Let's say kubernetes updates the string in 1.34 pkg. We don't actually move in lock step with kubernetes package version (i.e we don't update all our references to 1.34), we only test our existing CNS code to verify everything works and release CNS 1.8.x. For example device plugin was added with 1.30.7 pkg reference and it is still the same even though CNS 1.7.x is mapped to k8s 1.33.x.

In this particular case the perf bug will reappear and only after that we will upgrade the go.mod references, which would be the same work flow if we had used a hard coded string in our repo. 

 `The dependency costs you nothing` - is not true. We are parking the cruise ship in each developer's devbox.

isaac-dasan: I created [issue ](https://github.com/kubernetes/apimachinery/issues/189) in upstream

PR #3772, opened 2025-07-03
ci: cleanup Cilium cli and connectivity test usage

jpayne3506:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/networkobservability/pipeline.yaml
+++ b/.pipelines/networkobservability/pipeline.yaml
@@ -113,14 +113,9 @@ stages:
           displayName: "Wait for all pods to be running"
           retryCountOnTaskFailure: 3
 
-        - script: |
-              echo "Run Cilium Connectivity Tests"
-              cilium status
-              cilium connectivity test --connect-timeout 4s --request-timeout 30s --test '!pod-to-pod-encryption,!node-to-node-encryption,!check-log-errors'
-          retryCountOnTaskFailure: 3
-          name: "ciliumConnectivityTests"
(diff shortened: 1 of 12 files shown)


camrynl (approved):

.pipelines/networkobservability/pipeline.yaml
```diff
@@ -113,14 +113,9 @@ stages:
...
-              echo "Run Cilium Connectivity Tests"
-              cilium status
-              cilium connectivity test --connect-timeout 4s --request-timeout 30s --test '!pod-to-pod-encryption,!node-to-node-encryption,!check-log-errors'
-          retryCountOnTaskFailure: 3
-          name: "ciliumConnectivityTests"
-          displayName: "Run Cilium Connectivity Tests"
-          enabled: true
+        - template: ../templates/cilium-connectivity-tests.yaml
+          parameters:
+            skipTests: '!pod-to-pod-encryption,!node-to-node-encryption,!check-log-errors,!to-fqdns'
```
QxBytes, line 118: is skipping to-fqdns intentional here? it doesn't seem to be skipped before. Or will it essentially be the same behavior (skip) if cilium doesn't detect fqdn in this stage?

jpayne3506: There is an upstream change that I couldn't find that is no longer checking if L7proxy is enabled or disabled. Thus, we are defaulting to skipping the test. It happens on cli versions >= 0.18.4

PR #3776, opened 2025-07-04
fix: update error message for MTPNC not found and not ready in CNS

isaac-dasan:
**Reason for Change**:
This is a perf fix. The particular error message "network is not ready" will make kubelet retry with faster backOffInterval than default error messages. 
https://github.com/kubernetes/kubernetes/blob/efd2a0d1f514be96a2f012fc3cb40f7c872b4e67/pkg/kubelet/pod_workers.go#L1495C2-L1501C3

**Requirements**:
- [x] includes documentation
- [x] adds unit tests

**Notes**:


cns/middlewares/k8sSwiftV2_linux_test.go
```diff
@@ -39,6 +41,8 @@ var (
 
 	testPod9GUID = "2006cad4-e54d-472e-863d-c4bac66200a7"
 	testPod9Info = cns.NewPodInfo("2006cad4-eth0", testPod9GUID, "testpod9", "testpod9namespace")
+
+	kubeletExpectedString = "network is not ready"
```
MikeZappa87: This could be named better, its vague and hard to infer usage. Something like 'kubeletNetworkNotReadyMsg' relays to the reader of the code the who and what at least.

tamilmani1989:
@isaac-dasan  title reads mtpnc not found but we also cover dhcp discover failed case. I would like to have separate PR for returning network not ready if dhcp discover failed

cns/middlewares/k8sSwiftV2.go
```diff
@@ -13,17 +13,21 @@ import (
...
+	"k8s.io/kubernetes/pkg/kubelet"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var (
-	// In AKS, for kubelet to retry faster it is particularly looking for "network is not ready" error string.
-	// https://github.com/kubernetes/kubernetes/blob/efd2a0d1f514be96a2f012fc3cb40f7c872b4e67/pkg/kubelet/pod_workers.go#L1495C2-L1501C3
-	errMTPNCNotReady            = errors.New("network is not ready - mtpnc is not ready")
+	errMTPNCNotReady            = errors.New(kubelet.NetworkNotReadyErrorMsg + " - mtpnc is not ready")
+	errMTPNCNotFound            = errors.New(kubelet.NetworkNotReadyErrorMsg + " - mtpnc not found")
```
rbtr: these aren't exposed so we don't get any value out of them being predefined errors 🤔
maybe we should remove them and then wrap and rethrow inline?
```go
return errors.Wrap(err, kubelet.NetworkNotReadyErrorMsg)
```

cns/middlewares/k8sSwiftV2.go
```diff
@@ -13,17 +13,21 @@ import (
...
 	errInvalidSWIFTv2NICType    = errors.New("invalid NIC type for SWIFT v2 scenario")
 	errInvalidMTPNCPrefixLength = errors.New("invalid prefix length for MTPNC primaryIP, must be 32")
 )
 
+var getMTPNCNotFoundErr = func(err error) error {
+	return fmt.Errorf("%w: %v", errMTPNCNotFound, err.Error())
+}
+
```
rbtr: this indirection makes it less obvious what's happening, and the helper func takes up more space than having the fmt inline. I would get rid of it

cns/middlewares/k8sSwiftV2.go
```diff
@@ -74,7 +78,7 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
 	mtpnc := v1alpha1.MultitenantPodNetworkConfig{}
 	mtpncNamespacedName := k8stypes.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
 	if err := k.Cli.Get(ctx, mtpncNamespacedName, &mtpnc); err != nil {
-		return nil, errors.Wrapf(err, "failed to get pod's mtpnc from cache")
+		return nil, getMTPNCNotFoundErr(err)
```
rbtr: `Get` could return more error types than `NotFound`, right? that would be misleading while debugging

cns/middlewares/k8sSwiftV2.go
timraymond: Also the error handling is not right. Using `%v` for error strings destroys information. It seems like the intent here is to just wrap the error, so `errors.Wrap(err, "getting MTPNC")` will suffice. At the point the error occurs, the only things we know are:

1. There was an error
2. It happened while we were getting the MTPNC.

These are the only information that should be communicated--saying "not found" is a false implication that could mislead debugging efforts.

cns/middlewares/k8sSwiftV2.go
timraymond: I actually find it harmful to constantize error wrapping strings. When you're grepping for that string, it obscures where it's used. It also encourages reuse, which should not happen (since it also makes grepping for it harder).

isaac-dasan: We are using these errors in UTs for asserting. So it is better this way

cns/middlewares/k8sSwiftV2.go
isaac-dasan: Removed this func in latest iteration

cns/middlewares/k8sSwiftV2.go
isaac-dasan: replaced it with GetErr

rbtr (dismissed):

common/utils.go
```diff
@@ -30,6 +30,9 @@ const (
 	FiveSeconds                     = 5 * time.Second
 	JsonContent                     = "application/json; charset=UTF-8"
 	ContentType                     = "Content-Type"
+	// In AKS, for kubelet to retry faster it is particularly looking for "network is not ready" error string.
+	// https://github.com/kubernetes/kubernetes/blob/efd2a0d1f514be96a2f012fc3cb40f7c872b4e67/pkg/kubelet/pod_workers.go#L1495C2-L1501C3
+	KubeletNetworkNotReadyErrMsg = "network is not ready"
```
QxBytes: is this being used anywhere?

santhoshmprabhu (approved):

isaac-dasan:
@microsoft-github-policy-service rerun

isaac-dasan:
@microsoft-github-policy-service rerun

isaac-dasan:
@microsoft-github-policy-service rerun

PR #3779, opened 2025-07-08
feat: add azure iptables monitor binary and makefile changes

QxBytes:
**Reason for Change**:

Creates a new image similar to azure-ip-masq-merger which monitors the iptables rules on a machine and patches a node label if an iptables rule that does not match the provided regex patterns is found. See the README for details. The binary/image is meant to be run in a k8 environment, have the node name available as an environment variable, and the regex patterns in the specified folder. Pipeline modifications will be made in a future PR.

**Issue Fixed**:

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Notes**:


azure-iptables-monitor/iptables_monitor.go
```diff
@@ -0,0 +1,236 @@
...
+				break
+			}
+		}
+		if !ruleMatched {
+			klog.Infof("Unexpected rule: %s", rule)
+			foundUnexpectedRules = true
+		}
+	}
+
+	return foundUnexpectedRules
```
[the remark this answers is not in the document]
QxBytes: want to print out all the unexpected rules to the log

Makefile
```diff
@@ -230,6 +238,10 @@ azure-npm-binary:
 azure-ip-masq-merger-binary:
 	cd $(AZURE_IP_MASQ_MERGER_DIR) && CGO_ENABLED=0 go build -v -o $(AZURE_IP_MASQ_MERGER_BUILD_DIR)/azure-ip-masq-merger$(EXE_EXT) -ldflags "-X main.version=$(AZURE_IP_MASQ_MERGER_VERSION)" -gcflags="-dwarflocationlists=true"
 
+# Build the azure-iptables-monitor binary.
```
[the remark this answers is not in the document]
QxBytes: keeping as is to maintain consistency

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -0,0 +1,295 @@
...
+		// update label based on whether user iptables rules were found
+		err = patchLabel(dynamicClient, userIPTablesRulesFound, currentNodeName)
+		if err != nil {
+			klog.Errorf("failed to patch label: %v", err)
+		} else {
+			klog.V(2).Infof("Successfully updated label for %s: %s=%v", currentNodeName, label, userIPTablesRulesFound)
+		}
+
+		if *sendEvents && userIPTablesRulesFound {
+			err = createNodeEvent(clientset, currentNodeName, "UnexpectedIPTablesRules", "Node has unexpected iptables rules", corev1.EventTypeWarning)
```
santhoshmprabhu, line 288: We can keep this for the first iteration, but I think we should distinguish between two cases -
1. Attempted rule installation that is blocked by our iptables block, in which case we raise an event
2. We find that a rule installation has succeeded, while ebpf host routing is enabled (which is what this line is currently checking). Then the node is "broken" by some definition, so we could create a custom node condition rather than an event.

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -0,0 +1,295 @@
...
+		for _, pattern := range compiledPatterns {
+			if pattern.MatchString(rule) {
+				klog.V(3).Infof("MATCHED: '%s' -> pattern: '%s'", rule, pattern.String())
+				ruleMatched = true
+				break
+			}
+		}
+		if !ruleMatched {
+			klog.Infof("Unexpected rule: %s", rule)
+			foundUnexpectedRules = true
```
santhoshmprabhu, line 189: return?

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -0,0 +1,295 @@
...
+	return allRules, nil
+}
+
+// hasUnexpectedRules checks if any rules in currentRules don't match any of the allowedPatterns
+// Returns true if there are unexpected rules, false if all rules match expected patterns
+func hasUnexpectedRules(currentRules, allowedPatterns []string) bool {
+	foundUnexpectedRules := false
+
+	// compile regex patterns
+	compiledPatterns := make([]*regexp.Regexp, 0, len(allowedPatterns))
```
santhoshmprabhu, line 167: We could make this a bit more efficient by watching the configPath for files, and having a cache of compiled patterns. We update the patterns when the config changes.

Then, the periodic check can simply match the current patterns against the currently observed set of rules.

Given that we're also trying to get not just the logic done in the PR but also the pieces around building and publishing the image, I think we can go with what we currently have for the first iteration.

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -0,0 +1,295 @@
...
+		return fmt.Errorf("failed to create event for node %s: %w", nodeName, err)
+	}
+
+	klog.V(2).Infof("Created event for node %s: %s - %s", nodeName, reason, message)
+	return nil
+}
+
+type IPTablesClient interface {
+	ListChains(table string) ([]string, error)
+	List(table, chain string) ([]string, error)
```
santhoshmprabhu, line 139: Please verify that the implementation we use lists the rule fully, including fields like matches on marks. As I mentioned on one of the calls, iptables-save doesn't seem to do it, at least not by default.

santhoshmprabhu:
Let's also consider monitoring for the following (as future follow ups):
1. nftables rulesets that don't use the iptables tables
2. iptables-legacy

azure-iptables-monitor/iptables_monitor.go
QxBytes: For (1) I don't have an implementation in this code since I'm not sure how the blocking logic is supposed to report it-- is it via a file, for example? I'd probably add it once we have the binary ready and the method of communication set.
For (2) ah I see what you mean I'll look into making this change-- If we make the node not ready (or similar) with a condition, I wonder if that will affect the upgrade/downgrade flow since pods may not be able to be scheduled

azure-iptables-monitor/iptables_monitor.go
QxBytes: Wanted to always log the rules that don't match so if we (or the user) wanted to see what was unexpected they could pull up the logs-- if it runs every five mins I don't think looping through all the rules would be too big of a performance hit?

azure-iptables-monitor/iptables_monitor.go
QxBytes: sure we can look into it if performance starts becoming an issue

azure-iptables-monitor/iptables_monitor.go
santhoshmprabhu: maybe add this as a comment in the code

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -0,0 +1,296 @@
...
+	"k8s.io/component-base/logs"
+	"k8s.io/component-base/version/verflag"
+	"k8s.io/klog/v2"
+)
+
+// Version is populated by make during build.
+var version string
+
+var (
+	configPath    = flag.String("input", "/etc/config/", "Name of the directory with the allowed regex files")
```
camrynl, line 30: along with the monitor binary will we have to add these allowed regex files? how are the allowed rules determined?

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -0,0 +1,296 @@
...
+	checkInterval = flag.Int("interval", 300, "How often to check iptables rules (in seconds)")
```
camrynl, line 31: how did you come up with this interval?

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -0,0 +1,296 @@
...
+		return nil, fmt.Errorf("failed to open file %s: %w", filename, err)
+	}
+	defer file.Close()
+
+	var lines []string
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		// Skip empty lines
+		if line != "" {
```
camrynl, line 57: it seems like we add any line that is nonempty, does it matter if the file contains invalid regexes?

QxBytes: during the compilation stage in `hasUnexpectedRules`, if there is an error, we skip the bad rule and log it

azure-iptables-monitor/iptables_monitor.go
QxBytes: during the design review, five minutes came up as a good interval length

azure-iptables-monitor/iptables_monitor.go
QxBytes: yes-- the allowed rules are ones that we have "accounted for"/expect. Once we have mimicked the rule's functionality post-upgrade, we will add it to the allowed rules.

azure-iptables-monitor/iptables_monitor.go
QxBytes: discussed that we only need as much detail to identify the rule-- it is okay if the regex lets some rules through

santhoshmprabhu (approved):

PR #3790, opened 2025-07-10
Delegated nc update

NihaNallappagari:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/imds/client.go
+++ b/cns/imds/client.go
@@ -46,12 +46,14 @@ func RetryAttempts(attempts uint) ClientOption {
 const (
 	vmUniqueIDProperty    = "vmId"
 	imdsComputePath       = "/metadata/instance/compute"
+	imdsNetworkPath       = "/metadata/instance/network"
 	imdsComputeAPIVersion = "api-version=2021-01-01"
 	imdsFormatJSON        = "format=json"
 	metadataHeaderKey     = "Metadata"
 	metadataHeaderValue   = "true"
 	defaultRetryAttempts  = 3
 	defaultIMDSEndpoint   = "http://169.254.169.254"
+	ncVersion             = "ncVersion"
 )
 
 var (
@@ -79,7 +81,7 @@ func NewClient(opts ...ClientOption) *Client {
 func (c *Client) GetVMUniqueID(ctx context.Context) (string, error) {
 	var vmUniqueID string
 	err := retry.Do(func() error {
-		computeDoc, err := c.getInstanceComputeMetadata(ctx)
+		computeDoc, err := c.getInstanceMetadata(ctx, imdsComputePath)
 		if err != nil {
 			return errors.Wrap(err, "error getting IMDS compute metadata")
 		}
@@ -102,10 +104,50 @@ func (c *Client) GetVMUniqueID(ctx context.Context) (string, error) {
 	return vmUniqueID, nil
 }
 
-func (c *Client) getInstanceComputeMetadata(ctx context.Context) (map[string]any, error) {
+func (c *Client) GetNCVersionsFromIMDS(ctx context.Context) (map[string]string, error) {
+	var networkData NetworkMetadata
+	err := retry.Do(func() error {
+		networkMetadata, err := c.getInstanceMetadata(ctx, imdsNetworkPath)
+		if err != nil {
+			return errors.Wrap(err, "error getting IMDS network metadata")
+		}
+
+		// Try to parse the network metadata as the expected structure
+		// Convert the map to JSON and back to properly unmarshal into struct
+		jsonData, err := json.Marshal(networkMetadata)
+		if err != nil {
+			return errors.Wrap(err, "error marshaling network metadata")
+		}
+
+		if err := json.Unmarshal(jsonData, &networkData); err != nil {
+			return errors.Wrap(err, "error unmarshaling network metadata")
+		}
+		return nil
+	}, retry.Context(ctx), retry.Attempts(c.config.retryAttempts), retry.DelayType(retry.BackOffDelay))
+	if err != nil {
+		return nil, err
+	}
+
+	ncVersions := make(map[string]string)
+	for _, iface := range networkData.Interface {
+		// IMDS only returns compartment fields (interfaceCompartmentId, interfaceCompartmentVersion)
+		// We map these to NC ID and NC version concepts
+		// Standard fields (ncId, ncVersion) are ignored even if present
+		ncId := iface.InterfaceCompartmentId
+		ncVersion := iface.InterfaceCompartmentVersion
+
+		if ncId != "" {
+			ncVersions[ncId] = ncVersion
+		}
+	}
+
+	return ncVersions, nil
+}
+
+func (c *Client) getInstanceMetadata(ctx context.Context, imdsComputePath string) (map[string]any, error) {
 	imdsComputeURL, err := url.JoinPath(c.config.endpoint, imdsComputePath)
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to build path to IMDS compute metadata")
+		return nil, errors.Wrap(err, "unable to build path to IMDS metadata for path"+imdsComputePath)
 	}
 	imdsComputeURL = imdsComputeURL + "?" + imdsComputeAPIVersion + "&" + imdsFormatJSON
 
@@ -132,4 +174,17 @@ func (c *Client) getInstanceComputeMetadata(ctx context.Context) (map[string]any
 	}
 
 	return m, nil
+}
+
+// NetworkInterface represents a network interface from IMDS
+type NetworkInterface struct {
+	MacAddress string `json:"macAddress"`
+	// IMDS only returns compartment fields - these are mapped to NC ID and NC version concepts
+	InterfaceCompartmentId      string `json:"interfaceCompartmentId,omitempty"`
+	InterfaceCompartmentVersion string `json:"interfaceCompartmentVersion,omitempty"`
+}
+
+// NetworkMetadata represents the network metadata from IMDS
+type NetworkMetadata struct {
+	Interface []NetworkInterface `json:"interface"`
 }
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -25,6 +25,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// Known API names we care about
+	nmAgentSwiftV2API = "SwiftV2DhcpRehydrationFromGoalState"
+)
+
 // This file contains the internal functions called by either HTTP APIs (api.go) or
 // internal APIs (definde in internalapi.go).
 // This will be used internally (say by RequestController in case of AKS)
@@ -167,6 +172,7 @@ func (service *HTTPRestService) SyncHostNCVersion(ctx context.Context, channelMo
 	service.Lock()
 	defer service.Unlock()
 	start := time.Now()
+
 	programmedNCCount, err := service.syncHostNCVersion(ctx, channelMode)
 	// even if we get an error, we want to write the CNI conflist if we have any NC programmed to any version
 	if programmedNCCount > 0 {
@@ -202,6 +208,7 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
 			logger.Errorf("Received err when change nc version %s in containerstatus to int, err msg %v", service.state.ContainerStatus[idx].CreateNetworkContainerRequest.Version, err)
 			continue
 		}
+		logger.Printf("NC %s: local NC version %d, DNC NC version %d", service.state.ContainerStatus[idx].ID, localNCVersion, dncNCVersion)
 		// host NC version is the NC version from NMAgent, if it's smaller than NC version from DNC, then append it to indicate it needs update.
 		if localNCVersion < dncNCVersion {
 			outdatedNCs[service.state.ContainerStatus[idx].ID] = struct{}{}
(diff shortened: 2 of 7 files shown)


cns/imds/client.go
```diff
@@ -102,10 +104,50 @@ func (c *Client) GetVMUniqueID(ctx context.Context) (string, error) {
 	return vmUniqueID, nil
 }
 
-func (c *Client) getInstanceComputeMetadata(ctx context.Context) (map[string]any, error) {
+func (c *Client) GetNCVersionsFromIMDS(ctx context.Context) (map[string]string, error) {
```
timraymond: Naming: `*FromIMDS` is implied since this is an IMDS client.

```suggestion
func (c *Client) GetNCVersions(ctx context.Context) (map[string]string, error) {
```

cns/imds/client.go
```diff
@@ -102,10 +104,50 @@ func (c *Client) GetVMUniqueID(ctx context.Context) (string, error) {
...
+			return errors.Wrap(err, "error unmarshaling network metadata")
+		}
+		return nil
+	}, retry.Context(ctx), retry.Attempts(c.config.retryAttempts), retry.DelayType(retry.BackOffDelay))
+	if err != nil {
+		return nil, err
+	}
+
+	ncVersions := make(map[string]string)
+	for _, iface := range networkData.Interface {
```
timraymond: When writing a client at this level, don't try to assume how it will be used by the caller. Faithful reproduction of the data in the language's types should be the goal. This makes the client more reusable, if it returns `([]NetworkInterface, error)`. Leave it up to the caller to transform the data as it fits the business logic needs.

cns/imds/client.go
```diff
@@ -133,3 +175,16 @@ func (c *Client) getInstanceComputeMetadata(ctx context.Context) (map[string]any
 
 	return m, nil
 }
+
+// NetworkInterface represents a network interface from IMDS
+type NetworkInterface struct {
+	MacAddress string `json:"macAddress"`
```
timraymond: Use https://pkg.go.dev/net#HardwareAddr instead of `string`

cns/imds/client.go
NihaNallappagari: Addressed the comment

cns/imds/client.go
NihaNallappagari: Addressed comment. Thanks

cns/imds/client.go
```diff
@@ -102,10 +103,37 @@ func (c *Client) GetVMUniqueID(ctx context.Context) (string, error) {
...
+		return nil
+	}, retry.Context(ctx), retry.Attempts(c.config.retryAttempts), retry.DelayType(retry.BackOffDelay))
+	if err != nil {
+		return nil, errors.Wrap(err, "external call failed")
+	}
+
+	return networkData.Interface, nil
+}
+
+func (c *Client) getInstanceMetadata(ctx context.Context, imdsComputePath string) (map[string]any, error) {
```
kadevu: Change imdsComputePath to imdsMetadataPath or something similar

kadevu: Change references to the word "Compute" in this function

cns/imds/client.go
```diff
@@ -133,3 +161,15 @@ func (c *Client) getInstanceComputeMetadata(ctx context.Context) (map[string]any
 
 	return m, nil
 }
+
+// NetworkInterface represents a network interface from IMDS
+type NetworkInterface struct {
```
kadevu, line 234: Dont you need any other fields from the NetworkInterface?

cns/imds/client.go
```diff
@@ -102,10 +103,37 @@ func (c *Client) GetVMUniqueID(ctx context.Context) (string, error) {
 	return vmUniqueID, nil
 }
 
-func (c *Client) getInstanceComputeMetadata(ctx context.Context) (map[string]any, error) {
+func (c *Client) GetNCVersions(ctx context.Context) ([]NetworkInterface, error) {
```
kadevu: You are returning an array of NetworkInterface here, rename the method to GetNetworkInterfaces or something similar

cns/restserver/internalapi.go
```diff
@@ -221,18 +226,36 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
+	}
+
 	nmaNCs := map[string]string{}
 	for _, nc := range ncVersionListResp.Containers {
 		nmaNCs[strings.ToLower(nc.NetworkContainerID)] = nc.Version
 	}
-	hasNC.Set(float64(len(nmaNCs)))
+
+	// Consolidate both maps - NMA takes precedence, IMDS as fallback
+	consolidatedNCs := make(map[string]string)
```
kadevu: rename to programmedNCVersionMap

cns/restserver/internalapi.go
```diff
@@ -254,15 +277,17 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
+			logger.Errorf("NC version from consolidated sources is decreasing: have %d, got %d", localNCVersion, consolidatedNCVersion)
 			continue
 		}
 		if channelMode == cns.CRD {
-			service.MarkIpsAsAvailableUntransacted(ncInfo.ID, nmaNCVersion)
+			service.MarkIpsAsAvailableUntransacted(ncInfo.ID, consolidatedNCVersion)
 		}
-		logger.Printf("Updating NC %s host version from %s to %s", ncID, ncInfo.HostVersion, nmaNCVersionStr)
-		ncInfo.HostVersion = nmaNCVersionStr
+		//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
```
kadevu, line 292: why is nolint added?

cns/restserver/internalapi.go
```diff
@@ -221,18 +226,36 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
+	consolidatedNCs := make(map[string]string)
+	for ncID, version := range nmaNCs {
+		consolidatedNCs[ncID] = version
+	}
+	for ncID, version := range imdsNCVersions {
+		if _, exists := consolidatedNCs[ncID]; !exists {
+			consolidatedNCs[strings.ToLower(ncID)] = version
+		}
+	}
+	hasNC.Set(float64(len(consolidatedNCs)))
```
kadevu: where is hasNC used?

cns/restserver/internalapi.go
```diff
@@ -634,3 +659,59 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
 }
+
+// checkNMAgentAPISupport checks if specific APIs are supported by NMAgent using the existing client
+func (service *HTTPRestService) checkNMAgentAPISupport(ctx context.Context) (swiftV2Support bool, err error) {
+	// Use the existing NMAgent client instead of direct HTTP calls
+	if service.nma == nil {
+		return false, errors.New("NMAgent client is not available")
+	}
+
+	apis, err := service.nma.SupportedAPIs(ctx)
```
kadevu: Is SupportedAPIs not being called anywhere else in CNS repo? If it is can you use that method instead

cns/restserver/internalapi.go
```diff
@@ -634,3 +659,59 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
 func (service *HTTPRestService) SetVFForAccelnetNICs() error {
 	return service.setVFForAccelnetNICs()
 }
+
+// checkNMAgentAPISupport checks if specific APIs are supported by NMAgent using the existing client
+func (service *HTTPRestService) checkNMAgentAPISupport(ctx context.Context) (swiftV2Support bool, err error) {
```
kadevu: Have one method only to get the supportedapis from nma and the another method to check if swiftv2 nc programming support is enabled

cns/imds/client.go
NihaNallappagari: No. Version and id is enough
Update: I will add the macAddress as discussed

cns/restserver/internalapi.go
NihaNallappagari: Map represents the NC's that are programmed on NMA, not only the version. I would prefer something like nmaProgrammedNCs

cns/restserver/internalapi.go
NihaNallappagari: This is preexisting code, I just update the argument. Looks like .Set() reports the count of nc's to external sources.for things like debugging, monitoring health, understanding the scale

cns/restserver/internalapi.go
NihaNallappagari: logger.Printf is deprecated and causes linting errors in the newly added code, as the linter only scans the modified changes. However, the existing code in the CNS repo, including this file widely uses logger.Printf for logging. Addressing the deprecation consistently would require broader changes throughout the file. So suppressed the linting error and this can be a part of bigger change throughout the repo

cns/imds/client.go
NihaNallappagari: Updated

cns/restserver/internalapi.go
NihaNallappagari: Http handler [`nmAgentSupportedApisHandler`](https://github.com/Azure/azure-container-networking/blob/master/cns/restserver/api.go#L1223) is calling nmagent client, this looks more like used for external purpose where calls should be http post and whenever secure call happens. I'm making call to nma client instead

cns/imds/client.go
```diff
@@ -133,3 +161,15 @@ func (c *Client) getInstanceComputeMetadata(ctx context.Context) (map[string]any
...
+// NetworkInterface represents a network interface from IMDS
+type NetworkInterface struct {
+	// IMDS returns compartment fields - these are mapped to NC ID and NC version
+	InterfaceCompartmentID      string `json:"interfaceCompartmentID,omitempty"`
+	InterfaceCompartmentVersion string `json:"interfaceCompartmentVersion,omitempty"`
+}
+
+// NetworkMetadata represents the network metadata from IMDS
+type NetworkMetadata struct {
+	Interface []NetworkInterface `json:"interface"`
```
kadevu, line 242: Rename it to Interfaces as its an array

cns/imds/client.go
```diff
@@ -102,14 +103,41 @@ func (c *Client) GetVMUniqueID(ctx context.Context) (string, error) {
...
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, imdsComputeURL, http.NoBody)
+	return networkData.Interface, nil
+}
+
+func (c *Client) getInstanceMetadata(ctx context.Context, imdsMetadataPath string) (map[string]any, error) {
+	imdsRequestURL, err := url.JoinPath(c.config.endpoint, imdsMetadataPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to build path to IMDS metadata for path"+imdsMetadataPath)
+	}
+	imdsRequestURL = imdsRequestURL + "?" + imdsComputeAPIVersion + "&" + imdsFormatJSON
```
kadevu: The api version could be different. Check with Hitesh on the api version. Also rename the variable to not use the word 'Compute'.

cns/imds/client.go
kadevu: Add macaddress as well to the payload and also privateIpAddress of IPv4

cns/restserver/internalapi.go
```diff
@@ -25,6 +25,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// Known API names we care about
+	nmAgentSwiftV2API = "SwiftV2DhcpRehydrationFromGoalState"
```
kadevu: Confirm the name of the api with Hitesh

kadevu: rename nmAgentSwiftV2API to nmAgentSwiftV2IMDSMetadataSupportAPI

cns/restserver/internalapi.go
NihaNallappagari: Updated

cns/restserver/internalapi.go
```diff
@@ -221,18 +226,36 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
 	}
-	hasNC.Set(float64(len(nmaNCs)))
+
+	// Consolidate both nc's from NMA and IMDS calls
+	nmaProgrammedNCs := make(map[string]string)
+	for ncID, version := range nmaNCs {
+		nmaProgrammedNCs[ncID] = version
+	}
+	for ncID, version := range imdsNCVersions {
+		if _, exists := nmaProgrammedNCs[ncID]; !exists {
```
kadevu, line 248: Ideally this should never happen that the ncid in IMDS NetworkMetadata is same as the one which is returned by nma for swiftv1 ncs. Can you add a warning message here so that we can add some monitor incase this happens.

cns/imds/client.go
NihaNallappagari: Added the mac address as discussed

cns/restserver/internalapi.go
NihaNallappagari: Updated warning

cns/restserver/internalapi.go
```diff
@@ -634,3 +664,87 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
+
+	apis, err := service.nma.SupportedAPIs(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get supported APIs from NMAgent client")
+	}
+
+	return apis, nil
+}
+
+func (service *HTTPRestService) isSwiftV2NCSupported(ctx context.Context) bool {
```
miguelgoms: can we name this consistently with the feature flag we are checking? (IsSwiftV2NCGoalStateSupported or IsSwiftV2NCGoalStateDetailSupported)
this name is misleading if I'm looking at the apis without context - swiftv2 NCs are supported by NMA, what is new is returning the swiftv2 NC goal state through IMDS.

miguelgoms (approved):

cns/restserver/internalapi.go
```diff
@@ -634,3 +664,87 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
+		//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
+		logger.Errorf("Failed to get network interfaces from IMDS: %v", err)
+		return make(map[string]string), nil
+	}
+
+	// Build ncs map from the network interfaces
+	ncs := make(map[string]string)
+	for _, iface := range networkInterfaces {
+		// IMDS returns interfaceCompartmentID, as nc id guid has different context on nma. We map these to NC ID
+		ncID := iface.InterfaceCompartmentID
```
kadevu, line 715: Can you log the NCID and the corresponding MAC here which will help us in troubleshooting incase the NCID is wrong or is associated with the wrong MAC

cns/restserver/internalapi.go
```diff
@@ -634,3 +664,87 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
+// GetIMDSNCs gets NC versions from IMDS and returns them as a map
+func (service *HTTPRestService) GetIMDSNCs(ctx context.Context) (map[string]string, error) {
+	imdsClient := service.imdsClient
+	if imdsClient == nil {
+		//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
+		logger.Errorf("IMDS client is not available")
+		return make(map[string]string), nil
+	}
+	// Check NMAgent API support for SwiftV2, if it fails return empty map assuming support might not be available in that nma build
+	if !service.isSwiftV2NCGoalStateSupported(ctx) || !service.isNCDetailsAPIExists(ctx) {
```
kadevu: Do we even need the !service.isSwiftV2NCGoalStateSupported(ctx) check?

cns/restserver/internalapi.go
NihaNallappagari: As discussed, removed the supportedAPI function

cns/restserver/internalapi.go
NihaNallappagari: As discussed, removed the isSwiftV2NCGoalStateSupported check

cns/restserver/internalapi.go
```diff
@@ -221,18 +227,39 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
 		return len(programmedNCs), errors.Wrap(err, "failed to get nc version list from nmagent")
 	}
 
+	// Get IMDS NC versions for delegated NIC scenarios
+	imdsNCVersions, err := service.GetIMDSNCs(ctx)
+	if err != nil {
+		// If any of the NMA API check calls, imds calls fails assume that nma build doesn't have the latest changes and create empty map
+		imdsNCVersions = make(map[string]string)
+	}
+
```
tamilmani1989, line 236: how do we make sure this not invoked for swiftv1 case?

kadevu: The idea is not to differentiate between SwiftV1 and SwiftV2 and get the programming status for the NCs being programmed and avoiding any conditional code

PR #3824, opened 2025-07-14
Feat: Adding a CNI Telemetry sidecar to CNS pod to replace the azure-vnet-telemetry binary

behzad-mir:
Adding a CNI Telemetry sidecar to CNS pod to replace the azure-vnet-telemetry binary.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/service/main.go
```diff
@@ -470,11 +471,36 @@ func sendRegisterNodeRequest(ctx context.Context, httpClient httpDoer, httpRestS
...
+	}
 
 	tb := telemetry.NewTelemetryBuffer(nil)
-	err := tb.CreateAITelemetryHandle(config, false, false, false)
+
+	var err error
+	if cnsconfig.TelemetrySettings.AppInsightsInstrumentationKey != "" {
+		err = tb.CreateAITelemetryHandle(aiConfig, ts.DisableTrace, ts.DisableMetric, ts.DisableEvent)
+	} else {
+		logger.Printf("No Application Insights key provided for CNI telemetry service")
```
tamilmani1989: we should check other way and remove else?
```
if cnsconfig.TelemetrySettings.AppInsightsInstrumentationKey == "" {
  logger.Printf("No Application Insights key provided for CNI telemetry service")
  return
}
err = tb.CreateAITelemetryHandle(aiConfig, ts.DisableTrace, ts.DisableMetric, ts.DisableEvent)
```

santhoshmprabhu:
Discussed in the team meeting today - we may want to consider doing telemetry for CNI directly in CNI rather than CNS, to account for scenarios where CNI can't talk to CNS.
Tagging @tamilmani1989 as decided in the call.

Makefile
```diff
@@ -81,6 +81,10 @@ ACN_PACKAGE_PATH = github.com/Azure/azure-container-networking
 CNI_AI_PATH=$(ACN_PACKAGE_PATH)/telemetry.aiMetadata
 CNS_AI_PATH=$(ACN_PACKAGE_PATH)/cns/logger.aiMetadata
 NPM_AI_PATH=$(ACN_PACKAGE_PATH)/npm.aiMetadata
+CNI_TELEMETRY_SIDECAR_DIR = $(REPO_ROOT)/cns/cni-telemetry-sidecar
+CNI_TELEMETRY_SIDECAR_BUILD_DIR = $(BUILD_DIR)/cni-telemetry-sidecar
+CNI_TELEMETRY_SIDECAR_AI_ID = $(CNI_AI_ID)  # Reuse CNI AI ID
+CNI_TELEMETRY_SIDECAR_VERSION = $(CNS_VERSION)  # Version follows CNS
```
QxBytes: can we move the version to where all the other version variables are?

Makefile
```diff
@@ -123,7 +128,7 @@ all-binaries-platforms: ## Make all platform binaries
 
 # OS specific binaries/images
 ifeq ($(GOOS),linux)
-all-binaries: acncli azure-cni-plugin azure-cns azure-npm azure-ipam azure-ip-masq-merger ipv6-hp-bpf
+all-binaries: acncli azure-cni-plugin azure-cns azure-npm azure-ipam azure-ip-masq-merger ipv6-hp-bpf cni-telemetry-sidecar
 all-images: npm-image cns-image cni-manager-image azure-ip-masq-merger-image ipv6-hp-bpf-image
```
QxBytes: it looks like you're making a new image-- if this is like a new sidecar image/container image, there are other dockerfiles/scripts you may need to edit (ex: for the signed pipeline). would the image target be added here?

Makefile
```diff
@@ -275,6 +284,7 @@ CNI_IMAGE					= azure-cni
 CNS_IMAGE					= azure-cns
 NPM_IMAGE					= azure-npm
 AZURE_IP_MASQ_MERGER_IMAGE	= azure-ip-masq-merger
+CNI_TELEMETRY_SIDECAR_IMAGE = azure-cni-telemetry-sidecar
```
QxBytes: not sure if you want to keep the variable naming consistent? it's CNI_TELEMETRY here but the image name is azure-cni-telemetry-sidecar-- is the "sidecar" part of the name necessary (here and elsewhere)? If it's a container image I feel like it's already implied that it'll be running in a sidecar or similar.

Makefile
QxBytes: Also if you're intending this to be like azure ip masq merger with its own release versions etc. you'll need to add the exclusion to the tag above: `ACN_VERSION							?= $(shell git describe --exclude "azure-ip-masq-merger*" --exclude "azure-ipam*" --exclude "dropgz*" --exclude "zapai*" --exclude "ipv6-hp-bpf*" --tags --always)`

Makefile
```diff
@@ -783,6 +837,14 @@ ifeq ($(GOOS),linux)
 	cd $(IPV6_HP_BPF_BUILD_DIR) && $(ARCHIVE_CMD) $(IPV6_HP_BPF_ARCHIVE_NAME) ipv6-hp-bpf$(EXE_EXT)
 endif
 
+# Create a CNI Telemetry Sidecar archive for the target platform.
+.PHONY: cni-telemetry-sidecar-archive
+cni-telemetry-sidecar-archive: cni-telemetry-sidecar-binary
+ifeq ($(GOOS),linux)
```
QxBytes, line 860: if this binary is on windows and linux would we omit this linux check?

cns/cni-telemetry-sidecar/configmanager.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	// Validate batch size settings for Azure Application Insights
+	if config.TelemetrySettings.TelemetryBatchSizeBytes <= 0 {
+		logger.Printf("Warning: Invalid telemetry batch size, using default")
+	}
+
+	// Validate batch interval for optimal Azure ingestion
+	if config.TelemetrySettings.TelemetryBatchIntervalInSecs <= 0 {
+		logger.Printf("Warning: Invalid telemetry batch interval, using default")
```
QxBytes: does ai under the hood use the default automatically if it's not valid, or should this also modify the value to the default (also applies to above conditional)?

cns/cni-telemetry-sidecar/main.go
```diff
@@ -0,0 +1,51 @@
...
+var (
+	version    = "unknown"
+	configPath = flag.String("config", "/etc/cns/cns-config.json", "Path to CNS configuration file")
+)
+
+func main() {
+	flag.Parse()
+
+	// Initialize logging for the CNI telemetry sidecar
+	logger.InitLogger("azure-cns-cni-telemetry-sidecar", 1, 1, "/var/log/azure-cns-telemetry")
```
QxBytes: should the naming be the same-- would it be cns or cni telemetry? Like below the logger prints that it's azure cni telemetry sidecar but the log is to the file azure-cns-telemetry

cns/cni-telemetry-sidecar/sidecar.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/Azure/azure-container-networking/aitelemetry"
+	"github.com/Azure/azure-container-networking/cns/configuration"
+	"github.com/Azure/azure-container-networking/cns/logger"
+	"github.com/Azure/azure-container-networking/telemetry"
+)
+
+const (
+	// CNI telemetry constants aligned with azure-vnet-telemetry
+	cniTelemetryAppName = "azure-vnet-telemetry"
+	cniTelemetryVersion = "1.0.0"
```
QxBytes: nit: do we want to start with 0.0.x release train or 1.0.x?

QxBytes (commented):
based on my understanding, this pr replaces azure-vnet-telemetry binary on the node with a sidecar that does the same thing. So CNI would send AI events to this sidecar via a socket if it is present (rather than the azure vnet telemetry long running process). I believe right now azure-vnet-telemetry is still on the vhd (?) so CNI will still attempt to start that azure vnet telemetry binary service. Can you check that there's no race or issue with both CNS and Azure CNI trying to start the telemetry service at the same time?

.pipelines/pipeline.yaml
```diff
@@ -153,6 +153,14 @@ stages:
...
                 name: npm
                 os: windows
+              cni_telemetry_sidecar_linux_amd64:
+                arch: amd64
+                name: cni-telemetry-sidecar
+                os: linux
+              cni_telemetry_sidecar_windows_amd64:
+                arch: amd64
+                name: cni-telemetry-sidecar
+                os: windows
```
rbtr, line 163: instead of publishing a separate image, can we just build this in to the azure-cni image and run it from there?

rbtr (changes requested):
instead of a new image, let's package this in azure-cni - it's already cached and even embedded in most CNS daemonsets today

behzad-mir:
> based on my understanding, this pr replaces azure-vnet-telemetry binary on the node with a sidecar that does the same thing. So CNI would send AI events to this sidecar via a socket if it is present (rather than the azure vnet telemetry long running process). I believe right now azure-vnet-telemetry is still on the vhd (?) so CNI will still attempt to start that azure vnet telemetry binary service. Can you check that there's no race or issue with both CNS and Azure CNI trying to start the telemetry service at the same time?

We are aming to run this on Stateless CNI  in specific K8s version. So, it will not run the azure-vnet-telemtry.

Makefile
```diff
@@ -84,6 +84,10 @@ ACN_PACKAGE_PATH = github.com/Azure/azure-container-networking
 CNI_AI_PATH=$(ACN_PACKAGE_PATH)/telemetry.aiMetadata
 CNS_AI_PATH=$(ACN_PACKAGE_PATH)/cns/logger.aiMetadata
 NPM_AI_PATH=$(ACN_PACKAGE_PATH)/npm.aiMetadata
+CNI_TELEMETRY_SIDECAR_DIR = $(REPO_ROOT)/cns/cni-telemetry-sidecar
```
tamilmani1989: can we have `azure-` prefix?

tamilmani1989: can it be `acn-node-telemetry`?

tamilmani1989 (changes requested):

Makefile
behzad-mir: Per @rbtr request I moved it to azure-cni image. I removed all the Makefile changes as a result.

Makefile
behzad-mir: @tamilmani1989 I removed all the Makefile changes and add the binary to the CNI DOCker instead per @rbtr suggestion.

cns/cni-telemetry-sidecar/sidecar.go
behzad-mir: Probably need to decide on versioning and put it out of this file. Will be done in next commit

cns/cni-telemetry-sidecar/configmanager.go
```diff
@@ -0,0 +1,145 @@
...
+type ConfigManager struct {
+	configPath string
+	logger     *zap.Logger
+}
+
+// NewConfigManager creates a new ConfigManager.
+// Config path resolution priority:
+//  1. Command line flag (if provided)
+//  2. CNS_CONFIGURATION_PATH environment variable
+//  3. Default path ({executable_directory}/cns_config.json)
```
rbtr: Instead of rolling this ourselves we should use [Viper](https://github.com/spf13/viper)

cns/cni-telemetry-sidecar/configmanager.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	var config configuration.CNSConfig
+	if err := json.Unmarshal(data, &config); err != nil { //nolint:musttag // CNSConfig has json tags in configuration package
+		return nil, fmt.Errorf("failed to parse config file: %w", err)
+	}
+
+	cm.applyDefaults(&config)
+
+	if cm.logger != nil {
```
rbtr: logger should not be allowed to be nil by this point

cns/cni-telemetry-sidecar/main.go
```diff
@@ -0,0 +1,93 @@
...
+	"syscall"
+
+	"github.com/Azure/azure-container-networking/telemetry"
+	"go.uber.org/zap"
+)
+
+var (
+	configPath = flag.String("config", "", "Path to CNS configuration file")
+	logLevel   = flag.String("log-level", "info", "Log level (debug, info, warn, error)")
+	version    = "1.0.0" // Set at build time via -ldflags
```
rbtr: should remain unassigned if it's supposed to be set with ldflags

cns/cni-telemetry-sidecar/main.go
```diff
@@ -0,0 +1,93 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		sig := <-sigChan
+		logger.Info("Received shutdown signal", zap.String("signal", sig.String()))
+		cancel()
+	}()
```
rbtr, line 56: typically we would do this in an init() instead of halfway through main - if something hangs before we get here we don't have exit signal handling

cns/cni-telemetry-sidecar/main.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if err := sidecar.Run(ctx); err != nil && !errors.Is(err, context.Canceled) {
+		logger.Error("Sidecar execution failed", zap.Error(err))
+		return 1
```
rbtr: this looks like the [Cobra](https://github.com/spf13/cobra) command pattern so I'm wondering why not use Cobra and remove all this other boilerplate

Makefile
QxBytes: no longer relevant as this is no longer an independent image (now part of cni)

cni/Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet -trimpath -ldflags
 RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet-telemetry -trimpath -ldflags "-s -w -X main.version="$VERSION" -X "$CNI_AI_PATH"="$CNI_AI_ID"" -gcflags="-dwarflocationlists=true" cni/telemetry/service/telemetrymain.go
 RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet-ipam -trimpath -ldflags "-s -w -X main.version="$VERSION"" -gcflags="-dwarflocationlists=true" cni/ipam/plugin/main.go
 RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet-stateless -trimpath -ldflags "-s -w -X main.version="$VERSION"" -gcflags="-dwarflocationlists=true" cni/network/stateless/main.go
+RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-cni-telemetry-sidecar -trimpath -ldflags "-X main.version=$VERSION -X "$CNI_AI_PATH"="$CNI_AI_ID"" -gcflags="-dwarflocationlists=true" ./cns/cni-telemetry-sidecar
```
QxBytes: I know we don't have it for the others but can we escape the double quotes? or have it be consistent with line 22 where we put `$VERSION` in quotes?

cns/cni-telemetry-sidecar/configmanager.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+const (
+	defaultConfigName          = "cns_config.json"
+	defaultTelemetrySocketPath = "/var/run/azure-vnet-telemetry.sock"
+)
+
+// appInsightsEnvVars are the environment variables checked for AppInsights keys.
+var appInsightsEnvVars = []string{
+	"APPINSIGHTS_INSTRUMENTATIONKEY",
```
QxBytes: wondering why we have many env variables which all seem to be used in the same way? Or maybe not even at all (where is APPLICATIONINSIGHTS_CONNECTION_STRING used?). The instrumentation key is different from the connection string.

behzad-mir: Removed the multiple env vars array and replaced it with a single constant [appInsightsEnvVar = "APPINSIGHTS_INSTRUMENTATIONKEY"]

cni/Dockerfile
behzad-mir: Done

behzad-mir:
> Wanted to confirm: you won't come into conflict with an already-running azure-vnet-telemetry binary because stateless cni won't start it up, correct?
> 
> Also if this sidecar hangs/fails to come up, cns still runs right?
> 
> Can you confirm what scenarios this affects?
> 
> * Stateless CNI linux: yes/no
> * Stateless CNI windows: yes/no
>   and add those affected scenarios to the PR description. I believe it's both but might be helpful for future reference to include that in the description.
> 
> Agree that for configuration code we should try to use viper
> 
> If this affects stateless cni, please add it to its pipeline stage

1-No conflict with azure-vnet-telemetry binary?
Stateless CNI uses ConnectTelemetry() which only connects to an existing socket - it does NOT start the telemetry process
Stateful CNI uses StartAndConnectTelemetry() which spawns the azure-vnet-telemetry binary if not running
The sidecar creates the socket at /var/run/azure-vnet-telemetry.sock before any CNI calls happen, so stateless CNI will connect to it. Since stateless CNI never calls StartTelemetryService(), no azure-vnet-telemetry binary is spawned.

2-2. If sidecar hangs/fails, does CNS still run?
Yes, CNS continues to run.
The sidecar is a separate container in the pod. If it fails:
CNS container is unaffected (no container dependency)
Stateless CNI's ConnectTelemetry() will fail to connect but gracefully handles this - it just won't send telemetry
No pod crashes - telemetry is best-effort

3- Affected scenarios are all stateless CNI linux and Windows.

cns/cni-telemetry-sidecar/main.go
behzad-mir: Fixed. Now signal handling is set up at the very start of [run()], before logger initialization, config loading, or sidecar creation.

cns/cni-telemetry-sidecar/main.go
behzad-mir: Replace this with Cobra pattern.

cns/cni-telemetry-sidecar/configmanager.go
behzad-mir: since CNS already has a configuration.ReadConfig() function that handles all the config loading logic, it makes more sense to reuse that rather than introducing a separate Viper-based config for the sidecar. If CNS migrates to Viper in the future, the sidecar will automatically benefit from it. Let me know if this works.

cns/cni-telemetry-sidecar/configmanager.go
behzad-mir: Now requires logger in constructor, removed SetLogger() method and nil checks

cns/cni-telemetry-sidecar/main.go
behzad-mir: done

cns/cni-telemetry-sidecar/sidecar.go
```diff
@@ -0,0 +1,204 @@
...
+	if cnsConfig != nil && cnsConfig.TelemetrySettings.AppInsightsInstrumentationKey != "" {
+		return cnsConfig.TelemetrySettings.AppInsightsInstrumentationKey
+	}
+	return os.Getenv(appInsightsEnvVar)
+}
+
+func (s *TelemetrySidecar) startTelemetryService(ctx context.Context, config telemetry.TelemetryConfig, cnsConfig *configuration.CNSConfig) error {
+	aiKey := s.getAppInsightsKey(cnsConfig)
+	if aiKey != "" {
+		os.Setenv("APPINSIGHTS_INSTRUMENTATIONKEY", aiKey)
```
QxBytes: curious why we set this environment variable here? The places we use it only seem to be `getAppInsightsKey` and `hasAppInsightsKey` which are both called before this.

QxBytes (dismissed):

cns/cni-telemetry-sidecar/configmanager.go
rbtr: CNS config loading is so jank. You don't have to keep using something bad because it already exists

cns/configuration/configuration.go
```diff
@@ -89,6 +89,10 @@ type TelemetrySettings struct {
 	ConfigSnapshotIntervalInMins int
 	// AppInsightsInstrumentationKey allows the user to override the default appinsights ikey
 	AppInsightsInstrumentationKey string
+	// Flag to enable CNI telemetry collection via sidecar
+	EnableCNITelemetry bool
+	// Path to the CNI telemetry socket file that azure-vnet CNI connects to
+	CNITelemetrySocketPath string
```
rbtr: My assumption is that you added these here because the sidecar will have access to the CNS configmap, but these should not have been added to the _CNS_ config struct. CNS doesn't understand them or care about them.
Instead, we can exploit a feature of Go's JSON marshalling - you can marshal a JSON object with more data than the struct you're targeting. Then, adding these fields to the configmap doesn't change anything when it's parsed into the CNS config, they are discarded.
However, when we parse that same data into a new CNITelemetryConfig struct with only these fields, all of the _other_ data would be discarded.

rbtr (approved):

cns/configuration/configuration.go
behzad-mir: This has been corrected in the new commit.

rbtr (approved):

cni/Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet -trimpath -ldflags
 RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet-telemetry -trimpath -ldflags "-s -w -X main.version="$VERSION" -X "$CNI_AI_PATH"="$CNI_AI_ID"" -gcflags="-dwarflocationlists=true" cni/telemetry/service/telemetrymain.go
 RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet-ipam -trimpath -ldflags "-s -w -X main.version="$VERSION"" -gcflags="-dwarflocationlists=true" cni/ipam/plugin/main.go
 RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-vnet-stateless -trimpath -ldflags "-s -w -X main.version="$VERSION"" -gcflags="-dwarflocationlists=true" cni/network/stateless/main.go
+RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-cni-telemetry-sidecar -trimpath -ldflags "-X main.version="$VERSION" -X "$CNI_AI_PATH"="$CNI_AI_ID"" -gcflags="-dwarflocationlists=true" ./cns/cni-telemetry-sidecar
```
tamilmani1989, line 25: should we keep the name as `azure-cni-telemetry`. we may resuse this in non-aks scenario as well

behzad-mir: Decided to keep the name to avoid similarity with azure-vnet-telemetry which is being used by stateful CNI

QxBytes (approved):

PR #3825, opened 2025-07-14
CNS API contracts for NUMA-Aware Pods

pjohnst5:
**Reason for Change**:
This introduces two new APIs in CNS:
1. PUT
`/ibdevices/pod/{podname-podnamespace}`
Request body:
``` json
{
  "podID": "string", //podname-podnamespace
  "deviceIds": ["string"] // ["60:45:bd:a4:b5:7a", "7c:1e:52:07:11:36"] for example
}
```
Response:
```
{
    "response": {
        "cnsCode": -1, // Does this add any value for NRM?
        "message": "Device aa:bb:cc:dd:ee:ff is already assigned to pod other-pod-namespace"
    }
}
```

2. GET
`/ibdevices/{mac-of-ib-device}`
Request: <no-body>
Response:
``` json
{
  "deviceID": "string",
  "podID": "string",
  "status": "string",
  "errorCode": "int",  
  "msg": "string"
}
```

**Notes**:
This is to enable "NUMA-Aware" Pods


cns/api.go
```diff
@@ -382,3 +386,33 @@ type GetVMUniqueIDResponse struct {
...
+type AssignIBDevicesToPodResponse struct {
+	ErrorCode infiniband.ErrorCode `json:"errorCode"` // Error code if applicable
+	Message   string               `json:"message"`   // Additional message or error description
+}
+
+// GetIBDeviceStatusRequest represents the request to get the InfiniBand device programming status
+// GET /ibdevices/{mac-address-of-device} - no request body needed for GET
+type GetIBDeviceStatusRequest struct {
+	MACAddress string `json:"macAddress"` // MAC address of the device (from URL path)
+}
```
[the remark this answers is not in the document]
pjohnst5: Will do it but without the comment

cns/swagger.yaml
```diff
@@ -120,6 +120,78 @@ paths:
...
+        Retrieves the current status of the specified IB device.
+      parameters:
+        - name: macaddress
+          in: path
+          required: true
+          description: The MAC address of the IB device (with colons ":")
+          schema:
+            type: string
+      requestBody:
+        required: false
```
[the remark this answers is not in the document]
pjohnst5: Will do it but without the comment

pjohnst5:
Waiting for #3876 to merge first before merging this

cns/types/infiniband/errorcodes.go
```diff
@@ -0,0 +1,13 @@
+package infiniband
+
+type ErrorCode int
+
+const (
+	Success                  ErrorCode = 0
+	PodNotFound              ErrorCode = 1
+	DeviceUnavailable        ErrorCode = 2
+	DeviceNotFound           ErrorCode = 3
```
kmurudi: when would we use DeviceNotFound?

cns/types/infiniband/errorcodes.go
```diff
@@ -0,0 +1,13 @@
...
+	AnnotationNotFound       ErrorCode = 4
+	PodAlreadyAllocated      ErrorCode = 5
```
kmurudi: how would we determine this? should this be idempotent, or if we want to return that pod-request is already in-progress let's return that

cns/types/infiniband/status.go
```diff
@@ -0,0 +1,11 @@
+package infiniband
+
+type Status int
+
+const (
+	ProgrammingPending  Status = 0
+	ProgrammingFailed   Status = 1
+	ProgrammingComplete Status = 2
+	ReleasePending      Status = 3
+	Available           Status = 4
```
kmurudi: thinking if this should be 0? we should have a default state & once released we should let IB status reflect that initial state, since we aren't actually deleting an IB here just marking it available or programmed

cns/api.go
```diff
@@ -382,3 +386,28 @@ type GetVMUniqueIDResponse struct {
 	Response   Response `json:"response"`
 	VMUniqueID string   `json:"vmuniqueid"`
 }
+
+// IBDevice API Contracts
+
+// AssignIBDevicesToPodRequest represents the request to assign InfiniBand devices to a pod
+// POST /ibdevices/pod
+type AssignIBDevicesToPodRequest struct {
+	PodID        string   `json:"podID"`        // podname-podnamespace format
```
kmurudi: just check if the format is usable everywhere in your methods later, i.e. easy to decipher from here & then use it for checking with api-server if pod exists etc. we could also change this later if you discover it doesn't work & is efficient in other format like body/struct or separate fields

cns/types/infiniband/errorcodes.go
pjohnst5: Basically, if DNC goes to try and program the device, but, in pubsub, it's not there (like a mismatch between the physical devices on the VM and what's in pubsub), DNC-RC could write "DeviceNotFound" in the CRD status

cns/types/infiniband/errorcodes.go
pjohnst5: Basically, if NRM tries to assign the same pod twice, but with _different_ devicd IDs the 2nd time, we want to return this

This basically means, "hey for this pod, we've already allocated devices, and the current devices you are trying to assign are different than we've already programmed"

cns/types/infiniband/status.go
pjohnst5: Yeah that's true, I'll move it to 0

pjohnst5: Done

cns/api.go
pjohnst5: Yeah good point, in controller-runtime, I see they do "podnamespace/podname"
image

But seeing as this will be part of the API path on the GET, and also that this API is per pod, I'd rather lead with the podname (and make it "-")

cns/types/infiniband/errorcodes.go
kmurudi: what if they send request for the same pod-id & with same devices - do we want to ignore an idempotent request or what will our response be

pjohnst5: I would say our response would be `200` so yeah idempotent

pjohnst5: Would you like me to name this "ConflictingPodInfo" or something like that? (or "ConflictingDeviceInfo"?)

rbtr: try and omit error codes entirely (because you already have the HTTP ones). `PodAlreadyAllocated` could be covered by returning HTTP/409 with an error message, `DeviceNotFound` is simply a 404, etc.
aside, PodAlreadyAllocated is the wrong naming because the _Pod_ is not allocated, the IBDev is.

cns/api.go
rbtr: If the accepted convention is namespaced name, you should be adapting your implementation to that convention. Please don't do your own different quirky thing just because it's convenient, that's how we (already) end up with a dozen different Pod "IDs" in different parts CNS state that are irreconcilable.
It's generally more parseable and makes more sense when lexicographically sorted if you order by least -> most specificity. Pod is more specific than namespace, so it should go after.

cns/api.go
```diff
@@ -382,3 +386,28 @@ type GetVMUniqueIDResponse struct {
...
+// AssignIBDevicesToPodResponse represents the response for assigning InfiniBand devices to a pod
+type AssignIBDevicesToPodResponse struct {
+	ErrorCode infiniband.ErrorCode `json:"errorCode"` // Error code if applicable
+	Message   string               `json:"message"`   // Additional message or error description
+}
+
+// GET /ibdevices/{mac-address-of-device}
+// GetIBDeviceStatusResponse represents the response containing InfiniBand device programming status
+type GetIBDeviceStatusResponse struct {
+	MACAddress string               `json:"macAddress"` // MAC address of the device
```
rbtr: [there's a HardwareAddress type](https://pkg.go.dev/net#HardwareAddr)

cns/api.go
```diff
@@ -382,3 +386,28 @@ type GetVMUniqueIDResponse struct {
...
+	PodID      string               `json:"podID"`      // Pod that the device is assigned to
+	Status     infiniband.Status    `json:"status"`     // Device status (e.g., "pendingProgramming", "error", "programmed", "pendingDeletion", "available")
+	ErrorCode  infiniband.ErrorCode `json:"errorCode"`  // Error code if applicable
```
rbtr: what's the value of an errorCode instead of an error string? only adds an extra step to debugging imo

cns/types/infiniband/errorcodes.go
pjohnst5: Yeah I see your point about trying to get rid of error codes entirely
However, how would we distinguish between `PodNotFound` vs `DeviceNotFound`?
>the Pod is not allocated, the IBDev is

The case I'm trying to cover here is more like "Pod A has already been assigned devices B and C, but you're now trying to assign it devices D and E"

cns/api.go
pjohnst5: Okay haha

cns/types/infiniband/errorcodes.go
kmurudi: not Conflicting, maybe something that says pod is being allocated or in-progress, yeah whether devices given are same or different - if the given pod-id is in-progress we should return that & not take a new request for it

cns/types/infiniband/errorcodes.go
kmurudi: can we comment above on which error codes would be returned for POST pod api & which ones to be used for IBDevice status, i.e. first list the pod api errorcodes followed by ib device status, since they are being returned/used for different purpose
        `

cns/swagger.yaml
```diff
@@ -120,6 +120,69 @@ paths:
               schema:
                 $ref: "#/components/schemas/UnpublishNetworkContainerResponse"
 
+  /ibdevices/pod:
```
rbtr: I stared at this for a long time trying to figure out what bothered me about it and why it didn't seem like good REST - I thought it was because we must POST here but GET at /{mac}?... and I want to be able to GET the asset I have just created/updated.
But now I think it's simply that `pod` is not an `ibdevice`. The collection doesn't make sense - a bucket of ibdevices where I can get by mac but post by pod?
https://github.com/microsoft/api-guidelines/blob/vNext/graph/Guidelines-deprecated.md#93-collection-url-patterns

instead maybe: 
`/ibdevices/{mac}` is the collection and items, can GET
`/ibdevices/{mac}/pod` is the mapping of ibdev:pod, can GET or POST to set associated Pod

or inverted (but a collection of pods seems odd since we don't control those and it's not all of them)
`/pods/{pod}` collection of pods, can GET
`/pods/{pod}/ibdevices` mapped IB devices, can GET or POST

or maybe both, as long as they are bidirectionally linked and traversable

cns/types/infiniband/errorcodes.go
rbtr: > Yeah I see your point about trying to get rid of error codes entirely However, how would we distinguish between `PodNotFound` vs `DeviceNotFound`?
> 
> > the Pod is not allocated, the IBDev is
> 
> The case I'm trying to cover here is more like "Pod A has already been assigned devices B and C, but you're now trying to assign it devices D and E"

I think this smells because the api structure isn't mapping to the objects conceptually right, left a comment on the swagger with more thoughts.

cns/api.go
pjohnst5: Hey @rbtr , what would you say is best at this point in CNS, to start moving towards a single consolidated `PodInfo` way of identifying a pod?
I see in https://github.com/Azure/azure-container-networking/blob/2997eb49b0fd1f29ea6cd81cf232f4c9b95efc8d/cns/NetworkContainerContract.go#L217

There's a `PodInfo` struct, should I use that here as well?

So
```go
PodID PodInfo `json:podID"`
```

rbtr: you can URL encode the `/` if you want your asset to be "namespace/name" and not "namespace"/"name"

cns/swagger.yaml
pjohnst5: Interesting, interesting, okay yeah, here's the main thing we are trying to allow
1. Let's say, IB device A and B are assigned to Pod A
2. Let's say everything is programmed and Pod A is running
3. Pod A gets deleted
4. DNC successfully unprograms IB Device A, but not B
5. IB Device B is stuck completely due to some bug

In that case, we want to free up IB Device A, so that a new pod can use it while we try to fix IB Device B (or even if unprogramming device B is taking a long time)

But now I'm thinking about what you said

cns/api.go
pjohnst5: Yeah, so the partner team wanted us to give back error codes, so that when issues arose (let's say `PodNotFound` vs `DeviceNotFound`, they could try some extra steps on their end automatically to fix it, instead of involving DRIs

cns/api.go
pjohnst5: It has extra fields than what I need so just wondering what you think

pjohnst5: > you can URL encode the / if you want your asset to be "namespace/name" and not "namespace"/"name"

Oh yeah, this is actually not a problem anymore, because I moved the namespace/name from the url to the request body instead

rbtr: To your question on using PodInfo, do you currently have a problem if a Pod is deleted and a new one immediately created with same name?

cns/swagger.yaml
pjohnst5: Yeah it's kind of weird, because in this case, a pod can have multiple IBs, but we want to treat each IB device as independent in a way, in case something goes wrong (or is slow), to use the free IB devices as soon possible

cns/api.go
pjohnst5: That's a good question, here's how we've designed it:
1. Pod A is created
2. MTPNC is also created for it with a finalizer
3. Devices are programed, pod runs etc.
4. Pod is deleted
5. MTPNC is only deleted, after all devices for that pod were unprogrammed (this will be enforced in our other repo)

So now I'm wondering, is the pod also kept alive, while the MTPNC for it is?
If yes:
- We're good, since the new pod can't be created until the old one is fully deleted
If no:
- Then yeah we'd have a problem haha

cns/swagger.yaml
pjohnst5: So let's imagine for a minute we did the first way you said:
`/ibdevices/{mac}`
`/ibdevices/{mac}/pod`

In this case, how would the partner team, assign multiple IB devices to a single pod?

And then, let's imagine the other way:
`/pods/{pod}`
`/pods/{pod}/ibdevices`
This way I see how we could assign multiple IB devices to a single pod
But, like you said, we don't own all the pods nor control them, so it's just a bit odd
And then back to that situation where one IB device is slower to unprogram than another, we want to be able to get that info back to the partner team quickly, so in this case, the partner team would GET on `/pod/{pod}/ibdevice`, but have to remember the pod I guess, which would mean they would have to keep track of which pod they assigned it to or something like that 🤔

pjohnst5: Yeah, basically, when discussing this, they said:
"Oh yeah, we'll want to know the status of individual devices, to be able to use them for other pods easily, so let's make it GET on mac"
Just trying to relay that info here

cns/types/infiniband/errorcodes.go
pjohnst5: Yeah, here's a rough draft of which code would be used for what:
```go
const (
	Success                  ErrorCode = 0 // POST and GET
	PodNotFound              ErrorCode = 1 // POST
	DeviceUnavailable        ErrorCode = 2 // POST
	DeviceNotFound           ErrorCode = 3 // GET
	AnnotationNotFound       ErrorCode = 4 // POST
	PodAlreadyAllocated      ErrorCode = 5 // POST
	InternalProgrammingError ErrorCode = 6 // GET
)
```

Do you want me to simply move shuffle the lines of the error codes around? Or make separate error code enumerations for POST errors vs GET?

cns/swagger.yaml
rbtr: i think this is all manageable with query params on an ibdevices collection
GET `/ibdevices?state=assigned` to find all the assigned ibdevices
GET `/ibdevices?state=unassigned` unassigned
GET `/ibdevices?by-pod=podnamespacedname`
POST `/ibdevices:assign?ibdevs=mac1,mac2&pod=podnamespacedname`

cns/api.go
rbtr: is it just a finalizer or an ownerref?

cns/swagger.yaml
pjohnst5: Okay, one thing I should mention, is what the client of these APIs is doing

The client is going to scan for the MACs on the L1VH VM on startup, and choose devices that belong to the same NUMA nodes (within the L1VH VM)

CNS will not know which devices are part of the same NUMA nodes, the partner team's client will know that

So we weren't planning on having CNS scan for all devices on startup as well, since the client would be doing that

That being said, I do see a potential problem, where the client could give CNS garbage (like a non-existent device id), and CNS would just pass that along to DNC

DNC would be the one that discovers the device does not exist, and have to trickle that knowledge back to CNS

So I think with what you just shared, you're assuming CNS will know all the devices on the L1VH at any given point in time, is that correct?

cns/api.go
pjohnst5: Finalizer
`finalizers.acn.azure.com/dnc-operations`

pjohnst5: So yes, the pod may be deleted, but the mtpnc for it may still be around
And then, if a new pod with the exact same name in the same namespace were to be created right after, then we'd have a problem of the old mtpnc still existing

What was your point then regarding the pod info?

cns/swagger.yaml
pjohnst5: Hey @rbtr , what's the reason to have the `mac1`, `mac2` ids as query params instead of in the POST body?
```json
{
  "podID": "string" // podname-podnamespace
  "macAddresses": ["string"] // ["60:45:bd:a4:b5:7a", "7c:1e:52:07:11:36"] for example
}
```

rbtr: it's just not enough data to require a body. mac1,mac2 select the item in the collection that you are operating on, which is what query params are for.
_maybe_ there's an argument that the pod should be in a body instead of assigned out of the param but it's acceptable either way and putting it all one place is better than splitting imo

rbtr: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#performing-an-action

cns/api.go
pjohnst5: I can leave it out this PR, and add it in another one after we discuss it more

pjohnst5:
Alright @rbtr , I've made some changes, I just have 2 remaining questions:
- Why do you want the `assign` action specifically in the POST API?
  - I think it may be because POST alone (good rapper name?) with no action, implies you're creating a new resource (which isn't the case here)
- Why did you ask if a pod being created with the same name right after one was deleted would be a problem?
  - I've discovered that is a problem, but now want to know what you had in mind because we were talking about the `podID`
  - Were you thinking we need to pass more identification info such as `InfraContainerID` ?


https://github.com/Azure/azure-container-networking/blob/138e4f57a978e45be442ba7adf0ead8857c00239/cns/NetworkContainerContract.go#L219

cns/types/infiniband/errorcodes.go
pjohnst5: I am omiting these for now, and if so deemed appropriate will add them in a future PR

pjohnst5:
Hi @rbtr , lmk what you think now with the latest changes, thanks

cns/api.go
rbtr: that namespace/name is not a unique identifier and you may have toctou race problems where a pod could pick up an outdated MTPNC

rbtr:
> Alright @rbtr , I've made some changes, I just have 2 remaining questions:
> 
> * Why do you want the `assign` action specifically in the POST API?
>   
>   * I think it may be because POST alone (good rapper name?) with no action, implies you're creating a new resource (which isn't the case here)

Exactly, if you don't have an action in the URI, the implicit contract is that POSTing to `/ibdevices/<id>` is a create or update of _the whole object_.
Thinking about it from the other direction, if POSTing `/ibdevices/<id>` assigns a mac to a Pod...how would you create the ibdevice entity in the first place? 


> * Why did you ask if a pod being created with the same name right after one was deleted would be a problem?
>   
>   * I've discovered that is a problem, but now want to know what you had in mind because we were talking about the `podID`
>   * Were you thinking we need to pass more identification info such as `InfraContainerID` ?
> 
> https://github.com/Azure/azure-container-networking/blob/138e4f57a978e45be442ba7adf0ead8857c00239/cns/NetworkContainerContract.go#L219

There's maybe a race here, since namespaced name is not unique. A pod could get a stale MTPNC?

pjohnst5:
> Thinking about it from the other direction, if POSTing /ibdevices/<id> assigns a mac to a Pod...how would you create the ibdevice entity in the first place?

Yeah that's a good point, the ibdevices are provisioned at the Azure host level
So while CNS can't really _create_ one, are you saying you'd like me to have the POSTing of `/ibdevices/<id>` "register" a device with CNS?

> There's maybe a race here, since namespaced name is not unique. A pod could get a stale MTPNC?

Yes you're right, we could have a race condition met here

Lemme see what we do today (cuz this already happens)

pjohnst5:
Alright @rbtr , I ran some tests, and something that changes even when pods get the same names, is `UID`:
image

We could make this a required parameter for the POST API
- CNS could then add it to the MTPNC on creation
  - CNS will be creating the MTPNC in this scenario btw
- CNS could then check the MTPNC on each POST, and make sure the `UID` matches
  - If it matches, CNS knows the MTPNC is current
  - If it doesn't match, we know the MTPNC is stale, and CNS could return an error

Wdyt?

rbtr:
who creates the MTPNC? why don't we set an ownerref to the Pod so it gets GCed?

pjohnst5:
> who creates the MTPNC?

Today, DNC-RC creates the MTPNC, however, in this new NUMA scenario, CNS will create the MTPNC instead

> why don't we set an ownerref to the Pod so it gets GCed

For this new NUMA scenario, we could do that

> Why haven't we done that in the past with DNC-RC making the MTPNC?

I don't know

Basically @rbtr , is there anything else you want different about this PR, besides adding an ownerref to the pod so it is GCed? (that anyways will come in a future PR when I actually implement the APIs, this PR is to establish the contract between the partner team and us)

pjohnst5:
Hey wait we may actually set an ownerref already (I probably just missed it), here is one example MTPNC that DNC-RC already creates today:
```yaml

apiVersion: multitenancy.acn.azure.com/v1alpha1
kind: MultitenantPodNetworkConfig
metadata:
  creationTimestamp: "2025-05-29T17:51:35Z"
  finalizers:
  - finalizers.acn.azure.com/dnc-operations
  generation: 1
  name: ibpod1-deployment-5dbf8f965-sdgzr
  namespace: e2e-ns-d0s9t9ibfg06a1lnpsog
  ownerReferences:
  - apiVersion: v1
    blockOwnerDeletion: true
    controller: true
    kind: Pod
    name: ibpod1-deployment-5dbf8f965-sdgzr
    uid: 18a8b6b2-b247-4ae6-a351-5d56a203eab4
  resourceVersion: "14083"
  uid: 817a229e-24cd-4efc-b37c-75d4ff48cb5f
```

Looking more in the code to see where we do that

pjohnst5:
Yeah I found where it is added, I sent it to you privately

pjohnst5:
So, we know there's an owner ref on the MTPNC, of the Pod

Are you saying, we should also add, an ownerref on the pod, of the MTPNC?

cns/swagger.yaml
```diff
@@ -120,6 +120,87 @@ paths:
               schema:
                 $ref: "#/components/schemas/UnpublishNetworkContainerResponse"
 
+  /ibdevices:assign:
+    post:
```
pjohnst5, line 124: This would look like this:
`POST /ibdevices:assign?ibmacs=60:45:bd:a4:b5:7a,7c:1e:52:07:11:36&podnamespace=mynamespace&podname=mypod`

cns/swagger.yaml
```diff
@@ -120,6 +120,87 @@ paths:
...
+        '400':
+          description: >-
+            One of the IB devices specified is not available.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AssignIBDevicesToPodResponse"
+  
+  /ibdevices:
+    get:
```
pjohnst5, line 156: This would look like this:
`GET /ibdevices?ibmac=60:45:bd:a4:b5:7a`

cns/swagger.yaml
kmurudi: there could be up to 7/8 macs too for a single pod, don't you think this is a bit cumbersome to add in request URL? I'd prefer the macs to be as part of a request body

pjohnst5:
Hi @rbtr , Kshitija and I were talking and on the POST API, if the pod has 7 or 8 macs, the URL will get really long.. 😅
`POST /ibdevices:assign?ibmacs=e6:ed:09:b1:97:7e,ea:2a:3f:19:fd:49,e9:b1:8b:f9:d1:e0,00:7b:34:84:c5:0f,5f:26:ba:f7:2b:41,07:81:05:db:2a:36,a8:e9:a1:50:3b:2a&podnamespace=mynamespace&podname=mypod`

Can I move them into arequest body instead? 😅
`POST /ibdevices:assign`
```json
{
"MACs": ["e6:ed:09:b1:97:7e",
"ea:2a:3f:19:fd:49",
"e9:b1:8b:f9:d1:e0",
"00:7b:34:84:c5:0f",
"5f:26:ba:f7:2b:41",
"07:81:05:db:2a:36",
"a8:e9:a1:50:3b:2a"],
"podnamespace": "mynamespace"
"podname": "mypod"
}
```

cns/swagger.yaml
pjohnst5: Agreed, I am moving them to a request body instead

cns/api.go
```diff
@@ -382,3 +385,26 @@ type GetVMUniqueIDResponse struct {
 	Response   Response `json:"response"`
 	VMUniqueID string   `json:"vmuniqueid"`
 }
+
+// AssignIBDevicesToPodRequest represents the request for assigning InfiniBand devices to a pod
+type AssignIBDevicesToPodRequest struct {
+	IBMACs       []net.HardwareAddr `json:"ibmacs"`       // List of IB device MAC addresses to assign
```
kmurudi: nit: IBMACAddresses, same in swagger below

cns/api.go
```diff
@@ -382,3 +385,26 @@ type GetVMUniqueIDResponse struct {
...
+	PodNamespace string             `json:"podNamespace"` // Namespace of the target pod
```
kmurudi: can you check if we can use this struct from NetworkContainerContract.go - https://github.com/Azure/azure-container-networking/blob/master/cns/NetworkContainerContract.go#L239 ? this might be handy in generating OrchestratorContext() from existing methods, check the flow from here how CNS will use these fields from this API request body to create an MTPNC & set owner-ref, populate MTPNC Spec fields etc (Whatever DNC-RC does today basically)

pjohnst5: Yes checked, and sent you a message

cns/api.go
pjohnst5: Done

kmurudi (approved):

cns/api.go
pjohnst5: As discussed, there is no toctou race on the pod name/namespace since the pod will be the owner of the mtpnc

PR #3860, opened 2025-07-28
ci: Unblocking ADO build pipeline

apontejaj:
**Reason for Change**:

The ADO build pipeline is currently blocked with a reference to a "acn-devops" repo which is not refenced in the main pipeline file (in the ADO repository).

This PR reverts this reference. It also implements a fix for a building all the images in the OneBranch ADO pipeline.

**Issue Fixed**:

No GitHub issue associated to this PR.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/build/generate-manifest.steps.yaml
+++ b/.pipelines/build/generate-manifest.steps.yaml
@@ -12,7 +12,7 @@ steps:
         .args = [ (.platform | split("/")[0]), (.platform     | split("/")[1]) ] | 
         .args = [ ("--os "   + .args[0]     ), ("--arch "     + .args[1]     ) ] | 
         if .osVersion then .args += ["--os-version " + .osVersion] else . end    |
-        { image: .imageReference, annotate: .args }' | \
+        { image: .imageReference, annotate: (.args | join(" ")) }' | \
       jq -rcs)
     echo >&2 "##vso[task.setvariable variable=MANIFEST_JSON;isOutput=true]$MANIFEST_DATA"
     echo "$MANIFEST_DATA" | jq -r .
--- a/.pipelines/build/image.steps.yaml
+++ b/.pipelines/build/image.steps.yaml
@@ -33,6 +33,7 @@ parameters:
   default:
     - "--target $(os) "
     - "--platform $(os)/$(arch) "
+    - "--provenance false "
 
 - name: common_build_args
   type: object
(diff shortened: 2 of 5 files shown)


.pipelines/build/generate-manifest.steps.yaml
```diff
@@ -12,7 +12,7 @@ steps:
         .args = [ (.platform | split("/")[0]), (.platform     | split("/")[1]) ] | 
         .args = [ ("--os "   + .args[0]     ), ("--arch "     + .args[1]     ) ] | 
         if .osVersion then .args += ["--os-version " + .osVersion] else . end    |
-        { image: .imageReference, annotate: .args }' | \
+        { image: .imageReference, annotate: (.args | join(" ")) }' | \
```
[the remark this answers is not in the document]
apontejaj: This change is only relevant to internal ADO pipeline.

.pipelines/build/image.steps.yaml
```diff
@@ -33,6 +33,7 @@ parameters:
   default:
     - "--target $(os) "
     - "--platform $(os)/$(arch) "
+    - "--provenance false "
```
rohitams, line 36: is this the workaround for the image manifest list is not a image issue (when merging image into the list)?

.pipelines/build/images.jobs.yaml
```diff
@@ -127,12 +127,10 @@ jobs:
         OS: linux
       ${{ elseif eq(job_data.job, 'windows_amd64') }}:
```
rohitams, line 128: so switching to az linux 3 without emulator solves the arch issue?

rohitams (approved):

.pipelines/build/image.steps.yaml
apontejaj: Correct. That's the workaround provided by OneBranch.
The also suggested an alternative way that doesn't really suit us because we create individual images for each one of the os/archs.

.pipelines/build/images.jobs.yaml
apontejaj: It doesn't quite solve the issue, but we noticed that for ARM64 images, using ARM64 host is more stable. This was implemented for both Retina OSS and Enterprise, so I'm taking the chance now to bring it here.

PR #3871, opened 2025-07-29
feat: iptables block using LSM BPF

santhoshmprabhu:
**Reason for Change**:

This PR introduces a binary that will block iptable rule installation in the host network namespace. We are doing this as a sub-feature for eBPF host routing in Cilium. This is relevant because any iptables rules in the host network namespace are skipped when Cilium eBPF host routing is enabled. Further, we allow certain processes, such as cilium-agent and ip-masq-agent to install rules, even if those rules may get bypassed. We account for this difference in behavior in our design.

The block covers both iptables netlink and iptables legacy. iptables netlink block targets the netlink_send LSM BPF hook, which gets invoked when a netlink message is sent (such as when installing iptables rules). If the message is an iptables rule installation, and the network namespace is host network namespace, and the parent of the calling process is not one of the allow-listed processes, we block that installation. The switch to parent is needed because cilium-agent, ip-masq-agent etc invoke the iptables binary as a child to do rule installation. In the legacy case, we block the setsockopt call with the  IPT_SO_SET_REPLACE option, which is used for iptables rule installation.

Whenever we block a rule installation, we increment a counter in a map that is pinned at `/sys/fs/bpf/block-iptables/event_counter`. We can read the count from a CNS sidecar container and generate kube events, so that the customer has visibility into failing iptables rule installations.

The binary is meant for packaging into a systemd service. The service will be always running, but will only block rule installation depending on the contents of a config file. If the config file is missing, or is present and has contents, the service will not block iptables rules. If the config file is present but empty, the service will block. The idea is that any process/component that needs to stop the iptables block can append a unique value to the file, and remove it once it is done with the need for the service to stop.

The Makefile changes build the binary, we will separately make the changes needed for inclusion of the binary in the github release artifacts.

**Validation steps completed**
- Packaged this binary manually into a systemd service and applied on AKS nodes, validated that iptables rule installations are blocked when the config file is present and empty, and not blocked otherwise. On real AKS clusters, AgentBaker will be responsible to set up this systemd service.
- When the systemd service is running, validated that Cilium is able to install iptables rules.
- Verified that block counts increase when rule installation attempts are blocked.
- Verified that iptables rules can be installed from pods that are not running in hostNetwork mode.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


Makefile
```diff
@@ -201,6 +209,21 @@ else ifeq ($(GOARCH),arm64)
 	for dir in /usr/include/aarch64-linux-gnu/*; do sudo ln -sfn "$$dir" /usr/include/$$(basename "$$dir"); done
 endif
 
+# Build the block-iptables binary.
+block-iptables-binary:
+	cd $(BLOCK_IPTABLES_DIR) && CGO_ENABLED=0 go generate ./...
+	cd $(BLOCK_IPTABLES_DIR)/cmd/block-iptables && CGO_ENABLED=0 go build -v -o $(BLOCK_IPTABLES_BUILD_DIR)/block-iptables$(EXE_EXT) -ldflags "-X main.version=$(BLOCK_IPTABLES_VERSION)" -gcflags="-dwarflocationlists=true"
+
+# Libraries for block-iptables
+block-iptables-lib:
```
jpayne3506: `block-iptables-lib` and `ipv6-hp-bpf-lib` recipes match. Rename to `bpf-prog-lib`? Seems like the libs are generic to bpf.

.pipelines/templates/run-unit-tests.yaml
```diff
@@ -14,6 +14,8 @@ stages:
           - script: |
               set -e
               make tools
+              make bpf-lib
+              go generate ./...
```
QxBytes, line 18: the ipv6-hp-bpf program also requires generated code-- I'm assuming the ipv6 bpf program didn't need to run unit tests which is why this is the first time I see that we need to go generate?

Makefile
```diff
@@ -201,6 +209,11 @@ else ifeq ($(GOARCH),arm64)
 	for dir in /usr/include/aarch64-linux-gnu/*; do sudo ln -sfn "$$dir" /usr/include/$$(basename "$$dir"); done
 endif
 
+# Build the block-iptables binary.
+block-iptables-binary:
+	cd $(BLOCK_IPTABLES_DIR) && CGO_ENABLED=0 go generate ./...
+	cd $(BLOCK_IPTABLES_DIR)/cmd/block-iptables && CGO_ENABLED=0 go build -v -o $(BLOCK_IPTABLES_BUILD_DIR)/block-iptables$(EXE_EXT) -ldflags "-X main.version=$(BLOCK_IPTABLES_VERSION)" -gcflags="-dwarflocationlists=true"
```
QxBytes, line 215: do we use main.version in the bpf main package? There doesn't seem to be any variable named version there.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
```diff
@@ -0,0 +1,214 @@
...
+	if err := p.objs.EventCounter.Pin(pinPath); err != nil {
+		return fmt.Errorf("failed to pin event counter map to %s: %w", pinPath, err)
+	}
+
+	log.Printf("Event counter map pinned to %s", pinPath)
+	return nil
+}
+
+// unpinEventCounterMap unpins the event counter map from the filesystem
+func (p *Program) unpinEventCounterMap() error {
```
QxBytes, line 67: do we need to return an error here? It always seems to return nil.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
```diff
@@ -0,0 +1,214 @@
...
+	// Load BPF objects with the host namespace inode set
+	spec, err := blockservice.LoadBlockIptables()
+	if err != nil {
+		return fmt.Errorf("failed to load BPF spec: %w", err)
+	}
+
+	// Set the host_netns_inode variable in the BPF program before loading
+	// Note: The C program sets it to hostNetnsInode + 1, so we do the same
+	if err := spec.RewriteConstants(map[string]interface{}{
+		"host_netns_inode": hostNetnsInode,
```
QxBytes, line 116: what does the comment mean here? Is this value on 112 supposed to be +1'd?

bpf-prog/block-iptables/pkg/bpfprogram/program.go
```diff
@@ -0,0 +1,214 @@
...
+	options := &ebpf.CollectionOptions{
+		Maps: ebpf.MapOptions{
+			PinPath:        BPFMapPinPath,
+			LoadPinOptions: ebpf.LoadPinOptions{},
+		},
+	}
+	if err := spec.LoadAndAssign(objs, options); err != nil {
+		return fmt.Errorf("failed to load BPF objects: %w", err)
+	}
+	p.objs = objs
```
QxBytes, line 132: if there's an error below do we need to set this to nil after closing it?

bpf-prog/block-iptables/pkg/bpfprogram/program.go
```diff
@@ -0,0 +1,214 @@
...
+	for _, l := range p.links {
+		if err := l.Close(); err != nil {
+			log.Printf("Warning: failed to close link: %v", err)
+		}
+	}
+	p.links = nil
+
+	// Close objects
+	if p.objs != nil {
+		p.objs.Close()
```
QxBytes: should we log if there is an error on closing the objects (here and elsewhere)?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,181 @@
...
+        return 0;
+
+    if (!data)
+        return 0;
+
+    if (bpf_core_read(&skb_len, sizeof(skb_len), &skb->len) < 0)
+        return 0;
+
+    #pragma unroll
+    for (int i = 0; i < 4; i++) {
```
QxBytes: is this like looking at 4 netlink messages in the buffer at a time? Or what is the reason for this number?

santhoshmprabhu: BPF requires loops to be bounded, so we need to have some limit. Also, it's not really 4 messages at a time, it's 4 messages overall. In the experiments I've tried with iptables, the rule add message seems to be the second one always (after a "batch begin"). We could raise it or lower it, but we definitely need a limit.

.pipelines/templates/run-unit-tests.yaml
santhoshmprabhu: correct.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
santhoshmprabhu: This was flagged in linting too, fixing it.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
santhoshmprabhu: I was testing it to make sure it was being set correctly, and copilot went ahead and added a comment. Removed it now.

Makefile
santhoshmprabhu: No, but it seems we do set it for for all our binaries.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
santhoshmprabhu: I don't think it makes a difference, but I've set it to nil.

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,181 @@
...
+#include <bpf/bpf_tracing.h>
+
+#define sk_family __sk_common.skc_family
+#define EPERM 1
+#define IPT_SO_SET_REPLACE 64
+#define TASK_COMM_LEN 16
+#define COMM_COUNT 3
+
+char __license[] SEC("license") = "Dual MIT/GPL";
+volatile const u64 host_netns_inode = 4026531840; // Initialized by userspace
```
tamilmani1989, line 30: is this inode fixed for all nodes?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,181 @@
...
+    __uint(pinning, LIBBPF_PIN_BY_NAME);
+} event_counter SEC(".maps");
+
+int is_allowed_parent ()
+{
+    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+    struct task_struct *parent_task = NULL;
+
+    char parent_comm[TASK_COMM_LEN] = {};
+    const char target_prefixes[COMM_COUNT][TASK_COMM_LEN] = {"cilium-agent", "ip-masq", "azure-cns"};
```
tamilmani1989: can we define as #define constants?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    if (family != 16) // Not AF_NETLINK
+        return 0;
+
+    if (sk != NULL) {
+        bpf_probe_read_kernel(&proto, sizeof(proto), &sk->sk_protocol);
+    }
+
+    if (proto != 12) // Not NETLINK_NETFILTER
```
tamilmani1989: lets define numbers as constants

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,181 @@
...
+
+    if (!is_host_ns()) {
+        return 0;
+    }
+
+    struct nlmsghdr nlh = {};
+    void *data = NULL;
+    __u32 skb_len = 0;
+
+    if (bpf_core_read(&data, sizeof(data), &skb->data) < 0)
```
tamilmani1989, line 178: here size of data will be size of pointer and not actual structure. is that expected?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,181 @@
...
+
+        if (bpf_probe_read_kernel(&nlh, sizeof(nlh), data) < 0)
+            return 0;
+        __u16 type = nlh.nlmsg_type;
+        __u8 subsys_id = type >> 8;
+        __u8 cmd = type & 0xFF;
+        __u32 nlmsg_len = nlh.nlmsg_len;
+
+        if (subsys_id == 10 && cmd == 6) {
+            if(is_allowed_parent()) {
```
tamilmani1989: can we add more comments in code on what we are reading and what we are validating, why iterating 4 times etc

bpf-prog/block-iptables/cmd/block-iptables/main.go
```diff
@@ -0,0 +1,196 @@
...
+	case 1: // File is empty
+		log.Println("File is empty, attaching BPF program")
+		if err := bp.Attach(); err != nil {
+			return fmt.Errorf("failed to attach BPF program: %w", err)
+		}
+	case 0: // File has content
+		log.Println("Config file has content, BPF program will remain detached")
+	case -1: // File is missing
+		log.Println("Config file is missing, waiting for file to be created...")
+	}
```
tamilmani1989: why this is repetitive. i see same logic above?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: Yes. you can see we're reading to `&data`. We later read from the address that we're copying here (line 160).

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: This the value I've seen so far on Ubuntu nodes, for host net ns. I don't think there's any guarantee that it is this value, but it's a fairly okay default. 

We do have the loader program read and update the true value when the BPF program is loaded.

tamilmani1989: ok. how do we get real value?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,181 @@
...
+#define EPERM 1
+#define IPT_SO_SET_REPLACE 64
+#define TASK_COMM_LEN 16
+#define COMM_COUNT 3
+
+char __license[] SEC("license") = "Dual MIT/GPL";
+volatile const u64 host_netns_inode = 4026531840; // Initialized by userspace
+
+struct {
+    __uint(type, BPF_MAP_TYPE_ARRAY);
```
tamilmani1989, line 33: why this can't be percpu array?

bpf-prog/block-iptables/cmd/block-iptables/main.go
tamilmani1989: i think this runs on startup?

tamilmani1989 (commented):
can you also added a comment at top of file to indicate if generated file

Makefile
santhoshmprabhu: Printing out the version in the binary, as discussed with @QxBytes.

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: We read from `stat /proc/self/ns/net`.

See https://github.com/Azure/azure-container-networking/pull/3871/files#diff-3fe83f77f4cac2badd9ae3aa6810a1e3a3908c5c8d7c6990a89dec60370609b8R76

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: Done

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: Done

Makefile
```diff
@@ -111,6 +114,7 @@ AZURE_IPAM_ARCHIVE_NAME = azure-ipam-$(GOOS)-$(GOARCH)-$(AZURE_IPAM_VERSION).$(A
 AZURE_IP_MASQ_MERGER_ARCHIVE_NAME = azure-ip-masq-merger-$(GOOS)-$(GOARCH)-$(AZURE_IP_MASQ_MERGER_VERSION).$(ARCHIVE_EXT)
 AZURE_IPTABLES_MONITOR_ARCHIVE_NAME = azure-iptables-monitor-$(GOOS)-$(GOARCH)-$(AZURE_IPTABLES_MONITOR_VERSION).$(ARCHIVE_EXT)
 IPV6_HP_BPF_ARCHIVE_NAME = ipv6-hp-bpf-$(GOOS)-$(GOARCH)-$(IPV6_HP_BPF_VERSION).$(ARCHIVE_EXT)
+BLOCK_IPTABLES_ARCHIVE_NAME = block-iptables-$(GOOS)-$(GOARCH)-$(BLOCK_IPTABLES_VERSION).$(ARCHIVE_EXT)
```
vipul-21, line 117: curious to know what happens it `GOOS` is windows/darwin here ?

bpf-prog/block-iptables/cmd/block-iptables/main.go
santhoshmprabhu: One runs the first time if the file is already present, the other when a change is notified. I'll refactor the common code out.

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: Discussed offline - percpu is useful when there is high contention, in this case a single array isn't a big performance problem with the level of concurrency we have.

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+            return 0;
+
+        // type variable holds subsystem ID and command
+        // subsys_id is the upper byte and cmd is the lower byte of the type
+        __u16 type = nlh.nlmsg_type;
+        __u8 subsys_id = type >> 8;
+        __u8 cmd = type & 0xFF;
+        __u32 nlmsg_len = nlh.nlmsg_len;
+
+        if (subsys_id == NFNL_SUBSYS_NFTABLES && cmd == NFT_MSG_NEWRULE) {
```
tamilmani1989, line 206: is updaterule exists? should we look for that too?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+    if (bpf_core_read(&skb_len, sizeof(skb_len), &skb->len) < 0)
+        return 0;
+
+    // Check the first NETLINK_MSG_COUNT messages. We cap the number of messages
+    // at 4 to make the verifier happy. We have seen that typically NEWRULE messages
+    // appear as the second message. 
+    #pragma unroll
+    for (int i = 0; i < NETLINK_MSG_COUNT; i++) {
+        if (skb_len < sizeof(struct nlmsghdr))
+            return 0;
```
tamilmani1989, line 194: can the message be partial and we are skip reading it?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+            // If not allowed, increment the event counter and return -EPERM.
+            if(is_allowed_parent()) {
+                    return 0;
+            } else {
+                increment_event_counter();
+                return -EPERM;
+            }
+        }
+
+        data = data + nlmsg_len;
```
tamilmani1989, line 220: why we need to move data pointer? how we make sure its not going out of bounds

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: Netlink doesn't have fragmentation the way IP does, so each message is self contained.

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: No, only NEW, GET and DELETE

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: We're reading up to 4 messages, data points to the beginning of each message. skb_len tracks the number of bytes remaining, if it's smaller than a single message header size, we quit, we don't read from the data pointer. It is also worth pointing out that the verifier would block any out-of-bound memory reads.

tamilmani1989 (dismissed):

tamilmani1989 (dismissed):

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+    __uint(type, BPF_MAP_TYPE_ARRAY);
+    __uint(max_entries, 1);
+    __type(key, u32);
+    __type(value, u64);
+    __uint(pinning, LIBBPF_PIN_BY_NAME);
+} event_counter SEC(".maps");
+
+// This function checks if the parent process of the current task is allowed to install iptables rules.
+// It checks the parent's command name against a predefined list of allowed prefixes.
+int is_allowed_parent ()
```
anubhabMajumdar: make this Boolean for verbosity?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+char __license[] SEC("license") = "Dual MIT/GPL";
+volatile const u64 host_netns_inode = 4026531840; // Initialized by userspace
+
+struct {
+    __uint(type, BPF_MAP_TYPE_ARRAY);
+    __uint(max_entries, 1);
+    __type(key, u32);
+    __type(value, u64);
+    __uint(pinning, LIBBPF_PIN_BY_NAME);
+} event_counter SEC(".maps");
```
anubhabMajumdar: More verbose map name? Like `iptable_event_counter` or something similar?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+        int match = 1;
+        for (int i = 0; i < TASK_COMM_LEN && target_prefixes[p][i] != '\0'; i++) {
+            if (parent_comm[i] != target_prefixes[p][i]) {
+                match = 0;
+                break;
+            }
+        }
+
+        if(match) {
+            bpf_printk("Allowed netlink from parent: %s\n", parent_comm);
```
anubhabMajumdar: I think it would be better to surface this through the map, similar to how you will surface the count of installation stopped by the eBPF program.

```
event_counter[1] = 10; // 1 = cilium agent
event_counter[2] = 5; // 2 = azure cns
```

anubhabMajumdar: Also, is this log required in production? Seems like it's here for debugging purpose.

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+
+#define CILIUM_AGENT "cilium-agent"
+#define IP_MASQ "ip-masq"
+#define AZURE_CNS "azure-cns"
+
+char __license[] SEC("license") = "Dual MIT/GPL";
+volatile const u64 host_netns_inode = 4026531840; // Initialized by userspace
+
+struct {
+    __uint(type, BPF_MAP_TYPE_ARRAY);
```
anubhabMajumdar, line 33: Would it be useful to use something like `BPF_MAP_TYPE_LRU_HASH` and surface more verbose events? Like:
- how many iptable rules were installed by allows processes (keyed by process name)
- which processes recently tried to install iptable rules in host namespace and how many times

We can surface all the information we collect and userspace can decide which to present to user?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+
+    return 0;
+}
+
+// blocking hook for iptables-nftables rule installation
+SEC("lsm/netlink_send")
+int BPF_PROG(iptables_nftables_block, struct sock *sk, struct sk_buff *skb) {
+    __u16 family = 0, proto = 0;
```
anubhabMajumdar: Should we return immediately if `sk` or `skb` is nil?

anubhabMajumdar: Then we don't need the not nil checks after this.

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
```diff
@@ -0,0 +1,219 @@
...
+    if (sk != NULL) {
+        bpf_probe_read_kernel(&proto, sizeof(proto), &sk->sk_protocol);
+    }
+
+    // Check if the protocol is NETLINK_NETFILTER
+    // This is the protocol used for netfilter messages
+    if (proto != NETLINK_NETFILTER) 
+        return 0;
+
+    if (!is_host_ns()) {
```
anubhabMajumdar, line 169: Should we move this check up? This check can eliminate the need for `bpf_probe_read_kernel` before this. `bpf_probe_read_kernel` can be expensive.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
```diff
@@ -0,0 +1,223 @@
...
+
+	blockservice "github.com/Azure/azure-container-networking/bpf-prog/block-iptables/pkg/blockservice"
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/link"
+	"github.com/pkg/errors"
+)
+
+const (
+	// BPFMapPinPath is the directory where BPF maps are pinned
+	BPFMapPinPath = "/sys/fs/bpf/block-iptables"
```
anubhabMajumdar, line 20: Do we need a new directory for this? I would imagine one map would be enough for surfacing all the information from kernel. So we can just keep it under `/sys/fs/bpf`.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
```diff
@@ -0,0 +1,223 @@
...
+		return errors.Wrapf(err, "failed to remove pinned map %s", pinPath)
+	}
+
+	log.Printf("Event counter map unpinned from %s", pinPath)
+	return nil
+}
+
+func getHostNetnsInode() (uint64, error) {
+	var stat syscall.Stat_t
+	err := syscall.Stat("/proc/self/ns/net", &stat)
```
anubhabMajumdar: Make `/proc/self/ns/net` a constant?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: We _allow_ cilium agent, CNS and ip-masq-agent, so we aren't really tracking the attempts from those components. If any other process tries to install rules, then we block and raise the count.

Are you suggesting to remove the `printk`? We could, but then we likely want to track both allowed and disallowed rule installations in the map. Right now, the map is reflecting the number of blocks, to be read from CNS.

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: We could do that, it's more than what the PM ask is - to block iptables rule addition, and to raise events when it's attempted. We could enhance it of course, but we are also expecting that the component that is trying the rule installation will also fail visibly, so the customer will have this knowledge independently.

In any case, can we take this as a future enhancement?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: Done

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: is_host_ns has 3 `BPF_CORE_READ` calls, which apparently expand into `bpf_probe_read_kernel`. This is at the end of the day a judgement call of which will help you terminate faster, I can move it up if you think that'll be better.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
santhoshmprabhu: We already have 2 other dirs, so this basically just adds a third:
```
ls /sys/fs/bpf/
block-iptables  cilium  tc
```

bpf-prog/block-iptables/pkg/bpfprogram/program.go
santhoshmprabhu: Done

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: Named it `iptables_block_event_counter`

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: Done

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
anubhabMajumdar: Yeah, in prod, I think it would be better to surface allowed/blocked as metrics rather than logs. Metrics are persistent. And we should remove the log line because it seems like it's useful for debugging, but not for surfacing information.

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
anubhabMajumdar: Sure, this was more of a suggestion. One clarification is if it would be useful for customer to know which process is trying to apply `iptable` rule in host namespace - both from security standpoint and to fix the issue. Maybe this could be clarified from PM side and enhanced in future.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
anubhabMajumdar: Not a big deal, but `cilium` and `tc` has bunch of maps that form a logical group. Will `block-iptables` have more than one map?

bpf-prog/block-iptables/bpf/src/block_iptables.bpf.c
santhoshmprabhu: Removed the log, added separate tracking of blocked and allowed rule installs.

bpf-prog/block-iptables/pkg/bpfprogram/program.go
santhoshmprabhu: I don't yet see a reason for us to do that, but I think it's cleaner nevertheless. Unless you have a strong opposition to it, I'd rather keep the extra dir.

anubhabMajumdar (approved):

camrynl (approved):

Makefile
santhoshmprabhu: This particular variable will be defined, but we only invoke the Makefile target on Linux (this matches how we are building other bpf programs). See https://github.com/Azure/azure-container-networking/pull/3871/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R133 to see that.

tamilmani1989 (approved):

PR #3885, opened 2025-08-01
feat: add cns iptables reconciliation

QxBytes:
**Reason for Change**:

CNS currently programs snat iptables rules only on NNC update/create. This PR makes it so that it also does so during cns bootup. Additionally, CNS will now attempt to "reconcile" the iptables on the node to what it expects. This allows us to "rollback" future updates to CNS iptables rules to match the expected state each time CNS reboots or restarts.

It is currently possible for CNI to add iptables rules to the node using chain "SWIFT". We want to jump to the cns-controlled chain "SWIFT-POSTROUTING" before we jump to the chain "SWIFT". The first piece of logic is to migrate to that:

Case 1: CNI already added a jump to SWIFT chain in nat POSTROUTING chain and it is at a higher priority than any SWIFT-POSTROUTING jump we may have added: We find the position of the SWIFT jump rule, and insert our SWIFT-POSTROUTING jump at that position, pushing the SWIFT jump rule to a lower priority
```
POSTROUTING
jump to KUBE POSTROUTING
...
jump to SWIFT
jump to SWIFT-POSTROUTING
```
or
```
POSTROUTING
jump to KUBE POSTROUTING
...
jump to SWIFT
```

Case 2: The SWIFT-POSTROUTING jump rule exists before the SWIFT jump rule, or exists without any SWIFT jump rule: We do not modify the nat POSTROUTING chain
```
POSTROUTING
jump to KUBE POSTROUTING
...
jump to SWIFT-POSTROUTING
jump to SWIFT
```
or
```
POSTROUTING
jump to KUBE POSTROUTING
...
jump to SWIFT-POSTROUTING
```

Case 3: The nat POSTROUTING chain does not contain a SWIFT jump rule nor a SWIFT-POSTROUTING rule: We append the SWIFT-POSTROUTING jump rule to the nat POSTROUTING chain
```
POSTROUTING
jump to KUBE POSTROUTING
...
```

Now, packets will go to the nat SWIFT-POSTROUTING chain first. We then take a look at the SWIFT-POSTROUTING chain. 
Case 1: The nat SWIFT-POSTROUTING chain does not contain every single rule we expect exactly, or the # of rules in the SWIFT-POSTROUTING chain does not match how many rules we expect: Flush the chain and reinstall the iptables rules we expect.

Case 2: The nat SWIFT-POSTROUTING chain contains every single rule we expect exactly, and the # of rules matches how many rules we expect: We do not modify or flush the SWIFT-POSTROUTING chain

**Issue Fixed**:

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Notes**:
- Check iptables insert of bounds just in case
- When we list iptables rules, the first rule is always `-N ...` for custom chains or `-P ...` for built in chains (not shown in `iptables -nvL -t nat`). Feel free to check-- the library calls `iptables -t <table> -S <chain>` under the hood.
- When we list iptables rules, the rules use -A syntax when listed no matter how they were added (via insert, append, etc.)
- IPTables is 1-indexed, so insert at position 1 will be the first rule in the chain-- so when using `iptables -nvL -t nat` the first rule that shows up in a chain is at index 1.

### DIFF
--- a/cns/restserver/internalapi_linux.go
+++ b/cns/restserver/internalapi_linux.go
@@ -72,6 +72,7 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 	if swiftRuleIndex != -1 {
 		// jump SWIFT rule exists, insert SWIFT-POSTROUTING rule at the same position so it ends up running first
 		// first, remove any existing SWIFT-POSTROUTING rules to avoid duplicates
+		// note: inserting at len(rules) and deleting a jump to SWIFT-POSTROUTING is mutually exclusive
 		swiftPostroutingExists, err := ipt.Exists(iptables.Nat, iptables.Postrouting, "-j", SWIFT)
 		if err != nil {
 			return types.UnexpectedError, fmt.Sprintf("[Azure CNS] Error. Failed to check for existence of SWIFT-POSTROUTING rule: %v", err)
--- a/cns/restserver/internalapi_linux.go
+++ b/cns/restserver/internalapi_linux.go
@@ -39,30 +39,59 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 
 	chainExist, err := ipt.ChainExists(iptables.Nat, SWIFT)
 	if err != nil {
-		return types.UnexpectedError, fmt.Sprintf("[Azure CNS] Error. Failed to check for existence of SWIFT chain: %v", err)
+		return types.UnexpectedError, fmt.Sprintf("[Azure CNS] Error. Failed to check for existence of SWIFT-POSTROUTING chain: %v", err)
 	}
 	if !chainExist { // create and append chain if it doesn't exist
 		logger.Printf("[Azure CNS] Creating SWIFT Chain ...")
 		err = ipt.NewChain(iptables.Nat, SWIFT)
 		if err != nil {
-			return types.FailedToRunIPTableCmd, "[Azure CNS] failed to create SWIFT chain : " + err.Error()
-		}
-		logger.Printf("[Azure CNS] Append SWIFT Chain to POSTROUTING ...")
-		err = ipt.Append(iptables.Nat, iptables.Postrouting, "-j", SWIFT)
-		if err != nil {
-			return types.FailedToRunIPTableCmd, "[Azure CNS] failed to append SWIFT chain : " + err.Error()
+			return types.FailedToRunIPTableCmd, "[Azure CNS] failed to create SWIFT-POSTROUTING chain : " + err.Error()
 		}
 	}
 
-	postroutingToSwiftJumpexist, err := ipt.Exists(iptables.Nat, iptables.Postrouting, "-j", SWIFT)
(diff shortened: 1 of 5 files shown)


cns/restserver/internalapi_linux.go
```diff
@@ -37,32 +37,62 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
...
+			}
+		}
+
+		// slice index is 0-based, iptables insert is 1-based, but list also gives us the -P POSTROUTING ACCEPT
+		// as the first rule so swiftRuleIndex gives us the correct 1-indexed iptables position.
+		// Example:
+		// -P POSTROUTING ACCEPT is at swiftRuleIndex 0
+		// -A POSTROUTING -j SWIFT is at swiftRuleIndex 1, and iptables index 1
+		logger.Printf("[Azure CNS] Inserting SWIFT-POSTROUTING Chain at iptables position %d", swiftRuleIndex)
+		err = ipt.Insert(iptables.Nat, iptables.Postrouting, swiftRuleIndex, "-j", SWIFTPOSTROUTING)
```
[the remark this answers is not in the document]
QxBytes: rules correspond to what is received from List(), which includes the built-in -P or -N rule. So, if you see one rule with `iptables -nvL -t nat`, List gives us the built in rule (-P or -N) and also the one rule we see, leading to a length of 2. If we insert using iptables index 2, it ends up at the end of the chain (after the one rule we see from `iptables -nvL -t nat`.

cns/restserver/internalapi_linux.go
```diff
@@ -71,39 +101,47 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 		// put the ip address in standard cidr form (where we zero out the parts that are not relevant)
 		_, podSubnet, _ := net.ParseCIDR(v.IPAddress + "/" + fmt.Sprintf("%d", req.IPConfiguration.IPSubnet.PrefixLength))
 
-		snatUDPRuleExists, err := ipt.Exists(iptables.Nat, SWIFT, "-m", "addrtype", "!", "--dst-type", "local", "-s", podSubnet.String(), "-d", networkutils.AzureDNS, "-p", iptables.UDP, "--dport", strconv.Itoa(iptables.DNSPort), "-j", iptables.Snat, "--to", ncPrimaryIP.String())
-		if err != nil {
-			return types.UnexpectedError, fmt.Sprintf("[Azure CNS] Error. Failed to check for existence of pod SNAT UDP rule : %v", err)
+		// define all rules we want in the chain
+		rules := [][]string{
```
santhoshmprabhu, line 105: can we add this at the top?

santhoshmprabhu:
I think we could consider eliminating that blip from clearing the chain, but not a big deal.

cns/restserver/internalapi_linux.go
QxBytes: rules is dependent on `podSubnet.String()` which is dependent on the ip address we get in the for loop

santhoshmprabhu (approved):

rbtr (approved):

PR #3878, opened 2025-07-31
MTPNC CRD changes for NUMA-Aware Infiniband pods

pjohnst5:
**Reason for Change**:
MTPNC CRD changes for NUMA-Aware pod IB device programming

The partner team will call into CNS (still discussing how in #3825) but regardless, CNS will need to pass which IB devices to use for a pod, and DNC-RC will need to pass back to CNS the programming status of said devices

PR #3886, opened 2025-08-04
Updating base golang image

pjohnst5:
Just updating the base golang image again

I'm curious:
- Is it preferential that I do this in a separate PR every time? (to keep PRs scoped to just 1 thing)
- Are there any automations to do this sort of thing like dependabot does?

### DIFF
--- a/cni/Dockerfile
+++ b/cni/Dockerfile
@@ -6,7 +6,7 @@ ARG OS_VERSION
 ARG OS
 
 # mcr.microsoft.com/oss/go/microsoft/golang:1.23-azurelinux3.0
-FROM --platform=linux/${ARCH} mcr.microsoft.com/oss/go/microsoft/golang@sha256:7d33a8015c48c06e97ecd5139181594b550d4d4d6a9d7fb19083192541213753 AS go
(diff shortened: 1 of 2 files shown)


pjohnst5:
The reason i do these separately, is let's say I lump it into #3878

Then if we need to revert #3878, the base golang change would also be reverted

So I like to separate them just in case, but also curious others' points of view

pjohnst5:
@jpayne3506 would you be able to review this for me please? 🙏🏻

jpayne3506:
@rbtr This looks to be consistent pain point for people. IMO, I would want this to be updated before every release, but I also don't want this to be put onto every PR as it would impact backports ect. 

> Are there any automations to do this sort of thing like dependabot does?

I have seen solutions with Renovate, but ~I don't believe dependabot updates~ we have not been using dependabot correctly for our dockerfiles. We can test it out with a follow on PR. 
- relevant thread https://github.com/dependabot/dependabot-core/discussions/9120 
- https://github.com/Azure/azure-container-networking/network/updates/17351297/jobs

rbtr:
it's not a blocking check

pjohnst5:
@rbtr would you be willing to approve this one?

rbtr (approved):

PR #3940, opened 2025-08-15
fix: remove veth pair in vm ns if previously leaked and fix validation

QxBytes:
**Reason for Change**:

In transparent-vlan mode, removes the vnet veth interface and container veth interface in the vm namespace if they exist prior to creating the pair in the vm namespace. This won't disrupt existing connections because these pairs are one per container, and if either side of the veth pair were in the vm namespace, the container's networking would be broken. The vnet veth interface must be in the vnet namespace and the container veth interface must be in the container namespace in a working setup (otherwise it is broken and we need to clean up). Removing one side of the veth should remove the other.

The above shouldn't be necessary since we already validate the existence of the veth interfaces on ADD, but it seems like sometimes the veth creation can pass validation, but then disappear for a short period of time before re-appearing, bypassing the cleanup logic during the add.

Also fixes an improper validation check after moving the vnet veth into the vnet namespace (though this did not cause the issue that triggered this fix). Previously it would check the wrong interface name, now it will check the interface name passed in the namespace passed in.

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
Tested on a multitenancy linux transparent vlan setup with no issues
0c998649d31, consolidate logic and improve uts

### DIFF
--- a/network/transparent_vlan_endpointclient_linux.go
+++ b/network/transparent_vlan_endpointclient_linux.go
@@ -294,7 +294,7 @@ func (client *TransparentVlanEndpointClient) PopulateVM(epInfo *EndpointInfo) er
 		}, numRetries, sleepInMs)
 
 		if err != nil {
-			deleteNSIfNotNilErr = errors.Wrapf(err, "failed to get vlan interface interface:%s", client.vlanIfName)
+			deleteNSIfNotNilErr = errors.Wrapf(err, "failed to get vlan interface: %s", client.vlanIfName)
 			return deleteNSIfNotNilErr
 		}
 


network/transparent_vlan_endpointclient_linux.go
```diff
@@ -310,6 +310,16 @@ func (client *TransparentVlanEndpointClient) PopulateVM(epInfo *EndpointInfo) er
 		logger.Info("Failed to parse the mac address", zap.String("defaultHostVethHwAddr", defaultHostVethHwAddr))
 	}
 
+	// Proactively clean up any leftover veth interfaces before creating new ones
+	if err = client.netlink.DeleteLink(client.vnetVethName); err != nil {
+		logger.Info("Could not proactively clean up vnet veth (expected if not present)",
```
behzad-mir: Can we do better error handling here? If it is only due to the fact that veth does not exist then we can log it with a different message.

behzad-mir (dismissed):

network/transparent_vlan_endpointclient_linux.go
QxBytes: If the veth does not exist already it does not throw an error-- see netlink/link_linux.go > DeleteLink -- I've updated the logger message accordingly

behzad-mir (approved):
lgtm

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -194,9 +194,9 @@ func (client *TransparentVlanEndpointClient) setLinkNetNSAndConfirm(name string,
 	// confirm veth was moved successfully
 	err = RunWithRetries(func() error {
 		// retry checking in the namespace if the interface is not detected
-		return ExecuteInNS(client.nsClient, client.vnetNSName, func() error {
-			_, ifDetectedErr := client.netioshim.GetNetworkInterfaceByName(client.vlanIfName)
-			return errors.Wrap(ifDetectedErr, "failed to get vlan veth in namespace")
+		return ExecuteInNS(client.nsClient, nsName, func() error {
```
vipul-21, line 206: What's the difference in passing the `nsName` rather than using `client.vnetNSName` ? We are still passing `client.vnetNSName` as `nsName` always. Is the client different in `setLinkNetNSAndConfirm` calls ?

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -194,9 +194,9 @@ func (client *TransparentVlanEndpointClient) setLinkNetNSAndConfirm(name string,
...
+			_, ifDetectedErr := client.netioshim.GetNetworkInterfaceByName(name)
```
vipul-21, line 207: Does this mean it was checking the wrong interface? If yes, won't this fail consistently.
How was this function successful prior to this change?

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -310,6 +310,16 @@ func (client *TransparentVlanEndpointClient) PopulateVM(epInfo *EndpointInfo) er
 		logger.Info("Failed to parse the mac address", zap.String("defaultHostVethHwAddr", defaultHostVethHwAddr))
 	}
 
+	// Proactively clean up any leftover veth interfaces before creating new ones
+	if err = client.netlink.DeleteLink(client.vnetVethName); err != nil {
```
vipul-21: just fyi, this will be called for every container. And most likely we won't be deleting any interface. This will increase the log lines getting added? Any impact on app insight?

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -310,6 +310,16 @@ func (client *TransparentVlanEndpointClient) PopulateVM(epInfo *EndpointInfo) er
...
+		logger.Info("Could not proactively clean up vnet veth",
```
vipul-21: How does this cleanup different from the one we have during the ADD call ? It can happen at this point to ?referring to this:

> The above shouldn't be necessary since we already validate the existence of the veth interfaces on ADD, but it seems like sometimes the veth creation can pass validation, but then disappear for a short period of time before re-appearing, bypassing the cleanup logic during the add.

Curious to know if we know the reason for disappearing for a short period of time.

network/transparent_vlan_endpointclient_linux.go
QxBytes: yes it checked the wrong interface previously. By the time this function was called a second time, client.vlanIfName is guaranteed to exist since it is added earlier in the code. So it would always succeed without error (even if "name" did not exist).

network/transparent_vlan_endpointclient_linux.go
QxBytes: there isn't a difference right now but if in the future we want to move a link to a namespace other than client.vnetNSName, we could do so without changing this code.

network/transparent_vlan_endpointclient_linux.go
QxBytes: no error or log will be generated if the device does not exist-- see https://github.com/Azure/azure-container-networking/pull/3940#discussion_r2289432805

network/transparent_vlan_endpointclient_linux.go
QxBytes: yes we believe the delete link could silently fail, but if that happens, the next call to create the veth pair will fail the ADD call (file exists... ) and then the container runtime will retry, and eventually the leaked interfaces should be deleted. we are investigating the reason for the disappearance with the azure linux team at the moment.

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -110,11 +110,24 @@ func NewTransparentVlanEndpointClient(
 	return client
 }
 
+// cleanupInterfaceIfExists checks if an interface exists and deletes it if found
+// Returns an error if the deletion fails
+func (client *TransparentVlanEndpointClient) cleanupInterfaceIfExists(interfaceName string) error {
```
vipul-21, line 115: `TransparentVlanEndpointClien` does not feel like the right place to implement the `cleanupInterfaceIfExists` as it is not related to just `TransparentVlanEndpointClient`. But I see it uses `netlink` and `netioshim` so this might be the only place we can add without disrupting a lot of code but wanted to check if there is an alternative?

QxBytes: it is generalizable if we pass in the netio and netlink clients but right now only transparent vlan does this-- if others need it it can always be pulled out into utils or similar

PR #3947, opened 2025-08-19
[WIP] Populate home AZ cache if it's not filled

smittal22:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/homeazmonitor.go
+++ b/cns/restserver/homeazmonitor.go
@@ -50,14 +50,25 @@ func (h *HomeAzMonitor) updateCacheValue(resp cns.GetHomeAzResponse) {
 
 // readCacheValue reads home az cache value
 func (h *HomeAzMonitor) readCacheValue() cns.GetHomeAzResponse {
-	cachedResp, found := h.values.Get(homeAzCacheKey)
-	if !found {
-		return cns.GetHomeAzResponse{Response: cns.Response{
+	if cachedResp, found := h.values.Get(homeAzCacheKey); found {
+		return cachedResp.(cns.GetHomeAzResponse)
+	}
+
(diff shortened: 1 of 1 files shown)


cns/restserver/homeazmonitor.go
```diff
@@ -50,14 +50,27 @@ func (h *HomeAzMonitor) updateCacheValue(resp cns.GetHomeAzResponse) {
...
+	h.Populate(ctx)
+
+	if cachedResp, found := h.values.Get(homeAzCacheKey); found {
+		return cachedResp.(cns.GetHomeAzResponse)
+	}
+
+	// Still no cache value after populate attempt
+	return cns.GetHomeAzResponse{
+		Response: cns.Response{
 			ReturnCode: types.NotFound,
```
ZetaoZhuang, line 69: I think we should not return NotFound now. Initially we want to return NotFound when the cns start, but the request coming before the home az goroutine finish populating, and in that case, the cache is empty.  but now we know the populate function at least get execution once, then we can just return back cache values for GetHomeAzResponse.

smittal22: What do we return if the cache population fails even after we make an explicit attempt? DNC will still base its decision on whether homeAZ is supported or not in prepare.go

ZetaoZhuang: if cns already made an explicit attempt and failed, it would cache the error and returned whatever the error it got to dnc.

smittal22: Discussed with Zetao offline. Keeping it as is for now.

PR #3948, opened 2025-08-19
feat: update iptables monitor with ipv6 and bpf map reading capabilities

QxBytes:
**Reason for Change**:

Adds two new capabilities to azure-iptables-monitor:
- Reads ipv6 iptables rules if present on the node. To accomodate ipv6, will now read config files from a new directory for allowed ipv6 iptables rules-- these config files are separate from the ipv4 iptables allowlist. The ipv6 directory has the same structure as the ipv4 one (ex: an allowlist pattern file for each ipv6 table). If there are either ipv4 *or* ipv6 rules that are unexpected, we send an event and set the ciliumnode label as before.
- Can read a pinned bpf map at a configurable location. The bpf map records how times on the node an iptables rule add request was blocked by a separate iptables block binary. If the number of blocks increases between intervals, we create a new event.

**Issue Fixed**:

See above

**Requirements**:

- [x] includes documentation
- [X] adds unit tests

**Notes**:
Tested on a cilium dualstack cluster
- If ipv6 rules not in allowlist are found, confirmed user iptables rules true
- If ipv4 and ipv6 rules are all allowed, confirmed user iptables rules false
- If bpf map increases between intervals, confirmed event emitted
- Confirmed ipv6 uses an ipv6 client and shows ipv6 rules

### DIFF
--- a/azure-iptables-monitor/README.md
+++ b/azure-iptables-monitor/README.md
@@ -33,7 +33,7 @@ Follow the steps below to build and run the program:
     - The `-events` flag enables Kubernetes event creation for rule violations. Default: `false`
     - The `-ipv6` flag enables IPv6 ip6tables monitoring using the IPv6 allowlists. Default: `false`
     - The `-checkMap` flag enables checking the pinned bpf map specified in mapPath for increases. Default: `false`
-    - The `-mapPath` flag specifies the pinned bpf map path to check. Default: `/block-iptables/iptables_block_event_counter`
+    - The `-mapPath` flag specifies the pinned bpf map path to check. Default: `/azure-block-iptables/iptables_block_event_counter`
     - The program must be in a k8s environment and `NODE_NAME` must be a set environment variable with the current node.
 
 5. The program will set the `kubernetes.azure.com/user-iptables-rules` label to `true` on the specified ciliumnode resource if unexpected rules are found, or `false` if all rules match expected patterns. Proper RBAC is required for patching (patch for ciliumnodes, create for events, get for nodes).
--- a/azure-iptables-monitor/iptables_monitor.go
+++ b/azure-iptables-monitor/iptables_monitor.go
@@ -34,7 +34,7 @@ var (
 	sendEvents    = flag.Bool("events", false, "Whether to send node events if unexpected iptables rules are detected")
 	ipv6Enabled   = flag.Bool("ipv6", false, "Whether to check ip6tables using the ipv6 allowlists")
 	checkMap      = flag.Bool("checkMap", false, "Whether to check the bpf map at mapPath for increases")
-	pinPath       = flag.String("mapPath", "/block-iptables/iptables_block_event_counter", "Path to pinned bpf map")
+	pinPath       = flag.String("mapPath", "/azure-block-iptables/iptables_block_event_counter", "Path to pinned bpf map")
 )
 
 const label = "kubernetes.azure.com/user-iptables-rules"
--- a/bpf-prog/azure-block-iptables/pkg/bpfprogram/program.go
+++ b/bpf-prog/azure-block-iptables/pkg/bpfprogram/program.go
@@ -18,7 +18,7 @@ import (
 
 const (
 	// BPFMapPinPath is the directory where BPF maps are pinned
-	BPFMapPinPath = "/sys/fs/bpf/block-iptables"
+	BPFMapPinPath = "/sys/fs/bpf/azure-block-iptables"
 	// EventCounterMapName is the name used for pinning the event counter map
 	EventCounterMapName = "iptables_block_event_counter"
 	// IptablesLegacyBlockProgramName is the name used for pinning the legacy iptables block program


santhoshmprabhu (commented):
In the event description, let's add an explanation indicating that eBPF host routing is enabled. The user may not immediately have context on why iptables rules are being blocked. We could also link aka.ms/acnsperformance

azure-iptables-monitor/README.md
```diff
@@ -25,14 +25,20 @@ Follow the steps below to build and run the program:
...
-    - The `--input` flag specifies the directory containing allowed regex pattern files. Default: `/etc/config/`
-    - The `--interval` flag specifies how often to check iptables rules in seconds. Default: `300`
-    - The `--events` flag enables Kubernetes event creation for rule violations. Default: `false`
+    - The `-input` flag specifies the directory containing allowed regex pattern files. Default: `/etc/config/`
+    - The `-input6` flag specifies the directory containing allowed regex pattern files for IPv6 ip6tables. Default: `/etc/config6/`
+    - The `-interval` flag specifies how often to check iptables rules in seconds. Default: `300`
+    - The `-events` flag enables Kubernetes event creation for rule violations. Default: `false`
+    - The `-ipv6` flag enables IPv6 ip6tables monitoring using the IPv6 allowlists. Default: `false`
+    - The `-checkMap` flag enables checking the pinned bpf map specified in mapPath for increases. Default: `false`
+    - The `-mapPath` flag species the pinned bpf map path to check. Default: `/block-iptables/iptables_block_event_counter`
```
tamilmani1989: can it be `/azure-block-iptables/iptables_block_event_counter`

azure-iptables-monitor/README.md
```diff
@@ -25,14 +25,20 @@ Follow the steps below to build and run the program:
...
     - The program must be in a k8s environment and `NODE_NAME` must be a set environment variable with the current node.
 
-5. The program will set the `user-iptables-rules` label to `true` on the specified ciliumnode resource if unexpected rules are found, or `false` if all rules match expected patterns. Proper RBAC is required for patching (patch for ciliumnodes, create for events, get for nodes).
+5. The program will set the `kubernetes.azure.com/user-iptables-rules` label to `true` on the specified ciliumnode resource if unexpected rules are found, or `false` if all rules match expected patterns. Proper RBAC is required for patching (patch for ciliumnodes, create for events, get for nodes).
+
+6. The program will also send out an event if the bpf map value specified increases between checks
```
tamilmani1989, line 41: can we also specify how often the program check for increases?

azure-iptables-monitor/README.md
```diff
@@ -25,14 +25,20 @@ Follow the steps below to build and run the program:
...
-    - The `--interval` flag specifies how often to check iptables rules in seconds. Default: `300`
-    - The `--events` flag enables Kubernetes event creation for rule violations. Default: `false`
+    - The `-input` flag specifies the directory containing allowed regex pattern files. Default: `/etc/config/`
+    - The `-input6` flag specifies the directory containing allowed regex pattern files for IPv6 ip6tables. Default: `/etc/config6/`
+    - The `-interval` flag specifies how often to check iptables rules in seconds. Default: `300`
+    - The `-events` flag enables Kubernetes event creation for rule violations. Default: `false`
+    - The `-ipv6` flag enables IPv6 ip6tables monitoring using the IPv6 allowlists. Default: `false`
+    - The `-checkMap` flag enables checking the pinned bpf map specified in mapPath for increases. Default: `false`
+    - The `-mapPath` flag species the pinned bpf map path to check. Default: `/block-iptables/iptables_block_event_counter`
     - The program must be in a k8s environment and `NODE_NAME` must be a set environment variable with the current node.
```
tamilmani1989, line 37: why NODE_NAME is required?

azure-iptables-monitor/README.md
QxBytes: The default uses the path specified here: https://github.com/Azure/azure-container-networking/blob/master/bpf-prog/azure-block-iptables/pkg/bpfprogram/program.go#L20 . In any case, the bpf map is on the vm and needs to be mounted to the container, and that mount can be at any location on the container. The mount location is specific to the container with the iptables monitor image. The `mapPath` is configurable as well if we need to change it later.

azure-iptables-monitor/README.md
QxBytes: The increases are checked based on `-interval`, unless you mean you would like a second configurable interval option for checking the map?

azure-iptables-monitor/README.md
QxBytes: This is to create events for the current node the program is running on-- the field needs to be passed in for the program to know about it.

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -291,6 +338,28 @@ func main() {
...
+			if err != nil {
+				klog.Errorf("failed to get bpf map value: %v", err)
+			}
+			klog.V(2).Infof("IPTables rules blocks: Previous: %d Current: %d", previousBlocks, currentBlocks)
+		}
+		// if number of blocked rules increased since last time
+		blockedRulesIncreased := currentBlocks > previousBlocks
+		if *sendEvents && blockedRulesIncreased {
+			msg := fmt.Sprintf("Blocked iptables rules increased from %d to %d since last check. "+
+				"IPTables rules blocked because EBPF Host Routing is enabled: aka.ms/acnsperformance", previousBlocks, currentBlocks)
```
santhoshmprabhu: Nit: Let's use "Iptables". in other places, we're saying "iptables".

QxBytes: changed to "iptables" since we previously referenced it in lowercase

santhoshmprabhu (approved):

santhoshmprabhu (approved):

santhoshmprabhu (approved):

PR #3930, opened 2025-08-13
feat: snat azure dns traffic to node ip in cns linux

QxBytes:
**Reason for Change**:

Changes the ip CNS-added IPTables rules SNAT to from the primary ip to node ip for **linux** podsubnet scenarios (both azure and cilium cases). CNI-added iptables rules are not modified and windows behavior remains the same (will be modified in a future PR).

**Issue Fixed**:

**Requirements**:

- [X] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [X] adds unit tests
- [X] relevant PR labels added

**Notes**:
Tested:
- Podsubnet linux upgrade from 1.7.3 --> this version --> downgrade to 1.7.3 and dns snat is as expected
- Podsubnet linux cilium ugprade from 1.7.3 --> this version --> downgrade to 1.7.3 and dns snat is as expected

PR #3958, opened 2025-08-21
feat: remove jump to swift-postrouting in iptables legacy as rules already exist in iptables nftables

QxBytes:
**Reason for Change**:

Contains same changes as #3930 but without the behavioral change of snat to node ip

Deletes iptables legacy jump to swift postrouting rule **before** adding nftables rules because:
If we delete before:
If iptables maps to iptables nftables, we should already have nftables rules programmed on the node because cns 1.7.1 has fully rolled out, so deleting first should not cause any connectivity blip.
If iptables somehow maps to iptables legacy, we have a blip and then write the rule immediately after
 
If we delete afterwards:
If iptables maps to iptables nftables, we write the rules and delete the legacy jump after as expected
If iptables somehow maps to iptables legacy, we would add the rules and then delete the jump we just added, breaking the node until the cns restarts

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
Tested upgrade from 1.7.0 --> 1.7.1 --> 1.7.2 (this PR) for cns

### DIFF
--- a/cns/restserver/internalapi_linux.go
+++ b/cns/restserver/internalapi_linux.go
@@ -45,13 +45,14 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 	ncPrimaryIP, _, _ := net.ParseCIDR(req.IPConfiguration.IPSubnet.IPAddress + "/" + fmt.Sprintf("%d", req.IPConfiguration.IPSubnet.PrefixLength))
 
 	iptl, err := service.iptables.GetIPTablesLegacy()
-	if err != nil {
-		return types.UnexpectedError, fmt.Sprintf("[Azure CNS] Error. Failed to create iptables legacy interface : %v", err)
-	}
-	err = iptl.Delete(iptables.Nat, iptables.Postrouting, "-j", SWIFTPOSTROUTING)
-	// ignore if command fails
 	if err == nil {
-		logger.Printf("[Azure CNS] Deleted legacy jump to SWIFT-POSTROUTING Chain")
+		err = iptl.Delete(iptables.Nat, iptables.Postrouting, "-j", SWIFTPOSTROUTING)
+		// ignore if command fails
+		if err == nil {
+			logger.Printf("[Azure CNS] Deleted legacy jump to SWIFT-POSTROUTING Chain")
+		}
+	} else {
+		logger.Printf("[Azure CNS] Could not create iptables legacy interface, continuing : %v", err)
 	}
 
 	ipt, err := service.iptables.GetIPTables()
--- a/cns/restserver/internalapi_linux.go
+++ b/cns/restserver/internalapi_linux.go
@@ -3,6 +3,7 @@ package restserver
 import (
 	"fmt"
 	"net"
+	"os/exec"
 	"strconv"
 
 	"github.com/Azure/azure-container-networking/cns"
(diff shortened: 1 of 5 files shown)


cns/restserver/internalapi_linux.go
```diff
@@ -22,6 +23,16 @@ func (c *IPtablesProvider) GetIPTables() (iptablesClient, error) {
...
+func (c *IPtablesProvider) GetIPTablesLegacy() (iptablesLegacyClient, error) {
+	return &iptablesLegacy{}, nil
+}
+
+type iptablesLegacy struct{}
+
+func (c *iptablesLegacy) Delete(table, chain string, rulespec ...string) error {
+	cmd := append([]string{"-t", table, "-D", chain}, rulespec...)
+	return errors.Wrap(exec.Command("iptables-legacy", cmd...).Run(), "iptables legacy failed delete")
+}
```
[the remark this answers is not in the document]
QxBytes: command called is known at compile time

santhoshmprabhu (approved):

cns/restserver/internalapi_linux.go
```diff
@@ -32,6 +43,17 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 	// in podsubnet case, ncPrimaryIP is the pod subnet's primary ip
 	// in vnet scale case, ncPrimaryIP is the node's ip
 	ncPrimaryIP, _, _ := net.ParseCIDR(req.IPConfiguration.IPSubnet.IPAddress + "/" + fmt.Sprintf("%d", req.IPConfiguration.IPSubnet.PrefixLength))
+
+	iptl, err := service.iptables.GetIPTablesLegacy()
+	if err != nil {
+		return types.UnexpectedError, fmt.Sprintf("[Azure CNS] Error. Failed to create iptables legacy interface : %v", err)
```
tamilmani1989: should we throw error or silently log and return success?

QxBytes: GetIPTablesLegacy()'s current linux implementation never throws an error-- it has an error return to be consistent with how we call GetIPTables() for non-legacy scenarios (and also so that we throw an error if somehow windows calls this function)

tamilmani1989: how windows would call this? this is linux specific file?

QxBytes: yes but GetIPTablesLegacy() is a method of the interface iptablesLegacyClient, which has a windows implementation that does return an error https://github.com/Azure/azure-container-networking/pull/3958/files#diff-9178ea91dd1ced83380fd2bd131170d325f7d21263420b3585772100894acb96R28. Because of the interface definition, the linux implementation also returns an error, so we handle it here as well, even if the linux implementation will never result in an error here https://github.com/Azure/azure-container-networking/pull/3958/files#diff-dc925c2c559d453039876553a92efd8d78d4621eabe3e311c1f0f3134972541aR26.

MikeZappa87: This abstraction seems odd. A bigger discussion here of course however referencing to Linux constructs in Windows such as iptables is odd. The design here should be rethought.

tamilmani1989 (dismissed):

cns/restserver/internalapi_linux.go
```diff
@@ -32,6 +43,18 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
...
+	if err != nil {
+		logger.Printf("[Azure CNS] Could not create iptables legacy interface, not programming snat : %v", err)
+		return types.Success, ""
+	}
+	err = iptl.Delete(iptables.Nat, iptables.Postrouting, "-j", SWIFTPOSTROUTING)
+	// ignore if command fails
+	if err == nil {
+		logger.Printf("[Azure CNS] Deleted legacy jump to SWIFT-POSTROUTING Chain")
+	}
+
```
tamilmani1989: shouldn't we move to end of this function?

santhoshmprabhu (approved):

tamilmani1989 (approved):

PR #3967, opened 2025-08-27
fix: fixing Stateless CNI delete in SwiftV2 scenario

behzad-mir:
A number of changes is made to stateless CNI to fully support SWiftV2 in Linux:

1.  For Stateless CNI, the EndpointID should include ifName when the NicType is Delegated. This ensures it can be distinguished from the InfraNIC endpoint. The reason for this behavior is that Stateless CNI uses only the ContainerID as the endpoint ID when saving the state file in CNS, to maintain consistency with Cilium. However, within CNI’s in-memory representation, the EndpointID must uniquely identify each NIC. As a result:
    - We always pass the ContainerID to CNS as the endpoint state key.
    - Internally, CNI uses EndpointID (which includes ifName for delegated NICs) to differentiate endpoints. 
8.  Delete flow has been revised and  NetNSPath has been added to the epinfo since it is needed by the TransparentClient for Frontend NIC. This is needed to move the secondary interface out of the pod namespace.
9.  The Transparent mode used by statefull CNI for SWiftV1 and V2 and stateless CNI should follow the same. TransparentVlan which is the original value seems to be a mistake.
10.  Async delete file creation in Stateless CNI has been moved to CNS_invoker to be consistent with Stateful CNI.
11.  Removing secondary interface from pod namespace in SwiftV2 Linux in case of asynchronous delete. This is needed to be covered for stateless CNI since if CNS is down, CNI does not have the secondary interface name. A dedicated function has been added to list the interface in pod namespace and remove the secondary.

For validating the scenario ADD/Delete calls have been issues on SwiftV2 cluster and and logs and satefile has been analyzed to make sure it is consistent with Stateful CNI and also nothing gets leaked. 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cns/restserver/ipam.go
```diff
@@ -1313,12 +1313,16 @@ func updateIPInfoMap(iPInfo map[string]*IPInfo, interfaceInfo *IPInfo, ifName, e
 		iPInfo[ifName].MacAddress = interfaceInfo.MacAddress
 		logger.Printf("[updateEndpoint] update the endpoint %s with MacAddress  %s", endpointID, interfaceInfo.MacAddress)
 	}
+	if interfaceInfo.NetworkNameSpace != "" {
```
MikeZappa87: We protect against empty string, what about " " ? Is that a possible condition?

cns/restserver/ipam.go
```diff
@@ -1313,12 +1313,16 @@ func updateIPInfoMap(iPInfo map[string]*IPInfo, interfaceInfo *IPInfo, ifName, e
...
+		iPInfo[ifName].NetworkNameSpace = interfaceInfo.NetworkNameSpace
+		logger.Printf("[updateEndpoint] update the endpoint %s with NetworkNameSpace  %s", endpointID, interfaceInfo.NetworkNameSpace)
```
MikeZappa87: This statement has no level, debug, info, warning, error? I see this is what is currently done above, this makes troubleshooting hard does it not? We cant filter on the level. This should be a tech debt item.

cns/restserver/ipam.go
MikeZappa87: What happens if the pod is no longer present? I don't know how the netns path is being used here. Do we check this?

network/manager.go
```diff
@@ -115,7 +115,7 @@ type NetworkManager interface {
 	DetachEndpoint(networkID string, endpointID string) error
 	UpdateEndpoint(networkID string, existingEpInfo *EndpointInfo, targetEpInfo *EndpointInfo) error
 	GetNumberOfEndpoints(ifName string, networkID string) int
-	GetEndpointID(containerID, ifName string) string
+	GetEndpointID(containerID, ifName string, nicType cns.NICType) string
```
MikeZappa87: Do we have any tests?

cns/restserver/ipam.go
behzad-mir: This is where we write to the statefile so we can store the pod information when a delete call get issued. We are just checking for empty trying to avoid writing empty to the field in statefile. We read NetNs from whatever CNI produces so I don't think it can be " ".

behzad-mir: In swiftV2 when a delete call happens the Frontened Nic needs the netNspath. we have a secondry_transparentClient tha moves interface from the Nettwork Namespace.
This is already being used by Stateful CNI and we are just cirrecting it for stateless CNI.

cns/restserver/ipam.go
behzad-mir: You're correct. We need to add levels for all of these logging here. Will address some of them in the next commit.

network/manager.go
behzad-mir: Unfortunately we don't. I am planning to add a subsequent (sister) PR to this that add UTs for some of these funcs.

santhoshmprabhu (changes requested):
I don't have a lot of context on the need for these changes. The changes look fine, but we should improve the PR a bit -
1. Update the description to cover why the opMode change
2. Address other copilot comments (one about NICType in particular seems serious)
3. Add a description of what validation steps have been carried out. Include screenshots/logs if necessary.
4. Add tests.

network/manager.go
```diff
@@ -514,7 +514,7 @@ func (nm *networkManager) DeleteEndpointState(networkID string, epInfo *Endpoint
 	nw := &network{
 		Id:           networkID, // currently unused in stateless cni
 		HnsId:        epInfo.HNSNetworkID,
-		Mode:         opModeTransparentVlan,
+		Mode:         opModeTransparent,
```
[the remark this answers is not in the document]
behzad-mir: The Transparent mode beong used by statefull CNI for SWiftV1 and V2 and stateless CNI should follow the same. TransparentVlan which is the original value seems to be a mistake

cns/restserver/ipam.go
tamilmani1989: what's the issue if we write empty namespace?

cns/restserver/ipam.go
tamilmani1989: dont we have logger.Info or logger.Error?

cns/restserver/ipam.go
```diff
@@ -1313,12 +1313,16 @@ func updateIPInfoMap(iPInfo map[string]*IPInfo, interfaceInfo *IPInfo, ifName, e
...
+	}
 }
 
 // verifyUpdateEndpointStateRequest verify the CNI request body for the UpdateENdpointState API
 func verifyUpdateEndpointStateRequest(req map[string]*IPInfo) error {
 	for ifName, InterfaceInfo := range req {
-		if InterfaceInfo.HostVethName == "" && InterfaceInfo.HnsEndpointID == "" && InterfaceInfo.NICType == "" && InterfaceInfo.MacAddress == "" {
+		if InterfaceInfo.HostVethName == "" && InterfaceInfo.HnsEndpointID == "" && InterfaceInfo.NICType == "" && InterfaceInfo.MacAddress == "" && InterfaceInfo.NetworkNameSpace == "" {
```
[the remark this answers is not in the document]
tamilmani1989: @behzad-mir can you fix this?

network/manager.go
tamilmani1989: instead of changing existing api, can we define new api `GetEndpointIDByNicType` ?

network/manager.go
tamilmani1989, line 520: was this a bug in previous PR?

cni/network/network.go
```diff
@@ -712,11 +712,11 @@ func (plugin *NetPlugin) createEpInfo(opt *createEpInfoOpt) (*network.EndpointIn
 	if opt.ifInfo.NICType == cns.InfraNIC && !*opt.infraSeen {
 		// so we do not break existing scenarios, only the first infra gets the original endpoint id generation
 		ifName = opt.args.IfName
-		endpointID = plugin.nm.GetEndpointID(opt.args.ContainerID, ifName)
+		endpointID = plugin.nm.GetEndpointID(opt.args.ContainerID, ifName, opt.ifInfo.NICType)
 		*opt.infraSeen = true
 	} else {
 		ifName = "eth" + strconv.Itoa(opt.endpointIndex)
-		endpointID = plugin.nm.GetEndpointID(opt.args.ContainerID, ifName)
+		endpointID = plugin.nm.GetEndpointID(opt.args.ContainerID, ifName, opt.ifInfo.NICType)
```
QxBytes: what scenarios was this enabled in beforehand/currently?
Windows swift v1 (single nic)? y/n
Windows swift v2 (multi nic)? y/n
Linux swift v1 (single nic)? y/n
Linux swift v2 (multi nic)? y/n

cns/restserver/restserver.go
```diff
@@ -119,13 +119,14 @@ type EndpointInfo struct {
...
-	MacAddress    string      `json:",omitempty"`
-	NICType       cns.NICType
+	IPv4             []net.IPNet
+	IPv6             []net.IPNet `json:",omitempty"`
+	HnsEndpointID    string      `json:",omitempty"`
+	HnsNetworkID     string      `json:",omitempty"`
+	HostVethName     string      `json:",omitempty"`
+	MacAddress       string      `json:",omitempty"`
+	NICType          cns.NICType `json:",omitempty"`
+	NetworkNameSpace string      `json:",omitempty"`
```
QxBytes: To confirm, we are adding NetworkNameSpace to the cns and ip info because this information is required during linux endpoint deletion, right?

network/manager.go
```diff
@@ -732,8 +734,11 @@ func (nm *networkManager) GetNumberOfEndpoints(ifName string, networkId string)
 }
 
 // GetEndpointID returns a unique endpoint ID based on the CNI mode.
-func (nm *networkManager) GetEndpointID(containerID, ifName string) string {
+func (nm *networkManager) GetEndpointID(containerID, ifName string, nicType cns.NICType) string {
 	if nm.IsStatelessCNIMode() {
+		if nicType == cns.DelegatedVMNIC {
```
QxBytes: Can you explain the reasoning for adding this? What does it fix and what about other nic types (Frontend, etc.). How does it affect windows and linux (since I believe there are delegated vmnics in both)

network/manager.go
```diff
@@ -537,11 +538,12 @@ func (nm *networkManager) DeleteEndpointState(networkID string, epInfo *Endpoint
 		NetworkContainerID:       epInfo.NetworkContainerID, // we don't use this as long as AllowInboundFromHostToNC and AllowInboundFromNCToHost are false
 		NetNs:                    dummyGUID,                 // to trigger hnsv2, windows
 		NICType:                  epInfo.NICType,
+		NetworkNameSpace:         epInfo.NetNsPath,
 		IfName:                   epInfo.IfName, // TODO: For stateless cni linux populate IfName here to use in deletion in secondary endpoint client
```
QxBytes, line 545: Feel free to delete this comment if this pr fixes linux stateless cni

network/manager.go
behzad-mir: yes. So for stateless CNI in Windows this did not matter. For Linux the delete was actually executed fine since this was not multitenancy. It did remove the Veth interface instead of ip route since it will fall into bridge mode (default).

cns/restserver/restserver.go
behzad-mir: yes that's correct it's needed for fornetend nic removal form namespace.

tamilmani1989: wont namepsace be provided by containerd during cni delete?

cns/restserver/ipam.go
behzad-mir: This is just to make the statefile neater. There is no reason to add a field when it is empty.

cns/restserver/ipam.go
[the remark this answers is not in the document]
behzad-mir: I don't think this is actually correct as NICTYpe is not an enum and it is indeed string!

cns/restserver/ipam.go
behzad-mir: This is for old CNS logegr which does not  have a Info and instead it has printf.
All the logs in IPAM and other part of CNS code needs to be moved to the loggerV2 which has the proper logging levels. I used the oldlogger to have the PR be consistent with the rest of the file.

cni/network/network.go
behzad-mir: This has been enabled in all swiftV2 multi nic scenarios and also singularity.
In the new commit a new API defined GetEndpointIDByNicType to avoid touching the previous one and also it only make changes to the containerID in statelessCNI DElegatedNIC case.

cns/restserver/restserver.go
behzad-mir: I willcheck that. But currently in stateful case this has been supplied via statefile.

network/manager.go
behzad-mir: Addressed in the new commit

network/manager.go
behzad-mir: Addressed in the new commit

cns/restserver/restserver.go
behzad-mir: In the new commit I used the args from containerd to update the endpoint and no longer write the nspath to the state.

network/secondary_endpoint_client_linux.go
```diff
@@ -188,15 +189,30 @@ func (client *SecondaryEndpointClient) DeleteEndpoints(ep *endpoint) error {
 			logger.Error("Failed to exit netns with", zap.Error(newErrorSecondaryEndpointClient(err)))
 		}
 	}()
-	// TODO: For stateless cni linux, check if delegated vmnic type, and if so, delete using this *endpoint* struct's ifname
+	// For stateless cni linux, check if delegated vmnic type, and if so, delete using this *endpoint* struct's ifname
+	if ep.NICType == cns.DelegatedVMNIC {
```
QxBytes: I recall we had swiftv2 without stateless previously-- why was this not required then?

Additionally, why do we use cns.DelegatedVMNIC here when we use `cns.NodeNetworkInterfaceFrontendNIC` when selecting the endpoint in `endpoint_linux.go`. They both have the same value. It seems like secondary endpoint clients are created only when the nic type is of value FrontendNIC anyway so why this check?

if you're going by the comment it might be outdated

network/secondary_endpoint_client_linux.go
```diff
@@ -188,15 +189,30 @@ func (client *SecondaryEndpointClient) DeleteEndpoints(ep *endpoint) error {
...
+		if err := client.moveInterfaceToHostNetns(ep.IfName, vmns); err != nil {
+			logger.Error("Failed to move interface", zap.String("IfName", ep.IfName), zap.Error(newErrorSecondaryEndpointClient(err)))
+		}
+	}
+	// For Statefull cni linux, Use SecondaryInterfaces map to move all interfaces to host netns
+	// TODO: SecondaryInterfaces map should be retired and only IfName field and NICType should be used to determine the delgated NIC
 	for iface := range ep.SecondaryInterfaces {
-		if err := client.netlink.SetLinkNetNs(iface, uintptr(vmns)); err != nil {
+		if err := client.moveInterfaceToHostNetns(iface, vmns); err != nil {
```
QxBytes: the function mentions it moves the interface to the host ns but it seems like we still need to provide the ns id (an int)-- wondering if either 1) this should have a more general name (as it just moves an interface to any netns id) or 2) just keep it in-line w/o a new func?

network/secondary_endpoint_client_linux.go
```diff
@@ -188,15 +189,30 @@ func (client *SecondaryEndpointClient) DeleteEndpoints(ep *endpoint) error {
...
 	}()
-	// TODO: For stateless cni linux, check if delegated vmnic type, and if so, delete using this *endpoint* struct's ifname
+	// For stateless cni linux, check if delegated vmnic type, and if so, delete using this *endpoint* struct's ifname
+	if ep.NICType == cns.DelegatedVMNIC {
+		if err := client.moveInterfaceToHostNetns(ep.IfName, vmns); err != nil {
+			logger.Error("Failed to move interface", zap.String("IfName", ep.IfName), zap.Error(newErrorSecondaryEndpointClient(err)))
+		}
+	}
+	// For Statefull cni linux, Use SecondaryInterfaces map to move all interfaces to host netns
+	// TODO: SecondaryInterfaces map should be retired and only IfName field and NICType should be used to determine the delgated NIC
```
QxBytes: nit: spelling stateful and delegated

network/manager.go
```diff
@@ -116,6 +116,7 @@ type NetworkManager interface {
 	UpdateEndpoint(networkID string, existingEpInfo *EndpointInfo, targetEpInfo *EndpointInfo) error
 	GetNumberOfEndpoints(ifName string, networkID string) int
 	GetEndpointID(containerID, ifName string) string
+	GetEndpointIDByNicType(containerID, ifName string, nicType cns.NICType) string
```
QxBytes, line 120: nit: GetEndpointIDByNICType as nic is an acronym

cni/network/invoker_cns.go
```diff
@@ -301,6 +301,11 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 		PodInterfaceID:      GetEndpointID(args),
 		InfraContainerID:    args.ContainerID,
 	}
+	if options != nil {
+		if v, ok := options["asyncDeleteFileID"].(string); ok && v != "" {
+			ipConfigs.PodInterfaceID = v
```
QxBytes: why does this value need to differ between stateless and stateful? Is it for the async deletion path? what happens when we don't have this?

cni/network/network.go
```diff
@@ -1070,31 +1069,34 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
 				logger.Info("Endpoint Not found", zap.String("containerID", args.ContainerID), zap.Error(err))
 				return nil
+			default:
+				logger.Error("Get Endpoint State API returned error", zap.String("containerID", args.ContainerID), zap.Error(err))
+				return plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
+			}
+		} else {
+			for i, epInfo := range epInfos {
+				logger.Info("Found endpoint to delete", zap.String("IfName", epInfo.IfName), zap.String("EndpointID", epInfo.EndpointID), zap.Any("NICType", epInfo.NICType))
+				epInfos[i].NetNsPath = args.Netns // in case DelegatedNIC need to be moved to host namespace
```
QxBytes: if this is used in secondary endpoint client (I believe via epInfo.NetNsPath --> ep.NetworkNameSpace ) , could you add a comment mentioning that's how it is used?

QxBytes: would this modification after we get the state be better suited for GetEndpointState (similar to how we pass in networkID and then populate it in each epInfo?)?

QxBytes (commented):
preliminary suggestions-- could you also add how you tested and what happens without each of these changes?

cni/network/network.go
```diff
@@ -1078,6 +1078,13 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 			case errors.Is(err, network.ErrConnectionFailure):
 				logger.Error("Failed to connect to CNS", zap.Error(err))
 				logger.Info("Endpoint will be deleted from state file asynchronously", zap.String("containerID", args.ContainerID))
+				// In SwiftV2 Linux stateless CNI mode, if the plugin cannot connect to CNS,
+				// we asynchronously remove the secondary (delegated) interface from the pod’s network namespace.
+				// This is necessary because leaving the delegated NIC in the pod netns can cause the kernel to block rtnetlink operations.
+				// When that happens, kubelet and containerd hang during sandbox creation or teardown.
+				// The delegated NIC (SR-IOV VF) used by SwiftV2 for multitenant pods remains tied to the pod namespace,
+				// triggering hot-unplug/re-register events and leaving the node in an unhealthy state.
+				plugin.nm.RemoveSecondaryEndpointFromPodNetNS(args.IfName, args.Netns)
```
QxBytes: can you explain why we cannot delete using the normal delete flow and instead call this new method? What info is missing when we cannot retrieve data from CNS? If we can delete everything on CNI side without the CNS (which seems to be the case here), why have this new method?

cns/restserver/restserver.go
```diff
@@ -132,7 +132,7 @@ type IPInfo struct {
 	HnsNetworkID  string      `json:",omitempty"`
 	HostVethName  string      `json:",omitempty"`
 	MacAddress    string      `json:",omitempty"`
-	NICType       cns.NICType `json:",omitempty"`
+	NICType       cns.NICType
```
QxBytes: what is the reason for this change?

network/manager.go
```diff
@@ -860,3 +861,16 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
 
 	return ifNametoIPInfoMap
 }
+
+// RemoveSecondaryEndpointFromPodNetNS removes the secondary endpoint from the pod netns
+func (nm *networkManager) RemoveSecondaryEndpointFromPodNetNS(ifName string, netns string) error {
```
QxBytes: This seems really odd to me-- manager.go is meant to deal with general state, endpoint creation/deletion that can apply to all endpoint/nic type scenarios, but secondary endpoints (add/remove) are an implementation detail that applies to only one of the various scenarios. This logic should be at the client level (like changes should be in secondary_endpoint_client_linux only and it should decide the how to handle it based on the parameters passed in). This applies to all the `RemoveSecondaryEndpointFromPodNetNS` and `removeSecondaryEndpointFromPodNetNSImpl`

network/secondary_endpoint_client_linux.go
```diff
@@ -190,15 +192,15 @@ func (client *SecondaryEndpointClient) DeleteEndpoints(ep *endpoint) error {
 		}
 	}()
 	// For stateless cni linux, check if delegated vmnic type, and if so, delete using this *endpoint* struct's ifname
-	if ep.NICType == cns.DelegatedVMNIC {
-		if err := client.moveInterfaceToHostNetns(ep.IfName, vmns); err != nil {
+	if ep.NICType == cns.NodeNetworkInterfaceFrontendNIC {
```
QxBytes: reason for this particular nic type rather than delegated?

network/secondary_endpoint_client_linux.go
```diff
@@ -190,15 +192,15 @@ func (client *SecondaryEndpointClient) DeleteEndpoints(ep *endpoint) error {
...
+		if err := client.MoveInterfaceToHostNetns(ep.IfName, vmns); err != nil {
```
QxBytes: do we need to capitalize (export) this? It does not seem to be used outside of this file

network/secondary_endpoint_client_linux.go
```diff
@@ -209,10 +211,74 @@ func (client *SecondaryEndpointClient) DeleteEndpoints(ep *endpoint) error {
...
 	logger.Info("Moving interface to host netns", zap.String("IfName", ifName))
 	if err := client.netlink.SetLinkNetNs(ifName, uintptr(vmns)); err != nil {
 		return newErrorSecondaryEndpointClient(err)
 	}
 	return nil
 }
+
+// removeInterfacesfromNetnsPath finds and removes all interfaces from the specified netns path except the infra and non-eth interfaces.
+func (client *SecondaryEndpointClient) RemoveInterfacesfromNetnsPath(infraInterfaceName string, netnspath string) error {
+	// Get VM namespace
```
QxBytes, line 231: can we make a helper function for this namespace changing? Or if one already exists use that?

network/secondary_endpoint_client_linux.go
```diff
@@ -209,10 +211,74 @@ func (client *SecondaryEndpointClient) DeleteEndpoints(ep *endpoint) error {
...
+		}
+	}
+
+	return nil
+}
+
+// ListIfNamesInNetns returns all interface names in the given netns path.
+func listIfNamesInNetns(ns NamespaceInterface) ([]string, error) {
+	// Use the existing netlink interface to list links
+	links, err := vishnetlink.LinkList()
```
QxBytes: maybe this can be mocked? ex: in your test we have a mock vishnetlink with that one method you're using LinkList and we return something that satisfies the vishnetlink.Link interface

network/secondary_endpoint_client_linux.go
```diff
@@ -209,10 +211,74 @@ func (client *SecondaryEndpointClient) DeleteEndpoints(ep *endpoint) error {
...
+		}
+	}()
+	// For stateless cni linux, check if delegated vmnic type, and if so, delete using this *endpoint* struct's ifname
+	ifnames, err := listIfNamesInNetns(ns)
+	for _, iface := range ifnames {
+		// skip the infra interface as well as non-eth interfaces
+		if iface == infraInterfaceName || !strings.Contains(iface, "eth") {
+			continue
+		}
+		if err := client.MoveInterfaceToHostNetns(iface, vmns); err != nil {
```
QxBytes: hmm so why do we leave the infra eth0 in the pod namespace? Does the eth0 (infra) get cleaned up later on in the delete call?

QxBytes: Is it like during async delete, we only have 1 endpoint info received (since can't contact CNS for others) whereas if CNS were up, we'd get 1 endpoint info for each (1 for infra, 1 for each delegated) and so we'd be able to rely on the existing logic?

QxBytes (commented):
I noticed that cni/network/* has some issues with tests? Also please respond to the previous comments I made.

network/secondary_endpoint_client_linux.go
behzad-mir: That's correct. Eth0 does not suffer from this issue similar to secondary interface and once the NETNS get removed i is working fine. For delgated interface like eth1 we need to find it and moving it back to host

network/secondary_endpoint_client_linux.go
behzad-mir: I already created a helper function to remove the interface form the netnsPath

network/secondary_endpoint_client_linux.go
behzad-mir: In another comment you mentioned to use this newer version since the other one is absolute. They both resolve in forntend

cns/restserver/restserver.go
behzad-mir: I'm removing this change to have this whole PR only affcting CNI and network package. This is the original.

cni/network/network.go
behzad-mir: In normal Stateless CNI flow we need the delegated NIC Name which will be missing in this case. Since Stateless CNI cannot retrieve the state it cannot go through normal delete flow and we need this special case function

network/manager.go
behzad-mir: That's correct, but all of the call to platform related functions are called from manager.go. I did not see a direct call to Implementation from CNI package.

cni/network/network.go
behzad-mir: I did not want to change the function for getEndpointState(). The comment has been added.

network/secondary_endpoint_client_linux.go
behzad-mir: NodeNetworkInterfaceFrontendNIC has been used instead in new commit to be consistenet with CNS contract

network/secondary_endpoint_client_linux.go
behzad-mir: I moved it to a different function since this prosess is being called multiple times in this files

network/secondary_endpoint_client_linux.go
behzad-mir: addressed in the new commit

network/secondary_endpoint_client_linux.go
behzad-mir: I tired to mock this but still even mocking this is not enough to create a UT for the new releaseSecodnary function.

network/secondary_endpoint_client_linux.go
QxBytes: Can you change the name and comments for the function to like moveInterfaceToNetns? The int passed in is not necessarily the vm's namespace id. Sure it's called multiple times but we still need to handle the error-- you're replacing a function call with another function call that seems to do the exact same thing?

network/secondary_endpoint_client_linux.go
QxBytes: Could you address the first two questions?:
I recall we had swiftv2 without stateless previously-- why was this not required then?
It seems like secondary endpoint clients are created only when the nic type is of value FrontendNIC anyway so why this check?

cni/network/network.go
QxBytes: Can you mention it's used in the secondary endpoint client-- unless it's also being used in windows? Also if it does not change the logic, can you populate the epinfos right after you call GetEndpointState

cni/network/network.go
QxBytes: It seems like you can delete the secondary nics even without the delegated NIC name though, based on this function `RemoveSecondaryEndpointFromPodNetNS` -- why can't this be the normal flow?

network/secondary_endpoint_client_linux.go
QxBytes: I see that RemoveInterfacesfromNetnsPath is a helper, but noticed that the code here https://github.com/Azure/azure-container-networking/pull/3967/files#diff-a2a6265dc83a2c15c13849f9f3f4b1f2758e06f0973093002e0077dd8f20d4efR155 also opens a network namespace-- this is what I am referring to

cni/network/network.go
```diff
@@ -1070,31 +1069,45 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-			return plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
 		}
 	} else {
 		epInfos = plugin.nm.GetEndpointInfosFromContainerID(args.ContainerID)
 	}
 
-	// for when the endpoint is not created, but the ips are already allocated (only works if single network, single infra)
-	// this block is not applied to stateless CNI
+	// for Stateful CNI when the endpoint is not created, but the ips are already allocated (only works if single network, single infra)
+	// this block is applied to stateless CNI only if there was a connection failure in previous block and asynchronous delete by CNS will remover the endpoint from state file
```
QxBytes: <outdated>

QxBytes: can you clarify the comment (1109-1110) to:
the following block applies to:
- stateful when ...
- stateless when ...

cni/network/network.go
```diff
@@ -1069,32 +1068,45 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+				logger.Error("Failed to connect to CNS", zap.Error(err))
+				logger.Info("Endpoint will be deleted from state file asynchronously", zap.String("containerID", args.ContainerID))
+				// In SwiftV2 Linux stateless CNI mode, if the plugin cannot connect to CNS,
+				// we asynchronously remove the secondary (delegated) interface from the pod’s network namespace in the absence of the endpoint state.
+				// This is necessary because leaving the delegated NIC in the pod netns can cause the kernel to block rtnetlink operations.
+				// When that happens, kubelet and containerd hang during sandbox creation or teardown.
+				// The delegated NIC (SR-IOV VF) used by SwiftV2 for multitenant pods remains tied to the pod namespace,
+				// triggering hot-unplug/re-register events and leaving the node in an unhealthy state.
+				// This workaround mitigates the issue by removing the secondary NIC from the pod netns when CNS is unreachable during DEL to provide the endpoint state.
+				if err = plugin.nm.RemoveSecondaryEndpointFromPodNetNS(args.IfName, args.Netns); err != nil {
```
tamilmani1989: it should still create file for cns to release ip allocated for pod

tamilmani1989 (changes requested):
@behzad-mir lets discuss offline on these changes. i feel it can be done in another way

cni/network/network.go
behzad-mir: Addressed in the new commit

network/secondary_endpoint_client_linux.go
QxBytes: Previously in swiftv2 with stateful, the if statement above was not needed because we had the `for iface := range ep.SecondaryInterfaces {` block afterward, which took secondary interfaces from the endpoint and moved them. This is no longer the format we are using-- rather, in stateless we rely on this function being called for each secondary interface as well, and not being called for the infra (infra would be called with transparent endpoint client).

cni/network/network.go
```diff
@@ -1065,36 +1064,11 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-			if errors.Is(err, network.ErrEndpointStateNotFound) {
-				logger.Info("Endpoint Not found", zap.String("containerID", args.ContainerID), zap.Error(err))
-				return nil
-			}
-			logger.Error("Get Endpoint State API returned error", zap.String("containerID", args.ContainerID), zap.Error(err))
-			return plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
-		}
-	} else {
-		epInfos = plugin.nm.GetEndpointInfosFromContainerID(args.ContainerID)
+	epInfos, err := plugin.nm.GetEndpoint(networkID, args)
```
QxBytes: Can you rename this-- it currently isn't getting an endpoint object-- it's getting a slice of endpoint infos. Also please add a comment to that function

network/manager.go
```diff
@@ -847,3 +867,54 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
 }
+
+func (nm *networkManager) GetEndpoint(networkID string, args *cniSkel.CmdArgs) ([]*EndpointInfo, error) {
+	if nm.IsStatelessCNIMode() {
+		logger.Info("calling cns getEndpoint API")
+		epInfos, err := nm.GetEndpointState(networkID, args.ContainerID, args.Netns)
+		if err != nil {
+			switch {
+			case errors.Is(err, ErrConnectionFailure):
+				logger.Error("Failed to connect to CNS", zap.Error(err))
```
QxBytes: Can we combine these logs-- they are related

network/manager.go
```diff
@@ -847,3 +867,54 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
+				// In SwiftV2 Linux stateless CNI mode, if the plugin cannot connect to CNS,
+				// we still have to remove the secondary (delegated) interface from the pod’s network namespace in the absence of the endpoint state.
+				// This is necessary because leaving the delegated NIC in the pod netns can cause the kernel to block rtnetlink operations.
+				// When that happens, kubelet and containerd hang during sandbox creation or teardown.
+				// The delegated NIC (SR-IOV VF) used by SwiftV2 for multitenant pods remains tied to the pod namespace,
+				// triggering hot-unplug/re-register events and leaving the node in an unhealthy state.
+				// This workaround mitigates the issue by generating a minimal endpointInfo via containerd args and netlink APIs that can be then passed to DeleteEndpoint API.
+				epInfos, err = nm.generateEndpointLocally(args)
+				if err != nil {
+					logger.Error("Failed to fetch secondary endpoint from pod netns", zap.String("netns", args.Netns), zap.Error(err))
```
QxBytes, line 909: Would like to avoid logging and returning the error here since we should already log the error in the main Delete call

network/manager.go
```diff
@@ -847,3 +867,54 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
+					return nil, fmt.Errorf("failed to fetch secondary interfaces: %w", err)
```
QxBytes: for consistency can we stick with errors.Wrap or errors.Wrapf-- this applies elsewhere as well

network/manager.go
```diff
@@ -847,3 +867,54 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
+	}
+	return nm.GetEndpointInfosFromContainerID(args.ContainerID), nil
+}
+
+// generateEndpointLocally fetches the endpoint information using containerd args and netlink APIs
+func (nm *networkManager) generateEndpointLocally(args *cniSkel.CmdArgs) ([]*EndpointInfo, error) {
+	ep := &endpoint{
+		Id:               args.ContainerID,
+		NetworkNameSpace: args.Netns,
+		IfName:           args.IfName, // TODO: For stateless cni linux populate IfName here to use in deletion in secondary endpoint client
```
QxBytes: what is this todo for? it's also added elsewhere

cni/network/network.go
```diff
@@ -1132,15 +1106,25 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
 		if !nwCfg.MultiTenancy && (epInfo.NICType == cns.InfraNIC || epInfo.NICType == "") {
-			// Delegated/secondary nic ips are statically allocated so we don't need to release
-			// Call into IPAM plugin to release the endpoint's addresses.
-			for i := range epInfo.IPAddresses {
-				logger.Info("Release ip", zap.String("ip", epInfo.IPAddresses[i].IP.String()))
-				telemetryClient.SendEvent(fmt.Sprintf("Release ip: %s container id: %s endpoint id: %s", epInfo.IPAddresses[i].IP.String(), args.ContainerID, epInfo.EndpointID))
-				err = plugin.ipamInvoker.Delete(&epInfo.IPAddresses[i], nwCfg, args, nwInfo.Options)
-				if err != nil {
-					return plugin.RetriableError(fmt.Errorf("failed to release address: %w", err))
+			// This is an special case for stateless CNI when Asychronous DEL to CNS will take place
```
QxBytes: please address the linter

network/manager.go
```diff
@@ -847,3 +867,54 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
+				// triggering hot-unplug/re-register events and leaving the node in an unhealthy state.
+				// This workaround mitigates the issue by generating a minimal endpointInfo via containerd args and netlink APIs that can be then passed to DeleteEndpoint API.
+				epInfos, err = nm.generateEndpointLocally(args)
+				if err != nil {
+					logger.Error("Failed to fetch secondary endpoint from pod netns", zap.String("netns", args.Netns), zap.Error(err))
+					return nil, fmt.Errorf("failed to fetch secondary interfaces: %w", err)
+				}
+			case errors.Is(err, ErrEndpointStateNotFound):
+				logger.Info("Endpoint Not found", zap.String("containerID", args.ContainerID), zap.Error(err))
+				return nil, nil
```
QxBytes: by moving the code into this function, the behavior changes-- like if we return nil here, and a nil error, the `epInfos` var from GetEndpoint is now nil and we continue-- while len(nil) I think is still zero, would like to return at least an empty slice if we have no error. Also, previously I believe we would immediately return from Delete, but now we continue to the len(epInfos) == 0 block since err is nil, leading to extra extraneous log lines and calling plugin.ipamInvoker.Delete

network/manager.go
```diff
@@ -847,3 +867,54 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
 
 	return ifNametoIPInfoMap
 }
+
+func (nm *networkManager) GetEndpoint(networkID string, args *cniSkel.CmdArgs) ([]*EndpointInfo, error) {
+	if nm.IsStatelessCNIMode() {
+		logger.Info("calling cns getEndpoint API")
```
QxBytes: capitalize log messages

network/manager.go
```diff
@@ -847,3 +867,54 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
+}
+
+// generateEndpointLocally fetches the endpoint information using containerd args and netlink APIs
+func (nm *networkManager) generateEndpointLocally(args *cniSkel.CmdArgs) ([]*EndpointInfo, error) {
+	ep := &endpoint{
+		Id:               args.ContainerID,
+		NetworkNameSpace: args.Netns,
+		IfName:           args.IfName, // TODO: For stateless cni linux populate IfName here to use in deletion in secondary endpoint client
+	}
+	epInfo, err := nm.getEndpointInfoByIfNameImpl(ep)
```
QxBytes: did you confirm what happens on windows? If the error is nil, I would prefer returning an empty slice over "nil" in the windows implementation. I believe in windows this leads to GetEndpoint returning empty/nil, we going into the len(epInfos)==0 block, calling ipamInvoker.Delete-- please confirm this is intended

Also can we make the function/method names more consistent? If it returns a slice of endpoint infos, we getEndpointInfo**s** (plural)

The method name is also a bit confusing since we're using more than just the endpoint's ifname

network/endpoint_linux.go
```diff
@@ -547,3 +547,29 @@ func getDefaultGateway(routes []RouteInfo) net.IP {
...
+func (nm *networkManager) getEndpointInfoByIfNameImpl(ep *endpoint) ([]*EndpointInfo, error) {
+	epInfo := &EndpointInfo{
+		EndpointID: ep.Id,
+		NetNsPath:  ep.NetworkNameSpace,
+		NICType:    cns.InfraNIC,
+		IfName:     ep.IfName, // TODO: For stateless cni linux populate IfName here to use in deletion in secondary endpoint client
+	}
+	ret := []*EndpointInfo{}
+	ret = append(ret, epInfo)
+	logger.Info("Fetching Secondary Endpoint from", zap.String("NetworkNameSpace: ", ep.NetworkNameSpace))
```
QxBytes: You don't need to put the colon in "NetworkNameSpace: "-- it should automatically format

network/secondary_endpoint_client_linux.go
```diff
@@ -145,61 +144,113 @@ func (client *SecondaryEndpointClient) ConfigureContainerInterfacesAndRoutes(epI
...
 	// Open the network namespace.
-	logger.Info("Opening netns", zap.Any("NetNsPath", ep.NetworkNameSpace))
-	ns, err := client.nsClient.OpenNamespace(ep.NetworkNameSpace)
+	logger.Info("Opening netns", zap.Any("NetNsPath", nsName))
+	ns, err := client.nsClient.OpenNamespace(nsName)
 	if err != nil {
 		if strings.Contains(err.Error(), errFileNotExist.Error()) {
-			// clear SecondaryInterfaces map since network namespace doesn't exist anymore
-			ep.SecondaryInterfaces = make(map[string]*InterfaceInfo)
+			// clear SecondaryInterfaces map since network namespace doesn't exist anymore, Specific to legacy code in stateful cni
```
QxBytes: could you explain these clearings of the secondary interface map? This function should only deal with moving into or out of a network namespace based on its naming-- this is a hidden side effect that modifies the endpoint client and should be moved out of this function if possible.

network/errors.go
```diff
@@ -3,9 +3,11 @@ package network
...
-	errSubnetV6NotFound        = errors.New("Couldn't find ipv6 subnet in network info")                // nolint
-	errV6SnatRuleNotSet        = errors.New("ipv6 snat rule not set. Might be VM ipv6 address missing") // nolint
-	ErrEndpointStateNotFound   = errors.New("endpoint state could not be found in the statefile")
-	ErrConnectionFailure       = errors.New("couldn't connect to CNS")
-	ErrGetEndpointStateFailure = errors.New("failure to obtain the endpoint state")
+	errSubnetV6NotFound         = errors.New("Couldn't find ipv6 subnet in network info")                // nolint
+	errV6SnatRuleNotSet         = errors.New("ipv6 snat rule not set. Might be VM ipv6 address missing") // nolint
+	ErrEndpointStateNotFound    = errors.New("endpoint state could not be found in the statefile")
+	ErrConnectionFailure        = errors.New("couldn't connect to CNS")
+	ErrEndpointRemovalFailure   = errors.New("Failed to remove endpoint")
```
QxBytes: I know some of these are capitalized already but new error messages shouldn't be capitalized: https://go.dev/wiki/CodeReviewComments#error-strings

QxBytes (changes requested):
As I'm reading this it seems like a lot of the code is geared around accessing FetchInterfacesFromNetnsPath in `secondary_endpoint_client_linux.go`, requiring us to drill down through cni/network/network.go > network/manager.go > network/endpoint_linux.go > secondary_endpoint_client_linux.go, adding methods to interfaces which are only specific to swiftv2/secondary endpoint client.

What is it from secondary endpoint client that we actually need? At GetEndpoint() > nm.generateEndpointLocally, there is no guarantee we are even in swiftv2 right? Why can't we just make a helper function (doesn't even need to be part of a client) where generateEndpointLocally is called that does the following:
1) Enters the pod's netns
2) Searches for any swiftv2/secondary interfaces if they exist
3) Creates endpoint infos and returns them if found
Instead of going through the trouble of creating a secondary endpoint client etc.

cni/network/network.go
```diff
@@ -1132,15 +1106,25 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+			// At this point the endpoint is already deleted in previous block and CNS will release the IP whenever it is up
+			if epInfo.IPAddresses == nil && plugin.nm.IsStatelessCNIMode() {
+				logger.Warn("Release ip Asynchronously by CNS",
```
tamilmani1989: why releaseip only for stateless cni and not for stateful cni case? current code calling release for both cases

network/manager.go
```diff
@@ -116,11 +118,13 @@ type NetworkManager interface {
...
 	GetNumberOfEndpoints(ifName string, networkID string) int
 	GetEndpointID(containerID, ifName string) string
+	GetEndpointIDByNicType(containerID, ifName string, nicType cns.NICType) string
 	IsStatelessCNIMode() bool
 	SaveState(eps []*endpoint) error
 	DeleteState(epInfos []*EndpointInfo) error
+	GetEndpoint(networkID string, args *cniSkel.CmdArgs) ([]*EndpointInfo, error)
 	GetEndpointInfosFromContainerID(containerID string) []*EndpointInfo
-	GetEndpointState(networkID, containerID string) ([]*EndpointInfo, error)
+	GetEndpointState(networkID, containerID, netns string) ([]*EndpointInfo, error)
```
tamilmani1989: why do we need 2 apis? can we merge to 1 api, otherwise this is confusing and not sure about purpose of 2 apis

network/manager.go
```diff
@@ -745,6 +751,16 @@ func (nm *networkManager) GetEndpointID(containerID, ifName string) string {
 	return containerID + "-" + ifName
 }
 
+// GetEndpointIDByNicType returns a unique endpoint ID based on the CNI mode and NIC type.
+func (nm *networkManager) GetEndpointIDByNicType(containerID, ifName string, nicType cns.NICType) string {
+	// For stateless CNI, secondary NICs use containerID-ifName as endpointID.
+	if nm.IsStatelessCNIMode() && nicType != cns.InfraNIC {
```
tamilmani1989, line 756: for stateful cni, this is not an issue? what's the impact if we remove statelesscnimode check here?

network/manager.go
```diff
@@ -847,3 +867,54 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
+				logger.Error("Get Endpoint State API returned error", zap.String("containerID", args.ContainerID), zap.Error(err))
+				return nil, ErrEndpointRetrievalFailure
+			}
+		}
+		for _, epInfo := range epInfos {
+			logger.Info("Found endpoint to delete", zap.String("IfName", epInfo.IfName), zap.String("EndpointID", epInfo.EndpointID), zap.Any("NICType", epInfo.NICType))
+		}
+		return epInfos, nil
+	}
+	return nm.GetEndpointInfosFromContainerID(args.ContainerID), nil
```
tamilmani1989, line 926: if this api is called only for stateful cni, can we keep in else to avoid any error in future

network/manager.go
behzad-mir: One of the reasons I made this change is to go through plugin.ipamInvoker.Delete where the FSnotify and Asych delete have been initialized properly. I wanted the Asynch delete get handled within that function and d not have duplication. I don't think we need to return an empty endpointInfoSlice since the previous code in GetEndpointState we were still returing nil in case of NO endpointFOund.

network/manager.go
behzad-mir: That has been tested, we wante to Asych delete get handled by  ipamInvoker.Delete

cni/network/network.go
behzad-mir: When we generate the ednpoint locally for SwiftV2 stateless when CNS is not up, then we will get to this block at the end to release the IP. The datapath already has been cleand up and we just do a Asynch delete via IPAMInvoker.Delete.

network/manager.go
behzad-mir: We previous;y doscussed that, and you suggested to have two APIs.

behzad-mir: GetEndpointInfos will be a genrall API for both Statefull and Stateless and then GetEndpointState is just for stateless CNI

network/manager.go
behzad-mir: I wanted to make sure we only add - ifname for stateless. Stateful already has this.

network/manager.go
```diff
@@ -867,3 +882,56 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
 
 	return ifNametoIPInfoMap
 }
+
+// GetEndpointInfos gets all endpoint infos associated with a container id
```
QxBytes: this comment will be confused with the other method `GetEndpointInfosFromContainerID`. We pass in more than the container id here (and it isn't obvious that the container id is even passed in). would prefer  we pass in each of the args that we use separately as parameters since we only use ContainerID, Netns, and IfName. comment should also explain what each of the parameters is used for and how it differs from `GetEndpointInfosFromContainerID`.

network/manager.go
```diff
@@ -867,3 +882,56 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
+func (nm *networkManager) GetEndpointInfos(networkID string, args *cniSkel.CmdArgs, disableAsyncDelete bool) ([]*EndpointInfo, error) {
```
QxBytes, line 890: can we add tests for this

network/errors.go
QxBytes: it seems like you capitalized all these messages? they should _not_ be capitalized

network/manager.go
QxBytes: Previously we returned nil _as an error_ ,  but _we didn't return anything else_ (so there was no possibility of for example, indexing into a nil pointer). Now we return nil, and the caller of the function thinks everything has gone well and the returned values are valid, but now the slice is nil and if they try to access it will cause an error.

From my understanding, you want the function to continue after we return nil, nil right? This is because you want the async delete code to run, which occurs after the call to `GetEndpointInfos()`, right?

cni/network/network.go
```diff
@@ -1139,15 +1112,25 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-			// Call into IPAM plugin to release the endpoint's addresses.
-			for i := range epInfo.IPAddresses {
-				logger.Info("Release ip", zap.String("ip", epInfo.IPAddresses[i].IP.String()))
-				telemetryClient.SendEvent(fmt.Sprintf("Release ip: %s container id: %s endpoint id: %s", epInfo.IPAddresses[i].IP.String(), args.ContainerID, epInfo.EndpointID))
-				err = plugin.ipamInvoker.Delete(&epInfo.IPAddresses[i], nwCfg, args, nwInfo.Options)
-				if err != nil {
-					return plugin.RetriableError(fmt.Errorf("failed to release address: %w", err))
+			// This is an special case for stateless CNI when Asychronous DEL to CNS will take place for SwiftV2 after the endpoint is created locally
+			// At this point the endpoint is already deleted in previous block and CNS will release the IP whenever it is up
+			if epInfo.IPAddresses == nil && plugin.nm.IsStatelessCNIMode() {
```
QxBytes: what is the significance of epInfo.IPAddresses == nil here? When does this occur? I'm guessing it's when we try to generate the endpoint info locally when we need to async delete? If so, can you modify the comment to be "after the endpoint info is recreated locally"? Also include the fact that the ip addresses being nil means the ep infos were recreated locally due to a cns connection failure if that is the case.

cni/network/network.go
```diff
@@ -1139,15 +1112,25 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
+				logger.Warn("Release ip Asynchronously by CNS",
+					zap.String("containerID", args.ContainerID))
+				if err = plugin.ipamInvoker.Delete(nil, nwCfg, args, nwInfo.Options); err != nil {
+					return plugin.RetriableError(fmt.Errorf("failed to release address(no endpoint): %w", err))
+				}
+			} else {
+				for i := range epInfo.IPAddresses {
+					logger.Info("Release ip", zap.String("ip", epInfo.IPAddresses[i].IP.String()))
+					telemetryClient.SendEvent(fmt.Sprintf("Release ip: %s container id: %s endpoint id: %s", epInfo.IPAddresses[i].IP.String(), args.ContainerID, epInfo.EndpointID))
```
QxBytes: should we send a telemetry event if we go down the stateless path as well?

network/endpoint_linux.go
```diff
@@ -547,3 +548,29 @@ func getDefaultGateway(routes []RouteInfo) net.IP {
...
 	return epInfo, nil
 }
+
+// getEndpointInfoByIfNameImpl returns an array of EndpointInfo for the given endpoint based on the IfName(s) found in the network namespace.
+func (nm *networkManager) getEndpointInfoByIfNameImpl(ep *endpoint) ([]*EndpointInfo, error) {
+	epInfo := &EndpointInfo{
+		EndpointID: ep.Id,
+		NetNsPath:  ep.NetworkNameSpace,
+		NICType:    cns.InfraNIC,
+		IfName:     ep.IfName, // TODO: For stateless cni linux populate IfName here to use in deletion in secondary endpoint client
```
QxBytes: what does this TODO mean? Is this something you need to address before merging?

network/manager.go
```diff
@@ -867,3 +884,57 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
+		}
+		return epInfos, nil
+	}
+	// Stateful CNI mode
+	return nm.GetEndpointInfosFromContainerID(args.ContainerID), nil
+
+}
+
+// generateEndpointLocally fetches the endpoint information using containerd args and netlink APIs
+func (nm *networkManager) generateEndpointLocally(args *cniSkel.CmdArgs) ([]*EndpointInfo, error) {
```
QxBytes: Why are we putting the args into an endpoint struct, only for the fields in that endpoint struct to be put into an endpoint info? The ep struct just seems to be passed into `NewSecondaryEndpointClient` but isn't used-- (I do see it referenced in ExecuteInNS but please address my other comment about ExecuteInNS not having side effects). In windows the endpoint struct created isn't used either.

Why can't we just call nm.getEndpointInfoByIfName(container id, netns, ifname) directly from `GetEndpointInfos` ?

network/manager.go
```diff
@@ -456,7 +459,7 @@ func validateUpdateEndpointState(endpointID string, ifNameToIPInfoMap map[string
 // GetEndpointState will make a call to CNS GetEndpointState API in the stateless CNI mode to fetch the endpointInfo
 // TODO unit tests need to be added, WorkItem: 26606939
 // In stateless cni, container id is the endpoint id, so you can pass in either
-func (nm *networkManager) GetEndpointState(networkID, containerID string) ([]*EndpointInfo, error) {
+func (nm *networkManager) GetEndpointState(networkID, containerID, netns string) ([]*EndpointInfo, error) {
```
QxBytes, line 461: can you update the comment to include what the netns is for, and also remove the work item if you already completed the unit tests?

network/manager.go
```diff
@@ -827,7 +842,9 @@ func cnsEndpointInfotoCNIEpInfos(endpointInfo restserver.EndpointInfo, endpointI
 		epInfo.HNSNetworkID = ipInfo.HnsNetworkID
 		epInfo.MacAddress = net.HardwareAddr(ipInfo.MacAddress)
 		epInfo.NetworkContainerID = ipInfo.NetworkContainerID
-
+		if epInfo.NetNsPath == "" {
+			epInfo.NetNsPath = netns
```
QxBytes, line 845: can you explain this? add to the comment-- the "netns" variable is more like a defaultNetNs, which only populates if epInfo.NetNsPath is empty. Why do we add this (whereas we didn't need it before) and where is it used? Only place I can immediately see is in secondary endpoitn client if there is a cns connection failure and we perform async delete-- but generateEndpointLocally alrady takes the netns from args.Netns ?

behzad-mir: This is always needed for Stateless CNI whether there is Asynch delete or not. We are not string it in Statefile rn and only populating it from here.

network/manager.go
behzad-mir: It will cause a linter error, so I added a comment instead.

network/errors.go
behzad-mir: Misread your first comment and made the capitalized. It is corrected now.

network/secondary_endpoint_client_linux.go
behzad-mir: There is a comment for it that this is due to a the support for legacy code and resulted statefile in Statefull CNI.

network/manager.go
behzad-mir: This will be only shown in final error return from CNI and wanted to have actual one line log for this.

network/manager.go
behzad-mir: We need a complementary PR for UT in stateless CNI to mock statefile creation for CNS API to add and cover all the UTs for Stateless CNI.

network/endpoint_linux.go
behzad-mir: It is a leftover that I just removed.

network/manager.go
behzad-mir: Updated the comment

cni/network/network.go
behzad-mir: Yes, that is correct. I have updated the comment accordingly.

cni/network/network.go
behzad-mir: Added

network/manager.go
behzad-mir: Please refer to my other comment

network/manager.go
behzad-mir: The reason was for the reference to ep in ExecuteInNS as you  have mentioned. I needed that to support the legacy state if exists in the statefile. We can remove it at some point perhaps.

network/manager.go
behzad-mir: Yes, that is correct. Previously we also had nil endpointinfo in case of NOendpointFOund as I have mentioned above. Since we are hitting the next block of len=0 this will not cause an issue right now. But if you want to have safety in the code for future unwanted changes I can return empty structure as well

network/secondary_endpoint_client_linux.go
QxBytes: Talked about possibly removing this in ~1 month after talking with tamilmani to see if it still needs to be supported. The err file not exist would be checked from where ExecuteInNS is called and then if found, we would clear secondary interfaces. Even right now I don't think we reference SecondaryInterfaces after delete (since we delete the whole endpoint).

network/manager.go
QxBytes: other comment mentions this will be done in separate pr (mocking cnsClient)

network/manager.go
QxBytes: Discussed with behzad and we add this because previously, swiftv2 linux was not supported, and so we did not need to exec into the container namespace in linux to move the secondary nics into the vm namespace. Now, we are supporting this and the secondary endpoint client needs the netns to exec into it during deletion.

network/manager.go
behzad-mir: This has been resolved by removing the SecondaryInterface in the ep

QxBytes (dismissed):
During delete, if we go via async route+stateless, we always create a secondary endpoint client (even if it isn't swift v2 mode) and check for interfaces that are neither eth* nor the infra interface. Just a note if in the future in some other mode we create a new interface in the container namespace and it isn't related to secondaries/swiftv2, it must have a name like eth* or it might be mistakenly treated as a secondary interface during delete.

cni/network/network.go
```diff
@@ -1046,13 +1045,15 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
 	if nwInfo, err = plugin.nm.GetNetworkInfo(networkID); err != nil {
 		if !nwCfg.MultiTenancy {
-			logger.Error("Failed to query network",
-				zap.String("network", networkID),
-				zap.Error(err))
 			// Log the error if the network is not found.
 			// if cni hits this, mostly state file would be missing and it can be reboot scenario where
 			// container runtime tries to delete and create pods which existed before reboot.
-			// this condition will not apply to stateless CNI since the network struct will be crated on each call
+			// this error will not apply to stateless CNI since the network struct will be crated on Delete calls
```
tamilmani1989, line 1051: will be or will not be crafted?

cni/network/network.go
```diff
@@ -1139,15 +1114,26 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
 		if !nwCfg.MultiTenancy && (epInfo.NICType == cns.InfraNIC || epInfo.NICType == "") {
-			// Delegated/secondary nic ips are statically allocated so we don't need to release
-			// Call into IPAM plugin to release the endpoint's addresses.
-			for i := range epInfo.IPAddresses {
-				logger.Info("Release ip", zap.String("ip", epInfo.IPAddresses[i].IP.String()))
-				telemetryClient.SendEvent(fmt.Sprintf("Release ip: %s container id: %s endpoint id: %s", epInfo.IPAddresses[i].IP.String(), args.ContainerID, epInfo.EndpointID))
-				err = plugin.ipamInvoker.Delete(&epInfo.IPAddresses[i], nwCfg, args, nwInfo.Options)
-				if err != nil {
-					return plugin.RetriableError(fmt.Errorf("failed to release address: %w", err))
+			// This is an special case for stateless CNI when Asynchronous DEL to CNS will take place for SwiftV2 after the endpointinfo is recreated locally
```
tamilmani1989: didnt understand this comment?

network/manager.go
```diff
@@ -867,3 +883,45 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
...
+				return emptyEpInfos, ErrEndpointRetrievalFailure
+			}
+		}
+		for _, epInfo := range epInfos {
+			logger.Info("Found endpoint to delete", zap.String("IfName", epInfo.IfName), zap.String("EndpointID", epInfo.EndpointID), zap.Any("NICType", epInfo.NICType))
+		}
+		return epInfos, nil
+	}
+	// Stateful CNI mode
+	return nm.GetEndpointInfosFromContainerID(args.ContainerID), nil
```
tamilmani1989, line 926: what empty return here returns?

cni/network/network.go
```diff
@@ -719,7 +718,7 @@ func (plugin *NetPlugin) createEpInfo(opt *createEpInfoOpt) (*network.EndpointIn
 		*opt.infraSeen = true
 	} else {
 		ifName = "eth" + strconv.Itoa(opt.endpointIndex)
```
behzad-mir, line 720: ADD a comment to give an example for the enrpointID format

cni/network/network.go
```diff
@@ -1046,13 +1045,15 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-			logger.Error("Failed to query network",
-				zap.String("network", networkID),
-				zap.Error(err))
 			// Log the error if the network is not found.
 			// if cni hits this, mostly state file would be missing and it can be reboot scenario where
 			// container runtime tries to delete and create pods which existed before reboot.
-			// this condition will not apply to stateless CNI since the network struct will be crated on each call
+			// this error will not apply to stateless CNI since the network struct will be crated on Delete calls
+			if !plugin.nm.IsStatelessCNIMode() {
+				logger.Info("Failed to query network",
```
behzad-mir: this should be stil Error

network/secondary_endpoint_client_linux.go
```diff
@@ -153,53 +155,97 @@ func (client *SecondaryEndpointClient) ConfigureContainerInterfacesAndRoutes(epI
...
+		ret := make([]string, 0, len(ifnames))
+		// Filter out infra interface and non-eth interfaces
+		for _, iface := range ifnames {
+			if iface == infraInterfaceName || !strings.HasPrefix(iface, "eth") {
+				continue
+			}
+			ret = append(ret, iface)
+		}
+
+		logger.Info("Found interfaces in netns that needs to be moved back to host", zap.Any("interfaces", ret), zap.Int("vmns", vmns))
```
behzad-mir: Just correct the logs to say that we found the eth interface

behzad-mir: done

cni/network/network.go
behzad-mir: done

cni/network/network.go
behzad-mir: Added in the appropriate function

cni/network/network.go
behzad-mir: The special case here has been removed since we are now retrieving IP addresses of interfaces from NEtns itself when CNS is down. As a result we are building the whole epinfo without the need for state when CNS is down. We can then call to CNS to release the IP and if it cannot it will do it asynchronously.

cni/network/network.go
behzad-mir: Will not.

network/manager.go
behzad-mir: We are returning empty error since the GetEndpointInfosFromContainerID will not generate any error at all

network/secondary_endpoint_linux_test.go
```diff
@@ -414,3 +394,229 @@ func TestSecondaryConfigureContainerInterfacesAndRoutes(t *testing.T) {
...
+	}
+
+	netnspath := "testns"
+	infraInterfaceName := "eth0"
+
+	result, err := client.FetchInterfacesFromNetnsPath(infraInterfaceName, netnspath)
+
+	require.NoError(t, err)
+	// Result will be empty in test environment since no actual interfaces exist
+	require.NotNil(t, result)
```
QxBytes, line 418: this doesn't check if the result is empty-- since this runs with the real implementation, it's behavior can change based on what machine it runs on. For instance, on my machine I have 2 interfaces that start with "eth". The "proper" way to fix this would be to mock vishnetlink but not sure how feasible that is right now. This comment applies to the other test cases too which all use the real vishnetlink implementation and thus can fail or pass based on which computer the test runs on.

cni/network/network.go
```diff
@@ -1139,9 +1114,9 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 			zap.String("endpointID", epInfo.EndpointID))
 		telemetryClient.SendEvent("Deleting endpoint: " + epInfo.EndpointID)
 
+		// Delegated/secondary nic ips are statically allocated so we don't need to release
+		// Call into IPAM plugin to release the endpoint's addresses.
```
QxBytes, line 1118: for reference why do we no longer need the special case when epInfo.IPAddresses is nil or empty that we previously had?

cni/network/network.go
```diff
@@ -1071,37 +1072,11 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-			if errors.Is(err, network.ErrEndpointStateNotFound) {
-				logger.Info("Endpoint Not found", zap.String("containerID", args.ContainerID), zap.Error(err))
-				return nil
-			}
-			logger.Error("Get Endpoint State API returned error", zap.String("containerID", args.ContainerID), zap.Error(err))
-			return plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
-		}
-	} else {
-		epInfos = plugin.nm.GetEndpointInfosFromContainerID(args.ContainerID)
+	epInfos, err := plugin.nm.GetEndpointInfos(networkID, args, nwCfg.DisableAsyncDelete)
```
QxBytes, line 1075: for reference, why do we no longer have that fsnotify.Add file case anymore? could we also add the reason to the pr description?

cni/network/network.go
```diff
@@ -719,7 +718,7 @@ func (plugin *NetPlugin) createEpInfo(opt *createEpInfoOpt) (*network.EndpointIn
 		*opt.infraSeen = true
 	} else {
 		ifName = "eth" + strconv.Itoa(opt.endpointIndex)
-		endpointID = plugin.nm.GetEndpointID(opt.args.ContainerID, ifName)
+		endpointID = plugin.nm.GetEndpointIDByNicType(opt.args.ContainerID, ifName, opt.ifInfo.NICType)
```
QxBytes, line 721: Could you confirm this is accurate and if so add this to the PR descrption or in a comment for this block (and if you add a comment add a comment to the GetEndpointID, GetEndpointIDByNicType methods to modify the comment if the behavior changes)-- hopefully it makes things more clear

Image

network/manager.go
```diff
@@ -827,13 +841,15 @@ func cnsEndpointInfotoCNIEpInfos(endpointInfo restserver.EndpointInfo, endpointI
...
+		if epInfo.NetNsPath == "" {
+			epInfo.NetNsPath = netns
+		}
 		ret = append(ret, epInfo)
 	}
 	return ret
 }
 
-// gets all endpoint infos associated with a container id and populates the network id field
+// gets all endpoint infos associated with a container id and populates the network id field in Statefull CNI mode
```
QxBytes, line 852: nit: stateful

network/manager.go
```diff
@@ -867,3 +883,45 @@ func generateCNSIPInfoMap(eps []*endpoint) map[string]*restserver.IPInfo {
 
 	return ifNametoIPInfoMap
 }
+
+// GetEndpointInfos gets all endpoint infos associated with a container id and networkID
+// In stateless CNI mode, it calls CNS GetEndpointState API to get the endpoint infos or genreate them locally if CNS is unreachable in SwiftV2 mode and AsyncDelete is enabled
```
QxBytes, line 888: nit: generate

network/manager.go
```diff
@@ -769,10 +783,10 @@ func (nm *networkManager) DeleteState(epInfos []*EndpointInfo) error {
 	logger.Info("Deleting state")
 
 	// For AKS stateless cni, plugin.ipamInvoker.Delete takes care of removing the state in the main Delete function.
-	// For swiftv2 stateless cni, this call will delete the endpoint state from CNS.
-	if nm.IsStatelessCNIMode() {
+	// For swiftv2 stateless cni in Windows, this call will delete the endpoint state from CNS.
+	if nm.IsStatelessCNIMode() && platform.GetOSInfo() == "windows" {
```
QxBytes, line 787: previously this was also run on linux-- wondering why we did we not run into issues then? I thought we had at least a few scenarios with stateless for linux, unless prior to this _no_ scenario had stateless + linux?

network/manager.go
```diff
@@ -746,6 +750,16 @@ func (nm *networkManager) GetEndpointID(containerID, ifName string) string {
 	return containerID + "-" + ifName
 }
 
+// GetEndpointIDByNicType returns a unique endpoint ID based on the CNI mode and NIC type.
+func (nm *networkManager) GetEndpointIDByNicType(containerID, ifName string, nicType cns.NICType) string {
+	// For stateless CNI, secondary NICs use #containerID-#ifName as endpointID such as 12345678-eth1
+	if nm.IsStatelessCNIMode() && nicType != cns.InfraNIC {
+		return containerID + "-" + ifName
+	}
+	// For InfraNIC, use GetEndpointID() logic.
```
QxBytes, line 759: more than the infra nic use this logic right? Even secondaries use this logic if we are not in stateless mode

also for the comment it's based on the container id, ifname, and nic type (not just the mode and nic type)
maybe mention in what cases it returns containerID+"-"+ifName and what cases we use GetEndpointID-- it's a bit difficult to see at first glance

network/manager.go
```diff
@@ -456,7 +457,8 @@ func validateUpdateEndpointState(endpointID string, ifNameToIPInfoMap map[string
 // GetEndpointState will make a call to CNS GetEndpointState API in the stateless CNI mode to fetch the endpointInfo
 // TODO unit tests need to be added, WorkItem: 26606939
 // In stateless cni, container id is the endpoint id, so you can pass in either
```
QxBytes, line 459: for this comment, is it still true? container id is not always equal to endpoint id once we added that branch on the nic type

network/manager.go
behzad-mir: This is a new API added recently for Windows SwiftV2 scenario in which delete logic is different. Basically in ACI Delete is not ATOMIC and CNS/DNC IP release will be removed separately from CNI dataplane del Flow. As a result, this API was added to delete the endpoint state from CNS statefile. For Linux SwiftV2 scenario this is not needed and if we do not add this platform check it will fail in Asynch Del case and results in a failed deleted repeated Containerd calls.

network/secondary_endpoint_client_linux.go
```diff
@@ -153,53 +156,112 @@ func (client *SecondaryEndpointClient) ConfigureContainerInterfacesAndRoutes(epI
...
+			delete(ep.SecondaryInterfaces, iface)
+		}
+
+		return nil
+	})
+}
+
+// FetchInterfacesFromNetnsPath finds all interfaces from the specified netns path except non-eth interfaces,
+// and return slice of EndpointInfo that includes interface name, NICType, netnspath and IP addresses.
+func (client *SecondaryEndpointClient) FetchInterfacesFromNetnsPath(infraInterfaceName, netnspath string) ([]*EndpointInfo, error) {
```
behzad-mir, line 183: upon discussion with @QxBytes we decided to move this function and its dependencies to the endpoint_linux since this will now cover both SwiftV1 and SwiftV2 Asynch delete case.
We are going to fetch all intrfaces available and its IP addresses and complete the Dataplane cleanup to remain consistent with Stateful CNI

cni/network/network.go
tamilmani1989: can you update comment then "will not be created"

PR #4002, opened 2025-09-04
[WIP] New /ibdevices CNS API

pjohnst5:
Scaffolding for new CNS `/ibdevices` API

This is a quick exploration of how CNS would create the MTPNC on a /ibdevices POST

I see we already have some MTPNC reconciling logic in CNS if Swiftv2 is enabled, so will chat with stakeholders about that

PR #4003, opened 2025-09-04
Add zap logger to cns apis

pjohnst5:
I see we now have a zap logger v2 for cns

In #4002. I am getting:
`SA1019: logger.Printf is deprecated: The global logger is deprecated. Migrate to zap using the cns/logger/v2 package and pass the logger instead. (staticcheck)`

So this PR is to introduce the zap v2 logger into CNS APIs, by plumbing it through the CNS service config

Thoughts?

This would make
https://github.com/Azure/azure-container-networking/blob/f636b50e6086d42216bed8d5190cb8db0d47ef81/cns/restserver/api.go#L48

Into something like this:
```go
service.Logger.Info("[Azure CNS] setEnvironment")
```

I understand we already have a way to make `logger.Printf()` to point to a v2 zap logger underneath:
https://github.com/Azure/azure-container-networking/blob/f636b50e6086d42216bed8d5190cb8db0d47ef81/cns/service/main.go#L650-L653

But this is more about when adding _new_ APIs, and to avoid the linting errors

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -646,6 +646,7 @@ func main() {
 	}
 	host, _ := os.Hostname()
 	z = z.With(zap.String("hostname", host), zap.String("version", version), zap.String("kubernetes_apiserver", os.Getenv("KUBERNETES_SERVICE_HOST")))
+	config.Logger = z
 	// Set the v2 logger to the global logger if v2 logger enabled.
 	if cnsconfig.EnableLoggerV2 {
 		logger.Printf("hotswapping logger v2") //nolint:staticcheck // ignore new deprecation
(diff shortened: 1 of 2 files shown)


cns/service/main.go
```diff
@@ -646,6 +646,7 @@ func main() {
 	}
 	host, _ := os.Hostname()
 	z = z.With(zap.String("hostname", host), zap.String("version", version), zap.String("kubernetes_apiserver", os.Getenv("KUBERNETES_SERVICE_HOST")))
+	config.Logger = z
```
rbtr: don't pass the bare root logger, create a derived one (`z.WithFields(...)`) that indicates it's the `Service` Logger

rbtr (commented):
what if V2 Logger is not enabled?

cns/service/main.go
pjohnst5: Alright, I passed in a zap logger with fields now
To your question about v2 logger not being enabled, yeah let me think about that

rbtr (approved):

paulyufan2 (approved):

PR #4012, opened 2025-09-05
refactor: Update apipa nic as separate entry in podIPInfo

tamilmani1989:
This PR updates both CNS and CNI code to construct apipa nic as separate entry in podIpInfo if either of allowhostonc or allownctohost set. This allows CNI to treat this as separate endpoint and align with current cni design/model of 1 nic per endpoint info. CNI then iterates through endpoint info and creates one nic at a time.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -192,6 +192,11 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 			if err := addBackendNICToResult(&info, &addResult, key); err != nil {
 				return IPAMAddResult{}, err
 			}
+		case cns.ApipaNIC:
+			if err := configureApipaAddResult(&info, &addResult, &response.PodIPInfo[i].PodIPConfig, key); err != nil {
+				return IPAMAddResult{}, err
+			}
+
 		case cns.InfraNIC, "":
 			// if we change from legacy cns, the nicType will be empty, so we assume it is infra nic
 			info.nicType = cns.InfraNIC
@@ -503,6 +508,29 @@ func configureSecondaryAddResult(info *IPResultInfo, addResult *IPAMAddResult, p
 		NICType:           info.nicType,
 		MacAddress:        macAddress,
 		SkipDefaultRoutes: info.skipDefaultRoutes,
+	}
+
+	return nil
+}
+
+func configureApipaAddResult(info *IPResultInfo, addResult *IPAMAddResult, podIPConfig *cns.IPSubnet, key string) error {
+	ip, ipnet, err := podIPConfig.GetIPNet()
+	if ip == nil {
+		return errors.Wrap(err, "Unable to parse IP from response: "+info.podIPAddress+" with err %w")
+	}
+
+	addResult.interfaceInfo[key] = network.InterfaceInfo{
+		IPConfigs: []*network.IPConfig{
+			{
+				Address: net.IPNet{
+					IP:   ip,
+					Mask: ipnet.Mask,
+				},
+				Gateway: net.ParseIP(info.ncGatewayIPAddress),
+			},
+		},
+		NICType:           info.nicType,
+		SkipDefaultRoutes: true,
 	}
 
 	return nil
--- a/cns/restserver/ipam.go
+++ b/cns/restserver/ipam.go
@@ -149,6 +149,7 @@ func (service *HTTPRestService) requestIPConfigHandlerHelperStandalone(ctx conte
 
 	// assign NICType and MAC Address for SwiftV2. we assume that there won't be any SwiftV1 NCs here
 	podIPInfoList := make([]cns.PodIpInfo, 0, len(resp))
+	apipaIndex := -1
 	for i := range resp {
 		podIPInfo := cns.PodIpInfo{
 			PodIPConfig:                     resp[i].IPConfiguration.IPSubnet,
@@ -157,6 +158,21 @@ func (service *HTTPRestService) requestIPConfigHandlerHelperStandalone(ctx conte
 			NetworkContainerPrimaryIPConfig: resp[i].IPConfiguration,
 		}
 		podIPInfoList = append(podIPInfoList, podIPInfo)
+		if resp[i].AllowHostToNCCommunication || resp[i].AllowNCToHostCommunication {
+			apipaIndex = i
+		}
+	}
+
+	if apipaIndex != -1 {
+		apipaPodIPInfo := cns.PodIpInfo{
+			PodIPConfig:                     resp[apipaIndex].LocalIPConfiguration.IPSubnet,
+			NICType:                         cns.ApipaNIC,
+			NetworkContainerPrimaryIPConfig: resp[apipaIndex].LocalIPConfiguration,
+			SkipDefaultRoutes:               true,
+			AllowHostToNCCommunication:      resp[apipaIndex].AllowHostToNCCommunication,
+			AllowNCToHostCommunication:      resp[apipaIndex].AllowNCToHostCommunication,
+		}
+		podIPInfoList = append(podIPInfoList, apipaPodIPInfo)
 	}
 
 	ipConfigsResp := &cns.IPConfigsResponse{
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -93,6 +93,9 @@ const (
 	NodeNetworkInterfaceFrontendNIC NICType = "FrontendNIC"
 	// NodeNetworkInterfaceBackendNIC is the new name for BackendNIC
 	NodeNetworkInterfaceBackendNIC NICType = "BackendNIC"
+
+	// ApipaNIC is used for internal communication between host and container
+	ApipaNIC NICType = "ApipaNIC"
 )
 
 // ChannelMode :- CNS channel modes
@@ -516,6 +519,10 @@ type PodIpInfo struct {
 	PnPID string
 	// Default Deny ACL's to configure on HNS endpoints for Swiftv2 window nodes
 	EndpointPolicies []policy.Policy
+	// This flag is in effect only if nic type is apipa. This allows connection originating from host to container via apipa nic and not other way.
+	AllowHostToNCCommunication bool
+	// This flag is in effect only if nic type is apipa. This allows connection originating from container to host via apipa nic and not other way.
+	AllowNCToHostCommunication bool
 }
 
 type HostIPInfo struct {
--- a/network/network_windows.go
+++ b/network/network_windows.go
@@ -342,7 +342,8 @@ func (nm *networkManager) addIPv6DefaultRoute() error {
 // newNetworkImplHnsV2 creates a new container network for HNSv2.
 func (nm *networkManager) newNetworkImplHnsV2(nwInfo *EndpointInfo, extIf *externalInterface) (*network, error) {
 	// network creation is not required for IB
-	if nwInfo.NICType == cns.BackendNIC {
+	// For apipa nic, we create network as part of endpoint creation
+	if nwInfo.NICType == cns.BackendNIC || nwInfo.NICType == cns.ApipaNIC {
 		return &network{Endpoints: make(map[string]*endpoint)}, nil
 	}
 


tamilmani1989:
reopen

cni/network/network_windows.go
```diff
@@ -75,10 +75,13 @@ func (plugin *NetPlugin) getNetworkName(netNs string, interfaceInfo *network.Int
 	// Swiftv2 L1VH Network Name
 	swiftv2NetworkNamePrefix := "azure-"
 	if interfaceInfo != nil && (interfaceInfo.NICType == cns.NodeNetworkInterfaceFrontendNIC || interfaceInfo.NICType == cns.BackendNIC) {
-		logger.Info("swiftv2", zap.String("network name", interfaceInfo.MacAddress.String()))
 		return swiftv2NetworkNamePrefix + interfaceInfo.MacAddress.String(), nil
 	}
 
+	if interfaceInfo != nil && interfaceInfo.NICType == cns.ApipaNIC {
+		return swiftv2NetworkNamePrefix + "apipa", nil
```
akanso: we could use a constant here for "apipa"
apipaInterfacePrefix = "apipa"

cni/network/network.go
```diff
@@ -643,6 +644,8 @@ func (plugin *NetPlugin) findMasterInterface(opt *createEpInfoOpt) string {
 		// when the VF is dismounted, this interface will go away
 		// return an unique interface name to containerd
 		return ibInterfacePrefix + strconv.Itoa(opt.endpointIndex)
+	case cns.ApipaNIC:
+		return apipaInterfacePrefix + strconv.Itoa(opt.endpointIndex)
```
QxBytes, line 648: just wanted to point out that the returned value here if apipa nic is not constant if there are multiple endpoint infos-- in one run endpointIndex could be 0, the other it could be 1 based on how we traverse the loop-- if this is fine you can ignore this comment

cni/network/network.go
```diff
@@ -757,8 +760,11 @@ func (plugin *NetPlugin) createEpInfo(opt *createEpInfoOpt) (*network.EndpointIn
 		IPAddresses: addresses,
 		MacAddress:  opt.ifInfo.MacAddress,
 		// the following is used for creating an external interface if we can't find an existing network
-		HostSubnetPrefix: opt.ifInfo.HostSubnetPrefix.String(),
-		PnPID:            opt.ifInfo.PnPID,
+		HostSubnetPrefix:         opt.ifInfo.HostSubnetPrefix.String(),
+		PnPID:                    opt.ifInfo.PnPID,
+		NetworkContainerID:       opt.ifInfo.NetworkContainerID,
```
QxBytes, line 765: can we add a comment or line break after the PnPID: line-- don't want the comment "the following is used..." to make it look like it applies to the NC ID or AllowInboundFrom... fields

cns/restserver/ipam.go
```diff
@@ -149,14 +149,33 @@ func (service *HTTPRestService) requestIPConfigHandlerHelperStandalone(ctx conte
...
 			PodIPConfig:                     resp[i].IPConfiguration.IPSubnet,
 			MacAddress:                      resp[i].NetworkInterfaceInfo.MACAddress,
 			NICType:                         resp[i].NetworkInterfaceInfo.NICType,
 			NetworkContainerPrimaryIPConfig: resp[i].IPConfiguration,
+			NetworkContainerID:              resp[i].NetworkContainerID,
 		}
 		podIPInfoList = append(podIPInfoList, podIPInfo)
+		if resp[i].AllowHostToNCCommunication || resp[i].AllowNCToHostCommunication {
+			apipaIndex = i
+			break
```
QxBytes: If we had 3 items in resp:
0 is infra nic
1 is apipa nic (either allow host to nc communication or allow nc to host communication)
2 is frontend nic
when we get to 1, we set apipa index to 1 and break out of the loop. Does that mean we never append the podIPInfo for 2: frontend nic?

Previously it looked like we added all the items in resp to the pod ip info list.

cni/network/invoker_cns.go
```diff
@@ -508,6 +513,32 @@ func configureSecondaryAddResult(info *IPResultInfo, addResult *IPAMAddResult, p
 	return nil
 }
 
+func configureApipaAddResult(addResult *IPAMAddResult, info *cns.PodIpInfo, key string) error {
```
QxBytes, line 516: can we add a ut for the case when cns returns a response with the apipa nic?

cns/restserver/ipam.go
```diff
@@ -1123,11 +1142,76 @@ func (service *HTTPRestService) EndpointHandlerAPI(w http.ResponseWriter, r *htt
 		service.GetEndpointHandler(w, r)
 	case http.MethodPatch:
 		service.UpdateEndpointHandler(w, r)
+	case http.MethodDelete:
+		service.DeleteEndpointStateHandler(w, r)
```
QxBytes: wondering how we deleted before if we're only adding the delete state here now?

cni/netconfig.go
```diff
@@ -72,6 +72,7 @@ type NetworkConfig struct {
 	EnableExactMatchForPodName    bool            `json:"enableExactMatchForPodName,omitempty"`
 	DisableHairpinOnHostInterface bool            `json:"disableHairpinOnHostInterface,omitempty"`
 	DisableIPTableLock            bool            `json:"disableIPTableLock,omitempty"`
+	DisableAsyncDelete            bool            `json:"disableAsyncDelete,omitempty"`
```
QxBytes, line 75: are we planning to add this field to the cns's cni conflist generation code? when would it be used?

network/endpoint.go
```diff
@@ -128,17 +128,20 @@ type RouteInfo struct {
...
+	IPConfigs                  []*IPConfig
+	Routes                     []RouteInfo
+	DNS                        DNSInfo
+	NICType                    cns.NICType
+	SkipDefaultRoutes          bool
+	HostSubnetPrefix           net.IPNet // Move this field from ipamAddResult
+	NCResponse                 *cns.GetNetworkContainerResponse
+	PnPID                      string
+	EndpointPolicies           []policy.Policy
+	NetworkContainerID         string
```
QxBytes, line 143: could we add a comment for when the nc container id, allow nc to host, and allow host to nc are populated and used?

cni/network/network.go
QxBytes: also can we add these 3 new fields to the PrettyString()s in network/endpoint.go for interfaceInfo and endpointInfo if you think they would be helpful for debugging?

cns/endpointmanager/endpointmanager_windows.go
```diff
@@ -16,6 +16,7 @@ func (em *EndpointManager) ReleaseIPs(ctx context.Context, ipconfigreq cns.IPCon
 	if err := em.deleteEndpoint(ctx, ipconfigreq.InfraContainerID); err != nil {
 		logger.Errorf("failed to remove HNS endpoint %s", err.Error())
 	}
+
```
behzad-mir: Seems to be extra?

cni/network/network.go
```diff
@@ -1072,7 +1078,7 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 		epInfos, err = plugin.nm.GetEndpointState(networkID, args.ContainerID)
 		// if stateless CNI fail to get the endpoint from CNS for any reason other than  Endpoint Not found
 		if err != nil {
-			if errors.Is(err, network.ErrConnectionFailure) {
+			if errors.Is(err, network.ErrConnectionFailure) && !nwCfg.DisableAsyncDelete {
```
behzad-mir, line 1082: Is Asych delete always disabled in ACI? If yes can we have a comment about it?

network/manager.go
```diff
@@ -771,6 +772,21 @@ func (nm *networkManager) DeleteState(_ []*EndpointInfo) error {
 	// For stateless cni, plugin.ipamInvoker.Delete takes care of removing the state in the main Delete function
 
 	if nm.IsStatelessCNIMode() {
+		for _, epInfo := range epInfos {
+			// this cleanup happens only for standalone swiftv2 to delete endpoint state from CNS.
+			if epInfo.NICType == cns.NodeNetworkInterfaceFrontendNIC || epInfo.NICType == cns.NodeNetworkInterfaceAccelnetFrontendNIC {
```
behzad-mir, line 776: Since the Apipa endpoint and frontend endpoint are both part of the same State entry in statefile, is it safe to assume that Apipa endpoint can also be deleted from statefile and datapath once the frontend endpoint gets deleted? 
We need confirmaton from the logs that, how many deletState will be issued to the CNS for deleting a container that also has Apipa endpoint.

network/manager.go
```diff
@@ -771,6 +772,21 @@ func (nm *networkManager) DeleteState(_ []*EndpointInfo) error {
...
+				response, err := nm.CnsClient.DeleteEndpointState(context.TODO(), epInfo.EndpointID)
+				if err != nil {
+					if response != nil && response.ReturnCode == types.NotFound {
+						logger.Info("Endpoint state not found in CNS", zap.String("endpointID", epInfo.EndpointID))
+						return nil
+					}
+					return errors.Wrapf(err, "Delete endpoint API returned with error for endpoint %s", epInfo.EndpointID)
```
behzad-mir, line 786: This error here will fail the delete call, so for whatever reason (CNS is down for example) if the endpoint cannot get deleted form statefile the whole Delete will fall in a loop. We have to be careful about catching errors from CNS here.

cni/network/network.go
akanso: Asych delete is configurable in the conflist JSON file. for now it is disabled, but in the future this might change. What type of comment you recommend adding?

network/manager.go
```diff
@@ -771,6 +772,21 @@ func (nm *networkManager) DeleteState(_ []*EndpointInfo) error {
 	// For stateless cni, plugin.ipamInvoker.Delete takes care of removing the state in the main Delete function
 
 	if nm.IsStatelessCNIMode() {
+		for _, epInfo := range epInfos {
+			// this cleanup happens only for standalone swiftv2 to delete endpoint state from CNS.
+			if epInfo.NICType == cns.NodeNetworkInterfaceFrontendNIC || epInfo.NICType == cns.NodeNetworkInterfaceAccelnetFrontendNIC {
+				response, err := nm.CnsClient.DeleteEndpointState(context.TODO(), epInfo.EndpointID)
+				if err != nil {
+					if response != nil && response.ReturnCode == types.NotFound {
+						logger.Info("Endpoint state not found in CNS", zap.String("endpointID", epInfo.EndpointID))
```
QxBytes, line 783: Will there ever only be one NodeNetworkInterfaceFrontendNIC or NodeNetworkInterfaceAccelnetFrontendNIC ? Or did we want to call DeleteEndpointState for each of them?

Maybe for organization we can have a helper method, or a more thorough comment?

network/manager.go
```diff
@@ -771,6 +772,21 @@ func (nm *networkManager) DeleteState(_ []*EndpointInfo) error {
 	// For stateless cni, plugin.ipamInvoker.Delete takes care of removing the state in the main Delete function
```
QxBytes: can we update this comment. Also we have nm.DeleteEndpointState (called in DeleteEndpoint) which is different from nm.CnsClient.DeleteEndpointState (called here)-- is there a better way we can indicate which functions are only called during stateless cni?

Possibly:
nm.DeleteEndpointState -> nm.DeleteEndpointStateless
nm.cnsClient.DeleteEndpointState -> nm.cnsClient.DeleteEndpointStateStateless

cns/client/client.go
```diff
@@ -1100,3 +1100,35 @@ func (c *Client) UpdateEndpoint(ctx context.Context, endpointID string, ipInfo m
 
 	return &response, nil
 }
+
+func (c *Client) DeleteEndpointState(ctx context.Context, endpointID string) (*cns.Response, error) {
```
QxBytes, line 1106: can we add a comment for when it's used or what it should do on cns side?

network/manager.go
```diff
@@ -771,6 +772,21 @@ func (nm *networkManager) DeleteState(_ []*EndpointInfo) error {
...
 
 	if nm.IsStatelessCNIMode() {
+		for _, epInfo := range epInfos {
+			// this cleanup happens only for standalone swiftv2 to delete endpoint state from CNS.
```
QxBytes, line 775: Is the reason we need this explicit deletion because
1) Previously, in non-multitenancy scenarios ipamInvoker.Delete would automatically clean up state (cns finds infra container id)
and
2) Now, swiftv2 multitenancy does not call plugin.ipamInvoker.Delete and so state does not automatically clean up?
If not can we add the reason to the pr description

network/endpoint_windows.go
```diff
@@ -529,6 +549,13 @@ func (nw *network) deleteEndpointImpl(_ netlink.NetlinkInterface, _ platform.Exe
 		return nil
 	}
 
+	if ep.NICType == cns.ApipaNIC {
+		if err := nw.deleteHostNCApipaEndpoint(ep.NetworkContainerID); err != nil {
+			logger.Error("Failed to delete HostNCApipaEndpoint due to error", zap.Error(err))
```
QxBytes: nit: we log the error here, but don't we also return it and log it at the top level already?

cni/network/network_windows.go
```diff
@@ -75,10 +75,13 @@ func (plugin *NetPlugin) getNetworkName(netNs string, interfaceInfo *network.Int
 	// Swiftv2 L1VH Network Name
 	swiftv2NetworkNamePrefix := "azure-"
 	if interfaceInfo != nil && (interfaceInfo.NICType == cns.NodeNetworkInterfaceFrontendNIC || interfaceInfo.NICType == cns.BackendNIC) {
-		logger.Info("swiftv2", zap.String("network name", interfaceInfo.MacAddress.String()))
 		return swiftv2NetworkNamePrefix + interfaceInfo.MacAddress.String(), nil
 	}
 
+	if interfaceInfo != nil && interfaceInfo.NICType == cns.ApipaNIC {
```
QxBytes, line 81: can we add a ut case in TestGetNetworkNameSwiftv2FromCNS in network_windows_test.go?

cni/netconfig.go
tamilmani1989: nope. this is introduced for standalone scenario which use stateless cni but not to use async delete.

cni/network/network.go
tamilmani1989: ya should be fine. cni wont be using this field at all..just to satisfy traditional code path

cni/network/invoker_cns_test.go
```diff
@@ -2298,3 +2298,316 @@ func TestMultipleIBNICsToResult(t *testing.T) {
...
+					apipaInterfaceFound = true
+					actualApipaResult = ifInfo
+					break
+				}
+			}
+
+			require.True(apipaInterfaceFound, "ApipaNIC interface should be found in the result")
+
+			// Verify the ApipaNIC interface info
+			// Lines around 2586-2590 should be:
```
QxBytes, line 2584: nit: the line numbers might change so this comment probably isn't necessary

cni/network/network.go
tamilmani1989: couldnt understand need for 1st comment. not addressing it. Addressed second comment

cns/restserver/ipam.go
tamilmani1989: removed break

network/manager.go
tamilmani1989: ya i updated 1st. lets revisit 2nd later

network/manager.go
tamilmani1989: yes correct

network/manager.go
tamilmani1989: yes

tamilmani1989: should be only 1. have a break after delete endpoint succeeds

network/manager.go
tamilmani1989: one delete call will remove all endpoint info. will update in comments

network/manager.go
tamilmani1989: yes i filter out "not found" errors.  Did you see any other error to be filtered out?

behzad-mir: I think we should add a endpointResponse.Response.ReturnCode == types.ConnectionError as well. This should trigger an async delete.

QxBytes (dismissed):
Also just to confirm cns/restserver/ipam.go it's okay if we have multiple podIPInfos that set `resp[i].AllowHostToNCCommunication || resp[i].AllowNCToHostCommunication` to true, right? We would just take one of them _(and it shouldn't matter which one we take)_ to create the apipa pod ip info.

tamilmani1989:
> Also just to confirm cns/restserver/ipam.go it's okay if we have multiple podIPInfos that set `resp[i].AllowHostToNCCommunication || resp[i].AllowNCToHostCommunication` to true, right? We would just take one of them _(and it shouldn't matter which one we take)_ to create the apipa pod ip info.

ya should be fine. validation will be added in dnc side to prevent this

behzad-mir (approved):

cns/restserver/ipam.go
```diff
@@ -1123,11 +1141,76 @@ func (service *HTTPRestService) EndpointHandlerAPI(w http.ResponseWriter, r *htt
...
 	default:
 		logger.Errorf("[EndpointHandlerAPI] EndpointHandler API expect http Get or Patch method")
```
santhoshmprabhu: Nit: update this log to include delete method

network/endpoint_windows_test.go
```diff
@@ -668,13 +668,13 @@ func TestDeleteEndpointStateForInfraDelegatedNIC(t *testing.T) {
 
 	// mock DeleteEndpointState() to make sure endpoint and network is deleted from cache
```
santhoshmprabhu: Nit: Update the comments.

santhoshmprabhu (commented):
I don't have much context on Swiftv2 and the need for this change. The code changes look good. A couple of nits, and also, it seems we don't have tests for the new CNS deletion API. Could we please add those as well?

santhoshmprabhu (approved):

santhoshmprabhu (approved):

PR #4042, opened 2025-09-15
Chore: update k8s.io dependencies

paulyufan2:
**Reason for Change**:

1. Change 1: Healthz:
Since sigs.k8s.io/controller-runtime v0.18.x, the default RESTMapper is discovery-based. Because the NewHealthzHandlerWithChecks() builds a client with only Scheme: cli, err := client.New(cfg, client.Options{ Scheme: scheme })
controller-runtime now hits the API discovery endpoints to map GVK→GVR before doing cli.List(...).
Our mock API server from configureLocalAPIServer() doesn’t implement discovery (and likely doesn’t return a decodable body for the list), so mapping/decoding fails → health check returns error → test expects healthy (200) but gets unhealthy.

2.Change 2: k8s.io/kubernetes/pkg/kubelet
To solve this issue after we upgrade k8s packages:
go: github.com/Azure/azure-container-networking/cns/middlewares imports
        k8s.io/kubernetes/pkg/kubelet imports
        k8s.io/cri-client/pkg: reading k8s.io/cri-client/go.mod at revision v0.0.0: unknown revision v0.0.0
go: github.com/Azure/azure-container-networking/cns/middlewares imports
        k8s.io/kubernetes/pkg/kubelet imports
        k8s.io/kubernetes/pkg/kubelet/apis/podresources imports
        k8s.io/cri-client/pkg/util: reading k8s.io/cri-client/go.mod at revision v0.0.0: unknown revision v0.0.0
        
This error is the reason why we shouldn’t import k8s.io/kubernetes/pkg/kubelet. The k8s.io/kubernetes monorepo has local replace rules for its staging deps (like k8s.io/cri-client) that don’t work outside the tree, so Go tries to resolve a fake v0.0.0 and fails. Instead we should using the staged modules to fix this.
kubelet.NetworkNotReadyErrorMsg == NetworkNotReadyErrorMsg = "network is not ready"
so make this as a constant variable makes more sense

**Issue Fixed**:

This PR is to update k8s.io dependencies along with dependencies changes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,7 +42,7 @@ require (
 	google.golang.org/protobuf v1.36.8
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
 	k8s.io/api v0.34.1
-	k8s.io/apiextensions-apiserver v0.34.0
+	k8s.io/apiextensions-apiserver v0.34.1
 	k8s.io/apimachinery v0.34.1
 	k8s.io/client-go v0.34.1
 	k8s.io/klog v1.0.0
--- a/go.sum
+++ b/go.sum
@@ -606,8 +606,8 @@ honnef.co/go/tools v0.0.0-20190102054323-c2f93a96b099/go.mod h1:rf3lG4BRIbNafJWh
 honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
 k8s.io/api v0.34.1 h1:jC+153630BMdlFukegoEL8E/yT7aLyQkIVuwhmwDgJM=
 k8s.io/api v0.34.1/go.mod h1:SB80FxFtXn5/gwzCoN6QCtPD7Vbu5w2n1S0J5gFfTYk=
-k8s.io/apiextensions-apiserver v0.34.0 h1:B3hiB32jV7BcyKcMU5fDaDxk882YrJ1KU+ZSkA9Qxoc=
-k8s.io/apiextensions-apiserver v0.34.0/go.mod h1:hLI4GxE1BDBy9adJKxUxCEHBGZtGfIg98Q+JmTD7+g0=
+k8s.io/apiextensions-apiserver v0.34.1 h1:NNPBva8FNAPt1iSVwIE0FsdrVriRXMsaWFMqJbII2CI=
+k8s.io/apiextensions-apiserver v0.34.1/go.mod h1:hP9Rld3zF5Ay2Of3BeEpLAToP+l4s5UlxiHfqRaRcMc=
 k8s.io/apimachinery v0.34.1 h1:dTlxFls/eikpJxmAC7MVE8oOeP1zryV7iRyIjB0gky4=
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -41,14 +41,14 @@ require (
 	google.golang.org/grpc v1.75.0
 	google.golang.org/protobuf v1.36.8
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
-	k8s.io/api v0.30.14
-	k8s.io/apiextensions-apiserver v0.30.1
-	k8s.io/apimachinery v0.30.14
-	k8s.io/client-go v0.30.14
+	k8s.io/api v0.34.1
+	k8s.io/apiextensions-apiserver v0.34.0
```
jpayne3506: Any reason for `k8s.io/apiextensions-apiserver v0.34.0` to not be `k8s.io/apiextensions-apiserver v0.34.1`?

cns/multitenantcontroller/multitenantoperator/multitenantcrdreconciler_test.go
```diff
@@ -14,11 +14,19 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+type clientWithApply struct{ *mockclients.MockClient }
+
+// Satisfy controller-runtime v0.22.1 Writer.Apply:
+func (c clientWithApply) Apply(_ context.Context, _ runtime.ApplyConfiguration, _ ...client.ApplyOption) error {
+	return nil // not used in these tests
+}
+
```
jpayne3506, line 29: Can I get some more context here for this change?

paulyufan2: client.Client in controller-runtime v0.22.1 added a new Apply method to support server-side apply. Our existing mockclients.MockClient only satisfied the old interface, so without an Apply implementation the tests wouldn’t compile.

To fix this, I wrapped the mock client in clientWithApply and added a no-op Apply method. We don’t actually use Apply in these tests, but this ensures the mock still satisfies the client.Client interface after the dependency upgrade.

go.mod
paulyufan2: I believe there was no v0.34.1 the time when I upgraded :) I will upgrade to v0.34.1, thanks John

rbtr (approved):

pjohnst5:
Hi @rbtr , what version of k8s is this meant for?

rbtr:
@pjohnst5 the main branch is currently the 1.7.x release train and deploys to AKS from 1.33-1.35

jpayne3506 (approved):

pjohnst5:
Okay thanks @rbtr , I will be looking into if without these, k8s v1.34 would break for example, unless you know that already (were these necessary for 1.34-5? Or just good hygiene?)

rbtr:
https://github.com/kubernetes/client-go?tab=readme-ov-file#compatibility-matrix
https://github.com/kubernetes-sigs/controller-runtime?tab=readme-ov-file#compatibility

PR #4047, opened 2025-09-18
chore: CNS - Only log CreateOrUpdateNetworkContainer on failure

peterfelts:
**Reason for Change**:
CNS logs every time a Network Container is created or updated. DNC will constantly call into CNS to refresh its association, resulting in a very high level of the following lines being logged:

`[azure-cns] Sent *cns.CreateNetworkContainerResponse &{Response:{ReturnCode:Success Message:}}.`

This line offers little value as it doesn't have context about which NC was updated.

This change will result in this response line only being logged on error.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/api.go
+++ b/cns/restserver/api.go
@@ -567,9 +567,10 @@ func (service *HTTPRestService) createOrUpdateNetworkContainer(w http.ResponseWr
 	// If the NC was created successfully, log NC snapshot.
 	if returnCode == types.Success {
 		logNCSnapshot(req)
+	} else {
+		// Only emit response trace/log for non-success responses to avoid noisy success logs.
+		logger.Response(service.Name, reserveResp, resp.ReturnCode, err)
 	}
-
-	logger.Response(service.Name, reserveResp, resp.ReturnCode, err)
 }
 
 func (service *HTTPRestService) getNetworkContainerByID(w http.ResponseWriter, r *http.Request) {


cns/restserver/api.go
```diff
@@ -567,9 +567,10 @@ func (service *HTTPRestService) createOrUpdateNetworkContainer(w http.ResponseWr
 	// If the NC was created successfully, log NC snapshot.
 	if returnCode == types.Success {
 		logNCSnapshot(req)
+	} else {
+		// Only emit response trace/log for non-success responses to avoid noisy success logs.
+		logger.Response(service.Name, reserveResp, resp.ReturnCode, err)
```
ZetaoZhuang, line 572: as the Lint says, should we switch to cns/logger/v2?

timraymond, line 572: I like this, but want to note: if this logs, it will always result in a redundant log. The only definition of `err` that reaches here is from `common.Encode`, which has logging logic within it in the case when `err != nil`. We should remove that logging logic because, IMO, it should always be the caller's responsibility to do something useful with an error.

peterfelts: As discussed, I'll leave the logging in `common.Encode` as that doesn't appear to be happening often. Also, it looks like `service.saveNetworkContainerGoalState` will return `0, ""` is most cases, and Id on't see any case where it returns a success error code along with a message, so this should be safe.

PR #4053, opened 2025-09-29
fix: Add Ifname to IPConfigsRequest from CNI

jpayne3506:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rbtr:
i'm skeptical of this. why has it never needed to be set before, and what's different that requires it to be set now?

PR #4058, opened 2025-09-30
Extending the MTPNC status

shreyashastantram:
**Reason for Change**:

This PR focuses on extending the mtpnc spec and status fields to include the Pod UID in the spec and the Node name in the status.

Pod UID is added to the spec because controllers are aware of the pod’s unique identifier at the time of mtpnc creation.
Node name is added to the status because scheduling decisions do not depend on the node name parameter. The node name becomes available only after the scheduling decision has been made.
If Node name were part of the spec, it would act as an optional input from the creator of the mtpnc. This could lead to inconsistent states—for example, if the creator specifies a node name that differs from the actual node where the pod is scheduled.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/crd/multitenancy/manifests/multitenancy.acn.azure.com_multitenantpodnetworkconfigs.yaml
+++ b/crd/multitenancy/manifests/multitenancy.acn.azure.com_multitenantpodnetworkconfigs.yaml
@@ -71,7 +71,7 @@ spec:
                 description: name of PNI object from requesting cx pod
                 type: string
               podUID:
-                description: PodUID is the uid of the pod
+                description: PodUID is the UID of the pod
                 type: string
             required:
             - podNetwork


crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -105,6 +105,13 @@ type MultitenantPodNetworkConfigStatus struct {
 	// DefaultDenyACL bool indicates whether default deny policy will be present on the pods upon pod creation
 	// +kubebuilder:validation:Optional
 	DefaultDenyACL bool `json:"defaultDenyACL"`
+	// NodeName is the name of the node where the pod is scheduled
+	// +kubebuilder:validation:Optional
+	NodeName string `json:"nodeName,omitempty"`
+	// PodName is the name of the pod
+	PodName string `json:"podName,omitempty"`
```
ashvindeodhar: is this ownerKey ( podname + podnamespace ) or just podname?

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -105,6 +105,13 @@ type MultitenantPodNetworkConfigStatus struct {
...
+	// PodUID is the uid of the pod
+	PodUID string `json:"podUID,omitempty"`
```
ashvindeodhar: curious to know what this would be useful for?

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -105,6 +105,13 @@ type MultitenantPodNetworkConfigStatus struct {
 	// DefaultDenyACL bool indicates whether default deny policy will be present on the pods upon pod creation
 	// +kubebuilder:validation:Optional
 	DefaultDenyACL bool `json:"defaultDenyACL"`
+	// NodeName is the name of the node where the pod is scheduled
```
isaac-dasan, line 110: These new fields are essential info going forward and I wonder if these new fields should be in the spec itself, considering MTPNC wouldn't reconcile successfully if this information isn't present in the near future. 
For backward compat to existing MTPNCs in a given cluster, I know we don't mutate the spec as of now. But it is allowed and may be we do it for this just once and re update the spec?

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
isaac-dasan: I would assume just podName because MTPNC is also created in the same namepace as the pod.

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
isaac-dasan: To begin with we want to query MTPNC by podUID, because we have seen some cases where statefulSets create pods with same namd and the second pod sometimes reuses the MTPNC from the first pod. In the future we want name MTPNC based on podUID, so it is unique per pod.

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
isaac-dasan: After reviewing team, we’ve aligned on keeping nodeName in the status section.
While nodeName can be provided by the creator of the MTPNC, it is not required for the controller to function. In fact, allowing it as an optional input in spec could lead to inconsistent states, especially if the provided nodeName differs from the actual node where the Pod is scheduled.
By placing it in status, we make it clear that:

- It is not an input for the creator of the MTPNC.
- The MTPNC reconciler is responsible for observing the Pod and populating this field.
- This helps reduce the burden on the MTPNC creator (DNCRC, CNS, DRA) by automating part of the goal state construction.

Additionally, nodeName in status is required for MTPNC during deletion, as it serves as a memory of the node the Pod was scheduled on. If the Pod is already gone at the time of deletion, the controller cannot infer the nodeName from the current state, so persisting it in status ensures the reconciler has the necessary context.
We’re also aiming for a model where fields in spec are required inputs once we clean up existing MTPNC objects. Keeping nodeName in status aligns with that direction and avoids introducing optional inputs that could complicate validation and reconciliation logic.

PR #4060, opened 2025-10-01
 [feat] Prefix on NIC v6 windows support

NihaNallappagari:
**Reason for Change**:
This PR includes the required changes for Windows to support prefix on NIC  NIC Swift v2 scenario. The HNS team requires certain values to be added to the registry, and this PR sets those values under the appropriate registry key.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:
This PR is extension of [Prefix on nic linux support](https://github.com/Azure/azure-container-networking/pull/3658)

Verifying registry keys...
Reading all values from SYSTEM\CurrentControlSet\Services\HNS\wcna_state\config\PrefixOnNic
Registry verification - enabled: 1
Registry verification - infra_nic_mac_address: 60:45:bd:75:e3:73
Registry verification - infra_nic_ifname: vEthernet (Ethernet 3)
Reading all values from SYSTEM\CurrentControlSet\Services\HNS\wcna_state\config
Registry verification - EnableSNAT: 0

windows keys

### DIFF
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -229,7 +229,10 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
 		return len(programmedNCs), errors.Wrap(err, "failed to get nc version list from nmagent")
 	}
 	// Get IMDS NC versions for delegated NIC scenarios.
-	imdsNCVersions := service.getIMDSNCDetails(ctx)
+	imdsNCVersions, err := service.getIMDSNCDetails(ctx)
+	if err != nil {
+		return len(programmedNCs), errors.Wrap(err, "failed to get NC details from IMDS")
+	}
 
 	nmaNCs := map[string]string{}
 	for _, nc := range ncVersionListResp.Containers {
@@ -686,30 +689,35 @@ func (service *HTTPRestService) isNCDetailsAPIExists(ctx context.Context) bool {
 	return false
 }
 
-// gets nc details from IMDS and returns perform platform specific action
-func (service *HTTPRestService) getIMDSNCDetails(ctx context.Context) map[string]string {
+// getIMDSNCDetails retrieves network container information from IMDS for delegated NIC scenarios and performs platform-specific configuration.
+//
+// Returns a map of NC ID -> NC version for delegated NIC retrieved from IMDS.
+// Windows only:
+//   - Sets Windows registry keys required by HNS for Cilium route configuration in prefix-on-NIC scenarios.
+//
+// The returned map is used by syncHostNCVersion to consolidate NC version information from both NMAgent and IMDS sources.
+func (service *HTTPRestService) getIMDSNCDetails(ctx context.Context) (map[string]string, error) {
 	imdsClient := service.imdsClient
 	if imdsClient == nil {
-		//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
-		logger.Errorf("IMDS client is not available")
-		return map[string]string{}
+		return nil, errors.New("IMDS client is not available")
 	}
+
 	// Check NC version support
 	if !service.isNCDetailsAPIExists(ctx) {
-		//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
-		logger.Errorf("IMDS does not support NC details API")
-		return map[string]string{}
+		return nil, errors.New("IMDS does not support NC details API")
 	}
 
 	// Get all network interfaces from IMDS
 	networkInterfaces, err := imdsClient.GetNetworkInterfaces(ctx)
 	if err != nil {
-		//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
-		logger.Errorf("Failed to get network interfaces from IMDS: %v", err)
-		return map[string]string{}
+		return nil, errors.Wrap(err, "failed to get network interfaces from IMDS")
 	}
 
-	return service.processIMDSData(networkInterfaces)
+	ncs, err := service.processIMDSData(networkInterfaces)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to process IMDS data")
+	}
+	return ncs, nil
 }
 
 // IsCIDRSuperset returns true if newCIDR is a superset of oldCIDR (i.e., all IPs in oldCIDR are contained in newCIDR).
--- a/cns/restserver/internalapi_linux.go
+++ b/cns/restserver/internalapi_linux.go
@@ -35,7 +35,10 @@ func (c *iptablesLegacy) Delete(table, chain string, rulespec ...string) error {
 	return errors.Wrap(exec.Command("iptables-legacy", cmd...).Run(), "iptables legacy failed delete")
 }
 
-func (service *HTTPRestService) processIMDSData(networkInterfaces []imds.NetworkInterface) map[string]string {
+// processIMDSData, processes network interface data from IMDS for prefix-on-NIC scenarios on Linux.
+// It extracts the delegated NIC's information from IMDS network interfaces.
+// A delegated NIC is identified by having a non-empty InterfaceCompartmentID (NC ID).
+func (service *HTTPRestService) processIMDSData(networkInterfaces []imds.NetworkInterface) (map[string]string, error) {
 	ncs := make(map[string]string)
 
 	for _, iface := range networkInterfaces {
@@ -46,7 +49,7 @@ func (service *HTTPRestService) processIMDSData(networkInterfaces []imds.Network
 		}
 	}
 
-	return ncs
+	return ncs, nil
 }
 
 // nolint
--- a/cns/restserver/internalapi_windows.go
+++ b/cns/restserver/internalapi_windows.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/cns/imds"
-	"github.com/Azure/azure-container-networking/cns/logger"
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/netio"
 	"github.com/Microsoft/hcsshim"
@@ -102,7 +101,13 @@ func (service *HTTPRestService) isSwiftV2PrefixOnNicEnabled(ncID string) bool {
 	return false
 }
 
-func (service *HTTPRestService) processIMDSData(networkInterfaces []imds.NetworkInterface) map[string]string {
+// processIMDSData processes network interface data from IMDS and configures Windows registry for prefix-on-NIC scenarios.
+//
+//	It extracts the delegated NIC's information from IMDS network interfaces.
+//
+//	Configures Windows registry if an infra NIC is present, this function sets registry keys
+//	   required by HNS to configure Cilium routes for Windows nodes
+func (service *HTTPRestService) processIMDSData(networkInterfaces []imds.NetworkInterface) (map[string]string, error) {
 	ncs := make(map[string]string)
 	var infraNicMacAddress string
 	var isSwiftv2PrefixOnNic bool
@@ -123,17 +128,14 @@ func (service *HTTPRestService) processIMDSData(networkInterfaces []imds.Network
 		// Get the interface name from the MAC address
 		infraNicIfName, err := service.getInterfaceNameFromMAC(infraNicMacAddress)
 		if err != nil {
-			//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
-			logger.Errorf("Failed to get interface name from MAC address to set for windows registry: %v", err)
-		} else {
-			// Process Windows registry keys with the retrieved MAC address and interface name. It is required for HNS team to configure cilium routes specific to windows nodes
-			if err := service.setRegistryKeysForPrefixOnNic(isSwiftv2PrefixOnNic, infraNicMacAddress, infraNicIfName); err != nil {
-				//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
-				logger.Errorf("Failed to set windows registry keys: %v", err)
-			}
+			return nil, fmt.Errorf("failed to get interface name from MAC address to set for windows registry: %w", err)
+		}
+		// Process Windows registry keys with the retrieved MAC address and interface name. It is required for HNS team to configure cilium routes specific to windows nodes
+		if err := service.setRegistryKeysForPrefixOnNic(isSwiftv2PrefixOnNic, infraNicMacAddress, infraNicIfName); err != nil {
+			return nil, fmt.Errorf("failed to set windows registry keys for prefix on nic scenario: %w", err)
 		}
 	}
-	return ncs
+	return ncs, nil
 }
 
 // getInterfaceNameFromMAC retrieves the network interface name given its MAC address.
--- a/cns/restserver/internalapi_linux_test.go
+++ b/cns/restserver/internalapi_linux_test.go
@@ -385,7 +385,10 @@ func TestAddSNATRules(t *testing.T) {
 func TestProcessIMDSData_EmptyInterfaces(t *testing.T) {
 	service := &HTTPRestService{}
 
-	result := service.processIMDSData([]imds.NetworkInterface{})
+	result, err := service.processIMDSData([]imds.NetworkInterface{})
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
 
 	if len(result) != 0 {
 		t.Errorf("Expected empty result for empty input, got %d NCs", len(result))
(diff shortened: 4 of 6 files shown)


cns/restserver/internalapi_windows.go
```diff
@@ -75,3 +84,72 @@ func (service *HTTPRestService) getPrimaryNICMACAddress() (string, error) {
...
+
+func (service *HTTPRestService) setInfraNicIfName(ifName string) error {
+	return service.setRegistryValue(prefixOnNicRegistryPath, "infra_nic_ifname", ifName)
+}
+
+func (service *HTTPRestService) setEnableSNAT(isEnabled bool) error {
+	return service.setRegistryValue(hnsRegistryPath, "EnableSNAT", isEnabled)
+}
+
+func (service *HTTPRestService) setPrefixOnNICRegistry(enablePrefixOnNic bool, infraNicMacAddress string) error {
```
tamilmani1989: why this is required for hns?

cns/restserver/internalapi.go
```diff
@@ -724,10 +721,28 @@ func (service *HTTPRestService) GetIMDSNCs(ctx context.Context) (map[string]stri
 
 		if ncID != "" {
 			ncs[ncID] = PrefixOnNicNCVersion // for prefix on nic version scenario nc version is 1
+		} else if runtime.GOOS == "windows" && service.isPrefixonNicSwiftV2() {
```
tamilmani1989: GOOS=windows not required since you are defining in windows specific file, linux specific calls would be dummy.. i think its better we expose all regkey apis through interface.. @behzad-mir / @rbtr to comment on this

cns/restserver/internalapi_windows.go
tamilmani1989: since we are setting multiple regkeys under same function,  can function name be `setRegistryKeysForPrefixOnNic` and then expose all registry key functions in interface, so that unit test can be written mocking it

cns/restserver/internalapi.go
rbtr: Good feedback. IMO any explicit check to GOOS or stubbing an interface in an `*_<os>.go` file indicates that the abstraction is in the wrong place.
There was a similar discussion here https://github.com/Azure/azure-container-networking/pull/3286#discussion_r1906112932 which you should read so that I don't have to repeat it all.

The tl;dr is that our abstraction should live at the last boundary of common code _before_ we start splitting things up by OS. E.g. here: if `getIMDSNCs` is the final call possible before we have to do OS specific things, _that_ is what lives in **both** the `*_linux.go` and the `*_windows.go`. 

However - I haven't done a full review of this code, but at first glance `getIMDSNCs` is probably not the place to be doing OS specific things at all. That advertises that it just getting NCs. 
And if we think about duplicating this func in OS-specific files, we end up copying a bunch of common code to both locations that doesn't care about the OS at all. This probably all just means that the current func is overloaded and poorly defined. Consider how to separate things in to the minimum necessary complete behaviors.

cns/restserver/internalapi_windows.go
NihaNallappagari: HNS team need to know thinks like infra nic, delegated nic, swiftv2 prefix on nic scenario to probably add some routes specific to the scenario. HNS team might answer better, tagging @surajkrishnan14

NihaNallappagari: Refactored as per suggestion

cns/restserver/internalapi.go
NihaNallappagari: removed the os check and moved windowskey process call form getIMDSNCs . Thanks for the feedback

cns/restserver/internalapi_windows.go
surajkrishnan14: Yes, this is required for HNS to setup appropriate rules to forward packets correctly between the 2 interfaces and container in various scenarios, setup SNAT etc.

cns/restserver/restserver.go
```diff
@@ -239,6 +248,7 @@ func NewHTTPRestService(config *common.ServiceConfig, wscli interfaceGetter, wsp
 		dockerClient:             dc,
 		wscli:                    wscli,
 		iptables:                 iptg,
+		windowsRegistry:          newRegistryClient(),
```
QxBytes: wondering if we should pass this in like how we do with `iptablesGetter` for example

cns/restserver/internalapi_linux.go
```diff
@@ -34,6 +34,44 @@ func (c *iptablesLegacy) Delete(table, chain string, rulespec ...string) error {
...
+}
+
+// SetEnableSNAT is a no-op on Linux.
+func (*linuxRegistryClient) SetEnableSNAT(_ bool) error {
+	return nil
+}
+
+// newLinuxRegistryClient creates a no-op registry client for Linux.
+func newLinuxRegistryClient() windowsRegistryClient {
+	return &linuxRegistryClient{}
```
QxBytes: `newLinuxRegistryClient()` seems to be called only from `newRegistryClient()`-- can we just have `newRegistryClient()` return `&linuxRegistryClient` directly?

cns/restserver/internalapi.go
```diff
@@ -724,10 +724,24 @@ func (service *HTTPRestService) GetIMDSNCs(ctx context.Context) (map[string]stri
 
 		if ncID != "" {
 			ncs[ncID] = PrefixOnNicNCVersion // for prefix on nic version scenario nc version is 1
+		} else {
+			infraNicMacAddress = iface.MacAddress.String()
```
QxBytes: what does having an empty ncID mean and when does this happen? can this be added as a comment for the function? can we also clarify in the comment what mac address we are getting? will the infra nic mac address ever be empty if all the network interfaces have nc ids?

cns/kubecontroller/nodenetworkconfig/conversion_windows_test.go
```diff
@@ -79,4 +80,5 @@ var validVNETBlockRequest = &cns.CreateNetworkContainerRequest{
 			NCVersion: version,
 		},
 	},
+	SwiftV2PrefixOnNic: true,
```
QxBytes, line 142: could you check conversion_test.go passes when OS=windows-- this request is the expected for one of the tests but in the test we set swiftv2 to false

cns/restserver/internalapi.go
```diff
@@ -229,12 +229,12 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
-	// Get IMDS NC versions for delegated NIC scenarios
-	imdsNCVersions, err := service.GetIMDSNCs(ctx)
-	if err != nil {
-		// If any of the NMA API check calls, imds calls fails assume that nma build doesn't have the latest changes and create empty map
-		imdsNCVersions = make(map[string]string)
-	}
+	// Get IMDS NC versions for delegated NIC scenarios.
+	imdsNCVersions, infraNicMacAddress := service.getIMDSNCs(ctx)
+
+	// Process Windows registry keys with the retrieved MAC address (empty if not found). It is required for HNS team to configure cilium routes sepcific to windows nodes.
```
QxBytes: nit: cilium routes "specific" to windows nodes and nit: isPrefixOnNicSwiftV2() (capital "O" for camel case or isPrefixOnNICSwiftV2() for capitalized acronyms)

cns/restserver/internalapi.go
```diff
@@ -229,12 +229,12 @@ func (service *HTTPRestService) syncHostNCVersion(ctx context.Context, channelMo
...
+	isPrefixOnNic := service.isPrefixonNicSwiftV2()
+	service.processWindowsRegistryKeys(isPrefixOnNic, infraNicMacAddress)
```
QxBytes: can you confirm that it's intentional that this runs even for non-prefix on nic scenarios?

cns/restserver/internalapi.go
```diff
@@ -691,31 +691,31 @@ func (service *HTTPRestService) isNCDetailsAPIExists(ctx context.Context) bool {
 	return false
 }
 
-// GetIMDSNCs gets NC versions from IMDS and returns them as a map
-func (service *HTTPRestService) GetIMDSNCs(ctx context.Context) (map[string]string, error) {
+// GetIMDSNCs gets NC versions from IMDS and returns them as a map along with the MAC address
+func (service *HTTPRestService) getIMDSNCs(ctx context.Context) (ncs map[string]string, infraNicMacAddress string) {
```
QxBytes: since we're also getting the mac of the infra nic, maybe this should be something like getIMDSInfo now?

cns/restserver/internalapi.go
```diff
@@ -724,10 +724,24 @@ func (service *HTTPRestService) GetIMDSNCs(ctx context.Context) (map[string]stri
...
+			infraNicMacAddress = iface.MacAddress.String()
 		}
 	}
 
-	return ncs, nil
+	return ncs, infraNicMacAddress
+}
+
+// Check whether NC is SwiftV2 NIC associated NC and prefix on nic is enabled
+func (service *HTTPRestService) isPrefixonNicSwiftV2() bool {
```
QxBytes: For the comment it seems more like "check whether _any_ NC is a swiftv2 nic with prefix on nic enabled" rather than any particular NC-- when this is enabled is there typically only 1 NC (or is it multitenant?)?

cns/restserver/internalapi_windows.go
```diff
@@ -75,3 +93,86 @@ func (service *HTTPRestService) getPrimaryNICMACAddress() (string, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to set registry value '%s': %w", keyName, err)
+	}
+	fmt.Printf("[SetValue] Set %s\\%s = %v\n", registryPath, keyName, value)
+	return nil
+}
+
+// newWindowsRegistryClient creates a new Windows registry client.
+func newWindowsRegistryClient() windowsRegistryClient {
+	return &windowsRegistry{}
```
QxBytes: similar comment to the one I made about the `newLinuxRegistryClient` -- can we just return the windows registry directly from the `newRegistryClient`

QxBytes (commented):
could you also add to the pr description some background on prefix on nic (pon, particularly v6-- what PON is meant for/aims to accomplish), what this feature is allowing us to do that we couldn't do before, and what scenarios this affects (ex: standalone, multi/single tenant)

cns/restserver/internalapi_linux.go
rbtr: @NihaNallappagari I don't think you have understood my earlier feedback on OS abstraction/separations.

A func called `newLinuxRegistryClient` returning struct `linuxRegistryClient` as interface `windowsRegistryClient` is nonsensical. Stubbing _Windows only behavior_ in a Linux struct so that the special Windows code compiles on Linux is bad. This abstraction is **wrong**. 

If you need to make stubs like this, the _place that you are calling these stubs is OS specific code already_. THAT is what should be refactored in to Windows and Linux versions.

rbtr (dismissed):

cns/restserver/restserver.go
NihaNallappagari: Latest code doesn't have this change, we can ignore this

cns/restserver/internalapi.go
NihaNallappagari: Updated the comment, let me know if it requires more context

cns/kubecontroller/nodenetworkconfig/conversion_windows_test.go
NihaNallappagari: Good catch! There were some test failures in Windows mode, which I’ve now fixed. I did run the tests on Windows, but I might have missed this while switching between changes. I was under the assumption that any test failures would fail the pipeline, so I’m wondering why this wasn’t caught.

cns/restserver/internalapi.go
NihaNallappagari: Fixed

cns/restserver/internalapi.go
NihaNallappagari: It is required only for prefix on nic scenario. Fixed in latest code changes

cns/restserver/internalapi.go
NihaNallappagari: Updated the function names

cns/restserver/internalapi_windows.go
NihaNallappagari: Latest code changes removed this all together

cns/restserver/internalapi.go
NihaNallappagari: @rbtr I’ve refactored the code based on your suggestions. Please review it and let me know if any of the changes don’t fully address your feedback. Thanks

cns/restserver/internalapi_windows.go
```diff
@@ -75,3 +90,125 @@ func (service *HTTPRestService) getPrimaryNICMACAddress() (string, error) {
...
+				logger.Errorf("[Windows] Failed to get interface name from MAC address to set for windows registry: %v", err)
+				return
+			}
+			// Process Windows registry keys with the retrieved MAC address and interface name. It is required for HNS team to configure cilium routes specific to windows nodes
+			if err := service.setRegistryKeysForPrefixOnNic(isSwiftv2PrefixOnNic, infraNicMacAddress, infraNicIfName); err != nil {
+				//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
+				logger.Errorf("[Windows] Failed to set registry keys: %v", err)
+				return
+			}
+		}()
```
rbtr: is it safe for this to be async? what if it happens much much later than the NCs that we return begin to be used?

rbtr (commented):
latest updates to the OS abstraction lgtm. only one question

cns/restserver/internalapi_windows.go
NihaNallappagari: I initially thought the keys were needed to send data to Cilium/CNI to configure routes, which happens later in the process. However, there is a risk incorrect details could be sent to Cilium. I updated this to a synchronous approach to ensure correct ordering. Please let me know if you think there’s a more optimal way to handle this.

NihaNallappagari: @rbtr Could you please review the changes and let me know if any further updates are needed? Thanks

cns/restserver/internalapi.go
```diff
@@ -691,43 +686,30 @@ func (service *HTTPRestService) isNCDetailsAPIExists(ctx context.Context) bool {
 	return false
 }
 
-// GetIMDSNCs gets NC versions from IMDS and returns them as a map
-func (service *HTTPRestService) GetIMDSNCs(ctx context.Context) (map[string]string, error) {
+// gets nc details from IMDS and returns perform platform specific action
```
QxBytes: could we rewrite this comment? Like what does the returned map[string]string represent, and also add comments to the linux and windows specific implementations-- it's a bit difficult to tell what they are doing at a glance (seems to map _only_ secondary nic ncids -> prefix on nic version)

Also there are side effects of this method on windows side it seems-- I initially thought this was just getting some data/transforming it but it also sets registry keys-- can this be documented either here or in processIMDSData-- I would also say the name of this function is still not quite accurate since we do this registry modification.

cns/restserver/internalapi_windows.go
```diff
@@ -75,3 +90,122 @@ func (service *HTTPRestService) getPrimaryNICMACAddress() (string, error) {
...
+		// Get the interface name from the MAC address
+		infraNicIfName, err := service.getInterfaceNameFromMAC(infraNicMacAddress)
+		if err != nil {
+			//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
+			logger.Errorf("Failed to get interface name from MAC address to set for windows registry: %v", err)
+		} else {
+			// Process Windows registry keys with the retrieved MAC address and interface name. It is required for HNS team to configure cilium routes specific to windows nodes
+			if err := service.setRegistryKeysForPrefixOnNic(isSwiftv2PrefixOnNic, infraNicMacAddress, infraNicIfName); err != nil {
+				//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
+				logger.Errorf("Failed to set windows registry keys: %v", err)
```
QxBytes: is it okay that we don't set the registry keys? If this fails it'll be a silent fail since the error doesn't bubble up-- same thing with getInetrfaceNameFromMAC-- should it bubble up if there is an error?

NihaNallappagari: I had the same question for the HNS team, as I wasn’t sure what was expected in the Windows workflow. I waited for their response, and they confirmed that the flow can be stopped if at least one registry key fails to be set. I’ve updated the code accordingly.

cns/restserver/internalapi.go
NihaNallappagari: I agree with the confusion and have updated the code to clarify the flow. Please let me know if anything is unclear.

NihaNallappagari:
> also note that we have external dependencies in our unit tests (registry and getting interface by mac) which might cause the same test to fail on one computer but succeed on another
@QxBytes  Could you please elaborate on which test case you’re referring to? I needed to introduce dependency injection for the interface MAC address, which required production code changes. To handle this, I refactored the code and added test scenarios that don’t require mocking NetIO. The runners will cover the end-to-end testing.

QxBytes:
> > also note that we have external dependencies in our unit tests (registry and getting interface by mac) which might cause the same test to fail on one computer but succeed on another
> > @QxBytes  Could you please elaborate on which test case you’re referring to? I needed to introduce dependency injection for the interface MAC address, which required production code changes. To handle this, I refactored the code and added test scenarios that don’t require mocking NetIO. The runners will cover the end-to-end testing.

I believe you fixed this in your most recent push

QxBytes:
niha to rename getIMDSNCDetails to better reflect that it has side-effects as per side discussion

NihaNallappagari:
These changes are no longer needed. We are exploring swiftv1 path

PR #4069, opened 2025-10-08
Fix cns zap logger issue

paulyufan2:
**Reason for Change**:

Customers found CNS crash sometimes due to cns zap logging on RequestIPConfigsHander() which defers service.publishIPStateMetrics() → every request ends by signaling the recorder:

```
fatal error: unexpected signal during runtime execution
[signal 0xc0000005 code=0x1 addr=0x11 pc=0xb92bb6]  // Windows 
...
go.uber.org/zap/internal/stacktrace.Capture
go.uber.org/zap.(*Logger).check
go.uber.org/zap.(*Logger).Info
github.com/.../cns/restserver.(*asyncMetricsRecorder).record
github.com/.../cns/restserver.(*HTTPRestService).publishIPStateMetrics.func1
fatal error: unexpected signal during runtime execution
[signal 0xc0000005 code=0x1 addr=0x11 pc=0xb92bb6]  // Windows vilation
...
go.uber.org/zap/internal/stacktrace.Capture
go.uber.org/zap.(*Logger).check
go.uber.org/zap.(*Logger).Info
github.com/.../cns/restserver.(*asyncMetricsRecorder).record
github.com/.../cns/restserver.(*HTTPRestService).publishIPStateMetrics.func1
...
restserver.(*asyncMetricsRecorder).record
```

publishIPStateMetrics() spins go recorder.run() (once) and sends on recorder.sig.
run() calls record(), which logs the pool watermarks via:  logger.Printf("Allocated IPs: %d, ...", ...)

The crash stack shows it dies inside zap while capturing a stack trace. So zap's stacktrace capture on an Info log that fires immediately after handler returns, causing windows access violation.  On windows, stacktrace.Capture() can crash under load, so disabling it could remove the crash path.
CNS zap’s stacktrace.Capture() uses runtime.Callers, and we’ve seen rare access violations in that path when the stacktrace capture is aggressive. As the result, capturing stacktraces on hot Info paths can crash the process in your environment.

In CNS metrics.go, the sig is unbuffered:
recorder.sig = make(chan struct{})

When send with a non-blocking select dropping the signal if run() is not parked on receive. An unbuffered send succeeds only if a receiver is already waiting. If run() is not parked on <-a.sig, the default branch runs and the signal could be silently dropped. So making this buffered size 1 will bring up:
1. no drops while worker is busy: if run() is in record(), a new event will queue instead of being dropped.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/logger/cnslogger.go
+++ b/cns/logger/cnslogger.go
@@ -45,11 +45,11 @@ func New(fileName string, logLevel, logTarget int, logDir string) (loggershim, e
 	if err != nil {
 		l.Errorf("Failed to get zap Platform cores: %v", err)
 	}
-	
-    zapLogger := zap.New(
+
+	zapLogger := zap.New(
 		platformCore,
 		zap.AddCaller(),
-		zap.AddStacktrace(zapcore.WarnLevel), 
+		zap.AddStacktrace(zapcore.WarnLevel),
 	).With(zap.Int("pid", os.Getpid()))
 
 	return &logger{


cns/logger/cnslogger.go
```diff
@@ -45,7 +45,12 @@ func New(fileName string, logLevel, logTarget int, logDir string) (loggershim, e
 	if err != nil {
 		l.Errorf("Failed to get zap Platform cores: %v", err)
 	}
-	zapLogger := zap.New(platformCore, zap.AddCaller()).With(zap.Int("pid", os.Getpid()))
+
+	zapLogger := zap.New(
+		platformCore,
+		zap.AddCaller(),
+		zap.AddStacktrace(zapcore.WarnLevel),
```
rbtr, line 52: why Warn? imo we don't need stacks until Error+

cns/restserver/metrics.go
```diff
@@ -219,7 +219,7 @@ func (a *asyncMetricsRecorder) record() {
 func (service *HTTPRestService) publishIPStateMetrics() {
 	recorder.once.Do(func() {
 		recorder.podIPConfigSrc = service.PodIPConfigStates
-		recorder.sig = make(chan struct{})
+		recorder.sig = make(chan struct{}, 1)
```
rbtr, line 222: disagree with this change, unless this can't be called more than twice concurrently (and if that was the case, why twice instead of once and leave this unbuffered?). this smells and indicates some other structural problem.
all it does is _delay_ when you will get a blocking send by one more event. maybe sig should only have non-blocking sends?

rbtr (commented):
these changes seem superficial, so I'm confused about what the underlying issue is here. from the description it sounds like sometimes on Windows getting a stacktrace panics? but if we're getting a stacktrace I expect we're in a catastrophic failure already?
I don't think buffering a channel is going to help resolve that.

PR #4070, opened 2025-10-08
Add status to mtpnc crd

aggarwal0009:
**Reason for Change**:
Add Status to MTPNC crd

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:


crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -61,6 +61,23 @@ const (
 	Failed        InfinibandStatus = "Failed"
 )
 
+// MTPNCStatus indicates the high-level status of MultitenantPodNetworkConfig
+// +kubebuilder:validation:Enum=Ready;InternalError;PodNotFound;PodFailed;NodeNameMissing;NodeUnreachable;PNINotFound;PNINotReady;CreateNCFailed;GetReservationSetFailed
+type MTPNCStatus string
+
+const (
+	MTPNCStatusReady                MTPNCStatus = "Ready"
```
kmurudi, line 71: do we need a distinct status for all possible errors, curious if there a way to get it from pod describe or other crds? if some of these are covered in pod describe we could avoid adding more status fields here

aggarwal0009: Yes but in that case we need to map it to a status on MTPNC. Maybe we can club these statuses under a single status

aggarwal0009: ok I have come up with the following statuses which will be useful for metric and also provide customer with actionable status without providing internal details 

const (
	MTPNCStatusReady                	MTPNCStatus = "Ready"
	MTPNCStatusPending              	MTPNCStatus = "Pending"
	MTPNCStatusInternalError        	MTPNCStatus = "InternalError"
	MTPNCStatusNodeLinkedToAnotherMTPNC     MTPNCStatus = "NodeLinkedToAnotherMTPNC"
	MTPNCStatusNICCapacityExceeded 		MTPNCStatus = "NICCapacityExceeded"
	MTPNCSubnetIPExhaustionError 		MTPNCStatus = "SubnetIPExhaustionError"
)

where Pending could be any of the below. Here MTPNC is waiting for some event(node to be available, pod to be scheduled, PNI to be ready etc) rather than failing to get provisioned
	MTPNCStatusPodNotFound          MTPNCStatus = "PodNotFound"
	MTPNCStatusPodFailed            MTPNCStatus = "PodFailed"
	MTPNCStatusNodeNameMissing      MTPNCStatus = "NodeNameMissing"
	MTPNCStatusNodeUnreachable      MTPNCStatus = "NodeUnreachable"
	MTPNCStatusPNINotFound          MTPNCStatus = "PNINotFound"
	MTPNCStatusPNINotReady          MTPNCStatus = "PNINotReady"

 Internal Error -> createNCError, getReservationSet error, any other error

Below are the statuses we expect to be customer actionable
MTPNCStatusNodeLinkedToAnotherMTPNC     MTPNCStatus = "NodeLinkedToAnotherMTPNC"
MTPNCStatusNICCapacityExceeded 		MTPNCStatus = "NICCapacityExceeded"
MTPNCSubnetIPExhaustionError 		MTPNCStatus = "SubnetIPExhaustionError"

Need feedback re. 
1) naming of above three statuses
2) If these are enough from metrics/ latency point of view

aggarwal0009: cc @miguelgoms

isaac-dasan: I would also rename NodeNameMissing to PodNotScheduled so it makes more sense to external customer

aggarwal0009: I have clubbed it under Pending status

kmurudi: does PodFailed mean NC failed to provision?
+1 to nodelinked status - could be invalidNode, bad request or scheduling error on cx side

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -61,6 +61,19 @@ const (
...
+// +kubebuilder:validation:Enum=Ready;Pending;InternalError;NodeCapacityExceeded;SubnetExhaustionError;NodeLinkedToAnotherMTPNC
+type MTPNCStatus string
+
+const (
+	MTPNCStatusReady                    MTPNCStatus = "Ready"
+	MTPNCStatusPending                  MTPNCStatus = "Pending"
+	MTPNCStatusInternalError            MTPNCStatus = "InternalError"
+	MTPNCStatusNodeCapacityExceeded     MTPNCStatus = "NodeCapacityExceeded"
+	MTPNCStatusSubnetExhaustionError    MTPNCStatus = "SubnetExhaustionError"
+	MTPNCStatusNodeLinkedToAnotherMTPNC MTPNCStatus = "NodeLinkedToAnotherMTPNC"
```
isaac-dasan: Did we intend to tell NIC is linked to another MTPNC or Node is linked? We support multiple NICs per node and it doesn't make sense to report NodeLinkedToAnotherMTPNC when it is expected to happen.

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
aggarwal0009: Pls check the updated enums

aggarwal0009:
> Can you please explain the requirement for these individual statuses? Like for example `pending` - what is the use case for this? `subnetExhaustion` doesn't make sense in MTPNC. It should have been caught in PNI and it's probably more appropriate to say something like- IP exhausted for PNI.

@isaac-dasan I will update the 1 pager and link it here for requirements.

subnetExhaustion is for dynamic allocation not static/explicit allocation / pni related error

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
aggarwal0009: Removing this as discussed

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
miguelgoms: as discussed offline: 
- Lets bubble up user-actionable statuses: PNI Not found / Not ready
- Remove NodeLinktedToAnotherMTPNC - this check will change, and is ultimately a flavor of "NIC Capacity Exceeded" 
- Pending - to clarify, intended only for transient errors before pod is effectively scheduled. 

PodFailed - we could rename this since this is rather specific. this does not mean NC failed provision but that we found the pod was already on Failed state on the first reconcile. the intent here is to avoid a pod which terminated/failed early from taking over a NIC for no reason until deletion.

miguelgoms: we could skip PodFailed as a status - and leave this as either Pending (we just do not update status on this error, before the mtpnc gets deleted with the pod), or we can mark it as InternalError.

aggarwal0009:
> pending update of enums to separate user-actionable errors.

@miguelgoms  Added

isaac-dasan (approved):

PR #4074, opened 2025-10-13
CNS Change for Subnet Overlay Expansion Job

rejain456:
**Reason for Change**:

Without this CNS change, expanding an overlay subnet causes the CNS daemonset to crash, resulting in a PrimaryCANotSame error for all clusters. 
To enable overlay subnet expansion:

- only for overlay clusters, we bypass the PrimaryCANotSame error. 
- The subnet address in the NNC is updated.
- The CNS state file gets updated to reflect the new pod cidr from NNC

Simply restarting CNS does not regenerate the state file with updated information. Instead, this pr will update the state file to have the new subnet address (same as the one from updated nnc) which will prevent the error only for overlay clusters when there is a mismatch of pod cidrs to start with.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/internalapi.go
+++ b/cns/restserver/internalapi.go
@@ -746,8 +746,8 @@ func validateCIDRSuperset(newCIDR, oldCIDR string) bool {
 		return false
 	}
 
-	// Condition 2: Check if the base IP of oldCIDR is contained in newCIDR
-	if !newPrefix.Contains(oldPrefix.Addr()) {
+	// Condition 2: Check for Overlap - this will also ensure containment
+	if !newPrefix.Overlaps(oldPrefix) {
 		return false
 	}
 
--- a/cns/restserver/internalapi_test.go
+++ b/cns/restserver/internalapi_test.go
@@ -133,6 +133,7 @@ func TestReconcileNCStatePrimaryIPChangeShouldNotFail(t *testing.T) {
 		{"10.0.1.0/24", "10.0.2.0/22"},
 		{"10.0.1.0/20", "10.0.1.0/18"},
 		{"10.0.1.0/19", "10.0.0.0/15"},
+		{"10.0.1.0/18", "10.0.1.0/18"},
 	}
 
 	// Run test cases


cns/restserver/internalapi.go
```diff
@@ -636,7 +640,18 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
 				existingReq.IPConfiguration.IPSubnet.PrefixLength,
 				req.IPConfiguration.IPSubnet.IPAddress,
 				req.IPConfiguration.IPSubnet.PrefixLength)
-			return types.PrimaryCANotSame
+			// delete the existing azure-cns file
+			if err := os.Remove(cnsStateFilePath); err != nil && !os.IsNotExist(err) {
```
nddq: when CNI stateless is running, CNS manages all of IPAM state (ip allocation, ...) in its statefile, so this approach would not work.

cns/restserver/internalapi.go
```diff
@@ -636,7 +640,18 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
+				log.Error("Error deleting azure-cns state file", zap.Error(err))
+			}
+
+			// panic (this will force azure-cns daemonset to restart and pick up the new state with updated nnc)
+			panic(fmt.Sprintf("PrimaryCA mismatch for NCId %s: old CA %s/%d, new CA %s/%d",
```
nddq: Is there no other way for us to communicate to CNS that the overlay subnet is expanding so that CNS can handle this error accordingly without telling CNS to kill itself?

cns/restserver/internalapi.go
rbtr: stateless cni (all windows nodes, linux in 1.34+) and any cilium CNI also

cns/restserver/internalapi.go
rbtr: i think the statefile should be updateable, but i'm not sure how embedded in the various subcomponents state from the initial read of the statefile is. we don't have any hotreload support and that would be a major rearchitecture... so we might end up exiting to be restarted regardless, but not as a panic

cns/restserver/internalapi.go
```diff
@@ -630,13 +630,23 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
 		if !reflect.DeepEqual(existingReq.IPConfiguration.IPSubnet, req.IPConfiguration.IPSubnet) {
-			logger.Errorf("[Azure CNS] Error. PrimaryCA is not same, NCId %s, old CA %s/%d, new CA %s/%d",
-				req.NetworkContainerid,
-				existingReq.IPConfiguration.IPSubnet.IPAddress,
-				existingReq.IPConfiguration.IPSubnet.PrefixLength,
-				req.IPConfiguration.IPSubnet.IPAddress,
-				req.IPConfiguration.IPSubnet.PrefixLength)
-			return types.PrimaryCANotSame
+
+			logger.Debugf("Updating PrimaryCA for NCId %s", req.NetworkContainerid)
```
nairashu: You might get lint failures as logger.Debugf is deprecated

nairashu: Also add in that the CIDR range has been updated in the log line. Should we also limit this check to only the Overlay scenario by adding the condition to the if block on line 632?

cns/restserver/internalapi.go
```diff
@@ -630,13 +630,23 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
+			// Update the in-memory state
+			service.Lock()
+			ncInfo := service.state.ContainerStatus[req.NetworkContainerid]
+			ncInfo.CreateNetworkContainerRequest.IPConfiguration.IPSubnet = req.IPConfiguration.IPSubnet
+			service.state.ContainerStatus[req.NetworkContainerid] = ncInfo
+			service.Unlock()
+
+			// Persist the updated state to disk
+			if err := service.saveState(); err != nil {
+				logger.Errorf("Failed to save updated CNS state: %v", err)
```
nairashu: Possible lint failure

cns/restserver/internalapi.go
```diff
@@ -630,13 +630,23 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
...
+
+			// Update the in-memory state
+			service.Lock()
+			ncInfo := service.state.ContainerStatus[req.NetworkContainerid]
+			ncInfo.CreateNetworkContainerRequest.IPConfiguration.IPSubnet = req.IPConfiguration.IPSubnet
+			service.state.ContainerStatus[req.NetworkContainerid] = ncInfo
+			service.Unlock()
+
+			// Persist the updated state to disk
+			if err := service.saveState(); err != nil {
```
nairashu: Should we retry a couple of times to save the stateFile to avoid benign errors before failing out?

nairashu:
Just got notified that this PR is replaced with the PR here: https://github.com/Azure/azure-container-networking/pull/4083/files

cns/restserver/internalapi.go
rejain789: not relevant anymore, removed that logic

cns/restserver/internalapi.go
rejain789: the update state file logic is in saveNetworkContainer function call and if its an overlay cluster, that call will be made on it so removed all logic for update statefile as it is already taking place

cns/restserver/internalapi.go
```diff
@@ -35,6 +36,9 @@ const (
 // internal APIs (definde in internalapi.go).
 // This will be used internally (say by RequestController in case of AKS)
 
+// Initialize a zap logger instance
+var zapLogger, _ = zap.NewProduction()
```
rbtr: why creating a new logger here?

cns/restserver/internalapi.go
```diff
@@ -722,3 +727,29 @@ func (service *HTTPRestService) GetIMDSNCs(ctx context.Context) (map[string]stri
...
+	}
+
+	// Condition 1: Check if the new prefix length is smaller (larger range) than the old prefix length
+	if newPrefix.Bits() >= oldPrefix.Bits() {
+		return false
+	}
+
+	// Condition 2: Check if the base IP of oldCIDR is contained in newCIDR
+	if !newPrefix.Contains(oldPrefix.Addr()) {
+		return false
```
nairashu, line 751: Full containment vs base address only

Currently, it only checks if the starting IP of oldCIDR is inside newCIDR.
If oldCIDR spans multiple IPs, this doesn’t guarantee full containment.


```suggestion

	    // Check full containment: start and end of oldCIDR inside newCIDR
    oldStart := oldPrefix.Masked().Addr()
    oldEnd := oldStart.Next().Next().Prev() // compute last IP in oldCIDR
    return newPrefix.Contains(oldStart) && newPrefix.Contains(oldEnd)
```

nairashu (changes requested):
Let's verify full CIDR containment. Rest looks good.

cns/restserver/internalapi.go
rbtr: I don't think it is possible that subnets can partially overlap, because they must be aligned on a pow2.
If we think about this bitwise, it is impossible for an IP to belong to multiple prefixes which only partially overlap - the masked bits must match, so the subnets would necessarily fully overlap, right?

@nairashu do you have a counter example?

rbtr: actually I think this is what [`netip.Prefix#Overlaps`](https://cs.opensource.google/go/go/+/refs/tags/go1.25.3:src/net/netip/netip.go;l=1453-1489) does (and @rejain456 we should just use it) - 
It reduces both subnets to their minimum fixed bits and checks to see if they're equal. Because if so, the larger subnet MUST fully overlap the smaller subnet (or they are exactly equal), because it's a bitmask, and everything in 
`1111xxxxx` is also in 
`111xxxxxx`.

rejain789: @rbtr, updated the code
I agree that with the prefix length check and the overlap check, we can check for containment. 

I want to confirm that we do not to explicitly call oldPrefix.Masked() and newPrefix.Masked() before calling function Overlaps(). 

Example: Since Overlaps() internally normalizes both prefixes to the shorter prefix length for comparison, it should handle cases like 10.0.0.0/8 vs 10.0.5.200/16 by masking both to /8 and correctly returning true.

However, could there be any edge cases to explicitly mask beforehand, or is relying on Overlaps() internal masking sufficient for CIDR containment checks?
In other words, wanted to confirm an address like 10.0.5.200/16 will be normalized to its network boundary (e.g. 10.0.5.0/16) and not treated as a partial subnet starting at 10.0.5.200 so that the containment check does not fail due to misaligned addresses

rbtr: yes, it looks like `Overlaps` normalizes to the network address even if the CIDR passed uses a different base address so it is sufficient alone

rbtr (approved):

PR #4078, opened 2025-10-14
fix: Adding delete timestamp check

shreyashastantram:
**Reason for Change**:

Adding check to MTPNC to ensure that CNS doesn't respond back with the ipaddress if mtpnc is set for deletion.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/middlewares/k8sSwiftV2_linux_test.go
+++ b/cns/middlewares/k8sSwiftV2_linux_test.go
@@ -226,7 +226,7 @@ func TestValidateMultitenantIPConfigsRequestFailure(t *testing.T) {
 	failReq.OrchestratorContext = b
 	_, respCode, msg = middleware.GetPodInfoForIPConfigsRequest(context.TODO(), failReq)
 	assert.Equal(t, respCode, types.UnexpectedError)
-	assert.Assert(t, strings.Contains(msg, "MTPNC in terminating state"), "expected error message to contain '%s', got '%s'", "MTPNC in terminating state", msg)
+	assert.Assert(t, strings.Contains(msg, NetworkNotReadyErrorMsg), "expected error message to contain '%s', got '%s'", NetworkNotReadyErrorMsg, msg)
 }
 
 func TestGetSWIFTv2IPConfigSuccess(t *testing.T) {


cns/middlewares/k8sSwiftV2.go
```diff
@@ -28,6 +28,7 @@ var (
 	errGetMTPNC                 = errors.New(NetworkNotReadyErrorMsg + " - failed to get MTPNC")
 	errInvalidSWIFTv2NICType    = errors.New("invalid NIC type for SWIFT v2 scenario")
 	errInvalidMTPNCPrefixLength = errors.New("invalid prefix length for MTPNC primaryIP, must be 32")
+	errDeleteTimestampSet       = errors.New("MTPNC deletion timestamp is set")
```
miguelgoms: these are still user-visible messages as this will be the error shown after "network not ready" to the cx on CNI
message should be more meaningful to the Cx - e.g.: "found mtpnc for previous pod pending deletion, waiting for new mtpnc to be ready"

miguelgoms: arguably we could just return the "mtpncnotready" - but since the old mtpnc object is still visible, I think its better to have the separate error message.

cns/middlewares/k8sSwiftV2.go
```diff
@@ -193,6 +194,9 @@ func (k *K8sSWIFTv2Middleware) getMTPNC(ctx context.Context, podInfo cns.PodInfo
 	if err := k.Cli.Get(ctx, mtpncNamespacedName, &mtpnc); err != nil {
 		return v1alpha1.MultitenantPodNetworkConfig{}, types.UnexpectedError, errors.Wrap(err, errGetMTPNC.Error()).Error()
 	}
+	if mtpnc.DeletionTimestamp != nil {
```
ashvindeodhar: !IsZero() seems to be the way to check this. DeletionTimestamp is Time.  
it may be helpful to add a func something like IsTerminating similar to IsReady that we already have on mtpnc. 

I also feel we shouldn't be adding these checks inside getMTPNC func. This func should return the MTPNC object and the caller should look at it and decide how to proceed however it wants. (different callers may call getMTPNC and use different fields to decide what to do next)

shreyashastantram: I can add a function similar to isready, but we cannot use isZero since the data will be nil if it's not populated with timestamp. it's either nil or timestamp.

ashvindeodhar: if the time.Time is absent, it will default to zero and IsZero() will be true

cns/middlewares/k8sSwiftV2.go
rbtr: deletion timestamp being set is an implementation detail. the important thing is that the object is deleting.

cns/middlewares/k8sSwiftV2.go
rbtr: +1 @ashvindeodhar is right, this doesn't belong in the `getMTPNC`. Return the MTPNC and let the caller decide what to do if it's Deleting.

`.DeletionTimestamp.IsZero()` is canonically the correct way to determine if an Object is deleting. ref https://book.kubebuilder.io/reference/using-finalizers.html

rbtr (dismissed):

cns/middlewares/k8sSwiftV2.go
shreyashastantram: will change it to mtpnc terminating

cns/middlewares/k8sSwiftV2.go
isaac-dasan: nit: I would recommend a rename of isTerminating -> isDeleting everywhere (++ errMessage)
isDeleting ⇒ metadata.deletionTimestamp != nil
isTerminating ⇒ your CR’s .status.phase == "Terminating" (or whatever phase name your API defines) which doesn't apply to MTPNC

cns/middlewares/k8sSwiftV2.go
miguelgoms: this still applies to "mtpnc is terminating" - its not untrue, but it does not tell you why this is treated as a transient "network not ready" instead of a terminal error.

cns/middlewares/k8sSwiftV2.go
shreyashastantram: I do not see deleting being used anywhere in Kubernetes, keeping it, terminating will keep it consistent with the pod state terminating.

cns/middlewares/k8sSwiftV2.go
shreyashastantram: With that change, we are tagging this error with stateful set pod creation, if  user removes ownership ref and tries deleting mtpnc for some reason, pod restart would result in the same error which is not the case. Yes, it's not the expected behavior but if user can do it, should we not consider that case? I want to keep it general so that it doesn't change with context

cns/middlewares/k8sSwiftV2.go
isaac-dasan: The pod status changes to Terminating until it’s fully removed,
- Sends SIGTERM to containers.
- Waits for the termination grace period (default 30s).
- Runs any presto hooks.
- Finally sends SIGKILL if containers don’t exit in time.

Unlike pod (where there is an actual termination phase), MTPNC doesn't have a termination phase. So, there is no need for us to keep it consistent with pod status. We should keep the name to the face value of the operation, since the errMessage is something, customers can see.

cns/middlewares/k8sSwiftV2.go
miguelgoms: if user removes ownership and tries deleting before pod starts they'll get this error - which tells them this will not succeed without a new mtpnc on ready state. and if the mtpnc completes delete on the next CNI restart they'll get "mtpnc not ready"

cns/middlewares/k8sSwiftV2.go
```diff
@@ -28,6 +28,7 @@ var (
 	errGetMTPNC                 = errors.New(NetworkNotReadyErrorMsg + " - failed to get MTPNC")
 	errInvalidSWIFTv2NICType    = errors.New("invalid NIC type for SWIFT v2 scenario")
 	errInvalidMTPNCPrefixLength = errors.New("invalid prefix length for MTPNC primaryIP, must be 32")
+	errMTPNCTerminating         = errors.New("MTPNC in terminating state")
```
isaac-dasan: Update to deleting && add `NetworkNotReadyErrorMsg`
Without NetworkNotReadyErrorMsg kubelet won't retry faster for the new pod. You can check my PR when I added `NetworkNotReadyErrorMsg` for the reasoning behind this.

isaac-dasan (changes requested):

cns/middlewares/k8sSwiftV2.go
```diff
@@ -28,6 +28,7 @@ var (
 	errGetMTPNC                 = errors.New(NetworkNotReadyErrorMsg + " - failed to get MTPNC")
 	errInvalidSWIFTv2NICType    = errors.New("invalid NIC type for SWIFT v2 scenario")
 	errInvalidMTPNCPrefixLength = errors.New("invalid prefix length for MTPNC primaryIP, must be 32")
+	errMTPNCDeleting            = errors.New(NetworkNotReadyErrorMsg + "found mtpnc for previous pod pending deletion, waiting for new mtpnc to be ready")
```
isaac-dasan: You are making assumptions that this err msg will only be thrown if the previous mtpnc is deleting. It could also be the case that the current mtpnc is deleting. So i would keep the error message simple - `NetworkNotReadyErrorMsg + " - mtpnc is deleting"`

shreyashastantram: https://github.com/Azure/azure-container-networking/pull/4078#discussion_r2433305530 Why would we see this error when current mtpnc is deleting? can you elaborate ?

miguelgoms: IFF mtpnc is marked for deletion before the pod start, we *are* assuming this mtpnc is not the current one and failing by design - and we are asking CNI to wait + retry until a new mtpnc is ready. 
IFF mtpnc was incorrectly deleted - the pod will fail until a new mtpnc shows up that is valid and ready.

miguelgoms (approved):

crd/multitenancy/api/v1alpha1/utils.go
```diff
@@ -9,3 +9,7 @@ func (m *MultitenantPodNetworkConfig) IsReady() bool {
 	// Check if InterfaceInfos slice is not empty
 	return !reflect.DeepEqual(m.Status, MultitenantPodNetworkConfigStatus{})
 }
+
```
ashvindeodhar, line 12: IMO we should always add the info on what the func is doing (however simple the func is). There is an argument that for simple functions we can skip it but then the definition of 'what's a simple func' is subjective.

cns/middlewares/k8sSwiftV2.go
isaac-dasan: Why do we need to tell all this background to customer? what value are we providing here? The error message should just convey what went wrong. You'd be surprised how customers use the product we provide.

Here is a scenario I could think off - 
- the pod was not scheduled for a very long time, but MTPNC was created
- customer deleted MTPNC manually
- pod got scheduled to a node and now CNS is trying to allocate an IP from MTPNC

isaac-dasan: I partially agree with the previous comment from https://github.com/Azure/azure-container-networking/pull/4078#discussion_r2433305530
We cannot say "MTPNC deletion time is set" that is not actionable on meaningful. But `MTPNC is deleting` is meaningful.

ashvindeodhar: +1 to Isaac. why do we specify that 'waiting for new mtpnc to be ready'? It should just indicate mtpnc is terminating

isaac-dasan: format wise, i would like to keep the message consistent with the one above it (including hyphen and spaces between NetworkNotReadyErrormsg)

shreyashastantram: mtpnc is still from the old pod: "pod got scheduled to a node and now CNS is trying to allocate an IP from MTPNC". Comment still holds the logic correctly but question is what customer would gain out of this, I would simply put he would know why MTPNC is state of deletion. I am all for general comments but sometimes being specific does help

crd/multitenancy/api/v1alpha1/utils.go
shreyashastantram: Unless we add some Kubernetes detail which explicitly states that" For Kubernetes object, when delete timestamp is set, it's considered set for deletion". For this function just adding "IsDeleting checks if the MTPNC resource is being deleted" makes it redundant since function is self-explanatory.

isaac-dasan (dismissed):

ashvindeodhar (dismissed):

miguelgoms (dismissed):

crd/multitenancy/api/v1alpha1/utils.go
rbtr, line 1: this whole file should go away... you've replaced 1 line with 4, and it was already clear

miguelgoms (approved):

PR #4103, opened 2025-10-27
[CNS] Overlay Expansion Subnet Update Job Bug Fix

rejain789:
**Reason for Change**:
Previously, we called validateCIDRSuperset with only the IP address, which didn’t include the prefix length and resulted in incomplete CIDR notation, causing inaccurate superset validation for overlay subnet expansion.
For example, the function previously received just the IP address 10.240.0.0 instead of the full CIDR 10.240.0.0/24, which caused validateCIDRSuperset to read the logic inaccurately and led to incorrect subnet containment checks.
This happened because the IP and prefix length were stored separately in 2 different variables, and the code didn’t combine them before validation.
We considered explicitly building the CIDR string before validation. We chose to construct the full CIDR (IP/PrefixLength) for both the new and existing requests prior to calling validateCIDRSuperset, as this ensures correct subnet containment logic for subnet overlay expansion with minimal code change.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rejain789:

rejain789:
@microsoft-github-policy-service agree

PR #4109, opened 2025-11-05
ci: clean up disk space before build and skip windows port forward

QxBytes:
**Reason for Change**:

Previously, building npm images would cause the agent to run out of disk space. This PR adds a step before building the image which cleans up unneeded files (from: https://github.com/cilium/cilium/blob/v1.16.8/.github/actions/disk-cleanup/action.yaml).

Also splits npm into its own stage and displays disk usage before and after disk cleanup and image build.

Also skips windows port forward tests

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Run: https://msazure.visualstudio.com/One/_build/results?buildId=142753721&view=results

### DIFF
--- a/.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
+++ b/.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
@@ -71,7 +71,7 @@ stages:
           dependsOn: ${{ parameters.name }}
           datapath: true
           dns: true
-          portforward: true
+          portforward: ${{ eq(parameters.os, 'linux') }} # Unblock Pipeline, broken for all windows scenarios
           hostport: true
           service: true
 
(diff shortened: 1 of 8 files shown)


.pipelines/containers/container-template.yaml
```diff
@@ -15,11 +15,38 @@ steps:
...
+      /usr/local/julia* \
+      /usr/local/lib/android \
+      /usr/local/lib/node_modules \
+      /usr/local/share/chromium \
+      /usr/local/share/powershell \
+      /usr/share/dotnet \
+      /usr/share/swift
+    echo "Disk space after cleanup..."
+    df -h /
+  displayName: "Clean up disk space"
```
jpayne3506, line 37: nit: this could be a template

rayaisaiah (approved):

PR #4122, opened 2025-11-17
feat: add support for windows MAC hex dump

ecigar13:
**Reason for Change**:
The `net.ParseMAC` function does not support windows MAC address, which is a hex dump without delimiter, so we make a wrapper to add colon `:` delimiters to the MAC address before parsing if it doesn't have any delimiter.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/imds/client_test.go
+++ b/cns/imds/client_test.go
@@ -165,17 +165,15 @@ func TestGetNetworkInterfacesInvalidMAC(t *testing.T) {
         "interface": [
             {
                 "interfaceCompartmentID": "nc-12345-67890",
-                "macAddress": "00005e00530"
+                "macAddress": "00005e00530" // incorrect windows MAC address length
             },
         ]
     }`)
 
 	mockIMDSServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// request header "Metadata: true" must be present
 		metadataHeader := r.Header.Get("Metadata")
 		assert.Equal(t, "true", metadataHeader)
 
-		// verify path is network metadata
 		assert.Contains(t, r.URL.Path, "/metadata/instance/network")
 
 		w.WriteHeader(http.StatusOK)


cns/imds/client_test.go
```diff
@@ -111,7 +111,7 @@ func TestGetNetworkInterfaces(t *testing.T) {
             },
             {
                 "interfaceCompartmentID": "",
-                "macAddress": "00:00:5e:00:53:02"
+                "macAddress": "00005e005302"
```
thatmattlong, line 114: should we add a new test case instead of changing the existing one to ensure there are no regressions with the existing behavior?

ecigar13: Hi, @thatmattlong , the test is for a new behavior so we won't regress. This is the first time CNS queries to IMDS to get NIC and MAC from it, so it's the first time it encounters Windows type MAC address. The previous test was for this new behavior and this is a fix for the previous test gap:

https://github.com/Azure/azure-container-networking/commit/2f0311d4f21a06e68361b2b68a1d2e1840a420c5#diff-cfbb3c5eff132c8dd7051bffc74ddae0d3906cb35d98b7550653051e127890b9R105-R118

cns/imds/client_test.go
```diff
@@ -160,6 +160,37 @@ func TestGetNetworkInterfaces(t *testing.T) {
...
 
+func TestGetNetworkInterfacesInvalidMAC(t *testing.T) {
+	networkInterfaces := []byte(`{
+        "interface": [
+            {
+                "interfaceCompartmentID": "nc-12345-67890",
+                "macAddress": "00005e00530" // incorrect windows MAC address length
+            },
+        ]
+    }`)
```
jpayne3506, line 171: Should we also add an invalid character test in here as well to ensure both sides of the conditional are caught?

ecigar13: Hi @jpayne3506 , the `net.ParseMAC` function catches it with `xtoi2` . I only add a wrapper for Windows case.

https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/net/mac.go;l=55-74

rbtr:
@ecigar13 this is fixed upstream (we should have contributed it ourselves imo but it's done now anyway).
https://github.com/golang/go/issues/66682
https://go-review.googlesource.com/c/go/+/722720

my preference would be to abandon this wrapper and pull that instead
cc @thatmattlong @timraymond

NihaNallappagari:
> @ecigar13 this is fixed upstream (we should have contributed it ourselves imo but it's done now anyway). [golang/go#66682](https://github.com/golang/go/issues/66682) https://go-review.googlesource.com/c/go/+/722720
> 
> my preference would be to abandon this wrapper and pull that instead cc @thatmattlong @timraymond

Hello @rbtr, I’m not sure how we’re able to use this fix in the ACN repo without it being officially released. I can try testing it locally by pointing the Go path to master instead of a released version, if that’s possible.
My main question is, since this change hasn’t been included in any released Go version yet, how does the code works in canary environments and later in production? Does canary use Go master instead of a released version?
I’d appreciate any clarification or insights from anyone with more context.

PR #4136, opened 2025-11-20
fix dualnic windows duplicated routes issue

paulyufan2:
**Reason for Change**:

This PR is to fix dualnic windows duplicated route issues

Once DNS config sets skipDefaultRoute to True, then do not set default route on this interface. For example:
eth0(skipDefaultRoute=False) -> CNI sets default route to eth0 interface
eth1(skipDefaultRoute=True) -> CNI does not set default route to eth1 interface and leave Routes field as empty.

We are providing two HCN endpoints to HNS:
Endpoint 1 (eth0) – includes a default route when skipDefaultRoute=False
Endpoint 2 (eth1) – the Routes field will be added as a dummy default route to avoid HNS setting up a default route when skipDefaultRoute=True

Note that HNS will create a default route on apipa network. HNS team needs to fix this duplicated route issue as well.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/hnsclient/hnsclient_windows.go
+++ b/cns/hnsclient/hnsclient_windows.go
@@ -541,8 +541,7 @@ func configureHostNCApipaEndpoint(
 	// keep Apipa Endpoint gw as 169.254.128.1 to make sure NC to host connectivity work for both Linux and Windows containers
 	hcnRoute := hcn.Route{
 		NextHop:           hnsLoopbackAdapterIPAddress,
-		DestinationPrefix: "0.0.0.0/0",
-		Metric:            50,
+		DestinationPrefix: "169.254.128.0/17",
 	}
 
 	endpoint.Routes = append(endpoint.Routes, hcnRoute)


cni/network/multitenancy.go
```diff
@@ -159,6 +159,26 @@ func (m *Multitenancy) DetermineSnatFeatureOnHost(snatFile, nmAgentSupportedApis
...
+func (m *Multitenancy) addDefaultRoute(gwStr string, epInfo *network.EndpointInfo, result *network.InterfaceInfo) error {
+	gw := net.ParseIP(gwStr)
+	if gw == nil {
+		return fmt.Errorf("invalid gateway IP: %s", gwStr) //nolint
+	}
+
+	var dst net.IPNet
+	if gw.To4() != nil {
+		_, defaultIPNet, _ := net.ParseCIDR("0.0.0.0/0")
+		dst = net.IPNet{IP: net.IPv4zero, Mask: defaultIPNet.Mask}
```
paulyufan2, line 173: ipv6 is not supported right now

PR #4143, opened 2025-11-24
feat: add support for IP constraint

aranta-rokade:
**Reason for Change**:
Add a new optional property in preparation to support IPContstraint feature 

**Issue Fixed**:
The new property is added in preparation to support new feature to enforce an IP constraint on the reservation. A single reservation will be allocated using the new IPConstraint field.

*Testing*
1. Added UT for supported PNI crds

2. Tested older PNI compatibility on standalone:
- created PNI on standalone created with master pointing to older spec
- update the crd
- the PNIs will now have a new blank field IPConstraint: <blank>
- delete the existing PNIs

3. Tested new PNIs created:  [pni-tests.yaml](https://github.com/user-attachments/files/24951994/pni-tests.yaml)
image

tested delete old and new:
image

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/crd/multitenancy/api/v1alpha1/podnetworkinstance_test.go
+++ b/crd/multitenancy/api/v1alpha1/podnetworkinstance_test.go
@@ -14,8 +14,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 )
 
-var testEnv *envtest.Environment
-var k8sClient client.Client
+var (
+	testEnv   *envtest.Environment
+	k8sClient client.Client
+)
 
 func TestMain(m *testing.M) {
 	// Check if KUBEBUILDER_ASSETS is set
@@ -40,7 +42,7 @@ func TestMain(m *testing.M) {
 	k8sClient, _ = client.New(cfg, client.Options{Scheme: scheme})
 
 	code := m.Run()
-	testEnv.Stop()
+	_ = testEnv.Stop()
 	os.Exit(code)
 }
 


crd/multitenancy/api/v1alpha1/podnetworkinstance.go
```diff
@@ -17,6 +17,14 @@ import (
 // +kubebuilder:metadata:labels=managed=
 // +kubebuilder:metadata:labels=owner=
 // +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.status`
+//
+// Enforce immutability of .spec.
```
miguelgoms: lets remove the 'code comments' for the kubebuilder validation rule from the CRD description compiled by kubebuilder?  the consumer of the CRD doesn't need to know this is output from the tool, and the rule is already self-descriptive in the yaml.

miguelgoms: if we need to document the rule in more detail, that should be part of the message of the rule.

miguelgoms (changes requested):

crd/multitenancy/api/v1alpha1/podnetworkinstance.go
aranta-rokade: Removed the comments.

crd/multitenancy/api/v1alpha1/podnetworkinstance.go
```diff
@@ -49,6 +42,8 @@ type PodNetworkConfig struct {
 	// PodIPReservationSize is the number of IP address to statically reserve
 	// +kubebuilder:default=0
 	PodIPReservationSize int `json:"podIPReservationSize,omitempty"`
+	// IPConstraint specifies criteria for selecting IP addresses from the PodNetwork's subnet
```
miguelgoms: shouldn't this have additional kubebuilder markers? at min. // +kubebuilder:validation:Optional. 
check podnetwork for examples, https://book.kubebuilder.io/reference/markers/crd-validation.html?highlight=required#crd-validation for reference

miguelgoms: also document example format / usage.

kmurudi (dismissed):
immutable spec changes lgtm

crd/multitenancy/api/v1alpha1/podnetworkinstance.go
aranta-rokade: I had added this validation for cidr check, but the CEL validation cost was too high.
	// +kubebuilder:validation:XValidation:rule="self == '' || isIP(self) || isCIDR(self)",message="must be a valid IPv4/IPv6 address or CIDR"
	
So, I added this MaxLength:
	// +kubebuilder:validation:MaxLength=43

But the cost is still too high, copilot is suggesting to add MaxItems on PodNetworkConfig array -> but this is incorrect. 

image


adding just the Optional validation for now, which I had missed earlier

aranta-rokade: done, added the optional validation + description with example and format

crd/multitenancy/api/v1alpha1/podnetworkinstance.go
```diff
@@ -41,6 +42,15 @@ type PodNetworkConfig struct {
...
+	// Examples:
+	//   - IPv4 address: "10.0.0.1"
+	//   - IPv4 CIDR:    "10.0.0.0/24"
+	//   - IPv6 address: "2001:db8::1"
+	//   - IPv6 CIDR:    "2001:db8::/32"
+	// +kubebuilder:validation:Optional
+	IPConstraint string `json:"ipConstraint,omitempty"`
 }
 
 // PodNetworkInstanceSpec defines the desired state of PodNetworkInstance
```
miguelgoms, line 57: can you fix the deprecated comment for PodNetworkInstanceSpec.PodNetwork since we're touching this?

aranta-rokade: fixed in [enhancement(pni-crd): enforce immutable spec and add property for IP constraints by aranta-rokade ·…](https://github.com/Azure/azure-container-networking/pull/4143/commits/5cbc430bbaedfb8f011199b7bb53b00c3f1d117d)

crd/multitenancy/api/v1alpha1/podnetworkinstance.go
```diff
@@ -35,20 +36,32 @@ type PodNetworkInstanceList struct {
 }
 
 // PodNetworkConfig describes a template for how to attach a PodNetwork to a Pod
+// +kubebuilder:validation:XValidation:rule="!has(self.ipConstraint) || size(self.ipConstraint) == 0 || self.podIPReservationSize == 1",message="ipConstraint can only be specified when podIPReservationSize is 1"
 type PodNetworkConfig struct {
```
aranta-rokade: change, size check to == ''

.golangci.yml
```diff
@@ -1,6 +1,13 @@
 issues:
   max-same-issues: 0
   max-issues-per-linter: 0
+  exclude-rules:
+    - path: crd/multitenancy/api/v1alpha1/podnetwork\.go
```
aranta-rokade: make it inline

crd/multitenancy/api/v1alpha1/podnetworkinstance.go
```diff
@@ -35,20 +36,33 @@ type PodNetworkInstanceList struct {
 }
 
 // PodNetworkConfig describes a template for how to attach a PodNetwork to a Pod
+//nolint:lll
+// +kubebuilder:validation:XValidation:rule="!has(self.ipConstraint) || size(self.ipConstraint) == 0 || self.podIPReservationSize == 1",message="ipConstraint can only be specified when podIPReservationSize is 1"
```
isaac-dasan: Since ipConstraint has +kubebuilder:default="", the field is always present post-defaulting. This makes has(self.ipConstraint) redundant in the XValidation rule. Simplified the rule to validate purely on value semantics: `empty string is always allowed, and non-empty ipConstraint is only allowed when podIPReservationSize == 1`. This keeps the invariant clear and aligns with CRD defaulting behavior.

crd/multitenancy/api/v1alpha1/podnetworkinstance.go
```diff
@@ -35,20 +36,31 @@ type PodNetworkInstanceList struct {
...
+	// Must be a valid IPv4 address or CIDR notation with /32 prefix.
+	// This is an optional field.
+	// Examples:
+	//   - IPv4 address: "192.168.0.1"
+	//   - IPv4 CIDR:    "192.168.0.1/32"
+	// +kubebuilder:default=""
+	// +kubebuilder:validation:MaxLength=18
+	// +kubebuilder:validation:Pattern=`^$|^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.){3}(25[0-5]|(2[0-4]|1\d|[1-9]|)\d)(\/32)?$`
+	IPConstraint string `json:"ipConstraint,omitempty"`
 }
```
aranta-rokade, line 56: removed +kubebuilder:Optional, since omitemoty is present. (similar to podIPReservationSize)

image

aranta-rokade: ref: https://book.kubebuilder.io/reference/markers

crd/multitenancy/api/v1alpha1/podnetworkinstance.go
aranta-rokade: removed the immutability soec for now which was interfering with the deletes. The PR now only adds the support for IPConstraint. 
++ removed the `!has(self.ipConstraint) ` clause

isaac-dasan (approved):

PR #4144, opened 2025-11-24
chore: add swiftv2 windows conflist

jackieluc:
**Reason for Change**:

Add a simple default Windows SwiftV2 conflist for usage with Stateless CNI. SwiftV2 uses delegated NICs and has no need for AdditionalArgs like endpoint policies for OutBoundNAT, ROUTE, or ACL. This conflist would be bundled with Stateless CNI for Windows SwiftV2 in the official nuget package for partners to consume.

### DIFF
--- a/cni/azure-windows-swiftv2-stateless.conflist
+++ b/cni/azure-windows-swiftv2-stateless.conflist
@@ -1,23 +1,27 @@
 {
     "cniVersion": "1.0.0",
     "name": "azure",
-    "type": "azure-vnet",
-    "mode": "bridge",
-    "bridge": "azure0",
-    "capabilities": {
-        "portMappings": true,
-        "dns": true
-    },
-    "ipam": {
-        "type": "azure-cns"
-    },
-    "dns": {
-        "Nameservers": [
-            "168.63.129.16"
-        ],
-        "Search": [
-            "svc.cluster.local"
-        ]
-    },
-    "disableAsyncDelete": true
+    "plugins": [
+        {
+            "type": "azure-vnet",
+            "mode": "bridge",
+            "bridge": "azure0",
+            "capabilities": {
+                "portMappings": true,
+                "dns": true
+            },
+            "ipam": {
+                "type": "azure-cns"
+            },
+            "dns": {
+                "Nameservers": [
+                    "168.63.129.16"
+                ],
+                "Search": [
+                    "svc.cluster.local"
+                ]
+            },
+            "disableAsyncDelete": true
+        }
+    ]
 }


cni/azure-windows-swiftv2-stateless.conflist
```diff
@@ -0,0 +1,22 @@
...
+	},
+	"ipam": {
+		"type": "azure-cns"
+	},
+	"dns": {
+		"Nameservers": [
+			"168.63.129.16"
+		],
+		"Search": [
+			"svc.cluster.local"
```
tamilmani1989: do we need search? can remove this.

cni/azure-windows-swiftv2-stateless.conflist
```diff
@@ -0,0 +1,22 @@
+{
+    "cniVersion": "1.0.0",
+    "name": "azure",
+	"type": "azure-vnet",
+	"mode": "bridge",
```
tamilmani1989: add "disableAsyncDelete": true

cni/azure-windows-swiftv2-stateless.conflist
jackieluc: we don't _need_ Search functionally speaking for Standalone scenarios, but network_windows.go would error if there is a mismatch in length between NameServers and Search: https://github.com/jackieluc/azure-container-networking/blob/stateless-cni-conflist/cni/network/network_windows.go#L147-L150

cni/azure-windows-swiftv2-stateless.conflist
jackieluc: added

tamilmani1989 (dismissed):

cni/azure-windows-swiftv2-stateless.conflist
```diff
@@ -0,0 +1,27 @@
+{
+    "cniVersion": "1.0.0",
+    "name": "azure",
+    "plugins": [
+        {
+            "type": "azure-vnet",
+            "mode": "bridge",
+            "bridge": "azure0",
+            "capabilities": {
```
MikeZappa87, line 9: Does azure-vnet explicitly handle these or is this part of the runtime conf of the CNI?

jackieluc: i'm not familiar with the runtime conf for the CNI but this conflist was provided to me statically from Tamilmani, Behzad, and Alexander last year for Stateless CNI on Windows

jpayne3506 (approved):

cni/azure-windows-swiftv2-stateless.conflist
jackieluc: resolving for now as these fields are re-used from other conflists provided. @tamilmani1989 may know the answer to this one

PR #4155, opened 2025-12-03
fix: do not copy empty values into secondary IP configs

santhoshmprabhu:
**Reason for Change**:
For Cilium Nodesubnet tests, we need to create secondary ip configs on nodes, which is being done by copying the primary config, and editing some fields. During this process, we were leaving nil fields untouched. With a new AZ CLI update, this causes an error because some fields expect values to be set. In this PR, we skip over any values that are nil.

Validation:
Manually created a byocni cluster and verified that the script runs correctly.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/integration/cilium-nodesubnet/ipconfigupdate.go
+++ b/test/integration/cilium-nodesubnet/ipconfigupdate.go
@@ -104,11 +104,17 @@ func main() {
 			for i := 1; i <= secondaryConfigCount+1; i++ {
 				ipConfig := make(map[string]interface{})
 				for k, v := range primaryIPConfig {
-					// only the primary config needs loadBalancerBackendAddressPools. Azure doesn't allow
-					// secondary IP configs to be associated load balancer backend pools. Also skip nil values.
-					if k == "loadBalancerBackendAddressPools" || v == nil {
+					// Skip nil values.
+					if v == nil {
 						continue
 					}
+
+					// only the primary config needs loadBalancerBackendAddressPools. Azure doesn't allow
+					// secondary IP configs to be associated load balancer backend pools.
+					if i > 1 && k == "loadBalancerBackendAddressPools" {
+						continue
+					}
+
 					ipConfig[k] = v
 				}
 


test/integration/cilium-nodesubnet/ipconfigupdate.go
```diff
@@ -100,12 +100,13 @@ func main() {
 		}
 
 		if primaryIPConfig != nil {
-			for i := 2; i <= secondaryConfigCount+1; i++ {
+			ipConfigurations = []interface{}{}
```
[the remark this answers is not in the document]
santhoshmprabhu: Purposefully removing primary ipconfig and recreating it without nil values.

test/integration/cilium-nodesubnet/ipconfigupdate.go
```diff
@@ -117,6 +118,8 @@ func main() {
 					ipConfig["primary"] = false
 					usedIPConfigNames = append(usedIPConfigNames, ipConfigName)
 					secondaryConfigs = append(secondaryConfigs, ipConfig)
+				} else {
+					ipConfigurations = append(ipConfigurations, ipConfig)
```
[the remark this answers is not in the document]
santhoshmprabhu: The intention is to overwrite.

jpayne3506 (dismissed):

test/integration/cilium-nodesubnet/ipconfigupdate.go
```diff
@@ -100,12 +100,13 @@ func main() {
...
-			for i := 2; i <= secondaryConfigCount+1; i++ {
+			ipConfigurations = []interface{}{}
+			for i := 1; i <= secondaryConfigCount+1; i++ {
 				ipConfig := make(map[string]interface{})
 				for k, v := range primaryIPConfig {
 					// only the primary config needs loadBalancerBackendAddressPools. Azure doesn't allow
-					// secondary IP configs to be associated load balancer backend pools.
-					if k == "loadBalancerBackendAddressPools" {
+					// secondary IP configs to be associated load balancer backend pools. Also skip nil values.
+					if k == "loadBalancerBackendAddressPools" || v == nil {
```
tamilmani1989: why are we hardcoding `loadBalancerBackendAddressPools`

santhoshmprabhu: We do have a comment above, we are copying the primary ip config to create secondary ones, and we need to remove the field. I could define it as a variable, but this is anyway a utility script. Are you suggesting to do that?

rejain789 (dismissed):

test/integration/cilium-nodesubnet/ipconfigupdate.go
santhoshmprabhu: Resolving this conversation since the hard coding isn't new, can clean up in a follow up if needed.

PR #4157, opened 2025-12-04
Add HomeAZ info to NodeInfo CRD upon creation

timraymond:
Needed for podsubnet HomeAZ. Detailed description to come.

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -1664,13 +1664,16 @@ func getPodInfoByIPProvider(
 	return podInfoByIPProvider, nil
 }
 
-// createOrUpdateNodeInfoCRD polls imds to learn the VM Unique ID and then creates or updates the NodeInfo CRD
-// with that vm unique ID
 func createOrUpdateNodeInfoCRD(ctx context.Context, restConfig *rest.Config, node *corev1.Node) error {
 	imdsCli := imds.NewClient()
-	vmUniqueID, err := imdsCli.GetVMUniqueID(ctx)
+
+	nmaConfig, err := nmagent.NewConfig("")
 	if err != nil {
-		return errors.Wrap(err, "error getting vm unique ID from imds")
+		return errors.Wrap(err, "failed to create nmagent config")
+	}
+	nmaCli, err := nmagent.NewClient(nmaConfig)
+	if err != nil {
+		return errors.Wrap(err, "failed to create nmagent client")
 	}
 
 	directcli, err := client.New(restConfig, client.Options{Scheme: multitenancy.Scheme})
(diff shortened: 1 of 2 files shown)


cns/service/main.go
```diff
@@ -1694,29 +1694,68 @@ func getPodInfoByIPProvider(
 
 // createOrUpdateNodeInfoCRD polls imds to learn the VM Unique ID and then creates or updates the NodeInfo CRD
 // with that vm unique ID
```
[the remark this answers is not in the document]
nairashu: This seems like a reasonable suggestion to apply

cns/service/main.go
```diff
@@ -1694,29 +1694,68 @@ func getPodInfoByIPProvider(
...
-	vmUniqueID, err := imdsCli.GetVMUniqueID(ctx)
+
+	nmaConfig, err := nmagent.NewConfig("")
 	if err != nil {
-		return errors.Wrap(err, "error getting vm unique ID from imds")
+		return errors.Wrap(err, "failed to create nmagent config")
+	}
+	nmaCli, err := nmagent.NewClient(nmaConfig)
+	if err != nil {
+		return errors.Wrap(err, "failed to create nmagent client")
```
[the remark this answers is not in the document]
nairashu: Again a good suggestion to have a look at. I am spinning up codespace to see if cnsConfig.WireserverIP is being used.

cns/service/main.go
```diff
@@ -1698,29 +1694,68 @@ func getPodInfoByIPProvider(
...
+
+// HomeAzGetter is an interface for getting HomeAZ from NMAgent
+type HomeAzGetter interface {
+	GetHomeAz(ctx context.Context) (nmagent.AzResponse, error)
+}
+
+// NodeInfoCreator is an interface for creating/updating NodeInfo CRD
+type NodeInfoCreator interface {
+	CreateOrUpdate(ctx context.Context, nodeInfo *mtv1alpha1.NodeInfo, fieldOwner string) error
+}
```
nairashu, line 1704: I only see the interface here. Where are the implementations of these methods?

cns/service/main.go
```diff
@@ -1698,29 +1694,68 @@ func getPodInfoByIPProvider(
...
+		return errors.Wrap(err, "error getting vm unique ID from imds")
+	}
+
+	nodeInfoSpec := mtv1alpha1.NodeInfoSpec{
+		VMUniqueID: vmUniqueID,
+	}
+
+	homeAzResponse, err := nmaCli.GetHomeAz(ctx)
+	if err != nil {
+		z.Warn("failed to get HomeAZ from nmagent", zap.Error(err))
```
thatmattlong: I think we should just make this fatal and return the error

csfmomo: If we make the fatal and return the error here, I think we don't need to pass zap.Logger here.

cns/service/main.go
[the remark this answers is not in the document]
timraymond: Hallucination. The WireserverIP config was introduced for overriding in tests. It's not set in the CNS synth, so it becomes the default anyway. Won't fix.

cns/service/main.go
timraymond: They're used on the line below. Implementations are in the caller and in the tests.

cns/service/main.go
timraymond: Done in 289acb2

nairashu (approved):

timraymond:
@copilot Fix whatever is wrong with the `Docker Base Images / render (pull_request)` check

nairashu (approved):

rejain789 (approved):

PR #4163, opened 2025-12-11
fix: infinite loop in Device Plugin caused by stale SocketWatcher state

isaac-dasan:
### Description
This PR fixes a critical bug in the `SocketWatcher` that caused the device plugin to enter a tight restart loop if the socket file was deleted (e.g., by Kubelet) or if the plugin restarted for any reason.

### The Problem:
The `SocketWatcher` maintains a map `socketChans` of active watchers. When a watcher goroutine exited (due to socket deletion or context cancellation), it closed its notification channel but failed to remove the entry from the map.

Consequently, when the `PluginManager` attempted to restart the plugin:

1. It called `WatchSocket` with the same socket path.
2. `WatchSocket` found the existing entry in the map and returned the already-closed channel.
3. The plugin's main loop (Run) received from this closed channel immediately and exited.
4. The loop repeated indefinitely, causing the plugin to restart every ~100ms.

### Symptoms:

Logs showing `"starting device plugin for resource"` and `"registering with kubelet"` repeating rapidly with no intermediate errors.
Pods failing with `"UnexpectedAdmissionError: Allocate failed due to no healthy devices present"`. This occurred because the plugin was constantly churning, leaving no stable window for Kubelet to allocate devices.

### The Fix:
Updated SocketWatcher to ensure that the socket entry is deleted from the socketChans map when the watcher gor
outine exits. This ensures that subsequent calls to WatchSocket create a fresh watcher and channel.

### Testing:

Added a regression test TestWatchSocketCleanup in socketwatcher_test.go to verify that the map entry is cleaned up and new watchers can be established successfully.

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/cns/deviceplugin/socketwatcher_test.go
+++ b/cns/deviceplugin/socketwatcher_test.go
@@ -171,8 +171,8 @@ func TestWatchSocketCleanup(t *testing.T) {
 	}
 
 	// 2. Delete the socket to trigger watcher exit
-	if err := os.Remove(socket); err != nil {
-		t.Fatalf("failed to remove socket: %v", err)
+	if removeErr := os.Remove(socket); removeErr != nil {
+		t.Fatalf("failed to remove socket: %v", removeErr)
 	}
 
 	// 3. Wait for ch1 to close


cns/deviceplugin/socketwatcher_test.go
```diff
@@ -134,3 +134,61 @@ func TestWatchSocketTwice(t *testing.T) {
 		t.Fatal("socket watcher is still watching 5 seconds after file is deleted")
 	}
 }
+
+func TestWatchSocketCleanup(t *testing.T) {
+	// Create a temporary directory
+	tempDir, err := os.MkdirTemp("", "socket-watcher-test-")
```
thatmattlong: I tend to prefer adding a .gitignore file with entries for every filename that we might create so you don't accidentally check in test data

cns/deviceplugin/socketwatcher_test.go
```diff
@@ -134,3 +134,61 @@ func TestWatchSocketTwice(t *testing.T) {
...
+		t.Fatal("timed out waiting for watcher to detect socket deletion")
+	}
+
+	// 4. Recreate the socket
+	if _, err := os.Create(socket); err != nil {
+		t.Fatalf("error recreating test file %s: %v", socket, err)
+	}
+
+	// 5. Watch the socket again
+	// If the bug exists, this will return the OLD closed channel (ch1) or a closed channel
```
thatmattlong: not a useful comment IMO since the bug doesn't exist

cns/deviceplugin/socketwatcher_test.go
```diff
@@ -134,3 +134,61 @@ func TestWatchSocketTwice(t *testing.T) {
...
+	ch2 := s.WatchSocket(context.Background(), socket)
+
+	// 6. Verify ch2 is NOT closed immediately
+	select {
+	case <-ch2:
+		t.Fatal("WatchSocket returned a closed channel on the second call! The map entry was likely not cleaned up.")
```
thatmattlong: I wouldn't try to suggest internal implementation details of socketwatcher in this fatal log, just say "channel is closed but expected to be open", maybe the internals will change someday and you'd have to remember to change this log

cns/deviceplugin/socketwatcher_test.go
```diff
@@ -134,3 +134,61 @@ func TestWatchSocketTwice(t *testing.T) {
...
+	// 4. Recreate the socket
+	if _, err := os.Create(socket); err != nil {
+		t.Fatalf("error recreating test file %s: %v", socket, err)
+	}
+
+	// 5. Watch the socket again
+	// If the bug exists, this will return the OLD closed channel (ch1) or a closed channel
+	ch2 := s.WatchSocket(context.Background(), socket)
+
+	// 6. Verify ch2 is NOT closed immediately
```
thatmattlong: prefer "verify ch2 is open" since this comment eludes to a bug that doesn't exist but that's very nitpicky so feel free to ignore

cns/deviceplugin/socketwatcher_test.go
```diff
@@ -134,3 +134,61 @@ func TestWatchSocketTwice(t *testing.T) {
...
+	select {
+	case <-ch2:
+		t.Fatal("WatchSocket returned a closed channel on the second call! The map entry was likely not cleaned up.")
+	case <-time.After(200 * time.Millisecond):
+		// If we wait a bit and it's still open, that's good.
```
thatmattlong: might want to call out that we're waiting for at least one tick here rather than just "wait a bit" but also nitpicky

cns/deviceplugin/socketwatcher_test.go
isaac-dasan: This will create a file in `/tmp/socket-watcher-test-123456` and the defer os.RemoveAll(tempDir) will clean it up, so no need to add entries to .gitignore

thatmattlong: you can't guarantee that your os.RemoveAll will run, if your tests crash then the files will be left behind. On linux this is fine since the temp directory goes under /tmp but on windows I think os.MkdirTemp uses the current directory (not 100% sure about that). If you can guarantee that these temp dirs never get created in the scope of the git repo, then it is fine to not add a .gitignore

isaac-dasan: https://pkg.go.dev/os#MkdirTemp
https://pkg.go.dev/os#TempDir

Yes it is guaranteed to not be created in repo for windows as well. It typically looks like:
`C:\Users\<Username>\AppData\Local\Temp\socket-watcher-test-<random>\to-be-deleted.sock`

miguelgoms (approved):

cns/deviceplugin/socketwatcher_test.go
rbtr: do what matt said

rbtr: /tmp may not be accessible to a user with restricted permissions, may cross filesystem boundaries, may not actually be as temporary as you think (ie these files will persist until reboot, on Linux; and forever, on Windows). Etc. Also, this means your test has side-effects, and as you pointed out, escapes the repo scope - that's explicitly bad.

thatmattlong: conventionally tests use the `testdata` folder (go test will set the current working directory to the directory of your test file so you don't have to do anything tricky when building the base path), so just create `testdata/.gitignore` and add the filenames that your tests create, then you can just read from `testdata/<file>` in the test without needing to construct any special base paths

rbtr: At _least_ use [T.TempDir](https://pkg.go.dev/testing#T.TempDir). But keeping test filesystem I/O within the repo tree is best imo.

isaac-dasan: updated with t.TempDir() since it is designed to handle cleanup automatically, even if the test fails or panics.

isaac-dasan: created testdata/.gitignore

cns/deviceplugin/testdata/socket.sock
rbtr, line 1: should not be checked in

rbtr: if you added the gitignore after adding this file, it hid it from you but kept it staged and then committed it

isaac-dasan: This file is in master. I'm just deleting it.

rbtr: you haven't deleted it

rbtr: ah it's just showing up differently in the new GH review UI

rbtr (approved):

PR #4174, opened 2025-12-16
feat: select linux network mode based on conflist instead of statefile

QxBytes:
**Reason for Change**:

Previously, in linux, during ADD and DEL calls, in order to determine what networking mode and implementation we should use (ex: transparent, transparent-vlan, ovs, bridge etc.), we would read the statefile into memory, examine the network struct, and read the mode. On the first ADD,  we would get the network mode from the conflist and write it to the statefile. Because we only read from the conflist on the first ADD, subsequent changes to the networking mode in the conflist would have no effect.

This PR makes the CNI ignore the statefile's network mode completely, instead reading the networking mode from the conflist on each ADD and DEL call. This is already provided to us in ADD where the mode is added to the `endpointInfo` (now we just use that field), but we add a new "mode" parameter in DEL since this struct does not exist there.

**Issue Fixed**:

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Notes**:
ACN PR: https://msazure.visualstudio.com/One/_build/results?buildId=146222879&view=results
CNI Release: https://msazure.visualstudio.com/One/_build/results?buildId=146132532&view=results
- nwCfg.Mode should be the same as epInfo.Mode (all ep infos should have the same networking mode)
- If customer's netconflist mode != statefile mode, the behavior may change suddenly (jump between transparent and transparent-vlan mode for example), but that should only happen if they mess with the cni conflist (which shouldn't happen).
- PR shouldn't overwrite the statefile's network mode-- it's just not used anymore so if rolled back we'd just be using the existing mode in the statefile again. To clean up any eps created with the wrong networking mode, the most surefire way as always is to restart the vm.

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -609,7 +609,7 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 
 			// Delete all endpoints
 			for _, epInfo := range epInfos {
-				deleteErr := plugin.nm.DeleteEndpoint(epInfo.NetworkID, epInfo.EndpointID, epInfo, epInfo.Mode)
+				deleteErr := plugin.nm.DeleteEndpoint(epInfo.NetworkID, epInfo.EndpointID, epInfo, nwCfg.Mode)
 				if deleteErr != nil {
 					// we already do not return an error when the endpoint is not found, so deleteErr is a real error
 					logger.Error("Could not delete endpoint after detecting add failure", zap.String("epInfo", epInfo.PrettyString()), zap.Error(deleteErr))


cni/network/network.go
vipul-21, line 1: The code changes lgtm. Have a a question thouhg, now the statefile will be updated with `mode` with each call to CNI ? Or it is not updated at all in the statefile ?

QxBytes: Statefile previously gets the network mode once from the network config when the network struct is created (subsequent calls don't get the network mode from the network config because the statefile is already populated). This behavior does not change. Now, instead of reading the network mode from the statefile to choose an implementation, we always read the mode from the network config passed in.

PR #4177, opened 2025-12-18
ci: fix ebpf manifests and allow e2e usage with cilium 1.18 and up

QxBytes:
**Reason for Change**:

Adjusts manifests to be compatible with cilium distroless. Additionally, when the pipeline is run with cilium 1.17+ (including 1.18 etc.) ebpf host routing will respect the cilium version passed in and deploy the version used by the pipeline (previously it always used a 1.17 image). Additionally, if the pipeline is run with a cilium version < 1.17, the ebpf host routing test stages will be skipped at runtime because ebpf host routing is not supported on those versions.

**Issue Fixed**:

Distroless init does not contain bash. This PR replaces calls to bash with calls to sh.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
1.14 acn: https://msazure.visualstudio.com/One/_build/results?buildId=147032294&view=results
1.17 acn: https://msazure.visualstudio.com/One/_build/results?buildId=147032411&view=results
1.18 acn: https://msazure.visualstudio.com/One/_build/results?buildId=147032508&view=results

### DIFF
--- a/.pipelines/singletenancy/cilium-ebpf/cilium-e2e-step-template.yaml
+++ b/.pipelines/singletenancy/cilium-ebpf/cilium-e2e-step-template.yaml
@@ -32,7 +32,7 @@ steps:
         export EBPF_CILIUM_DIR=${DIR}
         export EBPF_CILIUM_VERSION_TAG=${CILIUM_VERSION_TAG}
         export AZURE_IPTABLES_MONITOR_IMAGE_REGISTRY=acnpublic.azurecr.io
-        export AZURE_IPTABLES_MONITOR_TAG=$(make azure-iptables-monitor-version)  
+        export AZURE_IPTABLES_MONITOR_TAG=$(make azure-iptables-monitor-version)
         make -C ./hack/aks deploy-ebpf-podsubnet-cilium
 
         kubectl get pods -Aowide


jpayne3506 (approved):

.pipelines/singletenancy/cilium-ebpf/cilium-e2e-step-template.yaml
```diff
@@ -21,6 +21,16 @@ steps:
         kubectl get pods -Aowide
 
         # see makefile
+        echo "install Cilium ${CILIUM_VERSION_TAG}"
+        export DIR=$(echo ${CILIUM_VERSION_TAG#v} | cut -d. -f1,2)
+        
+        # ebpf host routing is only supported on 1.17+
+        if [[ "${DIR}" == "1.13" || "${DIR}" == "1.14" || "${DIR}" == "1.15" || "${DIR}" == "1.16" ]]; then
```
vipul-21, line 28: Why do we need this if the job has the same condition already ?

QxBytes: this was just an extra check if the stage somehow were called from elsewhere so at least a message would print if we used an unsupported version

vipul-21: Shouldn't we skip the test if this is called with an unsupported version in that case ? Because this might fail the pipeline.

PR #4179, opened 2025-12-19
Enable scale tests with 20 pods

sivakami-projects:
**Reason for Change**:

This PR increases the scale test capacity from 15 to 20 pods while improving test reliability through better timeout handling, retry logic, and error reporting. The changes also include infrastructure updates to support the increased pod density and fixes to RBAC propagation issues in storage account setup.

Key Changes
Scale test now creates 20 pods (10 per cluster) instead of 15 (8+7 split)
Reduced deletion timeouts from 30-60 attempts to 10 attempts with consistent 20s command timeouts
Added RBAC propagation verification for storage accounts with retry logic
Changed scheduled test region from centraluseuap to eastus2

Test run - https://msazure.visualstudio.com/One/_build/results?buildId=147257463&view=results
[Run 2](https://msazure.visualstudio.com/One/_build/results?buildId=147557030&view=results)
**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/swiftv2-long-running/scripts/create_storage.sh
+++ b/.pipelines/swiftv2-long-running/scripts/create_storage.sh
@@ -49,10 +49,9 @@ for SA in "$SA1" "$SA2"; do
   
   # Retry blob upload with exponential backoff if RBAC hasn't propagated yet
   MAX_RETRIES=5
-  RETRY_COUNT=0
   SLEEP_TIME=10
   
-  while [ $RETRY_COUNT -lt $MAX_RETRIES ]; do
+  for i in $(seq 1 $MAX_RETRIES); do
     if az storage blob upload \
       --account-name "$SA" \
       --container-name "test" \
@@ -63,9 +62,8 @@ for SA in "$SA1" "$SA2"; do
       echo "[OK] Test blob 'hello.txt' uploaded to $SA/test/"
       break
     else
-      RETRY_COUNT=$((RETRY_COUNT + 1))
-      if [ $RETRY_COUNT -lt $MAX_RETRIES ]; then
-        echo "[WARN] Blob upload failed (attempt $RETRY_COUNT/$MAX_RETRIES). Waiting ${SLEEP_TIME}s for RBAC propagation..."
+      if [ $i -lt $MAX_RETRIES ]; then
+        echo "[WARN] Blob upload failed (attempt $i/$MAX_RETRIES). Waiting ${SLEEP_TIME}s for RBAC propagation..."
         sleep $SLEEP_TIME
         SLEEP_TIME=$((SLEEP_TIME * 2))
       else


.pipelines/swiftv2-long-running/pipeline.yaml
```diff
@@ -18,14 +18,13 @@ parameters:
   - name: location
     displayName: "Deployment Region"
     type: string
-    default: "centraluseuap"
+    default: "eastus2"
```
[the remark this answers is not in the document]
sivakami-projects: This is not an issue.

.pipelines/swiftv2-long-running/template/long-running-pipeline-template.yaml
```diff
@@ -277,7 +277,7 @@ stages:
                 assign
                 ${{ parameters.subscriptionId }}
                 $(rgName)
-                "$(DiscoverStorageAccounts.storageAccount1) $(DiscoverStorageAccounts.storageAccount2)"
+                "$(DiscoverStorageAccounts.storageAccount1)"
```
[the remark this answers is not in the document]
sivakami-projects: storage account 2 does not require RBAC assignment.

test/integration/swiftv2/longRunningCluster/datapath_scale_test.go
```diff
@@ -142,29 +134,36 @@ var _ = ginkgo.Describe("Datapath Scale Tests", func() {
...
+		ginkgo.By("Waiting 10 seconds for pods to stabilize")
+		time.Sleep(10 * time.Second)
 
 		ginkgo.By("Verifying all pods are in Running state")
 		podIndex = 0
+		var verificationErrors []error
 		for i, scenario := range scenarios {
 			for j := 0; j < scenario.podCount; j++ {
 				podName := fmt.Sprintf("scale-pod-%d", podIndex)
 				err := helpers.WaitForPodRunning(allResources[i].Kubeconfig, allResources[i].PNName, podName, 5, 10)
```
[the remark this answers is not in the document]
sivakami-projects: time for pod to be running state is reduced to 10 retries with 10 second intervals.

paulyufan2 (dismissed):
lgtm

test/integration/swiftv2/longRunningCluster/datapath.go
```diff
@@ -234,12 +236,20 @@ func CreateNamespaceResource(kubeconfig, namespace string) error {
 }
 
 func CreatePodNetworkInstanceResource(resources TestResources) error {
+	reservations := resources.Reservations
+	if reservations == 0 {
+		reservations = 2
```
kmurudi: will it be always testing explicit reservations in this pipeline? if not reset why would reservations be set to 0 at any point?

sivakami-projects: No, not all PNI are created with explicit reservations. Pods created for the networking connectivity tests have implicit reservation, when reservation is not set explicitly it defaults to 2.

kmurudi: we should define this in the caller then in resources.Reservations itself, not to alter here or use a different PNI for implicit vs explicit

.pipelines/swiftv2-long-running/scripts/create_storage.sh
```diff
@@ -46,14 +46,34 @@ for SA in "$SA1" "$SA2"; do
...
-    --auth-mode login \
-    --overwrite \
-    && echo "[OK] Test blob 'hello.txt' uploaded to $SA/test/"
+  
+  # Retry blob upload with exponential backoff if RBAC hasn't propagated yet
+  MAX_RETRIES=5
+  RETRY_COUNT=0
+  SLEEP_TIME=10
+  
+  while [ $RETRY_COUNT -lt $MAX_RETRIES ]; do
```
kmurudi: wouldn't easier way here to be check for max of 5 times & break if success? don't see the need of if-else & Retry_count

kmurudi: similar to loop below in amange_storage_rbac.sh

.pipelines/swiftv2-long-running/template/long-running-pipeline-template.yaml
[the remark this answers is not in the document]
kmurudi: what's the difference between the 2 storage accounts that storage1 needs & storage2 does not?

.pipelines/swiftv2-long-running/scripts/create_storage.sh
sivakami-projects: We are breaking if the upload succeeds, [here](https://github.com/Azure/azure-container-networking/pull/4179/files#diff-608de02835ffdf51cacfcd6f9edd8a7785d763a123f752ae32d28820a16b6b12R63-R64 )

This is different from manage_storage_rbac loop as we have exponential backoff here for the rbac role to propagate.

.pipelines/swiftv2-long-running/template/long-running-pipeline-template.yaml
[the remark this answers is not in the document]
sivakami-projects: Storage account 1 has a private endpoint and it is used in the private endpoint tests. Storage account 2 has a public endpoint. To run the private endpoint tests on storage account 1, pipeline has to generate a SAS token. To generate the SAS token, we are assigning the storage blob contributor role here.

test/integration/swiftv2/longRunningCluster/datapath.go
```diff
@@ -234,12 +236,16 @@ func CreateNamespaceResource(kubeconfig, namespace string) error {
...
+	namespace := resources.Namespace
+	if namespace == "" {
+		namespace = resources.PNName
+	}
 	err := CreatePodNetworkInstance(resources.Kubeconfig, PNIData{
 		PNIName:      resources.PNIName,
 		PNName:       resources.PNName,
-		Namespace:    resources.PNName,
+		Namespace:    namespace,
 		Type:         "explicit",
```
kmurudi: will this need to be set to "implicit" if re-used for any other tests, would be good to configure it that way if so

.pipelines/swiftv2-long-running/scripts/create_storage.sh
kmurudi: I see only the sleep time changing in that case, you can still keep the lines L69-L70, but can use a for loop for loops 1 to MAX_Retries , and if-else within to try upload else sleep & continue. RETRY_COUNT is a redundant variable update here, i.e. change can be made with just max_retries & multiplying sleep time in each loop

test/integration/swiftv2/longRunningCluster/datapath.go
sivakami-projects: With the reservations explicitly set we would not need this. Removed this field and committed the changes.

sharifnasser (approved):

PR #4183, opened 2025-12-26
ci: fix known cniv1 pipeline issue and improve log collection

QxBytes:
**Reason for Change**:

There is a known issue in the pipeline for cniv1 during ip allocation. A symptom of this is "Initializing HTTP client with connection timeout" showing up in the cni logs. This PR adds a script to check the contents of the logs for these known phrases and marks the stage as succeeded with warnings if so. If the phrase is not found but there is an error, we fail out as normal.

Additionally adds tolerations to the privileged pods so that they always are scheduled, even if cilium or other components add taints to the nodes.

Additionally moves cni/cns log collection steps to windows or linux specific scripts. The goal is that anyone can set their kubectx to a cluster, run the collection scripts with appropriate parameters and the logs will be downloaded automatically, even outside of pipeline environments.

The log checking script in the future may also be used to detect other known issues in the pipeline.

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Green: https://msazure.visualstudio.com/One/_build/results?buildId=147727074&view=results
Detect: https://msazure.visualstudio.com/One/_build/results?buildId=147893558&view=results

### DIFF
--- a/.pipelines/singletenancy/aks/e2e-job-template.yaml
+++ b/.pipelines/singletenancy/aks/e2e-job-template.yaml
@@ -52,7 +52,11 @@ stages:
       - job: ${{ parameters.name }}
         displayName: Singletenancy AKS - (${{ parameters.name }})
         timeoutInMinutes: 120
-        # continue on error only in linux case to workaround known cniv1 issue
+        # continue on error only in linux case to workaround known cniv1 issue where we leak an ip
+        # this occurs when the azure vnet ipam releases an ip from its statefile and sends a successful
+        # delete response but the azure cni gets a broken pipe error and doesn't delete the ip from its
+        # statefile. a symptom of this problem is 'Initializing HTTP client with connection timeout'
+        # appearing in the logs, which is checked below
         continueOnError: ${{ eq(parameters.os, 'linux') }}
         pool:
           name: $(BUILD_POOL_NAME_DEFAULT)


.pipelines/singletenancy/aks/e2e-job-template.yaml
```diff
@@ -66,6 +68,16 @@ stages:
...
+      
+      # can only succeed with issues in linux
+      - template: ../../templates/warning-handler-job-template.yaml
+        parameters:
+          dependsOnToEmitWarning: ${{ parameters.name }}
+          clusterName: ${{ parameters.clusterName }}-$(commitID)
+          fileSearchPattern: "azure-vnet*.log"
+          searchPattern: "Initializing HTTP client with connection timeout"
+          cni: cniv1
+          os: ${{ parameters.os }}
```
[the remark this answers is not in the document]
QxBytes: In windows, dependsOnToEmitWarning will depend on the previous stage to emit a warning when this runs, but the condition continueOnError is only true when the os is linux, thus in windows, the previous stage will never emit a warning and this block will never be run

hack/scripts/collect-windows-logs.sh
```diff
@@ -0,0 +1,71 @@
...
+done
+
+if [ ${cni} = 'cniv2' ]; then
+  echo "------ CNS work ------"
+
+
+  kubectl get pods -n kube-system -l k8s-app=azure-cns-win --no-headers
+  echo "Capture State Files from CNS pods"
+  managed=`kubectl get cm cns-win-config -n kube-system -o jsonpath='{.data.cns_config\.json}' | jq .ManageEndpointState`
+  for pod in $podList; do
```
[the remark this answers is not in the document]
QxBytes: This is already correct-- it is run from the privileged daemonset and that data can be retrieved from the host, even though it is for the cns pods. For example, we can query the cnsCache, azure-cns.json, and azure-endpoints.json since the k directory is mounted. also if the cns is down this can still retrieve the logs

hack/scripts/collect-linux-logs.sh
```diff
@@ -0,0 +1,93 @@
...
+fi
+
+if [ ${cni} = 'cilium' ] || [ ${cni} = 'cniv2' ]; then
+  echo "------ CNS work ------"
+
+
+  kubectl get pods -n kube-system -l k8s-app=azure-cns
+  echo "Capture State Files from CNS pods"
+  managed=`kubectl get cm cns-config -n kube-system -o jsonpath='{.data.cns_config\.json}' | jq .ManageEndpointState`
+  for pod in $podList; do
```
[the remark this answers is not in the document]
QxBytes: see comment in windows section- this should still get the logs since the directory is mounted to the host

.pipelines/templates/warning-handler-job-template.yaml
```diff
@@ -0,0 +1,38 @@
...
+          inlineScript: |
+            make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}
+        displayName: Set Up Kubeconfig
+        condition: always()
+      - script: |
+          cd hack/scripts
+          acnLogs=$(System.DefaultWorkingDirectory)/${{ parameters.clusterName }}_check_common_issues
+          acnLogs=$acnLogs cni=${{ parameters.cni }} bash collect-${{ parameters.os }}-logs.sh
+          echo "acnLogs directory: $acnLogs"
+          pwd
```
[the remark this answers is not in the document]
QxBytes: condition: always() means that the job/task always runs, and since check-cni-log-contents.sh exits 0 or 1 and is the last statement in the script, the job should fail if there is a non zero exit code from the script

test/integration/manifests/load/privileged-daemonset-windows.yaml
```diff
@@ -19,6 +19,8 @@ spec:
     spec:
       nodeSelector:
         kubernetes.io/os: windows
+      tolerations:
+      - operator: Exists
```
vipul-21, line 23: Is there a reason for adding a toleration for all taints ?
I see from the description that we want these pods to be scheduled always, no matter the taint on the node. Wondering what is the issue that blocked the pod to be scheduled. It should eventually be scheduled once the taint from the node is removed.

.pipelines/singletenancy/aks/e2e-job-template.yaml
```diff
@@ -52,6 +52,8 @@ stages:
       - job: ${{ parameters.name }}
         displayName: Singletenancy AKS - (${{ parameters.name }})
         timeoutInMinutes: 120
+        # continue on error only in linux case to workaround known cniv1 issue
```
vipul-21: nit: Can we point to the issue itself if there's one. Or add a brief here so that we know why we are continuing on error.

.pipelines/singletenancy/aks/e2e-job-template.yaml
```diff
@@ -66,6 +68,16 @@ stages:
...
               os_version: ${{ parameters.os_version }}
               scaleup: ${{ parameters.scaleup }}
+      
+      # can only succeed with issues in linux
+      - template: ../../templates/warning-handler-job-template.yaml
+        parameters:
+          dependsOnToEmitWarning: ${{ parameters.name }}
+          clusterName: ${{ parameters.clusterName }}-$(commitID)
+          fileSearchPattern: "azure-vnet*.log"
+          searchPattern: "Initializing HTTP client with connection timeout"
```
vipul-21, line 82: I understand the intention behind emitting a warning for this condition. My concern is that this makes the pipeline aware of a specific test case and its failure mode. Ideally, the pipeline should focus on orchestrating test execution, while the test itself determines whether a condition should result in a warning or failure. This would help keep responsibilities clearly separated and make the pipeline easier to maintain.

Instead of relying on a bash script in the pipeline to inspect log contents, we could add a Go test that validates this scenario directly.

hack/scripts/collect-windows-logs.sh
vipul-21, line 1: It might be worth merging the Linux and Windows log collection scripts now. While this wasn’t feasible earlier in the pipeline, we can now pass the OS type to the script and define log paths conditionally based on that. This would help reduce duplication.

vipul-21 (changes requested):

test/integration/manifests/load/privileged-daemonset-windows.yaml
QxBytes: I believe I encountered a scenario in cilium where the agent/operator was not ready/couldn't start up and left a taint on the node that wouldn't auto remove since it was dependent on the agent/operator coming up-- since we want logs no matter what if something fails, I specified this toleration.

.pipelines/singletenancy/aks/e2e-job-template.yaml
QxBytes: The actual test that is failing is failing for a legitimate reason (there _is_ a leaked ip). Integrating this exception into the golang test itself also seems unnatural (we'd be validating ips, and then checking the logs for a specific value for this specific case-- no other scenarios would have this logic and the issue has not been observed outside of our pipelines).

I went with the script while considering the other issue we had in the pipelines, null characters appearing in the statefile, where the test might fail even for the upstream tests but we probably also want to mark it with a warning. This issue has also only appeared inside our pipelines.

Since these workarounds are for pipeline issues I thought it would make sense to put them in the pipeline, rather than complicate the individual tests.

hack/scripts/collect-windows-logs.sh
QxBytes: hmm between the windows and linux scripts we have different shells, different means of getting the files (ls vs. find), cilium is only on linux etc. which I think at least for now justifies them being separate

.pipelines/templates/warning-handler-job-template.yaml
```diff
@@ -0,0 +1,38 @@
...
+          bash check-cni-log-contents.sh "$acnLogs" "${{ parameters.fileSearchPattern }}" "${{ parameters.searchPattern }}"
+        displayName: Check for known ${{ parameters.os }} issue
+        condition: always()
```
jpayne3506, line 38: If we enforce this step to be `- bash: |` then we do not need to leverage an additional shell on line 36. This would enforce consistent error reporting. 

Out of curiosity, if the script fails on line 36, does it bubble all the back up? Or does it have a false positive?

QxBytes: If the logs do not contain the search pattern on line 36, the bash script fails, fails the script (since it is last to execute), which fails the stage/pipeline as a whole

.pipelines/singletenancy/aks/e2e-job-template.yaml
vipul-21: >The actual test that is failing is failing for a legitimate reason (there is a leaked ip)

Shouldn't we fix this in that case ? 

>Integrating this exception into the golang test itself also seems unnatural (we'd be validating ips, and then checking the logs for a specific value for this specific case

Cilium does this with connectivity test. They check the log of the agent if there are any errors and fail the test. Pipeline inherently does not break the test, we are running it in such a manner that we see the failure in the pipeline. It would be worth to check if the test itself that causes the issue can fixed.

hack/scripts/collect-windows-logs.sh
vipul-21: Not a fan of scripts, so if we can do it in the golang that would nice and easier to manage imo.

.pipelines/singletenancy/aks/e2e-job-template.yaml
QxBytes: For the failing test we have already investigated and determined it is a race outside of our control during restart. What makes this different from the cilium checking of logs is that they check logs and fail if they find something. In this case, we have already failed the test and need to check logs after the test has failed to see if we pass.

vipul-21 (approved):

vipul-21:
Approved, discussed offline about the comments. The issue only occurred in pipeline so far so we will be skipping it as it has been discussed with @tamilmani1989 as per @QxBytes.

QxBytes:
This PR can't be merged in until the E2E pipelines stop warning on every stage

.pipelines/templates/log-template.yaml
```diff
@@ -53,6 +53,12 @@ steps:
       kubectl describe nodes
     displayName: Node Status
     condition: always()
+  
+  - bash: |
+      kubectl describe nnc -A
+    displayName: Describe Node Network Configs (NNCs)
+    condition: and(always(), ne('${{ parameters.cni }}', 'cniv1')) # cniv1/nodesubnet scenarios do not have node network configs
+    continueOnError: true # even if nnc doesn't exist we want to continue
```
jpayne3506, line 61: nit: we can conditionally insert this step instead of abusing the conditional `ne('${{ parameters.cni }}', 'cniv1')` .

QxBytes: will revisit if this conditional gets any more complex

PR #4184, opened 2025-12-30
ci: move cilium deploy logic to makefile

QxBytes:
**Reason for Change**:

Move cilium deploy logic for all scenarios to the deploy makefile. The goal is that both the pipeline and developers can use these targets after setting their kubectx to a byocni cilium (nokubeproxy) cluster and it'll automatically deploy. Also reduces repetition and cleans up the pipeline as all deploy logic for cilium should be in the makefiles now-- the pipeline just sets variables and runs the correct recipe. There should be no behavioral change in the pipeline. EBPF host routing cilium recipes are unaffected.

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
1.14 cilium ACN PR: https://msazure.visualstudio.com/One/_build/results?buildId=147897281
1.17 cilium ACN PR: https://msazure.visualstudio.com/One/_build/results?buildId=147903887
1.18 cilium ACN PR: https://msazure.visualstudio.com/One/_build/results?buildId=147966391&view=results
Release Test: https://msazure.visualstudio.com/One/_build/results?buildId=147907045&view=results
Nightly: https://msazure.visualstudio.com/One/_build/results?buildId=147911298
ACN Build: https://dev.azure.com/msazure/One/_build/results?buildId=147897317&view=results

### DIFF
--- a/hack/aks/deploy.mk
+++ b/hack/aks/deploy.mk
@@ -1,8 +1,8 @@
 # general cilium variables
-DIR ?= 1.17
-CILIUM_VERSION_TAG               	?= v1.17.7-250927
-CILIUM_IMAGE_REGISTRY           	?= mcr.microsoft.com/containernetworking
-IPV6_IMAGE_REGISTRY					?= mcr.microsoft.com/containernetworking
+DIR 									?= 1.17
+CILIUM_VERSION_TAG               		?= v1.17.7-250927
+CILIUM_IMAGE_REGISTRY           		?= mcr.microsoft.com/containernetworking
+IPV6_IMAGE_REGISTRY						?= mcr.microsoft.com/containernetworking
 IPV6_HP_BPF_VERSION               		?= v0.0.1
 
 # ebpf cilium variables
@@ -51,7 +51,7 @@ deploy-cilium-config-hubble:
 
 deploy-cilium-hubble: deploy-cilium-config-hubble deploy-cilium-agent deploy-cilium-operator wait-for-cilium
 
-# cilium nightly deployment  
+# cilium nightly deployment
 deploy-cilium-config-nightly:
 	kubectl apply -f ../../test/integration/manifests/cilium/cilium-nightly-config.yaml
 
--- a/hack/aks/README.md
+++ b/hack/aks/README.md
@@ -45,4 +45,12 @@ AKS Clusters
   windows-nodepool-up                       Add windows node pool
   down                                      Delete the cluster
   vmss-restart                              Restart the nodes of the cluster
+
+Deploy (onto a byocni cluster without kube-proxy)
+  deploy-cilium                             Deploy standard Cilium with agent, operator, and config
+  deploy-cilium-hubble                      Deploy Cilium with Hubble observability enabled
+  deploy-cilium-nightly                     Deploy Cilium using nightly builds
+  deploy-cilium-dualstack                   Deploy Cilium with dual-stack IPv4/IPv6 support
+  deploy-ebpf-overlay-cilium                Deploy Cilium with eBPF dataplane on overlay networks
+  deploy-ebpf-podsubnet-cilium              Deploy Cilium with eBPF dataplane on podsubnet networks
 ```


.pipelines/singletenancy/cilium-dualstack-overlay/cilium-dualstackoverlay-e2e-step-template.yaml
```diff
@@ -33,20 +33,14 @@ steps:
...
-
         export CILIUM_VERSION_TAG=${CILIUM_DUALSTACK_VERSION}
+        export DIR=$(echo ${CILIUM_VERSION_TAG#v} | cut -d. -f1,2)
         export IPV6_HP_BPF_VERSION=$(make ipv6-hp-bpf-version)
-        echo "install Cilium ${CILIUM_DUALSTACK_VERSION} onto Overlay Cluster"
-        envsubst '${CILIUM_VERSION_TAG},${CILIUM_IMAGE_REGISTRY},${IPV6_IMAGE_REGISTRY},${IPV6_HP_BPF_VERSION}' < test/integration/manifests/cilium/v${DIR}/cilium-agent/templates/daemonset-dualstack.yaml | kubectl apply -f -
-        envsubst '${CILIUM_VERSION_TAG},${CILIUM_IMAGE_REGISTRY}' < test/integration/manifests/cilium/v${DIR}/cilium-operator/templates/deployment.yaml | kubectl apply -f -
+        echo "IPV6_HP_BPF_VERSION: $IPV6_HP_BPF_VERSION"
+        echo "IPV6_IMAGE_REGISTRY: $IPV6_IMAGE_REGISTRY"
+        # IPV6_IMAGE_REGISTRY is already exported
```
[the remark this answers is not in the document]
QxBytes: previous code also assumed the value is exported

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
```diff
@@ -93,34 +93,19 @@ stages:
...
                     export IPV6_IMAGE_REGISTRY=acnpublic.azurecr.io
                     export IPV6_HP_BPF_VERSION=$(make ipv6-hp-bpf-version)
-                    envsubst '${CILIUM_VERSION_TAG},${CILIUM_IMAGE_REGISTRY},${IPV6_IMAGE_REGISTRY},${IPV6_HP_BPF_VERSION}' < test/integration/manifests/cilium/v${DIR}/cilium-agent/templates/daemonset-dualstack.yaml | kubectl apply -f -
+                    echo "IPV6_HP_BPF_VERSION: $IPV6_HP_BPF_VERSION"
+                    echo "IPV6_IMAGE_REGISTRY: $IPV6_IMAGE_REGISTRY"
+                    echo "installing cilium dualstack version ${CILIUM_VERSION_TAG} from directory ${DIR}"
+                    make -C ./hack/aks deploy-cilium-dualstack
                   else
-                    envsubst '${CILIUM_VERSION_TAG},${CILIUM_IMAGE_REGISTRY}' < test/integration/manifests/cilium/v${DIR}/cilium-agent/templates/daemonset.yaml | kubectl apply -f -
+                    export DIR=$(echo ${CILIUM_VERSION_TAG#v} | cut -d. -f1,2)
```
vipul-21, line 104: We can specify `DIR` outside of the conditions ? In that case we will only need to do it once ?

hack/aks/deploy.mk
vipul-21, line 1: let's update the readme too.

hack/aks/deploy.mk
```diff
@@ -1,19 +1,85 @@
...
+
+# ebpf
 export AZURE_IPTABLES_MONITOR_IMAGE_REGISTRY
 export AZURE_IPTABLES_MONITOR_TAG
 export AZURE_IP_MASQ_MERGER_IMAGE_REGISTRY
 export AZURE_IP_MASQ_MERGER_TAG
 
+wait-for-cilium:
+	@kubectl rollout status deployment/cilium-operator -n kube-system --timeout=1200s
+	@kubectl rollout status daemonset/cilium -n kube-system --timeout=1200s
```
vipul-21, line 32: nit: We can reduce the timeout i think. 20 minutes seems high. Maybe let's do 10 ?

QxBytes: I think when testing I initially set it at 10 minutes but some cases still timed out which is why it's at 20 minutes

.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml
QxBytes: the `DIR` is dependent on the `CILIUM_VERSION_TAG` which is different in the if vs. else block right?

vipul-21: > the DIR is dependent on the CILIUM_VERSION_TAG

Yes.
But the way it is defined it is same for both (if/else). 
Moreover I think it was outside the if/else block before too.

vipul-21: My bad, you are right.

PR #4175, opened 2025-12-17
ci: General Cilium Nightly fixes

jpayne3506:
**Reason for Change**:

Cilium Nightly failing due to a number of upstream changes
- acnpublic usage for azure-ipam v0.3.0 - Image no longer exists, Moved usage to MCR
- operator failure due to config validation checking for timeformat on `unmanaged-pod-watcher-interval: "15"` | https://github.com/cilium/cilium/pull/43344/files#diff-0d0e8dda37431037db0d692857b55f2926e7b4e162de80e889a5cf10b67d2a70
- Device (eth0) Change/drop | `systemctl restart systemd-networkd` restarts network which drops devices. [Panic is from missing DirectRoutingDevice check](https://github.com/cilium/cilium/commit/ce709caaa9b3210c78dd7b82afb95fd646aa3a64)
- zTunnel adoption forces RBAC onto operator 1.19+ | https://github.com/cilium/cilium/pull/42766

Ongoing failures that will not be fixed until more changes are done
- Device (eth0) Change/drop | Requires upstream changes to fix. 
- Error logged during k8sE2E test when running service conformance test. Non impacting, but caught in cilium agent logs which fails the pipeline. 

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [x] relevant PR labels added

**Notes**:

PR #4191, opened 2026-01-08
ci: General Cilium Nightly fixes ( cont. )

jpayne3506:
**Reason for Change**:

Resolves lingering issues from #4175.
Adds validate template in an effort to unify CI/CD.

Issues:
- `Device (eth0) Change/drop | Requires upstream changes to fix.`
  - Removed systemd-networkd restart test from cilium nightly until upstream fix is merged. 
  - This resulted in splitting out the test from `TestStateValidation`. Added make recipes to support and retain original intent
- `Error logged during k8sE2E test when running service conformance test. Non impacting, but caught in cilium agent logs which fails the pipeline.`
  - Testing confirmed that failure was a false negative from too many tests occurring at the same time.  Added a lever to control the amount of parallelized processes and restricted the nightly to be serial. This also has the side benefit of giving us complete logs from every test.
 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jpayne3506:
Will address `.pipelines/cni` templates in a follow up PR.

.pipelines/cni/cilium/nightly-release-test.yml
```diff
@@ -157,6 +157,7 @@ stages:
           dns: true
           portforward: true
           service: true
+          processes: 1 # Serialize testing to get granular data, prevent race conditions, false-negatives
```
vipul-21: What do you mean "prevent race conditions" ? Is that specific to the test cases running in a specific order ? It should be good to have non serialize right ? (to catch races)

.pipelines/cni/k8s-e2e/k8s-e2e-job-template.yaml
```diff
@@ -4,6 +4,7 @@ parameters:
   dependsOn: ""
   sub: ""
   cni: cni
+  processes: 8
```
vipul-21, line 7: magic number ?

.pipelines/singletenancy/azure-cni-overlay-stateless/azure-cni-overlay-stateless-e2e.steps.yaml
```diff
@@ -40,7 +40,7 @@ steps:
         fi
       done
       sudo -E env "PATH=$PATH" make test-load \
-        SCALE_UP=32 OS_TYPE=linux CNI_TYPE=stateless VALIDATE_STATEFILE=true VALIDATE_V4OVERLAY=true \
+        SCALE_UP=32 OS_TYPE=windows CNI_TYPE=stateless VALIDATE_STATEFILE=true VALIDATE_V4OVERLAY=true \
```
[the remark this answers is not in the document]
jpayne3506: stateless is restricted to windows

.pipelines/templates/validate-tests.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - task: AzureCLI@2
+      inputs:
+        azureSubscription: $(BUILD_VALIDATIONS_SERVICE_CONNECTION)
+        scriptLocation: "inlineScript"
+        scriptType: "bash"
+        addSpnToEnvironment: true
+        inlineScript: |
+          set -e
+          echo "Validating Networkd Restart"
+          make test-validate-networkd-restart OS_TYPE=${{ parameters.os }} CLEANUP=${{ parameters.cleanup }}
```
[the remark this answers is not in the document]
jpayne3506: does not require CNI_TYPE

camrynl:
is there an issue you can link for systemd-networkd restart fix?

camrynl (dismissed):

test/validate/validate.go
```diff
@@ -103,16 +103,24 @@ func (v *Validator) Validate(ctx context.Context) error {
 	}
 
 	if v.os == "linux" {
-		// We are restarting the systmemd network and checking that the connectivity works after the restart. For more details: https://github.com/cilium/cilium/issues/18706
 		log.Printf("Validating the restart network scenario")
-		err = v.validateRestartNetwork(ctx)
+		err = v.ValidateNetworkdRestart(ctx)
```
QxBytes, line 111: nit: didn't realize that Validate() actually validates other things like the networkd restart-- maybe this can be put in the comment for the function?

.pipelines/templates/cilium-nightly-checks.yaml
```diff
@@ -9,7 +9,7 @@ steps:
       echo "expect the identities to be deleted when the namespace is deleted"
       kubectl get ciliumidentity | grep cilium-test
 
-      make test-validate-state
+      make test-validate
```
QxBytes: so this is for the nightly pipeline right? If so, test-validate runs `test-validate: test-validate-state test-validate-networkd-restart` but based on your pr description didn't we want to skip `test-validate-networkd-restart` ? Also should we change the echo statement above that mentions the test checks systemd-networkd restart?

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e.steps.yaml
```diff
@@ -106,26 +107,12 @@ steps:
...
-        echo "Validating Node Restart"
-        make test-validate-state OS_TYPE=linux RESTART_CASE=true
-        kubectl delete ns load-test
-    displayName: "Validate Node Restart"
-    retryCountOnTaskFailure: 3
+  - template: ../../templates/validate-tests.yaml
+    parameters:
+      validate: "nodeRestart"
+      cni: "cilium"
+      os: ${{ parameters.os }}
```
QxBytes, line 114: is os a parameter passed in here? if this is an issue there might be other places without the os like `.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml` etc. unless I'm missing where it's defined (is it like a pipeline var?)

.pipelines/singletenancy/cilium-dualstack-overlay/cilium-dualstackoverlay-e2e.steps.yaml
```diff
@@ -124,15 +110,17 @@ steps:
...
+
       echo "delete cilium connectivity test resources and re-validate state"
       ciliumNamespace=`kubectl get ns | grep cilium-test | awk '{print $1}'`
       for namespace in ${ciliumNamespace}; do
         kubectl delete ns ${namespace}
       done
+
       kubectl get pod -owide -A
-      CNI_TYPE=cilium_dualstack go test -timeout 30m -tags load -run ^TestValidateState$
+      make test-validate-state CNI_TYPE=cilium_dualstack
```
QxBytes, line 123: since test-validate-state calls TestValidateState which now solely validates the state file, this is no longer testing networkd restart-- is this okay? (may also apply to `.pipelines/singletenancy/cilium-dualstack-overlay/cilium-dualstackoverlay-e2e-step-template.yaml`) Unless this is covered by the `make test-validate` above?

.pipelines/singletenancy/aks/e2e.steps.yaml
```diff
@@ -59,16 +59,10 @@ steps:
...
-      cd ../../..
 
-      make test-validate-state OS_TYPE=${{ parameters.os }} CNI_TYPE=cniv1
-
-      kubectl delete ns load-test
-    displayName: "Validate State"
-    retryCountOnTaskFailure: 3
+  - template: ../../templates/validate-tests.yaml
+    parameters:
+      validate: "nodeRestart"
```
QxBytes: if we want to retain the same behavior for non-cilium nightly pipeline, 
`make test-validate-state` would be equivalent to `validate: "nodeRestart,networkdRestart"` right?
Since test-validate-state in validate-tests.yaml > nodeRestart calls  make test-validate-state which calls `TestValidateState` which calls `ValidateStateFile` (but does not call networkd restart validation anymore)

.pipelines/singletenancy/cilium-dualstack-overlay/cilium-dualstackoverlay-e2e.steps.yaml
jpayne3506: `make test-validate` captures the original validation of both. There is no need to validate `networkd restart` a second time after deleting the namespace.

.pipelines/singletenancy/aks/e2e.steps.yaml
jpayne3506: Good catch. Yes, we would want to retain that original functionality.

jpayne3506: Fyi, this scenarios step template is reused for windows, but the go test will gracefully skip the `networkdRestart`

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e.steps.yaml
jpayne3506: If it is passed in from any parent template the `parameters.os` will be populated. Varying on the scenario it is passed in from `pipeline.yaml` or the job template.

test/validate/validate.go
jpayne3506: I definitely didn't realize until I started making the changes. I think @vipul-21 intent was for it to be a parent function to run all possible validations and gracefully handle unique scenarios. As such `TestValidateState` within `load_test.go` only calls `v.ValidateStateFile`. 

Not sure what comment you are looking for..? Do you mean above `Validate( )`?

.pipelines/templates/cilium-nightly-checks.yaml
jpayne3506: It slipped. Pipeline caught it as it hit the same error.

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e.steps.yaml
QxBytes: so if I had a parameter for a stage template, and then we had a template for a job, and a template for a task-- in the task I could use the stage template's parameters? wondering if it would be too much hassle to explicitly declare the parameters and pass them down into the template?

test/validate/validate.go
QxBytes: yeah like a comment for `Validate` mentioning that it validates the state file and restarts networkd

.pipelines/templates/validate-tests.yaml
```diff
@@ -0,0 +1,55 @@
...
+          # Scale Cluster Up/Down to confirm functioning CNS
+          ITERATIONS=2 SCALE_UP=${{ parameters.scaleup }} OS_TYPE=${{ parameters.os }} go test -count 1 -timeout 30m -tags load -run ^TestLoad$
+          kubectl get pods -owide -A
+
+          cd ../../..
+          echo "Validating Node Restart"
+          make test-validate-state OS_TYPE=${{ parameters.os }} RESTART_CASE=true CNI_TYPE=${{ parameters.cni }} CLEANUP=${{ parameters.cleanup }}
+      displayName: "Validate Node Restart"
+      retryCountOnTaskFailure: 3
+  - ${{ if contains(parameters.validate, 'networkdRestart') }}:
```
vipul-21, line 29: Should the order in which these validate step matter ? Earlier we were doing state validation and then networkd restart ?

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e.steps.yaml
jpayne3506: Just to make sure this covers all variations

> so if I had a parameter for a stage template, and then we had a template for a job, and a template for a task-- in the task I could use the stage template's parameters? 
- you cannot explicitly reference a given stage, job, task parameter. 
- a template call will take any parameter. Defined or not in the respective template that is being called
- If a given/expected parameter is not passed into the template then it will use the default parameters we define at the top of our template files. We leverage the defaults as empty strings as helpers to people using them to know that this template requires A,B,C,X,Y,Z parameters to function. 
https://github.com/Azure/azure-container-networking/blob/ab1c159b51969dee0a46ac49ae113db7fa7f2f78/.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml#L1-L7
- if you have 3 successive template calls A calls B, B calls C, C calls D.. that all pass the same `parameter.os`, then the 1st template call value will bubble all the way to the 3rd.
  - pass-ing here means that when using `- template: ` you explictily define what `parameter.os` is.
  - Note if the 2nd template call does not pass `parameter.os` then the 3rd template call will use the default of template file being called (if there even is one).  


> wondering if it would be too much hassle to explicitly declare the parameters and pass them down into the template?

I don't think it would be too much a hassle, but some parameters just don't need to be defined as the file name is descriptive enough to know that a given template should have a default value of `cilium` and `linux` for example. 

If we were to merge job/step templates into single files then we 100% should be descriptive as to what parameters we want to use.

.pipelines/cni/cilium/nightly-release-test.yml
jpayne3506: These tests are cleaning up namespace for the test suite before cilium can assign a CEP. The race is due to multiple parallel tests flooding the agent. The tests themselves have 0 to do with actual race conditions within cilium or the tests. It is just reporting as an error within cilium logs.

.pipelines/cni/k8s-e2e/k8s-e2e-job-template.yaml
jpayne3506: pretty much. Idea is to have each `processes` take care of potentially 2 tests as most test blocks only have at most 15-20 tests. I don't think we would benefit of having `(test count/processes) < 2`

.pipelines/templates/validate-tests.yaml
jpayne3506: There should be no impact. Best part of this template is that it is all conditional insertion. So, you can control the order by calling the template multiple times.

.pipelines/cni/cilium/nightly-release-test.yml
jpayne3506: Changed the processes from 1 back to 8 ( the default ) in favor of just not deleting the namespaces during testing.

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
```diff
@@ -29,32 +29,54 @@ steps:
...
+        addSpnToEnvironment: true
+        inlineScript: |
+          set -e
+          make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${{ parameters.clusterName }}
+          ls -lah
+          pwd
+          kubectl cluster-info
+          kubectl get po -owide -A
+          if [ "$CILIUM_VERSION_TAG" = "cilium-nightly-pipeline" ]; then
+            echo "Running nightly"
```
QxBytes, line 69: so we branch on the nightly variable, but we also have this branch here in the else (which seems to be for non-nightly pipeline runs). Are we concerned that this template could be called without nightly being properly specified? Is it complex to use the parameter passed in inside these scripts as a replacement for checking `CILIUM_VERSION_TAG` (also done below when we "Start Azilium E2E Tests on Overlay Cluster")

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e.steps.yaml
QxBytes: John to update the PR so we don't have a scenario where A calls B and B uses a parameter that A never passed in and B never set as a default/defined

.pipelines/cni/k8s-e2e/k8s-e2e-job-template.yaml
vipul-21: Can this be added to other pipelines too if not already present ? private/PR

.pipelines/singletenancy/azure-cni-overlay-stateless/azure-cni-overlay-stateless-e2e.steps.yaml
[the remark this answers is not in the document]
vipul-21: Why only windows ? Was this incorrect at the first place ? cc: @behzad-mir

.pipelines/singletenancy/azure-cni-overlay-stateless/azure-cni-overlay-stateless-e2e.steps.yaml
```diff
@@ -65,22 +65,12 @@ steps:
...
-        cd test/integration/load
-        clusterName=${{ parameters.clusterName }}
-        make -C ./hack/aks set-kubeconf AZCLI=az CLUSTER=${clusterName}
-        make -C ./hack/aks azcfg AZCLI=az REGION=$(REGION_AKS_CLUSTER_TEST)
-        kubectl get pods -owide -A
-        echo "Validating Node Restart"
-        CNI_TYPE=stateless RESTART_CASE=true go test -timeout 30m -tags load -run ^TestValidateState$
-    displayName: "Validate Node Restart"
-    retryCountOnTaskFailure: 3
+  - template: ../../templates/validate-tests.yaml
```
vipul-21, line 68: Now it does the load test and then run the noderestart etc. It should not impact anything but want to make sure we are aware of this change.

.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e-step-template.yaml
```diff
@@ -108,30 +131,47 @@ steps:
...
+            cat output.log
+
+            # Node Restart is expected to cause "Non-zero (1)" in cilium agent logs and
+            # a missing init container i.e. "install-cni-binaries in pod "cilium-fmjsb" not found"
+            if grep -q "Non-zero (1)" output.log; then
+              echo "Expected error log from Node Restart, proceed."
+              echo "Check for additional errors."
+
+              # Check for any other errors aside from node restart
+              if grep -v "Non-zero (1)" output.log | grep -v "previous terminated container" | grep "check-log-errors" | grep "Error"; then
```
vipul-21, line 163: `previous terminated container` what is the error line for this ?

vipul-21:
lgtm waiting for a green pipeline run

PR #4199, opened 2026-01-22
fix: init NNC from reconciler instead of directly

rbtr:
**Reason for Change**:
There's no clear reason that CNS initializes with an NNC that it gets out of band from the typical NNC reconciler flow.
This separate path for fetching the NNC leads to bugs such as ingesting an NNC for a previous Node of the same name and state corruption once the reconciler (which is smart enough to wait for the actual NNC for its Node) starts. 
Additionally, any caching, filtering, or other selection done to NNCs for efficiency is not applied or must be duplicated. 

Instead, initialization can be run from the first NNC that the reconciler is pushed. This consolidates all CNS NNC reads to a single location where list-watch optimizations and filtering can be centralized.

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/reconciler.go
+++ b/cns/kubecontroller/nodenetworkconfig/reconciler.go
@@ -27,8 +27,12 @@ type cnsClient interface {
 	MustEnsureNoStaleNCs(validNCIDs []string)
 }
 
+type nodenetworkconfigSink func(*v1alpha.NodeNetworkConfig) error
+
+// Deprecated: use nodenetworkconfigSink instead
 type nodeNetworkConfigListener interface {
-	Update(*v1alpha.NodeNetworkConfig) error
+	// Deprecated: use nodenetworkconfigSink instead
+	Update(*v1alpha.NodeNetworkConfig) error // phasing this out in favor of the sink
 }
 
 type nncGetter interface {
@@ -38,31 +42,33 @@ type nncGetter interface {
 // Reconciler watches for CRD status changes
 type Reconciler struct {
 	cnscli             cnsClient
-	ipampoolmonitorcli nodeNetworkConfigListener
+	ipampoolmonitorcli nodenetworkconfigSink
 	nnccli             nncGetter
 	once               sync.Once
-	started            chan interface{}
+	started            chan any
 	nodeIP             string
 	isSwiftV2          bool
+	initializer        nodenetworkconfigSink
 }
 
 // NewReconciler creates a NodeNetworkConfig Reconciler which will get updates from the Kubernetes
 // apiserver for NNC events.
 // Provided nncListeners are passed the NNC after the Reconcile preprocesses it. Note: order matters! The
 // passed Listeners are notified in the order provided.
-func NewReconciler(cnscli cnsClient, ipampoolmonitorcli nodeNetworkConfigListener, nodeIP string, isSwiftV2 bool) *Reconciler {
+func NewReconciler(cnscli cnsClient, initializer nodenetworkconfigSink, ipampoolmonitorcli nodeNetworkConfigListener, nodeIP string, isSwiftV2 bool) *Reconciler {
 	return &Reconciler{
 		cnscli:             cnscli,
-		ipampoolmonitorcli: ipampoolmonitorcli,
-		started:            make(chan interface{}),
+		ipampoolmonitorcli: ipampoolmonitorcli.Update,
+		started:            make(chan any),
 		nodeIP:             nodeIP,
 		isSwiftV2:          isSwiftV2,
+		initializer:        initializer,
 	}
 }
 
 // Reconcile is called on CRD status changes
 func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	listenersToNotify := []nodeNetworkConfigListener{}
+	listenersToNotify := []nodenetworkconfigSink{}
 	nnc, err := r.nnccli.Get(ctx, req.NamespacedName)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
@@ -88,6 +94,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		validNCIDs[i] = nnc.Status.NetworkContainers[i].ID
 	}
 	r.cnscli.MustEnsureNoStaleNCs(validNCIDs)
+
+	// call initFunc on first reconcile and never again
+	if r.initializer != nil {
+		if err := r.initializer(nnc); err != nil {
+			logger.Errorf("[cns-rc] initializer failed during reconcile: %v", err)
+			return reconcile.Result{}, errors.Wrap(err, "initializer failed during reconcile")
+		}
+		r.initializer = nil
+	}
 
 	// for each NC, parse it in to a CreateNCRequest and forward it to the appropriate Listener
 	for i := range nnc.Status.NetworkContainers {
@@ -134,7 +149,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 	// push the NNC to the registered NNC listeners.
 	for _, l := range listenersToNotify {
-		if err := l.Update(nnc); err != nil {
+		if err := l(nnc); err != nil {
 			return reconcile.Result{}, errors.Wrap(err, "nnc listener return error during update")
 		}
 	}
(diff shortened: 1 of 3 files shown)


cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -27,8 +27,10 @@ type cnsClient interface {
 	MustEnsureNoStaleNCs(validNCIDs []string)
 }
 
+type nodenetworkconfigSink func(*v1alpha.NodeNetworkConfig) error
+
 type nodeNetworkConfigListener interface {
-	Update(*v1alpha.NodeNetworkConfig) error
+	Update(*v1alpha.NodeNetworkConfig) error // phasing this out in favor of the sink
```
timraymond, line 35: Maybe add a `// Deprecated: use nodenetworkconfigSink`?

cns/service/main.go
```diff
@@ -1444,35 +1422,16 @@ func InitializeCRDState(ctx context.Context, z *zap.Logger, httpRestService cns.
...
-	// aks addons to come up so retry a bit more aggresively here.
-	// will retry 10 times maxing out at a minute taking about 8 minutes before it gives up.
-	attempt := 0
-	_ = retry.Do(func() error {
-		attempt++
-		logger.Printf("reconciling initial CNS state attempt: %d", attempt)
-		err = reconcileInitialCNSState(ctx, directscopedcli, httpRestServiceImplementation, podInfoByIPProvider, cnsconfig.EnableSwiftV2)
-		if err != nil {
-			logger.Errorf("failed to reconcile initial CNS state, attempt: %d err: %v", attempt, err)
-			nncInitFailure.Inc()
```
timraymond, line 1470: Do we still want to keep this counter around?

cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -89,6 +93,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	}
 	r.cnscli.MustEnsureNoStaleNCs(validNCIDs)
 
+	// call initFunc on first reconcile and never again
+	if r.initializer != nil {
+		if err := r.initializer(nnc); err != nil {
+			logger.Errorf("[cns-rc] initializer failed during reconcile: %v", err)
+			return reconcile.Result{}, errors.Wrap(err, "initializer failed during reconcile")
+		}
+		r.initializer = nil
```
timraymond, line 104: I think it's the same behavior for this to be wedged until initialization succeeds, but do we want to eventually circuit break? It seems like that was the original intention of the existing logic as I read the comment above it... whether that was true in practice may be dubious though because of swallowing the retry error and the `UntilSucceeded()`.

cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -89,6 +93,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	}
 	r.cnscli.MustEnsureNoStaleNCs(validNCIDs)
 
+	// call initFunc on first reconcile and never again
+	if r.initializer != nil {
+		if err := r.initializer(nnc); err != nil {
+			logger.Errorf("[cns-rc] initializer failed during reconcile: %v", err)
```
nddq, line 101: cns-rc? that's new 🙂

cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -38,31 +40,33 @@ type nncGetter interface {
...
 		nodeIP:             nodeIP,
 		isSwiftV2:          isSwiftV2,
+		initializer:        initializer,
 	}
 }
 
 // Reconcile is called on CRD status changes
 func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	listenersToNotify := []nodeNetworkConfigListener{}
+	listenersToNotify := []nodenetworkconfigSink{}
```
nddq, line 71: were there going to be any other listener other than the ipam pool monitor?

cns/kubecontroller/nodenetworkconfig/reconciler.go
rbtr: You read the prior intent correctly - we wanted to retry for a while and eventually fail out as a signal to the user that something was wrong. However, the retry was long enough that CNS wouldn't exit frequently enough to end up in CrashLoopBackoff and instead the Pod would just clock 5-6 restarts per hour.
Eventually someone might notice, but in practice this effectively was a retry-forever.

By tying this in to the Reconcile loop, we will be able to lean on the ctrlruntime machinery - it automatically does retry backoff and exposes metrics for failed reconciles which we can collect and use as a signal.

cns/service/main.go
rbtr: It will be implicitly counted in failed nnc reconciles

cns/kubecontroller/nodenetworkconfig/reconciler.go
rbtr: oops

cns/kubecontroller/nodenetworkconfig/reconciler.go
rbtr: Good point

cns/kubecontroller/nodenetworkconfig/reconciler.go
rbtr: there are two now 🤷

cns/kubecontroller/nodenetworkconfig/reconciler.go
rbtr: someone else started this actually, it's all over this file?

rbtr: maybe it was me https://github.com/Azure/azure-container-networking/commit/c970d07e8f17ea77ea7f20fee17a7225a856abd0#diff-694b7c4cfba89c7ad55426495e4aa08416a34859e2b01df8c5e140c2a9906db6

cns/service/main.go
```diff
@@ -1444,35 +1422,15 @@ func InitializeCRDState(ctx context.Context, z *zap.Logger, httpRestService cns.
...
+	initializerWrapper := func(nnc *v1alpha.NodeNetworkConfig) error {
+		logger.Printf("Reconciling initial CNS state")
+		if err := reconcileInitialCNSState(nnc, httpRestServiceImplementation, podInfoByIPProvider, cnsconfig.EnableSwiftV2); err != nil {
```
nairashu, line 1427: This used to be in a retrier. Wouldn't removing the retrier affect the initialization efforts here.

nairashu: The comment on the retrier seems that it was added intentionally.

// apiserver nnc might not be registered or api server might be down and crashloop backof puts us outside of 5-10 minutes we have for
	// aks addons to come up so retry a bit more aggresively here.
	// will retry 10 times maxing out at a minute taking about 8 minutes before it gives up.

cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -89,6 +95,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	}
 	r.cnscli.MustEnsureNoStaleNCs(validNCIDs)
 
+	// call initFunc on first reconcile and never again
+	if r.initializer != nil {
+		if err := r.initializer(nnc); err != nil {
+			logger.Errorf("[cns-rc] initializer failed during reconcile: %v", err)
```
nairashu, line 101: You have a few errors where logger.Errorf is mentioned as deprecated to using zap logger instead. Possibly check for those updates.

cns/service/main.go
nairashu: Now that you are relying on controller runtime's backoff will this affect Node Ready times and degrade performance at scale?

rbtr: Also mentioned in response to @timraymond earlier, but to clarify:
The explicit retry is no longer necessary.
 
This init is run during the first NNC reconcile; if it is unsuccessful we return an error and the NNC event gets requeued. We lean on the ctrlruntime machinery to do backoffs and retries for us.
 
There is a minor behavioral change - previously we would crash after a certain number of retries. Now it will requeue forever. Functionally this results in the same user experience - CNS never becomes ready, and no Pods ever get IPs.
 
The timeout-crash was intended to more visibly tell us something was broken, but it never worked out that way because we wouldn't crash fast enough (after retrying for 10+ minutes) to trigger any alerts or fail a rollout. This new way we have access to the same amount of observable data, it's just through the reconciliation failure metric instead.

rbtr: Additionally, there is a failure mode in the direct Get that we avoid here - if the NNC did not exist yet, the previous design would error, so we retry to give DNC/RC time to process our Node. Now, we open a watch and wait for that NNC event to get pushed to us.

vipul-21 (approved):

PR #4200, opened 2026-01-22
fix: only mark IPs Available for matching NCID

### LINKED ISSUE #4210: Replacing a Node with the same name in Overlay can corrupt CNS state
There's a state corruption bug in CNS when:
- The apiserver is overloaded so object garbage collection is delayed
- A node is deleted and replaced with one with the **same name**
- The CNI mode is Overlay, and the new Node gets the same secondary IP prefix as the old Node

The flow is roughly:
Node A is deleted. NNC A is queued for GC.
Node A' is created. NNC A exists, so NNC A' cannot be created yet.
CNS A' starts, reads NNC A. NNC A gets GC'd, NNC A' is created with the same secondary IP prefix. 
CNS tries to use the IPs from NNC A' but gets the ones from NNC A since they are the same set of IPs. This internal inconsistency causes IP assignment failure and this Node is unrecoverable.

<details><summary>More details</summary>

1) We get a stale NNC, and load the stale NC in to state, and the stale IPs in to PendingProgramming. 
2) We get a new/updated NNC. The new NC has the same IP addresses. We delete the stale NC from state, but not the stale IPs, which retain an association to the stale NC. 
3) We load the new NC into state and try to load the new IPs. However, CNS uniquely identifies IP by their address string in Overlay so we think they are already loaded.
4) The new NC is programmed, so we try to mark its IPs as available, but since we store them by address string, we accidentally mark the stale NC's IPs as available.
5) During the CNI IPAM assignment call, we assign an Available IP for the Pod. But when we try to change its state to Assigned, we have to look up its NC again, and that stale IP's NC has been delete from our state <- this is where we fail and return an Assignment error.
</details> 

There are several issues that stack to allow this to happen and need to be fixed:
1) During init we should use node UUID/ownerref to only load NNC for the actual Node we're running on.
2) When we purge a stale NC, we need to purge the IPs from that NC.
3) When we see an NC is programmed and mark IPs as available, we need to check the IP's NC association instead of only their ID string, and only mark the ones for that NC.
4) When we load IPs from the NNC, it's not good enough to ID them by address string. We can't fix the IP ID scheme today, but when we have an IP in state already from a different NC, we should error.

rbtr:
**Reason for Change**:

In block allocation modes, IPs are stored in CNS state not by an allocation UUID, but by their IP string as a key. This means that if a Node gets multiple NCs with overlapping blocks, when we program the new NC it is possible that the IPs in CNS state are for the wrong NC and we should not mark those as Available. This adds a last-minute sanity check to confirm that the IP is from the expected NC.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/ipam.go
+++ b/cns/restserver/ipam.go
@@ -673,6 +673,8 @@ func (service *HTTPRestService) MarkIpsAsAvailableUntransacted(ncID string, newH
 			for uuid, secondaryIPConfigs := range ncInfo.CreateNetworkContainerRequest.SecondaryIPConfigs {
 				if ipConfigStatus, exist := service.PodIPConfigState[uuid]; !exist {
 					logger.Errorf("IP %s with uuid as %s exist in service state Secondary IP list but can't find in PodIPConfigState", ipConfigStatus.IPAddress, uuid)
+				} else if ipConfigStatus.NCID != ncID {
+					logger.Errorf("IP %s with uuid as %s NCID %s mismatch with current processing NCID %s, skip updating its status", ipConfigStatus.IPAddress, uuid, ipConfigStatus.NCID, ncID)
 				} else if ipConfigStatus.GetState() == types.PendingProgramming && secondaryIPConfigs.NCVersion <= newHostNCVersion {
 					_, err := service.updateIPConfigState(uuid, types.Available, nil)
 					if err != nil {
--- a/cns/restserver/ipam.go
+++ b/cns/restserver/ipam.go
@@ -673,9 +673,7 @@ func (service *HTTPRestService) MarkIpsAsAvailableUntransacted(ncID string, newH
 			for uuid, secondaryIPConfigs := range ncInfo.CreateNetworkContainerRequest.SecondaryIPConfigs {
 				if ipConfigStatus, exist := service.PodIPConfigState[uuid]; !exist {
 					logger.Errorf("IP %s with uuid as %s exist in service state Secondary IP list but can't find in PodIPConfigState", ipConfigStatus.IPAddress, uuid)
-				} else if ipConfigStatus.NCID != ncID {
-					logger.Errorf("IP %s with uuid as %s NCID %s mismatch with current processing NCID %s, skip updating its status", ipConfigStatus.IPAddress, uuid, ipConfigStatus.NCID, ncID)
-				} else if ipConfigStatus.GetState() == types.PendingProgramming && secondaryIPConfigs.NCVersion <= newHostNCVersion {
+				} else if ipConfigStatus.NCID == ncID && ipConfigStatus.GetState() == types.PendingProgramming && secondaryIPConfigs.NCVersion <= newHostNCVersion {
 					_, err := service.updateIPConfigState(uuid, types.Available, nil)
 					if err != nil {
 						logger.Errorf("Error updating IPConfig [%+v] state to Available, err: %+v", ipConfigStatus, err)
--- a/cns/restserver/ipam_test.go
+++ b/cns/restserver/ipam_test.go
@@ -1427,6 +1427,30 @@ func TestIPAMMarkIPAsPendingWithPendingProgrammingIPs(t *testing.T) {
 	}
 }
 
+func TestMarkIpsAsAvailableUntransactedSkipsMismatchedNCID(t *testing.T) {
+	svc := getTestService(cns.KubernetesCRD)
+	secondaryIPConfigs := map[string]cns.SecondaryIPConfig{
+		testIP1: {
+			IPAddress: testIP1,
+			NCVersion: 0,
+		},
+	}
+
+	createAndValidateNCRequest(t, secondaryIPConfigs, testNCID, "0")
+	createAndValidateNCRequest(t, secondaryIPConfigs, testNCIDv6, "0")
+
+	ncInfo := svc.state.ContainerStatus[testNCIDv6]
+	ncInfo.HostVersion = "0"
+	svc.state.ContainerStatus[testNCIDv6] = ncInfo
+
+	svc.MarkIpsAsAvailableUntransacted(testNCIDv6, 1)
+
+	ipState := svc.PodIPConfigState[testIP1]
+	assert.Equal(t, types.PendingProgramming, ipState.GetState())
+	assert.Equal(t, testNCID, ipState.NCID)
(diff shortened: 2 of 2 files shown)


cns/restserver/ipam.go
```diff
@@ -658,33 +658,45 @@ func (service *HTTPRestService) updateIPConfigState(ipID string, updatedState ty
...
-					ncInfo.CreateNetworkContainerRequest.SecondaryIPConfigs[uuid] = secondaryIPConfigs
-					logger.Printf("Change ip %s with uuid %s from pending programming to %s, current secondary ip configs is %+v", ipConfigStatus.IPAddress, uuid, types.Available,
-						ncInfo.CreateNetworkContainerRequest.SecondaryIPConfigs[uuid])
-				}
-			}
+			logger.Errorf("Error updating IPConfig [%+v] state to Available, err: %+v", ipConfigStatus, err)
 		}
+		secondaryIPConfigs.NCVersion = newHostNCVersion
+		ncInfo.CreateNetworkContainerRequest.SecondaryIPConfigs[uuid] = secondaryIPConfigs
+		logger.Printf("Change ip %s with uuid %s from pending programming to %s, current secondary ip configs is %+v", ipConfigStatus.IPAddress, uuid, types.Available, ncInfo.CreateNetworkContainerRequest.SecondaryIPConfigs[uuid])
```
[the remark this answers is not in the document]
rbtr: actually it was wrong before, too

cns/restserver/ipam.go
```diff
@@ -658,33 +658,45 @@ func (service *HTTPRestService) updateIPConfigState(ipID string, updatedState ty
...
-		previousHostNCVersion, err := strconv.Atoi(ncInfo.HostVersion)
+		return
+	}
+	previousHostNCVersion, err := strconv.Atoi(ncInfo.HostVersion)
+	if err != nil {
+		logger.Printf("[MarkIpsAsAvailableUntransacted] Get int value from ncInfo.HostVersion %s failed: %v, can't proceed", ncInfo.HostVersion, err)
+		return
+	}
+	// We only need to handle the situation when dnc nc version is larger than programmed nc version
+	if previousHostNCVersion < newHostNCVersion {
```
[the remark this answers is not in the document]
rbtr: the conditional was flipped correctly

cns/restserver/ipam.go
```diff
@@ -658,33 +658,43 @@ func (service *HTTPRestService) updateIPConfigState(ipID string, updatedState ty
...
+	}
+	for uuid, secondaryIPConfigs := range ncInfo.CreateNetworkContainerRequest.SecondaryIPConfigs {
+		ipConfigStatus, exist := service.PodIPConfigState[uuid]
+		if !exist {
+			logger.Errorf("IP %s with uuid as %s exist in service state Secondary IP list but can't find in PodIPConfigState", secondaryIPConfigs.IPAddress, uuid)
+			continue
+		}
+		// In block allocations, IPs do not have UUIDs, so the ID is actually the IP string. In these scenarios, we need to double check that
+		// the IP is the one for the NC we are processing in case there is stale IP data in PodIPConfigState from a previous NC.
+		if ipConfigStatus.NCID != ncID {
```
paulyufan2: I think we should check both exist and ipConfigStatus.NCID != ncID

timraymond (dismissed):

cns/restserver/ipam.go
rbtr: the previous block checks !exist and continues to the next loop

cns/restserver/ipam.go
```diff
@@ -658,33 +658,43 @@ func (service *HTTPRestService) updateIPConfigState(ipID string, updatedState ty
...
+	previousHostNCVersion, err := strconv.Atoi(ncInfo.HostVersion)
+	if err != nil {
+		logger.Printf("[MarkIpsAsAvailableUntransacted] Get int value from ncInfo.HostVersion %s failed: %v, can't proceed", ncInfo.HostVersion, err)
+		return
+	}
+	// We only need to handle the situation when dnc nc version is larger than programmed nc version
+	if previousHostNCVersion >= newHostNCVersion {
+		return
+	}
+	for uuid, secondaryIPConfigs := range ncInfo.CreateNetworkContainerRequest.SecondaryIPConfigs {
```
QxBytes: it seems like it's possible the "key" here is a pod ip string-- can we change the name to reflect this possibility? Also if I recall the linked issue it mentions the bug is contained to overlay mode, but it seems like this pod ip string key only comes into play in vnet block mode, unless that isn't the case? Also an aside-- if the SecondaryIPConfigs key is not always the uuid can we change the comment in the `CreateNetworkContainerRequest` struct-- I think it currently says `// uuid is key`.

QxBytes (commented):
Wanted to confirm also the issue this guards against is: 
if the persisted state (ContainerStatus) has 2 NCs NC A and NC A' with the same IP. Then, for that IP key in the in-memory state (PodIPConfigState) we have NC ID A. Then when we try to "sync"/transfer the persisted state of NC A' to the slot in the in-memory state for the IP key, it skips/gets rejected with this PR, since in-memory[IP key] has data from NC A, not NC A' (even though both NCs can refer to the same IP key in-memory).

rbtr:
> Wanted to confirm also the issue this guards against is: if the persisted state (ContainerStatus) has 2 NCs NC A and NC A' with the same IP. Then, for that IP key in the in-memory state (PodIPConfigState) we have NC ID A. Then when we try to "sync"/transfer the persisted state of NC A' to the slot in the in-memory state for the IP key, it skips/gets rejected with this PR, since in-memory[IP key] has data from NC A, not NC A' (even though both NCs can refer to the same IP key in-memory).

Correct. We've previously loaded IPs from NC A. Now we're marking Available IPs from NC A' (but they're the same IPs). Here, we check which NC the IP is from and will fail.

cns/restserver/ipam.go
rbtr: In any static prefix modes (definitely overlay, maybe vnet block), the key _is_ the IP string, and that's why we aren't aware that the IP is from a different NC. 
In podsubnet, which is what this code was originally written for, the secondary IPs are individually attached to the NC and have their own UUIDs, so getting a new NC would necessarily mean a new set of uniquely identifiable IPs.

rbtr: I have a change staged to update the reference from UUID to just ID but it is a wider reaching change than only in this func

QxBytes (dismissed):

nairashu (dismissed):

PR #4203, opened 2026-01-23
ci: update cilium version to 1.17 and decouple from pipeline vars

QxBytes:
**Reason for Change**:

The pipeline currently runs cilium 1.14 during testing. This is inaccurate, as our e2e cluster k8s version is 1.33, which should use cilium 1.17. The cilium version is currently controlled by a pipeline variable set to 1.14. However, we cannot just change it to 1.17 because the pipeline variable affects ALL runs of the pipeline for all release branches, including release/v1.6, release/v1.5 etc. which still need to be run with cilium 1.14. The solution is to control the cilium version from the state of the repo/code, rather than through a pipeline variable. To do this, we unset the pipeline variable and use the makefile to set the default cilium version. This is similar to how we set the default k8s version to 1.33 on this release branch, but on other release branches the makefile sets it to an earlier k8s version. Thus, going forward, each release branch can have its own compatible cilium version.

**Issue Fixed**:

Runs the tests against 1.17 and re-enables ebpf host routing tests since they should now always run since the cilium version is 1.17+

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Does not affect ebpf host routing deployment or targets
.pipelines/cni/cilium/cilium-overlay-load-test-template.yaml put through a diff checker-- most of the delta is from CRLF -> LF
Check these runs:
ACN PR: https://msazure.visualstudio.com/One/_build/results?buildId=150406014&view=results
ACN PR 2: https://msazure.visualstudio.com/One/_build/results?buildId=150280427&view=results
CNI Release: https://msazure.visualstudio.com/One/_build/results?buildId=150285649&view=results
ACN Test Unofficial: https://dev.azure.com/msazure/One/_build/results?buildId=150406265&view=results
Nightly: https://msazure.visualstudio.com/One/_build/results?buildId=150456849&view=results

### DIFF
--- a/.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e.steps.yaml
+++ b/.pipelines/singletenancy/cilium-overlay/cilium-overlay-e2e.steps.yaml
@@ -40,7 +40,6 @@ steps:
         if [ "$CILIUM_VERSION_TAG" = "cilium-nightly-pipeline" ]; then
           echo "Running nightly"
           # makefile manages cilium version (nightly doesn't use DIR)
-          unset DIR
           unset CILIUM_VERSION_TAG
           make -C ./hack/aks deploy-cilium-nightly
         else


.pipelines/singletenancy/cilium/cilium-e2e.steps.yaml
```diff
@@ -33,9 +33,9 @@ steps:
         pwd
         kubectl cluster-info
         kubectl get po -owide -A
-        export DIR=$(echo ${CILIUM_VERSION_TAG#v} | cut -d. -f1,2)
-        # cilium version tag is already exported
-        echo "installing cilium version ${CILIUM_VERSION_TAG} from directory ${DIR}"
+        # makefile manages cilium version
+        unset DIR
```
vipul-21, line 37: Why do we have to unset `DIR` or `CILIUM_VERSION_TAG` ? Are they still being assigned the values in the pipelines somewhere ? Being safe before calling the makefile ?

vipul-21 (approved):

.pipelines/singletenancy/cilium/cilium-e2e.steps.yaml
QxBytes: Yes so because `DIR` and `CILIUM_VERSION_TAG` are variables in the ADO pipeline GUI (pipeline runtime variables), they are automatically populated as env variables in each script, even when set to "". Because they are technically set (*to the empty string), make will not use the makefile defaults, and will instead use "" as the `DIR` and "" as the `CILIUM_VERSION_TAG`. When I unset these variables, make's `?=` behavior sets them to the proper defaults (like `DIR=1.17` for example)

PR #4206, opened 2026-01-29
fix: add windows uts in pipeline for cns folder

QxBytes:
**Reason for Change**:

Unit tests in the cns folder for windows are not being run. This PR either makes them linux specific or fixes them. Then, it runs the uts in the pipeline. The behavior of CNS should not change with this PR.

Notable changes:
fsnotify_test: use temp directories to avoid elevated perms
conversion_test: move one unit test to linux specific file
conversion_windows_test: change expected ip for vnet block from primary to node ip (CHECK)
api_test: move unit test to linux specific file
ipam_test: move unit test to linux specific file
util_test: move unit test to linux specific file
routes_test: gate behind "pipeline" build tag so it does not run (pipeline build tag will be used later, but right now we don't run this ut because it requires elevated permissions)
service_test: pem file must be base64 encoded on windows, and dpapi can only be called from a windows specific test file (so service_linux_test and service_windows_test are created
podinfoprovider_test: fixed-- the test expects the windows cni binary path on windows
statefile_test: use filepath instead of path (path is only compatible with "/" paths)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
Run: https://msazure.visualstudio.com/One/CCE%20Apps/_build/results?buildId=151161946&view=logs&s=a76b775d-1ea1-5d2d-57a9-c68a58bed41d&j=1e95c5bf-f785-5c1d-a4c1-33b038e882fb

### DIFF
--- a/cns/service_windows_test.go
+++ b/cns/service_windows_test.go
@@ -6,13 +6,11 @@ import (
 
 // postProcessPEMCert encrypts the pem to base64 with dpapi on windows
 func postProcessPEMCert(pem []byte) ([]byte, error) {
-	var fileContent []byte
 	// On Windows, the TLS certificate retriever expects DPAPI-encrypted content
 	encrypted, encryptErr := dpapi.Encrypt(string(pem))
 	if encryptErr != nil {
 		return nil, encryptErr
 	}
-	fileContent = []byte(encrypted)
 
-	return fileContent, nil
+	return []byte(encrypted), nil
 }


cns/service_windows_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+// postProcessPEMCert encrypts the pem to base64 with dpapi on windows
+func postProcessPEMCert(pem []byte) ([]byte, error) {
+	var fileContent []byte
+	// On Windows, the TLS certificate retriever expects DPAPI-encrypted content
+	encrypted, encryptErr := dpapi.Encrypt(string(pem))
+	if encryptErr != nil {
+		return nil, encryptErr
+	}
+	fileContent = []byte(encrypted)
```
jackieluc: nit: with this approach, we don't really need `fileContent` anymore

rbtr (dismissed):

PR #4208, opened 2026-01-29
feat: add cilium log collector

QxBytes:
**Reason for Change**:

Adds cilium log collector image which is based on mcr fluent bit, with the added app insights output plugin. The plugin takes in an instrumentation key and enriches it with information from the imds, pushing it to app insights.

Moves cilium to server side apply so we can test the cilium log collector by applying it to the cilium daemonset (does not affect ebpf host routing since that might break a dependent pipeline).

To add cilium log collector to any cilium cluster, should be as simple as setting the log collector's image registry and version, and then running `make -C ./hack/aks add-cilium-log-collector`

pipeline.yaml put through a text diff checker the majority is a CRLF -> LF change

Builds cilium log collector in cni release, acn pr, and acn official build

Puts makefile targets in a separate included makefile. For the image, the dockerfile `$VERSION` is passed in from `$TAG` which is actually `$(CILIUM_LOG_COLLECTOR_PLATFORM_TAG)`. Added the log collector to ACN VERSION exclude, and added it to the all binaries target

Cilium log collector is only for linux

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
Round 1:
ACN PR Pass: https://msazure.visualstudio.com/One/_build/results?buildId=151173566&view=results
CNI Release Test Mostly Pass: https://msazure.visualstudio.com/One/_build/results?buildId=151174815&view=results
ACN Test Unofficial Pass: https://dev.azure.com/msazure/One/_build/results?buildId=151174219&view=results
Cilium Nightly Mostly Pass: https://msazure.visualstudio.com/One/_build/results?buildId=151143234&view=results
Cilium private 1.18 w/ upgrade doesn't pass on upgrade but that's not related this change (even without that ebpf stage fails): https://msazure.visualstudio.com/One/_build/results?buildId=151451236&view=results 
You can also confirm the image is being deployed by clicking on the "install cilium" job under each e2e stage

Round 2 After Merge Master In:
ACN PR: https://msazure.visualstudio.com/One/_build/results?buildId=151885712&view=results
CNI Release Test: https://msazure.visualstudio.com/One/_build/results?buildId=151887115&view=results
Cilium Nightly: https://msazure.visualstudio.com/One/_build/results?buildId=151885844&view=logs&j=31b419ed-604a-581c-8f21-7ebe3bcfa833
ACN Unofficial currently broken even on master during build


test/integration/manifests/cilium/v1.17/cilium-log-collector/cilium-log-collector-configmap.yaml
```diff
@@ -0,0 +1,46 @@
...
+        K8S-Logging.Parser On
+        K8S-Logging.Exclude Off
+        Buffer_Size 128KB
+        Use_tag_for_meta On
+
+    [OUTPUT]
+        Name                        azure_app_insights
+        debug                       true
+        imds                        true
+        instrumentation_key         aad482c8-f408-48e9-9380-e37fdb9ef4a3
```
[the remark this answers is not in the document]
QxBytes: not a secret

test/integration/manifests/cilium/v1.18/cilium-log-collector/cilium-log-collector-configmap.yaml
```diff
@@ -0,0 +1,46 @@
...
+        K8S-Logging.Parser On
+        K8S-Logging.Exclude Off
+        Buffer_Size 128KB
+        Use_tag_for_meta On
+
+    [OUTPUT]
+        Name                        azure_app_insights
+        debug                       true
+        imds                        true
+        instrumentation_key         aad482c8-f408-48e9-9380-e37fdb9ef4a3
```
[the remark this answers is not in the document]
QxBytes: not a secret

cilium-log-collector/out_azure_app_insights.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+var (
+	client       appinsights.TelemetryClient
+	debug        string
+	logKey       string
+	hostMetadata *common.Metadata
+)
+
+func convertToString(v interface{}) string {
```
santhoshmprabhu, line 65: Do you need to define this? I'd assume there's a utility function for this.

cilium-log-collector/out_azure_app_insights.go
```diff
@@ -0,0 +1,246 @@
...
+	imds := output.FLBPluginConfigKey(plugin, "imds")
+	fmt.Printf("[flb-azure-app-insights] version = '%s'\n", version)
+	fmt.Printf("[flb-azure-app-insights] plugin instrumentation key = '%s'\n", instrumentationKey)
+	fmt.Printf("[flb-azure-app-insights] using log key = '%s'\n", logKey)
+	fmt.Printf("[flb-azure-app-insights] debug = '%s'\n", debug)
+	fmt.Printf("[flb-azure-app-insights] imds = '%s'\n", imds)
+
+	telemetryConfig := appinsights.NewTelemetryConfiguration(instrumentationKey)
+	telemetryConfig.MaxBatchInterval = 1 * time.Second
+	telemetryConfig.MaxBatchSize = 10
```
santhoshmprabhu, line 138: What does this set? Batch size in MB? Please add comments.

cilium-log-collector/out_azure_app_insights.go
```diff
@@ -0,0 +1,246 @@
...
+//
+//export FLBPluginInit
+func FLBPluginInit(plugin unsafe.Pointer) int {
+	instrumentationKey := output.FLBPluginConfigKey(plugin, "instrumentation_key")
+	// the key that is identified as the log upon receiving the record in this plugin
+	logKey = output.FLBPluginConfigKey(plugin, "log_key")
+	if logKey == "" {
+		logKey = "log"
+	}
+	debug = output.FLBPluginConfigKey(plugin, "debug")
```
santhoshmprabhu, line 127: This is to debug the plugin itself, unrelated to the logs, correct?

cilium-log-collector/out_azure_app_insights.go
```diff
@@ -0,0 +1,246 @@
...
+			Fields:    record,
+		}, count, hostMetadata)
+		count++
+	}
+
+	return output.FLB_OK
+}
+
+// ProcessSingleRecord handles processing of an individual record
+func (rp *RecordProcessor) ProcessSingleRecord(record ProcessRecord, recordIndex int, metadata *common.Metadata) {
```
santhoshmprabhu, line 203: I recall you mentioned reusing the AppInsights logic from CNI. Is the code duplicated? If yes, it'd be good to move to a common location and reuse.

cilium-log-collector/out_azure_app_insights.go
santhoshmprabhu: Also, any chance of storing the filename here, to catch the exact container whose logs are being collected?

test/integration/manifests/cilium/v1.17/cilium-log-collector/daemonset-patch.yaml
```diff
@@ -0,0 +1,36 @@
+# Note: Daemonset must be server side applied prior. Then just do: kubectl apply --server-side daemonset-patch.yaml
```
santhoshmprabhu: I believe you do need to set the field-manager to log-collector, otherwise you might overwrite. I think that's how you are applying in this PR, but this comment may cause an unfamiliar person to try just apply, without the filed-manager setting. Please verify manual apply, and update the above comment if necessary.

test/integration/manifests/cilium/v1.18/cilium-log-collector/daemonset-patch.yaml
```diff
@@ -0,0 +1,36 @@
+# Note: Daemonset must be server side applied prior. Then just do: kubectl apply --server-side daemonset-patch.yaml
```
santhoshmprabhu: Same comment from the 1.17 file.

test/integration/manifests/cilium/v1.17/cilium-log-collector/daemonset-patch.yaml
QxBytes: yeah true-- maybe I'll just instruct them to use the make target instead

cilium-log-collector/out_azure_app_insights.go
QxBytes: I needed some specific behavior that the built-in functions don't provide-- for example the conversion would usually be just like Sprintf("%v", val) which is the default case here, but that by itself prints out bytes or map[field: value] which doesn't show up/parse well in azure data explorer/kusto. Also straight marshaling into a string doesn't work if the "key" value isn't a string.

cilium-log-collector/out_azure_app_insights.go
QxBytes: This is for debugging the plugin, yes. 

If we wanted a specific container's logs, we can modify the input section of the fluent bit config to regex match on it

If you're referring to storing something identifying the file the log came from, each log collected should be already enriched with the container name

cilium-log-collector/out_azure_app_insights.go
QxBytes: I am reusing the code to get the IMDS metadata so that part isn't duplicated. The logic to send the logs out doesn't need to be as complex as the CNI-- the below for loop is meant to parse through the record's fields and get them into the custom fields map. This would need to be done regardless.

The populating of the custom fields with metadata.* I believe is clearer than what we do in `aitelemetry/telemetrywrapper.go/TrackLog`, and the metadata fields we use here are different from TrackLog.

We then directly call new trace telemetry. Although you could import and use NewAITelemetry or similar, you would just be replacing this 4-5 line call to NewTraceTelemetry + Track with something of similar complexity.

cilium-log-collector/out_azure_app_insights.go
QxBytes: Maximum number of telemetry items that can be submitted in each request-- I'll add a comment

camrynl (dismissed):
discussed some questions offline for log storage. ~168 hours or potentially longer and no logs should be missed since this will go back to the start of the container

santhoshmprabhu:
The changes look good, I am wondering if we should call it "acn-log-collector", and only the fluent config makes it a "cilium" log collector. There is nothing in any other file that makes it cilium-specific, right?

santhoshmprabhu (approved):

PR #4212, opened 2026-02-03
Feat: necessary fixes for Stateless CNI Linux

behzad-mir:
A number of changes is made to stateless CNI to fully support SWiftV2 in Linux:

1.  For Stateless CNI, the EndpointID should include ifName when the NicType is Delegated. This ensures it can be distinguished from the InfraNIC endpoint. The reason for this behavior is that Stateless CNI uses only the ContainerID as the endpoint ID when saving the state file in CNS, to maintain consistency with Cilium. However, within CNI’s in-memory representation, the EndpointID must uniquely identify each NIC. As a result:
    - We always pass the ContainerID to CNS as the endpoint state key.
    - Internally, CNI uses EndpointID (which includes ifName for delegated NICs) to differentiate endpoints. 
8.  Delete flow has been revised and  NetNSPath has been added to the epinfo since it is needed by the TransparentClient for Frontend NIC. This is needed to move the secondary interface out of the pod namespace.
9.  The Transparent mode used by statefull CNI for SWiftV1 and V2 and stateless CNI should follow the same. TransparentVlan which is the original value seems to be a mistake.
10.  Async delete file creation in Stateless CNI has been moved to CNS_invoker to be consistent with Stateful CNI.
11.  The DeleteState function has been modified to only applies for Windows SwiftV2 scenario and do not affect the SWifv2 Linux since the delete flow is different in Linux and Windows.

For validating the scenario ADD/Delete calls have been issues on SwiftV2 cluster and and logs and satefile has been analyzed to make sure it is consistent with Stateful CNI and also nothing gets leaked. 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -18,7 +18,6 @@ import (
 	"github.com/Azure/azure-container-networking/cni/util"
 	"github.com/Azure/azure-container-networking/cns"
 	cnscli "github.com/Azure/azure-container-networking/cns/client"
-	"github.com/Azure/azure-container-networking/cns/fsnotify"
 	"github.com/Azure/azure-container-networking/common"
 	"github.com/Azure/azure-container-networking/dhcp"
 	"github.com/Azure/azure-container-networking/iptables"
@@ -719,7 +718,7 @@ func (plugin *NetPlugin) createEpInfo(opt *createEpInfoOpt) (*network.EndpointIn
 		*opt.infraSeen = true
 	} else {
 		ifName = "eth" + strconv.Itoa(opt.endpointIndex)
-		endpointID = plugin.nm.GetEndpointID(opt.args.ContainerID, ifName)
+		endpointID = plugin.nm.GetEndpointIDByNicType(opt.args.ContainerID, ifName, opt.ifInfo.NICType)
 	}
 
 	endpointInfo := network.EndpointInfo{
@@ -1046,13 +1045,15 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 	networkID, err = plugin.getNetworkID(args.Netns, nil, nwCfg)
 	if nwInfo, err = plugin.nm.GetNetworkInfo(networkID); err != nil {
 		if !nwCfg.MultiTenancy {
-			logger.Error("Failed to query network",
-				zap.String("network", networkID),
-				zap.Error(err))
 			// Log the error if the network is not found.
 			// if cni hits this, mostly state file would be missing and it can be reboot scenario where
 			// container runtime tries to delete and create pods which existed before reboot.
-			// this condition will not apply to stateless CNI since the network struct will be crated on each call
+			// this condition will not apply to stateless CNI since the network struct will be created on Delete calls
+			if !plugin.nm.IsStatelessCNIMode() {
+				logger.Error("Failed to query network",
+					zap.String("network", networkID),
+					zap.Error(err))
+			}
 			err = nil
 		}
 	}
@@ -1075,33 +1076,23 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
 	if plugin.nm.IsStatelessCNIMode() {
 		// network ID is passed in and used only for migration
 		// otherwise, in stateless, we don't need the network id for deletion
-		epInfos, err = plugin.nm.GetEndpointState(networkID, args.ContainerID)
-		// if stateless CNI fail to get the endpoint from CNS for any reason other than  Endpoint Not found
-		if err != nil {
-			// async delete should be disabled for standalone scenario
-			if errors.Is(err, network.ErrConnectionFailure) && !nwCfg.DisableAsyncDelete {
-				logger.Info("failed to connect to CNS", zap.String("containerID", args.ContainerID), zap.Error(err))
-				addErr := fsnotify.AddFile(args.ContainerID, args.ContainerID, watcherPath)
-				logger.Info("add containerid file for Asynch delete", zap.String("containerID", args.ContainerID), zap.Error(addErr))
-				if addErr != nil {
-					logger.Error("failed to add file to watcher", zap.String("containerID", args.ContainerID), zap.Error(addErr))
-					return errors.Wrap(addErr, fmt.Sprintf("failed to add file to watcher with containerID %s", args.ContainerID))
-				}
-				return nil
-			}
-			if errors.Is(err, network.ErrEndpointStateNotFound) {
-				logger.Info("Endpoint Not found", zap.String("containerID", args.ContainerID), zap.Error(err))
-				return nil
-			}
+		epInfos, err = plugin.nm.GetEndpointState(networkID, args.ContainerID, args.Netns)
+		// if stateless CNI fails to get the endpoint from CNS for any reason other than
+		// endpoint not found or connection failure, return error
+		if err != nil && !errors.Is(err, network.ErrEndpointStateNotFound) && !errors.Is(err, network.ErrConnectionFailure) {
 			logger.Error("Get Endpoint State API returned error", zap.String("containerID", args.ContainerID), zap.Error(err))
 			return plugin.RetriableError(fmt.Errorf("failed to delete endpoint: %w", err))
 		}
+		// set the error to nil if endpoint state is not found or connection failure to CNS;
+		// the next if block (len(epInfos) == 0) will handle that ip release if necessary
+		err = nil
 	} else {
 		epInfos = plugin.nm.GetEndpointInfosFromContainerID(args.ContainerID)
 	}
 
 	// for when the endpoint is not created, but the ips are already allocated (only works if single network, single infra)
-	// this block is not applied to stateless CNI
+	// this block applies to stateless CNI if either endpoint state is not found or there is a CNS connection failure
+	// if there is a connection failure to CNS, IP release will be handled asynchronously by the CNS invoker (via ipamInvoker.Delete)
 	if len(epInfos) == 0 {
 		endpointID := plugin.nm.GetEndpointID(args.ContainerID, args.IfName)
 		if !nwCfg.MultiTenancy {
--- a/network/manager.go
+++ b/network/manager.go
@@ -120,7 +120,8 @@ type NetworkManager interface {
 	SaveState(eps []*endpoint) error
 	DeleteState(epInfos []*EndpointInfo) error
 	GetEndpointInfosFromContainerID(containerID string) []*EndpointInfo
-	GetEndpointState(networkID, containerID string) ([]*EndpointInfo, error)
+	GetEndpointState(networkID, containerID, netns string) ([]*EndpointInfo, error)
+	GetEndpointIDByNicType(containerID, ifName string, nicType cns.NICType) string
 }
 
 // Creates a new network manager.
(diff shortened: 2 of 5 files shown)


cni/network/network.go
```diff
@@ -1075,24 +1076,9 @@ func (plugin *NetPlugin) Delete(args *cniSkel.CmdArgs) error {
...
-				}
-				return nil
-			}
-			if errors.Is(err, network.ErrEndpointStateNotFound) {
-				logger.Info("Endpoint Not found", zap.String("containerID", args.ContainerID), zap.Error(err))
-				return nil
-			}
+		epInfos, err = plugin.nm.GetEndpointState(networkID, args.ContainerID, args.Netns)
+		// if stateless CNI fail to get the endpoint from CNS for any reason other than Endpoint Not found or connection failure, return error
+		if err != nil && !errors.Is(err, network.ErrEndpointStateNotFound) && !errors.Is(err, network.ErrConnectionFailure) {
```
QxBytes, line 1082: why do we no longer have DisableAsyncDelete? As a side note: we don't have AddFile anymore because we just pass through in all cases and have cns invoker add the file? Please confirm if that is the case.

network/manager.go
```diff
@@ -768,12 +774,12 @@ func (nm *networkManager) DeleteState(epInfos []*EndpointInfo) error {
...
 
-	// For AKS stateless cni, plugin.ipamInvoker.Delete takes care of removing the state in the main Delete function.
+	// For AKS stateless cni Windows, plugin.ipamInvoker.Delete takes care of removing the state in the main Delete function.
 	// For swiftv2 stateless cni, this call will delete the endpoint state from CNS.
 	if nm.IsStatelessCNIMode() {
 		for _, epInfo := range epInfos {
-			// this cleanup happens only for standalone swiftv2 to delete endpoint state from CNS.
-			if epInfo.NICType == cns.NodeNetworkInterfaceFrontendNIC || epInfo.NICType == cns.NodeNetworkInterfaceAccelnetFrontendNIC {
+			// this cleanup happens only for standalone swiftv2 Windows to delete endpoint state from CNS.
+			if epInfo.NICType == cns.NodeNetworkInterfaceFrontendNIC || epInfo.NICType == cns.NodeNetworkInterfaceAccelnetFrontendNIC && platform.GetOSInfo() == "windows" {
```
QxBytes: so I recall you mentioned that when async delete is **disabled**, like it is in windows ASI, we can run this special block in plugin.nm.DeleteState because by that point, we will know CNS is up (otherwise it would have failed and retried much earlier in the delete call). 

Can we check in DeleteState that async delete is disabled, and if so, run the `if epInfo.NICType == cns.NodeNetworkInterfaceFrontendNIC || epInfo.NICType == cns.NodeNetworkInterfaceAccelnetFrontendNIC {... etc.` instead of checking if the os is windows? 

My understanding is that if async delete is disabled, in order to reach DeleteState, CNS must be up and Stateless is true; otherwise, it would fail when we try to get the endpoint infos since we go down the Stateless is true path.

Related: https://teams.microsoft.com/l/message/19:96023851-2aa2-4837-8647-6c24ac5799d9_e43e1ced-8231-4014-9d98-dab7031b16bd@unq.gbl.spaces/1769211669164?context=%7B%22contextType%22%3A%22chat%22%7D

network/secondary_endpoint_client_linux.go
```diff
@@ -191,7 +192,12 @@ func (client *SecondaryEndpointClient) DeleteEndpoints(ep *endpoint) error {
 			logger.Error("Failed to exit netns with", zap.Error(newErrorSecondaryEndpointClient(err)))
 		}
 	}()
-	// TODO: For stateless cni linux, check if delegated vmnic type, and if so, delete using this *endpoint* struct's ifname
+	// For stateless cni linux, check if delegated vmnic type, and if so, delete using this *endpoint* struct's ifname
+	if ep.NICType == cns.NodeNetworkInterfaceFrontendNIC {
```
QxBytes, line 196: it's not quite deleting here though? Seems more like it's moving it back to the vm namespace. Also why are you moving it back-- didn't you mention that in containerd (1.7.30)+newer AKS the NIC will move back to vm after containerd deletes netns automatically?

network/manager.go
```diff
@@ -456,18 +457,21 @@ func validateUpdateEndpointState(endpointID string, ifNameToIPInfoMap map[string
...
 	if err != nil {
+		emptyEpInfos := []*EndpointInfo{}
 		if endpointResponse.Response.ReturnCode == types.NotFound {
-			return nil, ErrEndpointStateNotFound
+			logger.Info("Endpoint Not found", zap.String("containerID", containerID), zap.Error(err))
+			return emptyEpInfos, ErrEndpointStateNotFound
 		}
 		if endpointResponse.Response.ReturnCode == types.ConnectionError {
-			return nil, ErrConnectionFailure
+			logger.Info("failed to connect to CNS", zap.String("containerID", containerID), zap.Error(err))
```
QxBytes, line 469: nit: can we capitalize logger messages and keep casing consistent "Endpoint not found"?

cni/network/network.go
QxBytes: also probably inconsequential for now but after this block it is possible for err to be not nil (it could beErrEndpointStateNotFound / ErrConnectionFailure )-- not sure if you wanted to explicitly set err = nil after checking if it's ErrEndpointStateNotFound / ErrConnectionFailure if you confirmed this isn't actually an error.

The reason I'm bringing this up is because err is actually examined by the defer statement on return so if the error you receive is expected, I would set the err to nil immediately afterward.

behzad-mir: The add file section is redundant and I simply refactored the code so that if the error is connection error or no state in CNS we will go to the .Also the Err will be set to nil before retrun nil within the next if block (len(epInfos) == 0) that will take care of asynch delete. 
I will add a err=nil eplicitly for these two errors as well for safeguarding.

network/secondary_endpoint_client_linux.go
behzad-mir: Yes, but I still wanted to do the right thing here and not rely on containerd.
CNI has brought this NIC to the netnsduring ADD and should take it out during delete

network/manager.go
behzad-mir: DeleteState is not special to Asynch delete scenario. It is designated to the SiftV2 windows since the delete is not Atomic and we have to delete the state by CNI at the end.
However in Linux scenario we do not need to call this function since CNS will always reply by state not found or an error.
As a result, we have to find a logic to prevent calling CNS in Linux swiftv2. I could have used asynch delete condition since it is right now enabled for Linux Swiftv2 and disabled for Windows. However this is not a n accurate condition. A correctc condition is checking platfrom to make sure it is only windows.

network/secondary_endpoint_client_linux.go
QxBytes: behzad to modify the comment

network/manager.go
QxBytes: behzad to mention in linux the state is cleaned up earlier when we call the cns invoker, and that the platform == windows check is for optimization-- if it had no check we would call this in swiftv2 stateless linux scenarios but it would be unecessary/redundant and incur a performance cost of 1-2 seconds

at this point the cns should be up (if the cns is down at the beginning of the call to Delete we would return no matter what at the `len(epInfos) == 0` block in `Delete`  of cni/network (whether we have async delete enabled or not) and would not need to continue to DeleteState)

cni/network/network.go
QxBytes: we don't have disable async delete check because that is covered by the len(epInfos == 0) when it calls into cns invoker and cns invoker checks the disable async delete flag

network/manager.go
```diff
@@ -768,12 +775,13 @@ func (nm *networkManager) DeleteState(epInfos []*EndpointInfo) error {
 
 	logger.Info("Deleting state")
 
-	// For AKS stateless cni, plugin.ipamInvoker.Delete takes care of removing the state in the main Delete function.
-	// For swiftv2 stateless cni, this call will delete the endpoint state from CNS.
+	// For AKS stateless cni Windows, plugin.ipamInvoker.Delete takes care of removing the state in the main Delete function.
+	// For swiftv2 stateless cni in Windows, this call will delete the endpoint state from CNS.
+	// For SwiftV2 stateless cni in Linux, plugin.ipamInvoker.Delete already deletes the endpoint state from CNS.
```
tamilmani1989, line 777: why there is difference between Linux and Windows

network/manager.go
```diff
@@ -768,12 +775,13 @@ func (nm *networkManager) DeleteState(epInfos []*EndpointInfo) error {
...
 	if nm.IsStatelessCNIMode() {
 		for _, epInfo := range epInfos {
-			// this cleanup happens only for standalone swiftv2 to delete endpoint state from CNS.
-			if epInfo.NICType == cns.NodeNetworkInterfaceFrontendNIC || epInfo.NICType == cns.NodeNetworkInterfaceAccelnetFrontendNIC {
+			// this cleanup happens only for standalone swiftv2 Windows to delete endpoint state from CNS.
+			if (epInfo.NICType == cns.NodeNetworkInterfaceFrontendNIC || epInfo.NICType == cns.NodeNetworkInterfaceAccelnetFrontendNIC) && runtime.GOOS == "windows" {
```
tamilmani1989: why special case for windows?

network/manager.go
behzad-mir: It has been addressed in the new commit

network/manager.go
```diff
@@ -766,13 +772,22 @@ func (nm *networkManager) DeleteState(epInfos []*EndpointInfo) error {
...
 	defer nm.Unlock()
 
-	logger.Info("Deleting state")
+	// For AKS stateless cni Windows, plugin.ipamInvoker.Delete takes care of removing the state in the main Delete function.
+	// For swiftv2 stateless cni in Windows, this call will delete the endpoint state from CNS.
+	// For SwiftV2 stateless cni in Linux, plugin.ipamInvoker.Delete already deletes the endpoint state from CNS.
+	if nm.IsStatelessCNIMode() {
+		for _, epInfo := range epInfos {
+			if epInfo.NICType == cns.InfraNIC {
+				// If the scanrio contains InfraNic delete it means the state has already been deleted in CNS by plugin.ipamInvoker.Delete
```
QxBytes: so this won't affect cases where we have an infra nic + delegated nic because they will supposedly be part of the same endpoint in azure-endpoints.json (stateless cni statefile located in cns), and deleting the infra will also delete all of the secondaries, right?

also can we move the 779-784 for loop inside of the if block on line 788? The if blocks are right after each other and seem to check the same conditional (keep the comment around 775).

can we also fix the comment on 781-- "scenario" and "If the scenario contains InfraNic it means the state has already been deleted..." (the "InfraNic delete" in the comment is confusing me a bit).

QxBytes: so this won't affect cases where we have an infra nic + delegated nic because they will supposedly be part of the same endpoint in azure-endpoints.json (stateless cni statefile located in cns), and deleting the infra will also delete all of the secondaries, right?

also can we move the 779-784 for loop inside of the if block on line 788? The if blocks are right after each other and seem to check the same conditional (keep the comment around 775).

can we also fix the comment on 781-- "scenario" and "If the scenario contains InfraNic it means the state has already been deleted..." (the "InfraNic delete" in the comment is confusing me a bit).

behzad-mir: done

tamilmani1989 (approved):

PR #4213, opened 2026-02-03
[MultiNIC] Propagate SkipDefaultRoutes to CNI and fix windows endpoint default route behavior

adjordjevic-microsoft:
**Reason for Change**:

1. _Change for SkipDefaultRoutes in ipam.go_:
_SkipDefaultRoutes_ flag was not fully propagated through the stack. While CNS received the value, it was not forwarded to CNI, and the decision was made at the CNI layer. This change updates _requestIPConfigHandlerHelperStandalone_ to propagate _SkipDefaultRoutes_ from the NC request to CNI, making its behavior consistent with other parameters.

2. _Change for adding dummy route in endpoint_windows.go_:
In multi-NIC scenarios on WCOW, CNI may end up adding multiple default routes. 
For example - one NIC has _SkipDefaultRoutes = false_ and correctly adds a default route (0.0.0.0/0) via its gateway. Another NIC has _SkipDefaultRoutes = true_, but Windows still implicitly add an additional default route for it. This results in two default routes, which can negatively impact routing. As a workaround, we explicitly add a dummy default route (0.0.0.0/0) with gateway 0.0.0.0 for the NIC where _SkipDefaultRoutes = true_, which ensures correct traffic behavior.

### DIFF
--- a/network/endpoint_windows.go
+++ b/network/endpoint_windows.go
@@ -346,7 +346,7 @@ func (nw *network) configureHcnEndpoint(epInfo *EndpointInfo) (*hcn.HostComputeE
 		hcnEndpoint.Routes = append(hcnEndpoint.Routes, hcnRoute)
 	}
 
-	if epInfo.SkipDefaultRoutes && len(hcnEndpoint.Routes) == 0 {
+	if epInfo.SkipDefaultRoutes {
 		logger.Info("Adding dummy default route for SkipDefaultRoutes=true",
 			zap.String("endpoint", infraEpName),
 			zap.String("nicType", string(epInfo.NICType)))


network/endpoint_windows.go
```diff
@@ -346,6 +346,17 @@ func (nw *network) configureHcnEndpoint(epInfo *EndpointInfo) (*hcn.HostComputeE
 		hcnEndpoint.Routes = append(hcnEndpoint.Routes, hcnRoute)
 	}
 
+	if epInfo.SkipDefaultRoutes && len(hcnEndpoint.Routes) == 0 {
```
behzad-mir: Please add UT coverage for this scenario

behzad-mir: Also I suggest instead of checking len(hcnEndpoint.Routes) == 0 you actually add a helper function and check wvi hns API if the route currently exist in datapath or not.

adjordjevic-microsoft: added

adjordjevic-microsoft: I checked this, but it always reports that the dummy route does not exist, so we would still always enter the branch that adds it. It looks the dummy route is not actually programmed into the datapath:

Image

Do you see any other concern with keeping it this way, or should we leave it as is?

behzad-mir: Although you may not see it right now, there might be clusters in prod that hit this issue, so please check the existence of dummyRoute via HNS in Datapath instead of the current condition.

network/endpoint_windows.go
```diff
@@ -346,6 +346,17 @@ func (nw *network) configureHcnEndpoint(epInfo *EndpointInfo) (*hcn.HostComputeE
...
+		logger.Info("Adding dummy default route for SkipDefaultRoutes=true",
+			zap.String("endpoint", infraEpName),
+			zap.String("nicType", string(epInfo.NICType)))
+
+		hcnEndpoint.Routes = append(hcnEndpoint.Routes, hcn.Route{
+			NextHop:           "0.0.0.0",   // Dummy gateway
+			DestinationPrefix: "0.0.0.0/0", // Default route
+		})
+	}
```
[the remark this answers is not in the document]
behzad-mir: can you also please add a comment and explaining why there will not be a windows implicit route?

network/endpoint_windows.go
tamilmani1989: why do we check `len(hcnEndpoint.Routes) == 0`?

adjordjevic-microsoft: Thank you for bringing this up. I originally added the check because, in our case, the endpoint had no routes and we needed to add a dummy route in that scenario. It makes more sense to remove this check.

PR #4216, opened 2026-02-04
[Draft] Ponv6 wireserver fix

NihaNallappagari:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/internalapi_linux.go
+++ b/cns/restserver/internalapi_linux.go
@@ -14,6 +14,7 @@ import (
 	goiptables "github.com/coreos/go-iptables/iptables"
 	"github.com/pkg/errors"
 	vishnetlink "github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 )
 
 const SWIFTPOSTROUTING = "SWIFT-POSTROUTING"
@@ -24,9 +25,6 @@ const WireserverIP = "168.63.129.16"
 // WireserverRulePriority is the priority for the ip rule that routes wireserver traffic via main table.
 // This ensures wireserver traffic goes through eth0 (infra NIC) even when other rules are added.
 const WireserverRulePriority = 0
-
-// MainRouteTable is the main routing table number (equivalent to unix.RT_TABLE_MAIN).
-const MainRouteTable = 254
 
 type IPtablesProvider struct{}
 
@@ -220,7 +218,7 @@ func (n *NetlinkIPRuleClient) RuleAdd(rule *IPRule) error {
 func (service *HTTPRestService) AddNodeIPRule() error {
 	// For delegated NIC scenarios, pod traffic goes through eth1. This rule ensures that traffic to
 	// the wireserver IP goes through the correct interface via the main routing table
-	return service.addIPRule(WireserverIP, MainRouteTable, WireserverRulePriority)
+	return service.addIPRule(WireserverIP, unix.RT_TABLE_MAIN, WireserverRulePriority)
 }
 
 // addIPRule programs an ip rule to route traffic for the given destination IP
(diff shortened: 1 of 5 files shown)


cns/restserver/restserver.go
```diff
@@ -76,12 +76,27 @@ type iptablesGetter interface {
...
 
+// IPRuleClient is an interface for netlink ip rule operations.
+// This is used to program ip rules for routing traffic to specific destinations.
+type IPRuleClient interface {
+	RuleList(family int) ([]IPRule, error)
+	RuleAdd(rule *IPRule) error
+}
+
+// IPRule represents an ip routing rule.
+type IPRule struct {
```
kadevu, line 87: Can you use vishnetlink.NewRule instead of creating a new struct

cns/restserver/internalapi_linux.go
```diff
@@ -177,6 +186,85 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
...
+			Table:    r.Table,
+			Priority: r.Priority,
+		}
+	}
+	return result, nil
+}
+
+// RuleAdd adds an ip rule.
+func (n *NetlinkIPRuleClient) RuleAdd(rule *IPRule) error {
+	nlRule := vishnetlink.NewRule()
```
kadevu, line 211: why do you have to create an IPRule struct and then copy into vishnetlink.NewRule? Can you directly use vishnetlink.NewRule?

NihaNallappagari: vishnetlink is a Linux-only package - it won't compile on Windows. The codebase needs to support both platforms.  Implemented using vishnetlink on linux and no op on windows. If the interface used vishnetlink.Rule directly, the Windows build would fail because it can't import the package. I created these to mock and add tests

cns/service/main.go
```diff
@@ -1315,9 +1315,13 @@ type ipamStateReconciler interface {
 	ReconcileIPAMStateForSwift(ncRequests []*cns.CreateNetworkContainerRequest, podInfoByIP map[string]cns.PodInfo, nnc *v1alpha.NodeNetworkConfig) cnstypes.ResponseCode
 }
 
+type nodeRuleProgrammer interface {
+	AddNodeIPRule() error
```
kadevu, line 1319: Rename it to AddRule

cns/service/main.go
```diff
@@ -1374,6 +1378,14 @@ func reconcileInitialCNSState(ctx context.Context, cli nodeNetworkConfigGetter,
 		return errors.Wrap(err, "failed to reconcile CNS IPAM state")
 	}
 
+	// Program node-level ip rules to ensure critical traffic is routed correctly.
+	if nodeProg != nil {
+		if err := nodeProg.AddNodeIPRule(); err != nil {
```
kadevu, line 1383: I think this is not the right place to add the rule. I think the 2 options are to add it around the programSNATRules or add it around the place in main.go where there are couple of things happening for SwiftV2.

if cnsconfig.EnableSwiftV2 {
            // No-op for linux, mapping is set for windows in aks swiftv2 scenario
            logger.Printf("Fetching backend nics for the node")
            }
 
            // No-op for linux, setting primary macaddress if VF is enabled on the nics for aks swiftv2 windows
            logger.Printf("Setting VF for accelnet nics if feature is enabled (only on windows VM & swiftV2 scenario)")
            if err = httpRemoteRestService.SetVFForAccelnetNICs(); err != nil {
                logger.Errorf("Failed to set VF for accelnet NICs: %v", err)
            }
        }

kadevu: @miguelgoms thoughts?

miguelgoms: I think this is different: cns.EnableSwiftv2 was added for multitenancy first and is enabling the mtpnc reconcilers and the swiftv2middleware which you are not using today. adding this node-level-route on the same if-statement feels mixes the two scenarios and i think it'll be more confusing.

miguelgoms: for that matter - if we're keeping the nodeRuleProgrammer, I'd think the choice of whether this is nil or applicable is made before calling this method.

PR #4217, opened 2026-02-05
fix: remove linux-headers-generic as it was causing pkg not found errors

QxBytes:
**Reason for Change**:

linux-headers-generic recently bumped to -100 (previously -94) and now the package can't be found: https://dev.azure.com/msazure/One/_build/results?buildId=151906478&view=results . This removes those packages so the acn official/nonofficial build pipeline can be unblocked.

Azure iptables monitor and ipv6 build without it being installed:
azure iptables monitor: https://dev.azure.com/msazure/One/CCE%20Apps/_build/results?buildId=151929965&view=logs&j=d3a3756c-a55d-5820-67ec-8fc39ccdb431&t=94931ea8-4cd5-5621-2a4c-70bae38b9e5a&l=710
ipv6 bpf prog: https://dev.azure.com/msazure/One/CCE%20Apps/_build/results?buildId=151937724&view=logs&j=4e720d8a-b7a6-5390-d2aa-2d74f6cff204&t=b8b50c28-2f32-558b-9982-d0c444ebf20b&l=71

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
ACN Unofficial: https://dev.azure.com/msazure/One/_build/results?buildId=151952637&view=results

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,10 +195,10 @@ ipv6-hp-bpf-binary: bpf-lib
 # Libraries for bpf
 bpf-lib:
 ifeq ($(GOARCH),amd64)
-	sudo apt-get update && sudo apt-get install -y llvm clang linux-libc-dev linux-headers-generic libbpf-dev libc6-dev nftables iproute2 gcc-multilib
+	sudo apt-get update && sudo apt-get install -y llvm clang linux-libc-dev libbpf-dev libc6-dev nftables iproute2 gcc-multilib
 	for dir in /usr/include/x86_64-linux-gnu/*; do sudo ln -sfn "$$dir" /usr/include/$$(basename "$$dir"); done
 else ifeq ($(GOARCH),arm64)
-	sudo apt-get update && sudo apt-get install -y llvm clang linux-libc-dev linux-headers-generic libbpf-dev libc6-dev nftables iproute2 gcc-aarch64-linux-gnu
+	sudo apt-get update && sudo apt-get install -y llvm clang linux-libc-dev libbpf-dev libc6-dev nftables iproute2 gcc-aarch64-linux-gnu
 	for dir in /usr/include/aarch64-linux-gnu/*; do sudo ln -sfn "$$dir" /usr/include/$$(basename "$$dir"); done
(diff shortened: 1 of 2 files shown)


.pipelines/build/scripts/ipv6-hp-bpf.sh
```diff
@@ -46,7 +46,7 @@ mkdir -p "$OUT_DIR"/lib
 # Package up Needed C Files
 if [[ -f /etc/debian_version ]];then
   apt-get update -y
-  apt-get install -y --no-install-recommends llvm clang linux-libc-dev linux-headers-generic libbpf-dev libc6-dev nftables iproute2
+  apt-get install -y --no-install-recommends llvm clang linux-libc-dev libbpf-dev libc6-dev nftables iproute2
```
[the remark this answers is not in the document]
QxBytes: binary.steps.yaml is called by binaries.jobs.yaml which doesn't seem to be used

.pipelines/build/scripts/azure-iptables-monitor.sh
```diff
@@ -24,7 +24,7 @@ echo "Building azure-block-iptables binary..."
 if [[ -f /etc/debian_version ]]; then
 
   apt-get update -y
-  apt-get install -y --no-install-recommends llvm clang linux-libc-dev linux-headers-generic libbpf-dev libc6-dev nftables iproute2
+  apt-get install -y --no-install-recommends llvm clang linux-libc-dev libbpf-dev libc6-dev nftables iproute2
```
jpayne3506, line 27: If I had to guess, without spending extra time. It was potentially never needed as the header files you require would be in `linux-libc-dev libbpf-dev libc6-dev`. If you have extra cycles you could confirm through using dive and splitting up the respective apt install commands in the matching dockerfile build.

QxBytes: some headers are still added by that layer but the pipeline seems to build and run the images just fine

PR #4221, opened 2026-02-10
backport: fix: init NNC from reconciler instead of directly

### LINKED ISSUE #4210: Replacing a Node with the same name in Overlay can corrupt CNS state

rbtr:
backport of https://github.com/Azure/azure-container-networking/pull/4199 to 1.6

>There's no clear reason that CNS initializes with an NNC that it gets out of band from the typical NNC reconciler flow.
This separate path for fetching the NNC leads to bugs such as ingesting an NNC for a previous Node of the same name and state corruption once the reconciler (which is smart enough to wait for the actual NNC for its Node) starts.
Additionally, any caching, filtering, or other selection done to NNCs for efficiency is not applied or must be duplicated.
>
>Instead, initialization can be run from the first NNC that the reconciler is pushed. This consolidates all CNS NNC reads to a single location where list-watch optimizations and filtering can be centralized.


cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -27,8 +27,12 @@ type cnsClient interface {
 	MustEnsureNoStaleNCs(validNCIDs []string)
 }
 
+type nodenetworkconfigSink func(*v1alpha.NodeNetworkConfig) error
+
+// Deprecated: use nodenetworkconfigSink instead
 type nodeNetworkConfigListener interface {
-	Update(*v1alpha.NodeNetworkConfig) error
+	// Deprecated: use nodenetworkconfigSink instead
+	Update(*v1alpha.NodeNetworkConfig) error // phasing this out in favor of the sink
```
vipul-21, line 35: nit: The deprecated comment can updated to point the sink

cns/kubecontroller/nodenetworkconfig/reconciler.go
```diff
@@ -87,6 +93,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 	r.cnscli.MustEnsureNoStaleNCs(validNCIDs)
 
+	// call initFunc on first reconcile and never again
+	if r.initializer != nil {
+		if err := r.initializer(nnc); err != nil {
+			logger.Errorf("[cns-rc] initializer failed during reconcile: %v", err)
+			return reconcile.Result{}, errors.Wrap(err, "initializer failed during reconcile")
+		}
+		r.initializer = nil
+	}
```
vipul-21, line 102: `CreateOrUpdateNetworkContainerInternal` is called twice per NC on the first reconcile - once inside the initializer's` ReconcileIPAMStateForSwift`, and once in the reconciler's NC loop that runs immediately after. So on the first reconcile now the work done is twice ?

cns/service/main.go
```diff
@@ -1452,36 +1430,13 @@ func InitializeCRDState(ctx context.Context, z *zap.Logger, httpRestService cns.
...
-	attempt := 0
-	err = retry.Do(func() error {
-		attempt++
-		logger.Printf("reconciling initial CNS state attempt: %d", attempt)
-		err = reconcileInitialCNSState(ctx, directscopedcli, httpRestServiceImplementation, podInfoByIPProvider)
-		if err != nil {
-			logger.Errorf("failed to reconcile initial CNS state, attempt: %d err: %v", attempt, err)
+	initializerWrapper := func(nnc *v1alpha.NodeNetworkConfig) error {
+		logger.Printf("Reconciling initial CNS state")
+		if err := reconcileInitialCNSState(nnc, httpRestServiceImplementation, podInfoByIPProvider); err != nil {
```
vipul-21: Under persistent initialization failures, controller-runtime's exponential backoff  degrades more compared to the old fixed 10s retry. Is that a scenario we have encountered where the NNC is failing consistently before succeeding ? (maybe in scale cases?)

cns/kubecontroller/nodenetworkconfig/reconciler.go
rbtr: the work was always done twice - once when we get the NNC directly, and then again when the reconciler starts

cns/service/main.go
rbtr: We shouldn't get an event in to reconcile before the NNC exists, but maybe it is not fully populated yet - those are two separated processes (Node then NNC reconciliation in RC). If RC is especially busy it is possible for there to be a delay. In this case we requeue to try it again.

cns/kubecontroller/nodenetworkconfig/reconciler.go
vipul-21: Oh okay, understood.

cns/service/main.go
vipul-21: Does that possible delay(which needs to be measured in the first place) want us to cap the max delay so we don't overshoot the RC delay window ?
Since this issue impacted OAI cluster, wondering if it is worth checking their cluster size so that on reconciliation they don't face any issue.

vipul-21 (dismissed):

PR #4224, opened 2026-02-17
[feat] Add wireserver ip rule for PONv6

kadevu:
**Reason for Change**:

In the Prefix on NIC v6 feature with SwiftV2/Cilium CNI, pod traffic is routed through eth1 (delegated NIC). This causes wireserver traffic (168.63.129.16) to also go through eth1, which breaks connectivity to the Azure Wireserver.

**Issue Fixed**:

When VnetBlockDualStackSwiftV2 is enabled in the CNS config, CNS programs an ip rule to route wireserver and Azure DNS traffic through eth0 (infra NIC) via the main routing table (table 254).

The rule programmed is equivalent to:

ip rule add to 168.63.129.16/32 table 254 priority 0
ip rule add to 169.254.169.254 table 254 priority 0

**Requirements**:

Results:

image

image

image

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/nodesetup/nodesetup_linux.go
+++ b/cns/nodesetup/nodesetup_linux.go
@@ -27,6 +27,11 @@ var (
 
 // Run performs one-time node-level setup.
 func Run(z *zap.Logger) error {
+	return programWireserverIPRules(z)
+}
+
+// programWireserverIPRules programs ip rules for infrastructure traffic.
+func programWireserverIPRules(z *zap.Logger) error {
 	// For scenarios like Prefix on NIC v6 with Cilium CNI based on SwiftV2, pod traffic may be routed
 	// through eth1 (delegated NIC). These rules ensure critical traffic (e.g. wireserver,
 	// IMDS) is routed through eth0 (infra NIC) via the main routing table.


common/config.go
```diff
@@ -89,6 +89,9 @@ const (
 	// Enable CNS to program SNAT iptables rules
 	OptProgramSNATIPTables = "program-snat-iptables"
 
+	// Enable CNS to program wireserver ip rule for VnetBlockDualStackSwiftV2 scenario
+	OptVnetBlockDualStackSwiftV2 = "vnet-block-dualstack-swiftv2"
```
miguelgoms: if we are checking for this within the rulesclient impl, the config option should be named based on the behavior/rules it enables. there may be more scenarios that need the ip rule going forward, and we should not need to update cns build with new flags per scenario where this would be applicable. the context for mapping the cluster configuration to this requirement is not part of azure-cns itself.

miguelgoms: at this point I'd also keep this simpler, define a higher-level flag for the feature - and check internalapi L#668 before calling .AddRules. skips this code earlier where not applicable, and keeps the conditions for now outside platform-specific impl. if/when we have more specific rules 

e.g.: this could be 'program-nodeiprules' - then its the responsibility of whoever is configuring CNS to set this to true/false if the feature is appropriate.

miguelgoms (dismissed):

cns/restserver/internalapi.go
```diff
@@ -665,6 +665,12 @@ func (service *HTTPRestService) CreateOrUpdateNetworkContainerInternal(req *cns.
 		}
 	}
 
+	if err := service.AddRules(); err != nil {
+		//nolint:staticcheck // SA1019: suppress deprecated logger.Errorf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
+		logger.Errorf("[Azure CNS] Failed to program ip rules: %v", err)
+		return types.UnexpectedError
+	}
```
tamilmani1989: why its part of createorupdatenetworkcontainer?

cns/restserver/internalapi_linux.go
```diff
@@ -177,6 +189,106 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
...
+func (service *HTTPRestService) AddRules() error {
+	if service.ipruleclient == nil {
+		//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
+		logger.Printf("[Azure CNS] IPRuleClient not configured, skipping ip rule programming")
+		return nil
+	}
+
+	var rules []IPRule
+
+	if service.Options[common.OptVnetBlockDualStackSwiftV2] == true {
```
tamilmani1989: why we need separate flag? can we set this for all swiftv2 case?

cns/restserver/internalapi_linux.go
```diff
@@ -9,14 +9,26 @@ import (
...
 	"github.com/pkg/errors"
+	vishnetlink "github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 )
 
 const SWIFTPOSTROUTING = "SWIFT-POSTROUTING"
 
+const (
+	// WireserverIP is the IP address of the Azure Wireserver (also used for DNS).
+	WireserverIP = "168.63.129.16"
```
rbtr: https://github.com/Azure/azure-container-networking/blob/92c604681fea0b14139b7f571b718a29ebe4cf94/cns/configuration/configuration.go#L229

cns/restserver/internalapi_windows.go
```diff
@@ -75,3 +86,8 @@ func (service *HTTPRestService) getPrimaryNICMACAddress() (string, error) {
 	}
 	return macAddress, nil
 }
+
+// AddRules is a no-op on Windows.
+func (service *HTTPRestService) AddRules() error {
+	return nil
+}
```
rbtr: noop implementations to make the compiler happy indicate that the OS abstraction is wrong. Once you get to the point that you need to execute this behavior (which is only valid on one OS), you are _already_ in OS-specific code, and _that's_ what needs to be separated in `_<os>.go` files.

rbtr: I've discussed this recently with Niha and there's more detail and context here https://github.com/Azure/azure-container-networking/pull/4060#discussion_r2670452829

cns/restserver/restserver.go
```diff
@@ -76,12 +76,27 @@ type iptablesGetter interface {
 	GetIPTablesLegacy() (iptablesLegacyClient, error)
 }
 
+// IPRuleClient is an interface for netlink ip rule operations.
+// This is used to program ip rules for routing traffic to specific destinations.
+type IPRuleClient interface {
+	RuleList(family int) ([]IPRule, error)
+	RuleAdd(rule *IPRule) error
+}
```
rbtr: this doesn't seem like behavior that the restserver needs to know about or be involved in

common/config.go
rbtr: > e.g.: this could be 'program-nodeiprules' - then its the responsibility of whoever is configuring CNS to set this to true/false if the feature is appropriate.

100% - directly configure the desired behavior, **not** the scenario

cns/restserver/internalapi_linux.go
```diff
@@ -177,6 +189,106 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 	return types.Success, ""
 }
 
+// NetlinkIPRuleClient implements IPRuleClient using the vishvananda/netlink package.
+type NetlinkIPRuleClient struct{}
+
+// RuleList lists all ip rules for the given address family.
+func (n *NetlinkIPRuleClient) RuleList(family int) ([]IPRule, error) {
```
rbtr: I wish there wasn't precedent for iptables interaction in this file 🙁

rbtr (changes requested):

cns/restserver/internalapi_windows.go
kadevu: Makes sense. Looking at the code the only option I see which is part of the CNS init is programSNATRules. But that does not seem to be a function where we could also "Add IP Rules". Thoughts @rbtr ? Any other recommendations you have where AddRules ideally should be called from within linux specific code.

cns/restserver/internalapi.go
kadevu: I think the 2 options are to add it around the programSNATRules or add it around the place in main.go where there are couple of things happening for SwiftV2.

if cnsconfig.EnableSwiftV2 {
// No-op for linux, mapping is set for windows in aks swiftv2 scenario
logger.Printf("Fetching backend nics for the node")
}

Discussed with Miguel and he mentioned

"I think this is different: cns.EnableSwiftv2 was added for multitenancy first and is enabling the mtpnc reconcilers and the swiftv2middleware which you are not using today. adding this node-level-route on the same if-statement feels mixes the two scenarios and i think it'll be more confusing."

Do you have any other suggestions?

cns/restserver/internalapi_linux.go
kadevu: I will create a behavior specific flag instead as suggested in other comment

cns/restserver/restserver.go
kadevu: I will move this into iprule_linux.go

cns/restserver/internalapi_linux.go
kadevu: Should we have this in a helper class instead?

cns/restserver/internalapi.go
miguelgoms: it could also be in main.go but **not** mixed with the check for cnsconfig.EnableSwiftv2

kadevu: I will make this change if @tamilmani1989 and @rbtr also agree

tamilmani1989: ya lets add in main.go in init/setup call and call based on same config you are planning to add `program-nodeiprules`

common/config.go
tamilmani1989: let field name be  RouteWireserverViaDefaultInterface and be part of cns config map ?

cns/restserver/internalapi_linux.go
```diff
@@ -177,6 +189,106 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
...
+	nlRule.Priority = rule.Priority
+	return errors.Wrap(vishnetlink.RuleAdd(nlRule), "failed to add ip rule")
+}
+
+// wireserverIPRules returns ip rules to route wireserver traffic through the main routing table.
+// For VnetBlockDualStackSwiftV2 (Prefix on NIC v6) with Cilium CNI, pod traffic is routed
+// through eth1 (delegated NIC). These rules ensure critical traffic (e.g. wireserver)
+// is routed through eth0 (infra NIC) via the main routing table.
+func wireserverIPRules() ([]IPRule, error) {
+	_, wireserverNet, err := net.ParseCIDR(WireserverIP + "/32")
```
tamilmani1989: did we test for metadata ip 169.254.169.254 via delegated nic?

cns/restserver/internalapi_linux.go
rbtr: yeah i would prefer this be encapsulated and ideally invisible to the core code

cns/restserver/internalapi_linux.go
```diff
@@ -177,6 +189,106 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
...
+	}
+	return []IPRule{
+		{Dst: wireserverNet, Table: unix.RT_TABLE_MAIN, Priority: WireserverRulePriority},
+	}, nil
+}
+
+// AddRules programs ip rules based on the current service configuration.
+// It checks which options are enabled and adds corresponding rules.
+// It is idempotent: rules that already exist are skipped.
+func (service *HTTPRestService) AddRules() error {
```
timraymond: Why is this all hanging off `*HTTPRestService`? It doesn't have anything to do with HTTP

cns/restserver/internalapi_linux.go
```diff
@@ -177,6 +189,106 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
 	return types.Success, ""
 }
 
+// NetlinkIPRuleClient implements IPRuleClient using the vishvananda/netlink package.
+type NetlinkIPRuleClient struct{}
```
timraymond: Having some wrapper around `vishvanda/netlink` is fine, but I think this is the wrong level of abstraction to be presenting to the rest of CNS. Why don't we have some kind of a `NodePreparer` whose purpose is to prepare the node for hosting the CNS API? This can be a separate phase of initialization.

It's completely conceivable that windows would have its own weird things (idk, HNS stuff?) that have to happen on boot. Trying to jam that into the no-op Windows implementation of this would feel awkward, but if there was something that operated on a higher level (e.g. "all I do is get the node ready, how that happens is an implementation detail"), then it's not so bizarre.

cns/restserver/internalapi_linux.go
```diff
@@ -177,6 +189,106 @@ func (service *HTTPRestService) programSNATRules(req *cns.CreateNetworkContainer
...
+			return nil
+		}
+	}
+
+	if err := service.ipruleclient.RuleAdd(rule); err != nil {
+		return errors.Wrapf(err, "failed to add ip rule to %s table %d priority %d", rule.Dst, rule.Table, rule.Priority)
+	}
+
+	//nolint:staticcheck // SA1019: suppress deprecated logger.Printf usage. Todo: legacy logger usage is consistent in cns repo. Migrates when all logger usage is migrated
+	logger.Printf("[Azure CNS] Added ip rule: to %s table %d priority %d", rule.Dst, rule.Table, rule.Priority)
```
timraymond: We constantly kick this can down the road and these nolints become comment pollution. We should either use the correct logger or remove the deprecation.

common/config.go
kadevu: Renamed it to RouteWireserverViaDefaultInterface

cns/restserver/internalapi.go
kadevu: Moved it into main.go

cns/restserver/internalapi_linux.go
kadevu: done

cns/restserver/internalapi_windows.go
kadevu: Created a new file iprule_linux and moved the ListRule, AddRule into that package

cns/restserver/internalapi_linux.go
kadevu: Moved in into iprule package

common/config.go
kadevu: done

cns/restserver/internalapi_linux.go
kadevu: I still need to test this

cns/restserver/internalapi_linux.go
kadevu: Refactored the code and this dependency is removed

cns/restserver/internalapi_linux.go
kadevu: This makes sense. Created a nodeSetup package and moved the iprule logic in there.

cns/restserver/internalapi_linux.go
kadevu: Used the v2 logger

PR #4228, opened 2026-02-18
feat: Add comprehensive unit tests for stateless CNI endpoint management

behzad-mir:
**Reason for Change**:
Add unit test coverage for stateless CNI operations across CNI plugin (common, Linux, Windows), improving test coverage for the endpoint state management flow used in SwiftV2 and overlay scenarios.

## New: CNSClient Interface for Testability

Added `CNSClient` interface to enable proper unit testing of real `networkManager` methods without requiring HTTP connections:

**File**: `network/cns_client_interface.go` (NEW)
- Defines interface with `GetEndpoint`, `UpdateEndpoint`, `DeleteEndpointState` methods
- `MockCNSEndpointClient` implements this interface
- `networkManager.CnsClient` changed from concrete `*cnsclient.Client` to `CNSClient` interface

## CNI Common Tests

**File**: `cni/network/stateless_cni_test.go`

| Test Function | Description |
|---------------|-------------|
| `TestStatelessCNI_Delete_CNSGetEndpointError` | Error handling when CNS endpoint lookup fails |
| `TestStatelessCNI_Delete_HappyPath` | InfraNIC, FrontendNIC, BackendNIC deletion scenarios |
| `TestStatelessCNI_Delete_MultiNIC` | Multi-NIC combinations (SwiftV2) with CNS state validation |
| `TestStatelessCNI_Delete_DualStack` | IPv4+IPv6 IP release |
| `TestStatelessCNI_Delete_IpamDeleteFails` | IPAM error handling during delete |

## CNI Linux Tests

**File**: `network/stateless_cni_linux_test.go`

| Test Function | Description |
|---------------|-------------|
| `TestStatelessCNI_Delete_Linux_TransparentMode` | Delete endpoint in transparent mode (CRI handles veth cleanup) |
| `TestStatelessCNI_Delete_Linux_FrontendNIC` | Delete FrontendNIC using SecondaryEndpointClient |
| `TestStatelessCNI_Delete_Linux_BackendNIC` | BackendNIC deletion intentionally skipped |
| `TestStatelessCNI_Delete_Linux_NetlinkErrorIgnored` | Netlink errors swallowed during cleanup |
| `TestNetworkManager_GetEndpointState` | **REAL** networkManager.GetEndpointState with mock CNS injected |
| `TestNetworkManager_DeleteState` | **REAL** networkManager.DeleteState conditional logic |
| `TestNetworkManager_UpdateEndpointState` | **REAL** networkManager.SaveState → UpdateEndpoint flow |

## CNI Windows Tests

**File**: `network/stateless_cni_windows_test.go`

| Test Function | Description |
|---------------|-------------|
| `TestStatelessCNI_Delete_Windows_WithHNSEndpointID` | Delete endpoint using HNS endpoint ID |
| `TestStatelessCNI_Delete_Windows_SwiftV2_MultipleNICs` | SwiftV2 multi-NIC deletion on Windows |
| `TestStatelessCNI_Delete_Windows_HNSNotFound` | DELETE idempotent when HNS endpoint missing |
| `TestStatelessCNI_Delete_Windows_BackendNIC` | BackendNIC deletion intentionally skipped |

## Verification

```bash
# Run CNI common tests
go test -v -run 'TestStatelessCNI' ./cni/network/

# Run CNI Linux tests (includes real networkManager tests)
go test -v -run 'TestStatelessCNI|TestNetworkManager' ./network/

### DIFF
--- a/cni/network/stateless_cni_test.go
+++ b/cni/network/stateless_cni_test.go
@@ -1,0 +1,389 @@
+package network
+
+import (
+	"fmt"
+	"net"
+	"testing"
+
+	"github.com/Azure/azure-container-networking/cni"
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/restserver"
+	"github.com/Azure/azure-container-networking/common"
+	acnnetwork "github.com/Azure/azure-container-networking/network"
+	"github.com/Azure/azure-container-networking/nns"
+	cniSkel "github.com/containernetworking/cni/pkg/skel"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	testNamespace = "test-ns"
+)
+
+// GetStatelessTestResources creates a test NetPlugin configured for stateless CNI mode.
+// It uses MockCNSEndpointClient so tests can set up endpoint state in CNS format
+// and verify the conversion to CNI EndpointInfo format.
+//
+// For simpler tests that don't need CNS format testing, you can instead:
+// 1. Call mockNM.SetStatelessCNIMode() without MockCNSClient
+// 2. Use mockNM.SetEndpointState(containerID, []*EndpointInfo{...}) directly
+func GetStatelessTestResources(t *testing.T) (*NetPlugin, *acnnetwork.MockNetworkManager) {
+	pluginName := "testplugin"
+	isIPv6 := false
+	config := &common.PluginConfig{}
+	grpcClient := &nns.MockGrpcClient{}
+	plugin, err := NewPlugin(pluginName, config, grpcClient, &Multitenancy{})
+	require.NoError(t, err, "NewPlugin should not fail")
+
+	// Create mock network manager with stateless mode enabled
+	mockNetworkManager := acnnetwork.NewMockNetworkmanager(acnnetwork.NewMockEndpointClient(nil))
+	err = mockNetworkManager.SetStatelessCNIMode()
+	require.NoError(t, err, "SetStatelessCNIMode should not fail")
+
+	// Initialize MockCNSClient for tests that need CNS format testing or call tracking
+	mockNetworkManager.MockCNSClient = acnnetwork.NewMockCNSEndpointClient()
+
+	plugin.nm = mockNetworkManager
+	plugin.ipamInvoker = NewMockIpamInvoker(isIPv6, false, false, false, false)
+
+	return plugin, mockNetworkManager
+}
+
+// createStatelessTestConfig creates a network config for stateless CNI tests
+func createStatelessTestConfig() cni.NetworkConfig {
+	return cni.NetworkConfig{
+		Name:              "test-stateless-nwcfg",
+		CNIVersion:        "0.3.0",
+		Type:              "azure-vnet",
+		Mode:              OpModeTransparent,
+		Master:            "eth0",
+		IPsToRouteViaHost: []string{"169.254.20.10"},
+		IPAM: cni.IPAM{
+			Type: "azure-cns",
+		},
+	}
+}
+
+// createTestCmdArgs creates CmdArgs for stateless CNI tests, reducing duplication across test cases
+func createTestCmdArgs(nwCfg cni.NetworkConfig, containerID, podName, podNamespace string) *cniSkel.CmdArgs {
+	return &cniSkel.CmdArgs{
+		StdinData:   nwCfg.Serialize(),
+		ContainerID: containerID,
+		Netns:       containerID,
+		Args:        fmt.Sprintf("K8S_POD_NAME=%v;K8S_POD_NAMESPACE=%v", podName, podNamespace),
+		IfName:      "eth0",
+	}
+}
+
+// TestStatelessCNI_Delete_CNSGetEndpointError tests DELETE when CNS returns errors
+func TestStatelessCNI_Delete_CNSGetEndpointError(t *testing.T) {
+	nwCfgStateless := createStatelessTestConfig()
+
+	tests := []struct {
+		name            string
+		setupError      func(*acnnetwork.MockCNSEndpointClient)
+		wantErr         bool
+		wantErrContains string
+	}{
+		{
+			name: "CNS returns EndpointStateNotFound - should succeed",
+			setupError: func(mockCNS *acnnetwork.MockCNSEndpointClient) {
+				mockCNS.GetEndpointErr = acnnetwork.ErrEndpointStateNotFound
+			},
+			wantErr: false, // EndpointStateNotFound is handled gracefully
+		},
+		{
+			name: "CNS returns ConnectionFailure - should succeed (async IP release)",
+			setupError: func(mockCNS *acnnetwork.MockCNSEndpointClient) {
+				mockCNS.GetEndpointErr = acnnetwork.ErrConnectionFailure
+			},
+			wantErr: false, // Connection failure handled with async release
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create fresh plugin and mocks for each test to avoid cross-test contamination
+			plugin, mockNM := GetStatelessTestResources(t)
+
+			// Set up error condition
+			tt.setupError(mockNM.MockCNSClient)
+
+			cmdArgs := createTestCmdArgs(nwCfgStateless, "error-test-container", "error-pod", testNamespace)
+			err := plugin.Delete(cmdArgs)
+
+			if tt.wantErr {
+				require.Error(t, err)
+				if tt.wantErrContains != "" {
+					require.Contains(t, err.Error(), tt.wantErrContains)
+				}
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+// TestStatelessCNI_Delete_HappyPath tests DELETE when CNS returns valid endpoint state
+func TestStatelessCNI_Delete_HappyPath(t *testing.T) {
+	nwCfgStateless := createStatelessTestConfig()
+	containerID := "happy-path-container"
+	podName := "test-pod"
+	podNamespace := testNamespace
+
+	tests := []struct {
+		name          string
+		setupState    func(*acnnetwork.MockCNSEndpointClient, *MockIpamInvoker)
+		validateAfter func(*testing.T, *MockIpamInvoker)
+		wantErr       bool
+		description   string
+	}{
+		{
+			name: "Delete InfraNIC endpoint - IP released",
+			setupState: func(mockCNS *acnnetwork.MockCNSEndpointClient, mockIpam *MockIpamInvoker) {
+				// Set up InfraNIC endpoint in CNS
+				mockCNS.SetEndpointStateWithIPInfo(containerID, podName, podNamespace, map[string]*restserver.IPInfo{
+					"eth0": acnnetwork.CreateMockIPInfo(cns.InfraNIC, "10.240.0.5/24", "", "", "veth-host", ""),
+				})
+				// Pre-populate ipam invoker with the IP so Delete validates it
+				mockIpam.ipMap["10.240.0.5/24"] = true
+			},
+			validateAfter: func(t *testing.T, mockIpam *MockIpamInvoker) {
+				// Verify IP was released from IPAM
+				_, exists := mockIpam.ipMap["10.240.0.5/24"]
+				require.False(t, exists, "InfraNIC IP should be released from IPAM")
+			},
+			wantErr:     false,
+			description: "InfraNIC endpoint should be deleted and IP released via ipamInvoker.Delete",
+		},
+		{
+			name: "Delete FrontendNIC endpoint - IP NOT released (delegated)",
+			setupState: func(mockCNS *acnnetwork.MockCNSEndpointClient, _ *MockIpamInvoker) {
+				// Set up FrontendNIC (delegated) endpoint in CNS
+				mockCNS.SetEndpointStateWithIPInfo(containerID, podName, podNamespace, map[string]*restserver.IPInfo{
+					"eth1": acnnetwork.CreateMockIPInfo(cns.NodeNetworkInterfaceFrontendNIC, "20.20.20.20/32", "", "", "", "aa:bb:cc:dd:ee:ff"),
+				})
+				// Do NOT add to ipam invoker - delegated IPs should not be released
+			},
+			validateAfter: func(t *testing.T, mockIpam *MockIpamInvoker) {
+				// Verify ipMap is unchanged (delegated NICs don't release IPs via IPAM)
+				require.Empty(t, mockIpam.ipMap, "FrontendNIC should not trigger IPAM release")
+			},
+			wantErr:     false,
+			description: "FrontendNIC (delegated) endpoint should be deleted but IP NOT released",
+		},
+		{
+			name: "Delete BackendNIC endpoint - IP NOT released",
+			setupState: func(mockCNS *acnnetwork.MockCNSEndpointClient, _ *MockIpamInvoker) {
+				// Set up BackendNIC endpoint in CNS
+				mockCNS.SetEndpointStateWithIPInfo(containerID, podName, podNamespace, map[string]*restserver.IPInfo{
+					"ib1": acnnetwork.CreateMockIPInfo(cns.BackendNIC, "", "", "", "", ""),
+				})
+				// BackendNIC has no IP to release
+			},
+			validateAfter: func(t *testing.T, mockIpam *MockIpamInvoker) {
+				// Verify ipMap is unchanged (BackendNIC has no IPs)
+				require.Empty(t, mockIpam.ipMap, "BackendNIC should not trigger IPAM release")
+			},
+			wantErr:     false,
+			description: "BackendNIC endpoint should be deleted (no IP release needed)",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create fresh plugin and mocks for each test
+			plugin, mockNM := GetStatelessTestResources(t)
+			mockIpam := NewMockIpamInvoker(false, false, false, false, false)
+			plugin.ipamInvoker = mockIpam
+
+			// Set up endpoint state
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: network/mock_cns_endpoint_client.go [hunk 1]
```diff
@@ -0,0 +1,221 @@
...
+
+		endpointInfo := &EndpointInfo{
+			ContainerID:   containerID,
+			IfName:        ifname,
+			NICType:       ipInfo.NICType,
+			HNSEndpointID: ipInfo.HnsEndpointID,
+			HNSNetworkID:  ipInfo.HnsNetworkID,
+			HostIfName:    ipInfo.HostVethName,
+			MacAddress:    net.HardwareAddr{},
+		}
```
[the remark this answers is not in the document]
behzad-mir: Addressed

cni/network/stateless_cni_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+			err := plugin.Delete(cmdArgs)
+
+			if tt.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
```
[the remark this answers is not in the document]
behzad-mir: Addressed

network/manager_mock.go
```diff
@@ -208,7 +215,11 @@ func (nm *MockNetworkManager) GetEndpointInfosFromContainerID(containerID string
 	return ret
 }
 
-func (nm *MockNetworkManager) GetEndpointState(_, _, _ string) ([]*EndpointInfo, error) {
+func (nm *MockNetworkManager) GetEndpointState(_, containerID, ifName string) ([]*EndpointInfo, error) {
+	// GetEndpointState is only called in stateless CNI mode
+	if nm.MockCNSClient != nil {
+		return nm.MockCNSClient.GetEndpointState(containerID, ifName)
+	}
 	return []*EndpointInfo{}, nil
```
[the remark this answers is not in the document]
behzad-mir: Addressed

cni/network/stateless_cni_test.go
```diff
@@ -0,0 +1,403 @@
...
+func createStatelessTestConfig() cni.NetworkConfig {
+	return cni.NetworkConfig{
+		Name:              "test-stateless-nwcfg",
+		CNIVersion:        "0.3.0",
+		Type:              "azure-vnet",
+		Mode:              OpModeTransparent,
+		Master:            "eth0",
+		IPsToRouteViaHost: []string{"169.254.20.10"},
+		IPAM: struct {
+			Mode          string `json:"mode,omitempty"`
```
QxBytes: why can't we use cni.IPAM ?

cni/network/stateless_cni_test.go
```diff
@@ -0,0 +1,403 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Reset mock
+			mockNM.MockCNSClient = acnnetwork.NewMockCNSEndpointClient()
+
+			// Set up error condition
+			tt.setupError()
+
+			cmdArgs := &cniSkel.CmdArgs{
+				StdinData:   nwCfgStateless.Serialize(),
```
QxBytes: this CmdArgs step looks very similar between the steps-- can we move this and other duplicated setup to a common func that returns all the test resources we need? Ex: getting the plugin, test config, cmd args, mocks, and pass in a param for what we need to change

cni/network/stateless_cni_test.go
```diff
@@ -0,0 +1,403 @@
...
+		}{
+			Type: "azure-cns",
+		},
+	}
+}
+
+// TestStatelessCNI_Delete_CNSGetEndpointError tests DELETE when CNS returns errors
+func TestStatelessCNI_Delete_CNSGetEndpointError(t *testing.T) {
+	plugin, mockNM := GetStatelessTestResources(t)
+	nwCfgStateless := createStatelessTestConfig()
```
QxBytes, line 79: is it intended that we share this between each test? If you wanted a separate NetPlugin for each test can we put it in the t.Run like the other tests?

cni/network/stateless_cni_test.go
```diff
@@ -0,0 +1,403 @@
...
+			setupState: func(mockCNS *acnnetwork.MockCNSEndpointClient, _ *MockIpamInvoker) {
+				// Set up FrontendNIC (delegated) endpoint in CNS
+				mockCNS.SetEndpointStateWithIPInfo(containerID, podName, podNamespace, map[string]*restserver.IPInfo{
+					"eth1": acnnetwork.CreateMockIPInfo(cns.NodeNetworkInterfaceFrontendNIC, "20.20.20.20/32", "", "", "", "aa:bb:cc:dd:ee:ff"),
+				})
+				// Do NOT add to ipam invoker - delegated IPs should not be released
+			},
+			validateAfter: func(t *testing.T, mockIpam *MockIpamInvoker) {
+				// Verify ipMap is unchanged (delegated NICs don't release IPs via IPAM)
+				require.Empty(t, mockIpam.ipMap, "FrontendNIC should not trigger IPAM release")
```
QxBytes, line 169: how do you verify that the ip 20.20.20.20/32 is not released? Above for infra you checked that it was released by checking that ip map (which I believe is whatever is NOT released/in use) did not have the ip 10.240.0.5/24. By that logic I expect 20.20.20.20/32 to still exist in ip map if it were not released. Is it empty because ipMap never received the delegated ip in the first place? If the previous statement is true, we should add validation that the 20.20.20.20/32 still exists somewhere in memory/state. Please clarify the description.

cni/network/stateless_cni_test.go
```diff
@@ -0,0 +1,403 @@
...
+			setupState: func(mockCNS *acnnetwork.MockCNSEndpointClient, _ *MockIpamInvoker) {
+				// Set up BackendNIC endpoint in CNS
+				mockCNS.SetEndpointStateWithIPInfo(containerID, podName, podNamespace, map[string]*restserver.IPInfo{
+					"ib1": acnnetwork.CreateMockIPInfo(cns.BackendNIC, "", "", "", "", ""),
+				})
+				// BackendNIC has no IP to release
+			},
+			validateAfter: func(t *testing.T, mockIpam *MockIpamInvoker) {
+				// Verify ipMap is unchanged (BackendNIC has no IPs)
+				require.Empty(t, mockIpam.ipMap, "BackendNIC should not trigger IPAM release")
```
QxBytes, line 185: same comment as above-- how do we validate backend nic is gone from the cns state? Can we use the mock GetEndpointState?

network/manager_mock.go
```diff
@@ -72,14 +75,18 @@ func (nm *MockNetworkManager) DeleteEndpoint(_, endpointID string, _ *EndpointIn
 	return nil
 }
 
-// SetStatelessCNIMode enable the statelessCNI falg and inititlizes a CNSClient
+// SetStatelessCNIMode enable the statelessCNI flag and initializes a CNSClient
 func (nm *MockNetworkManager) SetStatelessCNIMode() error {
+	nm.StatelessCNIMode = true
+	if nm.MockCNSClient == nil {
+		nm.MockCNSClient = NewMockCNSEndpointClient()
```
QxBytes: This is confusing me-- the purpose of mocking the network manager is so you can test everything "above" it. So you could choose what each network manager method returns when you specifically test cni/network.go "Add" and "Delete" when in stateless mode. Instead of creating a whole new mock cns client why not just add a function to the mock which we can call in GetEndpointState to get the []*EndpointInfo that you want?

If you want to test how the real networkManager interacts with CNS that should be in your stateless_cni_*_test.go which you seem to be doing

network/mock_cns_endpoint_client.go
```diff
@@ -0,0 +1,187 @@
+// Copyright 2017 Microsoft. All rights reserved.
```
QxBytes, line 1: It seems like this file is used in stateless_cni_test.go in that error test only-- based on this https://github.com/Azure/azure-container-networking/pull/4228/changes#r2835586594 I'm struggling to see where most of this is run as GetEndpointState seems to be the only part hit when I set breakpoints. Because you control MockNetworkManager > GetEndpointState and can control what it returns I believe this mock cns endpoint client file is not necessary.

QxBytes (changes requested):
Also were we going to test network manager GetEndpointState or DeleteState? Maybe I missed the tests for those

I think the main thing that seemed off to me was the `mock_cns_endpoint_client.go`

cni/network/stateless_cni_test.go
```diff
@@ -0,0 +1,403 @@
...
+			Type          string `json:"type"`
+			Environment   string `json:"environment,omitempty"`
+			AddrSpace     string `json:"addressSpace,omitempty"`
+			Subnet        string `json:"subnet,omitempty"`
+			Address       string `json:"ipAddress,omitempty"`
+			QueryInterval string `json:"queryInterval,omitempty"`
+		}{
+			Type: "azure-cns",
+		},
+	}
```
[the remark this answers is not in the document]
behzad-mir: Changed anonymous struct to use [cni.IPAM{Type: "azure-cns"}]

cni/network/stateless_cni_test.go
behzad-mir: Changed anonymous struct to use [cni.IPAM{Type: "azure-cns"}]

cni/network/stateless_cni_test.go
behzad-mir: Added [createTestCmdArgs() helper function to eliminate code duplication.
Moved plugin creation inside [t.Run] in CNSGetEndpointError test to prevent test pollution, and updated all test functions to use the helper.

cni/network/stateless_cni_test.go
behzad-mir: Moved plugin creation inside [t.Run] in CNSGetEndpointError test to prevent test pollution, and updated all test functions to use the helper.

cni/network/stateless_cni_test.go
behzad-mir: Updated [validateAfter] signature to include MockCNSEndpointClient.
Added validation that [GetEndpointStateCalls] contains expected containerID and added [GetEndpointStateCalls] tracking to MockCNSEndpointClient

cni/network/stateless_cni_test.go
behzad-mir: Updated [validateAfter] signature to include MockCNSEndpointClient.
Added validation that [GetEndpointStateCalls] contains expected containerID and added [GetEndpointStateCalls] tracking to MockCNSEndpointClient

network/manager_mock.go
behzad-mir: ✅ Addressed - Added CNSClient interface to enable testing of REAL networkManager methods:

Created [cns_client_interface.go] with CNSClient interface
Changed [networkManager.CnsClient] from concrete *cnsclient.Client to CNSClient interface
[MockCNSEndpointClient] now implements CNSClient interface (compile-time verified).

network/mock_cns_endpoint_client.go
behzad-mir: addressed above

network/manager.go
```diff
@@ -77,7 +77,7 @@ type EndpointClient interface {
 // NetworkManager manages the set of container networking resources.
 type networkManager struct {
 	statelessCniMode   bool
-	CnsClient          *cnsclient.Client
+	CnsClient          CNSClient // Interface for CNS client operations (injectable for testing)
```
QxBytes, line 80: can we change this comment to mention that CnsClient is used only in stateless cni scenarios-- the current comment doesn't seem to provide much information. Also wondering since cns client is now an interface, whether we can add it to the NewNetworkManager function since all the other interfaces are put in there. I know you only use it in stateless, but maybe it would make clearer the dependencies that NetworkManager has.

network/cns_client_interface.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/restserver"
+)
+
+// CNSClient defines the interface for CNS client operations used by networkManager
+// in stateless CNI mode. This interface allows for dependency injection and unit testing.
```
QxBytes, line 14: you don't need to mention it's for dependency injection/uts in the comment, and the comments above each of the methods don't seem to add much about what each is doing-- we probably don't need those comments

network/manager_mock.go
```diff
@@ -72,14 +75,29 @@ func (nm *MockNetworkManager) DeleteEndpoint(_, endpointID string, _ *EndpointIn
...
+	nm.StatelessCNIMode = true
+	// MockCNSClient is now optional - tests can use TestEndpointInfoMap directly
+	// via SetEndpointState() for simpler scenarios
 	return nil
 }
 
+// SetEndpointState is a helper to directly set endpoint state for testing.
+// This is the simpler alternative to using MockCNSEndpointClient.
+// Use this when you don't need CNS format conversion testing or call tracking.
+func (nm *MockNetworkManager) SetEndpointState(containerID string, epInfos []*EndpointInfo) {
```
QxBytes: where is this used?

network/manager_mock.go
```diff
@@ -208,8 +226,14 @@ func (nm *MockNetworkManager) GetEndpointInfosFromContainerID(containerID string
 	return ret
 }
 
-func (nm *MockNetworkManager) GetEndpointState(_, _, _ string) ([]*EndpointInfo, error) {
-	return []*EndpointInfo{}, nil
+func (nm *MockNetworkManager) GetEndpointState(_, containerID, netns string) ([]*EndpointInfo, error) {
+	// GetEndpointState is only called in stateless CNI mode.
```
QxBytes, line 230: why don't you just check nm.StatelessCNIMode or IsStatelessCNIMode() ?

network/stateless_cni_linux_test.go
```diff
@@ -0,0 +1,389 @@
+//go:build linux
+// +build linux
```
QxBytes, line 2: I don't believe this is required anymore-- go should be able to determine it's a linux file (and only build if it's linux) from the naming convention you used

network/stateless_cni_linux_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+	// Verify UpdateEndpoint was called with correct data
+	require.Len(t, mockCNS.UpdateEndpointCalls, 1)
+	require.Equal(t, "test-container", mockCNS.UpdateEndpointCalls[0].EndpointID)
+	require.Contains(t, mockCNS.UpdateEndpointCalls[0].IPInfo, "eth0")
+
+	ipInfo := mockCNS.UpdateEndpointCalls[0].IPInfo["eth0"]
+	require.Equal(t, cns.InfraNIC, ipInfo.NICType)
+	require.Equal(t, "hns-endpoint-123", ipInfo.HnsEndpointID)
+	require.Equal(t, "veth-host", ipInfo.HostVethName)
```
QxBytes, line 388: looks like you are testing network manager here-- shouldn't it go in manager_test.go ? Or create a new file with similar naming conventions? Like manager_stateless_cni_linux_test.go ?

network/stateless_cni_windows_test.go
```diff
@@ -0,0 +1,280 @@
+// Copyright 2017 Microsoft. All rights reserved.
```
QxBytes: Probably don't need this header in all your files (also since it's no longer 2017)

rbtr (approved):

network/stateless_cni_windows_test.go
QxBytes, line 1: probably don't need this header in all your files

cni/network/stateless_cni_test.go
QxBytes, line 1: This seems to test plugin.* methods, which is what network_test.go does-- can we either move these tests there or name this so it gets grouped with the other network files? maybe like network_stateless_test.go ?

cni/network/stateless_cni_test.go
QxBytes: This doesn't seem to answer my question. The only validation I see is the `require.Empty`. If we don't releasevia ipMap, where does the 20.20.0.20/32 end up-- is it still in memory somewhere?

cni/network/stateless_cni_test.go
QxBytes: same response as above

network/mock_cns_endpoint_client.go
```diff
@@ -0,0 +1,256 @@
...
+
+	return &cns.Response{
+		ReturnCode: types.Success,
+		Message:    "success",
+	}, nil
+}
+
+// DeleteEndpointState deletes the endpoint state from the mock store
+func (m *MockCNSEndpointClient) DeleteEndpointState(_ context.Context, endpointID string) (*cns.Response, error) {
+	m.DeleteEndpointStateCalls = append(m.DeleteEndpointStateCalls, endpointID)
```
QxBytes, line 147: We seem to have call tracking for DeleteEndpointState calls-- why don't we seem to use this in `stateless_cni_test.go` ? Rather we seem to use `GetEndpointState` call tracking there. 

Also I'd like to be clear on containerID vs. endpointID-- maybe I didn't mention this before but I believe in most stateless scenarios, the "endpointID" is actually the containerID and should be named as such. Please confirm this because they are different. 

containerID is the same throughout the lifecycle of the container and all endpoint infos from a single ADD call share the same containerID. endpointID during add is something that's generated during ADD and only unique during that one add call.

network/mock_cns_endpoint_client.go
```diff
@@ -0,0 +1,256 @@
...
+
+// SetConnectionError configures the mock to return a ConnectionError with the appropriate ReturnCode.
+// This helper ensures tests match production behavior where GetEndpoint returns types.ConnectionError
+// when CNS connectivity fails.
+func (m *MockCNSEndpointClient) SetConnectionError() {
+	m.GetEndpointErr = ErrConnectionFailure
+	m.GetEndpointReturnCode = types.ConnectionError
+}
+
+// SetEndpointState is a helper to set up mock endpoint state for testing
```
QxBytes, line 196: I don't think this comment is providing useful information-- can you go through all your comments and remove ones for functions/mechanics which don't need explaining?

PR #4229, opened 2026-02-19
dualstack swiftv1 multitenancy: added ipv6 support in CNS

estebancams:
This PR adds a new field IPv6Configuration to CNS state. The goal of this field is to pass ipv6 from DNC for standalone swiftv1 CNI multitenancy scenarios for dual stack feature.
The function of this field is analogous to the existent IPConfiguration for ipv4. When CNI calls CNS during an ADD call (getAllnetworkcontainers interface), CNS will pass this new field on to CNI.
This PR does not implement functionality on CNI side, except for receiving the new field and testing backward compatibility. Other PRs in this feature will implement the rest

### Comment on backward/forward compatibility:
- CNS state model has a new field IPv6Configuration analog to existent IPConfiguration for ipv4
- old DNC + new CNS + old CNI: DNC will not send the new field so CNS will zero it out in state, that means IPv6Configration will be ignored by CNI
- If CNS has to be rolledback, state file might have IPv6Configuration in it but old CNS will ignore the field during deserialization

### DIFF
--- a/cns/restserver/api_test.go
+++ b/cns/restserver/api_test.go
@@ -1991,6 +1991,7 @@ func TestGetHostLocalIP(t *testing.T) {
 // returns IPv6Configuration when present (ServiceFabric/Swiftv1 multitenancy scenario)
 func TestGetAllNetworkContainersWithIPv6Configuration(t *testing.T) {
 	setEnv(t)
+	cleanupAllNetworkContainers()
 	err := setOrchestratorType(t, cns.ServiceFabric)
 	require.NoError(t, err)
 
@@ -2030,6 +2031,7 @@ func TestGetAllNetworkContainersWithIPv6Configuration(t *testing.T) {
 // when IPv6Configuration is not provided (empty) in ServiceFabric scenario
 func TestGetAllNetworkContainersBackwardCompatibilityEmptyIPv6(t *testing.T) {
 	setEnv(t)
+	cleanupAllNetworkContainers()
 	err := setOrchestratorType(t, cns.ServiceFabric)
 	require.NoError(t, err)
 
@@ -2063,6 +2065,7 @@ func TestGetAllNetworkContainersBackwardCompatibilityEmptyIPv6(t *testing.T) {
 // in ServiceFabric/Swiftv1 multitenancy scenario
 func TestPostNetworkContainersWithIPv6(t *testing.T) {
 	setEnv(t)
+	cleanupAllNetworkContainers()
 	err := setOrchestratorType(t, cns.ServiceFabric)
 	require.NoError(t, err)
 
@@ -2102,5 +2105,19 @@ func TestPostNetworkContainersWithIPv6(t *testing.T) {
 	for i := 0; i < len(ncParamsWithIPv6); i++ {
 		err = deleteNetworkContainerWithParams(ncParamsWithIPv6[i])
 		require.NoError(t, err)
+	}
+}
+
+// cleanupAllNetworkContainers removes all NCs from the shared service state
+// so that tests relying on exact NC counts are not affected by stale state
+// left behind by prior test failures.
+func cleanupAllNetworkContainers() {
+	svc.Lock()
+	defer svc.Unlock()
+	for k := range svc.state.ContainerStatus {
+		delete(svc.state.ContainerStatus, k)
+	}
+	for k := range svc.state.ContainerIDByOrchestratorContext {
+		delete(svc.state.ContainerIDByOrchestratorContext, k)
 	}
 }


cni/network/multitenancy.go
```diff
@@ -228,6 +228,16 @@ func (m *Multitenancy) GetAllNetworkContainers(
 
 		ipconfig, routes := convertToIPConfigAndRouteInfo(ifInfo.NCResponse)
 		ifInfo.IPConfigs = append(ifInfo.IPConfigs, ipconfig)
+
+		// Add IPv6Configuration if present (used for CNI multitenancy in Swiftv1 scenario)
+		if ifInfo.NCResponse.IPv6Configuration.IPSubnet.IPAddress != "" {
```
QxBytes: I noticed in convertToIPConfigAndRouteInfo above we seem to have a branch where we check if the ip configuration is an ipv4 address. If it isn't, it looks like we use ipv6FullMask which is very similar to what you are adding here. Why do we need this duplicated logic? Is it because we can have both an ipv4 and ipv6 configuration for a single NC? If so I wonder what was the reasoning for having convertToIPConfigAndRouteInfo since if I recall dualstack was always ipv4 + ipv6, and currently, if 1 pod = 1 NC, then the current code would only support ipv6 (not ipv4).

cni/network/multitenancy_test.go
```diff
@@ -862,3 +862,190 @@ func TestGetMultiTenancyCNIResultNotFound(t *testing.T) {
...
+	for _, info := range ipamResult.interfaceInfo {
+		ifInfo = info
+		break
+	}
+
+	// Verify both IPv4 and IPv6 configurations are present
+	require.Len(t, ifInfo.IPConfigs, 2, "Expected 2 IP configs (IPv4 and IPv6)")
+
+	// Verify IPv6 configuration
+	ipv6Config := ifInfo.IPConfigs[1]
```
QxBytes, line 960: did you want to verify the ipv4 configuration too?

cni/network/multitenancy_test.go
```diff
@@ -862,3 +862,190 @@ func TestGetMultiTenancyCNIResultNotFound(t *testing.T) {
...
+	require.Equal(t, "2001:db8::5", ipv6Config.Address.IP.String())
+	require.Equal(t, 64, getPrefixLength(ipv6Config.Address.Mask))
+	require.Equal(t, "2001:db8::1", ipv6Config.Gateway.String())
+}
+
+// TestGetAllNetworkContainersEmptyIPv6Multitenancy verifies CNI handles
+// empty IPv6Configuration gracefully
```
QxBytes: I believe this test case is unnecessary-- TestGetMultiTenancyCNIResult already covers the ipv4 only case without ipv6

cni/network/multitenancy_test.go
```diff
@@ -862,3 +862,190 @@ func TestGetMultiTenancyCNIResultNotFound(t *testing.T) {
...
+// Helper function to get prefix length from mask
+func getPrefixLength(mask net.IPMask) int {
+	ones, _ := mask.Size()
+	return ones
+}
+
+// MockNetIOShim for testing
+type MockNetIOShim struct {
+	interfaceSubnet net.IPNet
+	callCount       int
```
QxBytes: it doesn't seem like we use this call count?

cni/network/multitenancy_test.go
```diff
@@ -862,3 +862,190 @@ func TestGetMultiTenancyCNIResultNotFound(t *testing.T) {
...
+
+	cnsclient := &MockCNSClient{
+		getAllNetworkContainersConfiguration: getAllNetworkContainersConfigurationHandler{
+			orchestratorContext: orchestratorContext,
+			returnResponse:      ncResponses,
+			err:                 nil,
+		},
+	}
+
+	mockNetIO := &MockNetIOShim{
```
QxBytes: in this file we have `mockNetIOShim` so why can't we reuse it here? I guess if the above comment about this case being redundant holds this struct won't be needed since I think this is the only place `MockNetIOShim` is used (capitalized). Also I changed it to `mockNetIOShim` with nothing passed in and it seems to still pass?

cns/restserver/api_test.go
```diff
@@ -1983,3 +1983,270 @@ func TestGetHostLocalIP(t *testing.T) {
...
+	ipv6Config.IPSubnet = ipv6Subnet
+
+	podInfo := cns.KubernetesPodInfo{PodName: "testpod", PodNamespace: "testpodnamespace"}
+	ctx, err := json.Marshal(podInfo)
+	if err != nil {
+		t.Fatalf("Failed to marshal pod info: %+v", err)
+	}
+
+	ncID := "a47ac10b-58cc-0372-8567-0e02b2c3d479"
+	createReq := cns.CreateNetworkContainerRequest{
```
QxBytes: can we reuse the code for setup between these tests? It seems very similar to what `createOrUpdateNetworkContainerWithParams` is doing already

QxBytes: similar comment with postAllNetworkContainers-- maybe we can add something so it can support ipv6 as well

QxBytes (commented):
could you also address the linter comments?

cni/network/multitenancy.go
estebancams: That is correct. I updated the code by making contvertTo... to handle both ipv4 and 6 cases

cni/network/multitenancy_test.go
estebancams: SInce I changed some common logic during this review, I added ipv4 verification as you suggests.

cni/network/multitenancy_test.go
estebancams: That's, true but I'd like to keep it since it is more targeted to testing the absence of any IPv6 configuration in the resulting objects (while existent test tests that ipv4 configurations are present). I prefer not to disrupt the existent UT and instead make the intention clear in the new one

cni/network/multitenancy_test.go
estebancams: Yes, I replaced it with the existent one

cni/network/multitenancy_test.go
estebancams: Removed

cns/restserver/api_test.go
estebancams: That's a good observation. Helper was extended and used in the new tests

cni/network/multitenancy.go
```diff
@@ -343,10 +352,10 @@ func convertToIPConfigAndRouteInfo(networkConfig *cns.GetNetworkContainerRespons
 
 	for _, ipRouteSubnet := range networkConfig.CnetAddressSpace {
 		routeIPnet := net.IPNet{IP: net.ParseIP(ipRouteSubnet.IPAddress), Mask: net.CIDRMask(int(ipRouteSubnet.PrefixLength), ipv4FullMask)}
-		routes = append(routes, network.RouteInfo{Dst: routeIPnet, Gw: ipconfig.Gateway})
+		routes = append(routes, network.RouteInfo{Dst: routeIPnet, Gw: ipconfigs[0].Gateway})
```
QxBytes: I believe this is fine because we always add the ipv4 configuration first, so ipconfigs[0] will always refer to the ipv4 gateway? If there are both ipv4 and ipv6 configurations, is it possible that they have different Gateway IP Addresses?

cni/network/multitenancy_test.go
QxBytes: In this test you verify:
When inputting a single nc response with a single ipv4 configuration...
1 ipconfig in ifInfo
the ipv4 configuration present matches what we expect

In TestGetMultiTenancyCNIResult:
When inputting two nc responses , each with their own ipv4 configuration...
ifInfo for certain keys exactly matches what we expect
We don't directly check the size of got.interfaceInfo like how we check the # of ifInfo in your test, but I feel like that shouldn't be too hard to add to this test? I think the older test would also benefit because if we have 2 ipv4 nc responses, we want to verify that the output interface infos are exactly 2.

cns/restserver/api_test.go
```diff
@@ -1983,3 +1986,145 @@ func TestGetHostLocalIP(t *testing.T) {
...
+
+	err = createOrUpdateNetworkContainerWithParams(ncParamsWithIPv6[0])
+	if err != nil {
+		t.Fatalf("createOrUpdateNetworkContainerWithParams failed with error:%+v", err)
+	}
+
+	// Retrieve all NCs and verify IPv6Configuration is present
+	ncResponses, err := getAllNetworkContainers(t, ncParamsWithIPv6)
+	if err != nil {
+		t.Fatalf("TestGetAllNetworkContainersWithIPv6Configuration failed with error:%+v", err)
```
QxBytes: nit: the error messages aren't consistent-- sometimes it says the name of the test that failed, and sometimes it says the call within that test that failed

cni/network/multitenancy.go
```diff
@@ -319,18 +319,27 @@ func convertToCniResult(networkConfig *cns.GetNetworkContainerResponse, ifName s
...
+		ipconfig := &network.IPConfig{}
+		ipAddr := net.ParseIP(cnsIPConfig.IPSubnet.IPAddress)
+
+		if ipAddr.To4() != nil {
+			ipconfig.Address = net.IPNet{IP: ipAddr, Mask: net.CIDRMask(int(cnsIPConfig.IPSubnet.PrefixLength), ipv4FullMask)}
+		} else {
+			ipconfig.Address = net.IPNet{IP: ipAddr, Mask: net.CIDRMask(int(cnsIPConfig.IPSubnet.PrefixLength), ipv6FullMask)}
+		}
+
+		ipconfig.Gateway = net.ParseIP(cnsIPConfig.GatewayIPAddress)
```
ZetaoZhuang, line 343: for ipv6, should we use cnsIPConfig.GatewayIPv6Address here?

estebancams: Reason I decided not to: If you see the cnsIP struct:

Image
IPv6 gateway was added to be able to provide a v6 gateway for secondary IPv6s in aks scenarios. This is because secondary IPs struct do not have a gateway field probably because the gateway would be the same for all, secondary and primary ips. Now, in standalone scenarios we don't use secondary so this becomes irrelevant because ipv6 is primary so it has its own gateway field, that is agnostic of v4/v6

cni/network/multitenancy.go
estebancams: Yes, that is risky, updated the code

cni/network/multitenancy_test.go
estebancams: I usually don't like to change other tests unless there is a change of behavior, but in this case it is reasonable. Updated

cns/restserver/api_test.go
estebancams: I decided to replace them with asssertions. I like that pattern better

csfmomo (approved):

PR #4231, opened 2026-02-24
fix: cilium status wait and other pipeline fixes

QxBytes:
**Reason for Change**:

- Fixes various pipeline issues.
- Retriggering a cluster create stage now checks if the cluster already exists, and if it does, tears it down before creating a new cluster. Previously, the cluster create stage would fail if the cluster already existed. This allows us to retrigger a scenario stage like overlay without waiting for all other stages and the "Delete Clusters" stage to finish.
- Shortens names and brings scenario names to the front so that it is obvious in the pipeline overview what scenario each stage is running (Instead of just seeing "Cluster Create..." for each stage).

**Issue Fixed**:

Previously -e was not set in the pipeline script so a failure in cilium wait would not be visible.
EBPF stages were also deployed with hubble enabled even though we don't have hubble present, so disabled hubble for these stages.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
See [here](https://teams.microsoft.com/l/message/19:148c83f5142d4f189176e792c1bf8b47@thread.skype/1772040116448?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=1eb27b57-8ace-4f0e-a9a9-e6db4e9829b7&parentMessageId=1772040116448&teamName=ContainerNetworking&channelName=Code%20review&createdTime=1772040116448) for verification

### DIFF
--- a/hack/aks/deploy.mk
+++ b/hack/aks/deploy.mk
@@ -54,8 +54,7 @@ print-ebpf-cilium-vars:
 	@echo "EBPF_CILIUM_VERSION_TAG: $(EBPF_CILIUM_VERSION_TAG)"
 
 wait-for-cilium:
-	@kubectl rollout status deployment/cilium-operator -n kube-system --timeout=1200s
-	@kubectl rollout status daemonset/cilium -n kube-system --timeout=1200s
+	cilium status --wait --wait-duration 20m
 
 # vanilla cilium deployment
 deploy-cilium-config:
--- a/.pipelines/singletenancy/cilium-dualstack-overlay/cilium-dualstackoverlay-e2e-step-template.yaml
+++ b/.pipelines/singletenancy/cilium-dualstack-overlay/cilium-dualstackoverlay-e2e-step-template.yaml
@@ -4,7 +4,6 @@ parameters:
   scaleup: ""
 
 steps:
-
   - bash: |
       go version
       go env
@@ -19,6 +18,8 @@ steps:
   - task: KubectlInstaller@0
     inputs:
       kubectlVersion: latest
+
+  - template: ../../templates/cilium-cli.yaml
 
   - task: AzureCLI@2
     inputs:
@@ -48,25 +49,12 @@ steps:
     name: "installCilium"
     displayName: "Install Cilium on AKS Dualstack Overlay"
 
-  - template: ../../templates/cilium-cli.yaml
-
   - script: |
       echo "Start Azilium E2E Tests on Overlay Cluster"
       sudo -E env "PATH=$PATH" make test-load SCALE_UP=32 OS_TYPE=linux CNI_TYPE=cilium_dualstack VALIDATE_STATEFILE=true INSTALL_CNS=true INSTALL_OVERLAY=true AZURE_IPAM_VERSION=$(make azure-ipam-version) CNS_VERSION=$(make cns-version) CLEANUP=true
     retryCountOnTaskFailure: 3
     name: "aziliumTest"
     displayName: "Run Azilium E2E on AKS Overlay"
-
-  - script: |
-      set -e
-      kubectl get pods -A
-      echo "Waiting < 2 minutes for cilium to be ready"
-      # Ensure Cilium is ready Xm\Xs
-      cilium status --wait --wait-duration 2m
-      kubectl get crd -A
-    retryCountOnTaskFailure: 3
-    name: "CiliumStatus"
-    displayName: "Cilium Status"
 
   - task: AzureCLI@2
     inputs:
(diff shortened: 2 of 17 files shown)


hack/aks/Makefile
```diff
@@ -462,8 +462,8 @@ linux-swiftv2-nodepool-up: ## Add linux node pool to swiftv2 cluster
 		--pod-subnet-id /subscriptions/$(SUB)/resourceGroups/$(GROUP)/providers/Microsoft.Network/virtualNetworks/$(VNET)/subnets/podnet
 
 down: ## Delete the cluster
-	$(AZCLI) aks delete -g $(GROUP) -n $(CLUSTER) --yes
-	@$(MAKE) unset-kubeconf
+	-$(AZCLI) aks delete -g $(GROUP) -n $(CLUSTER) --yes
+	-@$(MAKE) unset-kubeconf
```
jpayne3506, line 466: does this make these lines quiet..? What is the intent?

.pipelines/templates/create-cluster.yaml
```diff
@@ -41,6 +41,19 @@ jobs:
...
+            
+            if [ ! -z "$CLUSTER_EXISTS" ]; then
+              echo "Cluster ${{ parameters.clusterName }} already exists. Tearing it down first..."
+              make -C ./hack/aks down \
+              AZCLI=az REGION=${{ parameters.region }} SUB=$(SUB_AZURE_NETWORK_AGENT_BUILD_VALIDATIONS) \
+              CLUSTER=${{ parameters.clusterName }}
+              echo "Cluster teardown completed. Proceeding with recreation..."
+            else
+              echo "Cluster ${{ parameters.clusterName }} does not exist. Proceeding with creation..."
+            fi
```
jpayne3506, line 55: Intent?

jpayne3506 (commented):
Maybe we should change our order of installation. Cilium CLI first then Cilium so that the install cilium step can contain `cilium status --wait` Otherwise we should add `--wait to the cli install template. https://github.com/Azure/azure-container-networking/blob/45871288e8f3c54af52d71db55582555e01940fe/.pipelines/templates/cilium-cli.yaml#L24

.pipelines/templates/create-cluster.yaml
jpayne3506: >Retriggering a cluster create stage now checks if the cluster already exists, and if it does, tears it down before creating a new cluster. Previously, the cluster create stage would fail if the cluster already existed. This allows us to retrigger a scenario stage like overlay without waiting for all other stages and the "Delete Clusters" stage to finish.

I don't think this is entirely correct, but my understanding is that this change will no longer fail when cluster already exists. Originally if the cluster existed, you could just retrigger the respective E2E. Either completely or just the failed jobs.

hack/aks/Makefile
QxBytes: makes it so that if the cluster is already deleted, or the kubeconfig does not contain the cluster, the resource group for the cluster still gets deleted (the non-zero exit codes are ignored). This may occur when we retrigger the cluster create step-- the agent does not have any record of the cluster's kubeconfig from the previous run, so unset-kubeconf gives a non-zero exit code.

.pipelines/templates/create-cluster.yaml
QxBytes: Originally, if the cluster existed, yes, that is correct (you could retrigger the failed e2es), but if the cluster was unrecoverable/in a bad state (ex: null characters in statefile), and you wanted to completely recreate the cluster while the test was running by re-running create cluster+dependents, that would not be possible. There were times where it was obvious the test was not going to succeed due to flakiness (but the flakiness corrupted the state etc.), and I just wanted to rerun that stage, but couldn't because Delete Clusters had not yet run.

hack/aks/deploy.mk
```diff
@@ -54,8 +54,7 @@ print-ebpf-cilium-vars:
 	@echo "EBPF_CILIUM_VERSION_TAG: $(EBPF_CILIUM_VERSION_TAG)"
 
 wait-for-cilium:
-	@kubectl rollout status deployment/cilium-operator -n kube-system --timeout=1200s
-	@kubectl rollout status daemonset/cilium -n kube-system --timeout=1200s
+	cilium status --wait --wait-duration 20m
```
camrynl, line 57: why is the wait set to 20m?

camrynl: even before with this being 1200s, not sure why we have the wait set for that long

QxBytes: I previously saw cilium taking slightly less than 20m to come up for cases like ebpf (accounting for pulling the image, for example).

camrynl: hm but for example in .pipelines/templates/cilium-tests.yaml we were only waiting 2m

jpayne3506: I am in favor of a design where status check `--wait`s for a short time and the step retries. That way we get some debug output from each retry.

camrynl: discussed more in chat, can update status check later

PR #4234, opened 2026-02-24
[feat] Trigger Create/update NodeInfo CRD for Swift v1 dualstack podsubnet

NihaNallappagari:
**Reason for Change**:
The PR includes changes that trigger createOrUpdateNodeInfoCRD for swiftv1 scenario based on the flag cns synth passes (true for swiftv1dualstack, false for all other scenarios)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -1429,10 +1429,10 @@ func InitializeCRDState(ctx context.Context, z *zap.Logger, httpRestService cns.
 	}
 
 	// populate the NodeInfo CRD for Swift V1 dualstack scenario when enabled via config
-	if swiftV1Err := enableSwiftV1DualStackCRD(cnsconfig, func() error {
-		return createOrUpdateNodeInfoCRD(ctx, kubeConfig, node)
-	}); swiftV1Err != nil {
-		return swiftV1Err
+	if cnsconfig.EnableSwiftV1DualStack {
+		if nodeInfoErr := createOrUpdateNodeInfoCRD(ctx, kubeConfig, node); nodeInfoErr != nil {
+			return errors.Wrap(nodeInfoErr, "error creating or updating nodeinfo crd for swift v1 dualstack")
+		}
 	}
 
 	// perform state migration from CNI in case CNS is set to manage the endpoint state and has emty state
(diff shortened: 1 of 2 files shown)


csfmomo (approved):

cns/service/main.go
```diff
@@ -1428,6 +1428,13 @@ func InitializeCRDState(ctx context.Context, z *zap.Logger, httpRestService cns.
 		}
 	}
 
+	// populate the NodeInfo CRD for Swift V1 dualstack scenario when enabled via config
+	if swiftV1Err := enableSwiftV1DualStackCRD(cnsconfig, func() error {
```
estebancams: Why not just calling createOrUpdateNoterInfo and handle config logic here?

cns/configuration/configuration.go
```diff
@@ -36,6 +36,7 @@ type CNSConfig struct {
 	EnablePprof                 bool
 	EnableStateMigration        bool
 	EnableSubnetScarcity        bool
+	EnableSwiftV1DualStack      bool
```
estebancams, line 39: Why do we need this flag? where is it being used?

NihaNallappagari: We are using this [flag ](https://github.com/Azure/azure-container-networking/pull/4234/changes/BASE..add2659829b30a5addf9fc70c5e641ddea7ccc4b#diff-3c6ba0ad51f93dc5c172cd00978b989c8c63d2b8d799f8d58d56c983b6498cceR1747) to call createOrUpdateNodeInfoCRD specifically for the SwiftV1 dual-stack pod subnet cluster. Currently, createOrUpdateNodeInfoCRD is invoked only for SwiftV2. The flag received from CNS Synth in AKS-RP indicates that the cluster is the SwiftV1 dual-stack pod subnet and we are skipping the CRD call for other swifv1 scenarios such as overlay.

cns/service/main.go
NihaNallappagari: This change is purely for unit testing purposes. The injectable function parameter allows me to add unit tests. Currently, there are no existing tests for InitializeCRDState or createOrUpdateNodeInfoCRD because they have several external dependencies that require extensive mocking. At this point, I have two options: either invoke the method directly and skip adding tests and depend on runners, or ,move the logic into a separate, testable unit to enable proper test coverage.

estebancams: Ok, in this case logic is so simple that imo does not deserve testing (at the cost of complicating the code). Make your choice, I'm ok

NihaNallappagari: I even wasn’t fully convinced about making code changes solely for the sake of unit tests, but I thought code would be missing unit tests, so added. Since it’s acceptable to skip testing in this case, I’ve removed the test-specific cleanup code I had added.

estebancams: Yes, if coverage is a problem you can use an existing UT (that should exists) that executes that function and just repeat is with the flag on.

PR #4241, opened 2026-02-25
feat: added dual-stack support for transparent-vlan

estebancams:
**Reason for Change**:
This is part of the dualstack support for standalone swift lite feature.
This PR adds dual stack support for CNI transparent-vlan 
So far transparent-vlan only supported ipv4. Changes here do the analog for v6

Change implements the analog that is being done already for ipv4, but for ipv6:
- rules and iptables rules for tunneling
- static roues and neighbour cache
- ipv6 forwarding and ip6 enablement

CNI Delete call is already general enough to handle ipv6 so no changes were required for that flow

Related PR on CNS changes for the same feature: https://github.com/Azure/azure-container-networking/pull/4229

### DIFF
--- a/.pipelines/build/dockerfiles/cns.Dockerfile
+++ b/.pipelines/build/dockerfiles/cns.Dockerfile
@@ -11,11 +11,11 @@ ENTRYPOINT ["azure-cns.exe"]
 EXPOSE 10090
 
 # mcr.microsoft.com/azurelinux/base/core:3.0
-FROM --platform=linux/${ARCH} mcr.microsoft.com/azurelinux/base/core@sha256:2a97a634da103fbfc21697f0136231f17ad18f7880a5be94488c5d2b15eb02d3 AS build-helper
+FROM --platform=linux/${ARCH} mcr.microsoft.com/azurelinux/base/core@sha256:a452d39c91576f5a2c983c7d3b62521fabd08e16b4a7237e24bf2be3b06e1651 AS build-helper
 RUN tdnf install -y iptables
 
 # mcr.microsoft.com/azurelinux/distroless/minimal:3.0
-FROM --platform=linux/${ARCH} mcr.microsoft.com/azurelinux/distroless/minimal@sha256:c9a1ed9515e033bca0c0b3171eccec41790bfcb9e47f05130cddc0a458de46ba AS linux
+FROM --platform=linux/${ARCH} mcr.microsoft.com/azurelinux/distroless/minimal@sha256:22810fd97d6ad5ec7d5bdd5b00233a3050be01d9e26b47b16cb6f1a7f178834b AS linux
 ARG ARTIFACT_DIR .
 
 COPY --from=build-helper /usr/sbin/*tables* /usr/sbin/
--- a/cni/Dockerfile
+++ b/cni/Dockerfile
@@ -9,7 +9,7 @@ ARG OS
 FROM --platform=linux/${ARCH} mcr.microsoft.com/oss/go/microsoft/golang@sha256:bc7423b52b62e8f0281b5f7f564eb1862dc315bc57e1373c6a81e87ef3ac39ab AS go
 
 # mcr.microsoft.com/azurelinux/base/core:3.0
-FROM --platform=linux/${ARCH} mcr.microsoft.com/azurelinux/base/core@sha256:2a97a634da103fbfc21697f0136231f17ad18f7880a5be94488c5d2b15eb02d3 AS mariner-core
+FROM --platform=linux/${ARCH} mcr.microsoft.com/azurelinux/base/core@sha256:a452d39c91576f5a2c983c7d3b62521fabd08e16b4a7237e24bf2be3b06e1651 AS mariner-core
 
 FROM go AS azure-vnet
 ARG OS
--- a/cns/Dockerfile
+++ b/cns/Dockerfile
@@ -8,10 +8,10 @@ ARG OS
 FROM --platform=linux/${ARCH} mcr.microsoft.com/oss/go/microsoft/golang@sha256:bc7423b52b62e8f0281b5f7f564eb1862dc315bc57e1373c6a81e87ef3ac39ab AS go
 
 # mcr.microsoft.com/azurelinux/base/core:3.0
-FROM mcr.microsoft.com/azurelinux/base/core@sha256:2a97a634da103fbfc21697f0136231f17ad18f7880a5be94488c5d2b15eb02d3 AS mariner-core
+FROM mcr.microsoft.com/azurelinux/base/core@sha256:a452d39c91576f5a2c983c7d3b62521fabd08e16b4a7237e24bf2be3b06e1651 AS mariner-core
 
 # mcr.microsoft.com/azurelinux/distroless/minimal:3.0
-FROM mcr.microsoft.com/azurelinux/distroless/minimal@sha256:c9a1ed9515e033bca0c0b3171eccec41790bfcb9e47f05130cddc0a458de46ba AS mariner-distroless
+FROM mcr.microsoft.com/azurelinux/distroless/minimal@sha256:22810fd97d6ad5ec7d5bdd5b00233a3050be01d9e26b47b16cb6f1a7f178834b AS mariner-distroless
 
 FROM --platform=linux/${ARCH} go AS builder
 ARG OS


network/network_linux.go
```diff
@@ -98,10 +98,12 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
 		if err := nu.EnableIPV4Forwarding(); err != nil {
 			return nil, errors.Wrap(err, "ipv4 forwarding failed")
 		}
-		if err := nu.UpdateIPV6Setting(1); err != nil {
-			return nil, errors.Wrap(err, "failed to disable ipv6 on vm")
+
+		// Enable IPv6 in the VM namespace so vlan sub-interfaces can carry IPv6.
+		if err := nu.UpdateIPV6Setting(0); err != nil {
+			return nil, errors.Wrap(err, "failed to enable ipv6 on vm")
```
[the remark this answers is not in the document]
QxBytes: seconding this-- this will change the existing behavior for ipv4

network/transparent_endpointclient_linux.go
```diff
@@ -16,6 +16,8 @@ const (
 	virtualGwIPString     = "169.254.1.1/32"
 	defaultGwCidr         = "0.0.0.0/0"
 	defaultGw             = "0.0.0.0"
+	defaultGwv6Cidr       = "::/0"
+	defaultGwv6           = "::"
```
QxBytes: if we use these consts in transparent vlan endpoint client linux and nowhere else, shouldn't we put them in that file?

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -396,6 +396,10 @@ func (client *TransparentVlanEndpointClient) PopulateVnet(epInfo *EndpointInfo)
 	if err != nil {
 		return errors.Wrap(err, "transparent vlan failed to disable rp filter vlan interface in vnet")
 	}
+	// Enable IPv6 forwarding in VLAN namespace
+	if err := client.netUtilsClient.EnableIPV6Forwarding(); err != nil {
+		return errors.Wrap(err, "transparent vlan failed to enable ipv6 forwarding in vnet")
+	}
```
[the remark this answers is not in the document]
QxBytes: is there some way to check if ipv6 is enabled to start enabling these?

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -637,42 +686,47 @@ func (client *TransparentVlanEndpointClient) addDefaultRoutes(linkToName string,
...
+// Helper that creates arp/neighbor entries for the current NS which map the virtual
+// gateway to destMac on a particular interfaceName.
+// For IPv4: (169.254.2.1) at 12:34:56:78:9a:bc [ether] PERM on <interfaceName>
+// For IPv6: (fe80::1234:5678:9abc) at 12:34:56:78:9a:bc [ether] PERM on <interfaceName> (if hasIPv6 is true)
+func (client *TransparentVlanEndpointClient) AddDefaultArp(interfaceName, destMac string, hasIPv6 bool) error {
+	if err := client.addDefaultNeighbors(interfaceName, destMac, virtualGwIPVlanString); err != nil {
+		return err
+	}
+	if hasIPv6 {
+		if err := client.addDefaultNeighbors(interfaceName, destMac, virtualv6GwString); err != nil {
```
QxBytes: can we make it clearer that the v6 counter part of virtualGwIPVlanString is virtualv6GwString by naming something like virtualGwIPVlanStringv6 or similar?

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -605,26 +631,49 @@ func (client *TransparentVlanEndpointClient) GetVnetRoutes(ipAddresses []net.IPN
...
+		if err := client.addDefaultRoutesHelper(linkToName, table, virtualv6GwString, defaultGwv6Cidr, defaultGwv6); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// addDefaultRoutesHelper creates routing rules for the given address family.
+// It adds an on-link route for the virtual gateway and a default route via that gateway.
+func (client *TransparentVlanEndpointClient) addDefaultRoutesHelper(linkToName string, table int, virtualGwCIDR, defaultCIDR, defaultAddr string) error {
```
QxBytes: can we document that defaultCIDR is actually the default gateway cidr for either ipv4 or ipv6? similarly for defaultAddr (just explain what its value represents in the comment)

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -430,40 +434,55 @@ func (client *TransparentVlanEndpointClient) AddEndpointRules(epInfo *EndpointIn
 
 // Add rules related to tunneling the packet outside of the VM, assumes all calls are idempotent. Namespace: vnet
 func (client *TransparentVlanEndpointClient) AddVnetRules(epInfo *EndpointInfo) error {
-	// iptables -t mangle -I PREROUTING -j MARK --set-mark <TUNNELING MARK>
```
QxBytes, line 433: can we keep the comment in addVnetMangleAndTunnelingRules (you can omit the iptables part since it can be iptables or ip6tables)

network/network_linux.go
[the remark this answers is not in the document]
estebancams: @QxBytes , can you help clarify how this field works? I see from the code hat IP6Mode will be set based on: 

	if opt.ipamAddResult.ipv6Enabled { // not specific to this particular interface
		endpointInfo.IPV6Mode = string(util.IpamMode(opt.nwCfg.IPAM.Mode)) // TODO: check IPV6Mode field can be deprecated and can we add IsIPv6Enabled flag for generic working
	}
and this overwrites config setting opt.nwCfg.IPV6Mode

What is the preferred way in CNI to signal ipv6 enablement?

network/transparent_vlan_endpointclient_linux.go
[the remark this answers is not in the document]
estebancams: done

network/transparent_vlan_endpointclient_linux.go
estebancams: done

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -605,26 +630,49 @@ func (client *TransparentVlanEndpointClient) GetVnetRoutes(ipAddresses []net.IPN
...
+// to the virtual gateway ip on linkToName device interface.
+// For IPv4: 169.254.2.1 dev <linkToName>, default via 169.254.2.1 dev <linkToName>
+// For IPv6: fe80::1234:5678:9abc dev <linkToName>, default via fe80::1234:5678:9abc dev <linkToName> (if hasIPv6 is true)
+func (client *TransparentVlanEndpointClient) addDefaultRoutes(linkToName string, table int, hasIPv6 bool) error {
+	if err := client.addDefaultRoutesHelper(linkToName, table, virtualGwIPString, defaultGwCidr, defaultGw); err != nil {
+		return err
+	}
+	if hasIPv6 {
+		if err := client.addDefaultRoutesHelper(linkToName, table, virtualv6GwString, defaultGwv6Cidr, defaultGwv6); err != nil {
+			return err
```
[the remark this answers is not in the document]
estebancams: done

network/transparent_vlan_endpointclient_linux.go
estebancams: Updated constant names for clarify

network/transparent_vlan_endpointclient_linux.go
estebancams: done

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -637,25 +685,39 @@ func (client *TransparentVlanEndpointClient) addDefaultRoutes(linkToName string,
...
+// For IPv4: (169.254.2.1) at 12:34:56:78:9a:bc [ether] PERM on <interfaceName>
+// For IPv6: (fe80::1234:5678:9abc) at 12:34:56:78:9a:bc [ether] PERM on <interfaceName> (if hasIPv6 is true)
+func (client *TransparentVlanEndpointClient) AddDefaultArp(interfaceName, destMac string, hasIPv6 bool) error {
+	if err := client.addDefaultNeighbors(interfaceName, destMac, virtualGwIPString); err != nil {
+		return err
+	}
+	if hasIPv6 {
+		if err := client.addDefaultNeighbors(interfaceName, destMac, virtualv6GwString); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
estebancams: done

network/transparent_vlan_endpointclient_linux_test.go
```diff
@@ -615,14 +613,23 @@ func TestTransparentVlanDeleteEndpoints(t *testing.T) {
...
+	}
 
 	tests := []struct {
 		name       string
 		client     *TransparentVlanEndpointClient
 		ep         *endpoint
 		wantErr    bool
 		wantErrMsg string
-		routesLeft func() (int, error)
 	}{
```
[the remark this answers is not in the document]
estebancams: done

network/transparent_vlan_endpointclient_linux_test.go
```diff
@@ -1034,6 +1137,225 @@ func createFunctionWithFailurePattern(errorPattern []error) func() error {
...
+		},
+		{
+			name: "IPv6 neighbor fails with invalid mac",
+			client: &TransparentVlanEndpointClient{
+				netlink: netlink.NewMockNetlink(false, ""),
+			},
+			ifName:  "eth0",
+			destMac: "invalid-mac",
+			hasIPv6: true,
+			wantErr: true,
```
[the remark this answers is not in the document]
estebancams: done

network/transparent_vlan_endpointclient_linux_test.go
```diff
@@ -1034,6 +1029,225 @@ func createFunctionWithFailurePattern(errorPattern []error) func() error {
...
+func TestAddVnetRulesIPTables(t *testing.T) {
+	t.Run("IPv6 rules", func(t *testing.T) {
+		mockIPT := &mockIPTablesClient{}
+		client := &TransparentVlanEndpointClient{
+			vlanIfName:     "eth0.1",
+			iptablesClient: mockIPT,
+		}
+
+		client.addVnetMangleAndTunnelingRules(iptables.V6, vishnetlink.FAMILY_V6)
+
```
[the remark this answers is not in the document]
estebancams: done

network/transparent_vlan_endpointclient_linux_test.go
```diff
@@ -1034,6 +1137,225 @@ func createFunctionWithFailurePattern(errorPattern []error) func() error {
...
+	t.Run("IPv4 rules", func(t *testing.T) {
+		mockIPT := &mockIPTablesClient{}
+		client := &TransparentVlanEndpointClient{
+			vlanIfName:     "eth0.1",
+			iptablesClient: mockIPT,
+		}
+
+		client.addVnetMangleAndTunnelingRules(iptables.V4, vishnetlink.FAMILY_V4)
+
```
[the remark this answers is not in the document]
estebancams: done

network/network_linux.go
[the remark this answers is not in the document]
QxBytes: Just looking at the transparent case, it seems to check nwInfo.IPV6Mode, though I think we can probably use the `nwInfo.IsIPv6Enabled` field here. The TODO comment mentions we probably don't need to look at IPV6Mode anymore? Since nwInfo is an EndpointInfo I think you might even be able to use the same check for ipv6 you made in transparent vlan. As long as there is some guard and it works when you test it should be fine.

Looking at this further, was there a reason we didn't just do epInfo.isIPv6Enabled in transparent vlan?

network/transparent_vlan_endpointclient_linux.go
QxBytes: Okay looking at this again the const is actually used in transparent mode so this can stay as it was, sorry about that.

network/network_linux.go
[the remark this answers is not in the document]
estebancams: I found out that on CNI multitenancy mode (which is the case I'm targeting with this PR) is not setting ipv6 enablement flags since multitenancy does not use invoker interface.
I opted then for implementing this for multitenancy (traditionally ipv4-only) such ipv6 is added to ipamresult just like for invoker cases.
Based on that, the code now enables or not ipv6 in the host based on that flag (that propagates using existent code during ADD CALL flow)

@QxBytes can you review thi change?

network/transparent_endpointclient_linux.go
```diff
@@ -13,16 +13,16 @@ import (
...
+	virtualGwIPString       = "169.254.1.1/32"
+	defaultGwCidr           = "0.0.0.0/0"
+	defaultGw               = "0.0.0.0"
+	virtualGwIPVlanStringv6 = "fe80::1234:5678:9abc/128"
+	defaultv6Cidr           = "::/0"
+	ipv4Bits                = 32
+	ipv6Bits                = 128
+	ipv4FullMask            = 32
+	ipv6FullMask            = 128
+	defaultHostVethHwAddr   = "aa:aa:aa:aa:aa:aa"
```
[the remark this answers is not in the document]
estebancams: Won't fix.

network/transparent_vlan_endpointclient_linux_test.go
```diff
@@ -615,97 +613,73 @@ func TestTransparentVlanDeleteEndpoints(t *testing.T) {
...
 
 	tests := []struct {
-		name       string
-		client     *TransparentVlanEndpointClient
-		ep         *endpoint
-		wantErr    bool
-		wantErrMsg string
-		routesLeft func() (int, error)
+		name                 string
+		ep                   *endpoint
```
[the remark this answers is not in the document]
estebancams: done

network/network_linux.go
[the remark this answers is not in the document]
QxBytes: oh I see, since we're in multitenancy we use cni/network/multitenancy.go instead of cni/network/invoker_cns.go. Invoker cns sets the ipv6Enabled flag to true, but multitenancy does not. Is that what you saw when testing? I'm wondering if it would be easier to just modify multitenancy.go to set this flag in ipamAddResult so we can mimic invoker_cns.go's behavior in that case? Then we wouldn't need to have those hasIPv6Address calls.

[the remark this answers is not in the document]
estebancams: Well, that's why I'm doing, did you check my last change?
image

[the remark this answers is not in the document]
QxBytes: hmm okay I think I started reviewing before the newest changes pushed-- can we also add a check that the flag is set in multitenancy_test.go , and would it make sense to replace those has ipv6 address calls with checking the flag?

cni/network/multitenancy.go
```diff
@@ -228,6 +228,14 @@ func (m *Multitenancy) GetAllNetworkContainers(
 
 		ipconfigs, routes := convertToIPConfigAndRouteInfo(ifInfo.NCResponse)
 		ifInfo.IPConfigs = append(ifInfo.IPConfigs, ipconfigs...)
+
+		// Look for ipv6
+		for _, ipconfig := range ipconfigs {
+			if ipconfig.Address.IP.To4() == nil {
+				ipamResult.ipv6Enabled = true
+			}
+		}
```
ZetaoZhuang, line 237: nit: why not check the ncresponse[i].IPv6Configuration directly

network/transparent_vlan_endpointclient_linux.go
estebancams: I think it is ok, they are separated sceanrios but can or not share some constants. We could have a common place to put them but is not a  big deal atm

network/network_linux.go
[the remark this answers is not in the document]
estebancams: I'll check that, seems a good idea. Only thing is we need to modify the unit tests to force the flag every time since in prod flag is set upstream. Not terrible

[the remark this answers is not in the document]
estebancams: fixed

estebancams:
> looks good for the most part; just wanted a bit more explanation in comments for some of it. Were any changes needed for the apipa gateway or is that ipv4 only/already covered? Also please address the linter and any copilot comments.

There are no ipv6 requirements for apipa and I don't think there will be. IPv6 goal is to solve vnet's ipv4 scarcity while apipa is vm local and container limits per vm are well under ipv4 capabilities

cni/network/multitenancy.go
estebancams: Seems equivalent to me. What would be the improvement?

network/network_linux.go
```diff
@@ -98,10 +98,14 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
 		if err := nu.EnableIPV4Forwarding(); err != nil {
 			return nil, errors.Wrap(err, "ipv4 forwarding failed")
 		}
-		if err := nu.UpdateIPV6Setting(1); err != nil {
-			return nil, errors.Wrap(err, "failed to disable ipv6 on vm")
+
+		// Enable IPv6 in the VM namespace so vlan sub-interfaces can carry IPv6.
+		if nwInfo.IsIPv6Enabled {
```
msvik, line 103: how is this getting set?

network/network_linux.go
```diff
@@ -98,10 +98,14 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
...
+			if err := nu.UpdateIPV6Setting(0); err != nil {
+				return nil, errors.Wrap(err, "failed to enable ipv6 on vm")
+			}
 		}
-		// Blocks wireserver traffic from apipa nic
+		// Blocks wireserver traffic from apipa nic (IPv4 only)
```
msvik, line 108: why is this IPv4 only?

estebancams: wireserver is only ipv4 afaik

network/network_linux.go
estebancams: check changes in multitenancy.go . Then from ipamresult this is set to the each endpoint by preexisting code

network/transparent_endpointclient_linux.go
```diff
@@ -13,16 +13,16 @@ import (
...
-	defaultv6Cidr         = "::/0"
-	ipv4Bits              = 32
-	ipv6Bits              = 128
-	ipv4FullMask          = 32
-	ipv6FullMask          = 128
-	defaultHostVethHwAddr = "aa:aa:aa:aa:aa:aa"
+	virtualGwIPString       = "169.254.1.1/32"
+	defaultGwCidr           = "0.0.0.0/0"
+	defaultGw               = "0.0.0.0"
+	virtualGwIPVlanStringv6 = "fe80::1234:5678:9abc/128"
```
tamilmani1989: i would suggest to keep it `virtualv6GwString`  . This has nothing to do with vlan or non-vlan and gw remains same for both

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -19,6 +19,8 @@ import (
 )
 
 const (
+	defaultIPv6Prefix     = "::/0"
+	defaultIPv6Addr       = "::"
```
tamilmani1989: what ipv6 address is this. is it gateway? can we reflect in naming the variable?

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -606,25 +662,38 @@ func (client *TransparentVlanEndpointClient) GetVnetRoutes(ipAddresses []net.IPN
...
 	}
-	// Difference between interface name in addRoutes and DevName: in RouteInfo?
 	if err := addRoutes(client.netlink, client.netioshim, linkToName, []RouteInfo{routeInfo}); err != nil {
 		return err
 	}
 
-	// Add default route (ip route add default via 169.254.2.1 dev eth0)
-	_, defaultIPNet, _ := net.ParseCIDR(defaultGwCidr)
-	dstIP := net.IPNet{IP: net.ParseIP(defaultGw), Mask: defaultIPNet.Mask}
+	// Add default route via virtual gateway
```
tamilmani1989: lets not remove the ip route cmd instead we can also add for ipv6 default cmd

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -637,42 +706,47 @@ func (client *TransparentVlanEndpointClient) addDefaultRoutes(linkToName string,
...
-// Example: (169.254.2.1) at 12:34:56:78:9a:bc [ether] PERM on <interfaceName>
-func (client *TransparentVlanEndpointClient) AddDefaultArp(interfaceName, destMac string) error {
-	_, virtualGwNet, _ := net.ParseCIDR(virtualGwIPVlanString)
-	logger.Info("Adding static arp for",
-		zap.String("IP", virtualGwNet.String()), zap.String("MAC", destMac))
+// Helper that creates arp/neighbor entries for the current NS which map the virtual
+// gateway to destMac on a particular interfaceName.
+// For IPv4: (169.254.2.1) at 12:34:56:78:9a:bc [ether] PERM on <interfaceName>
+// For IPv6: (fe80::1234:5678:9abc) at 12:34:56:78:9a:bc [ether] PERM on <interfaceName> (if hasIPv6 is true)
+func (client *TransparentVlanEndpointClient) AddDefaultArp(interfaceName, destMac string, hasIPv6 bool) error {
```
tamilmani1989: instead of passing hasIPv6, is it better to pass gateway address and no need to have if check inside this api

tamilmani1989: also same comment for `addDefaultRoutes` api

network/transparent_endpointclient_linux.go
estebancams: done

network/transparent_vlan_endpointclient_linux.go
estebancams: done

network/transparent_vlan_endpointclient_linux.go
estebancams: Pass gateway address where? The gw will have to be determined based on hasIPv6 in any case. Can you detail your suggestion in code?

network/transparent_vlan_endpointclient_linux.go
estebancams: Resolved using default prefix. Removed unnecessary constant

estebancams:
> can you also consider adding possible UTs for these code changes?

What is not covered by UTs in this PR?

network/transparent_vlan_endpointclient_linux.go
```diff
@@ -637,42 +705,47 @@ func (client *TransparentVlanEndpointClient) addDefaultRoutes(linkToName string,
...
```
tamilmani1989:
```suggestion
func (client *TransparentVlanEndpointClient) AddDefaultArp(interfaceName, destMac string, gwIP string) error {
```

tamilmani1989:
> > can you also consider adding possible UTs for these code changes?
> 
> What is not covered by UTs in this PR?

can validate if ipv6 configurations are correctly set. ipv6 forwarding is enable correctly if ipv6 is enabled and you could think few more like this since you are PR author :)

network/transparent_vlan_endpointclient_linux.go
estebancams: done

estebancams:
> > > can you also consider adding possible UTs for these code changes?
> > 
> > 
> > What is not covered by UTs in this PR?
> 
> can validate if ipv6 configurations are correctly set. ipv6 forwarding is enable correctly if ipv6 is enabled and you could think few more like this since you are PR author :)

Since you are the reviewer you could have pointed out to that in the first place instead of jsut suggesting "adding more tests"

estebancams:
> > > > can you also consider adding possible UTs for these code changes?
> > > 
> > > 
> > > What is not covered by UTs in this PR?
> > 
> > 
> > can validate if ipv6 configurations are correctly set. ipv6 forwarding is enable correctly if ipv6 is enabled and you could think few more like this since you are PR author :)
> 
> Since you are the reviewer you could have pointed out to that in the first place instead of jsut suggesting "adding more tests"

* Added tests for ipv6 enablement in GetAllNetworkContainers (test presence of ipv6 in the result from CNS)
* Added UT for existent uncovered function (called during nw creation):
image

network/transparent_vlan_endpointclient_linux_test.go
```diff
@@ -615,97 +613,72 @@ func TestTransparentVlanDeleteEndpoints(t *testing.T) {
...
-			ep: &endpoint{
-				IPAddresses: IPAddresses,
-			},
-			routesLeft: func() (int, error) {
-				return numDefaultRoutes, nil
-			},
+			name:                 "Delete endpoint with IPv4 addresses",
+			ipAddresses:          IPAddresses,
+			wantDeleteRouteCount: 2,
+			wantDeleteLinkCalled: true,
```
tamilmani1989, line 637: what's the diff between ipv4 and dualstack? is delete route count 2 in both cases?

tamilmani1989:
> > > > can you also consider adding possible UTs for these code changes?
> > > 
> > > 
> > > What is not covered by UTs in this PR?
> > 
> > 
> > can validate if ipv6 configurations are correctly set. ipv6 forwarding is enable correctly if ipv6 is enabled and you could think few more like this since you are PR author :)
> 
> Since you are the reviewer you could have pointed out to that in the first place instead of jsut suggesting "adding more tests"

I think you got offended. My intention was not to offend you.  If there are some unit tests added, i could have pointed out specific tests if anything missing. When I reviewed, there was no unit tests to begin with, that's why i added generic comment "Add possible unit tests" for the code you added. I don't want to restrain you with list of UTs I shared, instead it's better if you can come up with what you think (may be better than what I'd have suggested) and then during second review, if i find anything missing, I would ve pointed out. Anyway, thanks for adding UTs and fixing other comments

network/transparent_vlan_endpointclient_linux_test.go
estebancams: it is the same in both cases, what is the question here?

tamilmani1989: so there is no difference between these 2 tests? should we validate routes (dst, gw, dev) are expected

estebancams: Still don't understand what your comment is about. There are 2 clearly different tests here, one has 2 ipv4 and the other 1 ipv4 and 1 ipv6. Then in both cases delete ip route is called 2 times

tamilmani1989: I know there are 2 tests but you haven't validated if expected routes are getting removed. do you know what those routes are? can you add comment on what routes getting removed.

estebancams: That is not counting routes but calls to deleteIproutes. The name might be misleading, I can fix that in a sperated PR since it is a minor non functional change

csfmomo (approved):

PR #4242, opened 2026-02-27
Harden state integration tests

rbtr:
**Reason for Change**:

Enhance the CNS state integration tests to set up for future state migrations and improvements.

These changes are AI assisted - I am also introducing guidance/docs in agents.md files at multiple levels in the repository, and hooks in the default spots so that they are used as context, which were used to guide the agent for these changes.

### DIFF
--- a/agents.md
+++ b/agents.md
@@ -1,0 +1,65 @@
+# agents.md
+
+Behavioral guidelines to reduce common LLM coding mistakes. Merge with project-specific instructions as needed.
+
+**Tradeoff:** These guidelines bias toward caution over speed. For trivial tasks, use judgment.
+
+## 1. Think Before Coding
+
+**Don't assume. Don't hide confusion. Surface tradeoffs.**
+
+Before implementing:
+- State your assumptions explicitly. If uncertain, ask.
+- If multiple interpretations exist, present them - don't pick silently.
+- If a simpler approach exists, say so. Push back when warranted.
+- If something is unclear, stop. Name what's confusing. Ask.
+
+## 2. Simplicity First
+
+**Minimum code that solves the problem. Nothing speculative.**
+
+- No features beyond what was asked.
+- No abstractions for single-use code.
+- No "flexibility" or "configurability" that wasn't requested.
+- No error handling for impossible scenarios.
+- If you write 200 lines and it could be 50, rewrite it.
+
+Ask yourself: "Would a senior engineer say this is overcomplicated?" If yes, simplify.
+
+## 3. Surgical Changes
+
+**Touch only what you must. Clean up only your own mess.**
+
+When editing existing code:
+- Don't "improve" adjacent code, comments, or formatting.
+- Don't refactor things that aren't broken.
+- Match existing style, even if you'd do it differently.
+- If you notice unrelated dead code, mention it - don't delete it.
+
+When your changes create orphans:
+- Remove imports/variables/functions that YOUR changes made unused.
+- Don't remove pre-existing dead code unless asked.
+
+The test: Every changed line should trace directly to the user's request.
+
+## 4. Goal-Driven Execution
+
+**Define success criteria. Loop until verified.**
+
+Transform tasks into verifiable goals:
+- "Add validation" → "Write tests for invalid inputs, then make them pass"
+- "Fix the bug" → "Write a test that reproduces it, then make it pass"
+- "Refactor X" → "Ensure tests pass before and after"
+
+For multi-step tasks, state a brief plan:
+```
+1. [Step] → verify: [check]
+2. [Step] → verify: [check]
+3. [Step] → verify: [check]
+```
+
+Strong success criteria let you loop independently. Weak criteria ("make it work") require constant clarification.
+
+---
+
+**These guidelines are working if:** fewer unnecessary changes in diffs, fewer rewrites due to overcomplication, and clarifying questions come before implementation rather than after mistakes.
--- a/cns/agents.md
+++ b/cns/agents.md
@@ -1,0 +1,58 @@
+# CNS Agent Brief (AKS/Azure CNI)
+
+This brief is for code contributors working on Container Networking Service (CNS) in this repo.
+
+## What CNS is
+CNS is the container networking daemon running on every AKS node for Azure CNI. It bridges the AKS control plane and node dataplane by managing pod IP allocation state and coordinating with the Azure networking control plane for routable VNET IPs.
+
+In Azure CNI modes where pods receive routable VNET IPs, CNS tracks the goal state from the NodeNetworkConfig (NNC) CRD and only exposes IPs to pods once the Network Container (NC) is programmed by NMAgent.
+
+## Core architecture (high level)
+- HTTP REST API used by CNI/azure-ipam for IP assignment and release.
+- Controller-runtime reconcilers that watch NNCs (and SwiftV2 CRDs) and reconcile desired NC/IP state.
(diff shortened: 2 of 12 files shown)


test/validate/cmd/summarydiff/main_test.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	stats := aggregate(summary)
+	assert.Equal(t, 2, stats.TotalChecks)
+	assert.Equal(t, 1, stats.FailedChecks)
+	assert.Equal(t, 1, stats.MissingIPs)
+	assert.Equal(t, 1, stats.UnexpectedIPs)
+	assert.Equal(t, 1, stats.DuplicateIPs)
+	assert.Equal(t, 5, stats.ExpectedIPsSum)
+	assert.Equal(t, 6, stats.ActualIPsSum)
```
jpayne3506, line 35: No `HasMismatch()` confirmation?

agents.md
```diff
@@ -0,0 +1,65 @@
+# agents.md
```
jpayne3506, line 1: Any intent to add this to our agents tab so we can leverage it instead of base copilot for PRs? Or does the `copilot-instruction.md` handle that?

Image

jpayne3506: Curious how the above would work with your design. Might need to inverse the flow to bring in the multiple files... Instead of root -> component. It goes component -> root.

My understanding of the intent is /agent.md + cns/agent.md would direct the agent if working within the cns/ directory.

But the agent tab I am talking about probably can only handle a single .md file. If you specify only the cns/agent.md file then it loses the root level guidelines..

cns/agents.md
```diff
@@ -0,0 +1,57 @@
+# CNS Agent Brief (AKS/Azure CNI)
```
jpayne3506, line 1: no conflist generator bit?

jpayne3506 (dismissed):
Validator and pipeline changes lgtm. Would like to see what the pipeline output changes to, if at all. 

We don't need to play with the agent tab in this PR, but it would be ideal that we could reference a custom agent tailor made for specific reviewers. Ideally a workflow could call them in automatically.

cns/agents.md
rbtr: i'm just starting them, will add more context as it comes up

agents.md
vipul-21: Seen the agent creating READMEs for every fix/refactoring. Is that something you saw with these guidelines ?

test/validate/utils_test.go
```diff
@@ -0,0 +1,39 @@
+package validate
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCompareIPsDetailed_Match(t *testing.T) {
```
vipul-21, line 9: Wondering why it created two separate test cases here and a single test for `main_test` ? Guidelines should apply for test files too as well, so why ?

vipul-21 (dismissed):
PR itself lgtm, but a bit curious about the process you followed with these guidelines to create the PR.
-  Did you ask the agent to plan first, ack for the implementation and then implement  ? How did it come up with the pipeline variables. For example `failOnRegression` is Boolean, I think this should be always `true`  so why that is added ?
- I see the `agents.md` mentions ask if there is anything unclear ? What counts as unclear from the perspective of the agent ? I see the cns agent file talks about DNC/RC, will that make agent ask us if there is a cns code change before progressing always ?
- If there is a suggestion that improves something along with the changes, should it atleast let us know ?

test/validate/utils_test.go
rbtr: not sure. I would have written these table driven, but the agents seem to strongly prefer more verbose, wholly separate tests. I think this is just something that needs stronger guidance?

rbtr:
> Did you ask the agent to plan first, ack for the implementation and then implement ? How did it come up with the pipeline variables. For example failOnRegression is Boolean, I think this should be always true so why that is added ?

Yes - but I was actually working a more complete state refactoring and had it break these test changes out to go in ahead of any substantial code change. That's why there's some dangling functionality like that bool, it was/is setting up for a future state platform migration.

> I see the agents.md mentions ask if there is anything unclear ? What counts as unclear from the perspective of the agent ? I see the cns agent file talks about DNC/RC, will that make agent ask us if there is a cns code change before progressing always ?

That is to prevent it from assuming things if given ambiguous instructions. It may ask clarifying questions around priority/style when asked to refactor, for example, instead of just doing things. 


> If there is a suggestion that improves something along with the changes, should it atleast let us know ?

I have seen that when unfocused the agents tend to make breaking "improvements" without fully considering the side-effects or why something was done the way it was done originally. I think it's better to start out as tightly scoped as possible - work on only exactly what the prompt was, nothing extra - so that the results stay high quality. It may be that we can loosen this up in the future.

cns/agents.md
```diff
@@ -0,0 +1,57 @@
...
+2) DNC allocates IPs as Network Containers (NCs) and publishes them to the fabric.
+3) NMAgent programs the NC; CNS watches for NC version readiness via IMDS/wireserver.
+4) CNS ingests NNC as goal state; once NC is ready, IPs become Available.
+5) CNI/azure-ipam calls CNS to RequestIPConfigs/ReleaseIPConfigs during pod lifecycle.
+6) CNS acts as the IPAM store for pod IP allocation and release.
+
+## Primary CNS APIs used by AKS
+Primary (current) IPAM APIs:
+- POST /network/requestipconfigs — allocate one or more IPs for a pod.
+- POST /network/releaseipconfigs — release IPs for a pod.
```
behzad-mir, line 28: You may want to add EndpointHandlerAPI here too, since it is called by Stateless CNI across platforms now.

cns/agents.md
```diff
@@ -0,0 +1,57 @@
...
+- IPAM pool monitor (dynamic IP scenarios) to resize IP pools on demand. Not used in fixed-prefix modes (e.g., overlay).
+- IMDS/wireserver integration to query NMAgent NC status.
+
+## Core flow (node → pod IP assignment)
+1) DNC/RC creates a NodeNetworkConfig (NNC) for the node.
+2) DNC allocates IPs as Network Containers (NCs) and publishes them to the fabric.
+3) NMAgent programs the NC; CNS watches for NC version readiness via IMDS/wireserver.
+4) CNS ingests NNC as goal state; once NC is ready, IPs become Available.
+5) CNI/azure-ipam calls CNS to RequestIPConfigs/ReleaseIPConfigs during pod lifecycle.
+6) CNS acts as the IPAM store for pod IP allocation and release.
```
behzad-mir, line 22: CNS also update the endpointState via EndpointHandlerAPI to store CNI state after Datapath provisioning

cns/agents.md
```diff
@@ -0,0 +1,57 @@
...
+- REST API handlers: cns/restserver/
+- Reconcilers: cns/kubecontroller/
+- IPAM pool monitor: cns/ipampool/
+
+## Tests to run for CNS changes
+- Tests in the packages above (restserver, kubecontroller, ipampool, service) are the first-line regression checks.
+
+## Critical caution areas
+CNS initialization and state management are high-risk code paths. Behavioral changes can cause large-scale impact. Keep changes minimal and validate thoroughly.
+
```
behzad-mir, line 59: We may want to also include the Asynch delete flow as well.

behzad-mir (changes requested):

cns/agents.md
rbtr: updated

cns/agents.md
rbtr: updated

cns/agents.md
rbtr: updated

PR #4243, opened 2026-03-02
AZP->GHA migration plan

rbtr:
**Reason for Change**:

As we migrate to managed Dalec, our hard requirement on AZP goes away. To improve iteration speed, CI reliability, and understandability (->and better support agentic workloads), we can begin migrating AZP pipelines in to native GHA workflows. 

AZP has always been a pain - the split GH/AZP sources of truth for pipeline results, the lack of visibility to external contributors, our enterprise changing settings without telling us that break our workflows or the challenge to even get them to run in the first place. Consolidating on GHA will be an entirely better devx.

The intent is that setting up the plan like this will provide context for much/all of this work to be implemented by the various agents.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Includes the same base agents.md proposed in other PRs...will rebase whichever one doesn't merge first.

### DIFF
--- a/.github/workflows/_reusable-go-setup.yaml
+++ b/.github/workflows/_reusable-go-setup.yaml
@@ -1,0 +1,41 @@
+name: Reusable Go Setup
+on:
+  workflow_call:
+    inputs:
+      generate-bpf:
+        description: "Run BPF library build and go generate"
+        type: boolean
+        default: false
+    outputs:
+      go-version:
+        description: "Go version resolved from go.mod"
+        value: ${{ jobs.setup.outputs.go-version }}
+
+permissions:
+  contents: read
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      go-version: ${{ steps.go.outputs.go-version }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Go
+        id: go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - name: Install BPF prerequisites and generate
+        if: inputs.generate-bpf
+        run: make bpf-lib && go generate ./...
+
+      - name: Upload generated BPF code
+        if: inputs.generate-bpf
+        uses: actions/upload-artifact@v6
+        with:
+          name: generated-bpf-code
+          path: ./bpf-prog/azure-block-iptables/pkg/blockservice
+          retention-days: 1
--- a/.github/workflows/_reusable-go-setup.yaml
+++ b/.github/workflows/_reusable-go-setup.yaml
@@ -20,11 +20,13 @@ jobs:
     outputs:
       go-version: ${{ steps.go.outputs.go-version }}
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
 
       - name: Set up Go
         id: go
-        uses: actions/setup-go@v6
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version-file: go.mod
 
@@ -34,7 +36,7 @@ jobs:
 
       - name: Upload generated BPF code
         if: inputs.generate-bpf
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: generated-bpf-code
           path: ./bpf-prog/azure-block-iptables/pkg/blockservice
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -1,0 +1,181 @@
+name: Unit Tests
+on:
+  workflow_dispatch:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+    branches:
+      - master
+      - "release/*"
+  merge_group:
+    types:
+      - checks_requested
+  push:
+    branches:
+      - master
+  schedule:
+    - cron: "0 2 * * *"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: unit-tests-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  # ---------------------------------------------------------------------------
+  # BPF code generation (Linux prerequisite)
+  # ---------------------------------------------------------------------------
+  generate:
+    name: Generate BPF Code
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - name: Build BPF library and run go generate
+        run: make bpf-lib && go generate ./...
+
+      - name: Upload generated code
+        uses: actions/upload-artifact@v6
+        with:
+          name: generated-bpf-code
+          path: ./bpf-prog/azure-block-iptables/pkg/blockservice
+          retention-days: 1
+
+  # ---------------------------------------------------------------------------
+  # Linux unit tests (PR fast lane + merge queue slow lane)
+  # ---------------------------------------------------------------------------
+  test-linux:
+    name: Linux Unit Tests
+    needs: generate
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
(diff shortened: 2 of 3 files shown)


.pipelines/migration.md
```diff
@@ -0,0 +1,477 @@
...
+- Every phase must be independently verifiable before starting the next.
+
+### Merge Queue Coverage Model
+
+- This repo uses merge queue; migration must preserve two execution lanes:
+  - `PR fast lane`: small/high-signal checks on `pull_request` for rapid developer feedback.
+  - `merge queue slow lane`: broader/slower checks on `merge_group` after approval.
+- A check is not considered migrated until both lanes are covered where applicable.
+- For required checks, ensure merge-queue-context checks are configured in branch protection,
+  not only PR-context checks.
```
jpayne3506, line 62: I think we should define what tests are fast lane. Otherwise it will choose what it thinks is "fast" without context of how long it takes.

.pipelines/migration.md
```diff
@@ -0,0 +1,477 @@
...
+
+### Migration Governance (Applies to All Phases)
+
+- Use a three-step rollout for each migrated check:
+  - `shadow`: run in GHA but do not gate merge, compare against ADO outcomes.
+  - `required`: add the GHA check to branch protection, keep ADO check required.
```
jpayne3506, line 68: Are we able to assign something to branch protection from the repo files..? I think this process will have to be manual

.pipelines/migration.md
```diff
@@ -0,0 +1,477 @@
...
+  log collection) to avoid drift across multiple workflow files.
+
+---
+
+### Phase 0 - CI Control Plane and Baseline
+
+**Value:** Highest leverage. Prevents migration drift and accidental coverage regressions.  
+**New files:**
+- `.github/workflows/_reusable-go-setup.yaml` (optional reusable setup)
+- `.github/workflows/_reusable-azure-auth.yaml` (OIDC login wrapper)
```
jpayne3506, line 85: Lets ensure that this cannot be changed per PR. Only allow reference to a version that exists on the default branch.

.pipelines/migration.md
```diff
@@ -0,0 +1,477 @@
...
+- `.github/workflows/required-checks-audit.yaml` (optional weekly audit)
```
jpayne3506, line 86: I think you mean for these to be `.github/actions/*`

.pipelines/migration.md
```diff
@@ -0,0 +1,477 @@
...
+
+#### Job structure
+
+1. **Create cluster**: `az aks create` with appropriate node SKU and CNI config
+2. **Deploy**: pull test image from GHCR (`ghcr.io/...:<tag>`) and deploy CNS/CNI via kubectl/helm
+3. **Run E2E**: execute e2e test suite (see `make test-k8se2e-only` and `hack/scripts/`)
+4. **Cleanup**: `az group delete --name <rg> --yes --no-wait` in an `always:` post-step
+
+Operational guardrails:
+- Use deterministic resource group naming (`acn-gha-<workflow>-<runid>`) and tag all resources
```
jpayne3506, line 318: workflow might need to be a unique identifier. Once we hit the issue we can fix it.

.pipelines/migration.md
```diff
@@ -0,0 +1,477 @@
...
+3. `cilium-overlay-withhubble`
+4. `cilium-dualstack-overlay`
+5. `cilium-nodesubnet`
+
+#### Reference templates
+
+- `.pipelines/templates/cilium-tests.yaml` — test sequence (status, restarts, scaling, connectivity)
+- `.pipelines/templates/cilium-connectivity-tests.yaml` — connectivity validation steps
+- `.pipelines/templates/cilium-cli.yaml` — Cilium CLI installation
+
```
jpayne3506, line 384: Would be in favor of migrating existing cilium-nightly pipeline to get an early signal on cilium testing.

.pipelines/migration.md
```diff
@@ -0,0 +1,477 @@
...
+| Cilium E2E variants | `.pipelines/singletenancy/cilium/*` | `e2e-cilium.yaml` | planned |
+| SwiftV2 long-running | `.pipelines/swiftv2-long-running/pipeline.yaml` | N/A or future | deferred |
+| NPM pipelines | `.pipelines/npm/*` | N/A | frozen |
+| OneBranch run pipeline | `.pipelines/run-pipeline.yaml` | N/A | retire with prod offload |
+
+## Open Decisions (Resolve Before Phase 4)
+
+- Azure auth model for AKS E2E: OIDC federation preferred; identify owning team and setup timeline.
+- AKS region and quota budget for nightly runs: define default region(s), VM SKUs, and max concurrent runs.
+- PR-on-demand E2E trigger model: label-triggered, comment-triggered, or manual-dispatch only.
```
jpayne3506, line 476: I like the idea of both label + comment triggers... I could be wrong, but wouldn't a label trigger be a 1 time operation? How would you retrigger? Manually? 

If we use both we would have the best of both worlds. 
- Apply e2e label when directory has been changed -> trigger respective E2E
- Comment which E2E we want triggered, retriggered, or generic testing

jpayne3506 (dismissed):
I like this phased approach. I do not expect this to work out of the box, but that is okay 🚀 . Key issues I foresee are:
- Unique naming patterns
- Under/Over scoping of tokens 
- Concurrency killing runs in the merge queue
- Wrong assumptions in what our ADO pipeline variable are providing, this then leads into hallucinations. 
- Referencing proprietary pipelines over our generic ADO pipelines, more hallucinations

.pipelines/migration.md
rbtr: I'm fine to leave these types of things ambiguous for now. I don't expect the plan to be comprehensive enough that it can be implemented fully autonomously yet.

.pipelines/migration.md
rbtr: You're right, I don't think rulesets can be checked in (but you can export/import them as JSON so they're stored _somewhere_)

.pipelines/migration.md
rbtr: PR workflows always run from the default branch AFAIK but references to this should be `@SHA` for safety

.pipelines/migration.md
rbtr: Either works. I think they didn't previously separate them clearly, but you're right actions is more correct for these today.

.pipelines/migration.md
rbtr: ack

.pipelines/migration.md
rbtr: ack

.pipelines/migration.md
rbtr: yeah you're probably right about label trigger, not sure why that is here. I think an /ok-to-test and /retest comment and manual triggers is sufficient

nairashu (dismissed):

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,181 @@
...
+        with:
+          go-version-file: go.mod
+
+      - name: Run Windows unit tests
+        shell: pwsh
+        run: |
+          # NPM is deprecated; exclude npm packages.
+          # cni packages with linux-only build tags are excluded for now.
+          go test -timeout 30m -covermode atomic -coverprofile=windows-coverage.out ./cns/... ./platform/...
+          go tool cover -func=windows-coverage.out
```
jpayne3506, line 133: https://github.com/Azure/azure-container-networking/blob/ee12aec5e61c82f128204c267c3b9f463ee2a04c/.pipelines/templates/run-unit-tests.yaml#L57-L58

Im confused, why are we excluding windows cni tests..?

jpayne3506: https://github.com/Azure/azure-container-networking/blob/ee12aec5e61c82f128204c267c3b9f463ee2a04c/cni/network/network_windows_test.go#L1-L31

.github/workflows/_reusable-go-setup.yaml
```diff
@@ -0,0 +1,41 @@
...
+    outputs:
+      go-version: ${{ steps.go.outputs.go-version }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Go
+        id: go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
```
jpayne3506, line 29: Might want to leverage this 

https://github.com/Azure/azure-container-networking/blob/039376ee2cedb4286ff7ca6f1d3491c344a0c914/.github/workflows/unit-tests.yml#L36-L44

cns/agents.md
```diff
@@ -0,0 +1,58 @@
+# CNS Agent Brief (AKS/Azure CNI)
```
behzad-mir, line 1: Aren't we having agents.md here too:
https://github.com/Azure/azure-container-networking/pull/4242/

behzad-mir (changes requested):

cns/agents.md
rbtr: yes, they were made from the same base branch with the agent docs. one will need to be rebased after the other merges

PR #4253, opened 2026-03-05
feat: allow istio, nfs iptables, add a route for health probe traffic

santhoshmprabhu:
**Reason for Change**:
We need to allow istio to be installed on clusters with ebpf host routing enabled. To do this, there are several changes made in this PR:
1. IPTables block logic is updated to add the istio executable, nfs scripts to the list of allowed processes that can install iptables rules.
2. In iptables monitor, a new flag is added which, when enabled, adds a route that sends reply packets from istio health probes to the loopback interface. This is essentially an implementation of the workaround described here: https://github.com/istio/istio/issues/52208#issuecomment-2488755452. We will enable this option in the sidecar on the cilium daemonset, when ebpf host routing is enabled.

Limitations:
1. istio allows the SNAT IP to be configured by the user. This PR installs routes the default IPs always.
2. There is no clean up of the route.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Validation**
1. Verified that routes are installed by the route manager, while manually running as a daemonset on a dualstack cluster.
2. Verified that istio is able to install iptables rules with the block enabled.
3. Verified that aznfs-mount is able to install DNAT rules, with the block enabled.

**Notes**:

### DIFF
--- a/azure-iptables-monitor/README.md
+++ b/azure-iptables-monitor/README.md
@@ -35,7 +35,7 @@ Follow the steps below to build and run the program:
     - The `-checkMap` flag enables checking the pinned bpf map specified in mapPath for increases. Default: `false`
     - The `-mapPath` flag specifies the pinned bpf map path to check. Default: `/azure-block-iptables-bpf-map/iptables_block_event_counter`
     - The `-terminateOnSuccess` flag, when set, will exit the program once there are no longer user iptables rules detected. Default: `false`
-    - The `-installIstioRoutes` flag causes routes to be installed that would send health-probe reply packets to the host loopback interface. Default: `false`
+    - The `-installRoutesForHealthProbeReply` flag causes routes to be installed that would send health-probe reply packets to the host loopback interface. Default: `false`
     - The program must be in a k8s environment and `NODE_NAME` must be a set environment variable with the current node.
 
 5. The program will set the `kubernetes.azure.com/user-iptables-rules` label to `true` on the specified ciliumnode resource if unexpected rules are found, or `false` if all rules match expected patterns. Proper RBAC is required for patching (patch for ciliumnodes, create for events, get for nodes).


azure-iptables-monitor/iptables_monitor.go
```diff
@@ -28,21 +29,27 @@ import (
...
+	configPath4            = flag.String("input", "/etc/config/", "Name of the directory with the ipv4 allowed regex files")
+	configPath6            = flag.String("input6", "/etc/config6/", "Name of directory with the ipv6 allowed regex files")
+	checkInterval          = flag.Int("interval", 300, "How often to check for user iptables rules and bpf map increases (in seconds)")
+	sendEvents             = flag.Bool("events", false, "Whether to send node events if unexpected iptables rules are detected")
+	ipv6Enabled            = flag.Bool("ipv6", false, "Whether to check ip6tables using the ipv6 allowlists")
+	checkMap               = flag.Bool("checkMap", false, "Whether to check the bpf map at mapPath for increases")
+	pinPath                = flag.String("mapPath", "/azure-block-iptables-bpf-map/iptables_block_event_counter", "Path to pinned bpf map")
+	terminateOnSuccess     = flag.Bool("terminateOnSuccess", false, "Whether to terminate the program when no user iptables rules found")
+	installIstioRoutesFlag = flag.Bool("install-istio-routes", false, "Whether to install loopback routes for Istio SNAT")
```
QxBytes: nit: should this be camelCase like the above option?

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -78,6 +85,27 @@ func NewEBPFClient() EBPFClient {
...
 
+// realRouteManager manages system routes via the ip command
+type realRouteManager struct{}
+
+func NewRouteManager() RouteManager {
+	return &realRouteManager{}
+}
+
+// EnsureRoute adds a route for the given IP to the loopback device, replacing any existing route
+func (r *realRouteManager) EnsureRoute(ip string, isIPv6 bool) error {
```
QxBytes: maybe I'm overthinking this but if we want to be very sure at a glance the ip is an ip and not an arbitrary string could we use a `net.IPNet`? I think it also carries ipv4/ipv6 and mask information.

azure-iptables-monitor/Dockerfile
```diff
@@ -46,12 +46,14 @@ RUN if [ "$ARCH" = "amd64" ]; then \
 RUN GOOS=$OS CGO_ENABLED=0 go generate ./bpf-prog/azure-block-iptables/...
 RUN GOOS=$OS CGO_ENABLED=0 go build -a -o /go/bin/azure-block-iptables -trimpath -ldflags "-s -w -X main.version="$AZURE_BLOCK_IPTABLES_VERSION"" -gcflags="-dwarflocationlists=true" ./bpf-prog/azure-block-iptables/cmd/azure-block-iptables
 
-FROM mariner-core AS iptables
-RUN tdnf install -y iptables
+FROM mariner-core AS iptools
```
QxBytes, line 49: I think you might need to also modify `.pipelines/build/dockerfiles/azure-iptables-monitor.Dockerfile` ? Can we also test building an image and running it on the acn official build pipeline-- I recall last time there was an issue where acn pr built the image differently.

azure-iptables-monitor/iptables_monitor.go
santhoshmprabhu: done.

azure-iptables-monitor/iptables_monitor.go
santhoshmprabhu: Tried out your suggestion, the code feels harder to read, and requires more conversions back and forth from/to strings, so I think it's okay to keep it this way.

azure-iptables-monitor/Dockerfile
santhoshmprabhu: done

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -37,7 +37,7 @@ var (
 	checkMap               = flag.Bool("checkMap", false, "Whether to check the bpf map at mapPath for increases")
 	pinPath                = flag.String("mapPath", "/azure-block-iptables-bpf-map/iptables_block_event_counter", "Path to pinned bpf map")
 	terminateOnSuccess     = flag.Bool("terminateOnSuccess", false, "Whether to terminate the program when no user iptables rules found")
-	installIstioRoutesFlag = flag.Bool("install-istio-routes", false, "Whether to install loopback routes for Istio SNAT")
+	installIstioRoutesFlag = flag.Bool("istallIstioRoutes", false, "Whether to install loopback routes for Istio SNAT")
```
QxBytes: `installIstioRoutes`?

santhoshmprabhu: fixed.

.pipelines/build/dockerfiles/azure-iptables-monitor.Dockerfile
```diff
@@ -6,13 +6,15 @@ FROM mcr.microsoft.com/azurelinux/base/core@sha256:9948138108a3d69f1dae62104599a
...
+RUN tdnf install -y iptables iproute
 
 FROM mariner-distroless AS linux
 ARG ARTIFACT_DIR
-COPY --from=iptables /usr/sbin/*tables* /usr/sbin/
-COPY --from=iptables /usr/lib /usr/lib
+COPY --from=iptools /usr/sbin/*tables* /usr/sbin/
+COPY --from=iptools /usr/sbin/ip /usr/sbin/
+COPY --from=iptools /usr/lib /usr/lib
+COPY --from=iptools /usr/lib64 /usr/lib64
```
rbtr, line 17: I was trying to _not_ just copy the entirety of `/usr/lib`, here

azure-iptables-monitor/interfaces.go
```diff
@@ -59,6 +65,7 @@ type Config struct {
 	PinPath            string
 	NodeName           string
 	TerminateOnSuccess bool
+	InstallIstioRoutes bool
```
rbtr: is there no more generic way to describe this? config on behavior, not on scenario

azure-iptables-monitor/iptables_monitor.go
```diff
@@ -28,21 +29,27 @@ import (
...
 )
 
 const (
 	label          = "kubernetes.azure.com/user-iptables-rules"
 	requestTimeout = 5 * time.Second
 )
 
+const (
+	istioIPv4 = "169.254.7.127"
+	istioIPv6 = "fd16:9254:7127:1337:ffff:ffff:ffff:ffff"
```
rbtr: perhaps the config could be `--plumb-routes-for-healthcheck-to=169.254.7.127,fd16:9254:7127:1337:ffff:ffff:ffff:ffff`
which makes the config for a specific behavior, not some named scenario, and makes these configurable too

azure-iptables-monitor/iptables_monitor.go
rbtr: why is this input arg not a `ipnet.Addr`?

azure-iptables-monitor/README.md
```diff
@@ -35,6 +35,7 @@ Follow the steps below to build and run the program:
     - The `-checkMap` flag enables checking the pinned bpf map specified in mapPath for increases. Default: `false`
     - The `-mapPath` flag specifies the pinned bpf map path to check. Default: `/azure-block-iptables-bpf-map/iptables_block_event_counter`
     - The `-terminateOnSuccess` flag, when set, will exit the program once there are no longer user iptables rules detected. Default: `false`
+    - The `-installIstioRoutes` flag causes routes to be installed that would send health-probe reply packets to the host loopback interface. Default: `false`
```
rbtr: that description is more like it. that's not istio-specific at all?

rbtr (approved):
i am not blocking but i would like this feedback addressed if possible

azure-iptables-monitor/interfaces.go
santhoshmprabhu: changed to installRoutesForHealthProbeReply

azure-iptables-monitor/iptables_monitor.go
santhoshmprabhu: I've renamed the option as described above, when discussing with TM, we decided that using the default IPs suffices. Renamed these IPs too.

azure-iptables-monitor/iptables_monitor.go
santhoshmprabhu: switched to netip.Addr

.pipelines/build/dockerfiles/azure-iptables-monitor.Dockerfile
santhoshmprabhu: iptables adds a very large number of files to /usr/lib, we can do this one as follow up I think.

santhoshmprabhu: Note that the copy of /usr/lib was there before this PR too.

santhoshmprabhu:
Disabled automerge, verifying the official build pipeline as well.

santhoshmprabhu:
Verified the image built in the official build pipeline. Merging.

PR #4265, opened 2026-03-12
[SwiftV2 IPv6] Add IPv6 dual-stack support for SwiftV2 in CNS and CNI

adjordjevic-microsoft:
**Reason for Change**:
SwiftV2 currently only supports single-stack IPv4. This change adds IPv6
dual-stack support by:

1. Adding `NetworkContainerIPv6Config` to `PodIpInfo` so CNS can carry
   IPv6 configuration alongside the existing IPv4 primary config.
2. Propagating `IPv6Configuration` in the standalone IPAM handler
   (`requestIPConfigHandlerHelperStandalone`).
3. Extending `IPResultInfo` with IPv6 fields (`ncSubnetV6Prefix`,
   `ncIPv6`, `ncGatewayIPv6Address`) and including them in structured
   logging via `MarshalLogObject`.
4. Appending an IPv6 `IPConfig` in `configureSecondaryAddResult` with
   full validation — invalid IPv6, IPv4-in-IPv6 rejection, and gateway
   checks using new sentinel errors.
5. Fixing the HCNv2 network configuration (`configureHcnNetwork`) to set
   the correct default route (`::/0`) for IPv6 subnets instead of the
   hardcoded `0.0.0.0/0`.

### DIFF
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"net/netip"
 
 	"github.com/Azure/azure-container-networking/cni"
 	"github.com/Azure/azure-container-networking/cni/log"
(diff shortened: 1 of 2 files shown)


network/network_windows.go
```diff
@@ -289,13 +289,19 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *EndpointInfo, extIf *exter
 
 	// Populate subnets.
 	for _, subnet := range nwInfo.Subnets {
+		// Choose route based on IP family
+		routeDest := defaultRouteCIDR
```
estebancams: Can you rename this to defaultIPv4Route to be symmetrical with ipv6 case?

cni/network/invoker_cns.go
```diff
@@ -149,19 +157,22 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
...
-			skipDefaultRoutes:  response.PodIPInfo[i].SkipDefaultRoutes,
-			routes:             response.PodIPInfo[i].Routes,
-			pnpID:              response.PodIPInfo[i].PnPID,
-			endpointPolicies:   response.PodIPInfo[i].EndpointPolicies,
+			podIPAddress:         response.PodIPInfo[i].PodIPConfig.IPAddress,
+			ncSubnetPrefix:       response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.IPSubnet.PrefixLength,
+			ncPrimaryIP:          response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.IPSubnet.IPAddress,
+			ncGatewayIPAddress:   response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.GatewayIPAddress,
+			ncSubnetV6Prefix:     response.PodIPInfo[i].NetworkContainerIPv6Config.IPSubnet.PrefixLength,
+			ncIPv6:               response.PodIPInfo[i].NetworkContainerIPv6Config.IPSubnet.IPAddress,
```
estebancams, line 166: Can you rename to ncSubnetPrefixIPv6?

cni/network/invoker_cns.go
```diff
@@ -510,6 +521,42 @@ func configureSecondaryAddResult(info *IPResultInfo, addResult *IPAMAddResult, p
...
+		}
+		if ipv6Gateway.To4() != nil {
+			return errors.Wrapf(errInvalidGatewayIPv6, "expected IPv6 gateway, got IPv4: %s", info.ncGatewayIPv6Address)
+		}
+
+		ifInfo := addResult.interfaceInfo[key]
+		ifInfo.IPConfigs = append(ifInfo.IPConfigs, &network.IPConfig{
+			Address: net.IPNet{
+				IP:   ipv6,
+				Mask: net.CIDRMask(int(info.ncSubnetV6Prefix), 128),
```
estebancams: Can you use a constant for 128?

cni/network/invoker_cns.go
```diff
@@ -510,6 +521,42 @@ func configureSecondaryAddResult(info *IPResultInfo, addResult *IPAMAddResult, p
...
+		ipv6 := net.ParseIP(info.ncIPv6)
+		if ipv6 == nil {
+			logger.Error("Invalid IPv6 address from NetworkContainerIPv6Config",
+				zap.String("ncIPv6", info.ncIPv6),
+				zap.String("macAddress", info.macAddress))
+			return errors.Wrap(errInvalidIPv6Address, info.ncIPv6)
+		}
+		if ipv6.To4() != nil {
+			return errors.Wrapf(errInvalidIPv6Address, "expected IPv6, got IPv4: %s", info.ncIPv6)
+		}
```
rbtr: can we use netip here? it's nicer in that `netip.ParseAddr()` returns an error if the netip.Addr is invalid, and the result can be interogated directly as `addr.IsV4()/IsV6()`.
netip.Prefix is similarly better than the old net.IPNet. 
you will have to convert them at the existing types, but everything in between gets improved.

adjordjevic-microsoft: changed to use netip

PR #4276, opened 2026-03-18
fix: restoreState should attempt endpoint state restoration even if CNS state read fails

jackieluc:
**Reason for Change**:

- Fixed an issue where `restoreState` returned early on CNS state read failure, skipping endpoint state restoration entirely when CNS is configured to `ManageEndpointState`
- Added nil guard on `EndpointStateStore` to prevent potential panic when the config is enabled but the store is unset

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/restserver/util.go
+++ b/cns/restserver/util.go
@@ -122,7 +122,8 @@ func (service *HTTPRestService) restoreState() {
 
 	if service.Options[acn.OptManageEndpointState] == true {
 		if service.EndpointStateStore == nil {
-			logger.Errorf("[Azure CNS]  OptManageEndpointState is enabled but EndpointStateStore is not initialized; endpoint state persistence/restoration is disabled.") //nolint:staticcheck // TODO: migrate to zap
+			//nolint:staticcheck // TODO: migrate to zap
+			logger.Errorf("[Azure CNS]  OptManageEndpointState is enabled but EndpointStateStore is not initialized; endpoint state persistence/restoration is disabled.")
 			return
 		}
 		err := service.EndpointStateStore.Read(EndpointStoreKey, &service.EndpointState)


cns/restserver/util.go
```diff
@@ -116,13 +116,11 @@ func (service *HTTPRestService) restoreState() {
 			logger.Errorf("[Azure CNS]  Failed to restore state, err:%v. Removing azure-cns.json", err)
 			service.store.Remove()
 		}
-
-		return
+	} else {
+		logger.Printf("[Azure CNS]  Restored state, %+v\n", service.state)
```
behzad-mir: Why the return at line 120  has been removed here? So if the error !=nil we do not return from funtion?

cns/restserver/util.go
```diff
@@ -116,13 +116,11 @@ func (service *HTTPRestService) restoreState() {
...
 	}
 
-	logger.Printf("[Azure CNS]  Restored state, %+v\n", service.state)
-
-	if service.Options[acn.OptManageEndpointState] == true {
+	if service.Options[acn.OptManageEndpointState] == true && service.EndpointStateStore != nil {
```
behzad-mir: Why we did not see this issue in AKS but have been seen in ACI?

cns/restserver/util.go
jackieluc: the two state files are unrelated and shouldn't be coupled. this came up when i was trying to contrive an example where there was a leak in HNS and no NC is available

cns/restserver/util.go
jackieluc: didn't see an issue but functions managing endpoint state guard for it and it isn't guarded here. don't mind if you think we don't need it and let it panic if it isn't initialized

jackieluc: removed it; added log message if it's not initialized

PR #4281, opened 2026-03-19
[POND] Add Windows default gateway v6 to cns.

csfmomo:
**Reason for Change**:
We need Windows default gateway v6 to make dual stack work.

**Issue Fixed**:
Fix Windows lack default gateway v6 info.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/conversion_test.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion_test.go
@@ -258,7 +258,7 @@ func TestCreateNCRequestFromDynamicNC(t *testing.T) {
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := CreateNCRequestFromStaticNC(tt.input, false, 0)
+			got, err := CreateNCRequestFromDynamicNC(tt.input)
 			if tt.wantErr {
 				assert.Error(t, err)
 				return
@@ -437,7 +437,7 @@ func TestCreateNCRequestFromStaticNC(t *testing.T) {
 				assert.Error(t, err)
 				return
 			}
-			assert.NoError(t, err)
+			require.NoError(t, err)
 			if tt.want != nil {
 				assert.Equal(t, tt.want, got)
 			}


cns/kubecontroller/nodenetworkconfig/conversion_windows.go
```diff
@@ -74,8 +74,9 @@ func createNCRequestFromStaticNCHelper(
 		NetworkContainerType: cns.Docker,
 		Version:              strconv.FormatInt(nc.Version, 10), //nolint:gomnd // it's decimal
 		IPConfiguration: cns.IPConfiguration{
-			IPSubnet:         subnet,
-			GatewayIPAddress: nc.DefaultGateway,
+			IPSubnet:           subnet,
+			GatewayIPAddress:   nc.DefaultGateway,
+			GatewayIPv6Address: nc.DefaultGatewayV6,
 		},
```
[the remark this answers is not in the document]
NihaNallappagari: @csfmomo can you add unit test coverage?

cns/kubecontroller/nodenetworkconfig/conversion_windows.go
```diff
@@ -74,8 +74,9 @@ func createNCRequestFromStaticNCHelper(
 		NetworkContainerType: cns.Docker,
 		Version:              strconv.FormatInt(nc.Version, 10), //nolint:gomnd // it's decimal
 		IPConfiguration: cns.IPConfiguration{
-			IPSubnet:         subnet,
-			GatewayIPAddress: nc.DefaultGateway,
+			IPSubnet:           subnet,
+			GatewayIPAddress:   nc.DefaultGateway,
+			GatewayIPv6Address: nc.DefaultGatewayV6,
```
ecigar13, line 80: Should we add test for this field? Probably just need to add the field to an existing test.

csfmomo: Added.

cns/kubecontroller/nodenetworkconfig/conversion_test.go
```diff
@@ -335,6 +335,19 @@ func TestCreateNCRequestFromStaticNC(t *testing.T) {
...
+		{
+			name: "GatewayIPv6Address is set from DefaultGatewayV6",
+			input: v1alpha.NetworkContainer{
+				AssignmentMode:     v1alpha.Static,
+				Type:               v1alpha.Overlay,
+				PrimaryIP:          overlayPrimaryIP,
+				NodeIP:             nodeIP,
+				SubnetName:         subnetName,
+				SubnetAddressSpace: subnetAddressSpace,
+				DefaultGatewayV6:   "fd00::1",
```
nairashu: **🟡 Test doesn't exercise the Windows code path in Linux CI**

`conversion_test.go` has no OS build tags, and `createNCRequestFromStaticNCHelper` is platform-specific (`_linux.go` / `_windows.go`). On Linux CI, this new test case calls the **Linux** helper — which already sets `GatewayIPv6Address`. So the actual Windows fix isn't exercised unless tests run on a Windows target.

This is a pre-existing architectural limitation (not introduced by this PR), but worth being aware of — this test passing in CI doesn't prove the Windows change works.

cns/kubecontroller/nodenetworkconfig/conversion_test.go
```diff
@@ -345,7 +358,12 @@ func TestCreateNCRequestFromStaticNC(t *testing.T) {
 				return
 			}
 			assert.NoError(t, err)
-			assert.EqualValues(t, tt.want, got)
+			if tt.want != nil {
```
nairashu, line 441: **🟡 New test case skips full structural assertion**

Since the new test case doesn't set `want`, this `if tt.want != nil` guard means we only check error absence and the `GatewayIPv6Address` field. Other fields in the response could regress silently.

Consider populating a full expected response (via a platform-specific `want` value in `conversion_windows_test.go` / `conversion_linux_test.go`, similar to how `validOverlayRequest` is defined per-platform), or at minimum add `assert.NotNil(t, got)` to guard against nil dereference on the `got.IPConfiguration` access below.

cns/kubecontroller/nodenetworkconfig/conversion_test.go
```diff
@@ -345,7 +358,12 @@ func TestCreateNCRequestFromStaticNC(t *testing.T) {
...
+				assert.Equal(t, tt.want, got)
+			}
+			if tt.input.DefaultGatewayV6 != "" {
```
nairashu: **🟢 Nit: Conditional assertion in shared test loop**

This `DefaultGatewayV6` check runs for every test case in the loop. It's harmless (silently skipped for existing cases with empty `DefaultGatewayV6`), but adds implicit behavior future authors might not expect.

A cleaner approach: include `GatewayIPv6Address: "fd00::1"` in the `want` struct of the new test case and rely on the existing full-struct comparison. That keeps the assertion path uniform across all table entries.

cns/kubecontroller/nodenetworkconfig/conversion_test.go
csfmomo: Updated the test to Windows.

cns/kubecontroller/nodenetworkconfig/conversion_test.go
csfmomo: Remove the unit test change to keep pr concise which only touch the windows part.

csfmomo:
azp run

nairashu (approved):

NihaNallappagari (approved):

PR #4282, opened 2026-03-19
fix: controller-runtime manager can't be restarted

rbtr:
**Reason for Change**:
There is no scenario in which a controller-runtime manager can be restarted. If the context is cancelled, it may exit without error; if it errors during Start, it will return that error - either scenario is unrecoverable and the Manager must be rebuilt.

This [snippet of the Start method](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/internal.go#L347) makes this very clear:

```go
// Start starts the manager and waits indefinitely.
// There is only two ways to have start return:
// An error has occurred during in one of the internal operations,
// such as leader election, cache start, webhooks, and so on.
// Or, the context is cancelled.
func (cm *controllerManager) Start(ctx context.Context) (err error) {
	cm.Lock()
	if cm.started {
		cm.Unlock()
		return errors.New("manager already started")
	}
	cm.started = true
...
...
...
}
```

It's clear that Start _cannot be called more than once_. That will never work. 

We could try to respring main, rebuild the manager, and Start the new one. But it's probably safer and definitely more reliable that we simply exit and let the Pod be restarted.

**Notes**:
I don't know if this never could have worked or if ctrlruntime has evolved...but it definitely doesn't work today and we need this fix to keep CNS from deadlocking during Manager init.

### DIFF
--- a/cns/service/main.go
+++ b/cns/service/main.go
@@ -1583,20 +1583,18 @@ func InitializeCRDState(ctx context.Context, z *zap.Logger, httpRestService cns.
 	// Start the Manager which starts the reconcile loop.
 	// The Reconciler will send an initial NodeNetworkConfig update to the PoolMonitor, starting the
 	// Monitor's internal loop.
+	// Note: controller-runtime Managers are not restartable. The *first* thing that Start() does is
+	// set the started semaphore, and *any* restart attempt will fail.
+	// The Manager is permanently marked as started and subsequent calls will always fail with
+	// "manager already started". If the Manager returns AT ALL, we must exit the process and
+	// let the kubelet restart the pod (or rebuild world, but exiting is more reliable).
 	go func() {
 		logger.Printf("Starting controller-manager.")
-		for {
-			if err := manager.Start(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


cns/service/main.go
```diff
@@ -1583,20 +1583,18 @@ func InitializeCRDState(ctx context.Context, z *zap.Logger, httpRestService cns.
 	// Start the Manager which starts the reconcile loop.
 	// The Reconciler will send an initial NodeNetworkConfig update to the PoolMonitor, starting the
 	// Monitor's internal loop.
+	// Note: controller-runtime Managers are not restartable. The *first* thing that Start() does is
+	// set the started semaphore, and *any* restart attempt will fail.
+	// The Manager is permanently marked as started and subsequent calls will always fail with
+	// "manager already started". If the Manager returns AT ALL, we must exit the process and
```
QxBytes, line 1589: if the manager start returns but without an error do we still exit the process? it looks like we don't but the comment seems to imply that we must

rbtr: Not directly, the only case that it exits without an error is if the context closes - which means the process is already exiting.

PR #4286, opened 2026-03-24
[feat]Implement the required CNI changes to support SwiftV1 pod subnet ipv6 on Windows

NihaNallappagari:
**Reason for Change**:
Implement the required CNI changes to support SwiftV1 pod subnetv6 on Windows

In the SwiftV1 prefix-on-NIC IPv6 scenario, only one NC contains both IPv4 and IPv6 details. The HNS network was incorrectly created for IPv6 because the destination prefix, prefix length, and gateway values from IPv4 were being used instead. The code has been updated to pass the correct IPv6 values during HNS network creation, and falls back to the previous behavior if those values are not available.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

**Payload sent to HNS network/Endpoint create before change:** 
`,{"IpAddressPrefix":"fd00::/16","Routes":[{"NextHop":"10.241.0.1","DestinationPrefix":"0.0.0.0/0"}]}`

**After changes:**
```
{"Subnets":[{"IpAddressPrefix":"fd00:aec6:6946:1::/64","Routes":[{"NextHop":"fd00:aec6:6946:1::1","DestinationPrefix":":
:/0"}]}]}]
```

pod to pod connectivity after the change 
pod to remote pod
[connectivity testing doc](https://microsoft.sharepoint.com/:w:/t/Aznet/cQohv4uJltPVRZJ1AnWfDa4ZEgUCEP5DQjXBQ1Swy-jpdgJV3w?isSPOFile=1&xsdata=MDV8MDJ8fDMzZjcyNGVlOWM5ZDQ5OTAzOWUxMDhkZThjNDBkNjIzfDcyZjk4OGJmODZmMTQxYWY5MWFiMmQ3Y2QwMTFkYjQ3fDB8MHw2MzkxMDI0MDcyMzE2MTc1MDV8VW5rbm93bnxWR1ZoYlhOVFpXTjFjbWwwZVZObGNuWnBZMlY4ZXlKRFFTSTZJbFJsWVcxelgwRlVVRk5sY25acFkyVmZVMUJQVEU5R0lpd2lWaUk2SWpBdU1DNHdNREF3SWl3aVVDSTZJbGRwYmpNeUlpd2lRVTRpT2lKUGRHaGxjaUlzSWxkVUlqb3hNWDA9fDF8TDJOb1lYUnpMekU1T21Ga016WTBaREl6TWpsaE56UmpZVEk1WVRRNVpERmhNV1ppWmpBd1pUUmtRSFJvY21WaFpDNTJNaTl0WlhOellXZGxjeTh4TnpjME5qUXpPVEl4T1RBNXxmMTc1YTk0MTM0NzY0Y2Y3Nzk3MDA4ZGU4YzQwZDYyMnwzOWQ5YmI5YTVlNTM0ODRkYmVhYWU3NDViYjgzMzcyOQ%3D%3D&sdata=d0pPUGNGTFlhZzVlb0pYQmJlMU1IdVhrV1Q5eGNuNVVBcWEyRUFjekVYYz0%3D&ovuser=72f988bf-86f1-41af-91ab-2d7cd011db47%2Cnihan%40microsoft.com)

### DIFF
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -199,10 +199,10 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 
 	for i := 0; i < len(response.PodIPInfo); i++ {
 		info := IPResultInfo{
-			podIPAddress:   response.PodIPInfo[i].PodIPConfig.IPAddress,
-			ncSubnetPrefix: getIPConfigPrefixLength(response.PodIPInfo[i].PodIPConfig.IPAddress, response.PodIPInfo[i].NetworkContainerPrimaryIPConfig),
 			// ncPrimaryIP intentionally stays IPv4 — it is only used for SNAT/iptables which are IPv4-only codepaths.
 			ncPrimaryIP:          response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.IPSubnet.IPAddress,
+			podIPAddress:         response.PodIPInfo[i].PodIPConfig.IPAddress,
+			ncSubnetPrefix:       getIPConfigPrefixLength(response.PodIPInfo[i].PodIPConfig.IPAddress, response.PodIPInfo[i].NetworkContainerPrimaryIPConfig),
 			ncGatewayIPAddress:   getIPConfigGatewayAddress(response.PodIPInfo[i].PodIPConfig.IPAddress, response.PodIPInfo[i].NetworkContainerPrimaryIPConfig),
 			ncSubnetPrefixIPv6:   response.PodIPInfo[i].NetworkContainerIPv6Config.IPSubnet.PrefixLength,
 			ncIPv6:               response.PodIPInfo[i].NetworkContainerIPv6Config.IPSubnet.IPAddress,


cni/network/invoker_cns_test.go
```diff
@@ -735,11 +737,16 @@ func TestCNSIPAMInvoker_Add(t *testing.T) {
 									},
 									NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
 										IPSubnet: cns.IPSubnet{
+											IPAddress:    "fd11:1234::",
```
kadevu: IPSubnet should be v4 address

kadevu: Also change the PrefixLength

kadevu: Actually you should keep this untouched (which is covering the overlay scenario) and add another test scenario which is covering the PONv6 payload

cni/network/invoker_cns_test.go
```diff
@@ -1046,12 +1053,14 @@ func TestRequestIPAPIsFail(t *testing.T) {
 										PrefixLength: 112,
 									},
 									NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
+										IPSubnetV6: cns.IPSubnet{},
```
kadevu, line 1230: Revert changes to this test scenario and add another test scenario for PONv6

NihaNallappagari: I’m intentionally updating the overlay case here, Kaushik. I want to update scenario where, if IPSubnetV6 is empty, the existing overlay functionality remains unchanged

kadevu: But you have also updated GatewayIPAddress and GatewayIPv6Address. In case of dualstack overlay wouldnt GatewayIPAddress have the IPv6 address and GatewayIPv6Address is empty?

cni/network/invoker_cns.go
```diff
@@ -409,7 +427,7 @@ func configureDefaultAddResult(info *IPResultInfo, addConfig *IPAMAddConfig, add
 				return err
 			}
 		} else if net.ParseIP(info.podIPAddress).To16() != nil {
-			ncgw = net.ParseIP(overlayGatewayV6IP)
+			ncgw = net.ParseIP(defaultGatewayV6IP)
```
kadevu: Why did you modify this? This will still be hit in overlay cases

cni/network/invoker_cns_test.go
NihaNallappagari: That’s correct. For that reason, in the overlay test cases I passed _GatewayIPv6Address: ""_ and _IPSubnetV6: cns.IPSubnet{}._ I updated the other cases (TestCNSIPAMInvoker_Add and TestRequestIPAPIsFail) with values that reflect the pod subnet scenarios. I’ll revert the changes in those two test cases and add a separate test to keep things clear.”

cni/network/invoker_cns_test.go
NihaNallappagari: Done

cni/network/invoker_cns.go
NihaNallappagari: I updated this for my local testing, this is pushed accidentally. Reverted the change

cni/network/invoker_cns.go
```diff
@@ -59,6 +59,24 @@ type IPResultInfo struct {
 	endpointPolicies   []policy.Policy
 }
 
+func getIPConfigGatewayAddress(podIP string, ipConfig cns.IPConfiguration) string {
+	parsedPodIP := net.ParseIP(podIP)
+	if parsedPodIP != nil && parsedPodIP.To4() == nil && net.ParseIP(ipConfig.GatewayIPv6Address) != nil {
+		return ipConfig.GatewayIPv6Address
```
ZetaoZhuang: for both ipv4 and ipv6, we are populating and reading the same field ipConfig.GatewayIPAddress for standalone scenario,  https://msazure.visualstudio.com/One/_git/Networking-Aquarius?path=/src/dnc/datamodel/networkmanager.go&version=GBmaster&line=2989&lineEnd=2990&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents . 
your pr is for aks scenario. might be different. just want to confirm

NihaNallappagari: for overlay yes, that changes for podsubnet scenario. Existing scenarios are not affected as cns sends empty GatewayIPv6Address, but for podsubnetv6 scenario dnc publishes GatewayIPv6Address field to NNC and cni consumes it from cns

cni/network/invoker_cns.go
```diff
@@ -150,9 +168,9 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 	for i := 0; i < len(response.PodIPInfo); i++ {
 		info := IPResultInfo{
 			podIPAddress:       response.PodIPInfo[i].PodIPConfig.IPAddress,
-			ncSubnetPrefix:     response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.IPSubnet.PrefixLength,
+			ncSubnetPrefix:     getIPConfigPrefixLength(response.PodIPInfo[i].PodIPConfig.IPAddress, response.PodIPInfo[i].NetworkContainerPrimaryIPConfig),
 			ncPrimaryIP:        response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.IPSubnet.IPAddress,
-			ncGatewayIPAddress: response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.GatewayIPAddress,
```
ZetaoZhuang: never mind

cni/network/invoker_cns.go
ZetaoZhuang: I see. Thanks for the explanation. I miss the condition check here

cni/network/invoker_cns.go
```diff
@@ -59,6 +59,24 @@ type IPResultInfo struct {
...
+	parsedPodIP := net.ParseIP(podIP)
+	if parsedPodIP != nil && parsedPodIP.To4() == nil && net.ParseIP(ipConfig.GatewayIPv6Address) != nil {
+		return ipConfig.GatewayIPv6Address
+	}
+
+	return ipConfig.GatewayIPAddress
+}
+
+func getIPConfigPrefixLength(podIP string, ipConfig cns.IPConfiguration) uint8 {
+	parsedPodIP := net.ParseIP(podIP)
```
nairashu: **🟡 Inconsistent sentinel check between the two helpers**

`getIPConfigGatewayAddress` (line 63) validates the v6 value with `net.ParseIP(ipConfig.GatewayIPv6Address) != nil` — checking if it's a valid IP.

`getIPConfigPrefixLength` here checks `ipConfig.IPSubnetV6.PrefixLength != 0` — using 0 as a sentinel for "not set."

These two checks answer different questions: "is it valid?" vs "is it non-zero?" A `/0` prefix is theoretically valid (though practically never an NC subnet prefix). More importantly, you could have `IPSubnetV6.PrefixLength` set to a non-zero value while `IPSubnetV6.IPAddress` is empty (a half-configured state).

Consider using a consistent check, e.g.:
```go
if parsedPodIP != nil && parsedPodIP.To4() == nil && ipConfig.IPSubnetV6.IPAddress != "" {
```
This mirrors the "is it actually configured?" intent and would be more robust.

cni/network/invoker_cns.go
```diff
@@ -59,6 +59,24 @@ type IPResultInfo struct {
 	endpointPolicies   []policy.Policy
 }
 
+func getIPConfigGatewayAddress(podIP string, ipConfig cns.IPConfiguration) string {
+	parsedPodIP := net.ParseIP(podIP)
```
nairashu: **🟢 Nit: Silent fallback on malformed `GatewayIPv6Address`**

If `GatewayIPv6Address` is non-empty but not a valid IP (e.g., `"bad-addr"`), `net.ParseIP` returns nil and the function silently falls back to the v4 gateway. This is safe but could mask a configuration error. Consider logging a warning when `GatewayIPv6Address != ""` but fails to parse, to aid debugging in production.

cni/network/invoker_cns_test.go
```diff
@@ -866,6 +868,125 @@ func TestCNSIPAMInvoker_Add(t *testing.T) {
 	}
 }
```
nairashu, line 869: **🟢 Nit: Missing IPv4 regression test case**

All three test cases use an IPv6 pod IP (`fd11:1234::1`). Consider adding a test case with an **IPv4 pod IP** where v6 fields are populated, to verify the helpers correctly ignore the v6 fields and use v4 values. This guards against accidentally returning v6 data for v4 pods.

cni/network/invoker_cns.go
NihaNallappagari: unfortunately, IPAddress is returned empty _ipSubnetV6":"{IPAddress: PrefixLength:64}_
We doesn't need ip address here, cni uses pod ip address and ip address here is redundant. Would be prefix ever being non zero other than the default value?

cni/network/invoker_cns.go
NihaNallappagari: Added a warning. Thanks

cni/network/invoker_cns_test.go
```diff
@@ -1046,6 +1206,7 @@ func TestRequestIPAPIsFail(t *testing.T) {
 										PrefixLength: 112,
 									},
 									NetworkContainerPrimaryIPConfig: cns.IPConfiguration{
+										IPSubnetV6: cns.IPSubnet{},
```
nairashu, line 1230: Re Kaushik's earlier comment about reverting this: the `IPSubnetV6: cns.IPSubnet{}` and `GatewayIPv6Address: ""` additions here are explicitly documenting that overlay sends empty v6 fields. This makes the test data more complete and is a valid change — not a functional modification. I'd resolve this thread.

cni/network/invoker_cns_test.go
nairashu: Addressed

network/network_windows.go
```diff
@@ -289,13 +289,17 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *EndpointInfo, extIf *exter
 
 	// Populate subnets.
 	for _, subnet := range nwInfo.Subnets {
+		defaultIpRouteCIDR := defaultRouteCIDR
+		if subnet.Family == platform.AfINET6 {
+			defaultIpRouteCIDR = defaultIPv6Route
+		}
```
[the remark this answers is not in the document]
nairashu: Addresed it looks like. Resolving the thread.

nairashu (commented):
## Code Review

### 🔴 Bug: Test subtests discard inner `*testing.T`

In `TestCNSIPAMInvoker_Add_podsubnetv6`, the subtest closure uses `func(_ *testing.T)` which discards the inner `t`. The `req` is bound to the outer `t` via `require.New(t)` at the top. This means:

- `require` failures **abort all remaining subtests** (the outer `t` is `FailNow`'d)
- Failures are **not correctly attributed** to the specific subtest in test output

```go
// Current (buggy):
t.Run(tt.name, func(_ *testing.T) {  // inner t discarded

// Should be:
t.Run(tt.name, func(t *testing.T) {
    req := require.New(t)  // bind to subtest's t
```

### 🟡 `ncPrimaryIP` not updated for IPv6

The PR makes `ncSubnetPrefix` and `ncGatewayIPAddress` IPv6-aware but leaves `ncPrimaryIP` always populated from `IPSubnet.IPAddress`. Downstream, `ncPrimaryIP` feeds into IPv4-guarded contexts (SNAT, iptables) so the CIDR construction itself is fine — it uses `podIPAddress`, not `ncPrimaryIP`.

However, the asymmetry is concerning: if `IPSubnet.IPAddress` is an IPv4 address while the pod has an IPv6 address, `ncPrimaryIP` would be incorrect for any future IPv6 usage. At minimum, add a comment explaining why `ncPrimaryIP` intentionally stays IPv4-only, or add a parallel `getIPConfigPrimaryIP` helper for consistency.

### 🟡 Missing standalone unit tests for helper functions

`getIPConfigGatewayAddress` and `getIPConfigPrefixLength` are pure functions — ideal for lightweight, isolated unit tests. The PR only tests them via the full `Add()` integration flow. Direct unit tests would be faster, simpler to maintain, and more precisely document edge-case behavior.

### 🟢 Minor: Inconsistent logging in fallback paths

`getIPConfigGatewayAddress` logs a warning when the IPv6 gateway is present but invalid. `getIPConfigPrefixLength` silently falls back when `IPSubnetV6.PrefixLength` is 0. Consider adding a debug/warn log in the prefix function too for operational visibility.

cni/network/invoker_cns.go
```diff
@@ -65,6 +65,31 @@ type IPResultInfo struct {
 	endpointPolicies     []policy.Policy
 }
 
+func getIPConfigGatewayAddress(podIP string, ipConfig cns.IPConfiguration) string {
+	parsedPodIP := net.ParseIP(podIP)
+	if parsedPodIP != nil && parsedPodIP.To4() == nil && ipConfig.GatewayIPv6Address != "" {
```
ecigar13: nit: you can use netip package `netip.ParseAddr`. It'll return an error on empty and invalid IP.

NihaNallappagari: done

NihaNallappagari:
> `ncPrimaryIP` not updated for IPv6

Fixed the test case
getIPConfigGatewayAddress and getIPConfigPrefixLength are pure functions  are private functions and I covered all the possible scenarios of these 2 functions in Add test which is calling these functions. so I haven't added separate tests. 
Added missing logs

cni/network/invoker_cns.go
```diff
@@ -65,6 +65,45 @@ type IPResultInfo struct {
 	endpointPolicies     []policy.Policy
 }
 
+func getIPConfigGatewayAddress(podIP string, ipConfig cns.IPConfiguration) string {
+	parsedPodIP, err := netip.ParseAddr(podIP)
+	if err != nil {
+		logger.Warn("Invalid pod IP address", zap.String("podIP", podIP))
+		return ipConfig.GatewayIPAddress
+	}
+	if parsedPodIP.IsValid() && parsedPodIP.Is6() && ipConfig.GatewayIPv6Address != "" {
```
nairashu: Nit: `parsedPodIP.IsValid()` is redundant here. `netip.ParseAddr` only returns a nil error when the result is a valid address, so after the `err != nil` guard above this is always true. Same applies to the identical check in `getIPConfigPrefixLength` (line 95). Safe to simplify to just `parsedPodIP.Is6()`.

cni/network/invoker_cns.go
```diff
@@ -65,6 +65,45 @@ type IPResultInfo struct {
...
+		if net.ParseIP(ipConfig.GatewayIPv6Address) != nil {
```
nairashu: Nit: This uses `net.ParseIP` to validate the gateway while the pod IP is validated with `netip.ParseAddr` a few lines above. Consider using `netip.ParseAddr` here too for consistency — `net.ParseIP` accepts some unusual formats (e.g. leading zeros) that `netip.ParseAddr` rejects.

cni/network/invoker_cns.go
```diff
@@ -159,9 +198,9 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 	for i := 0; i < len(response.PodIPInfo); i++ {
 		info := IPResultInfo{
 			podIPAddress:         response.PodIPInfo[i].PodIPConfig.IPAddress,
-			ncSubnetPrefix:       response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.IPSubnet.PrefixLength,
+			ncSubnetPrefix:       getIPConfigPrefixLength(response.PodIPInfo[i].PodIPConfig.IPAddress, response.PodIPInfo[i].NetworkContainerPrimaryIPConfig),
 			ncPrimaryIP:          response.PodIPInfo[i].NetworkContainerPrimaryIPConfig.IPSubnet.IPAddress,
```
nairashu, line 203: `ncPrimaryIP` is still always populated from the IPv4 `IPSubnet.IPAddress`, while its sibling fields (`ncSubnetPrefix`, `ncGatewayIPAddress`) are now IPv6-aware. Downstream this is safe today — `ncPrimaryIP` is only used in IPv4-guarded contexts (SNAT, iptables). But the asymmetry could bite a future reader or contributor. Would be good to at least add a comment here like:
```go
// ncPrimaryIP intentionally stays IPv4 — it is only used for SNAT/iptables which are IPv4-only codepaths.
```
Or add a `getIPConfigPrimaryIP` helper for consistency.

cni/network/invoker_cns.go
NihaNallappagari: done

cni/network/invoker_cns.go
NihaNallappagari: ya I missed it, updated now

cni/network/invoker_cns.go
NihaNallappagari: done

cni/network/invoker_cns.go
```diff
@@ -158,10 +199,11 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 
 	for i := 0; i < len(response.PodIPInfo); i++ {
 		info := IPResultInfo{
```
nairashu, line 201: Very minor nit: `podIPAddress` and `ncSubnetPrefix` have different alignment padding from the fields below (`ncPrimaryIP`, `ncGatewayIPAddress`, etc.). A quick `gofumpt` pass should normalize this.

nairashu (dismissed):
LGTM — all previous feedback has been addressed. Clean helper functions with consistent use of `netip.ParseAddr`, good fallback logging, and solid test coverage across the IPv4/IPv6 matrix.

One very minor nit: the struct literal alignment for `podIPAddress` and `ncSubnetPrefix` is inconsistent with the other fields (shorter padding). The linter will likely catch this, but worth a quick `gofumpt` pass.

ecigar13 (approved):

PR #4309, opened 2026-03-31
test: add swiftv2 long-running cluster tests

isaac-dasan:
## Overview
This PR introduces comprehensive long-running test infrastructure for SwiftV2 pod networking, enabling continuous validation of pod allocation, deallocation, and connectivity in eastus2euap region. Tests run automatically every hour on persistent test clusters.

## Key Changes

**Pipeline & Infrastructure (11 files)**
- New hourly scheduled pipeline with configurable zone-aware deployments
- Infrastructure setup template and stage for resource provisioning
- Zone node pool provisioning script with support for multiple regions
- Lease management scripts to prevent concurrent pipeline conflicts
- VNet/subnet validation and verification scripts
- Enhanced datapath test stage with improved error handling

**Long-Running Zone-Aware Pod Tests (3 new test files)**
- Rotating Pods: 6 deployments per zone continuously recycled to simulate pod churn
- DaemonSet Always-On Pods: Self-healing pods per zone for live migration testing
- Connectivity Tests: Validate pod-to-pod communication across zones

**Test Helpers & Utilities (2 files)**
- New Kubernetes client library with pod CRUD operations
- Enhanced Azure CLI helpers for MTPNC cleanup tracking and lease management

**Documentation & Manifests (5 files)**
- Comprehensive guide covering test design, zone allocation, and rotation algorithms
- Kubernetes manifests for DaemonSet and Deployment deployments

## Reviewers Should Know
- Tests has separate subnet and reuses PodNetwork/PodNetworkInstance infrastructure in everyr un
- Lease mechanism prevents simultaneous pipeline on swiftv2-datapath test case runs from interfering
- Zone-aware node pool creation is configurable per region

### DIFF
--- a/.pipelines/swiftv2-long-running/scripts/ensure_zone_nodepools.sh
+++ b/.pipelines/swiftv2-long-running/scripts/ensure_zone_nodepools.sh
@@ -163,8 +163,6 @@ for ZONE in $ZONES; do
 
   echo "==> Labeling and tainting nodes in pool $POOL_NAME"
   kubectl --kubeconfig "$KUBECONFIG_FILE" label nodes -l agentpool=$POOL_NAME \
-    nic-capacity=high-nic \
-    workload-type=swiftv2-linux \
     longrunning-zone-pool=true \
     --overwrite
 


test/integration/swiftv2/longRunningCluster/datapath_longrunning_alwayson_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+)
+
+func TestLongRunningAlwaysOn(t *testing.T) {
+	gomega.RegisterFailHandler(ginkgo.Fail)
```
miguelgoms, line 18: I don't think we are using ginkgo/gomega elsewhere - is this pipeline the only place with this dependecy? is there a reason this is using ginkgov1?

test/integration/swiftv2/longRunningCluster/k8s_client.go
```diff
@@ -0,0 +1,511 @@
+//go:build longrunning_rotating_test || longrunning_alwayson_test || longrunning_connectivity_test
+
+// Package-level nolint for unused: functions in this file are shared across
+// multiple build tags (longrunning_rotating_test, longrunning_alwayson_test, longrunning_connectivity_test).
+// Each tag only uses a subset, so the unused linter flags cross-tag symbols.
+//
+//nolint:unused // cross-build-tag shared helpers
+package longrunningcluster
```
miguelgoms: nit: naming and place - this is not tightly coupled to the longrunning cluster but also not a generic 'k8s client'. this seems analogous to the az_helpers.go but for k8 goclient insteand of CLIs. 
move this to the helpers folder, consider renaming.

test/integration/swiftv2/longRunningCluster/datapath_longrunning_alwayson_test.go
isaac-dasan: - There are other pipielines (npm, cni) that are already ginkgo in this repo
- The tests in this pipeline are not complicated enough to migrate to ginkgo/v2. I tried migrating only to the new files, but it is complicating some go.mod and linting dependencies, so I backed off

isaac-dasan: If you are wondering when to use ginkgo and when to use azure/go-workflow, here is an AI summary:
- Ginkgo is for expressing specs, assertions, failure reporting, suite lifecycle, and test runner behavior.
- go-workflow is for expressing execution graphs like “create namespace, then create PNI, then create deployment, then run connectivity checks,” especially when some steps can run in parallel.

miguelgoms: yes - we cannot use a mix of v1/v2 within the same package - https://[onsi.github.io/ginkgo/MIGRATING_TO_V2#can-i-mix-ginkgo-v1-and-ginkgo-v2](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#can-i-mix-ginkgo-v1-and-ginkgo-v2)

if the pipeline is already using v1, we don't need to include migrating onto the same pullrequest, we can tackle and review that as a separate change.

miguelgoms (approved):

PR #4314, opened 2026-04-02
feat: add cilium log collector support for multiple app insight outputs

QxBytes:
**Reason for Change**:
Currently, the cilium log collector plugin has one global variable for its app insights tracker, for example. The problem is if there are multiple instances of the plugin, they overwrite each other. They all share the same globals, so we add context to identify which instance of the plugin is being run. See https://github.com/fluent/fluent-bit-go/blob/master/examples/out_multiinstance/out.go for more information.

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
Tested using two outputs and it populates both app insights resources with their respective logs. Just for testing we send cns logs to one app insights, and cilium logs to another, and each app insights resource shows the logs of the application we specified only.

See pipeline page for full green run.

### DIFF
--- a/cilium-log-collector/out_azure_app_insights.go
+++ b/cilium-log-collector/out_azure_app_insights.go
@@ -111,37 +111,35 @@ func FLBPluginRegister(def unsafe.Pointer) int {
 	return output.FLBPluginRegister(def, "azure_app_insights", "Azure application insights")
 }
 
-// (fluentbit will call this)
-// plugin (context) pointer to fluentbit context (state/ c code)
-//
-//export FLBPluginInit
-func FLBPluginInit(plugin unsafe.Pointer) int {
-	fmt.Printf("[flb-azure-app-insights] version = '%s'\n", version)
+type configLookup func(key string) string
 
+type fileChecker func(path string) bool
+
+// initPluginContext builds a pluginContext from config values without depending on unsafe.Pointer
+func initPluginContext(lookup configLookup, fileExists fileChecker) (*pluginContext, int) {
 	ctx := &pluginContext{}
 
-	ctx.id = output.FLBPluginConfigKey(plugin, "id")
+	ctx.id = lookup("id")
 
 	// check disable flag
-	if _, err := os.Stat(disableFilePath); err == nil {
+	if fileExists(disableFilePath) {
 		fmt.Printf("[flb-azure-app-insights] Plugin disabled- file found at: %s\n", disableFilePath)
 		ctx.disabled = true
-		output.FLBPluginSetContext(plugin, ctx)
-		return output.FLB_OK
+		return ctx, output.FLB_OK
 	}
 	ctx.disabled = false
 
-	ctx.instrumentationKey = output.FLBPluginConfigKey(plugin, "instrumentation_key")
+	ctx.instrumentationKey = lookup("instrumentation_key")
 	if ctx.id == "" {
 		ctx.id = ctx.instrumentationKey
 	}
 	// the key that is identified as the log upon receiving the record in this plugin
-	ctx.logKey = output.FLBPluginConfigKey(plugin, "log_key")
+	ctx.logKey = lookup("log_key")
 	if ctx.logKey == "" {
 		ctx.logKey = "log"
 	}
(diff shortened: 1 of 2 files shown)


cilium-log-collector/out_azure_app_insights.go
```diff
@@ -151,24 +166,39 @@ func FLBPluginInit(plugin unsafe.Pointer) int {
...
 	fmt.Printf("[flb-azure-app-insights] App Insights client initialized with key: %s\n",
 		telemetryConfig.InstrumentationKey)
+	output.FLBPluginSetContext(plugin, ctx)
 	return output.FLB_OK
 }
 
 //export FLBPluginFlush
 func FLBPluginFlush(data unsafe.Pointer, length C.int, tag *C.char) int {
+	fmt.Println("[flb-azure-app-insights] Flush called for unknown instance")
+	return output.FLB_OK
```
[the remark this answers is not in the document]
QxBytes: If there's no context it wouldn't know where to send the logs anyway

cilium-log-collector/out_azure_app_insights.go
```diff
@@ -263,10 +293,21 @@ func (rp *RecordProcessor) ProcessSingleRecord(record ProcessRecord, recordIndex
 
 //export FLBPluginExit
 func FLBPluginExit() int {
-	if client != nil {
-		client.Channel().Flush()
+	fmt.Println("[flb-azure-app-insights] Exit called for unknown instance")
+	return output.FLB_OK
```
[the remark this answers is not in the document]
QxBytes: Follows example on fluent bit go's page

hack/aks/deploy.mk
```diff
@@ -70,6 +70,7 @@ add-cilium-log-collector:
 	@echo "CILIUM_LOG_COLLECTOR_IMAGE_REGISTRY: $(CILIUM_LOG_COLLECTOR_IMAGE_REGISTRY)"
 	kubectl apply --server-side -f ../../test/integration/manifests/cilium/v$(DIR)/cilium-log-collector/cilium-log-collector-configmap.yaml
 	envsubst '$${CILIUM_LOG_COLLECTOR_VERSION_TAG},$${CILIUM_LOG_COLLECTOR_IMAGE_REGISTRY}' < ../../test/integration/manifests/cilium/v$(DIR)/cilium-log-collector/daemonset-patch.yaml | kubectl apply --server-side --field-manager=cilium-log-collector -f -
+	kubectl rollout restart ds cilium -n kube-system
```
jpayne3506, line 73: Is this not duplicating effort..? PATCH/UPDATE restarts the daemonset, right? Expecting a SSA difference here.

QxBytes: During testing I often ran `add-cilium-log-collector` after just updating the log collector configmap. If only the configmap changes the cilium daemonset does not restart.

jpayne3506: Ahh, so this is a catch all... I don't know of a good pattern off the top of my head to ensure we always restart agent after cm changes without clubbing everything...

cilium-log-collector/out_azure_app_insights.go
```diff
@@ -104,42 +111,50 @@ func FLBPluginRegister(def unsafe.Pointer) int {
...
+
+// initPluginContext builds a pluginContext from config values without depending on unsafe.Pointer
+func initPluginContext(lookup configLookup, fileExists fileChecker) (*pluginContext, int) {
+	ctx := &pluginContext{}
+
+	ctx.id = lookup("id")
+
 	// check disable flag
-	if _, err := os.Stat(disableFilePath); err == nil {
+	if fileExists(disableFilePath) {
```
santhoshmprabhu, line 125: We can selectively disable fluent on some pods by configuring the mounts appropriately for the disable path, correct?

QxBytes: yes this is in the existing log collector and this pr should not change that behavior. Currently if you add the disable configmap it should disable log collection on all pods but that's because the same configmap path is mounted to all log collector pods.

PR #4316, opened 2026-04-02
[POND] Remove the ipv6 check when setting ipset v6 address

csfmomo:
**Reason for Change**:

We don't need the ipv6 address empty check since CNI doesn't need v6 address yet, it only needs gateway and ip subnet. Adding ip subnet v6 address is just for keeping consistent with v4 and future extension.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cns/kubecontroller/nodenetworkconfig/conversion.go
+++ b/cns/kubecontroller/nodenetworkconfig/conversion.go
@@ -35,10 +35,6 @@ func parseIPv6Subnet(nc v1alpha.NetworkContainer) (cns.IPSubnet, error) {
 	if !subnetV6Prefix.Addr().Is6() {
 		return cns.IPSubnet{}, errors.Errorf("SubnetAddressSpaceV6 %s is not an IPv6 prefix", nc.SubnetAddressSpaceV6)
 	}
-	// remove the condition to allow that v6 prefix set but IPs not ready.
-	// if nc.PrimaryIPV6 == "" {
-	// 	return cns.IPSubnet{}, errors.New("PrimaryIPV6 must be set when SubnetAddressSpaceV6 is specified")
-	// }
 	return cns.IPSubnet{
 		IPAddress:    nc.PrimaryIPV6,
 		PrefixLength: uint8(subnetV6Prefix.Bits()), //#nosec G115 -- prefix bits are 0-128, fits uint8


cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -35,9 +35,6 @@ func parseIPv6Subnet(nc v1alpha.NetworkContainer) (cns.IPSubnet, error) {
 	if !subnetV6Prefix.Addr().Is6() {
 		return cns.IPSubnet{}, errors.Errorf("SubnetAddressSpaceV6 %s is not an IPv6 prefix", nc.SubnetAddressSpaceV6)
 	}
-	if nc.PrimaryIPV6 == "" {
-		return cns.IPSubnet{}, errors.New("PrimaryIPV6 must be set when SubnetAddressSpaceV6 is specified")
-	}
```
rejain789, line 40: summary - getting rid of this comment as NNC does not have the field **primaryIPV6** and this field will be empty and is not getting used in cni for programming.

cns/kubecontroller/nodenetworkconfig/conversion.go
```diff
@@ -35,9 +35,6 @@ func parseIPv6Subnet(nc v1alpha.NetworkContainer) (cns.IPSubnet, error) {
...
 	return cns.IPSubnet{
 		IPAddress:    nc.PrimaryIPV6,
```
rejain789, line 39: can we get rid of this line: 	IPAddress:    nc.PrimaryIPV6 since we are always passing in empty here?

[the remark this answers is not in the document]
csfmomo: We don't need to do ip validation here.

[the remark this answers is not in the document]
NihaNallappagari: is it required to pass ip address here?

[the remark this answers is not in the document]
csfmomo: I don't think so since we are not using IP address to start a pod. Keep it here for future reference.

cns/kubecontroller/nodenetworkconfig/conversion.go
nairashu: Simply removing this will break the validation for Overlay use case. We need to update the logic such that this check is only implemented for Overlay NNCs.

nairashu, line 40: Simply removing this will break the validation for Overlay use case. We need to update the logic such that this check is only implemented for Overlay NNCs.

nairashu: Discussed offline that this is a newly introduced section. Resolving comment.

csfmomo: OK we can add the comment in a separate PR.

cns/kubecontroller/nodenetworkconfig/conversion.go
csfmomo: for current cni, yes, we don't need it.

PR #4320, opened 2026-04-07
ci: ignore status of azuresecuritylinuxagent during testing

QxBytes:
**Reason for Change**:

During testing we check that all pods are up and running, but a new external component out of our control is crashing a % of the time, causing flakiness. We ignore this namespace as the pod is also a host networking pod unrelated to our components.

Also modifies an error path to return a real error. Previously it always returned a wrapped nil value, which resolves to just nil (so if we hit the error path it would hide the error).

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Ran 5-6 full runs of the pipeline with this change and have yet to see the issue reappear.

### DIFF
--- a/test/internal/kubernetes/utils.go
+++ b/test/internal/kubernetes/utils.go
@@ -269,7 +269,7 @@ func MustSetupCNP(ctx context.Context, clientset *cilium.Clientset, cnpPath stri
 
 func Int32ToPtr(i int32) *int32 { return &i }
 
-func WaitForPodsRunning(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelselector string, excludeNamespaces ...string) error {
+func WaitForPodsRunning(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelselector string, excludeNamespaces []string) error {
 	podsClient := clientset.CoreV1().Pods(namespace)
 
 	excludeSet := make(map[string]struct{}, len(excludeNamespaces))
--- a/test/integration/datapath/datapath_linux_test.go
+++ b/test/integration/datapath/datapath_linux_test.go
@@ -131,7 +131,7 @@ func setupLinuxEnvironment(t *testing.T) {
 	})
 
 	t.Log("Waiting for pods to be running state")
-	err = kubernetes.WaitForPodsRunning(ctx, clientset, *podNamespace, podLabelSelector)
+	err = kubernetes.WaitForPodsRunning(ctx, clientset, *podNamespace, podLabelSelector, nil)
 	if err != nil {
 		t.Fatalf("Pods are not in running state due to %+v", err)
 	}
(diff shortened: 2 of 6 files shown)


test/internal/kubernetes/utils.go
```diff
@@ -269,9 +269,14 @@ func MustSetupCNP(ctx context.Context, clientset *cilium.Clientset, cnpPath stri
...
 
-func WaitForPodsRunning(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelselector string) error {
+func WaitForPodsRunning(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelselector string, excludeNamespaces ...string) error {
 	podsClient := clientset.CoreV1().Pods(namespace)
 
+	excludeSet := make(map[string]struct{}, len(excludeNamespaces))
+	for _, ns := range excludeNamespaces {
+		excludeSet[ns] = struct{}{}
+	}
+
```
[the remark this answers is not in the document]
QxBytes: this exclusion list is really only for when we specify all pods (i.e. not specifying a namespace), and there will almost certainly be system pods on the cluster.

test/internal/kubernetes/utils.go
```diff
@@ -269,9 +269,14 @@ func MustSetupCNP(ctx context.Context, clientset *cilium.Clientset, cnpPath stri
 
 func Int32ToPtr(i int32) *int32 { return &i }
 
-func WaitForPodsRunning(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelselector string) error {
+func WaitForPodsRunning(ctx context.Context, clientset *kubernetes.Clientset, namespace, labelselector string, excludeNamespaces ...string) error {
```
jpayne3506: While this will work for our internal testing, it will still block k8s_e2e testing as it also checks all pods. Are we able to exclude namespaces in its testing as well?

jpayne3506 (changes requested):

test/internal/kubernetes/utils.go
QxBytes: doesn't the upstream e2e only check kube-system namespace readiness?  https://github.com/kubernetes/kubernetes/blob/master/test/e2e/e2e.go#L229
If not can you point out where it checks all pods across all namespaces?

jpayne3506: Confirmed with local testing that it is constrained to kube-system.

jpayne3506 (approved):

test/internal/kubernetes/utils.go
vipul-21: Not in favor of adding a variadic parameter, it makes it harder to expand the function in future. Maybe pass in struct with these fields and the callers can pass in the fields.

vipul-21:
```
type Options struct {
    Namespace         string
    LabelSelector     string
    ExcludeNamespaces []string
}
```

QxBytes: replaced with []string parameter

PR #4321, opened 2026-04-07
ci: force cns test manifests to always pull images

QxBytes:
**Reason for Change**:

in the test pipeline, saw some unexpected behavior where cns pulled the wrong arch version from the multiarch manifest and could not recover. theorizing this was a transient error and would be fixed by repulling. however, currently, the daemonset will only pull once and never again-- this changes that behavior

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
See #20260407.12 -- cns is in crashloopbackoff with "exec format error"

### DIFF
--- a/test/integration/manifests/cns/daemonset-linux.yaml
+++ b/test/integration/manifests/cns/daemonset-linux.yaml
@@ -42,7 +42,7 @@ spec:
       containers:
         - name: cns-container
           image: acnpublic.azurecr.io/azure-cns:latest
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: Always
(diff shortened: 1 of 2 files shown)


QxBytes:
> > in the test pipeline, saw some unexpected behavior where cns pulled the wrong arch version from the multiarch manifest and could not recover. theorizing this was a transient error and would be fixed by repulling.
> 
> How did you confirm this?

If it weren't transient the error wouldn't just appear on one node. The confirmation that this issue exists is in the mentioned pipeline run. I'm trying to imagine a solution where repulling wouldn't work since other stages/scenarios pull successfully (and some, I assume, pull later), so if we keep retrying the pull, eventually we should get the same image the other (working) clusters use.

PR #4334, opened 2026-04-08
feat(pipeline): support multiple subscriptions per region in swiftv2 long-running tests

isaac-dasan:
## Summary

Enable the swiftv2 long-running pipeline to run independent stage chains per region×subscription pair, so the same region (e.g. `eastus2euap`) can have clusters in different Azure subscriptions tested in parallel. Includes numerous pipeline fixes discovered during rollout.

### Core multi-subscription support
1. **Multi-subscription scenarios matrix** — Refactor the pipeline from a single `subscriptionId` to a `scenarios` list where each entry carries its own `subscriptionId`, `location`, `label`, and feature flags (`enableLinuxBYON`, `enableDatapathLinux`). Each scenario gets its own full stage chain (infra → tests → cleanup).

2. **Subscription context in all scripts** — Add `az account set --subscription` to every provisioning script so commands target the correct subscription instead of relying on the CLI default.

3. **Rename `longrunningZones` → `longRunningPodZones`** — Align parameter naming with its actual purpose; fix stage template filename to match.

### Infra reliability
4. **Idempotent provisioning** — All infra scripts now check for existing resources before creating (AKS clusters, subnets, peerings, storage accounts, PEs, NSGs, VMSS). Re-runs after partial failures self-heal without duplicating resources.

5. **Storage account discover-then-reuse** — Replace random-name-per-run storage creation with prefix-based discovery (`sa1*`/`sa2*`). Downstream stages always discover accounts from the RG, eliminating the `stageDependencies` variable passthrough chain.

6. **Remove verify gate** — Delete `verify_infrastructure.sh` and the `VerifyInfrastructure`/`VerifyAccelnetBYON` jobs. Since scripts are idempotent, they always run and skip existing resources.

7. **Skip subnet delegation when SAL exists** — Check `serviceAssociationLinks` via `az rest` before calling the expensive subnet delegator. Fix jq path for the SAL check.

8. **Increase lease TTL to 300 min** — Prevent concurrent datapath test collisions when test runs exceed 2 hours.

9. **Remove runtime RBAC management** — Delete `manage_storage_rbac.sh`; require roles to be pre-assigned rather than granted at pipeline time.

### Pipeline fixes discovered during rollout
10. **Feature flags for per-scenario control** — Add `enableLinuxBYON` and `enableDatapathLinux` flags to skip Linux BYON provisioning or Linux datapath tests for subscriptions/regions that don't support them.

11. **Use `stageLabel` for artifact names** — Fix artifact name collisions across scenarios by incorporating the stage label; set subscription context in test jobs.

12. **Defer SSH key fetch** — Only fetch the SSH key from Key Vault when VMSS creation is actually needed, avoiding unnecessary failures when BYON is disabled.

13. **Boolean parameter handling** — Use `eq(x, true)` instead of bare string comparisons for boolean checks in ADO object parameters.

14. **Remove incorrect node labeling from `create_aks.sh`** — Stop labeling system pool nodes with workload labels that should only be set on BYON nodes.

### Test code fixes
15. **Wait for Deployment deletion before recreating** — `DeleteDeploymentAndWait` in `k8s_helpers.go` now polls until the Deployment object is fully deleted before returning, preventing `AlreadyExists` errors on recreation.

16. **Auto-remediate remediator-unschedulable taint** — `ensure_zone_nodepools.sh` detects and removes the `remediator=unschedulable:NoSchedule` taint on zone pool nodes that blocks pod scheduling.

17. **Scope MTPNC cleanup to deployment** — In the rotating pod test, scope the MTPNC cleanup wait to only the specific deployment's pods rather than the entire namespace.

18. **Single PodNetwork per region** — Replace per-zone, per-test-type PodNetworks (`pn-rotating-z<N>`, `pn-alwayson-z<N>`) with a single region-scoped PN (`pn-longrunning-<buildID>`) shared by all zones and both test scenarios. DNCRC will not reconcile different PNs with the same network and subnet combination. All PodNetworkInstances now reference the shared PN.


.pipelines/swiftv2-long-running/template/long-running-pipeline-template.yaml
```diff
@@ -31,42 +26,45 @@ variables:
...
+    - ${{ each zone in scenario.longRunningPodZones }}:
+      - template: long-running-pod-tests-stage.yaml
         parameters:
-          location: ${{ region.location }}
-          subscriptionId: ${{ parameters.subscriptionId }}
-          rgName: ${{ parameters.resourceGroupName }}-${{ region.location }}
+          location: ${{ scenario.location }}
+          subscriptionId: ${{ scenario.subscriptionId }}
+          stageLabel: ${{ scenario.label }}
+          rgName: ${{ parameters.resourceGroupName }}-${{ scenario.location }}
```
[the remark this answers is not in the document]
isaac-dasan: no it matches

.pipelines/swiftv2-long-running/scripts/create_vnets.sh
```diff
@@ -81,8 +88,20 @@ delegate_subnet() {
...
+
+    # Check if SAL already exists — skip expensive delegation if so
+    local sal
+    sal=$(az rest --method get \
+      --url "${subnet_id}?api-version=2024-05-01" \
+      2>/dev/null | jq -r '.properties.serviceAssociationLinks[0].properties.link // empty')
+    if [[ -n "$sal" ]]; then
+      echo "SAL already exists on $subnet in $vnet (link: $sal). Skipping delegation."
+      return 0
+    fi
```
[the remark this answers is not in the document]
isaac-dasan: This works without the suggester prefix

.pipelines/swiftv2-long-running/scripts/create_storage.sh
```diff
@@ -6,33 +6,41 @@ SUBSCRIPTION_ID=$1
...
-    echo "[OK] Storage account $SA verified successfully."
-  else
-    echo "[ERROR] Storage account $SA not found after creation!" >&2
-    exit 1
-  fi
-done
+
+# Discover existing storage accounts or create new ones
+SA1=$(az storage account list -g "$RG" --query "[?starts_with(name, 'sa1')].name | [0]" -o tsv 2>/dev/null || true)
+SA2=$(az storage account list -g "$RG" --query "[?starts_with(name, 'sa2')].name | [0]" -o tsv 2>/dev/null || true)
```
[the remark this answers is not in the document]
isaac-dasan: fixed

.pipelines/swiftv2-long-running/scripts/create_aks.sh
```diff
@@ -77,25 +80,37 @@ for i in $(seq 1 "$CLUSTER_COUNT"); do
...
+    if [[ "$EXISTING_STATE" == "Succeeded" ]]; then
+      echo "Cluster $CLUSTER_NAME already exists (state: $EXISTING_STATE). Skipping creation."
+    else
+      make -C ./hack/aks azcfg AZCLI=az REGION=$LOCATION
+      make -C ./hack/aks swiftv2-podsubnet-cluster-up \
+        AZCLI=az REGION=$LOCATION \
+        SUB=$SUBSCRIPTION_ID \
+        GROUP=$RG \
+        CLUSTER=$CLUSTER_NAME \
+        VM_SIZE=$VM_SKU_DEFAULT
```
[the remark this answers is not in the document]
isaac-dasan: out of scope

.pipelines/swiftv2-long-running/LONGRUNNING-TESTS.md
```diff
@@ -270,7 +276,7 @@ All operations are designed to be safe to re-run. PodNetworks, PodNetworkInstanc
 │   ├── long-running-pipeline-template.yaml    # Infra setup + datapath tests + long-running tests
 │   ├── infrastructure-setup-stage.yaml        # Per-location infra verify + conditional setup
 │   ├── datapath-tests-stage.yaml              # Per-workload datapath test stage
-│   └── longrunning-pod-tests-stage.yaml       # Per-zone: EnsureNodePool + rotating + always-on + connectivity
+│   └── long-running-pod-tests-stage.yaml      # Per-zone: EnsureNodePool + rotating + always-on + connectivity
 └── scripts/
     ├── verify_infrastructure.sh               # Smart infra check (skip setup if exists)
```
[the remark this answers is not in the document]
isaac-dasan: fixed

.pipelines/swiftv2-long-running/LONGRUNNING-TESTS.md
```diff
@@ -4,19 +4,25 @@ This document covers the **long-running zone-aware pod tests** — rotating pods
 
 These tests are **integrated into the main pipeline** (`pipeline.yaml`). The `longrunningRegions` parameter maps each region to its availability zones. Zone node pool creation and per-zone test stages run alongside the existing datapath tests.
```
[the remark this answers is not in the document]
isaac-dasan: fixed

.pipelines/swiftv2-long-running/template/datapath-tests-stage.yaml
```diff
@@ -194,14 +190,14 @@ stages:
...
+                echo "Discovered: STORAGE_ACCOUNT_1=$STORAGE_ACCOUNT_1, STORAGE_ACCOUNT_2=$STORAGE_ACCOUNT_2"
                 
                 if [ -z "$STORAGE_ACCOUNT_1" ] || [ -z "$STORAGE_ACCOUNT_2" ]; then
-                  echo "Storage account variables not set, discovering from resource group..."
-                  STORAGE_ACCOUNT_1=$(az storage account list -g ${{ parameters.rgName }} --query "[?starts_with(name, 'sa1')].name" -o tsv)
-                  STORAGE_ACCOUNT_2=$(az storage account list -g ${{ parameters.rgName }} --query "[?starts_with(name, 'sa2')].name" -o tsv)
-                  echo "Discovered: STORAGE_ACCOUNT_1=$STORAGE_ACCOUNT_1, STORAGE_ACCOUNT_2=$STORAGE_ACCOUNT_2"
+                  echo "##vso[task.logissue type=error]Could not discover storage accounts in ${{ parameters.rgName }}"
+                  exit 1
                 fi
```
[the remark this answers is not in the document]
isaac-dasan: fixed

sivakami-projects (dismissed):

.pipelines/swiftv2-long-running/scripts/create_aks.sh
```diff
@@ -77,40 +80,43 @@ for i in $(seq 1 "$CLUSTER_COUNT"); do
...
-    make -C ./hack/aks azcfg AZCLI=az REGION=$LOCATION
-    make -C ./hack/aks swiftv2-podsubnet-cluster-up \
-      AZCLI=az REGION=$LOCATION \
-      SUB=$SUBSCRIPTION_ID \
-      GROUP=$RG \
-      CLUSTER=$CLUSTER_NAME \
-      VM_SIZE=$VM_SKU_DEFAULT
-    wait_for_provisioning "$RG" "$CLUSTER_NAME"
+    # Check if cluster already exists and is healthy
+    EXISTING_STATE=$(az aks show -g "$RG" -n "$CLUSTER_NAME" --query provisioningState -o tsv 2>/dev/null || true)
```
miguelgoms, line 84: technically provisioningState can be "succeeded" with powerState => "Stopped" if your cluster was stopped intentionally vs failed. 
we could check for powerstate / other cases, or issue 'az aks update' / 'az aks start' without params on the existing cluster and let RP reconcile the goal state iff needed.

.pipelines/swiftv2-long-running/scripts/create_aks.sh
```diff
@@ -15,6 +15,9 @@ CLUSTER_COUNT=2
 PODS_PER_NODE=7
 CLUSTER_PREFIX="aks"
 
+echo "Setting active subscription to $SUBSCRIPTION_ID"
+az account set --subscription "$SUBSCRIPTION_ID"
```
miguelgoms, line 19: nit: if we're passing subscription_id explicitly in existing commands, why do we need to set this default instead instead of consistently passing this explicitly?

.pipelines/swiftv2-long-running/scripts/create_storage.sh
```diff
@@ -6,36 +6,62 @@ SUBSCRIPTION_ID=$1
...
-  && echo "Storage account $SA created successfully."
-  
-  if az storage account show --name "$SA" --resource-group "$RG" &>/dev/null; then
-    echo "[OK] Storage account $SA verified successfully."
-  else
-    echo "[ERROR] Storage account $SA not found after creation!" >&2
+
+# Deterministic storage account discovery: sort by name, fail if ambiguous
+discover_storage_account() {
+  local prefix="$1"
```
miguelgoms: if this pipeline owns creating and tracking these storage accounts - can we simplify this whole discovery thing using resource tags for search, filter, disambiguation?

.pipelines/swiftv2-long-running/template/long-running-pipeline-template.yaml
```diff
@@ -108,54 +108,69 @@ stages:
...
+  - ${{ each scenario in parameters.scenarios }}:
     - ${{ each workload in parameters.workloadType }}:
-      # Skip accelnet workloads in regions where accelnet is not enabled (currently only centraluseuap has quota)
-      - ${{ if or(not(contains(workload, 'l1vh-accelnet')), eq(loc, 'centraluseuap')) }}:
+      # Skip accelnet workloads unless the scenario has accelnet enabled.
+      # Skip linux-byon workloads unless the scenario has linux BYON enabled.
+      # Skip swiftv2-linux workloads unless the scenario has enableDatapathLinux enabled.
+      # NOTE: must use eq(x, true) — bare booleans from object params passed through
+      # 'extends' are coerced to strings, and "false" is truthy.
+      - ${{ if and(or(not(eq(workload, 'swiftv2-linux')), eq(scenario.enableDatapathLinux, true)), or(not(contains(workload, 'l1vh-accelnet')), eq(scenario.enableAccelnet, true)), or(not(eq(workload, 'swiftv2-linux-byon')), eq(scenario.enableLinuxBYON, true))) }}:
```
miguelgoms: this long condition is hard to parse and seems redundant to the inner checks - if these are now mutually exclusive scenarios, do we need this or can we have simpler if/checks that map to each of these "Skip x scenario under y"?

.pipelines/swiftv2-long-running/scripts/create_aks.sh
isaac-dasan: I don't think I did a full sweep of existing cmds to update them to use subscription_id. Also i thought it will increase the # lines changed for this PR substantially. I think setting the subscription explicitly also reduces the risk of future errors in case someone added a new line and forgot to add subscription parameter. By default ADO pipelines sets it's own subscription for every job. so this is a good guard

.pipelines/swiftv2-long-running/scripts/create_aks.sh
isaac-dasan: I'd prefer not to add speculative error handling for states we haven't observed failing in practice. These are bash scripts that aren't unit-testable against a real AKS control plane, so any "fix" for hypothetical Failed/Updating states would itself be unverified. If we hit one of these edge cases in a real run, I'd rather fix it then with confidence that the fix actually works, rather than guessing at the right recovery logic now. The az CLI behavior around re-creating existing resources can be subtle and version-dependent.

.pipelines/swiftv2-long-running/template/long-running-pipeline-template.yaml
isaac-dasan: These aren't mutually exclusive scenarios - it's a matrix of `scenario × workload` where each workload type is gated by its own feature flag. ADO's `${{ each }}` doesn't support `continue/skip`, so the only way to filter combinations is a single `${{ if }}` guard. Breaking it into separate blocks would mean duplicating the template inclusion for each workload type. The inline comments above (lines 116-119) explain the intent of each clause. Open to suggestions if there's a cleaner ADO syntax I'm missing.

miguelgoms: duplicating the template inclusion could be cleaner here. but if you are keeping the nested if - add a more accurate comment for the big-if-guard for the inclusion, and move the "skip if x" comment to the line of the corresponding gate.

.pipelines/swiftv2-long-running/scripts/create_storage.sh
isaac-dasan: resource tags are not unique as well. we could end up having multiple storage accounts with the same resource tags. 
Also if I were to add the resource tags, I need to manually add them for all provisioned storge accounts across all regions or delete all the storage accounts and let them get recreated. Why fix something that is not broken?

.pipelines/swiftv2-long-running/template/long-running-pipeline-template.yaml
isaac-dasan: fixed this

.pipelines/swiftv2-long-running/scripts/create_aks.sh
isaac-dasan: Added locks and skip gc tags so MC resource group won't get deleted. We'll do another PR if the cluster gets into an unhealthy state and we validated a fix for that.

.pipelines/swiftv2-long-running/scripts/create_storage.sh
isaac-dasan: Added resource tags to provide additional uniqueness to the storage account, since the prefix sa1, sa2 is common for storage accounts and could cause issues in future if someone manually added those

miguelgoms (approved):

sivakami-projects (approved):

PR #4339, opened 2026-04-09
feat: allow waagent to install iptables rule, in eBPF host routing mode

santhoshmprabhu:
**Reason for Change**:
Allow python scripts to install iptables rules. This is overapproximating waagent, which we need to allow to install iptables rules. We will look into ways to narrow down this. This is needed because in eBPF host routing mode, our ipttables block logic prevents rule installation, and the waagent script keeps trying. This in turn causes IPTablesRuleBlocked events to be created on the node, which is noise from the perspective of a customer. We are allowing python because in BPF, we don't see the actual script that's run, only the python executable.

**Issue Fixed**:

**Validation**:
Manually validated on an AKS cluster

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/bpf-prog/azure-block-iptables/bpf/src/block_iptables.bpf.c
+++ b/bpf-prog/azure-block-iptables/bpf/src/block_iptables.bpf.c
@@ -13,7 +13,7 @@
 #define EPERM 1
 #define IPT_SO_SET_REPLACE 64
 #define TASK_COMM_LEN 16
-#define COMM_COUNT 5
+#define COMM_COUNT 6
 #define IPPROTO_IP 0
 #define IPPROTO_IP6 41
 #define AF_NETLINK 16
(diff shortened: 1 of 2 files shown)


QxBytes (approved):
wondering how many more exceptions we'll need to add in the future? if any component complains the event is noise it might become difficult to keep managing all of these carveouts.

santhoshmprabhu:
> wondering how many more exceptions we'll need to add in the future? if any component complains the event is noise it might become difficult to keep managing all of these carveouts.

we need to rearchitect this. Allowing python is just to get this out for waagent, and buy us some time to figure out how to do it correctly.

tamilmani1989 (approved):

camrynl (approved):

PR #4345, opened 2026-04-10
ci: mitigate windows scenario flakiness

QxBytes:
**Reason for Change**:

Mitigates flakiness for windows and other test stages in the acn pr pipeline.

Ignore
- Extra stages created in pipeline.yaml -- these are for testing only
- Hardcoded 1.7.9-0 and MCR in .pipelines/templates/windows-overlay-e2e-step-template.yaml
- These MUST be reverted before merging though

All changes should only affect acn pr, but I have run load test/release test without any obvious issues.

Main Changes
- Moves k8s e2e job completely to a step template. Affects only the acn pr pipeline. The template intends to mimic the existing behavior completely and multiple test runs show that it runs without issue.  Affects windows and linux, all scenarios.
- Moves testing for each scenario to a single job, so it can be retried as a whole. I am aware this could increase the time for a test, but the benefit is that it is much easier to continueOnError and check a single job for a warning than multiple, which brings me to my next point
- Add e2e warning handler for the main windows tests-- dualstack-overlay, overlay, and stateless (3 total). This adds a job after the main testing to check if the logs mention any of the known issues in windows, such as null characters in the statefile (if windows bsod), the network not being found (occurs when the statefile has an hns network but it was somehow deleted from the vm). These issues have only occurred in the pipelines and are basically noise. Windows has already said byo cni for windows is not supported so these issues are unlikely to be fixed. If these log lines are detected, the stage is marked with a warning but still passes
- Adds exec: already started to the ubuntu cniv1 linux whitelist error messages-- this appears to be a bug with the cni framework-- exec.Run should never be called twice but the upstream cni framework code seems to make that a possible outcome making the test results unreliable if it occurs. 
- The check cni logs script now prints out the matched line, and failure logs will be collected even if the main testing stage succeeded with issues. Previously it only ran if the main testing job failed.
- Modifies dualstack test that checks hns network to accept either an HNSNetwork struct or an array of HNSNetwork structs. Previously, if only one HNSNetwork was returned, there would be an error unmarshalling into the array. The second network is the ext hns network which is not necessary for container networking (just a qol dummy network).
- Adds a script to collect logs from bad pods if they are not running for all sidecars and init containers, and prints out the status of all pods on failure.
- Update linux log collection script to get all cilium-related pod logs (sidecar, init containers) on failure if cilium scenario.
- Add retries to cilium install step.
- Increase timeout for konnectivity agent to get ready for k8s e2e test.

Mitigating Windows BYOCNI Scenarios
In the new windows overlay template, I aim to keep all existing tests but wanted to unify windows overlay into one template since previously there were 3 separate templates. Stateless step template was removed because it was windows-only. The others remain because they have linux counterparts.  I have added two new parts after investigating the issues with windows byocni nodes.
- Windows byo cni nodes do not seem to clean up the azure-vnet.json statefile when I restart the nodes
- Usually, windowsnodereset.ps1 will be called and then the cleanupnetwork.ps1 is called to remove stale statefiles like in linux
- However, cleanupnetwork.ps1 only runs the deletion if the cni name is set to "azure" in kubeclusterconfig.json. For byocni nodes, that value is set to "none"
- First change is to before tests run, patch the kubeclusterconfig.json on windows nodes to use plugin name "azure"
- Second change is to modify windowsnodereset.ps1 to use *>> instead of >> so the logs from the cleanupnetwork.ps1 script are visible in windowsnodereset.log-- otherwise we have no visibility into those logs
- I have also added collection of the windowsnodereset.log log to the failure logs collection step
- Witnessed the following for host networking pods during create which we determined were unrelated to our components (containerd?): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: hcs::CreateComputeSystem b9257d8579494625ee895a32e4be77da62991d162a858e44bcd29c116ff510b9: The parameter is incorrect.: unknown

**Issue Fixed**:

See above

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:


.pipelines/templates/windows-overlay-e2e-step-template.yaml
```diff
@@ -0,0 +1,124 @@
...
+            taint=`kubectl describe node $node | grep Taints | awk '{print $2}'`
+            if [ $taint == "node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule" ]; then
+                kubectl taint nodes $node node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule-
+            fi
+        done
+
+        # Run control plane scale tests
+        export CNI_IMAGE_REPO=MCR
+        export CNS_IMAGE_REPO=MCR
+        sudo -E env "PATH=$PATH" make test-load SCALE_UP=32 OS_TYPE=windows CNI_TYPE=${{ parameters.cniType }} VALIDATE_STATEFILE=true INSTALL_CNS=true ${{ parameters.testLoadArgs }} CNS_VERSION=v1.7.9-0 CNI_VERSION=v1.7.9-0 CLEANUP=true
```
jpayne3506: friendly comment to help remind to change

jpayne3506 (dismissed):

.pipelines/singletenancy/aks/e2e-job-template.yaml
```diff
@@ -72,36 +72,33 @@ stages:
...
+              hostport: true
       
       # can only succeed with issues in linux
       - template: ../../templates/warning-handler-job-template.yaml
         parameters:
           dependsOnToEmitWarning: ${{ parameters.name }}
           clusterName: ${{ parameters.clusterName }}-$(commitID)
           fileSearchPattern: "azure-vnet*.log"
-          searchPattern: "Initializing HTTP client with connection timeout"
+          searchPattern: "Initializing HTTP client with connection timeout|exec: already started"
```
vipul-21, line 97: I would say we should comment atleast what these errors are related to and why they might be skipped.

.pipelines/singletenancy/dualstack-overlay/dualstackoverlay-e2e-job-template.yaml
```diff
@@ -50,38 +50,58 @@ stages:
       - job: ${{ parameters.name }}_${{ parameters.os }}
         displayName: DualStack Overlay Test Suite | ${{ parameters.os }} - (${{ parameters.name }})
         timeoutInMinutes: 120
+        continueOnError: ${{ eq(parameters.os, 'windows') }}
```
vipul-21, line 54: So dual stack windows will never fail ?

.pipelines/singletenancy/dualstack-overlay/dualstackoverlay-e2e-job-template.yaml
```diff
@@ -50,38 +50,58 @@ stages:
...
+                regionClusterTest: $(REGION_DUALSTACKOVERLAY_CLUSTER_TEST)
+                scaleup: ${{ parameters.scaleup }}
+          - template: ../../templates/k8s-e2e-step-template.yaml
             parameters:
-              name: ${{ parameters.name }}
+              sub: $(BUILD_VALIDATIONS_SERVICE_CONNECTION)
               clusterName: ${{ parameters.clusterName }}-$(commitID)
               os: ${{ parameters.os }}
-              scaleup: ${{ parameters.scaleup }} # 50 in windows or 100 in linux
+              dualstack: ${{ eq(parameters.os, 'linux') }} # RUN IN LINUX not WINDOWS Currently broken for scenario and blocking releases, HNS is investigating. Covered by go test in E2E step template
```
vipul-21, line 83: Do we have an icm with HNS ? Are you tracking to revert this ?

hack/scripts/patch-kubeclusterconfig.sh
```diff
@@ -0,0 +1,59 @@
...
+
+podList=$(kubectl get pods -n kube-system -l os=windows,app=privileged-daemonset --no-headers -o custom-columns=NAME:.metadata.name)
+allSucceeded=true
+for pod in $podList; do
+  succeeded=false
+  for attempt in 1 2 3; do
+    echo "Attempt $attempt: Patching kubeclusterconfig.json on $pod"
+    if kubectl exec -n kube-system "$pod" -- powershell.exe -command \
+      'Get-Content "c:\k\kubeclusterconfig.json" -Raw | ConvertFrom-Json | % { $_.Cni.Name = "azure"; $_ } | ConvertTo-Json -Depth 20 | Set-Content "c:\k\kubeclusterconfig.json"'; then
+      echo "Successfully patched kubeclusterconfig.json on $pod"
```
vipul-21, line 19: When we patched this, does this have to be reverted back to `none` ?
I understand we want to clean up the state and this helps us do it, but does change anything else in terms on testing ? Or conflicts with how we test in byo windows cluster ?

.pipelines/singletenancy/dualstack-overlay/dualstackoverlay-e2e-job-template.yaml
QxBytes: If there is a failure, it will continue as SucceededWithIssues and then the next job `template: ../../templates/warning-handler-job-template.yaml` will determine if the logs contain any known errors. If they do not, that job fails. That job only runs if the main test job is marked as succeeded with issues. Similar pattern for the other windows e2es.

.pipelines/singletenancy/dualstack-overlay/dualstackoverlay-e2e-job-template.yaml
QxBytes: This comment was there even before this change-- I think John mentioned this issue at least 19 months ago. We can take a look at it again after this pr and see if it has been resolved.

hack/scripts/patch-kubeclusterconfig.sh
QxBytes: I don't believe so-- if we want to mimic what is in prod, they have their cni plugin name set to "azure". I guess to be sure we'd need to look at all the scripts on the /k/ directory on windows, but I think the e2e pipeline runs show that this change doesn't cause any immediately issues.

QxBytes: Just collected all the .ps1 scripts in /k/ and ran an analysis-- main changes
- cleanupnetwork.ps1 should clean up the azure hns network (which is how it should be done in prod)
- cleanupnetwork.ps1 should clean up statefiles (similar to prod/linux)
- kubeletstart.ps1 now restarts kubeproxy (behavior in prod)

vipul-21 (dismissed):

jpayne3506 (dismissed):

vipul-21 (dismissed):

jpayne3506 (dismissed):

tamilmani1989 (dismissed):

rbtr (approved):

jpayne3506 (approved):

PR #4351, opened 2026-04-14
ci: conditionally run e2es

QxBytes:
**Reason for Change**:

Does not run cilium e2es if we only modify `cni/` `ipam/` or `network/`
Does not run windows e2es if we only modify `linux` or `test` files (still runs windows e2es)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
If this is triggered in the merge queue, the PR is not classified as Pull Request and might not have target branch set, in which case it might diff against master instead of the branch it will merge into. I have made a change to detect the source branch if it is in the merge queue and extract the target branch that way, but even if it doesn't detect that it just means we might run more tests than necessary. If there are no files changed, should run all tests. If branch is refs/tag (tagged run) we also run all tests.


.pipelines/singletenancy/aks-swift/e2e-job-template.yaml
```diff
@@ -1,12 +1,28 @@
 parameters:
-  name: ""
-  displayName: ""
-  clusterType: ""
-  clusterName: ""
-  vmSize: ""
-  k8sVersion: ""
-  os: ""
-  dependsOn: ""
```
camrynl, line 9: seeing you updated these with default in all of the e2e templates - is this necessary?

QxBytes: shouldn't change the functionality of the pipeline but it does provide better validation-- right now I believe it can just take anything

.pipelines/templates/git-diff-check.yaml
```diff
@@ -0,0 +1,54 @@
+parameters:
+  - name: targetBranch
+    type: string
+    default: ""
+
+steps:
+  - checkout: self
+    displayName: 'Checkout Repository'
+    fetchDepth: 0
```
jpayne3506, line 9: nit: this done by default already.

.pipelines/templates/git-diff-check.yaml
```diff
@@ -0,0 +1,54 @@
...
+      echo "$OUTPUT"
+      
+      RUN_WINDOWS_TESTS=false
+      RUN_CILIUM_TESTS=false
+      echo "$OUTPUT" | grep -q '^RUN_WINDOWS_TESTS=true$' && RUN_WINDOWS_TESTS=true
+      echo "$OUTPUT" | grep -q '^RUN_CILIUM_TESTS=true$' && RUN_CILIUM_TESTS=true
+      
+      echo "##vso[task.setvariable variable=RunWindowsTests;isOutput=true]$RUN_WINDOWS_TESTS"
+      echo "##vso[task.setvariable variable=RunCiliumTests;isOutput=true]$RUN_CILIUM_TESTS"
+    name: GitDiffCheck
```
jpayne3506, line 59: We should skip to running all tests if it is a tag'd run imo. Even if there were no changes for a given directory in the diff

jpayne3506 (dismissed):
Talked to @QxBytes offline. We could have leveraged path filters + multiple pipelines to get the same result, but didn't want to deal with image generation and additional dependencies.

jpayne3506 (approved):

camrynl (approved):

PR #4379, opened 2026-04-29
ci: add retry to validate pods and ignore "completed" pods

QxBytes:
a completed pod already had its ip released, but the api server still seems to keep track of it so when we compare the actual ips to expected ips, actual ips has this completed pod's ip after a short period of time the completed pod disappears and the state becomes consistent again.

**Reason for Change**:

Fix flake in pipeline caused by new pod coming up and going into "completed" state (related to azuresecuritylinuxagent, not a component we control)

Example:
```
azuresecuritylinuxagent   azuresecuritylinuxagent-stale-pod-cleanup-29625080-rzbt2   0/1     Completed   0               84s   192.168.0.4     aks-nodepool1-15814823-vmss000001   <none>           <none>
```
will show up in the actual ips as `192.168.0.4` even though it would have already been cleaned up internally.

Example run:  
20260429.13 E2E-Cil EBPF on AKS Overlay Azure Linux stage > validate pods

```
2026/04/29 23:21:41 skipping pod azuresecuritylinuxagent/azuresecuritylinuxagent-stale-pod-cleanup-29625080-rzbt2 on node aks-nodepool1-15814823-vmss000001 in terminal phase Succeeded (IPs: [{192.168.0.4}]) when collecting pod IPs
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/templates/cilium-tests.yaml
+++ b/.pipelines/templates/cilium-tests.yaml
@@ -54,6 +54,7 @@ steps:
       make test-validate-state
     name: "validatePods"
     displayName: "Validate Pods"
+    retryCountOnTaskFailure: 3
 
   - script: |
       echo "Run wireserver and metadata connectivity Tests"
--- a/test/internal/kubernetes/utils_get.go
+++ b/test/internal/kubernetes/utils_get.go
@@ -2,6 +2,7 @@ package kubernetes
 
 import (
 	"context"
+	"log"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -46,6 +47,14 @@ func GetPodsIpsByNode(ctx context.Context, clientset *kubernetes.Clientset, name
 	}
 	ips := make([]string, 0, len(pods.Items)*2) //nolint
 	for index := range pods.Items {
+		// skip succeeded/failed pods since even though we released those ips internally,
+		// the api server still reports their existence
+		phase := pods.Items[index].Status.Phase
+		if phase == corev1.PodSucceeded || phase == corev1.PodFailed {
+			log.Printf("skipping pod %s/%s on node %s in terminal phase %s (IPs: %v) when collecting pod IPs",
+				pods.Items[index].Namespace, pods.Items[index].Name, nodeName, phase, pods.Items[index].Status.PodIPs)
+			continue
+		}
 		for _, podIP := range pods.Items[index].Status.PodIPs {
 			ips = append(ips, podIP.IP)
 		}


test/internal/kubernetes/utils_get.go
```diff
@@ -46,6 +47,14 @@ func GetPodsIpsByNode(ctx context.Context, clientset *kubernetes.Clientset, name
 	}
 	ips := make([]string, 0, len(pods.Items)*2) //nolint
 	for index := range pods.Items {
+		// skip succeeded/failed pods since even though we released those ips internally,
+		// the api server still reports their existence
+		phase := pods.Items[index].Status.Phase
+		if phase == corev1.PodSucceeded || phase == corev1.PodFailed {
```
vipul-21, line 53: Why pod failed check ? That can still have the IP address ?

.pipelines/templates/cilium-tests.yaml
```diff
@@ -54,6 +54,7 @@ steps:
       make test-validate-state
     name: "validatePods"
     displayName: "Validate Pods"
+    retryCountOnTaskFailure: 3
```
vipul-21, line 57: Is there a reason for this retry given we are fixing the validate pod step ?

camrynl: also if we immediately retry would it still fail?

QxBytes: mostly wanted to be consistent with the other steps which all have retries

test/internal/kubernetes/utils_get.go
vipul-21: Cilium endpoint list does not show the pod ip in case of failed state too.

.pipelines/templates/cilium-tests.yaml
vipul-21: I would say if the failure is fixed, we don't need the retries. Are there any other flakiness in this stage that needs retry ?

vipul-21 (approved):

.pipelines/templates/cilium-tests.yaml
QxBytes: If there is something transient where the validation fails, and then recovers after 30 seconds, the retries will catch it since they wait 15-30 secs between retries. This is just for the future, as all the other load tests (which this runs under the hood) run with retry count 3.

camrynl (approved):

PR #4404, opened 2026-05-11
fix: restrict host routes to InfraNIC in stateless CNI SwiftV2

behzad-mir:
## Summary

Fixes a transient `ENETUNREACH` error in SwiftV2 stateless CNI Linux during pod creation.

## Problem

In SwiftV2 stateless CNI with multiple NICs (InfraNIC + DelegatedVMNIC), the `Options[RoutesKey]` containing host gateway routes was being applied to **all** networks including the DelegatedVMNIC.

This caused kernel `ENETUNREACH` errors when trying to add routes like:
```
10.1.0.0/16 via 10.0.0.1 dev eth1
```

Where:
- `eth1` (DelegatedVMNIC) is in the **pod VNet** (10.1.0.0/16)
- Gateway `10.0.0.1` is in the **host VNet** (10.0.0.0/16)
- `eth1` has **no route to reach 10.0.0.1** → kernel returns `ENETUNREACH`

The route should only be applied to `eth0` (InfraNIC) which is connected to the host VNet.

## Root Cause

1. `setHostOptions()` populates `options[RoutesKey]` with a route to the host gateway for InfraNIC
2. All `EndpointInfo` structs get the **same options** via `shallowCopyIpamAddConfigOptions()`
3. `handleCommonOptions()` blindly applies these routes to every interface during network creation

## Fix

Moved the NICType check from handleCommonOptions (network/network_linux.go) to createEpInfo (cni/network/network.go). Non-infra NICs (DelegatedVMNIC, FrontendNIC) no longer receiveRoutesKey/IPTablesKey/SNATIPKey options from shallowCopyIpamAddConfigOptions() — these are infra-only host-namespace constructs. handleCommonOptions is reverted to its original unconditional behavior sincebad data is now prevented at the source.

## Why Transient

The error is transient because Go map iteration over interfaceInfo is non-deterministic. When InfraNIC processes first, its CreateNetwork installs the route on eth0; the subsequent DelegatedVMNIC attemptgets "file exists" and skips. When DelegatedVMNIC processes first, it hits ENETUNREACH. On retry, iteration order may differ.

## Testing

- [x] Code compiles successfully
- [ ] Manual testing on SwiftV2 cluster

## Requirements

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -134,18 +134,14 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
 
 func (nm *networkManager) handleCommonOptions(ifName string, nwInfo *EndpointInfo) error {
 	var err error
-	// Only apply RoutesKey routes for InfraNIC - host gateway routes (set by setHostOptions)
-	// are not reachable from secondary NICs like DelegatedVMNIC which are on different subnets.
-	// Empty NICType is treated as InfraNIC for legacy compatibility.
-	if routes, exists := nwInfo.Options[RoutesKey]; exists && (nwInfo.NICType == cns.InfraNIC || nwInfo.NICType == "") {
+	if routes, exists := nwInfo.Options[RoutesKey]; exists {
 		err = addRoutes(nm.netlink, nm.netio, ifName, routes.([]RouteInfo))
 		if err != nil {
 			return err
 		}
 	}
 
-	// IPTables rules (SNAT for DNS/IMDS) are also only relevant for InfraNIC traffic
-	if iptcmds, exists := nwInfo.Options[IPTablesKey]; exists && (nwInfo.NICType == cns.InfraNIC || nwInfo.NICType == "") {
+	if iptcmds, exists := nwInfo.Options[IPTablesKey]; exists {
 		err = nm.addToIptables(iptcmds.([]iptables.IPTableEntry))
 		if err != nil {
 			return err
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -134,14 +134,18 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
 
 func (nm *networkManager) handleCommonOptions(ifName string, nwInfo *EndpointInfo) error {
 	var err error
-	if routes, exists := nwInfo.Options[RoutesKey]; exists {
+	// Only apply RoutesKey routes for InfraNIC - host gateway routes (set by setHostOptions)
+	// are not reachable from secondary NICs like DelegatedVMNIC which are on different subnets.
+	// Empty NICType is treated as InfraNIC for legacy compatibility.
+	if routes, exists := nwInfo.Options[RoutesKey]; exists && (nwInfo.NICType == cns.InfraNIC || nwInfo.NICType == "") {
 		err = addRoutes(nm.netlink, nm.netio, ifName, routes.([]RouteInfo))
 		if err != nil {
 			return err
 		}
 	}
 
-	if iptcmds, exists := nwInfo.Options[IPTablesKey]; exists {
+	// IPTables rules (SNAT for DNS/IMDS) are also only relevant for InfraNIC traffic
+	if iptcmds, exists := nwInfo.Options[IPTablesKey]; exists && (nwInfo.NICType == cns.InfraNIC || nwInfo.NICType == "") {
 		err = nm.addToIptables(iptcmds.([]iptables.IPTableEntry))
 		if err != nil {
 			return err
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -736,6 +736,15 @@ func (plugin *NetPlugin) createEpInfo(opt *createEpInfoOpt) (*network.EndpointIn
 		endpointID = plugin.nm.GetEndpointIDByNicType(opt.args.ContainerID, ifName, opt.ifInfo.NICType)
 	}
 
+	// Only copy options (RoutesKey, IPTablesKey, SNATIPKey) for InfraNIC or legacy (empty NICType).
+	// These options are infra-only host-namespace constructs set by setHostOptions and are not
+	// applicable to delegated/secondary NICs on either Linux or Windows.
+	// For non-infra NICs, options remains nil so handleCommonOptions becomes a no-op.
+	var options map[string]interface{}
+	if opt.ifInfo.NICType == cns.InfraNIC || opt.ifInfo.NICType == "" {
+		options = opt.ipamAddConfig.shallowCopyIpamAddConfigOptions()
+	}
+
 	endpointInfo := network.EndpointInfo{
 		NetworkID:                     opt.networkID,
 		Mode:                          opt.ipamAddConfig.nwCfg.Mode,
@@ -744,7 +753,7 @@ func (plugin *NetPlugin) createEpInfo(opt *createEpInfoOpt) (*network.EndpointIn
 		BridgeName:                    opt.ipamAddConfig.nwCfg.Bridge,
 		NetworkPolicies:               networkPolicies, // nw and ep policies separated to avoid possible conflicts
 		NetNs:                         opt.ipamAddConfig.args.Netns,
-		Options:                       opt.ipamAddConfig.shallowCopyIpamAddConfigOptions(),
+		Options:                       options,
 		DisableHairpinOnHostInterface: opt.ipamAddConfig.nwCfg.DisableHairpinOnHostInterface,
 		IsIPv6Enabled:                 opt.ipv6Enabled, // present infra only
 
--- a/cni/network/network_linux_test.go
+++ b/cni/network/network_linux_test.go
@@ -472,9 +472,7 @@ func TestPluginLinuxAdd(t *testing.T) {
 							},
 							Suffix: "myDomain",
 						},
-						Options: map[string]interface{}{
-							"testflag": "copy",
-						},
+						Options: nil,
 						// matches with cns ip configuration
 						IPAddresses: []net.IPNet{
 							{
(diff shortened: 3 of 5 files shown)


network/network_linux.go
```diff
@@ -134,14 +134,18 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
 
 func (nm *networkManager) handleCommonOptions(ifName string, nwInfo *EndpointInfo) error {
 	var err error
-	if routes, exists := nwInfo.Options[RoutesKey]; exists {
+	// Only apply RoutesKey routes for InfraNIC - host gateway routes (set by setHostOptions)
+	// are not reachable from secondary NICs like DelegatedVMNIC which are on different subnets.
+	// Empty NICType is treated as InfraNIC for legacy compatibility.
+	if routes, exists := nwInfo.Options[RoutesKey]; exists && (nwInfo.NICType == cns.InfraNIC || nwInfo.NICType == "") {
 		err = addRoutes(nm.netlink, nm.netio, ifName, routes.([]RouteInfo))
```
tamilmani1989, line 138: what routes get added here?

network/network_linux.go
```diff
@@ -134,14 +134,18 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
 
 func (nm *networkManager) handleCommonOptions(ifName string, nwInfo *EndpointInfo) error {
 	var err error
-	if routes, exists := nwInfo.Options[RoutesKey]; exists {
+	// Only apply RoutesKey routes for InfraNIC - host gateway routes (set by setHostOptions)
+	// are not reachable from secondary NICs like DelegatedVMNIC which are on different subnets.
+	// Empty NICType is treated as InfraNIC for legacy compatibility.
+	if routes, exists := nwInfo.Options[RoutesKey]; exists && (nwInfo.NICType == cns.InfraNIC || nwInfo.NICType == "") {
```
isaac-dasan: I think this hides the bug but doesn't fully fix it. The actual race is in `EndpointCreate`: in stateless mode `ExternalInterfaces` is empty every ADD, so whichever NIC the Go map yields first becomes the network's `extIf`. If DelegatedVMNIC wins, the network gets bound to eth1 here. Now `handleCommonOptions` skips the route (good, no ENETUNREACH), but the InfraNIC iteration that follows hits `GetNetworkInfo == found` and skips `CreateNetwork` entirely.

with this fix in place, when DelegatedVMNIC wins map order in EndpointCreate, does the host route get installed elsewhere or do we end up with nodes that never install it? Worth checking ip route show on a stateless node post-rollout.

isaac-dasan (commented):
Diagnosis of the shared-options leak looks right and the gating is the right shape.
Two things:

1. The "transient because subsequent ADDs find existing networks in memory" line in the description is the stateful story. In stateless mode the race fires on every pod create; it just looks transient because kubelet retries eventually hit the lucky map order. Worth fixing in the description.
2. wrapping the `EndpointCreate` error with `kubelet.NetworkNotReadyErrorMsg` would drop kubelet's retry from ~10s to ~1s. Not needed once this PR is in (ENETUNREACH goes away), but any other failure on this path still costs us ~50s per affected pod in our 2000-node SwiftV2 runs

isaac-dasan (commented):
Can we also make sure this fix is backported to all previous CNI release going back to 1.6? For our test subscriptions, stateless CNI is enabled for all k8s version above 1.30

network/network_linux.go
behzad-mir: So in stateless mode, we call createNetwork every ADD call rather than reading it from persisted store like statefull. The current fix prevents going through route ADD again. 
I made a new commit to have the EndpointInfo sorted as well so we never process Delegated NIC 
before it.

network/network_linux.go
behzad-mir: This is a cross-VM pod-to-pod routing rule for the host namespace.  Without this route, when a pod on this VM sends traffic to a pod on a different VM in the same NC subnet, the host kernel wouldn't know to forward it via the infra gateway — it would try to deliver it locally (since 10.1.x.x looks like it belongs to a local subnet from the pod NIC's perspective).

network/network.go
```diff
@@ -316,6 +317,24 @@ func (nm *networkManager) GetNumEndpointsByContainerID(containerID string) int {
 func (nm *networkManager) EndpointCreate(cnsclient apipaClient, epInfos []*EndpointInfo) error {
 	eps := []*endpoint{} // save endpoints for stateless
 
+	// Sort epInfos so InfraNIC is processed first. In stateless mode, ExternalInterfaces is empty
+	// on every ADD, so whichever NIC is processed first determines which interface the network
+	// gets bound to. If DelegatedVMNIC wins the race, the network gets created with eth1 as extIf,
+	// and when InfraNIC is processed it finds the network already exists and skips CreateNetwork,
+	// resulting in host routes never being installed on eth0.
```
isaac-dasan: nit: This comment is not accurate and it was not the issue observed. Any of the k8s "system" pod's ADDs at node bootstrap install that route deterministically, so it's almost always present.
When `DelegatedVMNIC` wins map order, `SecondaryEndpointClient` moves the delegated NIC into the pod ns, then the subsequent `InfraNIC` iteration finds the network bound to that (now-gone-from-host) interface and `TransparentEndpointClient.AddEndpoints` fails with `GetNetworkInterfaceByName ... no such network interface`. Might be worth updating the comment to reflect that

behzad-mir: I think the comment is correct since we added a gate to prevent Delegated NIC get the IP route all together on first commit. So the sorting is just to prevent skipCreateNetwork for the first pod  ( which was as an issue for both stateful and Stateless)

isaac-dasan: My comment was more about this - `resulting in host routes never being installed on eth0`

isaac-dasan (dismissed):

network/network.go
```diff
@@ -316,6 +317,25 @@ func (nm *networkManager) GetNumEndpointsByContainerID(containerID string) int {
 func (nm *networkManager) EndpointCreate(cnsclient apipaClient, epInfos []*EndpointInfo) error {
 	eps := []*endpoint{} // save endpoints for stateless
 
+	// Sort epInfos so InfraNIC is processed first. In stateless mode, ExternalInterfaces is empty
+	// on every ADD, so whichever NIC is processed first determines which interface the network
+	// gets bound to. If DelegatedVMNIC wins the race, the network gets created with eth1 as extIf.
+	// SecondaryEndpointClient then moves eth1 into the pod namespace, and the subsequent InfraNIC
+	// iteration finds the network bound to that now-gone interface, causing
+	// TransparentEndpointClient.AddEndpoints to fail with "no such network interface".
+	slices.SortFunc(epInfos, func(a, b *EndpointInfo) int {
```
QxBytes: was this always an issue with a % chance of occurring or is this a new scenario? How did this work before? Haven't we had linux swiftv2 for a while now?

network/network.go
```diff
@@ -316,6 +317,25 @@ func (nm *networkManager) GetNumEndpointsByContainerID(containerID string) int {
...
+		// InfraNIC (or empty, for legacy) should come first
+		aIsInfra := a.NICType == cns.InfraNIC || a.NICType == ""
```
QxBytes: If we are going to start sorting the endpoint infos, I would do so after the createEpInfo loop in `cni/network/network.go` .

network/network_linux.go
QxBytes: If we want to skip adding routes, `nwInfo.Options[RoutesKey]` should not exist. I believe Options is set in `cni/network/network.go` > `createEpInfo`. Does it make sense for delegated/secondary nics to have iptables or routes (across windows + linux)? Please check with swiftv2 folks for this. If only infra nics have these constructs, I would prefer that we move this checking of the nic type to the `createEpInfo` function where if the nic type is not infra or empty, we just don't call `shallowCopyIpamAddConfigOptions`

network/network_linux.go
```diff
@@ -134,14 +134,18 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
...
 		err = addRoutes(nm.netlink, nm.netio, ifName, routes.([]RouteInfo))
 		if err != nil {
 			return err
 		}
 	}
 
-	if iptcmds, exists := nwInfo.Options[IPTablesKey]; exists {
+	// IPTables rules (SNAT for DNS/IMDS) are also only relevant for InfraNIC traffic
```
QxBytes: similar comment as above

network/network.go
behzad-mir: Not his was always an issue that was happening sometimes on stateful CNI too. So it was transient and next ADD call might have corrected it

network/network_linux.go
behzad-mir: @QxBytes I went through shallowCopyIpamAddConfigOptions() and all the options it copies (RoutesKey, IPTablesKey, SNATIPKey) are indeed InfraNIC-related. They're only populated insideconfigureDefaultAddResult → setHostOptions, which only runs for InfraNIC. FrontendNIC/DelegatedVMNIC paths (configureSecondaryAddResult) never write to the shared options map. I'm inclined to do your suggestion for cleaner code. Just need sign-off from @tamilmani1989  & @isaac-dasan  to make sure it's not affecting any other scenario. Singularity also does not use any of these option keys for theFrontendNIC type AFAIK.

network/network.go
behzad-mir: make sense. Will be moving it to createEpInfo loop before EndpointCreate.

network/network_linux.go
behzad-mir: Recall @tamilmani1989 mentioned we might have scanrios without InfraNic in the future and I was wondeing if those need any host Route

tamilmani1989: lets not worry about future scenario.. if what @QxBytes proposed cleaner way, lets do it

behzad-mir: addressed it in the new PR

QxBytes (approved):

isaac-dasan (approved):

PR #4405, opened 2026-05-12
fix: clean pre-installed Go before msft-go extraction

jpayne3506:
The tar overlay of msft-go onto the agent's pre-installed Go at /usr/local/go leaves behind source files from both versions. This causes build failures:
- xorBytes redeclared (crypto/internal/fips140/subtle)
- undefined math.MaxUint64 (internal/runtime/strconv)

Remove /usr/local/go before extracting msft-go to ensure a clean installation. Also set GOTOOLCHAIN=local to prevent auto-downloading upstream (non-FIPS) Go, and clear the build cache from any prior Go version on the agent.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/build/scripts/install-go.sh
+++ b/.pipelines/build/scripts/install-go.sh
@@ -49,8 +49,24 @@ fi
 
 ARCH="${ARCH:-amd64}"
 
+# Remove any pre-installed Go to prevent source file contamination.
+# tar extract overlays onto the existing directory without deleting files that
+# are absent from the archive. When the agent's pre-installed Go and msft-go
+# have different source files (e.g. crypto/internal/fips140, internal/runtime),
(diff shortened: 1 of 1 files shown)


rbtr:
Go supports multiple installed versions [now](https://go.dev/doc/manage-install), is there a way to use msft go like that?

rbtr (approved):

jpayne3506:
> Go supports multiple installed versions [now](https://go.dev/doc/manage-install), is there a way to use msft go like that?

I would have to test, but I don't think we can install msft-go with `go install`. Also, I don't think we can do a `mv` similar operation on the existing `go` toolchain, which would result in us having to change additional lines 🙃

mugeshsp (approved):

PR #4411, opened 2026-05-18
fix: wrap handleCommonOptions error with NetworkNotReadyErrorMsg

behzad-mir:
Wrap the error returned from handleCommonOptions in newNetworkImpl with NetworkNotReadyErrorMsg so kubelet treats transient failures (e.g. ENETUNREACH from route installation) as network-not-ready and retries the ADD call in ~1 second.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/network/network_linux.go
+++ b/network/network_linux.go
@@ -116,7 +116,7 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
 	err := nm.handleCommonOptions(ifName, nwInfo)
 	if err != nil {
 		logger.Error("handleCommonOptions failed with", zap.Error(err))
-		return nil, err
+		return nil, errors.Wrap(err, NetworkNotReadyErrorMsg)
 	}
 
 	// Create the network object.


isaac-dasan (approved):

network/network_linux.go
```diff
@@ -116,7 +116,7 @@ func (nm *networkManager) newNetworkImpl(nwInfo *EndpointInfo, extIf *externalIn
 	err := nm.handleCommonOptions(ifName, nwInfo)
 	if err != nil {
 		logger.Error("handleCommonOptions failed with", zap.Error(err))
-		return nil, err
+		return nil, errors.Wrap(err, NetworkNotReadyErrorMsg)
```
QxBytes, line 119: Curious how this particular string in the error message changes the behavior of kubelet?
If there is an ADD failure doesn't it already retry?

Also this error message string seems to be used in the context of swiftv2 mtpnc, rather than the vm's networking which applies to all scenarios (even non-swiftv2)

PR #4412, opened 2026-05-19
feat: Enahancing ETW logging for CNS

apontejaj:
### Summary

Enriches ETW (Event Tracing for Windows) log output with IMDS metadata and contextual fields (hostname, version, API server), and ensures `LogEvent` calls are captured by the zap/ETW pipeline in both the v1 and v2 logger paths.

#### `cns/logger/cnslogger.go`
- Moves the `disableTraceLogging` / `disableEventLogging` guard **before** the zap log call in `Printf`, `Debugf`, `Warnf`, `Errorf`, and `LogEvent`. Previously, zap would still emit even when trace/event logging was disabled.
- Adds a zap `Info` call in `LogEvent` so events are captured in structured logs (with `resource_id` and `properties` fields) in addition to Application Insights.

#### `cns/logger/log.go`
- Adds a `SetZapFields` helper to append persistent zap fields to the global v1 logger's underlying zap instance. This allows host metadata to be attached once and included in all subsequent zap output.

#### `cns/logger/v2/config_windows.go` / `config_linux.go`
- Adds `AppendETWFields` method to the logger v2 `Config` struct. On Windows this appends fields to the ETW core config; on Linux it is a no-op.

#### `cns/logger/v2/cores/etw_windows.go`
- Applies `cfg.Fields` to the ETW core via `core.With(cfg.Fields)` so that IMDS metadata fields are baked into every ETW event.
- Handles the error from `zapetw.New` explicitly instead of relying on the `//nolint:wrapcheck` suppression.

#### `cns/logger/v2/shim.go`
- Implements `LogEvent` on the v2 shim so that AI telemetry events are emitted as structured zap logs (previously was a no-op).

#### `cns/service/main.go`
- Consolidates host metadata fields (`hostname`, `version`, `kubernetes_apiserver`) into a single slice and applies them to both the v1 global logger (via `SetZapFields`) and the v2 logger (via `z.With`).
- Appends IMDS-derived fields to the ETW config via `AppendETWFields` so ETW traces include subscription, resource group, and VM metadata.

**Reason for Change**:

Previously, ETW traces lacked the IMDS metadata and identifying fields (hostname, version, Kubernetes API server) that were available in Application Insights telemetry. This made correlating ETW events to specific nodes/versions difficult. These changes ensure parity between the two telemetry sinks and make `LogEvent` observable in ETW output.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Tests**:
Tested on Service Fabric test cluster with Geneva Logs enabled.
image

### DIFF
--- a/cns/logger/cnslogger.go
+++ b/cns/logger/cnslogger.go
@@ -94,8 +94,11 @@ func (c *logger) SetAPIServer(apiserver string) {
 
 func (c *logger) Printf(format string, args ...any) {
 	c.logger.Logf(format, args...)
+	if c.disableTraceLogging {
+		return
+	}
 	c.zapLogger.Info(fmt.Sprintf(format, args...))
-	if c.th == nil || c.disableTraceLogging {
+	if c.th == nil {
 		return
 	}
 	msg := fmt.Sprintf(format, args...)
@@ -104,8 +107,11 @@ func (c *logger) Printf(format string, args ...any) {
 
 func (c *logger) Debugf(format string, args ...any) {
 	c.logger.Debugf(format, args...)
+	if c.disableTraceLogging {
+		return
+	}
 	c.zapLogger.Debug(fmt.Sprintf(format, args...))
-	if c.th == nil || c.disableTraceLogging {
+	if c.th == nil {
 		return
 	}
 	msg := fmt.Sprintf(format, args...)
@@ -114,8 +120,11 @@ func (c *logger) Debugf(format string, args ...any) {
 
 func (c *logger) Warnf(format string, args ...any) {
 	c.logger.Warnf(format, args...)
+	if c.disableTraceLogging {
+		return
+	}
 	c.zapLogger.Warn(fmt.Sprintf(format, args...))
-	if c.th == nil || c.disableTraceLogging {
+	if c.th == nil {
 		return
 	}
 	msg := fmt.Sprintf(format, args...)
@@ -124,8 +133,11 @@ func (c *logger) Warnf(format string, args ...any) {
 
 func (c *logger) Errorf(format string, args ...any) {
 	c.logger.Errorf(format, args...)
+	if c.disableTraceLogging {
+		return
+	}
 	c.zapLogger.Error(fmt.Sprintf(format, args...))
-	if c.th == nil || c.disableTraceLogging {
+	if c.th == nil {
 		return
 	}
 	msg := fmt.Sprintf(format, args...)
@@ -200,7 +212,11 @@ func (c *logger) sendTraceInternal(msg string, lvl ai.Level) {
 }
 
 func (c *logger) LogEvent(event ai.Event) {
-	if c.th == nil || c.disableEventLogging {
+	if c.disableEventLogging {
+		return
+	}
+	c.zapLogger.Info(event.EventName, zap.String("resource_id", event.ResourceID), zap.Any("properties", event.Properties))
+	if c.th == nil {
 		return
 	}
 	c.m.RLock()
--- a/cns/logger/log.go
+++ b/cns/logger/log.go
@@ -4,6 +4,7 @@ package logger
 import (
 	"github.com/Azure/azure-container-networking/aitelemetry"
 	"github.com/Azure/azure-container-networking/cns/types"
+	"go.uber.org/zap"
 )
 
 type loggershim interface {
@@ -28,6 +29,13 @@ var (
 	AppInsightsIKey = aiMetadata
 	aiMetadata      string // this var is set at build time.
 )
+
+// Deprecated: The global logger is deprecated. Migrate to zap using the cns/logger/v2 package and pass the logger instead.
+func SetZapFields(fields ...zap.Field) {
+	if l, ok := Log.(*logger); ok {
+		l.zapLogger = l.zapLogger.With(fields...)
+	}
+}
 
 // Deprecated: The global logger is deprecated. Migrate to zap using the cns/logger/v2 package and pass the logger instead.
 func Close() {
--- a/cns/logger/v2/config_linux.go
+++ b/cns/logger/v2/config_linux.go
@@ -16,3 +16,5 @@ type Config struct {
 }
 
 func (c *Config) normalize() {}
+
+func (c *Config) AppendETWFields([]zapcore.Field) {}
--- a/cns/logger/v2/cores/etw_windows.go
+++ b/cns/logger/v2/cores/etw_windows.go
@@ -45,5 +45,6 @@ func ETWCore(cfg *ETWConfig) (zapcore.Core, func(), error) {
 	encoderConfig := zap.NewProductionEncoderConfig()
 	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
 	jsonEncoder := zapcore.NewJSONEncoder(encoderConfig)
-	return zapetw.New(cfg.ProviderName, cfg.EventName, &ctrlzap.KubeAwareEncoder{Encoder: jsonEncoder}, cfg.level) //nolint:wrapcheck // ignore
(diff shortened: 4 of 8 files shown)


cns/logger/v2/cores/etw_windows.go
```diff
@@ -45,5 +45,9 @@ func ETWCore(cfg *ETWConfig) (zapcore.Core, func(), error) {
 	encoderConfig := zap.NewProductionEncoderConfig()
 	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
 	jsonEncoder := zapcore.NewJSONEncoder(encoderConfig)
-	return zapetw.New(cfg.ProviderName, cfg.EventName, &ctrlzap.KubeAwareEncoder{Encoder: jsonEncoder}, cfg.level) //nolint:wrapcheck // ignore
+	core, closer, err := zapetw.New(cfg.ProviderName, cfg.EventName, &ctrlzap.KubeAwareEncoder{Encoder: jsonEncoder}, cfg.level)
+	if err != nil {
+		return nil, closer, err
+	}
+	return core.With(cfg.Fields), closer, nil
```
rbtr: core.With is safe on nil cores
https://github.com/apontejaj/azure-container-networking/blob/6bbae181ec1cc05f4bac21fcef1219ce7e07f1db/cns/logger/v2/cores/ai.go#L66

```suggestion
	core, closer, err := zapetw.New(cfg.ProviderName, cfg.EventName, &ctrlzap.KubeAwareEncoder{Encoder: jsonEncoder}, cfg.level)
	return core.With(cfg.Fields), closer, err
```

cns/logger/cnslogger.go
```diff
@@ -200,6 +200,7 @@ func (c *logger) sendTraceInternal(msg string, lvl ai.Level) {
 }
 
 func (c *logger) LogEvent(event ai.Event) {
+	c.zapLogger.Info(event.EventName, zap.String("resource_id", event.ResourceID), zap.Any("properties", event.Properties))
 	if c.th == nil || c.disableEventLogging {
 		return
 	}
```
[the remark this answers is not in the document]
rbtr: copilot is right

cns/service/main.go
```diff
@@ -635,9 +635,22 @@ func main() {
 	// If this errors, we will not have metadata in the AI logs. Should we exit?
 	metadata, _ := acn.GetHostMetadata(aitelemetry.MetadataFile)
 	aifields := loggerv2.MetadataToFields(metadata)
+	host, _ := os.Hostname()
+	loggerFields := []zap.Field{
```
rbtr: loggerFields is not a sufficiently specific name... hostMetadataFields or something like that?

rbtr (dismissed):
couple nits - approved with suggestions

cns/logger/log.go
```diff
@@ -29,6 +30,13 @@ var (
 	aiMetadata      string // this var is set at build time.
 )
 
+// Deprecated: The global logger is deprecated. Migrate to zap using the cns/logger/v2 package and pass the logger instead.
+func SetZapFields(fields ...zap.Field) {
+	if l, ok := Log.(*logger); ok {
+		l.zapLogger = l.zapLogger.With(fields...)
+	}
+}
```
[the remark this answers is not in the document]
apontejaj: Leaving as it is since this only happens as startup.

cns/logger/v2/cores/etw_windows.go
```diff
@@ -45,5 +45,9 @@ func ETWCore(cfg *ETWConfig) (zapcore.Core, func(), error) {
 	encoderConfig := zap.NewProductionEncoderConfig()
 	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
 	jsonEncoder := zapcore.NewJSONEncoder(encoderConfig)
-	return zapetw.New(cfg.ProviderName, cfg.EventName, &ctrlzap.KubeAwareEncoder{Encoder: jsonEncoder}, cfg.level) //nolint:wrapcheck // ignore
+	core, closer, err := zapetw.New(cfg.ProviderName, cfg.EventName, &ctrlzap.KubeAwareEncoder{Encoder: jsonEncoder}, cfg.level)
+	if err != nil {
+		return nil, closer, err
```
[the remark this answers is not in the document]
apontejaj: Implemented.

cns/service/main.go
```diff
@@ -635,9 +635,22 @@ func main() {
...
+		zap.String("hostname", host),
+		zap.String("version", version),
+		zap.String("kubernetes_apiserver", os.Getenv("KUBERNETES_SERVICE_HOST")),
+	}
+
+	// Attach IMDS metadata and logger-level fields to the v1 logger's ETW output.
+	logger.SetZapFields(append(aifields, loggerFields...)...)
```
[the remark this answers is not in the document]
apontejaj: Implemented

cns/logger/cnslogger.go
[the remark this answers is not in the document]
apontejaj: I followed the same approach used in the other functions that implement the zap/ETW emission.
All of them bypass the checks on `c.th`, `c.disableTraceLogging`, `c.disableMetricLogging` and `c.disableEventLogging`.
 
I've implemented the checks for this and pre-existing functions.
 
Can you resolve this comment if your satisfied with the solution/answer?
Otherwise, please, provide further feedback.

cns/logger/v2/cores/etw_windows.go
apontejaj: This is the instantiation of the zapetw core, so I'll leave this as it is.

cns/service/main.go
apontejaj: Implemented.

cns/logger/v2/config_linux.go
```diff
@@ -16,3 +16,5 @@ type Config struct {
 }
 
 func (c *Config) normalize() {}
+
+func (c *Config) AppendETWFields(_ []zapcore.Field) {}
```
rbtr:
```suggestion
func (c *Config) AppendETWFields([]zapcore.Field) {}
```

cns/logger/v2/cores/etw_windows.go
rbtr: I don't understand, if the caller handles the error correctly (which we should expect it to - prefer to over-return and let the caller make the decisions), then my suggestion is equivalent/better because it provides access to the underlying objects and is fewer lines

cns/logger/v2/config_linux.go
apontejaj: Implemented.

cns/logger/v2/cores/etw_windows.go
apontejaj: What you're saying is true fur the App Insights Core. However, it's not true for the ETW core.

Compare the both core factory functions.

**App Insights (as you say, never returns `nil`):**
https://github.com/Azure/azure-container-networking/blob/2e7a01da3fb98ac73c33920de03e9a0229b63662/zapai/core.go#L40

**ETW (returns nil on error):**
https://github.com/Azure/azure-container-networking/blob/2e7a01da3fb98ac73c33920de03e9a0229b63662/zapetw/core_windows.go#L19

Since it is a small thing, I'm implementing a change in the ETW Core factory function so it never returns `nil` and has the same approach as App Insights.

One thing worth noting: if `etw.NewProviderWithOptions` fails and provider is `nil`, calling the returned `closer()` **would panic**. But this should be acceptable since we should expect the caller to handle the error. Also, existing callers `cnslogger_windows.go:22` and `logger_windows.go:29` both check `err` and discard the closer on error.

Could you resolve this comment if your satisfied with the solution/answer?
Otherwise, please, provide further feedback.

rbtr (approved):

estebancams (approved):

PR #4414, opened 2026-05-19
fix: find interface by mac address fix for swiftv2 delegated interfac…

kmurudi:
Reason for Change:
SwiftV2 delegated interface lookup can select the wrong host interface when multiple Linux interfaces share the same MAC address. On accelerated networking nodes, the netvsc interface (for example `eth1`) and the VF interface (for example `enP12217s2`) can both present the delegated MAC. Azure CNI currently returns the first interface match by MAC, which can select the VF instead of the intended upper interface.

Issue Fixed:
This change makes Azure CNI prefer the Linux upper/master interface when resolving delegated SwiftV2 FrontendNICs by MAC.

This addresses the failure seen in ICM where delegated endpoint setup bound to `enP12217s2` and later failed during endpoint configuration with:

`SecondaryEndpointClient Error: route ip+net: no such network interface`

Relevant trace from the incident:
- CNS returned delegated FrontendNIC info with MAC `00:22:48:d5:56:86`
- CNI generated endpoint info with `MasterIfName=enP12217s2`
- endpoint setup then failed with:
  `failed to create endpoint: SecondaryEndpointClient Error: route ip+net: no such network interface`

Requirements:
- On Linux, when multiple interfaces share the same MAC, prefer the interface that is not a lower/VF interface
- Avoid selecting VF interfaces such as `enP...` when the corresponding upper interface such as `eth1` exists
- Keep Windows behavior unchanged, since Windows nodes do not expose the same netvsc/VF interface model

Notes:
- Linux detection uses the `/sys/class/net/<ifname>/master` relationship to distinguish the upper interface from the VF
- If no upper interface is found, the existing matching interface is still used as fallback
- This change is intended to prevent delegated SwiftV2 endpoint creation failures caused by binding to the wrong interface

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -227,11 +227,11 @@ func (plugin *NetPlugin) findInterfaceByMAC(macAddress string) string {
 		}
 		master, err := resolveMasterInterface(iface.Name)
 		if err != nil {
-			logger.Error("failed to resolve master interface",
+			logger.Warn("failed to resolve master interface, falling back to interface name",
 				zap.String("name", iface.Name),
 				zap.String("mac", macAddress),
 				zap.Error(err))
-			return ""
+			return iface.Name
 		}
 		return master
 	}
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -233,6 +233,10 @@ func (plugin *NetPlugin) findInterfaceByMAC(macAddress string) string {
 				zap.Error(err))
 			return ""
 		}
+		logger.Info("found master interface by MAC",
+			zap.String("resolved-master", ifName),
+			zap.String("interface", iface.Name),
+			zap.String("mac", macAddress))
 		return ifName
 	}
 	logger.Error("failed to find interface by MAC", zap.String("macAddress", macAddress), zap.Strings("macs", macs))
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -225,7 +225,18 @@ func (plugin *NetPlugin) findInterfaceByMAC(macAddress string) string {
 		if iface.HardwareAddr.String() != macAddress {
 			continue
 		}
-		if !isInterfaceMaster(iface.Name) {
+		isMaster, err := isInterfaceMaster(iface.Name)
+		if err != nil {
+			logger.Warn("failed to determine interface master relationship",
+				zap.String("name", iface.Name),
+				zap.String("mac", macAddress),
+				zap.Error(err))
+			if fallback == "" {
+				fallback = iface.Name
+			}
+			continue
+		}
+		if !isMaster {
 			logger.Info("skipping non-master interface with matching MAC",
 				zap.String("name", iface.Name), zap.String("mac", macAddress))
 			if fallback == "" {
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -208,9 +208,10 @@ func (plugin *NetPlugin) Stop() {
 	logger.Info("Plugin stopped")
 }
 
-// findInterfaceByMAC returns the name of the upper (master) interface matching the given MAC.
+// findInterfaceByMAC returns the name of the master interface matching the given MAC.
 // With accelerated networking, both the netvsc upper device (e.g. eth1) and the VF
-// (e.g. enP12217s2) share the same MAC. This function prefers the upper device.
+// (e.g. enP12217s2) share the same MAC. When the matched interface is a VF, this
+// function resolves and returns its master.
 func (plugin *NetPlugin) findInterfaceByMAC(macAddress string) string {
 	interfaces, err := plugin.netClient.GetNetworkInterfaces()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


cni/network/network_windows.go
```diff
@@ -31,6 +31,12 @@ var (
 	dualStackCount = 2
 )
 
+// isInterfaceMaster reports whether the named interface is an upper (master) device.
+// Windows does not have the Linux netvsc + VF bonding model, so all interfaces are treated as master.
```
isaac-dasan: Does this mean, there won't be 2 interfaces for master and slave?

kmurudi: windows does not have it, we need to add this for linux

cni/network/network.go
```diff
@@ -208,23 +208,50 @@ func (plugin *NetPlugin) Stop() {
...
+		}
+		return iface.Name
 	}
-	// Failed to find a suitable interface.
-	logger.Error("Failed to find interface by MAC", zap.String("macAddress", macAddress), zap.Strings("interfaces", macs))
+	if fallback != "" {
+		logger.Info("no master interface found, using non-master",
+			zap.String("name", fallback), zap.String("mac", macAddress))
+		return fallback
+	}
```
tamilmani1989: can we change something like this:

if isSlaveInterface(ifName) {
    return getMasterInterface(ifName)
} 

return ifName

tamilmani1989 (changes requested):

cni/network/network.go
kmurudi: addressed

cni/network/network_linux.go
```diff
@@ -19,6 +21,27 @@ const (
...
+	if err != nil {
+		return "", fmt.Errorf("get link %q: %w", name, err)
+	}
+	masterIndex := link.Attrs().MasterIndex
+	if masterIndex == 0 {
+		return name, nil
+	}
+	master, err := vishnetlink.LinkByIndex(masterIndex)
+	if err != nil {
+		return "", fmt.Errorf("get master link by index %d: %w", masterIndex, err)
```
tamilmani1989:
```suggestion
		return "", fmt.Errorf("get master link by index %d failed with: %w", masterIndex, err)
```

cni/network/network_linux.go
```diff
@@ -19,6 +21,27 @@ const (
...
+// non-zero MasterIndex; the upper device does not. If name is already the
+// upper device, it is returned unchanged.
+func resolveMasterInterface(name string) (string, error) {
+	link, err := vishnetlink.LinkByName(name)
+	if err != nil {
+		return "", fmt.Errorf("get link %q: %w", name, err)
+	}
+	masterIndex := link.Attrs().MasterIndex
+	if masterIndex == 0 {
+		return name, nil
```
tamilmani1989: can you try using netlink.NetlinkInterface so that we can add units with mock

kmurudi: Image can I add a interface like this? (will rename) - it will introduce this in NetPlugin - 

Image

kmurudi: and mock for UTs will use this - 

Image

tamilmani1989: no need to add in netplugin..will touch multiple parts of code.. go with 1st option

cni/network/network.go
```diff
@@ -217,14 +220,22 @@ func (plugin *NetPlugin) findInterfaceByMAC(macAddress string) string {
...
+		if mac != macAddress {
+			continue
+		}
+		master, err := resolveMasterInterface(iface.Name)
+		if err != nil {
+			logger.Warn("failed to resolve master interface, falling back to interface name",
+				zap.String("name", iface.Name),
+				zap.String("mac", macAddress),
+				zap.Error(err))
 			return iface.Name
```
tamilmani1989: should return error.. cannot fallback

cni/network/network.go
```diff
@@ -217,14 +220,22 @@ func (plugin *NetPlugin) findInterfaceByMAC(macAddress string) string {
...
 	for _, iface := range interfaces {
-		// find master interface by macAddress for Swiftv2
-		macs = append(macs, iface.HardwareAddr.String())
-		if iface.HardwareAddr.String() == macAddress {
+		mac := iface.HardwareAddr.String()
+		macs = append(macs, mac)
+		if mac != macAddress {
+			continue
+		}
+		master, err := resolveMasterInterface(iface.Name)
```
tamilmani1989:
```suggestion
		ifName, err := resolveMasterInterface(iface.Name)
```

cni/network/network.go
kmurudi: added, checking if all UTs pass

cni/network/network_linux_test.go
```diff
@@ -547,3 +554,242 @@ func TestPluginLinuxAdd(t *testing.T) {
...
+	tests := []struct {
+		name       string
+		ifName     string
+		client     *mockNetlinkClient
+		wantResult string
+		wantErr    bool
+	}{
+		{
+			name:   "interface is already master (MasterIndex == 0)",
+			ifName: snatInterface,
```
tamilmani1989: why snatInterface? can we pass meaningful name instead of snatInterface

kmurudi: @copilot fix this using a variable for "eth1" as it is the original interface name, don't use snatInterface in this test file replace all

tamilmani1989 (approved):

isaac-dasan (approved):

santhoshmprabhu (approved):

PR #4421, opened 2026-05-21
fix: sort epInfos to process InfraNIC first before EndpointCreate

behzad-mir:
Sort epInfos cni/network/network.go) after the createEpInfo loop. In stateless mode, ExternalInterfaces is empty on every ADD so whichever NIC is processed first determines which interface the network gets bound to. If DelegatedVMNIC wins the race, SecondaryEndpointClient moves its interface into the pod namespace, then TransparentEndpointClient fails with "no such network interface".

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net"
 	"regexp"
+	"slices"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cni/network/network_linux_test.go
```diff
@@ -794,3 +795,41 @@ func (m *mockInterfaceGetter) GetNetworkInterfaces() ([]net.Interface, error) {
 func (m *mockInterfaceGetter) GetNetworkInterfaceAddrs(_ *net.Interface) ([]net.Addr, error) {
 	return nil, errNotImplemented
 }
+
+func TestSortEpInfosInfraNICFirst(t *testing.T) {
```
QxBytes: We wanted to test if Add() sorts the endpoint infos right? If you do this you're just testing the copied sorting func works (and if tomorrow someone changes the sort function in the network.go code, the bug won't be caught). I would call Add and have the ipam return several delegated nics and one infra, and ensure that the first call to create (or the processed epInfos) is the infra nic.

cni/network/network.go
```diff
@@ -640,6 +641,24 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
...
 	}()
 
+	// Sort epInfos so InfraNIC is processed first. In stateless mode, ExternalInterfaces is empty
+	// on every ADD, so whichever NIC is processed first determines which interface the network
+	// gets bound to. If DelegatedVMNIC wins the race, the network gets created with eth1 as extIf.
+	// SecondaryEndpointClient then moves eth1 into the pod namespace, and the subsequent InfraNIC
+	// iteration finds the network bound to that now-gone interface, causing
+	// TransparentEndpointClient.AddEndpoints to fail with "no such network interface".
+	slices.SortStableFunc(epInfos, func(a, b *network.EndpointInfo) int {
+		aIsInfra := a.NICType == cns.InfraNIC || a.NICType == ""
```
QxBytes: can we move this to a function-- I feel like the comparison where we check if "" or InfraNIC is getting repeated a lot (ex: in your previous PR we use that conditional as well)

cni/network/network_linux_test.go
behzad-mir: Addressed

cni/network/network.go
behzad-mir: added the helper

QxBytes (approved):

isaac-dasan (approved):

PR #4439, opened 2026-06-02
feat: implemented fallback mechanism for when CNI does not provide an…

estebancams:
HNS team has made changes that allow to create a dual-stack network when an ipv4-only network has already been associated with an interface. Previously this was not possible (although it was possible to create such networks in the reverse order).

This change has a side effect in HNS behavior for when the Adapter Name is not passed as part of the creation payload in a node **with multiple adapters:**

Before the change:
- User creates a ipv4 network
- User creates a dual-stack network but is does not indicate the name of the adapter to which it should be associated
- HNS selects an adapter different from the one associated to the ipv4 network to create the dual-stack network

After the change:
- User creates a ipv4 network
- User creates a dual-stack network but is does not indicate the name of the adapter to which it should be associated
- HNS could select the same adapter as the one associated to the ipv4 network to create the dual-stack network. Current users are not expecting the dual-stack network to be created in the same adapter, so HNS will require explicitely passing the adapter name if the user intention is to associate the dual-stack network with the same ipv4's adapter.

Coincidentally, CNI is passing an empty adapter name when there is any network already created in the node. This is because HNS fails to associate a new network with a virtual (vEthernet) adapter. This happens because the original Ethernet adapter becomes the uplink of the vSwitch where the network is created on and a new network has to be associated with that adapter.

HNS team proposed extending HNS creation API so CNI can pass instead of the adapter name, the IP of the original adapter.

The change in this PR uses that new API when adapter name is going to be passed empty to HNS during network creation. Note the change does not migrate from adapter name to IP APIs, but it only covers the existent gap. Also, this change will continue to be a best effort action and if the IP address cannot be passed to HNS for whatever reason, the call to HNS will still be made regardless (CNI should not fail in scenarios where it previously didn't)


network/endpoint.go
```diff
@@ -106,12 +106,13 @@ type EndpointInfo struct {
...
 	Subnets                       []SubnetInfo
 	BridgeName                    string
-	NetNs                         string // used in windows
+	NetNs                         string                 // used in windows
 	Options                       map[string]interface{} // populated only for InfraNIC; nil for non-infra NICs
 	DisableHairpinOnHostInterface bool
 	IsIPv6Enabled                 bool
 	HostSubnetPrefix              string // can be used later to add an external interface
 	PnPID                         string
+	PrimaryInterfaceIdentifier    string
```
QxBytes: This term is used in other parts of the codebase-- if you are adding a field with the same name here explain what its value is, what it looks like, and when it is used. I assume this is the exact same value as PrimaryInterfaceIdentifier mentioned in the cns codebase?

network/network_windows.go
```diff
@@ -250,13 +252,43 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *EndpointInfo, extIf *exter
...
 			logger.Error("Failed to serialize network adapter policy due to", zap.Error(err))
 			return nil, err
 		}
 
 		hcnNetwork.Policies = append(hcnNetwork.Policies, netAdapterNamePolicy)
+		isNetAdapterNamePolicyAdded = true
+	}
+
+	// Best effort to add IP policy in case adapter name is empty
+	// TODO estebanca improve logger messages with coordinates
```
QxBytes: what are coordinates?

network/network_windows.go
```diff
@@ -250,13 +252,43 @@ func (nm *networkManager) configureHcnNetwork(nwInfo *EndpointInfo, extIf *exter
...
+	if !isNetAdapterNamePolicyAdded {
```
QxBytes: can we make this into an if else with a comment that we bind to an adapter based on either 1) the name or 2) the ip/primary interface identifier-- there are two policies here that control what adapter we bind to right (correct if wrong)?
1) We bind based on net adapter name if it is provided 
2) We bind based on primary interface identifier (IP)
Also what happens if hns doesn't support this new `Type: hcn.ProviderAddress` policy? Can you confirm the code functions as well as before?

network/endpoint.go
QxBytes: Curious... is it possible this is ipv6?

network/policy/policy_windows.go
```diff
@@ -318,6 +318,31 @@ func GetHcnNetAdapterPolicy(networkAdapterName string) (hcn.NetworkPolicy, error
...
+	ProviderAddress string `json:"ProviderAddress"`
+}
+
+func GetHcnNetAdapterAddressPolicy(networkAdapterAddress string) (hcn.NetworkPolicy, error) {
+
+	networkAdapterAddressPolicy := hcn.NetworkPolicy{
+		Type: hcn.ProviderAddress,
+	}
+
+	netAdapterAddressPolicySetting := &ProviderAddressNetworkPolicySetting{
```
QxBytes, line 333: Using `GetHcnNetAdapterNamePolicy` as reference you mention this should be from hcsshim-- are you planning on changing this before merge?

network/network_windows_test.go
```diff
@@ -560,6 +561,60 @@ func TestConfigureHCNNetworkInfraNIC(t *testing.T) {
...
+
+	extIf := externalInterface{
+		Name: "vEthernet (eth0)",
+	}
+
+	nwInfo := &EndpointInfo{
+		NetworkID:                  "d3e97a83-ba4c-45d5-ba88-dc56757ece28",
+		MasterIfName:               "eth0",
+		NICType:                    cns.InfraNIC,
+		PrimaryInterfaceIdentifier: "10.240.0.4/24",
```
QxBytes: can we have a case for if the `primaryInterfaceIdentifier` is just a regular ip? since `// Based on cns/NetworkContainerContract.go, PrimaryInterfaceIdentifier can be either an IP or a CIDR`

QxBytes (commented):
Please fix the linter issues and todos before merge.

Also are you saying that previously on windows dualstack we always only had EXACTLY 2 adapters we could bind to such that the ipv4 would bind to one and the ipv6 would always bind to an adapter that wasn't the first, which just so happened to be the other? If we had more than 2 adapters wouldn't that mean the ipv6 network could bind to a 3rd adapter that was not intended?

Current understanding is that this PR is attempting to use the new hns api policy setting to bind the network to a particular adapter via IP. When will/has this hns change gone in and is it currently breaking? This only affects swiftv1 windows dualstack I assume?

network/endpoint.go
tamilmani1989: lets name it as PrimaryInterfaceIP to be specific

network/policy/policy_windows.go
estebancams: I guess you refer to ProviderAddressNetworkPolicySetting ? If so yes, I commented that on top of the struct. We still don't have hcsshim types from go deps, so I temporarily added it here

network/endpoint.go
estebancams: no, ipv4-only

estebancams: done

network/network_windows.go
estebancams: I rather keep this as a separated if to emphasize this is a fallback mechanism. It is equivalent logically

estebancams: I confirmed also with HNS team that current HNS will not break nor change behavior if we pass the new IP policy

network/network_windows_test.go
estebancams: added

network/network_windows.go
QxBytes: You have `isNetAdapterPolicyAdded` which is set to true if the previous if block succeeded, and then check if the previous block ran and ignore it if so. Is this not what an if else is for? This flag value seems off to me.

estebancams: I agree is equivalent. Seems more indicative of a fallback mechanism imo

network/endpoint.go
```diff
@@ -106,12 +106,13 @@ type EndpointInfo struct {
...
 	BridgeName                    string
-	NetNs                         string // used in windows
+	NetNs                         string                 // used in windows
 	Options                       map[string]interface{} // populated only for InfraNIC; nil for non-infra NICs
 	DisableHairpinOnHostInterface bool
 	IsIPv6Enabled                 bool
 	HostSubnetPrefix              string // can be used later to add an external interface
 	PnPID                         string
+	PrimaryInterfaceIP            string
 }
```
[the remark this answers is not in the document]
estebancams: Name suggested by TMN

cni/network/network_linux_test.go
```diff
@@ -279,7 +279,8 @@ func TestPluginLinuxAdd(t *testing.T) {
 				// should match with GetTestCNSResponse3
 				{
 					epInfo: &network.EndpointInfo{
-						ContainerID: "test-container",
+						PrimaryInterfaceIdentifier: "20.240.0.4/24",
+						ContainerID:                "test-container",
 						Data: map[string]interface{}{
```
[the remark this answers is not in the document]
estebancams: fixed

network/policy/policy_windows.go
```diff
@@ -318,6 +318,31 @@ func GetHcnNetAdapterPolicy(networkAdapterName string) (hcn.NetworkPolicy, error
...
+		ProviderAddress: networkAdapterAddress,
+	}
+
+	netAdapterAddressPolicySettingBytes, err := json.Marshal(netAdapterAddressPolicySetting)
+	if err != nil {
+		return networkAdapterAddressPolicy, errors.Wrap(err, "Error during policy marshalling")
+	}
```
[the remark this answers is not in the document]
estebancams: fixed

network/network_windows.go
QxBytes: please use the if else format here-- the flag variable has risk of desync and adds complexity. you can keep the comment to note that the branch runs as a fallback if the other one does not trigger

network/policy/policy_windows.go
QxBytes: okay should be fine if tamilmani decided this interim temp struct is okay (though when can we expect to replace this with the real hcs type?)

network/endpoint.go
```diff
@@ -106,12 +106,13 @@ type EndpointInfo struct {
...
 	Subnets                       []SubnetInfo
 	BridgeName                    string
-	NetNs                         string // used in windows
+	NetNs                         string                 // used in windows
 	Options                       map[string]interface{} // populated only for InfraNIC; nil for non-infra NICs
 	DisableHairpinOnHostInterface bool
 	IsIPv6Enabled                 bool
 	HostSubnetPrefix              string // can be used later to add an external interface
 	PnPID                         string
+	PrimaryInterfaceIP            string
```
QxBytes, line 115: can we add a comment saying that this is coming from the PrimaryInterfaceIdentifier in cns?

network/policy/policy_windows.go
estebancams: Yes, we'll do when they release the package

network/network_windows.go
estebancams: What do you mean by desync risk?

network/endpoint.go
estebancams: What's the need for it? All these fields are computed in certain way and their values come from other places. In this case I don't see why we need that. We can trace the field origin easily with copilot these days

network/network_windows.go
QxBytes: I mean it's easier for the flag behavior to change accidentally if this were to be modified in the future (ex: `isNetAdapterNamePolicyAdded` is moved or the above block changes its flow slightly to skip that variable being set).

network/endpoint.go
QxBytes: wanted to make it clear that it's the cni counterpart to PrimaryInterfaceIdentifier in cns since the name changed but not that big of a deal, resolving

QxBytes (dismissed):

network/policy/policy_windows.go
```diff
@@ -318,6 +318,32 @@ func GetHcnNetAdapterPolicy(networkAdapterName string) (hcn.NetworkPolicy, error
 	return networkAdapterNamePolicy, nil
 }
 
+// TODO: estebanca: This type must be from hcsshim. HNS team did not provide the updated version of hcsshim.
+// The team decided to lave this type here until we have the new hcsshim version available (6/19/2026)
```
tamilmani1989:
```suggestion
// The team decided to leave this type here until we have the new hcsshim version available (6/19/2026)
```

tamilmani1989: lets add `TODO` prefix.. this should be removed in future after hns team release new package

tamilmani1989 (dismissed):

network/policy/policy_windows.go
estebancams: TODO in previous line

tamilmani1989 (approved):

QxBytes (approved):

PR #4450, opened 2026-06-04
fix: exclude delegated NIC IPs from NNC reconcile on CNS restart

behzad-mir:
On SwiftV2 Windows nodes, PR #4375 started persisting IPv4 addresses for delegated NICs (FrontendNIC/BackendNIC) in the endpoint state. When CNS restarts, endpointStateToPodInfoByIP emits both the InfraNIC and FrontendNIC IPs under the same pod key, causing errMultipleIPPerFamily in newPodKeyToPodIPsMap and a permanent reconcile failure loop.

Fix: filter non-InfraNIC interfaces in endpointStateToPodInfoByIP so only InfraNIC IPs enter NNC reconcile. Add NICType.IsInfraOrLegacy() method for reuse across packages, and refactor the existing isInfraOrLegacyNICType helper in cni/network to use it.

Empty NICType is treated as InfraNIC for backward compatibility with legacy endpoint state that predates the NICType field.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cni/network/network.go
+++ b/cni/network/network.go
@@ -660,21 +660,16 @@ func (plugin *NetPlugin) Add(args *cniSkel.CmdArgs) error {
 // entries come before all other NIC types, preserving relative order among equals.
 func sortInfraNICFirst(epInfos []*network.EndpointInfo) {
 	slices.SortStableFunc(epInfos, func(a, b *network.EndpointInfo) int {
-		if isInfraOrLegacyNICType(a.NICType) && !isInfraOrLegacyNICType(b.NICType) {
+		if a.NICType.IsInfraOrLegacy() && !b.NICType.IsInfraOrLegacy() {
 			return -1
 		}
-		if !isInfraOrLegacyNICType(a.NICType) && isInfraOrLegacyNICType(b.NICType) {
+		if !a.NICType.IsInfraOrLegacy() && b.NICType.IsInfraOrLegacy() {
 			return 1
 		}
 		return 0
 	})
 }
 
-// isInfraOrLegacyNICType returns true if the NIC type is InfraNIC or empty (legacy).
-// Empty NICType is treated as infra for backward compatibility with older CNS responses.
-func isInfraOrLegacyNICType(nicType cns.NICType) bool {
-	return nicType.IsInfraOrLegacy()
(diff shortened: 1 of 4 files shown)


rbtr (dismissed):

cni/network/network.go
```diff
@@ -673,7 +673,7 @@ func sortInfraNICFirst(epInfos []*network.EndpointInfo) {
 // isInfraOrLegacyNICType returns true if the NIC type is InfraNIC or empty (legacy).
 // Empty NICType is treated as infra for backward compatibility with older CNS responses.
 func isInfraOrLegacyNICType(nicType cns.NICType) bool {
-	return nicType == cns.InfraNIC || nicType == ""
+	return nicType.IsInfraOrLegacy()
 }
```
tamilmani1989: can we remove this function and call nicType.IsInfraOrLegacy() directly?

cns/stateprovider/cns/podinfoprovider.go
```diff
@@ -41,6 +41,11 @@ func endpointStateToPodInfoByIP(state map[string]*restserver.EndpointInfo) (map[
 	podInfoByIP := map[string]cns.PodInfo{}
 	for containerID, endpointInfo := range state { // for each endpoint
 		for _, ipinfo := range endpointInfo.IfnameToIPMap { // for each IP info object of the endpoint's interfaces
+			// Only InfraNIC IPs participate in NNC reconcile; delegated NIC IPs
+			// (FrontendNIC/BackendNIC) are owned by per-pod MTPNC.
+			if !ipinfo.NICType.IsInfraOrLegacy() {
```
tamilmani1989, line 46: should we also stop saving delegated nic ip to state file?

behzad-mir: that is needed per Jacki's change

cni/network/network.go
behzad-mir: done

tamilmani1989 (approved):

rbtr (approved):

PR #4456, opened 2026-06-09
CNS Changes to Support PrefixOnNic allocations with DRA

ravisraju:
**Reason for Change**:

feat: Support for shared delegated nic allocations with DRA

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -13,6 +13,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 )
 
 // Container Network Service DNC Contract
@@ -34,7 +35,7 @@ const (
 	DetachContainerFromNetwork               = "/network/detachcontainerfromnetwork"
 	RequestIPConfig                          = "/network/requestipconfig"
 	RequestIPConfigs                         = "/network/requestipconfigs"
-	RequestIPConfigsAndNICResources          = "/network/requestipconfigsandnicresources"
+	RequestClaimConfig                       = "/network/requestclaimconfig"
 	ReleaseIPConfig                          = "/network/releaseipconfig"
 	ReleaseIPConfigs                         = "/network/releaseipconfigs"
 	PathDebugIPAddresses                     = "/debug/ipaddresses"
@@ -583,12 +584,19 @@ type IPConfigsResponse struct {
 	Response  Response    `json:"response"`
 }
 
-// IPConfigsAndNICResourcesResponse is the response for the RequestIPConfigsAndNICResources API.
+// ClaimConfigRequest is the request for the RequestClaimConfig API. ClaimUID identifies
+// the DRA ResourceClaim; CNS resolves it to the owning pod by finding the pod's MTPNC
+// whose Spec.ResourceClaims contains this claim.
+type ClaimConfigRequest struct {
+	ClaimUID k8stypes.UID `json:"claimUID"`
+}
+
+// ClaimConfigResponse is the response for the RequestClaimConfig API.
 // It returns the pod's IP configs (the same set RequestIPConfigs produces, but
 // WITHOUT the IsScheduledWithDRA filtering, so delegated NIC configs are always
 // included) together with the resource-slice properties (networkID/subnetGUID/
 // subnetName/capacity/...) of every NIC allocated to the pod.
-type IPConfigsAndNICResourcesResponse struct {
+type ClaimConfigResponse struct {
 	Response     Response      `json:"response"`
 	PodIPInfo    []PodIpInfo   `json:"podIPInfo"`
 	NICResources []NICResource `json:"nicResources"`
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -21,6 +21,7 @@ const (
 	GetHomeAz                                = "/homeaz"
 	GetNCList                                = "/nclist"
 	GetVMUniqueID                            = "/metadata/vmuniqueid"
+	GetNICResources                          = "/network/nicresources"
 	CreateOrUpdateNetworkContainer           = "/network/createorupdatenetworkcontainer"
 	DeleteNetworkContainer                   = "/network/deletenetworkcontainer"
 	PublishNetworkContainer                  = "/network/publishnetworkcontainer"
@@ -33,6 +34,7 @@ const (
 	DetachContainerFromNetwork               = "/network/detachcontainerfromnetwork"
 	RequestIPConfig                          = "/network/requestipconfig"
 	RequestIPConfigs                         = "/network/requestipconfigs"
+	RequestIPConfigsAndNICResources          = "/network/requestipconfigsandnicresources"
 	ReleaseIPConfig                          = "/network/releaseipconfig"
 	ReleaseIPConfigs                         = "/network/releaseipconfigs"
 	PathDebugIPAddresses                     = "/debug/ipaddresses"
@@ -525,6 +527,8 @@ type PodIpInfo struct {
 	InterfaceName              string
 	// MacAddress of interface
 	MacAddress string
+	// SharedNIC indicates the delegated NIC is shared and should not be managed by CNI.
+	SharedNIC bool `json:"sharedNic,omitempty"`
 	// SkipDefaultRoutes is true if default routes should not be added on interface
 	SkipDefaultRoutes bool
 	// Routes to configure on interface
@@ -577,6 +581,17 @@ type IPConfigResponse struct {
 type IPConfigsResponse struct {
 	PodIPInfo []PodIpInfo `json:"podIPInfo"`
 	Response  Response    `json:"response"`
+}
+
+// IPConfigsAndNICResourcesResponse is the response for the RequestIPConfigsAndNICResources API.
+// It returns the pod's IP configs (the same set RequestIPConfigs produces, but
+// WITHOUT the IsScheduledWithDRA filtering, so delegated NIC configs are always
+// included) together with the resource-slice properties (networkID/subnetGUID/
+// subnetName/capacity/...) of every NIC allocated to the pod.
+type IPConfigsAndNICResourcesResponse struct {
+	Response     Response      `json:"response"`
+	PodIPInfo    []PodIpInfo   `json:"podIPInfo"`
+	NICResources []NICResource `json:"nicResources"`
 }
 
 // GetIPAddressesRequest is used in CNS IPAM mode to get the states of IPConfigs
--- a/cns/NetworkContainerContract.go
+++ b/cns/NetworkContainerContract.go
@@ -35,7 +35,7 @@ const (
 	DetachContainerFromNetwork               = "/network/detachcontainerfromnetwork"
 	RequestIPConfig                          = "/network/requestipconfig"
 	RequestIPConfigs                         = "/network/requestipconfigs"
-	RequestClaimConfig                       = "/network/requestclaimconfig"
+	RequestClaimResourceInfo                 = "/network/requestclaimresourceinfo"
 	ReleaseIPConfig                          = "/network/releaseipconfig"
 	ReleaseIPConfigs                         = "/network/releaseipconfigs"
 	PathDebugIPAddresses                     = "/debug/ipaddresses"
@@ -584,19 +584,19 @@ type IPConfigsResponse struct {
 	Response  Response    `json:"response"`
 }
 
-// ClaimConfigRequest is the request for the RequestClaimConfig API. ClaimUID identifies
+// ClaimResourceInfoRequest is the request for the RequestClaimResourceInfo API. ClaimUID identifies
 // the DRA ResourceClaim; CNS resolves it to the owning pod by finding the pod's MTPNC
 // whose Spec.ResourceClaims contains this claim.
-type ClaimConfigRequest struct {
+type ClaimResourceInfoRequest struct {
 	ClaimUID k8stypes.UID `json:"claimUID"`
 }
 
-// ClaimConfigResponse is the response for the RequestClaimConfig API.
+// ClaimResourceInfoResponse is the response for the RequestClaimResourceInfo API.
 // It returns the pod's IP configs (the same set RequestIPConfigs produces, but
 // WITHOUT the IsScheduledWithDRA filtering, so delegated NIC configs are always
 // included) together with the resource-slice properties (networkID/subnetGUID/
 // subnetName/capacity/...) of every NIC allocated to the pod.
-type ClaimConfigResponse struct {
+type ClaimResourceInfoResponse struct {
 	Response     Response      `json:"response"`
 	PodIPInfo    []PodIpInfo   `json:"podIPInfo"`
 	NICResources []NICResource `json:"nicResources"`
--- a/cns/kubecontroller/nicnetworkconfig/reconciler.go
+++ b/cns/kubecontroller/nicnetworkconfig/reconciler.go
@@ -6,8 +6,6 @@ import (
 	"github.com/Azure/azure-container-networking/crd/multitenancy/api/v1alpha1"
 	"github.com/pkg/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -15,24 +13,9 @@ import (
 // purpose is to start and keep the controller-runtime informer running so the
 // manager cache stays warm for NICNetworkConfig objects, which the
 // K8sSWIFTv2Middleware reads via the cached client.
-//
-// The predicate scopes reconcile events to NICNetworkConfigs whose Spec.NodeName
-// matches this node. The reconciler is a noop today, so this has no functional
-// effect yet; it keeps the controller scoped to this node's objects for when
-// reconcile logic is added. Note it filters only reconcile events, not what the
-// informer watches/caches (the middleware still node-filters its cached reads).
-func SetupWithManager(mgr ctrl.Manager, nodeName string) error {
+func SetupWithManager(mgr ctrl.Manager) error {
 	err := ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.NICNetworkConfig{}).
-		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
-			nicnc, ok := object.(*v1alpha1.NICNetworkConfig)
-			if !ok {
-				return false
-			}
-			// Filter to only reconcile NICNetworkConfigs for this node.
-			// TODO - If we set owner references on NICNetworkConfigs to the Node CRD, we could use that instead of filtering by node name.
-			return nicnc.Spec.NodeName == nodeName
-		})).
 		Complete(reconcile.Func(func(context.Context, ctrl.Request) (ctrl.Result, error) { return ctrl.Result{}, nil }))
 	return errors.Wrap(err, "failed to set up nicnetworkconfig reconciler")
 }
--- a/cns/kubecontroller/nicnetworkconfig/reconciler.go
+++ b/cns/kubecontroller/nicnetworkconfig/reconciler.go
@@ -6,6 +6,8 @@ import (
 	"github.com/Azure/azure-container-networking/crd/multitenancy/api/v1alpha1"
 	"github.com/pkg/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -13,9 +15,24 @@ import (
 // purpose is to start and keep the controller-runtime informer running so the
 // manager cache stays warm for NICNetworkConfig objects, which the
 // K8sSWIFTv2Middleware reads via the cached client.
-func SetupWithManager(mgr ctrl.Manager) error {
+//
+// The predicate scopes reconcile events to NICNetworkConfigs whose Spec.NodeName
+// matches this node. The reconciler is a noop today, so this has no functional
+// effect yet; it keeps the controller scoped to this node's objects for when
+// reconcile logic is added. Note it filters only reconcile events, not what the
+// informer watches/caches (the middleware still node-filters its cached reads).
+func SetupWithManager(mgr ctrl.Manager, nodeName string) error {
 	err := ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.NICNetworkConfig{}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
+			nicnc, ok := object.(*v1alpha1.NICNetworkConfig)
+			if !ok {
+				return false
+			}
+			// Filter to only reconcile NICNetworkConfigs for this node.
+			// TODO - If we set owner references on NICNetworkConfigs to the Node CRD, we could use that instead of filtering by node name.
+			return nicnc.Spec.NodeName == nodeName
+		})).
 		Complete(reconcile.Func(func(context.Context, ctrl.Request) (ctrl.Result, error) { return ctrl.Result{}, nil }))
 	return errors.Wrap(err, "failed to set up nicnetworkconfig reconciler")
 }
--- a/cns/kubecontroller/nicnetworkconfig/reconciler.go
+++ b/cns/kubecontroller/nicnetworkconfig/reconciler.go
@@ -1,0 +1,21 @@
+package nicnetworkconfig
+
+import (
+	"context"
+
+	"github.com/Azure/azure-container-networking/crd/multitenancy/api/v1alpha1"
+	"github.com/pkg/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// SetupWithManager registers a noop NICNetworkConfig reconciler. Its only
+// purpose is to start and keep the controller-runtime informer running so the
+// manager cache stays warm for NICNetworkConfig objects, which the
+// K8sSWIFTv2Middleware reads via the cached client.
+func SetupWithManager(mgr ctrl.Manager) error {
+	err := ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.NICNetworkConfig{}).
+		Complete(reconcile.Func(func(context.Context, ctrl.Request) (ctrl.Result, error) { return ctrl.Result{}, nil }))
+	return errors.Wrap(err, "failed to set up nicnetworkconfig reconciler")
+}
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -32,6 +32,7 @@ var (
 	errInvalidSWIFTv2NICType    = errors.New("invalid NIC type for SWIFT v2 scenario")
 	errInvalidMTPNCPrefixLength = errors.New("invalid prefix length for MTPNC primaryIP, must be 32")
 	errMTPNCDeleting            = errors.New(NetworkNotReadyErrorMsg + " - mtpnc for previous pod is being deleted, waiting for new mtpnc to be ready")
+	errClaimNotFound            = errors.New(NetworkNotReadyErrorMsg + " - no mtpnc found for resource claim")
 )
 
 type K8sSWIFTv2Middleware struct {
@@ -256,6 +257,32 @@ func (k *K8sSWIFTv2Middleware) getMTPNC(ctx context.Context, podInfo cns.PodInfo
 	return mtpnc, types.Success, ""
 }
 
+// GetPodInfoByClaimUID finds the MTPNC on this node whose Spec.ResourceClaims contains
+// claimUID and returns PodInfo for the pod that owns it. It is used by RequestClaimConfig
+// to resolve a DRA ResourceClaim to its pod, and is scoped to this node's MTPNCs.
+func (k *K8sSWIFTv2Middleware) GetPodInfoByClaimUID(ctx context.Context, claimUID k8stypes.UID) (cns.PodInfo, types.ResponseCode, string) {
+	var mtpncList v1alpha1.MultitenantPodNetworkConfigList
+	if err := k.Cli.List(ctx, &mtpncList); err != nil {
+		return nil, types.UnexpectedError, errors.Wrap(err, "failed to list mtpncs").Error()
+	}
+	for i := range mtpncList.Items {
+		mtpnc := &mtpncList.Items[i]
+		// Only consider MTPNCs scheduled on this node.
+		if mtpnc.Status.NodeName != k.NodeName {
+			continue
+		}
+		for _, claim := range mtpnc.Spec.ResourceClaims {
+			if claim == string(claimUID) {
+				if mtpnc.IsDeleting() {
+					return nil, types.UnexpectedError, errMTPNCDeleting.Error()
+				}
+				return cns.NewPodInfo("", "", mtpnc.Spec.PodName, mtpnc.Namespace), types.Success, ""
+			}
+		}
+	}
+	return nil, types.UnexpectedError, errClaimNotFound.Error()
+}
+
 // Updates Ip Config Request
 func (k *K8sSWIFTv2Middleware) UpdateIPConfigRequest(mtpnc v1alpha1.MultitenantPodNetworkConfig, req *cns.IPConfigsRequest) (
 	respCode types.ResponseCode,
@@ -386,7 +413,7 @@ const dedicatedNICDRACapacity = 1
 func (k *K8sSWIFTv2Middleware) GetNICNCInfoByMAC(ctx context.Context) (map[string]*cns.NICResourceSliceInfo, error) {
 	var nicNCList v1alpha1.NICNetworkConfigList
 	if err := k.Cli.List(ctx, &nicNCList); err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "failed to list nicnetworkconfigs")
 	}
 
 	result := make(map[string]*cns.NICResourceSliceInfo, len(nicNCList.Items))
@@ -429,7 +456,7 @@ func (k *K8sSWIFTv2Middleware) GetNICNCInfoByMAC(ctx context.Context) (map[strin
 func (k *K8sSWIFTv2Middleware) GetMTPNCInfoByMAC(ctx context.Context) (map[string]*cns.NICResourceSliceInfo, error) {
 	var mtpncList v1alpha1.MultitenantPodNetworkConfigList
 	if err := k.Cli.List(ctx, &mtpncList); err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "failed to list mtpncs")
 	}
 
 	result := make(map[string]*cns.NICResourceSliceInfo)
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -2,6 +2,8 @@ package middlewares
 
 import (
 	"context"
+	"net"
+	"strings"
 
 	"github.com/Azure/azure-container-networking/cns"
 	"github.com/Azure/azure-container-networking/cns/configuration"
@@ -10,6 +12,7 @@ import (
 	"github.com/Azure/azure-container-networking/cns/types"
 	"github.com/Azure/azure-container-networking/crd/multitenancy/api/v1alpha1"
 	"github.com/pkg/errors"
+	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 	k8stypes "k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -32,7 +35,18 @@ var (
 )
 
 type K8sSWIFTv2Middleware struct {
-	Cli client.Client
+	Cli      client.Client
+	NodeName string
+	Logger   *zap.Logger
+}
+
+// log returns the middleware's logger, or a no-op logger when one was not set
+// (e.g. in unit tests that construct the middleware directly).
+func (k *K8sSWIFTv2Middleware) log() *zap.Logger {
+	if k.Logger != nil {
+		return k.Logger
+	}
+	return zap.NewNop()
 }
 
 // Verify interface compliance at compile time
@@ -72,6 +86,13 @@ func (k *K8sSWIFTv2Middleware) GetPodInfoForIPConfigsRequest(ctx context.Context
 
 // getIPConfig returns the pod's SWIFT V2 IP configuration.
 func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodInfo) ([]cns.PodIpInfo, error) {
+	return k.getIPConfigHelper(ctx, podInfo, false)
+}
+
+// getIPConfigHelper builds the pod's SWIFT V2 delegated IP configs from its MTPNC.
+// When includeDRAAllocations is false, pods scheduled with DRA are skipped.
+// when true, DRA-delegated NICs are included.
+func (k *K8sSWIFTv2Middleware) getIPConfigHelper(ctx context.Context, podInfo cns.PodInfo, includeDRAAllocations bool) ([]cns.PodIpInfo, error) {
 	// Check if the MTPNC CRD exists for the pod, if not, return error
 	mtpnc := v1alpha1.MultitenantPodNetworkConfig{}
 	mtpncNamespacedName := k8stypes.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
@@ -84,6 +105,16 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
 		return nil, errMTPNCNotReady
 	}
 	logger.Printf("[SWIFTv2Middleware] mtpnc for pod %s is : %+v", podInfo.Name(), mtpnc)
+
+	// When the pod was scheduled with Dynamic Resource Allocation (DRA), dranet
+	// owns the delegated NIC via NRI and configures it out of band. CNS must not
+	// return the delegated NIC IP config to the CNI invoker, so skip it here unless the
+	// caller explicitly asks to include DRA allocations.
+	// A swiftv2 pod is either scheduled with DRA using ResourceClaims or using the legacy PN/PNI annotation, regardless of how many NICs/Networks it connects to.
+	if !includeDRAAllocations && mtpnc.IsScheduledWithDRA() {
+		k.log().Debug("pod scheduled with DRA; skipping delegated NIC IP config for CNI", zap.String("pod", podInfo.Name()))
+		return []cns.PodIpInfo{}, nil
+	}
 
 	var podIPInfos []cns.PodIpInfo
 
@@ -140,6 +171,7 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
 					},
 					MacAddress:        interfaceInfo.MacAddress,
 					NICType:           nicType,
+					SharedNIC:         interfaceInfo.SharedNIC,
 					SkipDefaultRoutes: false,
 					// InterfaceName is empty for DelegatedVMNIC and AccelnetFrontendNIC
 				}
@@ -157,6 +189,27 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
 	}
 
 	return podIPInfos, nil
+}
+
+// GetAllIPConfigs returns the pod's SWIFT V2 delegated IP configs INCLUDING DRA-scheduled
+// NICs (unlike getIPConfig, which skips them).
+func (k *K8sSWIFTv2Middleware) GetAllIPConfigs(ctx context.Context, podInfo cns.PodInfo) ([]cns.PodIpInfo, error) {
+	return k.getIPConfigHelper(ctx, podInfo, true)
+}
+
+// GetPodNICMACs returns the MAC addresses of the NICs allocated to the pod, read
+// from its MultitenantPodNetworkConfig. It is self-contained and does not share code
+// with the existing getIPConfig path.
+func (k *K8sSWIFTv2Middleware) GetPodNICMACs(ctx context.Context, podInfo cns.PodInfo) ([]string, error) {
+	mtpnc := v1alpha1.MultitenantPodNetworkConfig{}
+	mtpncNamespacedName := k8stypes.NamespacedName{Namespace: podInfo.Namespace(), Name: podInfo.Name()}
+	if err := k.Cli.Get(ctx, mtpncNamespacedName, &mtpnc); err != nil {
+		return nil, errors.Wrap(err, errGetMTPNC.Error())
+	}
+	if !mtpnc.IsReady() {
+		return nil, errMTPNCNotReady
+	}
+	return mtpncMACs(&mtpnc), nil
 }
 
 func (k *K8sSWIFTv2Middleware) Type() cns.SWIFTV2Mode {
@@ -277,4 +330,138 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
 	v6Cidrs = append(v6Cidrs, serviceCIDRsV6...)
 
 	return v4Cidrs, v6Cidrs, nil
+}
+
+// mtpncMACs returns the NIC MAC addresses referenced by an MTPNC, taken from
+// Status.InterfaceInfos and falling back to the deprecated Status.MacAddress.
+func mtpncMACs(mtpnc *v1alpha1.MultitenantPodNetworkConfig) []string {
+	var macs []string
+	for i := range mtpnc.Status.InterfaceInfos {
+		if mac := mtpnc.Status.InterfaceInfos[i].MacAddress; mac != "" {
+			macs = append(macs, mac)
+		}
+	}
+	if len(macs) == 0 && mtpnc.Status.MacAddress != "" {
+		macs = append(macs, mtpnc.Status.MacAddress)
+	}
+	return macs
+}
+
+// subnetNameFromResourceID extracts the trailing subnet name from an ARM resource ID.
+// e.g. ".../subnets/mySubnet" → "mySubnet"
+func subnetNameFromResourceID(resourceID string) string {
+	if resourceID == "" {
+		return ""
+	}
+	parts := strings.Split(resourceID, "/")
+	if len(parts) < 2 {
+		return resourceID
+	}
+	return parts[len(parts)-1]
+}
+
+// canonicalMAC parses raw and returns its canonical (lowercase, colon-separated)
+// form so map keys built from different CRDs compare equal regardless of the
+// original formatting. Returns false if raw is empty or not a valid MAC address.
+func canonicalMAC(raw string) (string, bool) {
+	hw, err := net.ParseMAC(raw)
+	if err != nil {
+		return "", false
+	}
+	return hw.String(), true
+}
+
+// sharedNICDRACapacity is the resource-slice capacity advertised for a DRA-managed
+// prefix-on-NIC (shared) NIC: the scheduler can place up to this many pods on it.
+const sharedNICDRACapacity = 16
+
+// dedicatedNICDRACapacity is the resource-slice capacity advertised for a dedicated
+// (single-allocation) NIC scheduled with DRA: the scheduler tallies one resource claim
+// against the slice.
+const dedicatedNICDRACapacity = 1
+
+// GetNICNCInfoByMAC lists the NICNetworkConfigs on this node and returns a map keyed by
+// canonical NIC MAC address with the NIC's network/subnet info and resource-slice
+// capacity from Spec.
+func (k *K8sSWIFTv2Middleware) GetNICNCInfoByMAC(ctx context.Context) (map[string]*cns.NICResourceSliceInfo, error) {
+	var nicNCList v1alpha1.NICNetworkConfigList
+	if err := k.Cli.List(ctx, &nicNCList); err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]*cns.NICResourceSliceInfo, len(nicNCList.Items))
+	for i := range nicNCList.Items {
+		spec := &nicNCList.Items[i].Spec
+		// Only consider NICs on this node.
+		if spec.NodeName != k.NodeName {
+			continue
+		}
+		key, ok := canonicalMAC(spec.MACAddress)
+		if !ok {
+			continue
+		}
+
+		// A DRA-managed NIC advertises the shared capacity; otherwise it has none. Presence of NICNC indicates that the NIC is created on a PN with PrefixBlock Allocation.
+		// If its not scheduled with DRA, it is not shared and has no capacity to advertise to scheduler. we will mark it as 0 so the scheduler will not try to put further pods on this NIC.
+		capacity := 0
+		if spec.ScheduledByDRA {
+			capacity = sharedNICDRACapacity
+		}
+		result[key] = &cns.NICResourceSliceInfo{
+			NetworkID:  spec.NetworkID,
+			SubnetGUID: spec.SubnetGUID,
+			SubnetName: subnetNameFromResourceID(spec.SubnetResourceID),
+			Capacity:   capacity,
+		}
+	}
+
+	return result, nil
+}
+
+// GetMTPNCInfoByMAC lists the MTPNCs scheduled on this node and returns a map keyed by
+// canonical NIC MAC address with the NIC's network/subnet info and resource-slice
+// capacity from the MTPNC Spec. Dedicated NICs (single-allocation PodNetworks) usually
+// have no NICNetworkConfig and are served from here as a fallback.
+//
+// It intentionally does NOT wait for the MTPNC to be Ready: a not-yet-ready MTPNC may
+// have empty Spec NetworkID/SubnetGUID/SubnetResourceID, which are surfaced as-is (empty)
+// rather than treated as an error. Entries whose MAC is empty or unparseable are skipped.
+func (k *K8sSWIFTv2Middleware) GetMTPNCInfoByMAC(ctx context.Context) (map[string]*cns.NICResourceSliceInfo, error) {
+	var mtpncList v1alpha1.MultitenantPodNetworkConfigList
+	if err := k.Cli.List(ctx, &mtpncList); err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]*cns.NICResourceSliceInfo)
+	for i := range mtpncList.Items {
+		mtpnc := &mtpncList.Items[i]
+		// Only consider MTPNCs scheduled on this node.
+		if mtpnc.Status.NodeName != k.NodeName {
+			continue
+		}
+
+		// A dedicated NIC scheduled with DRA advertises the dedicated capacity; a NIC
+		// created without DRA has no capacity to advertise to the scheduler (marked 0).
+		capacity := 0
+		if mtpnc.IsScheduledWithDRA() {
+			capacity = dedicatedNICDRACapacity
+		}
+		info := &cns.NICResourceSliceInfo{
+			NetworkID:  mtpnc.Spec.NetworkID,
+			SubnetGUID: mtpnc.Spec.SubnetGUID,
+			SubnetName: subnetNameFromResourceID(mtpnc.Spec.SubnetResourceID),
+			Capacity:   capacity,
+		}
+		for _, mac := range mtpncMACs(mtpnc) {
+			key, ok := canonicalMAC(mac)
+			if !ok {
+				continue
+			}
+			if _, exists := result[key]; !exists { // first-seen wins
+				result[key] = info
+			}
+		}
+	}
+
+	return result, nil
 }
--- a/cns/middlewares/k8sSwiftV2.go
+++ b/cns/middlewares/k8sSwiftV2.go
@@ -32,7 +32,7 @@ var (
 	errInvalidSWIFTv2NICType    = errors.New("invalid NIC type for SWIFT v2 scenario")
 	errInvalidMTPNCPrefixLength = errors.New("invalid prefix length for MTPNC primaryIP, must be 32")
 	errMTPNCDeleting            = errors.New(NetworkNotReadyErrorMsg + " - mtpnc for previous pod is being deleted, waiting for new mtpnc to be ready")
-	errClaimNotFound            = errors.New(NetworkNotReadyErrorMsg + " - no mtpnc found for resource claim")
+	errMTPNCNotFoundForClaim    = errors.New(NetworkNotReadyErrorMsg + " - no mtpnc found for resource claim")
 )
 
 type K8sSWIFTv2Middleware struct {
@@ -281,7 +281,7 @@ func (k *K8sSWIFTv2Middleware) GetPodInfoByClaimUID(ctx context.Context, claimUI
 			}
 		}
 	}
-	return nil, types.UnexpectedError, errClaimNotFound.Error()
+	return nil, types.UnexpectedError, errMTPNCNotFoundForClaim.Error()
 }
 
 // Updates Ip Config Request
@@ -408,16 +408,16 @@ const sharedNICDRACapacity = 16
 // against the slice.
 const dedicatedNICDRACapacity = 1
 
-// GetNICResourceSliceInfoByMAC lists the NICNetworkConfigs on this node and returns a map keyed by
+// GetNICResourceNetworkInfoFromNICNC lists the NICNetworkConfigs on this node and returns a map keyed by
 // canonical NIC MAC address with the NIC's network/subnet info and resource-slice
 // capacity from Spec.
-func (k *K8sSWIFTv2Middleware) GetNICResourceSliceInfoByMAC(ctx context.Context) (map[string]*cns.NICResourceSliceInfo, error) {
+func (k *K8sSWIFTv2Middleware) GetNICResourceNetworkInfoFromNICNC(ctx context.Context) (map[string]*cns.NICResourceNetworkInfo, error) {
 	var nicNCList v1alpha1.NICNetworkConfigList
 	if err := k.Cli.List(ctx, &nicNCList); err != nil {
 		return nil, errors.Wrap(err, "failed to list nicnetworkconfigs")
 	}
 
-	result := make(map[string]*cns.NICResourceSliceInfo, len(nicNCList.Items))
+	result := make(map[string]*cns.NICResourceNetworkInfo, len(nicNCList.Items))
 	for i := range nicNCList.Items {
 		spec := &nicNCList.Items[i].Spec
 		// Only consider NICs on this node.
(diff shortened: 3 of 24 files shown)


cni/network/invoker_cns.go
```diff
@@ -227,6 +231,13 @@ func (invoker *CNSIPAMInvoker) Add(addConfig IPAMAddConfig) (IPAMAddResult, erro
 		key := invoker.getInterfaceInfoKey(info.nicType, info.macAddress)
 		switch info.nicType {
 		case cns.NodeNetworkInterfaceFrontendNIC:
+			if info.sharedNIC {
```
shreyashastantram: It should be no-op for all interfaces managed by dranet

cns/middlewares/k8sSwiftV2.go
```diff
@@ -85,6 +87,12 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
 	}
 	logger.Printf("[SWIFTv2Middleware] mtpnc for pod %s is : %+v", podInfo.Name(), mtpnc)
 
+	// When the pod was scheduled with Dynamic Resource Allocation (DRA), dranet
+	// owns the NIC via NRI. CNS must still publish the IP config but mark each
+	// frontend NIC as SharedNIC so the CNI invoker skips it and leaves the
+	// NIC for dranet to configure.
+	scheduledWithDRA := mtpnc.IsScheduledWithDRA()
```
shreyashastantram: if it's scheduled with DRA, do we need to return this information back to CNI ? we can ignore at CNI right

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -52,6 +52,24 @@ type MultitenantPodNetworkConfigSpec struct {
 	IBMACAddresses []string `json:"IBMACAddresses,omitempty"`
```
kmurudi, line 52: remove all CRD changes here since we have different PRs

cni/network/invoker_cns.go
```diff
@@ -59,6 +59,7 @@ type IPResultInfo struct {
 	hostGateway          string
 	nicType              cns.NICType
 	macAddress           string
+	sharedNIC            bool
```
miguelgoms: why do we need to change the CNI/invoker_cns code, return extra information and filter client side? 
if we have a separate endpoint for secondary nics on DRA, this should be purely a server-side (CNS) change.

cns/middlewares/k8sSwiftV2.go
miguelgoms: +1

cns/multitenancy-rbac.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: azure-cns-multitenancy
+rules:
+- apiGroups: ["multitenancy.acn.azure.com"]
+  resources: ["nodeinfo", "nicnetworkconfigs", "multitenantpodnetworkconfigs"]
```
miguelgoms: why is this new role needed, if cns already reads nodeinfo/mtpnc?

miguelgoms (changes requested):

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
ravisraju: will be rebased away once those PR gets in. until then its needed here for testing purposes.

cni/network/invoker_cns.go
ravisraju: removing

cns/multitenancy-rbac.yaml
ravisraju: adding in aks-rp. removing this change

cns/restserver/api.go
```diff
@@ -1314,6 +1315,195 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
...
 }
 
+func (service *HTTPRestService) getNICResources(w http.ResponseWriter, r *http.Request) {
+	logger.Request(service.Name, "getNICResources", nil)
+	ctx := r.Context()
+
+	switch r.Method {
+	case http.MethodGet:
+		// Step 1: Get list of NICs (MACs) from NodeInfo CRD.
+		logger.Printf("[Azure CNS] getNICResources: fetching NodeInfo CRD for node %s", service.nodeName)
```
ravisraju: too many log lines throughout this function. needs to be cleaned up since we are out of poc mode and dont need as much logging here. also check for lint errors when you introduce logger.printf / errorf statements

cns/restserver/api.go
```diff
@@ -1314,6 +1315,195 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
...
+			logger.Printf("[Azure CNS] getNICResources: listed %d host network interfaces", len(ifaces))
+			for _, iface := range ifaces {
+				if len(iface.HardwareAddr) == 0 {
+					continue
+				}
+				key := iface.HardwareAddr.String() // already normalized
+				if res, ok := nicByMAC[key]; ok {
+					// On Azure VMs with Accelerated Networking, a NIC appears as both a
+					// synthetic netvsc interface (e.g. eth1) and an SR-IOV VF interface
+					// (e.g. enP56082s2). Prefer the synthetic name because it is stable
```
ravisraju: comment needs to reflect what we ended up doing. we dont prefer ifNames with "eth" prefix. we only consider them and if we dont have a name in that format, we just dont set InterfaceName

cns/restserver/api.go
```diff
@@ -1314,6 +1315,195 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
...
+			}
+			rawMAC := res.MacAddress
+			normalizedKey := normalizedMAC.String()
+
+			// Try NICNetworkConfig first (lookup by both raw and normalized key).
+			if info := lookupNICNCInfo(nicNCByMAC, rawMAC, normalizedKey); info != nil {
+				res.NetworkID = info.NetworkID
+				res.SubnetID = info.SubnetID
+				res.PrimaryIP = info.PrimaryIP
+				res.Capacity = 16
```
ravisraju: 16 if ScheduledByDRA is true. if ScheduledByDRA is false, set this to 0.

cns/restserver/api.go
```diff
@@ -1314,6 +1315,195 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
...
+				logger.Errorf("[Azure CNS] getNICResources: failed to fetch NICNetworkConfig data: %v", err)
+			} else {
+				logger.Printf("[Azure CNS] getNICResources: fetched %d NICNetworkConfig entries", len(nicNCByMAC))
+			}
+		} else {
+			logger.Printf("[Azure CNS] getNICResources: nicNCClient is nil, skipping NICNetworkConfig")
+		}
+
+		// Step 4: Populate NetworkID, SubnetID, and Capacity from MTPNC CRD (fallback).
+		var mtpncByMAC map[string]*cns.NICNCInfo
```
ravisraju: need better variable name as this is not really a map of MAC -> MTPNC, just that MTPNC is source of truth here.

cns/restserver/api.go
```diff
@@ -1314,6 +1315,195 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
...
+
+			// Fall back to MTPNC.
+			if info := lookupNICNCInfo(mtpncByMAC, rawMAC, normalizedKey); info != nil {
+				res.NetworkID = info.NetworkID
+				res.SubnetID = info.SubnetID
+				// A NIC reported via MTPNC fallback is bound to a pod. If that
+				// pod was not scheduled with DRA the NIC is dedicated to a
+				// non-DRA workload and must not be advertised as having
+				// capacity for additional DRA pods.
+				if info.ScheduledWithDRA {
```
ravisraju: loigically incorrect. shared pods will also have MTPNC with ScheduledWithDRA to be true. 

when we lookup NICNCInfo by MTPNC MAC, we should no-op if that MAC info is already populated from NICNC. we only care about dedicated NICs (using Single allocation mode PN) that are assigned with or without DRA. With DRA, set capacity to 1, since scheduler witll tally up the number of resourceclaims against this slice. if IsScheduledWithDRA is false, set it to 0, since that pod is created without resourceclaims. but do not override information we have collected from NICNC

cns/restserver/api.go
```diff
@@ -1314,6 +1315,195 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
...
+				res.Capacity = 1
+				continue
+			}
+			rawMAC := res.MacAddress
+			normalizedKey := normalizedMAC.String()
+
+			// Try NICNetworkConfig first (lookup by both raw and normalized key).
+			if info := lookupNICNCInfo(nicNCByMAC, rawMAC, normalizedKey); info != nil {
+				res.NetworkID = info.NetworkID
+				res.SubnetID = info.SubnetID
```
ravisraju: SubnetID is not a valid field anymore. replace it with SubnetGUID, look up recent CRDs and verify all column validities. 

We also have SubnetResourceID from which we should extract the SubnetName and set it in the resourceslice. 

also these fields have moved from Status to Spec so do a whole review with recent CRDs to make sure you dont miss anything.

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +289,213 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+		return nil, err
+	}
+
+	result := make(map[string]*cns.NICNCInfo, len(nicNCList.Items))
+	for i := range nicNCList.Items {
+		mac := nicNCList.Items[i].Status.MACAddress
+		if mac == "" {
+			continue
+		}
+		// Build the host-underlay primary IP as a CIDR using the *subnet*
```
ravisraju: we are not going to be using / assigning primaryIP for the NIC itself in the final POC so remove all this code and also PrimaryIP from the GetNICResources response. we do not need the buildSubnetWidth... function either.

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +289,213 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+		// Status.SubnetAddressSpace (e.g., "165.0.0.0/20" → /20).
+		//
+		// Result: PrimaryIP = "165.0.0.16/20" — kernel-ready for `ip addr add`.
+		primaryIP := buildSubnetWidthPrimaryIP(
+			nicNCList.Items[i].Status.PrimaryIP,
+			nicNCList.Items[i].Status.SubnetAddressSpace,
+		)
+		result[mac] = &cns.NICNCInfo{
+			NetworkID: nicNCList.Items[i].Spec.NetworkID,
+			SubnetID:  nicNCList.Items[i].Spec.SubnetGUID,
```
ravisraju: rename the field in contract to SubnetID, add a new field SubnetName and populate it with the name that you extract from NICNC.Spec.SubnetResourceId - check latest crd in this branch for consistency

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +289,213 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+			PrimaryIP: primaryIP,
+		}
+	}
+
+	logger.Printf("[SWIFTv2Middleware] fetched %d NICNetworkConfigs, %d with MAC addresses", len(nicNCList.Items), len(result))
+	return result, nil
+}
+
+// buildSubnetWidthPrimaryIP combines the IP portion of Status.PrimaryIP
```
ravisraju: goes away

cns/middlewares/k8sSwiftV2.go
ravisraju: we should consider only NICNC in this node similar to what was done in MTPNC. every cns can scope its task to the NICs and Pods that are on its node.

ravisraju: any reason to cast a wider net here?

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +289,213 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+	pnCache := make(map[string]*v1alpha1.PodNetwork)
+	for i := range mtpncList.Items {
+		mtpnc := &mtpncList.Items[i]
+
+		// Only consider MTPNCs that are scheduled on this node.
+		if mtpnc.Status.NodeName != k.NodeName {
+			continue
+		}
+
+		pnNames, err := k.resolveMTPNCPodNetworkNames(ctx, mtpnc)
```
ravisraju: dont need these anymore, NetworkID, SubnetGUID and SubnetResourceID are all part of MTPNC Spec. remove the code resolveMTPNCPodNetworkNames, 

pn := k.podNetworkForInterface(ctx, pnNames, j, pnCache)

function nicNCInfoFromPodNetwork(pn)

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +289,213 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+			info.ScheduledWithDRA = scheduledWithDRA
+			result[mtpnc.Status.MacAddress] = info
+		}
+	}
+
+	logger.Printf("[SWIFTv2Middleware] fetched %d MTPNCs on node %s, %d unique MACs with info", len(mtpncList.Items), k.NodeName, len(result))
+	return result, nil
+}
+
+func (k *K8sSWIFTv2Middleware) resolveMTPNCPodNetworkNames(ctx context.Context, mtpnc *v1alpha1.MultitenantPodNetworkConfig) ([]string, error) {
```
ravisraju: goes away. use MTPNC spec instead

cns/restserver/api.go
```diff
@@ -1314,6 +1316,206 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
...
+			}
+		}
+
+		// Step 5: Enrich each NIC discovered from NodeInfo, preferring NICNetworkConfig
+		// and falling back to MTPNC. The map key is the canonical MAC used for lookups.
+		for mac, res := range nicByMAC {
+			enrichNICFromCRDs(res, mac, nicNCInfoByMAC, mtpncInfoByMAC)
+		}
+
+		// Step 6: Add NICs known only to MTPNC (not present in NodeInfo). MTPNC keys are
```
ravisraju: dont need this since its already handled as part of enrichNICFromCRDs

cns/kubecontroller/nicnetworkconfig/reconciler.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// SetupWithManager registers a noop NICNetworkConfig reconciler. Its only
+// purpose is to start and keep the controller-runtime informer running so the
+// manager cache stays warm for NICNetworkConfig objects, which the
+// K8sSWIFTv2Middleware reads via the cached client.
+func SetupWithManager(mgr ctrl.Manager) error {
+	err := ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.NICNetworkConfig{}).
+		Complete(reconcile.Func(func(context.Context, ctrl.Request) (ctrl.Result, error) { return ctrl.Result{}, nil }))
```
miguelgoms, line 19: This should have predicates for filtering - do we need to observe all NicNC objects and all operations in the cluster?

cns/restserver/api.go
```diff
@@ -1314,6 +1316,153 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
 	}
 }
 
+func (service *HTTPRestService) getNICResources(w http.ResponseWriter, r *http.Request) {
```
miguelgoms: can we move all this business logic to its own nicresources.go file? 
there is enough logic here that is more than api-handling / response construction, and this part of nic-resources is evolving and I could see us optimizing differenlty (e.g.: caching this map) without touching this api surface.

miguelgoms: for example, we could have a consumer for NicResources device info w/o IPAM information (e.g.: if we followed the byodp pattern in the future). I wouldn't refactor prematurely but if we encapsulate this it'll be easier to evolve / reuse.

miguelgoms (changes requested):

cns/service/main.go
```diff
@@ -1506,6 +1512,26 @@ func InitializeCRDState(ctx context.Context, z *zap.Logger, httpRestService cns.
...
+
+		// MTPNC is namespaced and per-pod; its node identity lives in status.nodeName,
+		// which DNC-RC writes atomically with the interface data CNS reads (and only
+		// after the pod is scheduled). Scope the informer to this node's MTPNCs. A not-
+		// yet-ready MTPNC has no status.nodeName and is simply absent from cache, but
+		// every CNS reader already rejects a not-ready MTPNC, so this only turns a
+		// "found-but-not-ready" read into a "not-found" one - both retryable.
+		cacheOpts.ByObject[&mtv1alpha1.MultitenantPodNetworkConfig{}] = cache.ByObject{
+			Field: fields.SelectorFromSet(fields.Set{"status.nodeName": nodeName}),
+		}
```
miguelgoms: can we double-check our assumptions on how this behaves is not defined as 'selectable' on the CRD? 
there are some subtleties with selectable fields on CRDs. depending on the above, we may need to split this scale fix into separate PR.

tldr; 
a) Selectable fields on CRDs is only enabled by default on 1.32+ (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#crd-selectable-fields)
b) Above is fine for enabling DRA scenarios (1.36+) and 1.8.x CNS/CNI support (1.35+)
   and *should* be fine for the earliest supported CNS/CNI for swiftv2 (1.7.x ->1.33)
c) *but* currently CRDs are installed centrally from the latest definition and considered backward compatible.
so as long as we have LTS versions w/o supporting this feature we need to make sure CRD install is fine. 

I had pullrequest w/ change accounting for that as part of updating all CRDs on https://github.com/Azure/azure-container-networking/pull/4377 - intended to resurrect that only after DRA changes merged.

cns/restserver/restserver.go
```diff
@@ -102,6 +119,10 @@ type HTTPRestService struct {
 	PnpIDByMacAddress          map[string]string
 	imdsClient                 imdsClient
 	nodesubnetIPFetcher        *nodesubnet.IPFetcher
+	nicNCClient                nicNCClient
+	mtpncCli                   mtpncClient
```
ashvindeodhar: nit; should be Client instead of Cli for naming consistency for both mtpnc and nodeInfo

cns/restserver/restserver.go
```diff
@@ -57,6 +58,22 @@ type imdsClient interface {
 	GetIMDSVersions(ctx context.Context) (*imds.APIVersionsResponse, error)
 }
 
+// nicNCClient enriches NICResource data with NICNetworkConfig CRD information (e.g., NetworkID, SubnetGUID).
+type nicNCClient interface {
+	GetNICNCInfoByMAC(ctx context.Context) (map[string]*cns.NICResourceSliceInfo, error)
+}
+
+// mtpncClient enriches NICResource data with MTPNC CRD information for dedicated NICs,
+// which usually have no NICNetworkConfig.
```
ashvindeodhar: nit: drop 'usually' - as it will never have NICNC.

cns/restserver/restserver.go
```diff
@@ -102,6 +119,10 @@ type HTTPRestService struct {
 	PnpIDByMacAddress          map[string]string
 	imdsClient                 imdsClient
 	nodesubnetIPFetcher        *nodesubnet.IPFetcher
+	nicNCClient                nicNCClient
```
ashvindeodhar: nit naming casing: inline with mtpncClient, should we name this to nicncClient?

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
+// Copyright 2017 Microsoft. All rights reserved.
```
ashvindeodhar: nit: copyright year incorrect.

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+				},
+			}
+			respondJSON(w, http.StatusInternalServerError, resp)
+			return
+		}
+
+		// Step 1: Get list of NICs (MACs) from NodeInfo CRD.
+		nodeInfo, err := service.nodeInfoCli.Get(ctx, service.nodeName)
+		if err != nil {
+			l.Error("failed to fetch NodeInfo CRD", zap.String("node", service.nodeName), zap.Error(err))
```
ashvindeodhar: is it conclusive that we failed to fetch the CRD or can it be extracting the info from the fetched CRD?

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		vmUniqueID := nodeInfo.Spec.VMUniqueID
+
+		// Initialize map keyed by normalized MAC for every device in NodeInfo.
+		nicByMAC := make(map[string]*cns.NICResource, len(nodeInfo.Status.DeviceInfos))
+		for _, device := range nodeInfo.Status.DeviceInfos {
+			normalizedMAC, parseErr := net.ParseMAC(device.MacAddress)
+			if parseErr != nil {
+				l.Warn("failed to parse MAC from NodeInfo", zap.String("mac", device.MacAddress), zap.Error(parseErr))
+				continue
```
ashvindeodhar, line 91: how can we bubble up this error / alert on this? This may get silently ignored with just the warning

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+			nicByMAC[key] = &cns.NICResource{
+				MacAddress: device.MacAddress,
+				VMUniqueID: vmUniqueID,
+			}
+		}
+
+		// Step 2: Populate InterfaceName from host network interfaces.
+		ifaces, ifErr := net.Interfaces()
+		if ifErr != nil {
+			l.Warn("failed to list host network interfaces", zap.Error(ifErr))
```
ashvindeodhar: if ifErr is non-nil, we won't assign InterfaceName in nicByMAC but will proceed further in the code. Is that ok?

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+					if strings.HasPrefix(iface.Name, "eth") {
+						res.InterfaceName = iface.Name
+					}
+				}
+			}
+		}
+
+		// Step 3: Fetch per-NIC network/subnet/capacity from NICNetworkConfig (shared
+		// prefix-on-NIC NICs).
+		var nicNCInfoByMAC map[string]*cns.NICResourceSliceInfo
```
ashvindeodhar: nit: same comment as above about nicncInfoByMAC

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+		if service.nicNCClient != nil {
+			nicNCInfoByMAC, err = service.nicNCClient.GetNICNCInfoByMAC(ctx)
+			if err != nil {
+				l.Warn("failed to fetch NICNetworkConfig data", zap.Error(err))
+			}
+		}
+
+		// Step 4: Fetch per-NIC data from MTPNC. Dedicated NICs usually have no
```
ashvindeodhar: nit: drop 'usually'

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+		returnCode := types.UnsupportedVerb
+		service.setResponse(w, returnCode, cns.GetNICResourcesResponse{
+			Response: cns.Response{ReturnCode: returnCode, Message: returnMessage},
+		})
+	}
+}
+
+// enrichNICResource populates res.NetworkID/SubnetGUID/SubnetName/Capacity for a NIC,
+// looked up by canonical MAC. NICNetworkConfig is preferred (shared prefix-on-NIC NICs);
+// MTPNC is the fallback for dedicated NICs, which usually have no NICNetworkConfig. MTPNC
```
ashvindeodhar: nit: drop usually

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+// never overrides NICNetworkConfig, and a NIC found in neither keeps zero capacity.
+func enrichNICResource(res *cns.NICResource, mac string, nicNCInfoByMAC, mtpncInfoByMAC map[string]*cns.NICResourceSliceInfo) {
+	info := nicNCInfoByMAC[mac]
+	if info == nil {
+		info = mtpncInfoByMAC[mac]
+	}
+	if info == nil {
+		res.Capacity = "0"
```
ashvindeodhar: will we hit this line of the code when that NIC has no MTPNC / NICNC associated yet ( free nic )? In that case, shouldn't we set the capacity to 1 for placeholder?
The capacity 0 is when the mtpnc associated with the nic is created without DRA, no?

cns/restserver/nicresources_test.go
```diff
@@ -0,0 +1,168 @@
...
+			{MacAddress: testMAC03}, // MTPNC dedicated DRA           → 1
+			{MacAddress: testMAC04}, // MTPNC dedicated non-DRA       → 0
+			{MacAddress: testMAC05}, // no NICNetworkConfig/MTPNC     → 0
+			{MacAddress: testMAC07}, // NICNC via canonical MAC match → 16
+			{MacAddress: testMAC08}, // in NICNC and MTPNC; NICNC wins → 16
+		}},
+	}
+	nicNC := map[string]*cns.NICResourceSliceInfo{
+		testMAC01: {NetworkID: testNet1, SubnetGUID: testGUID1, SubnetName: testSubnet1, Capacity: 16},
+		testMAC02: {NetworkID: testNet2, SubnetGUID: "guid2", SubnetName: "subnet2", Capacity: 0},
```
ashvindeodhar: why do we have one set of global const such as guid1, subnet1 while test adds local values? can we remove global consts?

cns/restserver/nicresources_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+type fakeMTPNCClient struct {
+	m map[string]*cns.NICResourceSliceInfo
+}
+
+func (f *fakeMTPNCClient) GetMTPNCInfoByMAC(context.Context) (map[string]*cns.NICResourceSliceInfo, error) {
+	return f.m, nil
+}
+
+func TestGetNICResources(t *testing.T) {
```
ashvindeodhar, line 60: I highly recommend rewriting this in the table-driven test form for better readability. copilot should help. Posting the screenshot to show better readability with it -

Image

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (service *HTTPRestService) getNICResources(w http.ResponseWriter, r *http.Request) {
+	l := service.Logger
+	if l == nil {
+		l = zap.NewNop()
+	}
+	ctx := r.Context()
+
+	switch r.Method {
+	case http.MethodGet:
```
ashvindeodhar: in requestClaimConfig, you have used `if r.Method != http.MethodXX {` to error out early. I like that over switch case as it removes the extra indentation for majority of this func.

cns/restserver/requestclaimconfig.go
```diff
@@ -0,0 +1,127 @@
...
+	var req cns.ClaimConfigRequest
+	if err := common.Decode(w, r, &req); err != nil {
+		l.Error("failed to decode request", zap.Error(err))
+		return
+	}
+
+	mw, ok := service.IPConfigsHandlerMiddleware.(swiftV2NICMiddleware)
+	if !ok {
+		respondJSON(w, http.StatusServiceUnavailable, cns.ClaimConfigResponse{
+			Response: cns.Response{ReturnCode: types.UnexpectedError, Message: "requestClaimConfig: SWIFT v2 middleware is not configured"},
```
ashvindeodhar: Swift v2 'NIC' middleware?

cns/restserver/requestclaimconfig.go
```diff
@@ -0,0 +1,127 @@
...
+		})
+		return
+	}
+
+	// Resolve the DRA ResourceClaim to the pod that owns it via the pod's MTPNC.
+	podInfo, respCode, message := mw.GetPodInfoByClaimUID(ctx, req.ClaimUID)
+	if respCode != types.Success {
+		respondJSON(w, http.StatusBadRequest, cns.ClaimConfigResponse{
```
ashvindeodhar: why is it a bad request? can it be internal error?

cns/restserver/requestclaimconfig.go
```diff
@@ -0,0 +1,127 @@
...
+			Response: cns.Response{ReturnCode: respCode, Message: message},
+		})
+		return
+	}
+
+	// 1) IP configs: the pod's SWIFT v2 delegated configs, including DRA allocations.
+	podIPInfo, err := mw.GetAllIPConfigs(ctx, podInfo)
+	if err != nil {
+		l.Error("failed to get SWIFT v2 IP configs", zap.Error(err))
```
ashvindeodhar: should we include the pod info in the error log for debugging?

cns/restserver/requestclaimconfig.go
```diff
@@ -0,0 +1,127 @@
...
+		respondJSON(w, http.StatusInternalServerError, cns.ClaimConfigResponse{
+			Response: cns.Response{ReturnCode: types.FailedToAllocateIPConfig, Message: errors.Wrap(err, "failed to get SWIFT v2 IP configs").Error()},
+		})
+		return
+	}
+
+	// 2) NIC resources: resource-slice properties for every NIC allocated to the pod.
+	nicResources, err := service.podNICResources(ctx, l, mw, podInfo)
+	if err != nil {
+		l.Error("failed to get pod nic resources", zap.Error(err))
```
ashvindeodhar: same comment as above

cns/restserver/requestclaimconfig.go
```diff
@@ -0,0 +1,127 @@
...
+	GetAllIPConfigs(ctx context.Context, podInfo cns.PodInfo) ([]cns.PodIpInfo, error)
+	GetPodNICMACs(ctx context.Context, podInfo cns.PodInfo) ([]string, error)
+}
+
+// requestClaimConfig is the handler for the RequestClaimConfig API. It
+// returns, in a single response, the pod's IP configs (the same set RequestIPConfigs
+// produces, but WITHOUT the IsScheduledWithDRA filtering, so delegated NIC configs
+// are always included) plus the resource-slice properties of every NIC allocated to
+// the pod.
+func (service *HTTPRestService) requestClaimConfig(w http.ResponseWriter, r *http.Request) {
```
ashvindeodhar: what's the rationale behind naming this as requestClaim'Config'? IMO ClaimConfig would be like a spec / input to the DRA but looks like you are returning more of a response / claimAllocationResult here

cns/restserver/requestclaimconfig.go
```diff
@@ -0,0 +1,127 @@
...
+// GetNICResources. The pod's NIC MACs come from its MTPNC; the per-NIC network/subnet/
+// capacity comes from NICNetworkConfig, falling back to MTPNC for dedicated NICs.
+func (service *HTTPRestService) podNICResources(ctx context.Context, l *zap.Logger, mw swiftV2NICMiddleware, podInfo cns.PodInfo) ([]cns.NICResource, error) {
+	macs, err := mw.GetPodNICMACs(ctx, podInfo)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get pod NIC MACs")
+	}
+
+	var nicNCInfoByMAC, mtpncInfoByMAC map[string]*cns.NICResourceSliceInfo
+	if service.nicNCClient != nil {
```
ashvindeodhar: can we always intialize these clients with CNS and drop these checks from multiple places in the code?

cns/restserver/requestclaimconfig.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	nicResources := make([]cns.NICResource, 0, len(macs))
+	for _, mac := range macs {
+		res := cns.NICResource{MacAddress: mac}
+		key := mac
+		if hw, parseErr := net.ParseMAC(mac); parseErr == nil {
+			key = hw.String()
+		}
+		enrichNICResource(&res, key, nicNCInfoByMAC, mtpncInfoByMAC)
```
ashvindeodhar: looks like you change the result ( res ) into slice in both the places this enrichNICResource is called.  Why not just return slice?

cns/restserver/requestclaimresourceinfo_test.go
```diff
@@ -0,0 +1,128 @@
...
+	return nil, nil
+}
+
+func (f *fakeSwiftV2NICMiddleware) GetPodNICMACs(context.Context, cns.PodInfo) ([]string, error) {
+	return f.macs, nil
+}
+
+// podNICResources enriches each of the pod's NICs from NICNetworkConfig, falling back to
+// MTPNC for dedicated NICs; the pod's NIC MACs come from its MTPNC via the middleware.
+func TestPodNICResources(t *testing.T) {
```
ashvindeodhar, line 35: can we rewrite this as table-driven test pls?

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+			Response: cns.Response{ReturnCode: returnCode, Message: returnMessage},
+		})
+	}
+}
+
+// enrichNICResource populates res.NetworkID/SubnetGUID/SubnetName/Capacity for a NIC,
+// looked up by canonical MAC. NICNetworkConfig is preferred (shared prefix-on-NIC NICs);
+// MTPNC is the fallback for dedicated NICs, which usually have no NICNetworkConfig. MTPNC
+// never overrides NICNetworkConfig, and a NIC found in neither keeps zero capacity.
+func enrichNICResource(res *cns.NICResource, mac string, nicNCInfoByMAC, mtpncInfoByMAC map[string]*cns.NICResourceSliceInfo) {
```
ashvindeodhar: the func name mislead me in thinking 'enrich' implying decorating already complete object with extra / optional data. The func is actually populating NIC resource. populateNICResourceInfo?

ashvindeodhar (changes requested):
Posting comments after reviewing restserver review. Will continue with other code

cns/restserver/nicresources.go
ashvindeodhar: is res supposed to be NICResourceSliceInfo or NICResource?

cns/NetworkContainerContract.go
```diff
@@ -525,6 +528,8 @@ type PodIpInfo struct {
 	InterfaceName              string
 	// MacAddress of interface
 	MacAddress string
+	// SharedNIC indicates the delegated NIC is shared and should not be managed by CNI.
```
ashvindeodhar, line 531: don't we want any delegated nic to be managed by cni, including dedicated delegated nic?

cns/middlewares/k8sSwiftV2.go
```diff
@@ -29,10 +32,22 @@ var (
 	errInvalidSWIFTv2NICType    = errors.New("invalid NIC type for SWIFT v2 scenario")
 	errInvalidMTPNCPrefixLength = errors.New("invalid prefix length for MTPNC primaryIP, must be 32")
 	errMTPNCDeleting            = errors.New(NetworkNotReadyErrorMsg + " - mtpnc for previous pod is being deleted, waiting for new mtpnc to be ready")
+	errClaimNotFound            = errors.New(NetworkNotReadyErrorMsg + " - no mtpnc found for resource claim")
```
ashvindeodhar: errClaimNotFound or errMtpncNotFoundForClaim?

cns/middlewares/k8sSwiftV2.go
```diff
@@ -85,6 +107,16 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
 	}
 	logger.Printf("[SWIFTv2Middleware] mtpnc for pod %s is : %+v", podInfo.Name(), mtpnc)
 
+	// When the pod was scheduled with Dynamic Resource Allocation (DRA), dranet
+	// owns the delegated NIC via NRI and configures it out of band. CNS must not
+	// return the delegated NIC IP config to the CNI invoker, so skip it here unless the
+	// caller explicitly asks to include DRA allocations.
+	// A swiftv2 pod is either scheduled with DRA using ResourceClaims or using the legacy PN/PNI annotation, regardless of how many NICs/Networks it connects to.
+	if !includeDRAAllocations && mtpnc.IsScheduledWithDRA() {
```
ashvindeodhar, line 114: do we need `includeDRAAllocations`? can we rely on just `mtpnc.IsScheduledWithDRA()`?

cns/middlewares/k8sSwiftV2.go
```diff
@@ -159,6 +192,27 @@ func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodI
 	return podIPInfos, nil
 }
 
+// GetAllIPConfigs returns the pod's SWIFT V2 delegated IP configs INCLUDING DRA-scheduled
+// NICs (unlike getIPConfig, which skips them).
+func (k *K8sSWIFTv2Middleware) GetAllIPConfigs(ctx context.Context, podInfo cns.PodInfo) ([]cns.PodIpInfo, error) {
+	return k.getIPConfigHelper(ctx, podInfo, true)
```
ashvindeodhar: trying to understand this... where will we have mix of dra and non-dra allocated nics for the same pod?

cns/middlewares/k8sSwiftV2.go
```diff
@@ -203,6 +257,32 @@ func (k *K8sSWIFTv2Middleware) getMTPNC(ctx context.Context, podInfo cns.PodInfo
...
+// to resolve a DRA ResourceClaim to its pod, and is scoped to this node's MTPNCs.
+func (k *K8sSWIFTv2Middleware) GetPodInfoByClaimUID(ctx context.Context, claimUID k8stypes.UID) (cns.PodInfo, types.ResponseCode, string) {
+	var mtpncList v1alpha1.MultitenantPodNetworkConfigList
+	if err := k.Cli.List(ctx, &mtpncList); err != nil {
+		return nil, types.UnexpectedError, errors.Wrap(err, "failed to list mtpncs").Error()
+	}
+	for i := range mtpncList.Items {
+		mtpnc := &mtpncList.Items[i]
+		// Only consider MTPNCs scheduled on this node.
+		if mtpnc.Status.NodeName != k.NodeName {
```
ashvindeodhar, line 271: should we add a todo here to remove this when we'll have the code to watch the mtpnc on the current node only?

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +358,137 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+	var nicNCList v1alpha1.NICNetworkConfigList
+	if err := k.Cli.List(ctx, &nicNCList); err != nil {
+		return nil, errors.Wrap(err, "failed to list nicnetworkconfigs")
+	}
+
+	result := make(map[string]*cns.NICResourceSliceInfo, len(nicNCList.Items))
+	for i := range nicNCList.Items {
+		spec := &nicNCList.Items[i].Spec
+		// Only consider NICs on this node.
+		if spec.NodeName != k.NodeName {
```
ashvindeodhar, line 433: same todo comment as above

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +358,137 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+		}
+		key, ok := canonicalMAC(spec.MACAddress)
+		if !ok {
+			continue
+		}
+
+		// A DRA-managed NIC advertises the shared capacity; otherwise it has none. Presence of NICNC indicates that the NIC is created on a PN with PrefixBlock Allocation.
+		// If its not scheduled with DRA, it is not shared and has no capacity to advertise to scheduler. we will mark it as 0 so the scheduler will not try to put further pods on this NIC.
+		capacity := 0
+		if spec.ScheduledByDRA {
```
ashvindeodhar, line 447: isn't this guaranteed since nicnc will only be created with DRA?

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +358,137 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+
+// dedicatedNICDRACapacity is the resource-slice capacity advertised for a dedicated
+// (single-allocation) NIC scheduled with DRA: the scheduler tallies one resource claim
+// against the slice.
+const dedicatedNICDRACapacity = 1
+
+// GetNICNCInfoByMAC lists the NICNetworkConfigs on this node and returns a map keyed by
+// canonical NIC MAC address with the NIC's network/subnet info and resource-slice
+// capacity from Spec.
+func (k *K8sSWIFTv2Middleware) GetNICNCInfoByMAC(ctx context.Context) (map[string]*cns.NICResourceSliceInfo, error) {
```
ashvindeodhar: is it possible to rename the name? imo it's a bit misleading - I thought this was returning a single NICNC info for a given mac. This looks like it's extracting NICResourceSliceInfo from all existing NICNCs

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +358,137 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+		// Only consider MTPNCs scheduled on this node.
+		if mtpnc.Status.NodeName != k.NodeName {
+			continue
+		}
+
+		// A dedicated NIC scheduled with DRA advertises the dedicated capacity; a NIC
+		// created without DRA has no capacity to advertise to the scheduler (marked 0).
+		capacity := 0
+		if mtpnc.IsScheduledWithDRA() {
+			capacity = dedicatedNICDRACapacity
```
ashvindeodhar, line 484: if this mtpnc has a corresponding nicnc, the capacity should be a shared capacity right?

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +358,137 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+			SubnetGUID: mtpnc.Spec.SubnetGUID,
+			SubnetName: subnetNameFromResourceID(mtpnc.Spec.SubnetResourceID),
+			Capacity:   capacity,
+		}
+		for _, mac := range mtpncMACs(mtpnc) {
+			key, ok := canonicalMAC(mac)
+			if !ok {
+				continue
+			}
+			if _, exists := result[key]; !exists { // first-seen wins
```
ashvindeodhar, line 498: can we short circuit the main loop of the result already contains the mac?

cns/middlewares/k8sSwiftV2_test.go
```diff
@@ -0,0 +1,197 @@
...
+const (
+	testNode       = "node1"
+	testSubnetName = "mySubnet"
+	canonMAC01     = "aa:bb:cc:dd:ee:01"
+	canonMAC02     = "aa:bb:cc:dd:ee:02"
+	canonMAC03     = "aa:bb:cc:dd:ee:03"
+)
+
+// GetMTPNCInfoByMAC must node-scope, compute capacity from DRA state, and tolerate a
+// not-ready MTPNC (empty Spec network/subnet) without erroring — empty values flow
```
ashvindeodhar: can we write this as table-driven tests?

cns/restserver/nicresources.go
```diff
@@ -0,0 +1,165 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/Azure/azure-container-networking/cns"
+	"github.com/Azure/azure-container-networking/cns/types"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+)
+
+func (service *HTTPRestService) getNICResources(w http.ResponseWriter, r *http.Request) {
```
ashvindeodhar, line 20: instead of always constructing NIC snapshots from NICNC and MTPNC, why not maintains a snapshot of NIC resources in the CNS cache that we just return when dranet driver calls this?
Whenever a NICNC / MTPNC is created / deleted, NIC resource snapshot can be updated in the cache.

ashvindeodhar (changes requested):

cns/kubecontroller/nicnetworkconfig/reconciler.go
ravisraju: since nicnc is a new crd, we will modify it to have selectable fields and use that to filter what is cached. reconciler here would only have visibility into what is cached.

cns/middlewares/k8sSwiftV2.go
ravisraju: NICNC will not be created only through DRA. we discussed the rollback scenario where if customer turns off DRA and still allocates with PrefixBlock PN, we will create a NICNC to hold the 16 ips. but mark the resourceslice with capacity 0

cns/middlewares/k8sSwiftV2.go
ravisraju: its both depending on how you look at it. DRANET is asking CNS webhook for IpConfig of a Claim. If CNS has not seen a MTPNC with that Claim it has no record of the MTPNC or the claim. we can modify the error message to say something like "no ip config found for the claim" if we dont want to expose mtpnc (internal) to dranet

cns/middlewares/k8sSwiftV2.go
ravisraju: We cannot. if we just rely on IsScheduledWithDRA(), CNS will not return IPConfig for any DRA allocated MTPNC, even to DRANET. What we want is CNS to not return this only if the caller invokes with includeDRAAllocations = false, which is CNI and anyone else who currently calls /requestIpConfigs in prod.

cns/middlewares/k8sSwiftV2.go
ravisraju: I am moving logic from getIPConfig to getIPConfigHelper so that I dont duplicate the 100 lines of code. 

only change to this function is the logic to not return anything to any CNI if the pod is scheduled with DRA. I am controlling that piece of logic using the flag. 

only concern here would be testing the new codeflow for existing CNI / DNC components with new CNS. to make sure older CRD definition ignores lack of the resourceClaims property and continue to return podIpInfo to CNI

cns/middlewares/k8sSwiftV2.go
ravisraju: Yes, we should remove this in future. added.

cns/middlewares/k8sSwiftV2.go
ravisraju: Yes this should be renamed to GetNICResourceSliceInfoByMAC. this was missed when copilot renamed NICNCInfo -> NICResourceSliceInfo

cns/middlewares/k8sSwiftV2.go
ravisraju: not always. see above comment where we disable DRA but continue to use a PrefixBlock PN

cns/middlewares/k8sSwiftV2.go
ravisraju: not sure if we can do that. the assumption then would be that a pod connected to multiple NIC needs to be all sharedNIC or be all dedicated NICs. User can have a RCT with two Claims, one for dedicated and one for shared, correct?

cns/restserver/nicresources.go
ravisraju: Yes we definitely want to do this so that NICNCReconciler and MTPNCReconciler can update the cache when the CRDs get updated.  but maybe not in this PR.

cns/middlewares/k8sSwiftV2.go
ravisraju: added

cns/restserver/restserver.go
ravisraju: it does when dra is turned off but we refer to a prefixblock PN

cns/restserver/requestclaimconfig.go
ravisraju: we init mtpnc only for swiftv2. follow same pattern here?

cns/middlewares/k8sSwiftV2.go
kmurudi: can the comment then reflect for the function that it is _extending_ to cover both type of ip configs or DRA and non-DRA pods?

cns/middlewares/k8sSwiftV2.go
ashvindeodhar: we are already exposing mtpnc in the other error messages. Since dranet is using a claimid to make a request, the claim is present from caller perspective. it's failing to get the config (mtpnc) for the given claim

cns/middlewares/k8sSwiftV2.go
ashvindeodhar: let's discuss this in a call. Not sure I understand this completely. Why do we need both? Can we simply this by encapsulating it as 'isDelegatedNicIPConfig' to differentiate between CNI and NRI?

When CNI calls, mtpnc.IsScheduledWithDRA() will prevent returning delegated nic ips to it. It will still get swiftv1 ip with nnc?

cns/middlewares/k8sSwiftV2.go
ashvindeodhar: this is not mix of dra and non-dra allocated nics for the same pod, right?
The true / false switch is controlling if you get IPs for delegated nic or not? What;s "All" ips here?

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +359,138 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+
+// GetMTPNCResourceSliceInfoByMAC lists the MTPNCs scheduled on this node and returns a map keyed by
+// canonical NIC MAC address with the NIC's network/subnet info and resource-slice
+// capacity from the MTPNC Spec. Dedicated NICs (single-allocation PodNetworks) usually
+// have no NICNetworkConfig and are served from here as a fallback.
+//
+// It intentionally does NOT wait for the MTPNC to be Ready: a not-yet-ready MTPNC may
+// have empty Spec NetworkID/SubnetGUID/SubnetResourceID, which are surfaced as-is (empty)
+// rather than treated as an error. Entries whose MAC is empty or unparseable are skipped.
+func (k *K8sSWIFTv2Middleware) GetMTPNCResourceSliceInfoByMAC(ctx context.Context) (map[string]*cns.NICResourceSliceInfo, error) {
```
ashvindeodhar: this and the above func are returning `NICResourceSliceInfo` as there is no `MTPNCResourceSliceInfo`. Can we fix the naming?
may I suggest `GetNICResourceSliceInfoFromNICNC` and `GetNICResourceSliceInfoFromMTPNC`? or you can use something else that adds more clarity to the func names

cns/middlewares/k8sSwiftV2.go
ashvindeodhar: got it, so it can be either dedicated or shared

cns/middlewares/k8sSwiftV2.go
ashvindeodhar: correct but for a given NIC - it will be either dedicated or shared. So the `result` map which is key'ed on `mac` would have single/common 'info' across different mtpncs that refer to it. So we if we have seen that mac in the map, no need to construct it again

cns/restserver/nicresources.go
ashvindeodhar: is there any particular reason to not do it in this PR?

ashvindeodhar (changes requested):

crd/multitenancy/api/v1alpha1/nicnetworkconfig.go
```diff
@@ -0,0 +1,136 @@
...
+type NICNetworkConfigSpec struct {
+	// PodNetwork is the name of the PodNetwork
+	PodNetwork string `json:"podNetwork"`
+	// NodeName is the name of the node this NIC belongs to
+	NodeName string `json:"nodeName"`
+	// MACAddress is the MAC address of the NIC, used to create the network container
+	MACAddress string `json:"macAddress"`
+	// customer subnet id
+	SubnetResourceID string `json:"subnetResourceID"`
+	// customer subnet guid
```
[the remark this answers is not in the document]
ravisraju: This file change is not part of this PR. and SubnetResourceID is always set when a NICNC is created as it gets created when first pod lands on a NIC already knowing which Subnet it wants to connect to.

cni/network/invoker_cns.go
ravisraju: removed

cns/restserver/api.go
ravisraju: addressed

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +288,145 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+func nicNCInfoFromPodNetwork(pn *v1alpha1.PodNetwork) *cns.NICNCInfo {
+	networkID := pn.Spec.NetworkID
+	if networkID == "" {
+		networkID = pn.Spec.VnetGUID
+	}
+	return &cns.NICNCInfo{
+		NetworkID: networkID,
+		SubnetID:  subnetNameFromResourceID(pn.Spec.SubnetResourceID),
+	}
+}
```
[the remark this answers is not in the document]
ravisraju: we dont have SubnetID anymore. its SubnetGUID and SubnetName. they are not ambiguous names.

cns/restserver/api.go
ravisraju: outdated

cns/restserver/api.go
ravisraju: outdated

cns/middlewares/k8sSwiftV2.go
ravisraju: done

cns/middlewares/k8sSwiftV2.go
ravisraju: reconciler doing this for nicnc. in future we'd only cache nicnc / mtpnc that belongs to this node

cns/middlewares/k8sSwiftV2.go
ravisraju: removed

cns/middlewares/k8sSwiftV2.go
ravisraju: removed

cns/restserver/api.go
ravisraju: fixed up

cns/kubecontroller/nicnetworkconfig/reconciler.go
ravisraju: added predicate to filter based on spec.nodeName. not using IsControlledBy as we do not see OwnerReference on a NICNC to the node today. Discuss to close on that call to see if any change here is needed.

cns/restserver/api.go
ravisraju: done

cns/service/main.go
ravisraju: not introducing selectableFields as part of this PR

cns/restserver/restserver.go
ravisraju: done

cns/restserver/restserver.go
ravisraju: but removing the second line here as mtpncClient is mainly to read mtpnc for populating dedicatedNICs. I see no need to talk about corner cases everywhere

cns/restserver/restserver.go
ravisraju: done

cns/restserver/nicresources.go
ravisraju: fixed

cns/restserver/nicresources.go
ravisraju: not extracting info here. just doing client.Get against the informer cache here

cns/restserver/nicresources.go
ravisraju: adding a metric to track this. we will have to come look at the logs to understand what macaddress is not getting parsed. but we will have a way to alert ourselves

cns/restserver/nicresources.go
ravisraju: yes, interfaceName is a metadata that's only added for viewing purposes (some cx might want it). it bears no role towards any of the scheduling path. we will always rely on macaddress for closing the allocation loop (something we need to keep in mind when we want to merge private dranet to upstream as the existing hooks are name based in most places.)

cns/restserver/nicresources.go
ravisraju: changed to nicResourceSliceInfoByMAC

cns/restserver/nicresources.go
ravisraju: removed

cns/restserver/nicresources.go
ravisraju: removed since its a edge case that we see when dranet is rolledback

cns/middlewares/k8sSwiftV2.go
miguelgoms: lets not hide useful information here. mtpnc is a visible object in the apiserver and if troubleshooting this error the first step is to confirm if the mtpnc object exists and has the claims.

cns/middlewares/k8sSwiftV2.go
miguelgoms: i think we covered this offline, but yes we can rename this function to GetSwiftv2IPConfigs to be more obvious.

cns/restserver/nicresources.go
ashvindeodhar: we discussed this - resolving

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +359,138 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+	hw, err := net.ParseMAC(raw)
+	if err != nil {
+		return "", false
+	}
+	return hw.String(), true
+}
+
+// sharedNICDRACapacity is the resource-slice capacity advertised for a DRA-managed
+// prefix-on-NIC (shared) NIC: the scheduler can place up to this many pods on it.
+const sharedNICDRACapacity = 16
```
miguelgoms, line 412: I'd expect this to be sourced from the NicNC resource instead of hardcoding this expectation on CNS (client-side).

cns/kubecontroller/nicnetworkconfig/reconciler.go
ravisraju: removing this. will address this as part of perf pr. Final conclusion - no predicates here. we will trigger a reconcile for all nicncs on the cache - its a no op however. when cacheOpts are introduced, this code automatically gets the perf bump

cns/restserver/requestclaimconfig.go
ravisraju: From a DRA point of view this gets all config needed to update DRA resource state based on a resource'claim' processing.

it includes IPAllocation info of the pod and NICState that might change based on this Pod being scheduled. its equivalent of other DRANET functions such as GetDeviceConfig, GetProfileConfig etc.. This function essentially does the job of GetProfileCofig + GetDeviceAttributes

But I do understand this might be interpreted as resourceclaim's device's config. resourceclaim itself would not have config but if you have better naming i'm open to change it.

cns/restserver/requestclaimconfig.go
ravisraju: i can pass in resources after the outer loop finishes building but i am curious why this is not ok

cns/restserver/nicresources.go
ravisraju: The NIC in its essence starts with the MAC, Name and VMUniqueId or nodeName. All additions here are networking related that are not attributes of the NIC itself - its a fleeting state of the NIC (Subnet, Vnet, SubnetGUID, capacity changes) are not guaranteed for any NIC except swiftv2 nics that are in use. 

Maybe name this function as populateNICNetworkingInfo if thats better or enrichNICResourceFromMultiTenantCRDs - to capture everything that we want from MTPNC and NICNC now and in future. but IMO any name is ok for this helper function.

ravisraju: NICResource. NICResourceSliceInfo is just the NetworkInfo (Subnet, VNET and capacity derived from CRD). I am renaming it to NICResourceNetworkInfo

cns/middlewares/k8sSwiftV2.go
ravisraju: agreed - syncing variable name to error message

cns/middlewares/k8sSwiftV2.go
ravisraju: GetIPConfig function is also tehcnically GetSwiftv2IPConfigs (without the DRA scheduled pods) 

This is the comment on top of GetIPConfig

// getIPConfig returns the pod's SWIFT V2 IP configuration.
func (k *K8sSWIFTv2Middleware) getIPConfig(ctx context.Context, podInfo cns.PodInfo) ([]cns.PodIpInfo, error) {
	return k.getIPConfigHelper(ctx, podInfo, false)
}

I feel we should disambiguate / qualify this further. Ideally we change getIPConfig that CNI calls via requestipconfigs to something like getNonDRASwiftV2IpConfig

cns/middlewares/k8sSwiftV2.go
ravisraju: we can do len(AvailableIPs) + len(PodAllocations) to come up with this number or have a CRD field in status of NICNC to call out how many IPs are being shared on that NIC. I Prefer the latter. thoughts?

cns/middlewares/k8sSwiftV2.go
ravisraju: yes thats what happens here. for every mtpnc we go through a list of mac that it refers to. if mac is already present in the result, we dont update. we still have to process rest of the macs on this mtpnc because they might not be already seen. 

for example, mtpnc 1 -> mac 2 might be same as mtpnc 4 -> mac 3 (but we still have to look at mac 4 and mac 5 of mtpnc 4 because they might be dedicated.

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +359,138 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+
+		// A dedicated NIC scheduled with DRA advertises the dedicated capacity; a NIC
+		// created without DRA has no capacity to advertise to the scheduler (marked 0).
+		capacity := 0
+		if mtpnc.IsScheduledWithDRA() {
+			capacity = dedicatedNICDRACapacity
+		}
+		info := &cns.NICResourceSliceInfo{
+			NetworkID:  mtpnc.Spec.NetworkID,
+			SubnetGUID: mtpnc.Spec.SubnetGUID,
```
ravisraju: I see an issue here. mtpnc will not necessarily have one NetworkID or SubnetGUID. Every InterfaceInfo of a MTPNC can have different NetworkID and SubnetGUID, correct?

@ashvindeodhar @miguelgoms

cns/middlewares/k8sSwiftV2.go
```diff
@@ -278,3 +359,138 @@ func (k *K8sSWIFTv2Middleware) GetInfravnetAndServiceCidrs() ([]string, []string
...
+		// Only consider MTPNCs scheduled on this node.
+		if mtpnc.Status.NodeName != k.NodeName {
+			continue
+		}
+
+		// A dedicated NIC scheduled with DRA advertises the dedicated capacity; a NIC
+		// created without DRA has no capacity to advertise to the scheduler (marked 0).
+		capacity := 0
+		if mtpnc.IsScheduledWithDRA() {
+			capacity = dedicatedNICDRACapacity
```
ravisraju, line 484: this is not always dedicatedNICDRACapacity. It needs to be determined based on whether the InterfaceInfo. SharedNIC is true or if a NICNC for this NIC was already seen and if capacity is already updated in the map leave it alone.

cns/middlewares/k8sSwiftV2_test.go
ravisraju: done

cns/middlewares/k8sSwiftV2.go
ravisraju: done. renamed to GetNICResourceNetworkInfoFromMTPNC and GetNICResourceNetworkInfoFromNICNC

ravisraju: done. since i renamed sliceInfo to NetworkInfo these functions will be called GetNICResourceNetworkInfoFromNICNC and GetNICResourceNetworkInfoFromMTPNC

cns/restserver/nicresources.go
ravisraju: addressed

cns/restserver/nicresources.go
ravisraju: you are right. this should be 1 here. but i am marking the default value of a NICResource as 1 when we assign MAC address and VMUniqueId from NodeInfo. Later based on the ResourceSliceInfo (now renamed ResourceNetworkInfo, this capacity can change to "0" or "16" or stay "1". I am letting the last line of this function drive that change

cns/restserver/nicresources_test.go
ravisraju: addressed

cns/restserver/nicresources_test.go
ravisraju: removed

cns/restserver/requestclaimconfig.go
ravisraju: existing middleware. i will leave it as it is

ravisraju: and it is not specific to nic. nics and pods.

cns/restserver/requestclaimconfig.go
ravisraju: no way to determine all we know is that the claim UID supplied was not found in cns. i am ok either way

cns/restserver/requestclaimresourceinfo_test.go
ravisraju: addressed

cns/restserver/requestclaimconfig.go
ravisraju: naming this to requestClaimResourceInfo

cns/restserver/api.go
```diff
@@ -1314,6 +1315,206 @@ func (service *HTTPRestService) getVMUniqueID(w http.ResponseWriter, r *http.Req
...
+	logger.Request(service.Name, "getNICResources", nil)
+	ctx := r.Context()
+
+	switch r.Method {
+	case http.MethodGet:
+		// Step 1: Get list of NICs (MACs) from NodeInfo CRD.
+		logger.Printf("[Azure CNS] getNICResources: fetching NodeInfo CRD for node %s", service.nodeName)
+		nodeInfo, err := service.nodeInfoCli.Get(ctx, service.nodeName)
+		if err != nil {
+			resp := cns.GetNICResourcesResponse{
```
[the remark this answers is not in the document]
ravisraju: outdated and resolved

cns/middlewares/k8sSwiftV2.go
ravisraju: naming this as GetSwiftV2IpConfigs which calls -> GetSwiftV2IpConfigsHelper

cns/middlewares/k8sSwiftV2.go
ravisraju: sticking with constant 16 with a todo to fix it later once CRDs are updated.

cns/middlewares/k8sSwiftV2.go
ravisraju: Moving the Networking info into InterfaceInfos away from Spec in a separate CRD change. Not taking dependency on that PR here since these details are not necessary for scheduling dedicated / shared NICs with DRA

cns/middlewares/k8sSwiftV2.go
ravisraju: we completely ignore MACs of a MTPNC whose IsShared is set to true while filling out the NICResourceInfo. Only determination here to be made is if this MTPNC was created for a ResourceClaim or not (capacity = 1 or 0). The assumption here is that later in the code, we filter out mac addresses who are not dedicated (InterfaceInfo.IsShared == true)

cns/middlewares/k8sSwiftV2.go
miguelgoms: yes lets track CRD change separately.

cns/middlewares/k8sSwiftV2.go
miguelgoms: I'm ok with that rename if getIPConfig is only invoked through the wrapper and within the middleware implementation; else just rename the helper to the above.

miguelgoms (approved):

ashvindeodhar (approved):

cns/middlewares/k8sSwiftV2.go
ravisraju: yes, getIpConfig is only invoked within the swiftv2 middleware. it serves swiftv2 ips exclusively.

cns/restserver/api.go
ravisraju: done

cns/middlewares/k8sSwiftV2.go
ravisraju: not applicable anymore. we are using all ips on the nic

cns/restserver/nicresources.go
ravisraju: we are setting the default value of capacity as 1. and set it to 0 if nicnc / mtpnc is not scheduled by dra. set it to 16 if there is a nicnc that is scheduled by dra

cns/restserver/requestclaimconfig.go
ravisraju: included

cns/restserver/requestclaimconfig.go
ravisraju: added more debug info

PR #4458, opened 2026-06-11
fix: restore DisableAsyncDelete check in CNS IPAM invoker Delete

behzad-mir:
## Summary

The `disableAsyncDelete` conflist flag was previously checked in the stateless CNI Delete path to prevent writing to the fsnotify watcher directory when CNS is unreachable. This guard was accidentally removed during a refactor (ed48bcf36 — "Feat: necessary fixes for Stateless CNI Linux #4212"), leaving standalone/ACI scenarios (which have no CNS daemon to process async deletes) writing orphan files to `/var/run/azure-vnet/deleteIDs`.

## Changes

Re-adds the `nwCfg.DisableAsyncDelete` check in both connection-error branches of `CNSIPAMInvoker.Delete` (`cni/network/invoker_cns.go`):
- When `disableAsyncDelete: true` in the conflist, the CNI now returns the connection error directly instead of deferring to the watcher.
- When `disableAsyncDelete` is false (default), behavior is unchanged — async delete files are still written for CNS to pick up later.

## Motivation

In ACI/standalone scenarios (e.g., `azure-windows-swiftv2-stateless.conflist`), there is no CNS daemon running to process async deletes. Without this guard, orphan files accumulate in the deleteIDs directory with no consumer.

## Testing

- `go build ./cni/...` passes
- `TestCNSIPAMInvoker_Delete*` tests pass

### DIFF
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -381,7 +381,7 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 
 			if err = invoker.cnsClient.ReleaseIPAddress(context.TODO(), ipConfig); err != nil {
 				if errors.As(err, &connectionErr) {
-					if nwCfg.DisableAsyncDelete {
+					if nwCfg != nil && nwCfg.DisableAsyncDelete {
 						logger.Error("Failed to release IP address and async delete is disabled",
 							zap.String("podInterfaceID", ipConfigs.PodInterfaceID), zap.String("containerID", args.ContainerID), zap.Error(log.NewErrorWithoutStackTrace(err)))
 						return errors.Wrap(err, "failed to release IP address using ReleaseIPAddress and async delete is disabled")
@@ -402,7 +402,7 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 			}
 		} else {
 			if errors.As(err, &connectionErr) {
-				if nwCfg.DisableAsyncDelete {
+				if nwCfg != nil && nwCfg.DisableAsyncDelete {
 					logger.Error("Failed to release IP address and async delete is disabled",
 						zap.String("podInterfaceID", ipConfigs.PodInterfaceID), zap.String("containerID", args.ContainerID), zap.Error(log.NewErrorWithoutStackTrace(err)))
 					return errors.Wrap(err, "failed to release IP address using ReleaseIPs and async delete is disabled")
--- a/cni/network/invoker_cns.go
+++ b/cni/network/invoker_cns.go
@@ -381,6 +381,11 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 
 			if err = invoker.cnsClient.ReleaseIPAddress(context.TODO(), ipConfig); err != nil {
 				if errors.As(err, &connectionErr) {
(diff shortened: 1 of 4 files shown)


cni/network/invoker_cns.go
```diff
@@ -381,6 +381,11 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 
 			if err = invoker.cnsClient.ReleaseIPAddress(context.TODO(), ipConfig); err != nil {
 				if errors.As(err, &connectionErr) {
+					if nwCfg.DisableAsyncDelete {
+						logger.Error("Failed to release IP address and async delete is disabled",
+							zap.String("podInterfaceID", ipConfigs.PodInterfaceID), zap.String("containerID", args.ContainerID), zap.Error(log.NewErrorWithoutStackTrace(err)))
+						return errors.Wrap(err, "failed to release IP address using ReleaseIPAddress and async delete is disabled")
+					}
```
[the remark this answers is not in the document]
jackieluc: agree with this comment. please add a unit test so it doesn't regress again

cni/network/invoker_cns_test.go
```diff
@@ -1515,6 +1516,129 @@ func TestCNSIPAMInvoker_Delete_NotSupportedAPI(t *testing.T) {
...
+		releaseIPsErr    error
+		wantErr          bool
+		wantErrContains  string
+	}{
+		{
+			name:            "connection failure with async delete disabled returns error",
+			disableAsync:    true,
+			releaseIPsErr:   connErr,
+			wantErr:         true,
+			wantErrContains: "async delete is disabled",
```
jackieluc: nit: error log message assertion seems brittle. can we check what kind of error instead with `require.ErrorAs`?

behzad-mir: done

cni/network/invoker_cns.go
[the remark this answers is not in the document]
behzad-mir: done

jackieluc (approved):

cni/network/invoker_cns.go
```diff
@@ -397,6 +402,11 @@ func (invoker *CNSIPAMInvoker) Delete(address *net.IPNet, nwCfg *cni.NetworkConf
 			}
 		} else {
 			if errors.As(err, &connectionErr) {
+				if nwCfg != nil && nwCfg.DisableAsyncDelete {
```
QxBytes, line 405: we should simplify these branches in a follow up-- they look like they are doing very similar things-- put similar logic for unsupported/supported api in its own func?

cni/network/invoker_cns_test.go
```diff
@@ -1515,6 +1516,129 @@ func TestCNSIPAMInvoker_Delete_NotSupportedAPI(t *testing.T) {
...
+			disableAsync: true,
+			releaseIPErr: connErr,
+			wantErr:      true,
+			wantConnErr:  true,
+		},
+		{
+			name:         "unsupported API path with connection failure and async delete enabled attempts async write",
+			disableAsync: false,
+			releaseIPErr: connErr,
+			wantErr:      true,
```
QxBytes, line 1604: I believe this test is counting on an error being thrown from fsnotify add file (because the fsnotify dir doesn't exist or permission denied). But it's possible that the write succeeds if the fsnotify directory _does_ exist and in that case we'd return nil when we want an error.

QxBytes (approved):
comments are non blocking

rbtr (approved):

PR #4459, opened 2026-06-11
MTPNC changes to support DRA Pod Scheduling

ravisraju:
feat: add information to mtpnc for DRANET processing

**Reason for Change**:
It introduces the ability to identify whether a pod's network config was configured via DRA

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -142,6 +157,14 @@ type MultitenantPodNetworkConfigStatus struct {
 	// Status represents the overall status of the MTPNC
 	// +kubebuilder:validation:Optional
 	Status MTPNCStatus `json:"status,omitempty"`
+	// ResourceClaims lists the ResourceClaim names allocated to the pod by the
+	// scheduler via Dynamic Resource Allocation (DRA). An empty or nil list
+	// indicates the pod was scheduled without DRA. When non-empty, dataplane
+	// programming is owned by the DRA driver (NRI), and CNS will skip the
+	// CNI pod-info delivery path for this pod. Use IsDRAScheduled() to read.
```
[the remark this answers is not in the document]
miguelgoms: rather: remove the implementation details of azure-cni/cns from the comment altogether. 
"when non-empty dataplane programming is owned by the DRA driver (NRI), otherwise for legacy pods it is owned by azure-cni".

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -142,6 +157,14 @@ type MultitenantPodNetworkConfigStatus struct {
...
+	// +kubebuilder:validation:Optional
+	// +listType=set
+	ResourceClaims []string `json:"resourceClaims,omitempty"`
```
miguelgoms: I'm not opposed to changing the mtpnc IsReady() semantics for new .status fields, if we track the CS version dependency for clusters enabling DRA. But these all seem to be copied or inferred from the pod spec. 

can't we add these fields into .spec instead of .status? 
they are all denormalized inputs from the PNI / PN and Pod spec, not the output from the reconcile operation -> network allocation.

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
```diff
@@ -52,6 +52,23 @@ type MultitenantPodNetworkConfigSpec struct {
...
+	SubnetGUID string `json:"subnetGUID,omitempty"`
+	// SubnetName is the short subnet name (the trailing segment of PodNetwork.Spec.SubnetResourceID,
+	// e.g. "pod-subnet"). Provided for human-readable identification and debuggability; the
+	// authoritative identifier is SubnetGUID.
+	// +kubebuilder:validation:Optional
+	SubnetName string `json:"subnetName,omitempty"`
+	// ResourceClaims lists the ResourceClaim names allocated to the pod.
+	// +kubebuilder:validation:Optional
+	// +listType=set
+	ResourceClaims []string `json:"resourceClaims,omitempty"`
```
kmurudi, line 71: will there be multiple resourceclaims tied to one MTPNC? in case of multiple PNs in PNI?

miguelgoms (dismissed):

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
ravisraju: Good point. yeah that was the reason I made this an array to support multiple PNs. but from yday's discussion, we decided to create a single ResourceClaimTemplate for every PNI and a single ResourceClaimTemplate always creates a single ResourceClaim.

I'd still prefer to keep this as a list for the sake of extensibility.

kmurudi: single RCT for every PN or PNI? 

once a pod is created -> scheduler creates resourceClaims & then this list would basically have the resourceClaim(not Template) created for the pod & copy it over to MTPNC?

crd/multitenancy/api/v1alpha1/utils.go
```diff
@@ -15,3 +15,9 @@ func (m *MultitenantPodNetworkConfig) IsReady() bool {
 func (m *MultitenantPodNetworkConfig) IsDeleting() bool {
 	return !m.DeletionTimestamp.IsZero()
 }
+
+// IsDRAScheduled reports whether the pod has any ResourceClaims allocated
+// (i.e. was scheduled with Dynamic Resource Allocation).
+func (m *MultitenantPodNetworkConfig) IsDRAScheduled() bool {
```
ashvindeodhar: nitpicking on wording here - as you mentioned in your comment, the better name would be IsScheduledWithDRA

crd/multitenancy/api/v1alpha1/utils.go
```diff
@@ -15,3 +15,9 @@ func (m *MultitenantPodNetworkConfig) IsReady() bool {
...
+	return len(m.Spec.ResourceClaims) > 0
```
ashvindeodhar, line 24: this will still return true if they are not using our DRA driver but using say GPU DRA driver. We should check if the claim generated is for our DRA driver.

ashvindeodhar (changes requested):

crd/multitenancy/api/v1alpha1/multitenantpodnetworkconfig.go
ravisraju: single PNI -> Single RCT -> Single RC. This will be set by PodReconciler when MTPNC object is created. Will leave this as a list as discussed with the squad

crd/multitenancy/api/v1alpha1/utils.go
ravisraju: I was in between these two. went for the shorter version but I do like this better, will change

crd/multitenancy/api/v1alpha1/utils.go
ravisraju: as discussed, we would only add to this list if the ResourceClaim was created for our custom multitenancy driver. I have edited the description of this field to reflect that change.

crd/multitenancy/manifests/multitenancy.acn.azure.com_multitenantpodnetworkconfigs.yaml
```diff
@@ -152,6 +170,17 @@ spec:
               primaryIP:
                 description: Deprecated - use InterfaceInfos
                 type: string
+              resourceClaims:
+                description: |-
+                  ResourceClaims lists the ResourceClaim names allocated to the pod by the
+                  scheduler via Dynamic Resource Allocation (DRA). An empty or nil list
+                  indicates the pod was scheduled without DRA. When non-empty, dataplane
+                  programming is owned by the DRA driver (NRI), and CNS will skip the
+                  CNI pod-info delivery path for this pod. Use IsDRAScheduled() to read.
```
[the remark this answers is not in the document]
ravisraju: not relevant with latest changes

crd/multitenancy/api/v1alpha1/utils.go
```diff
@@ -15,3 +15,12 @@ func (m *MultitenantPodNetworkConfig) IsReady() bool {
...
 	return !m.DeletionTimestamp.IsZero()
 }
+
+// IsDRAScheduled reports whether this pod was scheduled with Dynamic Resource
+// Allocation (DRA). It is derived from Status.ResourceClaims: a non-empty list
+// means a DRA driver owns dataplane programming for the pod (via NRI), and CNS
+// should skip CNI pod-info delivery. Callers must use this helper rather than
+// inspecting the field directly, so the contract has a single source of truth.
+func (m *MultitenantPodNetworkConfig) IsDRAScheduled() bool {
+	return len(m.Status.ResourceClaims) > 0
```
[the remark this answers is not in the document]
ravisraju: moved these fields to spec from status. so we do not have this risk anymore

ravisraju:

ashvindeodhar (dismissed):

miguelgoms (approved):

ashvindeodhar (approved):

PR #4469, opened 2026-06-22
fix: treat CNS DeleteEndpointState NotFound as a successful no-op

jackieluc:
**Reason for Change:**

In Stateless CNI mode, endpoint state lives in CNS (with `ManageEndpointState` enabled) rather than in a CNI-local state file. On teardown, CNI calls into CNS to delete that endpoint state.

When multiple CNI DEL calls arrive for the same pod (e.g. duplicate/retried teardowns from containerd/Service Fabric), they race to delete the same endpoint state in CNS. The first call wins and removes the state; the later calls find nothing to delete and CNS responds with `NotFound`.

Previously the CNS client treated that `NotFound` as a hard error. That error bubbled up through `network.DeleteState` as a **retriable** error to CNI, then back to containerd/SF, which would keep retrying a delete that can never succeed -- there is nothing left to delete. The teardown was actually complete, but the caller saw repeated failures.

**Issue Fixed:**

Deleting an endpoint that is already absent is now idempotent. `DeleteEndpointState` returns the CNS response alongside the error so the caller (`network.DeleteState`) can inspect the `ReturnCode`. A `NotFound` is logged and treated as a successful no-op rather than a retriable failure. This keeps the error-handling decision in CNI and mirrors the idempotent-delete behavior already used for HNS network/endpoint deletions.

**Notes:**

Also corrects a misleading log line in `restoreState` that incorrectly claimed it was removing `endpoints.json`.

### DIFF
--- a/cns/client/client.go
+++ b/cns/client/client.go
@@ -1132,7 +1132,11 @@ func (c *Client) DeleteEndpointState(ctx context.Context, endpointID string) (*c
 		return nil, errors.Wrap(err, "failed to decode CNS Response")
 	}
 
-	if response.ReturnCode != 0 {
+	if response.ReturnCode == types.NotFound {
+		return &cns.Response{}, nil
+	}
+
+	if response.ReturnCode != types.Success {
 		return nil, errors.New(response.Message)
 	}
 
--- a/cns/client/client.go
+++ b/cns/client/client.go
@@ -1132,12 +1132,9 @@ func (c *Client) DeleteEndpointState(ctx context.Context, endpointID string) (*c
 		return nil, errors.Wrap(err, "failed to decode CNS Response")
 	}
 
-	if response.ReturnCode == types.NotFound {
-		return &cns.Response{}, nil
-	}
-
 	if response.ReturnCode != types.Success {
-		return nil, errors.New(response.Message)
+		// return the response so the caller can inspect ReturnCode (e.g. NotFound) and decide how to handle it
+		return &response, errors.New(response.Message)
 	}
 
 	return &response, nil
--- a/network/manager.go
+++ b/network/manager.go
@@ -789,7 +789,11 @@ func (nm *networkManager) DeleteState(epInfos []*EndpointInfo) error {
 				// swiftv2 multitenancy does not call plugin.ipamInvoker.Delete and so state does not automatically clean up. this call is required to
 				// cleanup state in CNS
 				// One Delete call for endpointID will remove all interface info associated with that endpointID in CNS
-				if _, err := nm.CnsClient.DeleteEndpointState(context.TODO(), epInfo.EndpointID); err != nil {
+				if response, err := nm.CnsClient.DeleteEndpointState(context.TODO(), epInfo.EndpointID); err != nil {
+					if response != nil && response.ReturnCode == types.NotFound {
+						logger.Info("Delete called on the endpoint state which doesn't exist in CNS, treating as success", zap.String("endpointID", epInfo.EndpointID))
+						return nil
+					}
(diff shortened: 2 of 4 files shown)


network/manager.go
```diff
@@ -789,15 +789,10 @@ func (nm *networkManager) DeleteState(epInfos []*EndpointInfo) error {
...
 				// One Delete call for endpointID will remove all interface info associated with that endpointID in CNS
-				response, err := nm.CnsClient.DeleteEndpointState(context.TODO(), epInfo.EndpointID)
-				if err != nil {
-					if response != nil && response.ReturnCode == types.NotFound {
-						logger.Info("Endpoint state not found in CNS", zap.String("endpointID", epInfo.EndpointID))
-						return nil
-					}
+				if _, err := nm.CnsClient.DeleteEndpointState(context.TODO(), epInfo.EndpointID); err != nil {
 					return errors.Wrapf(err, "Delete endpoint API returned with error for endpoint %s", epInfo.EndpointID)
 				}
```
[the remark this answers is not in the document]
jackieluc: done

cns/client/client.go
```diff
@@ -1128,7 +1128,11 @@ func (c *Client) DeleteEndpointState(ctx context.Context, endpointID string) (*c
 		return nil, errors.Wrap(err, "failed to decode CNS Response")
 	}
 
-	if response.ReturnCode != 0 {
+	if response.ReturnCode == types.NotFound {
+		return nil, nil
+	}
+
+	if response.ReturnCode != types.Success {
 		return nil, errors.New(response.Message)
```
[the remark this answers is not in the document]
jackieluc: return empty response object instead

network/manager.go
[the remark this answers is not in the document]
rbtr: the [skills](https://github.com/Azure/azure-container-networking/blob/master/.github/skills/acn-go-errors-logging/SKILL.md#error-string-rules) are working! 😍

rbtr (dismissed):
lgtm

cns/client/client.go
```diff
@@ -1128,7 +1128,11 @@ func (c *Client) DeleteEndpointState(ctx context.Context, endpointID string) (*c
 		return nil, errors.Wrap(err, "failed to decode CNS Response")
 	}
 
-	if response.ReturnCode != 0 {
+	if response.ReturnCode == types.NotFound {
+		return &cns.Response{}, nil
```
behzad-mir: Isn't it better to return the do not found error and then CNi process that error? I was wondering if we return success we kind of lose that information in CNI logs

jackieluc: fixed

cns/client/client.go
```diff
@@ -1128,8 +1128,9 @@ func (c *Client) DeleteEndpointState(ctx context.Context, endpointID string) (*c
 		return nil, errors.Wrap(err, "failed to decode CNS Response")
 	}
 
-	if response.ReturnCode != 0 {
-		return nil, errors.New(response.Message)
+	if response.ReturnCode != types.Success {
+		// return the response so the caller can inspect ReturnCode (e.g. NotFound) and decide how to handle it
+		return &response, errors.New(response.Message)
```
rbtr, line 1137: the bug was that the response was not returned, correct? the client code looks like it already expected it to be returned and there was an attempt to handle a NotFound already?

rbtr (approved):

cns/client/client.go
jackieluc: correct. originally it returned `nil, err` and the handling of it in CNI basically was dead code

behzad-mir (approved):

PR #4499, opened 2026-07-02
ci: add CI/CD for cilium image version drift

jpayne3506:
**Reason for Change**:

Adds automation to keep up with drift. Similar dev experience to `make dockerfile`

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.github/workflows/cilium-versions.yaml
+++ b/.github/workflows/cilium-versions.yaml
@@ -1,0 +1,47 @@
+name: 'Cilium Image Versions'
+on:
+  workflow_dispatch:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+  merge_group:
+    types:
+      - checks_requested
+
+permissions:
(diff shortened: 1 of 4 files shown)


rbtr (approved):

QxBytes (approved):
Should be fine but if mcr has a bad tag, this will auto update our deploy.mk to always use the latest even if we roll back. In the past we have had scenarios where a bad tag was pushed to mcr and this would make the script less useful until we pushed an updated tag/version.

jpayne3506:
> Should be fine but if mcr has a bad tag, this will auto update our deploy.mk to always use the latest even if we roll back. In the past we have had scenarios where a bad tag was pushed to mcr and this would make the script less useful until we pushed an updated tag/version.

Hopefully because the version is scoped to the minor version that is present with `CILIUM_VERSION_TAG` we will avoid the bad image tags.

PR #4502, opened 2026-07-02
fix: prevent leaked CNS endpoint records from late ADDs after DELETE

rbtr:
Once ADD and DEL are each serialized, the split brain is gone — but one ordering
still leaves bad state. If DEL runs to completion *before* its ADD, DEL is a pure
no-op, and the ADD that lands afterwards assigns an IP and writes endpoint state
for a sandbox that no longer exists. Nothing will ever delete that record.

That leak is not merely wasted address space. `ReconcileIPAssignment` skips an
endpoint IP that is not in the current NC pool but leaves the endpoint record in
place, so if the pool shrinks and later reintroduces that IP, IPAM offers it as
`Available` while the stale record still claims it. The next pod to take it
produces the second endpoint record — the same `duplicate IP detected in CNS
initialization` failure, reached by a slower path.

## Fix

A durable write-ahead delete intent keyed by infra container ID, stored alongside
endpoint state.

- **DEL:** persist the delete intent, remove endpoint state, release the IP.
- **ADD:** reject while an unexpired intent exists for that container.
- **PATCH:** also rejected. `UpdateEndpointHelper` *creates* a record when one is
  absent, so an ungated late PATCH would resurrect exactly the leaked record this
  change exists to prevent.
- **Startup:** load endpoint and intent state fail-closed, replay active intents
  (deleting the corresponding endpoint records) and prune expired ones, all before
  the pod-info provider reads the store. Statefiles predating the intent key are
  tolerated.

### Intent lifetime

TTL is 10 minutes. It only needs to outlive a CNI request that was already in
flight when the DELETE arrived, plus a CNS restart. An intent that outlives the
racing request can only reject work that should have succeeded, so it is
deliberately far shorter than a pod lifetime.

The intent keeps the timestamp of the *first* delete. Refreshing it on every DELETE
would be self-renewing: a rejected ADD makes the runtime retry DELETE, which would
extend the intent indefinitely and permanently block that container.

## Scope

This prevents new corruption. It does **not** repair nodes whose statefile is
already corrupt — those records have no delete intents, so replay does not touch
them and startup still fails. Recovering an already-bricked node still requires
clearing its endpoint statefile. A repair path needs authoritative runtime state
to decide which of the duplicate records is live, and is left as follow-up.

## Performance under pod churn

Each DELETE now performs two store writes (the intent, then the endpoint state).
`jsonFileStore.Write` re-serializes the whole file, so both writes marshal the
endpoints and intents blobs together, and all of it happens under the global
service lock that serializes every ADD and DELETE on the node.

Measured cost of one DELETE on a 250-endpoint node, varying the number of
outstanding intents:

| outstanding intents | DELETE | vs. no intents |
|---|---|---|
| 0 | 0.44 ms | 1.0x |
| 250 | 0.69 ms | 1.6x |
| 1 000 | 1.41 ms | 3.4x |
| 5 000 | 5.32 ms | 12.7x |
| 86 400 | 99.7 ms | 227x |

Cost is linear in the number of outstanding intents, which is why the TTL is
minutes rather than hours. With a 10 minute TTL the intent count is the number of
*distinct* containers deleted in that window; a node cycling its entire 250-pod
complement every 10 minutes sits at roughly 0.69 ms, which is immaterial next to
the multi-second ADDs that trigger this bug class. The 86 400 row is what a 24
hour TTL would reach at ~1 delete/sec, and would itself have caused the ADD
timeouts this change exists to contain.

Repeated DELETEs for the same container do not add entries or refresh
timestamps, so a rejection storm cannot inflate the map.

If this ever needs to be cheaper, the fix is to fold both writes into a single
flush rather than to shorten the TTL further. That halves the I/O and is strictly
safer than the current two-flush sequence, but it needs a multi-key write on
`KeyValueStore` and is left as follow-up.

## Tests

- DEL-before-ADD rejection, sequential and under forced lock contention
  (`TestConcurrentDeleteThenAddDoesNotLeakIP`)
- repeated DELETE does not extend intent lifetime
- startup replay, expiry pruning, and fail-closed on unreadable state
- late PATCH rejection
- statefile with no intent key

## Validation

- `go test -race -count=1 ./cns/restserver/... ./store/...`


store/json.go
```diff
@@ -122,9 +122,19 @@ func (kvs *jsonFileStore) Write(key string, value interface{}) error {
 		return err
 	}
 
-	kvs.data[key] = &raw
+	nextData := make(map[string]*json.RawMessage, len(kvs.data)+1)
```
pjohnst5: My gut tells me we should split up json.go and json_test.go changes into a separate PR first (just to make the tombstone related changes smaller)

cns/restserver/util.go
```diff
@@ -124,22 +124,31 @@ func (service *HTTPRestService) restoreState() {
 		if service.EndpointStateStore == nil {
 			//nolint:staticcheck // TODO: migrate to zap
 			logger.Errorf("[Azure CNS]  OptManageEndpointState is enabled but EndpointStateStore is not initialized; endpoint state persistence/restoration is disabled.")
-			return
+			return nil
```
pjohnst5: I know this isn't part of this PR, but it's odd to me / doesn't make sense yet, why `OptManageEndpointState` would be true, but `service.EndpointStateStore` would be nil 🤔

https://github.com/Azure/azure-container-networking/blob/7738d7e1fb2590f42e7d23c2c3f97b4ebc29b069/cns/service/main.go#L773

Is where it is initialized, and it only proceeds if  `err != nil`, but I digress

pjohnst5: Yeah, probably something for a different PR, but we should probably? return an error here (I fail to see why not)

cns/restserver/ipam.go
```diff
@@ -522,18 +583,24 @@ func (service *HTTPRestService) ReleaseIPConfigsHandler(w http.ResponseWriter, r
 }
 
 func (service *HTTPRestService) removeEndpointState(podInfo cns.PodInfo) error {
```
pjohnst5: Should we just get rid of this function now, all together?

Image

I only see it referenced in tests now (after you switched it out for `releaseIPConfigsWithDeleteIntent()` in `ipam.go`

cns/restserver/endpoint_delete_intent.go
```diff
@@ -0,0 +1,151 @@
...
+			delete(endpointState, containerID)
+			endpointStateChanged = true
+		}
+	}
+
+	if endpointStateChanged {
+		if err := service.EndpointStateStore.Write(EndpointStoreKey, endpointState); err != nil {
+			return fmt.Errorf("replaying endpoint delete intents: %w", err)
+		}
+		service.EndpointState = endpointState
```
pjohnst5: nit: Write a comment here that says "Save it back to memory" or something, just so we know that we were operating on a clone (I know you named it cloned before, but just to be super explicit)

pjohnst5: Same for line 40 up above:
`	service.EndpointDeleteIntents = intents // Save it in-memory`

Idk, maybe it's redundant

cns/restserver/ipam.go
```diff
@@ -73,11 +77,21 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
 
 	// record a pod requesting an IP
 	service.podsPendingIPAssignment.Push(podInfo.Key())
-	podIPInfo, err := requestIPConfigsHelper(service, ipconfigsRequest) //nolint:contextcheck // appease linter for revert PR
```
pjohnst5: This is awesome! `TimedSet`

cns/restserver/ipam.go
```diff
@@ -115,6 +115,41 @@ func (service *HTTPRestService) requestIPConfigHandlerHelper(ctx context.Context
 	}, nil
 }
 
+func (service *HTTPRestService) requestIPConfigsWithEndpointState(ctx context.Context, ipconfigsRequest cns.IPConfigsRequest, podInfo cns.PodInfo) ([]cns.PodIpInfo, error) {
```
pjohnst5, line 118: Okay, (just thinking out loud here), so this is the function that first we check _if_ there's a delete intent for the container that we are trying to request IPs for

If there _is_ a delete intent for it, our "ADD" is blocked

cns/restserver/ipam.go
```diff
@@ -333,67 +368,78 @@ func (service *HTTPRestService) RequestIPConfigsHandler(w http.ResponseWriter, r
 }
 
 func (service *HTTPRestService) updateEndpointState(ipconfigsRequest cns.IPConfigsRequest, podInfo cns.PodInfo, podIPInfo []cns.PodIpInfo) error {
```
pjohnst5: `updateEndpointState()` maybe we should also just get rid of it:

Image

cns/restserver/ipam.go
```diff
@@ -1010,6 +1087,13 @@ func (service *HTTPRestService) AssignDesiredIPConfigs(podInfo cns.PodInfo, desi
 // Assigns an available IP from each NC on the NNC. If there is one NC then we expect to only have one IP return
 // In the case of dualstack we would expect to have one IPv6 from one NC and one IPv4 from a second NC
 func (service *HTTPRestService) AssignAvailableIPConfigs(podInfo cns.PodInfo) ([]cns.PodIpInfo, error) {
```
pjohnst5: Same (maybe just delete?)

Image

pjohnst5 (dismissed):

pjohnst5 (dismissed):
Let's GO!!

cns/restserver/endpoint_delete_intent.go
```diff
@@ -0,0 +1,122 @@
...
+
+	nextIntents := cloneEndpointDeleteIntents(service.EndpointDeleteIntents)
+	nextEndpointState := cloneEndpointState(service.EndpointState)
+	intentsChanged := false
+	endpointStateChanged := false
+	for containerID, intent := range nextIntents {
+		if endpointDeleteIntentExpired(intent, now) {
+			delete(nextIntents, containerID)
+			intentsChanged = true
+			continue
```
behzad-mir, line 90: what if CNS has been restarted or crashed for longer than 10 min, this "continue" means CNS will skip the endpoint deletion since the intent is expired and we will have an stale endpoint. Right?

cns/restserver/ipam.go
```diff
@@ -123,6 +126,13 @@ func (service *HTTPRestService) requestIPConfigsWithEndpointState(ctx context.Co
 	if err := ctx.Err(); err != nil {
 		return nil, fmt.Errorf("ip config request canceled: %w", err)
 	}
+	blocked, err := service.endpointDeleteIntentBlocksAddLocked(podInfo.InfraContainerID(), time.Now())
```
behzad-mir, line 136: This might affect Swiftv2 Rollback process. if swiftv2 getipconfig or interface process fails while the infra part was successful then there will be a rollback call which will be recorded as delete intent. if the runtime retries ADD with same containerid again then this might get blocked for 10 min. probably need to avoid marking internal fail add compensation of swiftv2 as a delete intent.

behzad-mir (changes requested):

PR #4510, opened 2026-07-07
bug: Fixing ETW metadata bug in CNI

apontejaj:
**Reason for Change**:

Commit `09cd3b2c2` ("feat: Enhancing ETW logging for CNI #4426", CNI bundle 1.8.9) added a metadata-enrichment block at the top of `rootExecute()` that calls `common.GetHostMetadata()` before the legacy logger is redirected from stderr to a file. On a metadata-cache miss (cold boot / node reimage), the legacy logger writes `[Telemetry] Request metadata from wireserver` to stderr.

CNS invokes the CNI binary via `cmd.CombinedOutput()` to fetch endpoint state, merging stderr into stdout. The stray log line pollutes the JSON response, causing `json.Unmarshal` to fail with `invalid character '/' after top-level value`, which crashes CNS → CrashLoopBackOff.

This PR moves the legacy logger initialization above the `GetHostMetadata()` call in both CNI entry points so stderr is redirected to a log file before anything can write to it.

**Issue Fixed**:
No issue associated to this PR.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

- **`cni/network/stateless/main.go`** — Reordered: logger redirect now precedes `GetHostMetadata()`
- **`cni/network/plugin/main.go`** — Added legacy `log` import and logger redirect before `GetHostMetadata()` (previously had no legacy logger setup at all)
- Both binaries compile cleanly; on cache miss the `[Telemetry]` line now goes to `azure-vnet.log` instead of stderr

**Tests**:
**ETW logs working correctly**
image

**Failure reproduction**
Overlay BYOCNI cluster created
image

Removing `azuremetadata.json` file
image

Test were performed twice:
a. With `EnableStateMigration:true` and `ManageEndpointState:true` as per IcM
b. With `EnableStateMigration:false` and `ManageEndpointState:false` as per default values

Deploying CNS/CNI v1.8.9 (bad version) in AKS BYOCNI cluster
a. Configmap `EnableStateMigration:true` and `ManageEndpointState:true`
image

a. Error reproduction `EnableStateMigration:true` and `ManageEndpointState:true`
image

b. Configmap  `EnableStateMigration:false` and `ManageEndpointState:false`
image

b. Error reproduction `EnableStateMigration:false` and `ManageEndpointState:false`
image

Test CNI image build/push (acnpublic.azurecr.io/azure-cni:v1.8.9-11-g96e8addde): https://msazure.visualstudio.com/One/Quota%20and%20Capacity%20Management/_build/results?buildId=171170956&view=logs&j=70019b04-dce6-53f1-c9a2-8bfe2dbe6823&t=ec0b9213-fa84-5643-8208-b10307f124c9&l=29

Test CNS image build/push (acnpublic.azurecr.io/azure-cns:v1.8.9-11-g96e8addde): https://msazure.visualstudio.com/One/Quota%20and%20Capacity%20Management/_build/results?buildId=171170956&view=logs&j=58f5182c-d509-5a71-bd8e-7b4908aff171&t=69962e21-bd97-56a3-0e68-084f3e045884&l=29

Error not present anymore
a. `EnableStateMigration:true` and `ManageEndpointState:true`
image

b. `EnableStateMigration:false` and `ManageEndpointState:false`
image

### DIFF
--- a/cni/network/plugin/main.go
+++ b/cni/network/plugin/main.go
@@ -51,6 +51,8 @@ func printVersion() {
 }
 
 func rootExecute() error {
+	var config common.PluginConfig
+
 	// Enrich all CNI loggers with host metadata so ETW events carry VM identity for diagnostics.
 	metadataFile := filepath.Join(os.TempDir(), "azuremetadata.json")
 	if metadata, err := common.GetHostMetadata(metadataFile); err == nil {
@@ -70,8 +72,6 @@ func rootExecute() error {
 			zap.String("os_type", metadata.OsType),
 		)
 	}
-
-	var config common.PluginConfig
 
 	config.Version = version
 
--- a/cni/network/stateless/main.go
+++ b/cni/network/stateless/main.go
@@ -51,6 +51,15 @@ func printVersion() {
 }
 
 func rootExecute() error {
+	var config common.PluginConfig
+
+	log.SetName(name)
+	log.SetLevel(log.LevelInfo)
+	if err := log.SetTargetLogDirectory(log.TargetLogfile, ""); err != nil {
+		fmt.Printf("Failed to setup cni logging: %v\n", err)
+	}
+	defer log.Close()
+
 	// Enrich all CNI loggers with host metadata so ETW events carry VM identity for diagnostics.
 	metadataFile := filepath.Join(os.TempDir(), "azuremetadata.json")
 	if metadata, err := common.GetHostMetadata(metadataFile); err == nil {
@@ -70,15 +79,6 @@ func rootExecute() error {
 			zap.String("os_type", metadata.OsType),
 		)
 	}
-
-	var config common.PluginConfig
-
-	log.SetName(name)
-	log.SetLevel(log.LevelInfo)
-	if err := log.SetTargetLogDirectory(log.TargetLogfile, ""); err != nil {
-		fmt.Printf("Failed to setup cni logging: %v\n", err)
-	}
-	defer log.Close()
 
 	config.Version = version
 	config.Stateless = stateless
--- a/common/utils.go
+++ b/common/utils.go
@@ -178,8 +178,6 @@ func GetHostMetadata(fileName string) (Metadata, error) {
 		}
 	}
 
-	log.Printf("[Telemetry] Request metadata from wireserver")
-
 	req, err := http.NewRequest("GET", metadataURL, nil)
 	if err != nil {
 		return Metadata{}, err


cni/network/plugin/main.go
```diff
@@ -51,6 +52,17 @@ func printVersion() {
...
+	var config common.PluginConfig
+
+	// The legacy GetHostMetadata function uses the log package to log errors.
+	// Logging needs to be set up before calling it to avoid issues.
+	log.SetName(name)
+	log.SetLevel(log.LevelInfo)
+	if err := log.SetTargetLogDirectory(log.TargetLogfile, ""); err != nil {
+		fmt.Printf("Failed to setup cni logging: %v\n", err)
+	}
+	defer log.Close()
```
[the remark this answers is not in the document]
apontejaj: Not implementing.

cni/network/stateless/main.go
```diff
@@ -51,6 +51,15 @@ func printVersion() {
...
 
 func rootExecute() error {
+	var config common.PluginConfig
+
+	log.SetName(name)
+	log.SetLevel(log.LevelInfo)
+	if err := log.SetTargetLogDirectory(log.TargetLogfile, ""); err != nil {
+		fmt.Printf("Failed to setup cni logging: %v\n", err)
+	}
+	defer log.Close()
```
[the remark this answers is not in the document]
apontejaj: Not implementing.

behzad-mir (dismissed):

mugeshsp (dismissed):

behzad-mir (approved):

cni/network/plugin/main.go
```diff
@@ -51,6 +52,17 @@ func printVersion() {
 }
 
 func rootExecute() error {
+	var config common.PluginConfig
+
+	// The legacy GetHostMetadata function uses the log package to log errors.
+	// Logging needs to be set up before calling it to avoid issues.
+	log.SetName(name + "-legacy")
```
behzad-mir: WHy are we changing the log name here?

behzad-mir (changes requested):

cni/network/plugin/main.go
behzad-mir: We need to have the same file format and azure-vnet.log

apontejaj: I've address your comment.
I've done a small refactor to the the `GetHostMetadata` function so it takes an optional logger.

Pipeline is passing now and the manual test on a BYOCNI cluster is also showing the correct CombinedOutput even after deleting the `azuremetada.json` file from the node.

cni/network/plugin/main.go
```diff
@@ -51,9 +51,11 @@ func printVersion() {
 }
 
 func rootExecute() error {
+	var config common.PluginConfig
+
 	// Enrich all CNI loggers with host metadata so ETW events carry VM identity for diagnostics.
 	metadataFile := filepath.Join(os.TempDir(), "azuremetadata.json")
-	if metadata, err := common.GetHostMetadata(metadataFile); err == nil {
+	if metadata, err := common.GetHostMetadata(metadataFile, logger); err == nil {
```
behzad-mir: is logger initiated by the time it is passed to this function?

apontejaj: Yes, logger is a package-level variable initialized at load time in both `plugin/main.go` and `stateless/main.go` files.
```
var logger = zaplog.CNILogger.With(zap.String("component", "cni-main"))
```

And CNILogger is itself a package-level variable in `logger.go`
```
var CNILogger = initZapLog(zapCNILogFile)
```

Go guarantees package-level variables are initialized before any function in the package runs. By the time `rootExecute()` is called from `main()`, `logger` is fully initialized.

behzad-mir: I recommend you log the line in here rather than pass the logger to the GetHostMetadata. If we only log one line in that function, I don't see a reason to pass the logger to it. WE can just get rid of that line in the GetHostMetadata.

apontejaj: I didn't think that removing the line from `GetHostMetadata` was a good idea, since it's part of legacy code that I'm not sure who else might be calling.

But that's OK. I'll remove the line from the `GetHostMetadata` function. 

Now, there is not point adding the line in any of the `main.go` files, since a) that only happens when the temporal files are not present, which is not the most common case, and b) I don't see a reason to add lines of logs in the other callers.

This is addressed now.

behzad-mir (approved):

behzad-mir (approved):

tamilmani1989 (approved):

PR #4514, opened 2026-07-08
fix(azure-iptables-monitor): build with Go 1.25 to remediate CVE-2026-39825

rejain789:
## Summary

The published `azure-iptables-monitor` image (`mcr.microsoft.com/containernetworking/azure-iptables-monitor`) ships the `iptables-monitor` and `azure-block-iptables` binaries built with **Go stdlib 1.24.13**, which is affected by **CVE-2026-39825** (gobinary, published 2026-05-07). The fix is available in Go **1.25.10+** / **1.26.3+**.

This PR bumps **only** this component's Go builder to the **1.25** line, leaving the shared `GO_IMG` (1.24) and all other components unchanged.

## Why this scope works

`.pipelines/build/scripts/install-go.sh` resolves the host Go toolchain **per-component** by reading the `FROM ... golang ... AS go` line from `${name}/Dockerfile`. So pinning `azure-iptables-monitor/Dockerfile` to Go 1.25 fixes **both** build paths:
- **Pipeline (production image):** `install-go.sh` (name=`azure-iptables-monitor`) installs Go 1.25, then `.pipelines/build/scripts/azure-iptables-monitor.sh` compiles the binaries.
- **Local/dev:** the component `Dockerfile` `FROM go` stages compile with Go 1.25 directly.

Precedent: `npm` is already pinned to Go 1.25 independently of the shared image.

## Version pinned

`mcr.microsoft.com/oss/go/microsoft/golang:1.25-azurelinux3.0` → digest `sha256:a5480db3...` = **Go 1.25.12** (≥ 1.25.10 ✓).

## FIPS / crypto handling

Per the MS Go FIPS docs, **since Go 1.25 `systemcrypto` (cgo/OpenSSL) is the default backend**. With `CGO_ENABLED=0` and crypto stdlib dependencies (both binaries import `crypto/*`), the build now **fails** unless systemcrypto is disabled — the cgo-less OpenSSL backend does not exist until Go 1.26 (`ms_nocgo_opensslcrypto`) / 1.27 (default).

We set **`MS_GO_NOSYSTEMCRYPTO=1`** on every `CGO_ENABLED=0` `go build`/`go generate` for this image's binaries (mirroring the existing `npm` Dockerfiles). This preserves the exact pure-Go, static-binary crypto posture the component already had on Go 1.24 — no FIPS regression, just patched stdlib.

## Files changed (azure-iptables-monitor only)

| File | Change |
|---|---|
| `azure-iptables-monitor/Dockerfile.tmpl` | Pin Go builder to 1.25; add `MS_GO_NOSYSTEMCRYPTO=1` to CGO=0 build/generate |
| `azure-iptables-monitor/Dockerfile` (generated) | Regenerated to match template |
| `.pipelines/build/scripts/azure-iptables-monitor.sh` | `export MS_GO_NOSYSTEMCRYPTO=1` (production build path) |
| `Makefile` | `MS_GO_NOSYSTEMCRYPTO=1` on `azure-iptables-monitor-binary` and `azure-block-iptables-binary` |

## Validation

- `install-go.sh` correctly parses the new pinned image from the Dockerfile.
- Rendered template == committed `Dockerfile` (baseimages `make dockerfiles` diff check passes).
- Smoke build of `iptables-monitor` with `MS_GO_NOSYSTEMCRYPTO=1 CGO_ENABLED=0` succeeds.

## Notes

- The shared `GO_IMG` remains 1.24; the go-version-check digest-refresh automation targets the shared image only and won't touch this pin. When the repo moves to 1.25/1.26 repo-wide, this component should be re-templated back onto the shared `{{.GO_PIN}}`.

### DIFF
--- a/.pipelines/build/scripts/azure-iptables-monitor.sh
+++ b/.pipelines/build/scripts/azure-iptables-monitor.sh
@@ -5,6 +5,9 @@ set -eux
 FILE_EXT=''
 
 export CGO_ENABLED=0
+# MS Go 1.25 makes systemcrypto (cgo/OpenSSL) the default backend; with CGO_ENABLED=0
+# and crypto stdlib deps the build fails, so disable it to keep the pure-Go static build.
+export MS_GO_NOSYSTEMCRYPTO=1
 export C_INCLUDE_PATH=/usr/include/bpf
 
 mkdir -p "$OUT_DIR"/bin
--- a/Makefile
+++ b/Makefile
@@ -207,8 +207,8 @@ endif
 
 # Build the azure-block-iptables binary.
 azure-block-iptables-binary: bpf-lib
-	cd $(AZURE_BLOCK_IPTABLES_DIR) && CGO_ENABLED=0 go generate ./...
-	cd $(AZURE_BLOCK_IPTABLES_DIR)/cmd/azure-block-iptables && CGO_ENABLED=0 go build -v -o $(AZURE_BLOCK_IPTABLES_BUILD_DIR)/azure-block-iptables$(EXE_EXT) -ldflags "-X main.version=$(AZURE_BLOCK_IPTABLES_VERSION)" -gcflags="-dwarflocationlists=true"
+	cd $(AZURE_BLOCK_IPTABLES_DIR) && MS_GO_NOSYSTEMCRYPTO=1 CGO_ENABLED=0 go generate ./...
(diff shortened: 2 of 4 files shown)


timraymond (approved):
I'm pretty sure this is a false positive CVE since it deals with reverse proxies. `azure-iptables-monitor` doesn't appear to directly use `httputil.ReverseProxy` at all, and it's doubtful that any of its dependencies do in the way that they're used.  I've found these scanners have a tendency to cry wolf a lot and don't actually look at which packages are used. I'm curious what `govulncheck` has to say about this one (which does the right thing).

That said though, upgrading to Go 1.25 is a good thing regardless, so I support this.

PR #4525, opened 2026-07-13
fix: skip aks-preview install if already present to avoid checksum failures

behzad-mir:
## Problem
The `az extension add --name aks-preview` command fails with:
```
ERROR: The checksum of the extension does not match the expected value.
```
This blocks all integration tests that create clusters.

## Fix
- Skip install if `aks-preview` is already present on the agent
- Tolerate `az extension update` failures gracefully (CDN propagation issues)

## Change
```makefile
# Before
@$(AZCLI) extension add --name aks-preview --yes
@$(AZCLI) extension update --name aks-preview

# After
@$(AZCLI) extension show --name aks-preview &>/dev/null || $(AZCLI) extension add --name aks-preview --yes
@$(AZCLI) extension update --name aks-preview || true
```

### DIFF
--- a/hack/aks/Makefile
+++ b/hack/aks/Makefile
@@ -86,8 +86,8 @@ azlogin: ## Login and set account to $SUB
 	@$(AZCLI) account set -s $(SUB)
 
 azcfg: ## Set the $AZCLI to use aks-preview
-	@$(AZCLI) extension add --name aks-preview --yes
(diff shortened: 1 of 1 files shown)


behzad-mir:
Closing in favor of #4526 which provides a more comprehensive fix (fallback version pinning + centralizes all aks-preview installs through make target).

PR #4531, opened 2026-07-14
feat: Default Deny for Non Swift pods

kaustubh0x77:
**Reason for Change**:
Add default deny support for non swift pods on L1VH.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


rejain789:
pr looks good - were you able to test this end to end via creating a pod with this label and ensuring that pod can't communicate with another pod and vice versa?

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -243,3 +251,35 @@ func (k *K8sSWIFTv2Middleware) getInfraRoutes(podIPInfo *cns.PodIpInfo) ([]cns.R
...
+		logger.Errorf("default-deny label check: failed to get pod %s: %v", nsName, err)
+		return
+	}
+	if _, ok := pod.Labels[configuration.LabelPodDefaultDeny]; !ok {
+		return
+	}
+	for i := range resp.PodIPInfo {
+		ipInfo := &resp.PodIPInfo[i]
+		// Match InfraNIC explicitly or the legacy/empty NICType, mirroring cns.NICType.IsInfraOrLegacy().
+		if ipInfo.NICType == cns.InfraNIC || ipInfo.NICType == "" {
```
pjohnst5, line 295: Why the "if NICType == "" " ?

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -243,3 +251,35 @@ func (k *K8sSWIFTv2Middleware) getInfraRoutes(podIPInfo *cns.PodIpInfo) ([]cns.R
...
+			ipInfo.EndpointPolicies = append(ipInfo.EndpointPolicies, defaultDenyEgressPolicy, defaultDenyIngressPolicy)
```
pjohnst5, line 296: What testing was done in general for this change? (What must we do to make sure it works for NPM-Lite?)

kaustubh0x77:
@microsoft-github-policy-service agree [company="Microsoft"]

kaustubh0x77:

cns/middlewares/k8sSwiftV2_windows.go
pjohnst5: This is how we handle the nic type elsewhere (see https://github.com/Azure/azure-container-networking/commit/2ab9cfe8236150147b2f551700a9d9e55585f236)

cns/middlewares/k8sSwiftV2_windows.go
pjohnst5: Kaustubh ran this on an NPM-lite enabled cluster, tested egress and ingress, failed, created allow-list network policy, traffic could then get through

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -243,3 +251,35 @@ func (k *K8sSWIFTv2Middleware) getInfraRoutes(podIPInfo *cns.PodIpInfo) ([]cns.R
...
+			logger.Printf("[DefaultDenyLabel] applied default-deny ACLs to InfraNIC for pod %s", nsName)
+			return
```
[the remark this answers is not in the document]
pjohnst5: This is fine, we will need to see when this is enabled on pods to make sure it is working

cns/middlewares/k8sSwiftV2_windows.go
```diff
@@ -130,7 +130,7 @@ func createEndpointPolicy(direction string) ([]byte, error) {
 // and release IP configs handlers.
 func (k *K8sSWIFTv2Middleware) IPConfigsRequestHandlerWrapper(defaultHandler, failureHandler cns.IPConfigsHandlerFunc) cns.IPConfigsHandlerFunc {
 	return func(ctx context.Context, req cns.IPConfigsRequest) (*cns.IPConfigsResponse, error) {
-		podInfo, respCode, message := k.GetPodInfoForIPConfigsRequest(ctx, &req)
+		pod, podInfo, mtpnc, respCode, message := k.GetPodAndNetworkingCrds(ctx, &req)
```
pjohnst5: Instead of changing this function, I think we should just call `k.GetPodInfo` after it:

```suggestion
		podInfo, respCode, message := k.GetPodInfoForIPConfigsRequest(ctx, &req)
		if respCode != types.Success {
			return &cns.IPConfigsResponse{
				Response: cns.Response{
					ReturnCode: respCode,
					Message:    message,
				},
			}, errors.New("failed to validate IP configs request")
		}
        podInfo, pod, respCode, message := k.GetPodInfo(ctx, req)
	        if respCode != types.Success {
		        return nil, respCode, message
	        }
```

pjohnst5: That way, we are touching less stuff (the `GetPodInfoForIPConfigsRequest` which is also used in linux (even though, we only care about windows in this case)

pjohnst5: If anything, in a future PR, we should fix `GetPodInfoForIPConfigsRequest` to *take in* `pod` as a parameter, but that is outside the scope of this PR

pjohnst5 (commented):
Hi Kaustubh, I think we should change this

pjohnst5 (dismissed):

pjohnst5 (approved):

kaustubh0x77:
Validation done - 
1. Create non swift pod with default deny label - 
2. Test ping to a pod on a different node - fails as expected
3. Create NP allowing connectivity to 1 pod 
3. Test ping to the same pod - ping succeeds
4. Remove NP
5. Test ping to the same pod - fails as expected

csfmomo (approved):

PR #4555, opened 2026-07-22
ci: install cluster networking during creation

jpayne3506:
## What this changes

- create clusters and install networking sequentially in the same create-stage job
- use shared `create-cluster-steps.yaml` and `network-install.yaml` step templates
- migrate all single-tenancy Azure CNI and Cilium scenarios
- support standard and OneBranch entrypaths, Linux/Windows, Hubble, dual-stack, eBPF, pod-subnet, and node-subnet
- preserve downstream load, restart, state, connectivity, Kubernetes E2E, diagnostics, and failure-analysis tests without reinstalling networking

## Why

When E2E agent pools are saturated, a new BYOCNI cluster can wait 20-30 minutes before its network installation job starts. During that gap, nodes remain without networking and can restart. Keeping network installation on the cluster-create agent removes that gap.

## Validation

- migrated 23 single-tenancy create jobs
- no direct single-tenancy callers of the old create job templates remain
- parsed all 33 changed Azure Pipeline YAML files
- `git diff --check` passes
- verified later E2E templates retain validation but no longer deploy components or reinstall CNS
- independently reviewed standard/OneBranch dependencies, image variables, Windows nodepool ordering, and Cilium-before-CNS sequencing


jpayne3506:
Superseded by https://github.com/Azure/azure-container-networking/pull/4558, which uses a branch directly in Azure/azure-container-networking.

PR #4563, opened 2026-07-23
feat(platform): add boot identity helpers

rbtr:
## Summary

- add an uncalled cross-platform `platform.BootID()` helper for future CNS state lifecycle work
- read Linux boot identity from `/proc/sys/kernel/random/boot_id`
- read the kubelet-compatible Windows `BootId` registry DWORD through native Go APIs
- keep OS-specific behavior isolated in `_linux.go` and `_windows.go` files

## Scope

This is an additive prerequisite only. It has no runtime caller and does not change CNS behavior.

## Validation

- `go test -race ./platform`
- `go vet ./platform`
- Windows platform test cross-compilation
- independent code review found no correctness or platform issues

### DIFF
--- a/platform/boot_id_linux.go
+++ b/platform/boot_id_linux.go
@@ -1,0 +1,34 @@
+// Copyright 2017 Microsoft. All rights reserved.
+// MIT License
+
+package platform
+
+import (
+	"errors"
+	"fmt"
+	"os"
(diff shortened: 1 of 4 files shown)


jpayne3506 (dismissed):
lgtm. ping for reapproval after lint fix

PR #4573, opened 2026-07-23
deps: remediate reachable Go vulnerabilities

rbtr:
## What this does

Updates dependency graphs for the root module, azure-ipam, azure-iptables-monitor, and tools/azure-npm-to-cilium-validator to fixed releases for the current reachable advisories.

Key floors are:

- github.com/cilium/cilium v1.17.16
- golang.org/x/net v0.56.0
- golang.org/x/text v0.39.0

Required x/crypto, x/sync, x/sys, x/term, and x/tools transitive upgrades follow from MVS. No dependency is downgraded.

## Validation

- go mod tidy and go mod verify for all four modules
- complete offline module-graph loading
- component module builds and tests
- govulncheck reports no reachable target vulnerabilities
- root scan reports zero called vulnerabilities after excluding pre-existing missing-BPF-codegen packages
- independent dependency review passed


behzad-mir:
@dependabot recreate

PR #4574, opened 2026-07-23
fix: [NPM] make policy-owned IPSet names unique per policy and namespace

rayaisaiah:
**Reason for Change**:

Two IPSet name generators joined variable fields with a delimiter that is also
legal inside those fields, so distinct inputs could serialize to the same
pre-hash name and end up sharing a single kernel IPSet:

- **ipblock CIDR sets** used `"%s-in-ns-%s-%d-%d%s"`. Since `-in-ns-` is a valid
  substring of a DNS label, two different `(policy, namespace)` pairs could
  produce the same name.
- **multi-value pod-selector nested sets** used `"%s-%s"` (`policyKey`, `matchKey`).
  Since `-` is valid in both, two different `(policy, matchKey)` pairs in the same
  namespace could produce the same name.

This separates the variable fields with `:`, which is not a valid character in a
Kubernetes name, namespace, or label key, so distinct inputs always produce
distinct set names:

```
ipblock:  "%s:in-ns:%s-%d-%d%s"   // <policy>:in-ns:<ns>-<setIndex>-<peerIndex><direction>
nested:   "%s:%s"                 // <policyKey>:<matchKey>
```

`:` was chosen over length-prefixing because the set name also appears in iptables
comments (255-byte limit), so keeping the name short matters. The translated sets
and their ACL / selector references come from the same generators, so references
stay consistent; nested child member sets are unchanged. It also hardens
`IPSetManager.AddReference` so a CIDR ipblock set is only referenced by a single
NetworkPolicy. Existing expected-name literals are updated and regression tests are
added for both generators. Scoped to the v2 translation path (the only enabled NPM
datapath).

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/) (`fix:`)
- [x] includes documentation (inline rationale on both format generators)
- [x] adds unit tests (`TestIPBlockSetNameUnambiguous`, `TestNestedSetNameUnambiguous`, `TestIPBlockTwoPolicyIsolation`, `TestNestedSelectorTwoPolicyIsolation`, `TestAddReferenceCIDRSingleOwner`, `TestAddReferenceCIDROwnerReleasedOnDelete`, `TestCIDRSetsDoNotMergeInManager`; existing literals updated)

**Notes**:

- `go test ./npm/...` passes except two environment-only failures that also fail on
  the base branch: `npm/cmd` (`iptables-nft-save` needs root) and `controllers/v1`
  (Prometheus metric registration) — neither is related to this change.
- Both name changes only affect the pre-hash set name, which is hashed to a
  fixed-length kernel name, so iptables chain names are unchanged. NPM rebuilds all
  IPSets and ACL references from the NetworkPolicy objects on startup, so the
  renamed sets reconcile on a pod restart.

### DIFF
--- a/npm/pkg/controlplane/translation/parseSelector.go
+++ b/npm/pkg/controlplane/translation/parseSelector.go
@@ -278,8 +278,11 @@ func parsePodSelector(policyKey string, selector *metav1.LabelSelector) ([]label
 				setType = ipsets.KeyValueLabelOfPod
 			} else {
 				// "(!) + matchKey + : + multiple matchVals" case
-				// see caveat in definition of TranslatedIPSet for why the policy key must be included in the set name
-				setName = fmt.Sprintf("%s-%s", policyKey, req.Key)
+				// The policy key must be included so each policy owns its own nested list
+				// (see caveat in TranslatedIPSet). Separate the policy key and match key
+				// with ":", which neither can contain, so distinct (policy, key) pairs
+				// always produce distinct set names.
+				setName = fmt.Sprintf("%s:%s", policyKey, req.Key)
 				for _, val := range req.Values {
 					setName = util.GetIpSetFromLabelKV(setName, val)
 					members = append(members, util.GetIpSetFromLabelKV(req.Key, val))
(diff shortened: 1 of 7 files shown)


matmerr (approved):

rayaisaiah:
Forward-ported to `master` in #4611 (single PR forward-porting #4574, #4587, #4593 — cherry-picked cleanly with zero conflicts).

PR #4590, opened 2026-07-27
test: parameterize LRP test resources per node pool

vipul-21:
Make the LRP integration test runnable per-node-pool in parallel without resource-name collisions or cross-pool node selection. All behavior is gated on optional env vars; when unset the test behaves exactly as before.

- Template metadata.name in lrp.yaml, node-local-dns-ds.yaml and client-ds.yaml with a ${POOL} suffix, rendered by the test to "-<pool>" when POOL is set and "" otherwise.
- Add an optional nodeSelector (${POOL_NODESELECTOR}) to both DaemonSet pod specs; empty renders to "{}" (no constraint).
- Add a renderManifest helper that always substitutes template tokens (so ${POOL}/${POOL_NODESELECTOR} never leak into applied manifests), replacing the previous one-off node-local-dns temp-file rewrite.
- Restrict the candidate node list to the pool via GetNodeListByLabelSelector when POOL_LABEL is set, falling back to the full node list otherwise.

### DIFF
--- a/test/integration/lrp/lrp_test.go
+++ b/test/integration/lrp/lrp_test.go
@@ -46,7 +46,6 @@ const (
 var (
 	defaultRetrier                 = retry.Retrier{Attempts: retryAttempts, Delay: retryDelay}
 	nodeLocalDNSDaemonsetPath      = ciliumManifestsDir + "node-local-dns-ds.yaml"
-	tempNodeLocalDNSDaemonsetPath  = ciliumManifestsDir + "temp-daemonset.yaml"
 	nodeLocalDNSConfigMapPath      = ciliumManifestsDir + "config-map.yaml"
 	nodeLocalDNSServiceAccountPath = ciliumManifestsDir + "service-account.yaml"
 	nodeLocalDNSServicePath        = ciliumManifestsDir + "service.yaml"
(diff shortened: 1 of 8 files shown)


camrynl (dismissed):

test/integration/lrp/lrp_test.go
```diff
@@ -481,17 +500,20 @@ func deleteAndRecreateResources(t *testing.T, ctx context.Context, cs *k8sclient
 	require.NoError(t, err)
 
 	// Recreate LRP
-	_, cleanupLRP := kubernetes.MustSetupLRP(ctx, ciliumCS, lrpPath)
+	_, cleanupLRP := kubernetes.MustSetupLRP(ctx, ciliumCS, lrpPath, func(l *ciliumv2.CiliumLocalRedirectPolicy) { l.Name += suffix })
```
jpayne3506, line 503: Wonder how this would error out.

jpayne3506 (approved):

test/integration/lrp/lrp_test.go
vipul-21: mustSetupLRP panics out if there's a malformed manifest.

camrynl (approved):

PR #4595, opened 2026-07-27
hack/aks: add LTS=auto to auto-detect AKS Long Term Support

vipul-21:
## Problem

`hack/aks/Makefile`'s `LTS` flag was **true/false only**. So every caller (OneBranch pipelines, other pipelines, local dev) had to compute AKS Long Term Support eligibility *itself* and pass `LTS=true/false` in. That detection is easy to get wrong — e.g. the cilium-private pipeline was grepping AKS patch versions for the legacy `.100` convention, which today only matches `1.31.100` and silently misclassified LTS-only minors like `1.32`/`1.33` as non-LTS. Since those minors have left community support, `az aks create` **requires** `--k8s-support-plan AKSLongTermSupport --tier premium`, so the create would fail.

## Change

Add an `auto` mode to `LTS`, **and make it the default**:

```make
LTS ?= auto             # auto | true | false
```

`auto` queries `az aks get-versions` for the requested `K8S_VER`'s support plan in `REGION` and enables the LTS support-plan + premium tier **only for LTS-only minors** — those whose `supportPlan` contains `AKSLongTermSupport` but **not** `KubernetesOfficial`. Still-supported minors (which also carry `KubernetesOfficial`) stay on the standard tier, so `auto` never forces Premium unnecessarily.

The signal is the version's `capabilities.supportPlan` (now populated for every version by AKS), not the patch-number shape — so it handles both the old `.100` numbering (`1.31.100`) and current sequential patches (`1.32.11`, `1.33.13`).

## ⚠️ Default behavior change (for maintainers)

The default flips from `false` to `auto`. **No currently-green path regresses:**

- Still-supported minors (e.g. `1.34`+) resolve to the **standard tier** under `auto` — identical `az aks create` args to today's `LTS=false` default.
- LTS-only minors (`1.31`/`1.32`/`1.33`) that **fail** to create today (missing required LTS flags) now **succeed** — the whole point.
- Cost is **one** extra `az aks get-versions` call per create (only when `auto` is in effect).
- `LTS=false` remains as an escape hatch (force standard tier / fully offline), and `LTS=true` forces LTS on.

### Design notes
- `LTS_ARGS` stays a **recursively-expanded** variable, so the `az` call is **deferred to recipe time**. `make help`, offline invocations, and `LTS=false`/`LTS=true` never invoke `az`. A cluster create with `auto` costs exactly **one** `az aks get-versions` call.
- The requested minor is derived with Make string functions (`$(word)`/`$(subst)`), so the `$(shell ...)` az query needs no shell variables and no Make `$$` escaping.
- Uses `$(AZCLI)` (consistent with the rest of the Makefile), so the dockerized-az default and `AZCLI=az` override both work.

## Validation

Local `GNU Make 4.3`, via a probe target `@echo "[$(LTS_ARGS)]"`:

| invocation | az calls | result |
|---|---|---|
| `help` (default auto) | 0 | (deferred — target doesn't use LTS_ARGS) |
| `K8S_VER=1.33` (default auto) | 1 | `--k8s-support-plan AKSLongTermSupport --tier premium` |
| `K8S_VER=1.34` (default auto) | 1 | standard |
| `LTS=false K8S_VER=1.33` | 0 | standard |
| `LTS=true K8S_VER=1.34` | 0 | LTS |
| `LTS=auto K8S_VER=1.32.100` | 1 | LTS (normalized to minor `1.32`) |

Confirmed against **real** `az aks get-versions` (eastus2): `1.33`→LTS, `1.34`→standard.

## Follow-up

With `auto` as the default, callers can drop their bespoke LTS detection entirely and no longer need to pass anything (e.g. cilium-private's `cluster-create.yaml`, which currently greps patch versions).


vipul-21:
@copilot resolve the merge conflicts in this pull request

jpayne3506 (approved):

rbtr (approved):

